PR #104, opened 2021-05-20
Fix wizard responsive design issues

FlyersPh9:
- All step descriptions have a set width of `20vw`.  In React we will override this to take the width of the longest step name & set that as the max-width for all step names.
- Description text is centered if bumped onto a second line.
- Circle remains centered on the track if multiple lines are introduced to the description.

Screen Recording 2021-05-20 at 4 41 36 PM

Closes #77 if we are ok with handling more narrow screens in a later PR.  In #109 we can do the class name refactor, better responsive design, and maybe offer a new layout for phone wide screens like below.
Screen Shot 2021-05-20 at 4 44 54 PM

PR #109, opened 2021-05-21
refactor: Wizard refactor

FlyersPh9:
- Built upon #104.
- HTML restructured like [this guide](https://csslayout.io/patterns/wizard) to use flex better.
- Fewer classes, track is now implemented using `::before` and `::after`.
- Removes `z-index` and `!important`.
- Added aria properties.
- `.iui-long` and `.iui-workflow` are modifiers applied to `.iui-wizard`.


src/wizard/wizard.scss
```diff
@@ -2,164 +2,142 @@
...
         }
       }
 
-      > .iui-wizards-step-track {
-        @include themed {
-          background-color: t(iui-color-foreground-positive) !important;
-        }
+      &:first-child .iui-track:first-child,
+      &:last-child .iui-track:last-child {
+        background-color: transparent;
```
mayank99, line 140: Well this isn't very nice
image

I would think those tracks should just not be present in the html. But then it messes up the centering.

What if we move those tracks outside the steps? I think it would even simplify the css and you wouldn't need the `.iui-idk` class 😄

backstop/tests/wizard.html
```diff
@@ -29,192 +29,201 @@
...
-            <span
-              class="iui-wizards-step-circle"
-              tabindex="0"
-            >
-              <span>2</span>
-            </span>
+      <div class="iui-wizard">
+        <nav
+          class="iui-wizard-container"
+          aria-label="Progress"
```
mayank99, line 39: When everything else gets finalized, we should consider adding some aria attributes.

I would expect `role=tablist` on the wizard, `role=tab` on the steps, `aria-selected` on the current step, and maybe `role=tabpanel` on the content.

src/wizard/wizard.scss
FlyersPh9: I don't think this is that bad.  It helps keep each step set to an equal width no matter the text underneath.  We want to avoid situations like this where a very long label VS short label off-centers the entire wizard.
Screen Shot 2021-05-25 at 11 40 23 AM

1 option I thought of but think is not a good idea is having the text labels be within their own flex box.  That way the first / last step's circle could go all the way to the edge.  However, then you'll end up with something like this.
Screen Shot 2021-05-25 at 11 47 36 AM

BTW the wizard refactor inspiration came from [here](https://csslayout.io/patterns/wizard).

backstop/tests/wizard.html
```diff
@@ -33,175 +33,247 @@
...
-            <span class="iui-wizards-step-title">Completed step</span>
-            <span class="iui-wizards-step-track iui-wizards-step-track-after"></span>
-            <span
-              class="iui-wizards-step-circle"
-              tabindex="0"
-            >
-              <span>2</span>
-            </span>
+      <div class="iui-wizard">
+        <nav
```
bentleyvk, line 41: Can we really use `<nav>` here? It is meant to be used for navigation, also W3 is using `<ol>` in their examples https://www.w3.org/WAI/tutorials/forms/multi-page/#using-the-html5-progress-element

src/wizard/wizard.scss
bentleyvk: But when I start to think, it really simplifies React implementation 😃

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
     margin-bottom: $iui-baseline * 0.5;
 
-    > .iui-wizards-step {
+    .iui-wizard-step {
+      flex: 1 1 0;
       align-items: center;
       display: flex;
-      flex-direction: column-reverse;
+      flex-direction: column;
       font-size: $iui-font-size-small;
```
bentleyvk, line 22: Maybe we have a mixin for font stuff?

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
       font-weight: $iui-font-weight-normal;
       line-height: $iui-line-height;
       position: relative;
```
bentleyvk, line 25: Seems it is not needed.

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
-        z-index: 1;
-        user-select: none;
-        @include themed {
-          border: 1px solid t(iui-color-background-5);
-          background-color: t(iui-color-background-1);
-          color: t(iui-text-color-muted);
+        .iui-circle {
+          align-items: center;
+          border-radius: 50%;
+          display: flex;
```
bentleyvk, line 46: Can you check whether all these flex rules are needed?

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
-      position: absolute;
-      top: $iui-l * 0.5;
-      z-index: 0;
 
-      &.iui-wizards-step-track-before,
-      &.iui-wizards-step-track-after,
-      &.iui-wizards-step-track-main {
+      .iui-step-name {
+        text-align: center;
+        width: 100%;
```
bentleyvk, line 67: Seems to do nothing.

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
-        font-weight: $iui-font-weight-semibold;
-      }
-
-      > .iui-wizards-step-circle {
-        @include themed {
-          color: t(iui-color-foreground-accessory);
-          border: 1px solid t(iui-color-background-positive);
-          background-color: t(iui-color-background-positive);
+  
+        .iui-before {
```
bentleyvk, line 91: I couldn't found it being used.

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
...
+      flex-direction: column;
       font-size: $iui-font-size-small;
       font-weight: $iui-font-weight-normal;
       line-height: $iui-line-height;
       position: relative;
-      width: 20%;
-
       @include themed {
         color: t(iui-color-background-positive);
       }
```
bentleyvk, line 29: Should we try to reduce nesting? E.g. `.iui-step-track` is 4 levels deep.

src/wizard/wizard.scss
```diff
@@ -2,168 +2,147 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 
-@mixin iui-wizards {
+@mixin iui-wizard {
   @include iui-reset;
   @include iui-font-family;
 
-  > .iui-wizards-wrapper {
+  > .iui-wizard-container {
```
bentleyvk, line 9: Hm, is this class needed? Can't it be merget with the root one?

src/wizard/workflow.scss
```diff
@@ -34,17 +28,11 @@
       }
     }
 
-    > .iui-wizards-step-track,
-    > .iui-wizards-step > .iui-wizards-step-track {
-      &.iui-wizards-step-track-before,
```
bentleyvk, line 39: This is nice!

backstop/tests/wizard.html
FlyersPh9: Swapped from `<nav>` & `<span>` to `<ol>` & `<li>` in the next push.

src/wizard/wizard.scss
FlyersPh9: Replaced `font-size`, `font-weight`, and `line-height` with `@include iui-text-small;` in the next push.

src/wizard/wizard.scss
FlyersPh9: `position: relative;` is removed in the next push.

src/wizard/wizard.scss
FlyersPh9: `width: 100%;` removed in next push.

src/wizard/wizard.scss
FlyersPh9: `.iui-before` is removed in the next push.

src/wizard/wizard.scss
FlyersPh9: I don't believe the container can be merged with the parent.  The parent contains an `<ol class="iui-wizard-container">` (in red) as well as `<p class="iui-wizard-steps">` (in pink).
Screen Shot 2021-08-04 at 2 02 56 PM

src/wizard/wizard.scss
FlyersPh9: I tried reducing the nesting in my latest push.  Let me know if you think I should keep reducing or if this is good enough.  cc @mayank99

src/wizard/wizard.scss
FlyersPh9: I was able to remove the flex rules and instead use line-height to center the text within the circle.

src/wizard/wizard.scss
mayank99: When in doubt, just look at the compiled css.
image

Seems you should be able to remove the top level nesting (i.e. `.iui-wizard`) from all selectors.
After that, I don't think the `.iui-step-name` needs to be inside `.iui-wizard-step`. And `.iui-step-track` does not need to be inside `.iui-wizard-step`.

And so on...

src/wizard/wizard.scss
mayank99: line-height is worse tho

FlyersPh9: Why is line-height worse?

mayank99: It's less predictable and will randomly stop working if you change something.

src/wizard/wizard.scss
```diff
@@ -60,110 +59,79 @@
...
       }
     }
+
+    &:first-child .iui-track:first-child,
+    &:last-child .iui-track:last-child {
+      background-color: transparent;
+    }
+  }
+
+  .iui-wizard-steps {
```
bentleyvk, line 134: I don't think this one is needed, maybe just don't add steps where they are not needed.

src/wizard/wizard.scss
```diff
@@ -1,57 +1,56 @@
...
 
-      > .iui-wizards-step-circle {
-        align-items: center;
+      .iui-circle {
         border-radius: 50%;
-        display: flex;
-        height: $iui-l;
         width: $iui-l;
-        justify-content: center;
+        height: $iui-l;
```
bentleyvk, line 46: Is this a valid scenario? 😃 If yes, then we need to remove height in workflow and add `word-break: break-word`

image

src/wizard/wizard.scss
bentleyvk: Missed that part.

src/wizard/wizard.scss
bentleyvk: Should we add `iui-wizard` prefix to each class here?

src/wizard/wizard.scss
FlyersPh9: Reverted back to flex, but was able to keep `flex-shrink: 0` removed.

src/wizard/wizard.scss
FlyersPh9: If I remove this line of CSS and keep the HTML structure the same between the default and long wizard, the default wizard will look like this:
Screen Shot 2021-08-05 at 1 21 01 PM

If you'd rather handle showing / hiding that portion in React I can remove this.

src/wizard/wizard.scss
FlyersPh9: I think reverting `.iui-circle` back to flex fixed this:
Screen Shot 2021-08-05 at 1 23 41 PM

FlyersPh9: I was able to get this working.
Screen Shot 2021-08-05 at 1 30 04 PM

src/wizard/wizard.scss
FlyersPh9: I've made some class name changes and tried lowering specificity.

mayank99: changing classnames without reducing nesting does not lower specificity
image

FlyersPh9: I did reduce nesting as well, that's why I renamed the classes so by themselves they make a little more sense.  There are some parts that are still nested, I'm unsure if they can be un-nested & still work correctly.

mayank99: I think you can remove the top level `iui-wizard` from all places, no?

src/wizard/wizard.scss
bentleyvk: We are already doing that now in React.

FlyersPh9: Ok, I have removed this part.  Should I also remove line 6-8 within `long.scss` as it's doing something similar?

bentleyvk: Yeah, remove from there too.

FlyersPh9: Ok, removing in the next push.

src/wizard/wizard.scss
mayank99: Much better now!
image

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
-        }
-        font-weight: $iui-font-weight-semibold;
-      }
+  &.iui-completed {
+    &.iui-clickable .iui-wizard-circle {
+      @include iui-focus(iui-focus-positive-box-shadow);
 
-      > .iui-wizards-step-circle {
+      &:hover {
+        cursor: pointer;
```
mayank99, line 59: weird place for this rule. It can be moved up because cursor automatically implies hover

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
+    align-items: center;
+    border-radius: 50%;
+    display: flex;
+    width: $iui-l;
+    height: $iui-l;
+    justify-content: center;
+    overflow-wrap: break-word;
+    padding: 0;
+    position: relative;
+    text-align: center;
```
mayank99, line 120: I think these 3 rules (padding, position, text-align) can be removed

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
+    @include themed {
+      border: 1px solid t(iui-color-background-5);
+      background-color: t(iui-color-background-1);
+      color: t(iui-text-color-muted);
     }
   }
 }
+
+@mixin iui-wizard-step-name {
+  text-align: center;
```
mayank99, line 132: This can be removed too, I think

src/wizard/workflow.scss
```diff
@@ -2,49 +2,37 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 
-@mixin iui-wizards-workflow {
-  @include iui-wizards;
-
-  > .iui-wizards-wrapper {
+@mixin iui-wizard-workflow {
+  ol {
```
mayank99, line 6: `> ol` here too

src/wizard/workflow.scss
```diff
@@ -2,49 +2,37 @@
...
     align-items: center;
```
mayank99, line 7: Does this rule change anything?

src/wizard/workflow.scss
```diff
@@ -2,49 +2,37 @@
...
-        @include themed {
-          color: t(iui-color-foreground-accessory);
-          border-color: t(iui-color-background-positive);
-        }
-      }
+  .iui-wizard-circle {
+    padding-left: $iui-m;
+    padding-right: $iui-m;
+    border-radius: $iui-sm;
+    word-break: break-word;
```
mayank99, line 14: Is this needed? I see we already have `overflow-wrap: break-word` in the base wizard circle

src/wizard/workflow.scss
```diff
@@ -2,49 +2,37 @@
...
+  }
 
-    > .iui-wizards-step-track,
-    > .iui-wizards-step > .iui-wizards-step-track {
-      &.iui-wizards-step-track-before,
-      &.iui-wizards-step-track-after,
-      &.iui-wizards-step-track-main {
-        top: 0;
+  .iui-wizard-step.iui-current .iui-wizard-circle {
+    border-radius: 0;
```
mayank99, line 24: I know this is not changing from how it was before but... Are we sure about this?

Sharp corners and rounded corners in different components was already bad, but now we have them in the same component 🤢

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
-      align-items: center;
-      display: flex;
-      flex-direction: column-reverse;
-      font-size: $iui-font-size-small;
-      font-weight: $iui-font-weight-normal;
-      line-height: $iui-line-height;
-      position: relative;
-      width: 20%;
+@mixin iui-wizard-step {
+  flex: 1 1 0;
```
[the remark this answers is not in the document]
FlyersPh9: Fixed in next push.

src/wizard/workflow.scss
FlyersPh9: Fixed in next push.

src/wizard/wizard.scss
FlyersPh9: I think this is needed, but only when the wizard gets really cramped for space horizontally.  Screenshot of how it looks when removed.
Screen Shot 2021-08-11 at 10 30 18 AM

src/wizard/workflow.scss
FlyersPh9: I don't believe this is needed, removed in next push.

src/wizard/workflow.scss
FlyersPh9: The shapes are following [industry standard designs](https://www.gliffy.com/blog/guide-to-flowchart-symbols).

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
-      }
-
-      &.iui-wizards-step-track-main {
-        display: block;
-        flex-basis: fill;
-        flex-grow: 1;
-        overflow: visible;
-        position: relative;
+    .iui-wizard-circle {
+      font-weight: $iui-font-weight-semibold;
```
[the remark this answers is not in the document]
FlyersPh9: Good suggestion, fixed.

src/wizard/wizard.scss
```diff
@@ -1,168 +1,137 @@
...
+  }
 
-    > .iui-wizards-step-current {
-      > .iui-wizards-step-title {
-        @include themed {
-          color: t(iui-color-foreground-positive);
-        }
-        font-weight: $iui-font-weight-semibold;
-      }
+  &.iui-completed {
```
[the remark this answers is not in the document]
FlyersPh9: This is now implemented.  I kept the `.iui-clickable` class incase of situations where someone might not want past steps to be clickable.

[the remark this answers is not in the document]
mayank99: but we don't really allow toggling the clickable class in React

src/wizard/wizard.scss
```diff
@@ -27,68 +26,98 @@
...
+    align-items: center;
+    border-radius: 50%;
+    display: flex;
+    width: $iui-l;
+    height: $iui-l;
+    justify-content: center;
+    overflow-wrap: break-word;
+    padding: 0;
+    position: relative;
+    text-align: center;
```
mayank99, line 40: padding, position and maybe text-align can all be removed.

(Duplicating comment since code moved)

src/wizard/wizard.scss
```diff
@@ -27,68 +26,98 @@
...
+          background-color: t(iui-color-background-positive);
+        }
       }
     }
   }
 
   &:first-child .iui-wizard-track-content::before,
   &:last-child .iui-wizard-track-content::after {
-    background-color: transparent;
+    background-color: transparent !important;
```
mayank99, line 120: Do you intend to keep this? 😄

src/wizard/workflow.scss
```diff
@@ -3,29 +3,33 @@
...
-    @include themed {
-      color: t(iui-color-foreground-accessory);
-      border-color: t(iui-color-background-positive);
+    &:first-child .iui-wizard-circle,
+    &:last-child .iui-wizard-circle {
+      border-radius: $iui-sm;
+      @include themed {
+        background-color: t(iui-color-background-1);
+        border-color: t(iui-color-foreground-positive);
+        color: t(iui-color-foreground-positive);
```
mayank99, line 31: These should already be set to foreground-positive.

src/wizard/wizard.scss
```diff
@@ -27,68 +26,98 @@
...
-
-      &:hover {
-        cursor: pointer;
-        @include themed {
-          background-color: t(iui-color-background-positive);
-          border-color: t(iui-color-background-positive);
-          color: t(iui-color-foreground-accessory);
-        }
+  // Future steps
+  &.iui-current ~ .iui-wizard-step {
```
mayank99, line 76: Probably shouldn't use name of the class explicitly here.

Try this:
```suggestion
  &.iui-current ~ & {
```

or
```suggestion
  &.iui-current ~ #{&} {
```

src/wizard/wizard.scss
```diff
@@ -27,68 +26,98 @@
     color: t(iui-color-background-positive);
   }
 
+  // Completed steps
+  .iui-wizard-circle {
+    align-items: center;
+    border-radius: 50%;
```
mayank99, line 32: Dont we have a variable for 50%?

src/wizard/workflow.scss
bentleyvk: Hm, but these are flow chart standards and for me wizard is not the same as a flowchart. Flowchart is more complex. And even if we want to follow this standard, we also have bad implementation, as oval steps are only for beginning and ending, and we are using them for not current steps.
image

mayank99: > oval steps are only for beginning and ending, and we are using them for not current steps.

This was fixed in latest. Oval is only used for start/end. Everything else gets green rectangle. "current" is ignored.

src/wizard/wizard.scss
FlyersPh9: Removed in next push.

src/wizard/workflow.scss
FlyersPh9: But border-color and color are set to this up above:
```
border-color: t(iui-color-background-positive);
```

src/wizard/wizard.scss
FlyersPh9: Good suggestion, I will go with `&.iui-current ~ &`.

src/wizard/wizard.scss
FlyersPh9: No, this was just temporary until I figured out how to fix the issue.  It's been fixed in the latest push.

mayank99 (approved):

backstop/tests/wizard.html
```diff
@@ -33,175 +33,195 @@
...
-        </p>
+        </div>
       </div>
+
     </section>
 
     <hr />
-    <h2 id="workflow">Workflow diagram</h2>
 
+    <h2>Workflow diagram</h2>
```
bentleyvk, line 187: Missing `iui-current` on one of the steps.
Also, should upcoming oval step look like a completed step?
image

mayank99: This does not look right. There should be no "iui-current" in workflow diagram

FlyersPh9: The workflow diagram is kind of unique (and probably shouldn't be within wizard) as it doesn't have the concept of a current step.  In the previous implementation I was using the `.iui-current` class to apply the styling of all the boxes that weren't first or last.  Now it's using `:first-child` and `:last-child` to do the styling instead.

PR #110, opened 2021-05-21
Update build.yml to host demo pages

mayank99:
Added deploy job that hosts demo pages when workflow runs on main branch. Preview will be available at https://itwin.github.io/iTwinUI/

---

Notable changes:
- index.html was added to redirect component links to `backstop/minified/` folder.
- node_modules will now be cached
- merging to main will now delete html report from all folders in gh_pages branch **except the open PRs**
- deploying to gh pages no longer preserves git history and will now use github-actions bot as the commit author
- checks are now created using [github-script action](https://github.com/actions/github-script)


.github/workflows/build.yml
```diff
@@ -37,19 +43,100 @@ jobs:
...
+              conclusion: '${{ job.status }}',
+              name: 'Test results',
+              output: {
+                "title": "Test results",
+                "summary": "See backstop html report at https://itwin.github.io/iTwinUI/${{ github.event.number }}/html_report/"
+              }
+            })
+
+  deploy:
+    name: Deploy demo html pages
```
mayank99, line 74: I was thinking of making a separate workflow file for this job but it might be better this way because it's similar to the build job (they use the same actions and everything). Also it could get confusing if there is a "deploy" and another "publish" (for npm release).

---

Other than that, this PR should be good. This is what the gh-pages branch looks like after a fresh deployment: https://github.com/iTwin/iTwinUI/tree/3c5a080c1e10aabb73e1e505bb0dd58a807a94fb

bentleyvk (approved):

PR #111, opened 2021-05-21
Declare $fallback before using it

mayank99:
Tackles this deprecation warning:

> DEPRECATION WARNING: As of Dart Sass 2.0.0, !global assignments won't be able to declare new variables. Consider adding `$fallback: null` at the root of the stylesheet.
> 
> 616 │   $fallback: true !global;
> 

image

### DIFF
--- a/src/style/theme.scss
+++ b/src/style/theme.scss
@@ -418,6 +418,7 @@ $fallbackvars: getfallbackvars();
   }
 }
 
+$fallback: null;
 
 //  Adds a CSS variable with a fallback value.
 //


veekeys:
I see in the examples, there is no exclamation mark before global.
image

Wonder, we need it and what it changes?

mayank99:
> I see in the examples, there is no exclamation mark before global.

@veekeys  That is just an example of a value, not syntax. 😄 

The `!` is used for changing the value of a global variable from a local scope. See https://sass-lang.com/documentation/variables#shadowing

veekeys (approved):

PR #117, opened 2021-05-27
Separate time picker

veekeys:
Moved Time picker component out of date picker, so it could be used alone.
Added examples with tests.

Moved time-picker into separate folder.


src/date-picker/date-picker.scss
```diff
@@ -145,11 +138,37 @@
 }
 
 @mixin iui-time {
+  @include iui-reset;
+  @include iui-font-family;
+  user-select: none;
+  text-align: center;
+  height: $iui-baseline * 26;
```
FlyersPh9, line 145: I think I set this height so that the time picker was the same height as the calendar.  If there is no calendar present, I don't suppose we _need_ to set a height.

backstop/tests/date-picker.html
```diff
@@ -1504,74 +1504,75 @@ <h2>Date & time</h2>
             </div>
           </div>
         </div>
+        <div class="iui-time-picker">
```
FlyersPh9, line 1507: Does it need to have it's own class?  We have `.iui-date-picker` when showing a calendar & a calendar + time combo.  Why not reuse `.iui-date-picker` and only show time within it?

FlyersPh9, line 1507:
```suggestion
        <div class="iui-time-picker" role="dialog" aria-modal="true">
```
Copying the accessibility from lines 32-35.

FlyersPh9: Ah wait, I see.  So above we have:
```
<date picker>
  <calendar />
  <time picker />
</date picker>
```

You're proposing:
```
<time picker />
```
where it is not nested within date picker at all?  So you're not really introducing a new class?

src/date-picker/date-picker.scss
veekeys: So we show 24 hours without scroll?

backstop/tests/date-picker.html
veekeys: I have a new wrapper class, which has few styles taken from Date picker to have same styling..
Unless you figure out the way not to have that class (but I think we need it).

src/date-picker/date-picker.scss
FlyersPh9: We would definitely want a scroll, I was more so saying I'm not sure it needs to be capped at this height if no calendar is present.

backstop/tests/date-picker.html
FlyersPh9: I am able to pull of something complex like this (not suggesting this layout, just showing how complex it can get) and it can be as simple as a single time column.

Screen Shot 2021-05-28 at 12 10 32 PM Screen Shot 2021-05-28 at 12 11 07 PM

src/date-picker/date-picker.scss
```diff
@@ -2,10 +2,10 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 @import '../icon/index';
+@import '../menu/index';
 
 @mixin iui-date-picker {
-  @include iui-reset;
-  @include iui-font-family;
+  @include iui-dropdown;
```
FlyersPh9, line 8: This is using the same mixin used on the dropdown menu & column filter menu.

mayank99: What does it give in this case?

FlyersPh9: Not a whole lot, but it gives [the properties we want to be consistent across all dropdowns](https://github.com/iTwin/iTwinUI/blob/main/src/menu/menu.scss#L7):
- Reset
- Font family
- Background color
- Drop shadow

backstop/tests/date-picker.html
veekeys: Have in mind, that the code we write for TimePicker (as single component) should be put under date picker and it should work.
Will this html work then after your changes??
<iui-date-picker>
....
 <iui-date-picker>
   <iui-time>
   ....
   </iui-time>
 </iui-date-picker>
</iui-date-picker>

?

veekeys: I have moved back separate time picker...

src/time-picker/time-picker.scss
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../menu/index';
+
+@mixin iui-time-picker {
+  @include iui-reset;
+  @include iui-font-family;
```
mayank99, line 8: Maybe should use `iui-dropdown` mixin here to be consistent

src/time-picker/time-picker.scss
```diff
@@ -0,0 +1,74 @@
...
+  &:first-child {
+    box-shadow: $iui-elevation-2;
+  }
+
+  > .iui-time,
+  > .iui-period {
+    padding: $iui-baseline $iui-s;
+    overflow-y: auto;
+
+    &:not(:first-child) {
```
mayank99, line 32: Does it need to be repeated?

src/date-picker/date-picker.scss
```diff
@@ -16,13 +16,6 @@
   }
 
   @include iui-calendar;
-  @include iui-time;
-
-  > div:not(:last-child) {
-    @include themed {
-      border-right: 1px solid t(iui-color-background-4);
```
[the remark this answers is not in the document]
veekeys: no

src/time-picker/time-picker.scss
veekeys: Should not.. it adds shadow, which brakes things a little..

src/time-picker/time-picker.scss
veekeys: well yes.. Unless you can find a better way, how to handle different situations (time picker alone, time picker in date picker, etc...)

backstop/tests/date-picker.html
```diff
@@ -2057,112 +2059,729 @@ <h3>24hr format</h3>
...
+          </div>
         </div>
+
       </div>
     </section>
 
+    <br />
+
+    <h3>Everything</h3>
+    <section id="demo-everything">
```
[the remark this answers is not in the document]
veekeys: removed it instead

mayank99 (approved):

PR #119, opened 2021-05-28
Wizard responsive design bug fixes

FlyersPh9:
- Default wizard's step uses a % rather than the step's description using VW.
- Workflow Diagram handles responsive design a little bit better:

Before:
Screen Shot 2021-05-28 at 1 10 45 PM

After:
Screen Shot 2021-05-28 at 1 09 52 PM


src/wizard/long.scss
```diff
@@ -5,9 +5,13 @@
 @mixin iui-wizards-long {
   @include iui-wizards;
 
-  > .iui-wizards-wrapper > .iui-wizards-step > .iui-wizards-step-title {
-    display: none;
-  }
+  > .iui-wizards-wrapper > .iui-wizards-step {
+    width: auto;
```
mayank99, line 9: Is this the magical fix? 😄

mayank99: It changes the demo image slightly 🤔

FlyersPh9: Yeah, at least for the long wizard.

For the default wizard we needed to set a fixed width (20%) because we can't know how long the step's text description might be.

For the long wizard the text description is not underneath the step, so we can reset the width back to auto.

mayank99: wait this is in long.scss. What does it give?

mayank99: oh it's to reset

backstop/tests/wizard.html
```diff
@@ -190,15 +190,12 @@ <h2 id="workflow">Workflow diagram</h2>
...
           <span class="iui-wizards-step">
-            <span class="iui-wizards-step-track iui-wizards-step-track-after"></span>
             <span class="iui-wizards-step-circle">Start</span>
           </span>
 
           <span class="iui-wizards-step-track iui-wizards-step-track-main"></span>
 
           <span class="iui-wizards-step iui-wizards-step-current">
-            <span class="iui-wizards-step-track iui-wizards-step-track-before"></span>
-            <span class="iui-wizards-step-track iui-wizards-step-track-after"></span>
```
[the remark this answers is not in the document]
FlyersPh9: I think they are leftover from copy / pasting an example from above.  In the image below you can see the top-aligned lines poking out from behind the start & complete steps.  They are under the middle step as well, we just can't see them.
Screen Shot 2021-05-28 at 1 18 55 PM

Actually, after reading this comment I inspected the long wizard and these spans were also not needed.  It appears these are only needed in the default wizard.

src/wizard/workflow.scss
```diff
@@ -35,6 +39,8 @@
       &.iui-wizards-step-track-before,
       &.iui-wizards-step-track-after,
       &.iui-wizards-step-track-main {
+        top: 0;
```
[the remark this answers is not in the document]
FlyersPh9: We're using `top: 12px` to align the line with the center of the circles.  If I don't set top to 0, we end up with this.
Screen Shot 2021-05-28 at 1 25 40 PM

PR #122, opened 2021-06-01
Remove iui-icon to be clickable in button

maxime4000:
Mostly this remove the svg to be clickable in button which make mouseevent target the button element and never the svg or path element.


veekeys:
@maxime4000 thank you for your PR!
As per the the instructions in our [Contributing guide](https://github.com/iTwin/iTwinUI/blob/main/CONTRIBUTING.md#committing-your-work), there is a need to have an [issue](https://github.com/iTwin/iTwinUI/issues/new?assignees=&labels=bug&template=bug_report.md&title=) created, so we could better understand what is the bug and what this PR fixes.
With your current description and a fix, it is really hard to understand and review.
Thanks!

maxime4000:
I'm messing with button and mouseevent right now and when you click on a iconbutton on the icon part, the mouseevent will target path or svg, though the svg or path part is not interesting and if I'm not clicking on the svg, i will get button.iui-button. So this will force click on button where ever you click on the button, you will get an event targeting the button.iui-button. MouseEvent will be consistent with this fix.

maxime4000:
Same thing would happen for iui-label in an iui-button. I'm doing it for iui-icon because I think it's fair to say that it will never be the case where you want to know which part of the SVG you click on. I'm not interested in iui-label because I don't have this use case. If you want to add iui-label to that, feel free to edit this PR. 

My use case is that I have a parent that is listening to mouseevent and I want to exclude button click to this listener. It's simplier to know that mouseevent.target is iui-button than path from iui-button>svg>path

bentleyvk:
Have you tried `event.currentTarget`? https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

maxime4000:
> Have you tried `event.currentTarget`? https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

current target is null

veekeys:
@maxime4000 Have you tried applying your custom class (or inline styles) on the icon you are passing to the button component?
I am just afraid it is really specific and if we disable it, someone might come and say he needs it.. So we better do not remove things (and let users remove what they do not need) than vice versa.
And again, if we start removing from here ,we need to do it in other million places, which we do not really want to do now (otherwise, inconsistent behavior expected in different components).

```
.my-custom-icon {
  pointer-events: none;
}

...
<IconButton>
  <MySvg className="my-custom-icon" />
</IconButton>
```

maxime4000:
I think it's consistency to always get the same result of mouseevent where ever you click in an icon button. We are not talking about custom button or other component and we are not talking about other children possibility in a button. In 90% of the case, you never use the mouseevent in button and when you do, 99% of the time it's about e.preventDefault() or switch(event.button). So I think my change is very specifics, but shouldn't be ignore. Anyway, can you give me a reason why would you want to know which part of a SVG you click on so you do a different behavior ? Pretty sure that goes against every rule of UX.

mayank99:
> _We are not talking about custom button or other component and we are not talking about other children possibility in a button._

Yes we are. You can pass any children in a button, and there can be a need for pointer events on those children. And if we change the pointer-events behavior in our button, it does some expectation that other components in our library would behave similarly.

> _why would you want to know which part of a SVG you click on so you do a different behavior_

Maybe not specific parts of the svg but there can be a use case for knowing whether the svg itself was hovered (and then acting on that, e.g. changing fill or something).

Either way, you're missing the point. The svg is not part of the component, and we should not mess with the events there, especially since it's so easy for users to add this rule themselves.

maxime4000:
Are we reviewing the same code, because it seem like you guys are searching any reason to not allow this simple change to be push...
```.scss
.iui-button > .iui-icon:not(.iui-user-icon) {
    width: $iui-icons-default;
    height: $iui-icons-default;
    transition: fill $iui-speed-fast ease-out;
    pointer-events: none;  // This is the only added line
}
```
The change is only apply to iui-icon which is a direct child of iui-button so it should only be the case if you use a icon button... This change does not affect anything else.

----

@veekeys 
> if we disable it, someone might come and say he needs it.
Well he can do the same by enabled in it back with his custom classname. This change is considering that 99% of the time, it won't bother anyone. To the 1% chance a dev has to reset the pointer-events, he can add custom classname...

> And again, if we start removing from here ,we need to do it in other million places, which we do not really want to do now

A million ? Where ? Give me 10 place that this change should be apply somewhere else which has a use case of having a onClick event listener in the first place. And if you do find 10 place to do it, fix it while your at it !
---

@mayank99 
> Yes we are. You can pass any children in a button, and there can be a need for pointer events on those children. 

No we don't ! Anything else other than a icon as direct child of button will keep his pointer-events.

> And if we change the pointer-events behavior in our button, it does some expectation that other components in our library would behave similarly.

No it does not make expectation that every other component should work the same because it's only apply to basic IconButton and nothing else. 

> Maybe not specific parts of the svg but there can be a use case for knowing whether the svg itself was hovered (and then acting on that, e.g. changing fill or something).

Dude we are talking about javascript or SCSS ? Javascript: Hover event can be done. SCSS : `.iui-icon:hover {}` will still work. What are you saying...

> Either way, you're missing the point. The svg is not part of the component, and we should not mess with the events there, especially since it's so easy for users to add this rule themselves.

No I'm not ! SVG is part of the component as this rule imply it : `.iui-button > .iui-icon:not(.iui-user-icon) {}` So unless I'm very very very very very stupid and I don't understand this source code after 1 year of making change to it and using it every day to build a website. I think I know the source code. 

---

I'm sorry if I'm arrogant right now, but we are debating on a freaking line that any front-end developer should already understand after 1 minutes of looking at it that this change won't do anything bad to iui-button... Can we just approve and never talk about this again ?

maxime4000:
Here's my final comment, You can close this PR if it's not going to happen.
I still think this code under is never going to happen, because it make no sense for UX to suggest this type of behavior nor make sense to ever want to add onClick listener directly on iui-icon in a button. So I don't see why this fix could do anything wrong to iTwinUI
```tsx
<Button onClick={(e) => {Do behavior1}}>
  <SvgSomething onClick={(e) => {Do behavior2}}/>
</Button>
// Or
const btnOnClick = (e) => { 
  if(e.target.classList.include("iui-icon")){
    // Do behavior2
  } else {
    // Do behavior1
  }
};
```
And I'm pushing the change here (while it's already apply in my app) because it offer consistency on mouseevent in iui-button which I think is an improvement for iui-button, but will also avoid other the need to investigate this quirky case that can happen if they click at the wrong place in the button.

Issue #131, opened 2021-06-03
Fix DEPRECATION WARNING when using with Dart Sass


bsteinbk:
#### Environment
- __Package version(s)__: 0.17.1
- __Operating System__: Windows 10
- __Browser name and version__:  NA - warnings during build

#### Code Sandbox
None

#### Steps to reproduce
1. use scss files in product that uses dart Sass - `sass: ^1.32.0` (such as iModeljs)

#### Actual behavior
```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
More info and automated migrator: https://sass-lang.com/d/slash-div
   ╷
856 │ $iui-inputs-padding: (($iui-inputs-height - $iui-line-height) / 2) - 1;
    │                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

#### Expected behavior
No warning

#### Additional information
Easy fix - convert ` / 2` -> ` * 0.5`

PR #132, opened 2021-06-03
Address sass division deprecation warnings when used in Dart Sass

bsteinbk:
Fix DEPRECATION WARNING when using with Dart Sass #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - **Improved wizard responsive design** by using percentage width.
 - **Fixed user-icon size not working when placed in icon button.**
+- **Fixed `DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`.** when using scss files in project with Dart Sass 1.32.0 or later.
 
 ## [0.17.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+`Date`
+
+### Fixes
+
+- **Fixed Dart Sass deprecation warning about division operator** by replacing / with \*.
+
 ## [0.17.1]
 
 `2021-05-28`
@@ -8,7 +16,6 @@
 
 - **Improved wizard responsive design** by using percentage width.
 - **Fixed user-icon size not working when placed in icon button.**
-- **Fixed `DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`.** when using scss files in project with Dart Sass 1.32.0 or later.
 
 ## [0.17.0]
 
--- a/src/tooltip/tooltip.scss
+++ b/src/tooltip/tooltip.scss
@@ -14,7 +14,7 @@
   width: max-content;
   overflow-wrap: break-word;
   box-sizing: content-box;
-  padding: round($iui-baseline / 3) $iui-s;
+  padding: round($iui-baseline * 0.3333) $iui-s;
   z-index: 999;
   box-shadow: $iui-elevation-8;
   pointer-events: none;
--- a/src/tooltip/tooltip.scss
+++ b/src/tooltip/tooltip.scss
@@ -14,7 +14,7 @@
   width: max-content;
   overflow-wrap: break-word;
   box-sizing: content-box;
-  padding: round($iui-baseline * 0.3333) $iui-s;
+  padding: round($iui-baseline * 0.33333333333333333) $iui-s;
   z-index: 999;
   box-shadow: $iui-elevation-8;
   pointer-events: none;
--- a/src/tooltip/tooltip.scss
+++ b/src/tooltip/tooltip.scss
@@ -14,7 +14,7 @@
   width: max-content;
   overflow-wrap: break-word;
   box-sizing: content-box;
-  padding: round($iui-baseline * 0.33333333333333333) $iui-s;
+  padding: round($iui-baseline * 0.33) $iui-s;
   z-index: 999;
   box-shadow: $iui-elevation-8;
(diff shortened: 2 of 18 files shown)


src/user-icon/user-icon.scss
```diff
@@ -81,8 +81,8 @@
       @include themed {
         fill: t(iui-icons-color);
       }
-      width: calc(100% / 1.5);
-      height: calc(100% / 1.5);
+      width: calc(100%* 0.6666667);
+      height: calc(100%* 0.6666667);
```
mayank99, line 85: This really gives the same value? 😮

src/tooltip/tooltip.scss
```diff
@@ -14,7 +14,7 @@
   width: max-content;
   overflow-wrap: break-word;
   box-sizing: content-box;
-  padding: round($iui-baseline / 3) $iui-s;
+  padding: round($iui-baseline * 0.3333) $iui-s;
```
mayank99, line 17: And this

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - **Improved wizard responsive design** by using percentage width.
 - **Fixed user-icon size not working when placed in icon button.**
+- **Fixed `DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`.** when using scss files in project with Dart Sass 1.32.0 or later.
```
mayank99, line 11: Create a new section for this like so:

https://github.com/iTwin/iTwinUI-react/blob/272fd7fc33b1302d8aaa9a8c09bd4a5e1a623754/CONTRIBUTING.md#L153-L157

Also could reword it to focus on the :
"Fixed Sass deprecation warning about division operator** by replacing `/` with `*`.

mayank99 (commented):
Thanks for the PR! We were thinking about how to tackle this warning (and `math.div` was not a good option since we use node-sass currently)

src/user-icon/user-icon.scss
bsteinbk: yep 1/1.5 = 0.6666667 - I messed up spacing I will fix that.

src/tooltip/tooltip.scss
bsteinbk: 1/3 = 0.33333

CHANGELOG.md
bsteinbk: good point - will fix

src/user-icon/user-icon.scss
mayank99: Well, I'm more concerned about the precision

src/tooltip/tooltip.scss
veekeys: So is it 4 or 5 symbols after coma? :D

src/user-icon/user-icon.scss
bsteinbk: I have updated to use 0.66666666666666667 which produces the same results as 1/1.5 in test values I tried from 1 - 9600.

src/tooltip/tooltip.scss
bsteinbk: ditto here, I updated to 0.33333333333333333 to match values returned when dividing by 3

CHANGELOG.md
bsteinbk: updated

src/tooltip/tooltip.scss
veekeys: oh...
@mayank99, could you quickly test from this branch generated css vs the current one if values produced are same?
Just to be sure that we need these "huge" numbers :D

mayank99: So I checked the generated css, and then realized we are using `round` here so the precision would not matter after like 5 points.

And in that other example, we have `calc` so I think we don't even need to switch to `*` (because for [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc()), `/` is a supported operator in native css).

src/user-icon/user-icon.scss
```diff
@@ -67,8 +67,8 @@
     position: absolute;
     right: -2px;
     bottom: -2px;
-    width: calc(100% / 4);
-    height: calc(100% / 4);
+    width: calc(100% * 0.25);
```
bentleyvk, line 70: Maybe just `25%`? 😀 Same with `66.66...%`

bsteinbk: so there really is no need for calc.... just `width: 25%` right?

src/tooltip/tooltip.scss
bsteinbk: @mayank99 - so change it to 0.33?

src/user-icon/user-icon.scss
mayank99: Yeah. I would also go through all other instances of `calc` that you changed, to see if they can be simplified or undone (because using `/` in calc should not give you the deprecation warning).

src/tooltip/tooltip.scss
mayank99: Yup

src/user-icon/user-icon.scss
mayank99: Turns out "all other instances" are all in this file 😄

src/tooltip/tooltip.scss
bsteinbk: OK, I think I have it cleaned up now.

src/user-icon/user-icon.scss
bsteinbk: Updated to 25% and reverted others back to calc(100% / 1.5) since it was in calc function should be ignored by sass.

mayank99 (approved):

PR #137, opened 2021-06-04
Increase specificity of sidenav-button selectors

### LINKED ISSUE #125: Sidenav buttons style is overriden by button
#### Steps to reproduce
Seems to be depending on the order of css files imports, no easy reproduction steps.

#### Expected behavior
Borders on hover are not this much dark @FlyersPh9 

#### Additional information
Button style overrride sidenav button styles:

image

mayank99:
Sidenav buttons should have a specificity higher than that of default button so that the order of imports won't matter (to fix #125). This PR increases the specificity by adding redundant selectors in a few places.

### DIFF
--- a/src/side-navigation/side-navigation.scss
+++ b/src/side-navigation/side-navigation.scss
@@ -80,7 +80,7 @@
 
     &,
     &:enabled:hover,
-    &:disabled,
+    &[disabled],
     &:focus-visible {
       @include themed {
         border-color: t(iui-color-background-5);
--- a/src/side-navigation/side-navigation.scss
+++ b/src/side-navigation/side-navigation.scss
@@ -76,7 +76,7 @@
     height: $iui-baseline * 5;
 
     &,
-    &:hover {
+    &:enabled:hover {
       @include themed {
         border-color: t(iui-color-background-5);
       }
--- a/src/side-navigation/side-navigation.scss
+++ b/src/side-navigation/side-navigation.scss
@@ -36,7 +36,7 @@
         border-bottom: none;
       }
     }
-    
+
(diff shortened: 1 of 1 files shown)


src/side-navigation/side-navigation.scss
```diff
@@ -76,7 +76,7 @@
     height: $iui-baseline * 5;
 
     &,
-    &:hover {
+    &:enabled:hover {
```
veekeys, line 79: yeah, I guess we could live with this for some time before we figure out about bigger refactoring..
Have you actually tested it helps?
does not work before your fix and works after...?

mayank99: Well, I could not reproduce the problem at all, because it depends on the order of imports.

Maybe I can change from all.css to individual imports to test locally

veekeys: Yeah, thats what I was asking...

veekeys: Its kinda obvious, what import order is needed to reproduce (its 2 files only and I think this was cought by Vilius test)

mayank99: Oh wow, I tested with different order and found a few more instances where sidenav-button needed extra specificity.

So I added redundant selectors for setting button height and icon size

veekeys (approved):

src/side-navigation/side-navigation.scss
mayank99: Ah damnit, border styling on focused button is still broken
image

bentleyvk (approved):

PR #150, opened 2021-06-14
Updated postcss

bentleyvk:
Postcss back-ported fix to v7 however it may take several days until it is seen by `yarn audit`.
https://github.com/stylelint/stylelint/issues/4942#issuecomment-859924007

Creating draft PR so others won't work on the same until `yarn audit` sees the fix.


mayank99 (commented):
Should be good to go!

We could even use `yarn audit` now? `audit-ci` has always looked weird 😄

mayank99 (approved):

PR #152, opened 2021-06-14
Fix table spacing, fill and cursor

mayank99:
- Added margin-right to filter icon
- Added iui-actionable class for cursor: pointer in header cells
- Added iui-icons-color fill for sort icon hover state
- Formatting fixes

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -103,6 +103,7 @@
     visibility: hidden;
     @include themed {
       fill: t(iui-icons-color-actionable);
+      opacity: t(iui-opacity-3);
     }
   }
 
@@ -114,6 +115,7 @@
 
     .iui-sort {
       visibility: visible;
+      opacity: 1;
     }
   }
 }
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -23,7 +23,7 @@
   @include iui-table-cell-icon;
 }
 
-@mixin iui-table-header { 
+@mixin iui-table-header {
   user-select: none;
   @include themed {
     background-color: t(iui-color-background-3);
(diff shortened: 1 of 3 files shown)


src/table/table.scss
```diff
@@ -36,10 +36,14 @@
...
     @include iui-focus;
-    cursor: pointer;
+
+    &.iui-actionable {
+      cursor: pointer;
+    }
 
     > .iui-filter-button {
       margin-left: $iui-xs;
+      margin-right: $iui-xs;
```
mayank99, line 46: @FlyersPh9 Can you check what the value here should be?

0 (current):
image

xs (4px):
image

s (8px)
image

FlyersPh9: My initial thought is to have the spacing between the button and the right edge of the column to be equal to either the red or orange spacing in the image below.
Screen Shot 2021-06-14 at 11 45 53 AM

I know the red gap is 16px, which is probably not our best use of space.  And the orange gap is probably too close to the edge of the column and a user might accidentally hover over the column resizer.  So I'd say let's go with the 8px gap.

src/table/table.scss
```diff
@@ -99,6 +103,7 @@
     visibility: hidden;
     @include themed {
       fill: t(iui-icons-color-actionable);
+      opacity: t(iui-opacity-3);
```
FlyersPh9, line 106: Another option could be to use `iui-icons-color` for the hover state & `iui-icons-color-actionable` for the active sort state.

mayank99: Done

src/table/table.scss
mayank99: Changed to `iui-s`

FlyersPh9 (approved):

PR #154, opened 2021-06-14
Add reduced motion media query

FlyersPh9:
- Wrapped all `transition` & `animation` within `@media (prefers-reduced-motion: no-preference)`.
  -  **Purposefully excluded progress indicators.**
  - Can confirm it is working on modal, expandable block, header, side navigation.  Can not confirm if it is working on toast notifications.
- Added `@include iui-reset` on a couple of components that were missing it.
- Added missing `.iui-fill` in linear progress indicator


src/style/global.scss
```diff
@@ -10,7 +10,7 @@
 
   font-size: $iui-font-size;
   line-height: $iui-line-height;
-
+  scroll-behavior: smooth;
```
mayank99, line 13: If this applies to all components, I think this could lead to unintended behavior in some places.

For example, in Select, we automatically scroll to the selected option. Setting `smooth` would cause an unnecessary "delay" while the animation is happening.

mayank99: Oh and it would also affect user apps. So this is definitely not the right place.

FlyersPh9: I've removed this for now, will rely on the individual app teams to implement this on their own as needed.

src/style/global.scss
```diff
@@ -55,3 +55,14 @@ html {
     }
   }
 }
+
+@media (prefers-reduced-motion: reduce) {
+  *:not(.iui-fill),
+  *::before,
+  *::after {
```
mayank99, line 62: 1. Why is `.iui-fill` excluded?
2. Applying to `*` will impact user styles.
3. I think we could do `[class*='iui-']` instead, but then it won't work for child elements of our components.
4. If we do `[class*='iui-'], [class*='iui-'] *`, then it will still impact user styles (`*` will target everything under `.iui-body`).

I think we might need to move this to a mixin, which we can include in our components. Or even inside `iui-reset`.

FlyersPh9: `.iui-fill` is the animated part within a progress indicator.  I now realize that the `.iui-fill` is only applied on the radials, the linears had no class name and the CSS was targeting `> *`.  Next push will change that so the linear also use `.iui-fill`.

mayank99 (approved):

PR #155, opened 2021-06-14
Optimize backstop test results

FlyersPh9:
- All `<section>`s include padding, column flex direction, and gap.  Some demo pages override these defaults as needed.
- All tests target a selector, no longer targeting full browser window (except for sticky alert, footer).
- No tests include non-UI component elements within the results (exclude `<h1>`-`<h6>`, `<hr>`, theme switcher button).
- All tests follow the same name format.
- When possible, test results width reduced to ~800px via scenario.js, not via demo HTML.
- Added test for badge component.

This PR helps us towards #200.


mayank99:
Since this one is changing all images, I think we can add more stuff here:
- Update backstopjs to latest version
- Add `light_` prefixes to image files for better sorting

PR #156, opened 2021-06-15
Refactor tabs

FlyersPh9:
- Tabs use `<button>` instead of `<a>`.
- Added `.iui-large` to default & borderless buttons.
- `.iui-large` tabs support secondary line.
- All tabs support icons.
- Tabs can now be disabled.
- Visual changes to align styling within tabs and within other components (buttons).
- All tabs support `.iui-not-animated`, only borderless & pill support `.iui-animated`.
- Introduced `.iui-tabs-wrapper` class.

127 0 0 1_5500_backstop_minified_tabs html


src/tabs/classes.scss
```diff
@@ -2,18 +2,15 @@
...
-.iui-tabs-borderless {
-  @include iui-tabs-borderless;
-}
-
-.iui-tabs-pill {
-  @include iui-tabs-pill;
-}
+.iui-tabs {
+  @include iui-tabs;
+  @include iui-tabs-default;
```
mayank99, line 7: Lets have a separate class for this so we don't need those long selectors to target default tabs.

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import './tabs';
+
+@mixin iui-tabs-default {
+  @include iui-tabs;
```
mayank99, line 7: `iui-tabs-default` is a modifier, so it shouldn't need to include base `iui-tabs` styles again.

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  > li {
+    > button {
```
mayank99, line 10: This can be flattened to reduce specificity (by adding a new classname)

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
...
+      height: $iui-baseline * 3;
```
mayank99, line 11: Shouldn't this use one of our height variables?

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
...
+      padding: 0 $iui-m;
```
mayank99, line 12: This could also use padding variables

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
...
+      box-sizing: border-box;
+      @include themed {
+        background-color: t(iui-color-background-2);
+        border: 1px solid t(iui-color-background-5);
+      }
+
+      .iui-label {
+        text-align: left;
+
+        .iui-description {
```
mayank99, line 22: These classnames are too generic

src/tabs/default.scss
```diff
@@ -0,0 +1,66 @@
...
+    &:not(:last-child) > button {
+      border-right: none;
+    }
+  }
+
+  &:not(.iui-borderless):not(.iui-pill) ~ .iui-tabs-content {
+    box-sizing: border-box;
+    padding-left: $iui-m;
+    padding-right: $iui-m;
+    margin-top: -1px;
```
mayank99, line 54: Why is -1 necessary?

src/tabs/pill.scss
```diff
@@ -6,98 +6,54 @@
 
 @mixin iui-tabs-pill {
   @include iui-tabs;
-  display: flex;
   flex-flow: row;
   justify-content: space-evenly;
-  margin: $iui-baseline/2 0;
+  border-radius: $iui-border-radius;
+  gap: 2px;
```
mayank99, line 12: We avoid using gap for IE support

src/tabs/pill.scss
```diff
@@ -6,98 +6,54 @@
...
+  @include themed {
+    background-color: t(iui-color-background-2);
+    border: 2px solid t(iui-color-background-2);
```
mayank99, line 15: Why 2px here?

src/tabs/pill.scss
```diff
@@ -6,98 +6,54 @@
...
-      }
+    > button {
+      height: $iui-baseline * 4;
+      background-color: transparent;
+      border-radius: $iui-border-radius;
+      width: 100%;
+      justify-content: center;
 
-      > svg {
+      &:enabled:hover {
```
mayank99, line 28: This has an extremely high specificity!

src/tabs/pill.scss
```diff
@@ -6,98 +6,54 @@
...
-
-        svg {
-          @include themed {
-            fill: t(iui-icons-color-primary);
-          }
+          background-color: rgba(
+            t(iui-color-foreground-primary-rgb),
+            t(iui-opacity-6)
+          );
+          box-shadow: inset 0 calc(#{$iui-xxs} * -1) 0 0;
```
mayank99, line 43: 🤔

src/tabs/tabs.scss
```diff
@@ -4,100 +4,99 @@
...
-
-      &::after {
-        content: '';
-        position: absolute;
-        z-index: 1;
-        transition: all $iui-speed-fast;
-        height: $iui-xxs;
-        opacity: 0;
 
+      .iui-icon:not(.iui-user-icon) {
```
mayank99, line 33: Why `:not(.iui-user-icon)`?

src/tabs/borderless.scss
```diff
@@ -49,4 +44,18 @@
...
     }
   }
+
+  &.iui-large {
+    > li > button {
+      height: $iui-baseline * 6;
+    }
+  }
+
+  ~ .iui-tabs-content {
```
mayank99, line 54: Does content area really need to have border for other types? I feel like it would add too much visual noise

mayank99 (commented):
Initial comments are mostly about reducing specificity by _not_ mimicking DOM structure in css.

If this PR is really meant to be a "refactor", then we should do it the right way, since we have sort of a "clean slate". And it doesn't depend on our button.css, so there is nothing stopping us.

src/tabs/classes.scss
FlyersPh9: I've added `.iui-default` class.

src/tabs/default.scss
FlyersPh9: Oops, removed all instances of `@include iui-tabs;` from within modifiers.

src/tabs/default.scss
FlyersPh9: Flatted by adding `.iui-tab` class to all `<button>`s.

src/tabs/pill.scss
FlyersPh9: Swapped `gap` with `margin-right: 2px` except on the last child.

src/tabs/default.scss
FlyersPh9: Without it we end up with this:
Screen Shot 2021-06-15 at 4 40 10 PM

mayank99: Why not remove border-top from content then?

src/tabs/default.scss
FlyersPh9: I'm using the same class names / structure from within the header component.  Should I use `.iui-tab-label` and `.iui-tab-description` instead?  I'm assuming if yes, we should eventually (in a separate PR) make the same change within the header?

mayank99: Yeah, I would say so. The `.iui-label` class is especially bad here, because a user might have a reason to place, say, an `.iui-button` (which includes its own `.iui-label`) inside tab.

Also, longer names would reduce specificity.

src/tabs/pill.scss
FlyersPh9: Without the `:enabled` part, we will accidentally give the disabled tab a hover state.

I suppose an alternative would be to place this further down:
```
&[disabled]:hover {
}
```

mayank99: Yeah lets add/move all the disabled styles at the bottom

src/tabs/default.scss
mayank99: I see the change, but now all the heights have changed drastically. Is this intentional?

src/tabs/default.scss
FlyersPh9: Well then we end up with something like this.
Screen Shot 2021-06-15 at 8 12 01 PM

src/tabs/pill.scss
bentleyvk: `-#{$iui-xxs}` works too. Better change to this as it is easier to read.

src/tabs/borderless.scss
```diff
@@ -4,49 +4,49 @@
 @import './tabs';
 
 @mixin iui-tabs-borderless {
-  @include iui-tabs;
-  margin: $iui-baseline/2 0;
+  .iui-tab {
+    height: $iui-component-height;
```
bentleyvk, line 8: We could just have `padding: 11px 12px` and it looks the same for default size and a bit bigger on large. But then we wouldn't need to set height.

src/tabs/tabs.scss
```diff
@@ -4,100 +4,104 @@
...
-  @include iui-font-family;
   display: flex;
+  align-items: center;
   list-style: none;
   padding: 0;
-  margin: round($iui-baseline/2) 0 0 0;
   user-select: none;
 
   > li {
     display: inline-block;
```
bentleyvk, line 14: These rules seem to do nothing now.

src/tabs/borderless.scss
bentleyvk: Oh, I see why you did this. I saw `None` tab is not fully taking height. But you could apply `display: flex` to the `li` element and it will take full height.

backstop/tests/tabs.html
```diff
@@ -18,247 +18,843 @@
...
+            class="iui-tab"
+            role="tab"
+            aria-selected="false"
+            aria-controls="none-tab"
+          >
+            <span class="iui-tab-label">None</span>
+          </button>
+        </li>
+        <li>
+          <button class="iui-button iui-borderless iui-small">
```
bentleyvk, line 114: Why not using the same tab styling with icon?
image

src/tabs/pill.scss
```diff
@@ -5,100 +5,56 @@
 @import './tabs';
 
 @mixin iui-tabs-pill {
-  @include iui-tabs;
-  display: flex;
   flex-flow: row;
   justify-content: space-evenly;
-  margin: $iui-baseline/2 0;
+  border-radius: $iui-border-radius;
```
bentleyvk, line 10: For me it looks odd seeing that small curve on an active tab 😀

src/tabs/pill.scss
mayank99: also, `xxs` is 2px, and we are using the value directly everywhere else in this PR

backstop/tests/tabs.html
FlyersPh9: The plus button I have currently is not in any standards nor has it been requested by any devs.  I was trying to think through "should the tabs use the baseline or input-height variables".  Here's another example of placing an input (select) inline with the tabs where the tabs are using baseline (44px) and the input is using the input-height (38px).
Screen Shot 2021-06-16 at 12 02 15 PM

I am still questioning which variables we should use for the tabs.

FlyersPh9: Decided against baseline, using the input height variables for all tabs.

src/tabs/default.scss
FlyersPh9: I changed all the tabs to use the input height and padding variables.  The borderless tab I multiplied the horizontal padding by 2 to get better visuals.

src/tabs/pill.scss
FlyersPh9: I've removed all `:enabled` and styled `[disabled` at the bottom.

src/tabs/pill.scss
FlyersPh9: Good suggestion, updated.

src/tabs/borderless.scss
FlyersPh9: We talked about this in MS Teams chat, seems like we are in agreement on making the top border an option as some times they may want it and other times may not.

src/tabs/tabs.scss
FlyersPh9: You are correct, removed.

src/tabs/pill.scss
mayank99: Does `-$iui-xxs` not work? The `#{}` syntax is usually not necessary for numeric operations like `-`.

FlyersPh9: `-$iui-xxs` appears to work on line 53 of `borderless.scss`, but does not work on line 46 of `pill.scss`.

Line 46 & results:
`box-shadow: inset 0 -$iui-xxs 0 0;`
Screen Shot 2021-06-16 at 3 13 22 PM

FlyersPh9: This appears to work: `box-shadow: inset 0 (-$iui-xxs) 0 0;`

src/tabs/borderless.scss
FlyersPh9: Is there any downside to using height?  By using height I can guarantee that the tab height = buttons, text fields, selects height.

backstop/tests/tabs.html
```diff
@@ -24,241 +24,832 @@
...
+              <path d="M8,0C3.6,0,0,3.6,0,8s3.6,8,8,8s8-3.6,8-8S12.4,0,8,0z M13,9H9v4H7V9H3V7h4V3h2v4h4V9z" />
+            </svg>
+          </button>
+        </li>
+      </ul>
+      <div
+        class="iui-tabs-content"
+        role="tabpanel"
+        id="active-tab"
+        aria-labelledby="active"
```
mayank99, line 125: There is no `"active"`. Did you forget to add the `id` to active tab?

src/tabs/pill.scss
FlyersPh9: Fixed it, using `::after { content: ' '; }` instead of `box-shadow` but only on pill tabs.  I could use `::after { content: ' '; }` on all tabs if we'd like to stay consistent.
Screen Shot 2021-06-16 at 4 06 13 PM

src/tabs/borderless.scss
```diff
@@ -4,49 +4,55 @@
...
-    > a:hover,
-    &.iui-tabs-active > a {
+    &.iui-active {
       @include themed {
         background-color: rgba(
-          t(iui-color-foreground-positive-rgb),
+          t(iui-color-foreground-primary-rgb),
           t(iui-opacity-6)
         );
+        box-shadow: inset 0 calc(#{$iui-xxs} * -1) 0 0;
```
mayank99, line 27: Seems you forgot to simplify this one

src/tabs/tabs.scss
```diff
@@ -4,100 +4,103 @@
...
   display: flex;
+  align-items: center;
   list-style: none;
   padding: 0;
-  margin: round($iui-baseline/2) 0 0 0;
   user-select: none;
 
   > li {
-    display: inline-block;
     position: relative;
```
mayank99, line 14: Is this needed?

src/tabs/pill.scss
mayank99: Check my other comment: https://github.com/iTwin/iTwinUI/pull/156#discussion_r652999113

If we change the style, all these hacks might not be necessary since we can just remove the border-radius

src/tabs/tabs.scss
mayank99: Previously, the `a` was positioned absolutely but now I think it can be removed

src/tabs/pill.scss
```diff
@@ -5,100 +5,62 @@
...
-        @include themed {
-          color: t(iui-color-foreground-primary);
-        }
-
-        &:hover {
-          @include themed {
-            color: t(iui-color-foreground-primary);
-          }
-        }
+  .iui-tab {
```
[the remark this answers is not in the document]
mayank99: @veekeys @bentleyvk Would `Tab` be a separate component or would it be similar to `SelectOption` where it's an object that can be part of a list (passed into `labels`)?

src/tabs/tabs.scss
FlyersPh9: I can move `position: relative;` off of the `<li>` and instead apply it to the `.iui-tab`.  Without it, the `::after` part does this.
Screen Shot 2021-06-16 at 4 46 52 PM

If I end up reverting the `::after` and going with `box-shadow` like originally I do think this can be removed then.

src/tabs/borderless.scss
FlyersPh9: You are correct, fixed.

src/tabs/borderless.scss
```diff
@@ -4,49 +4,55 @@
...
           t(iui-opacity-6)
         );
       }
     }
-  }
 
-  &.iui-green > li {
-    > a:hover,
-    &.iui-tabs-active > a {
+    &.iui-active {
```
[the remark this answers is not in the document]
FlyersPh9: I think that would be really cool on pill and borderless tabs, similar to [this](https://material-components.github.io/material-components-web-catalog/#/component/tabs).

src/tabs/pill.scss
```diff
@@ -5,100 +5,80 @@
...
-        bottom: $iui-xxs;
+        content: '';
+        position: absolute;
         left: $iui-xxs;
-        right: $iui-xxs;
-      }
-
-      &:hover {
+        bottom: $iui-xxs;
+        width: calc(100% - #{$iui-xs});
```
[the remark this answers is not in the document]
FlyersPh9: It's the same in [our button groups SCSS](https://github.com/iTwin/iTwinUI/blob/2de28c5b190c524d7f05b03cda8253e44fd255bf/src/button/button-group.scss#L70).  But yeah, I agree it's not very clean.

src/tabs/pill.scss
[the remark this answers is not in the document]
bentleyvk: I expect it to be a separate component as we accept `ReactNode` and we couldn't really do like `SelectOptions` without breaking changes.

src/tabs/pill.scss
bentleyvk: `::after` is not really needed here, just remove `border-radius` or set it to 0.

src/tabs/tabs.scss
bentleyvk: What about `padding` and `margin`?

src/tabs/borderless.scss
bentleyvk: Maybe it is ok to have controlled height. But then need to think whether we want to hide description when using default size (like in Header breadcrumbs) or it is ok to have default sized Tab with description.

src/tabs/borderless.scss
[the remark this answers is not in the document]
bentleyvk: If we want to have animation with a stripe then we need to bring back `::after` 😀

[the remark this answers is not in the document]
bentleyvk: This is the example of what we could do:
tab-animation
With this code:
image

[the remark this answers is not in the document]
mayank99: If we use a separate div for the active stripe, we could have an animation like material (from Jon's link).

src/tabs/borderless.scss
FlyersPh9: It's getting a little too tight to display a secondary line when in the default tab size.
Screen Shot 2021-06-17 at 12 42 46 PM

I've changed the CSS so the secondary line only displays when `.iui-large` is applied to the tab group.

src/tabs/tabs.scss
FlyersPh9: Doesn't look like it is needed either.  I was also able to remove `padding: 0;` from `.iui-tabs` because `@include iui-reset;` handles that.

src/tabs/tabs.scss
```diff
@@ -4,100 +4,103 @@
...
+    cursor: pointer;
+    transition: background-color $iui-speed-fast ease-out, box-shadow $iui-speed-fast ease-out;
+    @include themed {
+      color: t(iui-text-color);
+      box-shadow: inset 0 0 0 0 t(iui-color-foreground-primary);
+    }
 
+    &.iui-active {
+      cursor: default;
+      pointer-events: none;
```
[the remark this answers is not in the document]
FlyersPh9: If I remove `pointer-events: none;` & click on the green active tabs, it loses it's stripe.  Weird that it only seems to effect the green ones.  Also does not effect the pill tabs.
Screen Shot 2021-06-17 at 12 58 22 PM

src/tabs/borderless.scss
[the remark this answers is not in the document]
FlyersPh9: I think the animation from material design would look good on borderless & pill tabs, but would not look as nice on the default tabs because of the clear dividing lines.  I'd be ok with using Materials animation on borderless & pills but using Vilius animation on default.  Would that be too sloppy?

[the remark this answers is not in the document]
mayank99: they're different styles so different animations make sense

backstop/tests/tabs.html
```diff
@@ -698,212 +706,237 @@ <h2>Borderless</h2>
...
+          stripe.style.left = `${activeEl.offsetLeft}px`;
+          stripe.style.width = `${activeEl.offsetWidth}px`;
+        }
+        setStripe(activeTab);
+
+        [...tabs.children].forEach(({ firstElementChild: tab }) => {
+          tab.addEventListener('click', () => {
+            tabs.querySelector('.iui-active')?.classList.remove('iui-active');
+            tab.classList.add('iui-active');
+            setStripe(tab); // needs to be set in react for animation
```
mayank99, line 936: @veekeys @bentleyvk Need opinion about this. These inline styles may not be "nice" but the resulting animation is.

The alternative is to use css variables for `left` and `width`, that we can update in react.

src/tabs/borderless.scss
```diff
@@ -24,7 +24,7 @@
           t(iui-color-foreground-primary-rgb),
           t(iui-opacity-6)
         );
-        box-shadow: inset 0 (-$iui-xxs) 0 0;
+        //box-shadow: inset 0 (-$iui-xxs) 0 0;
```
mayank99, line 27: Leftover

src/tabs/pill.scss
```diff
@@ -35,49 +39,13 @@
         );
       }
     }
-
-    &.iui-active {
+
```
mayank99, line 42: Remove extra space from empty line

src/tabs/tabs.scss
```diff
@@ -18,19 +19,16 @@
...
     }
 
     &.iui-active {
-      cursor: default;
-      pointer-events: none;
       @include themed {
         color: t(iui-color-foreground-primary);
       }
-
+
```
mayank99, line 31: Extra space here too

mayank99 (commented):
Looks pretty good!

backstop/tests/tabs.html
bentleyvk: That animation look nice! I don't think there should be too much code to handle this in React.

backstop/tests/tabs.html
```diff
@@ -24,241 +24,919 @@
...
+    </section>
+
+    <script>
+      document.querySelectorAll('.iui-tabs:not(.iui-default)').forEach((tabs) => {
+        const wrapper = tabs.parentElement;
+        const activeTab = tabs.querySelector('.iui-active');
+        const stripe = wrapper.querySelector('.iui-tab-stripe');
+        const setStripe = (activeEl) => {
+          stripe.style.left = `${activeEl.offsetLeft}px`;
+          stripe.style.width = `${activeEl.offsetWidth}px`;
```
bentleyvk, line 928: These numbers not always match the real element:
image
As you can se on the right it goes outside the tab a bit. Because tab's width is `118.72px` so `offsetWidth` returns `119px`. We might need to use `getComputedStyle` or `getBoundingClientRect`.

src/tabs/default.scss
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  li:not(:first-child) .iui-tab {
+    border-left: none;
+  }
+
+  &.iui-green .iui-tab {
+    @include themed {
+      box-shadow: inset inherit inherit inherit inherit t(iui-color-foreground-positive);
```
bentleyvk, line 42: Is it ok that there is so many `box-shadow` rules that overrides each other?
image

src/tabs/tabs.scss
```diff
@@ -4,100 +4,139 @@
...
   }
+
+  .iui-tab-stripe {
+    display: none;
+  }
+}
+
+@mixin iui-tab-animated {
+  ~ .iui-tab-stripe {
+    width: 100%;
```
bentleyvk, line 122: If we are setting `width` and `left` dynamically, then I don't think we need to have rules for them in CSS.

src/tabs/classes.scss
```diff
@@ -2,18 +2,30 @@
...
+    @include iui-tabs-pill;
+  }
+
+  &.iui-animated {
+    @include iui-tab-animated;
+  }
 
-.iui-tabs-content-area {
-  @include iui-tabs-content-area;
+  &.iui-not-animated {
```
bentleyvk, line 28: Do we need this class? What if `iui-animated` would reset not animated values?

src/tabs/tabs.scss
mayank99: Removed

backstop/tests/tabs.html
mayank99: I changed width to use `getComputedStyle`, and it seemed better at first, but now I'm noticing different behavior on different zoom levels:

image
image

src/tabs/default.scss
mayank99: Removed. The default tabs don't the "not-animated" class (because there is no other variant, there is no stripe), so I've simplified it by setting the box-shadow and transition in default.

src/tabs/classes.scss
mayank99: But then there will be tons of extra box-shadow values being overridden, and specificity could also be a problem.

Also I think `not-animated` option should not really be used when `animated` has the preferred styles. But making `animated` default also doesn't make sense because it requires JS. So we end up with separate classes 🤷

backstop/tests/tabs.html
mayank99: Changed to `getBoundingClientRect`. But then I was zooming in and out, and noticed that this JS approach doesn't work well for pill tabs because those are using flex (their width doesn't change).

backstop/tests/tabs.html
```diff
@@ -24,241 +24,832 @@
...
+            </svg>
+            <span class="iui-tab-label">Active</span>
+          </button>
+        </li>
+        <li>
+          <button
+            class="iui-tab"
+            role="tab"
+            aria-selected="false"
+            aria-controls="left-tab"
```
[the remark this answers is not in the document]
FlyersPh9: I've added something like this to all of the demo tabs for each of the active tabs…unsure if I should add it for the disabled tab as well:
```
<div class="iui-tabs-content" role="tabpanel" id="left-tab" aria-labelledby="left" hidden>
  This is the left tab's content.
</div>
```
Do you know the required JS to make the tab content actually switch when the tab is clicked on?

src/tabs/tabs.scss
```diff
@@ -4,100 +4,151 @@
...
-  padding: 0;
-  margin: round($iui-baseline/2) 0 0 0;
   user-select: none;
 
-  > li {
-    display: inline-block;
+  .iui-tab {
+    @include iui-reset;
+    @include iui-font-family;
+    @include iui-focus;
```
FlyersPh9, line 16: Tab focus is working now that I'm using `::after { content` to create the active tab stripe.  However, we do have some clipping, also present in other parts of iTwinUI (header, grouped buttons, etc).  This could be fixed in #166.
Screen Shot 2021-06-22 at 6 28 22 PM

backstop/tests/tabs.html
bentleyvk: Yeah, we would need to observe resizes.

src/tabs/classes.scss
bentleyvk: Idk, then I would say to remove `not-animated` because there are some components that require JS e.g. `Toast`, `ExpandableBlock`.

src/tabs/classes.scss
```diff
@@ -2,18 +2,30 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import './index';
 
-.iui-tabs-horizontal {
-  @include iui-tabs-horizontal;
+.iui-tabs-wrapper {
```
bentleyvk, line 5: This class is only needed for the stripe that can be placed inside `iui-tabs`.

src/tabs/default.scss
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    &.iui-active {
+      border-bottom-color: transparent;
+      @include themed {
+        background-color: t(iui-color-background-1);
+      }
+
+      &::after {
+        height: 2px;
```
bentleyvk, line 42: Variable?

src/tabs/classes.scss
mayank99: It's only few lines and does not have tests. Could be useful for blazor or something, idk

src/tabs/classes.scss
mayank99: `iui-tabs` is a list, so placing stripe inside would not make sense and could mess up `:last-child`

src/tabs/default.scss
FlyersPh9: Switched to a variable.

mayank99 (approved):

PR #158, opened 2021-06-16
Fix column resizer activation

mayank99:
Changed `display: none`/`block` to use `opacity: 0`/`1`. This makes the resizer always be present in the DOM tree, allowing activating it when moving mouse from right to left.
resizer

After this, I was able to set z-index to 1 (the resizer needs to be on top of the next column which comes later in the tree).
Can't remove it completely because then it won't be activable when moving from right side:
resizer2

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -59,8 +59,8 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
-      z-index: 100;
-      display: none;
+      z-index: 1;
+      opacity: 0;
 
       > .iui-resizer-bar {
         height: 100%;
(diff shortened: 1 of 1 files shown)


src/table/table.scss
```diff
@@ -59,8 +59,8 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
-      z-index: 100;
-      display: none;
+      z-index: 1;
+      opacity: 0;
```
bentleyvk, line 64: It is a bit strange when your cursor is on right side column but left side column is highlighted. But I think this is minor as a cursor changes.

bentleyvk (approved):

src/table/table.scss
mayank99: It makes _some_ sense because you are resizing the left column. Any changes to the right column size would just be incidental (due to flex).

PR #163, opened 2021-06-18
Changed body background color

veekeys:
- Changed `<body>` background color from `$iui-color-background-1` to `$iui-color-background-2`.
- Tweaked fieldset background & border color to display better over background-2. *Discussed between @veekeys & @FlyersPh9.*


FlyersPh9:
How did only 2 files change?  Wouldn't nearly every single test image change?

bentleyvk:
> How did only 2 files change? Wouldn't nearly every single test image change?

I think there is a bug in BackstopJS that it is not checking or have very high threshold for color differences.

veekeys:
> > How did only 2 files change? Wouldn't nearly every single test image change?
> 
> I think there is a bug in BackstopJS that it is not checking or have very high threshold for color differences.

Yeah, so after changing the background color, only single test failed..
If I look at the passed ones, obviously there is a color change:
image

But that did not fail anywhere else.. So the only issue I have, all other images are going to be with wrong bacgkround color and maybe in future that could cause failures..
Ill try to play around to have all the test failing.

veekeys:
@FlyersPh9 
I have updated all images by deleting and running tests...
Mostly looks okeish I guess, but.. these components in particular did not look good on background-2
fieldset
fileupload
tags

So.. I am not sure, what we do.
- We could have background 1 in all the html files, just have background-2 on body in css applied.
- We can leave as I updated now (review please)
- We can do something else to make them look good on background-2.. So I am not sure, whats the intention. I know most of the components should appear on different background, but.. all this stuff is getting really confusing.

FlyersPh9:
> fieldset

Fieldset I anticipate only being used above background 1.  But if we can change the fieldset to use background-4 instead of background-3 so that it displays better.
Screen Shot 2021-07-07 at 11 41 13 AM

> fileupload

When file upload is needed above background-2, I am thinking it would be done similar to how [iTwin Design Review](https://review.itwin.bentley.com/home) handled it.
Screen Shot 2021-07-07 at 11 51 54 AM

When used above background-1, no changes needed.
Screen Shot 2021-07-07 at 11 52 46 AM

> tags

For tags, we have the `.iui-visible` class which adds a little bit of padding and changes the background color to 2.  If they needed to use this over background 2, they could either remove `.iui-visible` or override it to use background 1.
Screen Shot 2021-07-07 at 12 04 14 PM

I'd really like to set some time aside soon to work on [some fake full page UIs](https://vfble.csb.app/) so we can see all the UI components in context of the UI.  I think it will help us make visual design decisions.

src/style/global.scss
```diff
@@ -4,7 +4,7 @@
 
 .iui-body {
   @include themed {
-    background-color: t(iui-color-background-1);
+    background-color: t(iui-color-background-2);
```
veekeys, line 7: @FlyersPh9

FlyersPh9 (approved):
I made the adjustments that @veekeys and I spoke about on the phone to fieldset.  PR should be good to pass.

mayank99:
Need to remove these html overrides

https://github.com/iTwin/iTwinUI/blob/c68f2508dbaf301255542ae19bbae2f73d824eec/backstop/tests/header.html#L58-L59
https://github.com/iTwin/iTwinUI/blob/c68f2508dbaf301255542ae19bbae2f73d824eec/backstop/tests/side-navigation.html#L25-L26

mayank99 (approved):

PR #164, opened 2021-06-18
Add user icon lists & SVG support

FlyersPh9:
- Add user icon lists (animated and non animated options)
- Add support for SVG to be used in place of initials.

127 0 0 1_5500_backstop_minified_user-icon html


src/user-icon/user-icon.scss
```diff
@@ -116,3 +127,54 @@
     }
   }
 }
+
+@mixin iui-user-icon-list {
+  display: flex;
+
+  .iui-user-icon:not(:first-child),
+  > .iui-tooltip-container > .iui-user-icon {
```
mayank99, line 135: This is not necessary. Tooltip content can just be placed in the main user list element (after the last user-icon).

src/user-icon/user-icon.scss
```diff
@@ -23,13 +23,15 @@
...
+      box-shadow: inset 0 0 0 $iui-xxs
         rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-6));
     }
   }
 
   > .iui-initials {
-    display: block;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
mayank99, line 34: You can use `place-content: center`

backstop/scenarios/user-icon.js
```diff
@@ -1,3 +1,13 @@
...
+
+module.exports = [
+  // User icon types
+  scenario('Type default', {
+    selectors: ['#demo-default'],
+  }),
+  scenario('Type online status', {
+    selectors: ['#demo-status'],
+  }),
+];
```
mayank99, line 13: Forgot tests?

backstop/tests/user-icon.html
```diff
@@ -24,15 +24,46 @@
       #demo-status > div > span {
         margin-right: 4px;
       }
+
+      .right {
```
mayank99, line 28: Call it demo-right to avoid confusion

src/user-icon/user-icon.scss
FlyersPh9: Are you sure?  I'm getting this after replacing line 34:
Screen Shot 2021-06-18 at 2 36 41 PM

mayank99: Ah nevermind, I thought you used `justify-content` and `align-content` (not `align-items`).

src/user-icon/user-icon.scss
FlyersPh9: I've removed line 135 from the CSS and changed the HTML so the last user icon is no longer wrapped with the tooltip container.  I'm not sure I understand how to make the tooltip appear on hover over the last user icon tho.

backstop/scenarios/user-icon.js
FlyersPh9: Added tests but did not add hover tests yet.

backstop/tests/user-icon.html
FlyersPh9: Done.

backstop/scenarios/user-icon.js
mayank99: the animated one definitely needs hover, otherwise it's a pointless test

FlyersPh9: Added test for animated hover on large size.

src/user-icon/user-icon.scss
mayank99: Tooltips are easy in JS (We don't even use `.iui-tooltip-container` in React)

If really want, you could do some CSS hacks here (in demo html), like positioning the tooltip container over the "+5" icon (and giving it some random static width/height). No need to make it perfect.

mayank99: Or you can add this to the end of the html
```html
    <script>
      document.querySelectorAll('#demo-list .iui-user-icon-list').forEach((list) => {
        const lastIcon = list.querySelector('.iui-user-icon:nth-last-child(2)');
        const tooltip = list.querySelector('.iui-tooltip');
        tooltip.style.visibility = 'visible';
        lastIcon.addEventListener('mouseenter', () => { tooltip.style.opacity = 1; });
        lastIcon.addEventListener('mouseleave', () => { tooltip.style.opacity = 0; })
      });
    </script>
```

backstop/scenarios/user-icon.js
mayank99: Is the small size supposed to look like this?
image

src/user-icon/user-icon.scss
```diff
@@ -43,9 +45,18 @@
     text-decoration: none;
     text-transform: uppercase;
     background-color: $iui-color-dataviz-celery;
+    overflow: hidden;
     @include themed {
       color: t(iui-color-foreground-accessory);
     }
+
+    > .iui-icon {
```
bentleyvk, line 53: Why SVGs are placed differently than img? I don't think we can handle that in React properly.

src/user-icon/user-icon.scss
```diff
@@ -116,3 +127,53 @@
...
+
+  .iui-user-icon:not(:first-child) {
+    margin-left: -$iui-xs;
+
+    &.iui-small {
+      margin-left: -$iui-xxs;
+    }
+
+    &.iui-large {
+      margin-left: -6px;
```
bentleyvk, line 142: Why not `$iui-s` (`8px`)?

src/user-icon/user-icon.scss
mayank99: I added it, and it works

src/user-icon/user-icon.scss
FlyersPh9: Large is now using `$iui-s` and x-large is now using `$iui-sm`.

backstop/scenarios/user-icon.js
FlyersPh9: Good catch.  I've fixed this by changing the stroke around the small user icons.  All other sizes have the same thickness stroke.  I tried scaling the stroke with the size of the user icon & it looked like too much on the larger sizes.  Here's what that would have looked like:
Screen Shot 2021-06-22 at 12 08 16 PM

src/user-icon/user-icon.scss
FlyersPh9: I've updated the code so the `<svg>` is placed in the same location as the ``.  The only other difference is the `<abbr>` changes to a `<div>` for the SVG user icons.

mayank99: Why does `abbr` change to `div`?

Also, cant you use `display: inline-flex` to center the svg? Instead of this hacky transform

FlyersPh9: I replaced the `abbr` with `div` because I thought it made sense, we are no longer using that element to display initials.  I suppose it wouldn't hurt to stay as `abbr` and display no content within it. I'll change them back to `abbr`.

Regarding the `inline-flex`, I tried that plus removing `absolute` from `.iui-icon` and am getting something like this.  Any idea why?
Screen Shot 2021-06-22 at 6 14 30 PM

src/user-icon/user-icon.scss
```diff
@@ -116,3 +131,59 @@
...
   }
 }
+
+@mixin iui-user-icon-list {
+  display: flex;
+
+  .iui-user-icon:not(:first-child) {
+    margin-left: -$iui-xs;
+
+    .iui-initials {
```
bentleyvk, line 141: Do we really need to override these values?

src/user-icon/user-icon.scss
mayank99: I think the `position: absolute` needs to stay

src/user-icon/user-icon.scss
FlyersPh9: This `box-shadow` override is only needed for the smallest size because the shadow was larger than the margin between circles.  So even when the animated user icons were hovered, they would still slightly cover each other.  Mayank left a screenshot in a previous comment that has since been resolved.

mayank99 (approved):

Issue #85, opened 2021-05-14
Menu: multiline list items


danleydmello:
#### Environment
- __Package version(s)__: 0.13.1

If you look at Itwin.bentley.com or plansight the typical header dropdown has multiline list items:
![image](https://user-images.githubusercontent.com/16119774/118290246-357db880-b4a4-11eb-9b99-ef4b921d3f94.png)

I  tried adding multiple divs inside the <span> but it shows only the first <div>:
  ```
<ul class="iui-menu">
            @if (_items != null)
            {
                foreach (var item in _items)
                {
                    <li @onclick="() => ItemClicked(item)">
                        <span class="iui-content">
                            **<div >
                                @item.ItemName
                            </div>
                            <div>
                                @item.ItemDescription
                            </div>**
                        </span>
                    </li>
```

mayank99:
multi-line menu items are now available in 0.24.0 🎉
https://itwin.github.io/iTwinUI/backstop/minified/menu.html

PR #80, opened 2021-05-13
Table refactor

FlyersPh9:
- Class name refactor.
- Table header is now wrapped within `.iui-table-header`.
- ~~Table can support multiple headers.~~
- Added column resize grab handles in the header following [React Table examples](https://react-table.tanstack.com/docs/examples/column-resizing).
- Using borderless button through out the table & paginator.
- Paginator moved into it's own SCSS file.
- Paginator moved outside of the `.iui-table` container to allow overflow X to scroll if desired (not default setting). _I have not tried this in the demo page yet._
- Added status colors to rows & individual cells.
- ~~Added `.iui-editable` to `.iui-cell` for individually editable cells.~~
- Selected rows next to each other style the dividing line differently .
- Added 2 different types of expanded rows (link [1](https://react-table.tanstack.com/docs/examples/expanding) & [2](https://react-table.tanstack.com/docs/examples/sub-components)):
  - ~~Rows that expand indented rows.~~
  - Rows that expand empty space, not continuing the table's column structure (no transition for now)

127 0 0 1_5500_backstop_minified_table html

PR #167, opened 2021-06-23
Add menu divider, secondary line, and visual tweaks

FlyersPh9:
- Added menu divider.
- Added secondary line support.
- Visual changes to disabled line & focus.
- All sizes of select, text field, text area have matching horizontal padding so that the icon or label aligned with the menu.  This allows us to remove `$iui-component-padding-horizontal-small` and `$iui-component-padding-horizontal-large`.
- Special `.iui-header-menu-icon` modifier applied to `.iui-icon`, gives the extra spacing to align with header button's 24px thumbnail area.

Screen Shot 2021-06-30 at 5 10 45 PM


src/menu/menu.scss
```diff
@@ -94,6 +119,15 @@
       }
     }
   }
+
+  > .iui-divider {
```
mayank99, line 123: I would call this `iui-menu-divider` or something, and move it outside. It's specificity should be very low

src/menu/menu.scss
```diff
@@ -94,6 +119,15 @@
...
+    height: 1px;
+    width: calc(100% - #{$iui-l});
```
mayank99, line 125: You don't need `#{}`

src/menu/menu.scss
```diff
@@ -58,13 +68,24 @@
...
+        outline: 1px solid t(iui-color-foreground-primary);
+        outline-offset: -1px;
+      }
+    }
+  
+    &:focus:not(:focus-visible) {
       outline: none;
     }
+  
+    &:focus-visible {
```
mayank99, line 81: This is not necessary

src/menu/menu.scss
```diff
@@ -78,13 +99,17 @@
           fill: t(iui-icons-color-primary);
         }
       }
+
+      &:focus,
+      &:focus-visible {
+        outline-offset: -2px;
```
mayank99, line 105: use variable for 2px.

Also, `:focus, :focus-visible` breaks compatibility with older browsers. I don't think you even need focus-visible here

src/menu/menu.scss
```diff
@@ -36,19 +37,28 @@
...
+        }
+      }
 
       + .iui-icon {
         margin-right: 0;
         margin-left: $iui-s;
       }
+
+      &,
+      > * {
```
mayank99, line 54: Not sure about `*`

src/menu/menu.scss
```diff
@@ -19,13 +19,14 @@
   margin: 0;
   padding: 0;
 
-  > li {
+  > li:not(.iui-divider) {
```
mayank99, line 22: This increases the specificity 🤔

src/menu/menu.scss
```diff
@@ -36,19 +37,28 @@
 
     > .iui-content {
       flex: 1 1 auto;
-      white-space: nowrap;
-      overflow: hidden;
-      text-overflow: ellipsis;
-      height: $iui-line-height;
+
+      .iui-menu-description {
```
mayank99, line 41: I don't think this needs to be nested

src/menu/menu.scss
```diff
@@ -19,13 +19,14 @@
   margin: 0;
   padding: 0;
 
-  > li {
+  > li:not(.iui-divider) {
     display: flex;
     align-items: center;
     font-size: $iui-font-size;
-    padding: 0 $iui-sm;
+    padding: $iui-component-padding-vertical $iui-sm;
```
mayank99, line 26: How come it was 0 before?

mayank99 (commented):
@FlyersPh9  Before any further review, can you verify the visuals? The spacing just looks off to me.

src/menu/menu.scss
FlyersPh9: I think it is required, notice the width of the dividing line.  With `#{ }`:
Screen Shot 2021-06-28 at 11 06 41 AM

Without `#{ }`:
Screen Shot 2021-06-28 at 11 07 08 AM

src/menu/menu.scss
FlyersPh9: Renamed to `iui-menu-divider`.  Also introduced `iui-menu-item` so I can target the class directly rather than targeting the `li` element.  This allows me to remove the part that says `li:not(.iui-menu-divider)`.

src/menu/menu.scss
FlyersPh9: Removed.

src/menu/menu.scss
FlyersPh9: Using the variable and removed focus-visible.

src/menu/menu.scss
FlyersPh9: Added a `.iui-menu-item` class so I can avoid using this `:not`.

src/menu/menu.scss
mayank99: Oh right, because it's inside `calc`

src/menu/menu.scss
mayank99: The specificity is still high. I'd move both of those things out into their own flat mixins, and declare in classes.scss

src/menu/menu.scss
mayank99: The specificity went from 0-1-1 to 0-2-1 to 0-2-0.

If you remove nesting, it will be 0-1-0.

src/menu/menu.scss
FlyersPh9: Reverted back to 0, using component height vars instead.  Does now require `.iui-large` to be applied to enable the secondary line within a menu item.

src/menu/menu.scss
FlyersPh9: Is no longer nested.

src/menu/menu.scss
FlyersPh9: Removed this, gave every nested div a class name instead.

src/menu/menu.scss
FlyersPh9: > @FlyersPh9 Before any further review, can you verify the visuals? The spacing just looks off to me.

Since replacing the padding with a fixed height variable, the menu item height should match the height of the medium & large select component.

mayank99:
> > @FlyersPh9 Before any further review, can you verify the visuals? The spacing just looks off to me.
> 
> Since replacing the padding with a fixed height variable, the menu item height should match the height of the medium & large select component.

image

I think this looks even worse now 😄 There is more vertical space between the label and description than between them and the edges.

Also, what about the horizontal space between the icon and text? Probably should be higher for a larger menu item?

And what about that last menu item? I think that one looks out of place. Maybe the divider needs some margin above and below?

FlyersPh9:
> I think this looks even worse now 😄 There is more vertical space between the label and description than between them and the edges. Looks better when there's no description though.

I am trying to match the line height used tabs. I inspected the CSS for tabs and didn't realize it was inheriting the browser default styling `line-height: normal`, so after applying that here it looks to match better.

Screen Shot 2021-06-28 at 3 07 18 PM

mayank99:
Looks much better now (vertically).

I would suggest changing the right margin of icon from 8px to 12px (so it aligns with the left padding of menu item). I think that will make it look good on all sizes:
| 12px | 8px |
| --- | --- |
| image | image |

FlyersPh9:
> I would suggest changing the right margin of icon from 8px to 12px (so it aligns with the left padding of menu item). I think that will make it look good on all sizes

I made some changes so the first icon is indented to match the indentation of the select component.  Screenshots:
Screen Shot 2021-06-29 at 9 51 30 AM
Screen Shot 2021-06-29 at 10 14 15 AM
Screen Shot 2021-06-29 at 10 14 04 AM

bentleyvk:
> > I would suggest changing the right margin of icon from 8px to 12px (so it aligns with the left padding of menu item). I think that will make it look good on all sizes
> 
> I made some changes so the first icon is indented to match the indentation of the select component. Screenshots:
> Screen Shot 2021-06-29 at 9 51 30 AM
> Screen Shot 2021-06-29 at 10 14 15 AM
> Screen Shot 2021-06-29 at 10 14 04 AM

Is it really that bad to have the same spacing? For me it looks ok and requires less code. Only the dropdown button and the borderless select don't align with the menu but I don't think it would look nice with such small paddings.
image

FlyersPh9:
> Is it really that bad to have the same spacing? For me it looks ok and requires less code. Only the dropdown button and the borderless select don't align with the menu but I don't think it would look nice with such small paddings.
> image

Latest push makes selects, text fields, text areas, of all sizes use the same horizontal padding so that icons & labels now align with the menu items.

src/header/header.scss
```diff
@@ -219,6 +219,10 @@
 
     > .iui-icon {
       flex-shrink: 0;
+
+      &:not(.iui-header-button-icon) {
```
bentleyvk, line 223: Are these changes really necessary for this PR?

src/inputs/input.scss
```diff
@@ -21,12 +21,14 @@
   }
 
   &.iui-small {
-    padding: $iui-component-padding-vertical-small $iui-component-padding-horizontal-small;
+    padding-top: $iui-component-padding-vertical-small;
```
bentleyvk, line 24: Why removed left and right padding?

src/menu/menu.scss
```diff
@@ -19,89 +19,141 @@
...
+@mixin iui-menu-divider {
+  height: 1px;
+  width: calc(100% - #{$iui-l});
+  margin: auto;
+  @include themed {
+    background-color: t(iui-color-background-4);
+  }
+}
+
 @mixin iui-menu-icon {
```
bentleyvk, line 148: Spacing is missing.
image

src/style/variables.scss
```diff
@@ -13,8 +13,6 @@ $iui-component-height-small: $iui-baseline * 3 - $iui-component-offset * 2; // 2
 $iui-component-height-large: $iui-baseline * 5 - $iui-component-offset * 2; // 49px
 
 $iui-component-padding-horizontal: $iui-sm; // 12px
-$iui-component-padding-horizontal-small: $iui-sm * 0.5; // 6px
```
bentleyvk, line 16: Are these really not needed anymore?

bentleyvk:
> > Is it really that bad to have the same spacing? For me it looks ok and requires less code. Only the dropdown button and the borderless select don't align with the menu but I don't think it would look nice with such small paddings.
> > image
> 
> Latest push makes selects, text fields, text areas, of all sizes use the same horizontal padding so that icons & labels now align with the menu items.

It is hard to tell whether these changes are good or not 😁 It is good because default size select, button and inputs content aligns vertically but other sizes don't so I am not sure if it is worth changing.
What do you think @veekeys @mayank99

veekeys:
> > > Is it really that bad to have the same spacing? For me it looks ok and requires less code. Only the dropdown button and the borderless select don't align with the menu but I don't think it would look nice with such small paddings.
> > > image
> > 
> > 
> > Latest push makes selects, text fields, text areas, of all sizes use the same horizontal padding so that icons & labels now align with the menu items.
> 
> It is hard to tell whether these changes are good or not 😁 It is good because default size select, button and inputs content aligns vertically but other sizes don't so I am not sure if it is worth changing.
> Looked at other UI libs and they are not matching spacing of dropdown menu with button or other components only with select.
> What do you think @veekeys @mayank99

I think we should have added here only divider and secondary line.. not change the sizes of it and other components too...
If that is really needed, we could do it in separate PR. Now if its really needed, I dont know. Would need to research and hear @FlyersPh9 cent here..

src/style/variables.scss
FlyersPh9: They are needed if we want different left / right padding for small, medium, large sizes.  If we choose to have all 3 sizes match like the PR does currently, the large & small variables aren't needed anymore.

src/menu/menu.scss
FlyersPh9: Yikes, that's embarrassing.  It is fixed with the next push.

src/header/header.scss
FlyersPh9: We could remove this if it will help it pass.  Without this, the icons do not align.  We also won't be able to show a photo thumbnail, we'll only have icons for now.

**With this CSS:**
header-menu-with

**Without this CSS:**
header-menu-without

Worth pointing out tho that [the initial request / bug filed](https://github.com/iTwin/iTwinUI/issues/85) is using the double line menu items from the context of the header button.

src/inputs/select.scss
```diff
@@ -28,7 +28,7 @@ $iui-after-border: 5px solid;
       @include iui-menu-icon;
     }
 
-    > .iui-content {
+    > .iui-label {
```
mayank99, line 31: Why was this renamed?

src/menu/menu.scss
```diff
@@ -41,6 +39,7 @@
   cursor: pointer;
   box-sizing: border-box;
   line-height: normal;
+  border: 1px solid transparent;
```
mayank99, line 42: Flashbacks from #80 😬

backstop/tests/menu.html
```diff
@@ -29,162 +40,1879 @@
...
+            class="iui-icon"
+            viewBox="0 0 16 16"
+            aria-hidden="true"
+          >
+            <path
+              d="m6.00263 4.80681v6.39946a.27938.27938 0 0 0 .44292.24059l3.45479-3.17567a.34022.34022 0 0 0 .00046-.4807l-.00046-.00044-3.45479-3.22382a.269.269 0 0 0 -.44292.24058z"
+            />
+          </svg>
+        </li>
+        <li class="iui-menu-divider"></li>
```
mayank99, line 332: Add `aria-hidden` to dividers

backstop/tests/menu.html
```diff
@@ -29,162 +40,1879 @@
...
-              dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
-              ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
-              fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-              mollit anim id est laborum.
-            </span>
-            <!-- Use checkmark.svg from @itwin/itwinui-icons package. -->
+            <span class="iui-label">App name</span>
+          </div>
+
+          <div class="iui-divider"></div>
```
mayank99, line 551: Oh, I forgot about this class 😄

src/inputs/select.scss
```diff
@@ -68,7 +68,7 @@ $iui-after-border: 5px solid;
       }
     }
 
-    &.iui-active {
+    &.iui-expanded {
```
mayank99, line 71: Do not like this

src/menu/menu.scss
```diff
@@ -19,89 +19,140 @@
...
+
+  .iui-label,
+  .iui-menu-label,
+  .iui-menu-description {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .iui-label {
```
mayank99, line 55: Need to avoid these generic names

mayank99 (commented):
Looking a lot better now! But I don't like some of the class names

src/inputs/select.scss
FlyersPh9: Button, select, menu, header button are now all aligned to use `.iui-icon + .iui-label`.  We could re-use the `@include iui-menu-icon` across more components now.

backstop/tests/menu.html
FlyersPh9: `aria-hidden="true"` added to all `.iui-menu-divider`s.

src/inputs/select.scss
bentleyvk: I also don't like this rename. Because now we have `label` inside `label`, `content` makes more sense here.
image

backstop/scenarios/menu.js
```diff
@@ -1,7 +1,26 @@
-const { scenario } = require('../scenarioHelper');
+const { scenario, hover, focus } = require('../scenarioHelper');
 
 module.exports = [
-  scenario('basic', {
+  // Menu types
+  scenario('Type Default', {
+    selectors: ['#demo-default'],
+  }),
+  scenario('Type Header', {
```
bentleyvk, line 8: Do we really want to have tests for context? Because I really like this idea of having context but maybe don't flood our tests.

src/style/variables.scss
bentleyvk: I asked because you have used multiplication in buttons:
image

So maybe we need to have `small` and `large` variables here just to avoid different sizing in the future.

src/header/header.scss
bentleyvk: Huh, how this affects alignment?

src/inputs/select.scss
bentleyvk: Don't start refactoring...

backstop/tests/menu.html
```diff
@@ -29,162 +40,1915 @@
...
+          </span>
+        </li>
+      </ul>
+
+    </section>
+
+    <hr />
+
+    <h2>In context</h2>
+    <h3>Header</h3>
```
bentleyvk, line 351: Does not align.
image

src/button/default.scss
```diff
@@ -43,6 +43,15 @@
     &.iui-large {
       padding-right: $iui-sm;
     }
+
+    // Flipping the expand arrow
```
bentleyvk, line 47: Unrelated with current PR.

src/inputs/select.scss
FlyersPh9: Reverted back to `.iui-content` for the select component.  It should probably be renamed in another PR to more closely align with `.iui-menu-item-content`.

The menu icon mixin is now targeting `[class^='iui-'][class$='-content']` which should effect `.iui-menu-item-content`, `.iui-content` (as it is now), and `.iui-select-content` if / when we decide to rename that class.

src/inputs/select.scss
FlyersPh9: Reverted class name change and removed addition of `.iui-expanded` from button and header button.

src/menu/menu.scss
FlyersPh9: Renamed to `.iui-menu-item-content`.

src/button/default.scss
FlyersPh9: Removed all of this.

backstop/scenarios/menu.js
FlyersPh9: I've removed the tests for selects & dropdown buttons because the menu itself was the same as the default test.  I left the header test in because it is styled slightly different to get the image / icon to align properly.

backstop/tests/menu.html
FlyersPh9: I can't figure out why the test images have the menu's not aligned properly with the header buttons.  They look fine in the demo before:
Screen Shot 2021-07-07 at 4 37 26 PM

They menus are placed using `margin-left` inline and then some random value to make it look like they align.

src/header/header.scss
FlyersPh9: Sorry, I must have misunderstood what the initial comment was questioning.

I've removed this CSS since we also removed `.iui-expanded` from the header button.

src/style/variables.scss
FlyersPh9: I've added those variables back in exclusively for the button to use now.

backstop/tests/menu.html
bentleyvk: For me it is not only in tests but also when I open HTML

bentleyvk: Fixed menu placement, for some reason margin should change based on screen size.

src/inputs/select.scss
bentleyvk: I reverted back to `iui-content`. I hope Jon won't be mad about this 🤞

bentleyvk (approved):

src/inputs/select.scss
FlyersPh9: I'm not mad, I'm just happy this PR is finally about to pass! ☺️

mayank99:
> I would suggest changing the right margin of icon from 8px to 12px (so it aligns with the left padding of menu item). I think that will make it look good on all sizes:
> 
> | 12px | 8px |
> | --- | --- |
> | image | image |

So we don't want to make this change? And yet we are changing all the buttons for no reason?




Also we need to fix multi-select borders at some point.
image

mayank99 (commented):
I added some JS to toggle selection (it's not perfect but it does the job) and found a bug with focus. Other than that, this PR is good to go!

src/menu/menu.scss
```diff
@@ -18,90 +18,148 @@
...
     }
+  }
 
-    > .iui-icon {
-      @include iui-menu-icon;
+  > .iui-icon {
+    @include iui-menu-icon;
+  }
+
+  &:hover:not(.iui-disabled):not(.iui-active) {
```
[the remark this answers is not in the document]
FlyersPh9: I moved `iui-disabled` above this line.  I am able to simplify the selector to `&:hover:not(.iui-disabled)`.  If I remove the not disabled part, you'll see the background color change on hover.
Screen Shot 2021-07-12 at 9 48 05 AM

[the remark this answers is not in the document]
FlyersPh9: Oh no, wait… if I remove `:not(.iui-active)` the hover overrides the active state.  
hover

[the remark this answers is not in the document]
mayank99: 😦 Okay, I guess leave it

src/button/button.scss
```diff
@@ -62,23 +62,11 @@ $iui-button-padding-large: $iui-xs * 6;
...
 
   &.iui-large {
     padding: 0 $iui-button-padding-large;
     height: $iui-component-height-large;
     font-size: $iui-font-size-leading;
-
-    > .iui-icon + .iui-label,
-    > .iui-label + .iui-icon,
-    > .iui-icon + .iui-icon {
-      margin-left: $iui-sm;
```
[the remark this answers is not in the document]
FlyersPh9: Latest push reverts this.

mayank99 (approved):

PR #169, opened 2021-06-25
Animate expandable table rows

FlyersPh9:
- When a table row expands / collapses it animates.
- Fixed a bottom border of a selected expanded row.
- All row borders transition.

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -134,6 +134,7 @@
     display: flex;
     flex-grow: 1;
     border: solid 1px transparent;
+    transition: border $iui-speed-fast ease-out;
     @include themed {
       border-bottom-color: t(iui-color-background-4);
     }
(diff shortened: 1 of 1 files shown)


src/table/table.scss
```diff
@@ -161,6 +161,20 @@
...
     }
 
+    &.iui-expanded-content {
+      transform: scaleY(0);
+      transform-origin: top;
+      transition: transform $iui-speed-fast ease-in-out;
+      overflow: hidden;
+      opacity: 0;
+      height: 0;
+      margin-top: -2px;
```
FlyersPh9, line 171: Unsure why this is needed.

src/table/table.scss
```diff
@@ -161,6 +161,20 @@
       visibility: visible;
     }
 
+    &.iui-expanded-content {
+      transform: scaleY(0);
+      transform-origin: top;
+      transition: transform $iui-speed-fast ease-in-out;
+      overflow: hidden;
+      opacity: 0;
```
mayank99, line 169: This approach will not work.

This is what the DOM currently looks like:
table

I'm making some changes in React that should give you classes for animating.

FlyersPh9: We collaborated and are now using the correct classes.

mayank99 (approved):
Works perfectly with https://github.com/iTwin/iTwinUI-react/pull/169 🎉

src/table/table.scss
```diff
@@ -180,6 +186,24 @@
       }
     }
 
+    &.iui-expanded-content {
```
bentleyvk, line 189: Can we have this in HTML as an example?

mayank99: These classes come from react-transition-group, so I think not

PR #172, opened 2021-06-29
Add slider component

bsteinbk:
- Initial implementation of Progress (single thumb) and Range (two thumb) sliders.
- Backstop slider.html file that demonstrates capabilities.
- Note: I did not attempt to support IE,


mayank99:
Since the original PR was already reviewed on Azure, it would be nice to have a list of changes made since then, including any design quirks.

bsteinbk:
> Since the original PR was already reviewed on Azure, it would be nice to have a list of changes made since then, including any design quirks.

There is not too much remaining from the original azure code, but it is definitely what I started with. Here are some of the things I remembered hitting with the original code on azure.
1. If Firefox the thumbs worked fine but in Chrome once the track height was changed the mouse pointer needed to be within the width of the track before you could actually hold the mouse button down to drag it to a new location.
2. If you added an additional div wrapper with display of flex the vertical position of the slider would not be consistent between Firefox and Chrome.
3. The original design did not display nicely when tick marks are added due to how the thumbs are positioned on track. The only way to get the tick marks to line up was to position them inset from the track 1/2 the thumb diameter and that was not very pleasing to the eye.
4. The two thumb "range" slider was not working or displaying for me when using code from Azure PR.

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+@import '../icon/index';
+
+$thumb-height: $iui-m;  // 16px
+$track-height: $iui-xs;  // 4px
+$tick-height: $iui-s+$iui-xxs;
+
+@mixin iui-slider-component-container {
+  display: inline-flex;
+
+  span.iui-left,
```
veekeys, line 13: Is there a reason span is used for css rules? It increases specificity which is not what we might want

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+  span.iui-right {
+    @include iui-font-family;
+    user-select: all;
+    margin-top: $iui-xxs;
+
+    .iui-icon {
+      @include iui-icons-default;
+      margin-top: 4px;
```
veekeys, line 21: Use variable

backstop/tests/slider.html
```diff
@@ -0,0 +1,349 @@
...
+      href="../assets/demo.css"
+    />
+    <style>
+      section[id^='demo'] {
+        width: auto;
+      }
+
+      /* Gives left & right labels an equal width so the slider is centered. */
+      .iui-left,
+      .iui-right {width: 25px;}
```
veekeys, line 30: Which part of centering you mean?
Cause disabling it just makes slider full width and aligned with title:
image

If something is still needed, would be good to tackle not direct classes, cause then its confusing which part is coming from styles, which part is hardcoded for demo purposes...

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+$thumb-height: $iui-m;  // 16px
+$track-height: $iui-xs;  // 4px
+$tick-height: $iui-s+$iui-xxs;
```
veekeys, line 8: spacing around '+'

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+$thumb-height: $iui-m;  // 16px
```
veekeys, line 6: Not sure about variables which refer just a single other variable...
Could have used `$iui-m` just directly.. Unless it gives really a better understanding of the code..?

src/slider/slider.scss
veekeys: or actually..Better apply flex to the container and `align-items: center` and this margin would be not needed anymore..

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+$track-height: $iui-xs;  // 4px
+$tick-height: $iui-s+$iui-xxs;
+
+@mixin iui-slider-component-container {
+  display: inline-flex;
```
veekeys, line 11: Reasoning for using `inline-flex`?
If it was flex, wouldnt need to apply 100% width in htmls...
I am wondering, if we could get rid of container.. Should be possible to place 3 elements in one line without it. Just then width 100% is not really correct value on the slider..

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+    text-align: right;
+  }
+
+  span.iui-right {
+    margin-left: $iui-s;
+  }
+}
+
+@mixin iui-slider-container {
+  display: flex;
```
veekeys, line 36: Is flex actually needed? I removed it and visually nothing changed...
If not needed, we could get rid of flex-direction and justify-content too (they do nothing also...)

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+  position: relative;
+  z-index: 1; // above ticks
+  width: 100%;
```
veekeys, line 39: As I mentioned above, this value is not really correct.. cause in same container there are left and right elements, so this one just cannot take 100% of whole container..
It works, cause container has flex applied..so kinda hacky

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+  pointer-events: none;
+  flex-direction: column;
+  justify-content: center;
+
+  .iui-tooltip {
+    display: none;
```
veekeys, line 45: hmm.. I dont think this is needed in the css (can be in html). Cause when implementing, we will be handling the tooltip in react side anyway (so this should not be needed...)

src/slider/slider.scss
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  // hidden input[type=range] that is used to process events and hold value
+  .iui-slider {
```
veekeys, line 49: so if it is supposed to be hidden, cant we just apply same rules like in other custom input elements (checkbox, radio, etc...)?
Now there is a bunch of those and seems only appearance does the difference..

veekeys: oh, i see for it to work need to have input same as the visual one (meaning 100% width, so it covers whole slider and can react to events..)

veekeys:
hmm..
Slider does not work in IE, at all.. and it wont, if we want to use native input element.
I was checking also the library imodel.js uses and it is done without native slider. This gives ability to fully control stuff and add what ever we want versus fighting with the limitation of native input (now range slider needs to use actually 2 elements...). Even w3 example is not using native input element https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html  We could take idea from it, apply all the accessibility stuff, etc..
So I am thinking, should we just do the styling here with simple divs and implement functionality itself in js?
Especially, if there is going to be some more advanced stuff required, we would not need to refactor everything again...

image

src/slider/slider.scss
bsteinbk: align-items: center does not look nice when tick marks are displayed. I like the text/icons to be aligned with slider track.

src/slider/slider.scss
bsteinbk: I used it specifically to increase specificity, as there are  other items that use `iui-left` that are decedents of `iui-slider-component-container` that I do not want this rule applied.

src/slider/slider.scss
veekeys: Oh yeah.. I didnt notice it before.. Its because span takes full height (including marks...).

src/slider/slider.scss
bsteinbk: I like it because is make it clear when I use something like, $thumb-height * 0.5, exactly what I am trying to achieve. If we decide to make $thumb-height = $iui-m + $iui-xxs; nothing else has to change.

src/slider/slider.scss
bsteinbk: Don't you want your css generic so it can be use with or without iTwinUI-react?

src/slider/slider.scss
bsteinbk: > hmm..
> Slider does not work in IE, at all.. and it wont, if we want to use native input element.
> I was checking also the library imodel.js uses and it is done without native slider. This gives ability to fully control stuff and add what ever we want versus fighting with the limitation of native input (now range slider needs to use actually 2 elements...). Even w3 example is not using native input element https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html We could take idea from it, apply all the accessibility stuff, etc..
> So I am thinking, should we just do the styling here with simple divs and implement functionality itself in js?
> Especially, if there is going to be some more advanced stuff required, we would not need to refactor everything again...
> 
> image

@FlyersPh9 @veekeys - That is definitely an option.  I was basing my implementation on what was done in the original Azure repo.  We need to make a decision on which way to proceed as soon as possible. We keyboard and touch processing for free with the current implementation. If we go to totally js solution does that need to be implemented in both backstop and iTwinUI-react?

bsteinbk:
> hmm..
> Slider does not work in IE, at all.. and it wont, if we want to use native input element.
> I was checking also the library imodel.js uses and it is done without native slider. This gives ability to fully control stuff and add what ever we want versus fighting with the limitation of native input (now range slider needs to use actually 2 elements...). Even w3 example is not using native input element https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html We could take idea from it, apply all the accessibility stuff, etc..
> So I am thinking, should we just do the styling here with simple divs and implement functionality itself in js?
> Especially, if there is going to be some more advanced stuff required, we would not need to refactor everything again...
> 
> image

@FlyersPh9 @veekeys - That is definitely an option.  I was basing my implementation on what was done in the original Azure repo.  We need to make a decision on which way to proceed as soon as possible. We keyboard and touch processing for free with the current implementation. If we go to totally js solution does that need to be implemented in both backstop and iTwinUI-react?

src/slider/slider.scss
bsteinbk: I only chose it because I heard inline-flex was more efficient to render than flex.

src/slider/slider.scss
bsteinbk: You are correct, once I changed to use position absolute I should have changed this.

backstop/tests/slider.html
FlyersPh9: If the left & right labels do not have a shared fixed width, the slider will become off center depending on how long the labels are.
Screen Shot 2021-07-02 at 4 09 21 PM
width-issue

mayank99:
> If we go to totally js solution does that need to be implemented in both backstop and iTwinUI-react?

I would say no. The primary purpose of backstop html is visual tests. JS should be minimal, for adding main interactivity. Things like keyboard support, proper focus management, etc are not worth adding here.

src/slider/slider.scss
veekeys: hmm... Should we think about having a unique class instead? We already faced issues with the repeatance of the classes..

src/slider/slider.scss
```diff
@@ -0,0 +1,222 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+$thumb-height: $iui-m;  // 16px
+$track-height: $iui-xs;  // 4px
+$tick-height: ($iui-s + $iui-xxs);
```
veekeys, line 8: This is 10px... Any reason? It is just not really up to standards.. Could we change it to 8 or 12..?

src/slider/slider.scss
```diff
@@ -0,0 +1,222 @@
...
+        bottom: 0;
+        top: 12px;
+        pointer-events: none;
+        @include themed {
+          background-color: t(iui-color-foreground-primary);
+        }
+      }
+    }
+
+    .iui-slider-thumb.iui-left:focus,
```
veekeys, line 128: setting track-height on border-radius looks really weird...

src/slider/slider.scss
```diff
@@ -0,0 +1,222 @@
...
+        border-radius: $track-height;
+        left: 0;
+        top: 12px;
+        z-index: 1;
+        @include themed {
+          background-color: t(iui-color-foreground-primary);
+        }
+      }
+    }
+  }
```
veekeys, line 106: translate value is hard to understand..

src/slider/slider.scss
bsteinbk: It needs to be greater than 8, since thumb has radius of 8 which hides the tick.   12 is OK, just bigger than it needs to be.  Shouldn't this be a decision for Jon?

FlyersPh9: I changed `$tick-height` to use `$iui-baseline` = 11px.  Baseline is intended for spacing on the Y axis, so I think it makes sense here.

src/slider/slider.scss
bsteinbk: I removed border-radius it was left over when I was testing a rounded track.

src/slider/slider.scss
bsteinbk: Not sure what this means?

bsteinbk:
@FlyersPh9 - based on your changes, the disabled state only works if we wrap the whole slider inside a label element, is that by design?

src/slider/slider.scss
bsteinbk: I no longer use display flex. Not sure what is "hacky" please explain.

src/slider/slider.scss
bsteinbk: What class name would you like? I could change it to something like "iui-slider-min" and "iui-slider-max" if you prefer.

src/slider/slider.scss
veekeys: oh...comment was put on the wierd place.. now its on 67 line:
image
we can replace the top with the top value, but need to check what to do with horizontal placemenet and how we would implement that in react (if needed to change?)

src/slider/slider.scss
veekeys: Flex is still being used on `iui-slider-component-container`, which is the parent of `iui-slider-container`.
So `iui-slider-component-container` has `iui-left` and `iui-right` in children too. So all I am saying that practically `iui-slider-container` cannot take 100% width of the parent, cause parent has also 2 more elements which it needs to fit. Just now it works because of flex, so.. Better here would be to just apply `flex-grow: 1` on `iui-slider-container` so it has same outcome.

src/slider/slider.scss
veekeys: I think just flex would suit more here, so we ensure, that slider takes 100% of the parent width unless specified differently..
With inline-flex we would need to always add 100% width so the slider would work.
With `inline-flex`:
image
With `flex`:
image

Have in mind, I was using here slider WITHOUT input container wrapper.

backstop/tests/slider.html
veekeys: well, not sure in our examples its critical.. When removing this rule, we only get a bit longer slider itself (labels are still aligned).
image
Compared with fixed width:
image

veekeys: Cause if we expect users to have it also aligned the way we have it, we will need to add it in react somehow (if not provided from css by default...).

src/slider/slider.scss
veekeys: Yeah, I think that would suit better and reduce the specificity too.

src/slider/slider.scss
bsteinbk: I have removed the y adjustment and added a 'top' setting. The X adjustment is still necessary.

src/slider/slider.scss
veekeys: I renamed

src/slider/slider.scss
veekeys: Switched to flex and removed occurances of width: 100% in html

src/slider/slider.scss
veekeys: Changed

FlyersPh9:
> @FlyersPh9 - based on your changes, the disabled state only works if we wrap the whole slider inside a label element, is that by design?

It should be working now, I just tested it.  There is an `.iui-disabled` class applied to both `.iui-input-container` as well as `.iui-slider-component-container`.

src/slider/slider.scss
```diff
@@ -0,0 +1,204 @@
...
+
+  .iui-slider-thumb {
+    outline: none;
+    position: absolute;
+    pointer-events: all;
+    height: $thumb-height;
+    width: $thumb-height;
+    border-radius: 100%;
+    top: $track-height + 1; // position center of thumb in center of track
```
FlyersPh9, line 67: The regular slider's thumb is aligned correctly:
Screen Shot 2021-07-08 at 11 03 08 AM

But the ranges are off center, they are 2px further down than they should be.
Screen Shot 2021-07-08 at 11 05 26 AM

I believe the `top` is the culprit, but changing it to fix the ranges then breaks the sliders. Why is the range getting more height than the slider?

veekeys: fixed

src/slider/slider.scss
veekeys: Maybe I want, but we might need to remove tooltip handling from this css, cause in React it will work and be applied differently and these rules might mess it up..

src/slider/slider.scss
```diff
@@ -0,0 +1,177 @@
...
+  flex-grow: 1;
+  pointer-events: all;
+  cursor: pointer;
+
+  .iui-tooltip {
+    display: none;
+  }
+
+  .iui-slider-rail {
+    pointer-events: all;
```
veekeys, line 51: any specific reason for this one here?

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+    transform: translateX((-$thumb-height - 2) * 0.5); // Adjust the left (set in code) by radius of thumb so thumb center is at value position
+    z-index: 1;
+    cursor: grab;
+    @include themed {
+      background: t(iui-color-background-1);
+      border: 1px solid t(iui-color-background-5);
+    }
+
+    &:active {
+      cursor: grabbing;
```
mayank99, line 71: If we don't want to increase the thumb touch target, then lets remove this and add a class to change the  cursor for the whole container (not just thumb). After that, we can toggle that class based on JS drag events.

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+  user-select: none;
+
+  .iui-slider-tick {
+    position: relative;
+    display: flex;
+    pointer-events: none;
+    justify-content: center;
+    width: 1px;
+    height: $tick-height;
+    line-height: $tick-height * 4.25;
```
mayank99, line 104: is this what the value is supposed to be? 4.25 seems a bit random

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+    @include themed {
+      background: t(iui-color-background-1);
+      border: 1px solid t(iui-color-background-5);
+    }
+
+    &:active {
+      cursor: grabbing;
+    }
+
+    @include iui-focus();
```
mayank99, line 74: I don't think this is working. I'm able to make the focus ring show up when I click.

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  .iui-slider-thumb {
+    position: absolute;
+    pointer-events: all;
+    height: $thumb-height;
+    width: $thumb-height;
+    border-radius: 100%;
+    top: $track-height + 1; // position center of thumb in center of track
+    transform: translateX((-$thumb-height - 2) * 0.5); // Adjust the left (set in code) by radius of thumb so thumb center is at value position
+    z-index: 1;
```
mayank99, line 63: If we place the thumb at the end of the div (after the ticks), we could get rid of the z-index.

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+@mixin iui-slider-component-container {
+  display: flex;
+
+  .iui-slider-min,
+  .iui-slider-max {
+    @include iui-font-family;
+    user-select: all;
+    margin-top: $iui-xxs;
+
+    .iui-icon {
```
mayank99, line 19: Here we could use a longer class name to reduce nesting (and specificity).

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  &.iui-disabled {
+    @include iui-slider-disabled;
+  }
+}
+
+@mixin iui-slider-container {
+  position: relative;
+  flex-grow: 1;
+  pointer-events: all;
```
mayank99, line 42: What does this give us?

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+  .iui-slider-track {
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+    }
+  }
+
+  .iui-slider-min,
+  .iui-slider-max,
+  .iui-slider-tick {
+    user-select: none;
```
mayank99, line 141: `iui-slider-tick` already has user-select: none, so this is repeated, but maybe that's fine

src/slider/slider.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  .iui-slider-thumb {
+    cursor: not-allowed;
+    @include themed {
+      background-color: t(iui-color-background-4);
+      border-color: t(iui-color-background-4);
+    }
+
+    &:active {
+      cursor: not-allowed;
```
mayank99, line 128: If we remove the cursor above, we can remove it here as well

backstop/tests/slider.html
```diff
@@ -0,0 +1,590 @@
...
+        var min = parseInt(minAtt);
+        return min;
+      }
+
+      function getPercentFromThumb(thumb) {
+        var valueAtt = thumb.getAttribute("aria-valuenow");
+        var value = parseInt(valueAtt);
+        var min = getMinValueFromThumb(thumb);
+        var max = getMaxValueFromThumb(thumb);
+        var percent = ((value - min) / (max - min)) * 100;
```
mayank99, line 443: I think this might not be accurate enough for unbound sliders (like the volume one). When you slide the thumb, you see "jitters". Maybe this could be fixed through some animation though 🤔

See continuous sliders at mui for better behavior: https://material-ui.com/components/slider/#continuous-sliders

src/slider/slider.scss
mayank99: gif of issue: cursor

backstop/tests/slider.html
mayank99: gif of issue: 
jitter

bsteinbk: There is no jittering in my react implementation. If you want the native js based control to work smoother I will need to update the js. I thought the backstop was mostly to verify css not the js logic.

mayank99: Oh that's good then! I was assuming react would have a similar JS implementation.

src/slider/slider.scss
veekeys: In React it works well... lets keep it for now?

src/slider/slider.scss
veekeys: hmm..
Thats interesting.. I tried to have just focus and focus-visible rules like that, it still got the style on click:
image
Not sure, but might be some different case and browser decides that it needs to get the focus?
https://css-tricks.com/almanac/selectors/f/focus-visible/
There is a bunch of articles that sometimes it can be different... (inputs, buttons are the safe bet)

src/slider/slider.scss
veekeys: @bsteinbk you know this one?

bsteinbk: It was determined by trial and error.... it is defined to provides enough room for the ticks to be drawn above the text. Maybe it should be based on em and not px - a value of 3em also looks good.

src/slider/slider.scss
bsteinbk: I probably added it at some point to ensure the pointer was displayed. If it is no longer needed, it can be removed.

src/slider/slider.scss
bsteinbk: it allows slider to grow to fill available space between min/max labels

mayank99: ...pointer-events makes the slider grow?

src/slider/slider.scss
mayank99: Seems your react implementation requires the user to apply this class. https://github.com/iTwin/iTwinUI-react/pull/197/files#diff-73f4c3c7fc71dbb9d680fddeb2ede7bbe574448f82da3ba5e3e4f6cc197c91deR27

I would have expected that you clone it and apply it yourself, which is why I suggested the longer name.

An alternative is to target `svg` instead.

src/slider/slider.scss
mayank99: we plan to switch to rem units in the future, but we cant have a mix of them.

My concern here is that a line height of 46.75px could cause problems on different screen sizes.

src/slider/slider.scss
veekeys: removed

src/slider/slider.scss
veekeys: even a specific slider lib has same issue:
https://slider-react-component.vercel.app/demo/handle

src/slider/slider.scss
veekeys: Lets leave it for now..

src/slider/slider.scss
veekeys: Changed to 44px...

src/slider/slider.scss
bsteinbk: @mayank99 - Jon added the iui-icon rule and I added the class in the storybook example so icon was properly styled. I am unsure what you mean by cloning or targeting svg.

mayank99: cloning:
```tsx
React.cloneElement(startIcon, { className: cx('iui-icon', startIcon.props.className) }
```

targeting svg:
```scss
svg {
  @include iui-icons-default();
  ...
}
```

src/slider/slider.scss
mayank99: but we can do better 😕

veekeys: well..depends what we want to do firstly.. still have a grab cursot? then we for sure will need to increate the toucharea.. which will still not save the bug with the touch of the track on certain intervals

src/slider/slider.scss
bsteinbk: removed

backstop/tests/slider.html
FlyersPh9: It is pretty minor and only becomes more obvious the wider the range is (1-1,000,000 is obvious where 100,000-1,000,000 is not).  We can choose to skip this for now.

mayank99 (approved):

PR #177, opened 2021-07-06
User icon with count inside has different visual styling

FlyersPh9:
The user icon within a list that shows the count has the new `.iui-user-icon-count` class applied which gives slightly different visual styling.  Adding this in response to [this comment in iTwinUI-React](https://github.com/iTwin/iTwinUI-react/pull/186#discussion_r662961797).

Also, we will be updating the documentation to match what is said [here](https://github.com/iTwin/iTwinUI-react/pull/186#issuecomment-874949869).

Screen Shot 2021-07-06 at 3 49 40 PM
Screen Shot 2021-07-06 at 3 49 55 PM


mayank99 (approved):
I am ok with this, but I think we might need more changes for https://github.com/iTwin/iTwinUI-react/pull/186

Specifically:
- class for non-stacked list
- inline-flex to avoid taking full width
- allow last icon to be wrapped (maybe)

But those can come in a separate PR.

PR #179, opened 2021-07-07
Add font family fallback options

FlyersPh9:
- Added system font fallback options for `$iui-sans`.
- Added system font fallback options for `$iui-monospace` and changed the order slightly.
- Codeblock now uses `$iui-monospace` correctly. 

Created in response to [this PR comment in iTwinUI-React](https://github.com/iTwin/iTwinUI-react/issues/182#issuecomment-875758633). Used [this blog post](https://qwtel.com/posts/software/the-monospaced-system-ui-css-font-stack/), [this blog post](https://infinnie.github.io/blog/2017/systemui.html), & [this blog post](https://markdotto.com/2018/02/07/github-system-fonts/#the-stack) to come up with the correct stack.  We removed `system-ui` so that tests would pass, plan to re-add it back when we bundle Open Sans.

### DIFF
--- a/src/style/typography.scss
+++ b/src/style/typography.scss
@@ -5,8 +5,8 @@
 
 /// Constants ------------------------------------------------------------------
 
-$iui-sans: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
-$iui-monospace: Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+$iui-sans: 'Open Sans', BlinkMacSystemFont, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', sans-serif;
+$iui-monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
 
 $iui-font-family: $iui-sans;
 
--- a/src/style/typography.scss
+++ b/src/style/typography.scss
@@ -5,7 +5,7 @@
 
 /// Constants ------------------------------------------------------------------
 
-$iui-sans: 'Open Sans', -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Helvetica Neue', system-ui, Arial, sans-serif;
+$iui-sans: 'Open Sans', -apple-system, BlinkMacSystemFont, 'Segoe UI', system-ui, 'Roboto', 'Helvetica Neue', Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


src/style/typography.scss
```diff
@@ -5,8 +5,8 @@
 
 /// Constants ------------------------------------------------------------------
 
-$iui-sans: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
-$iui-monospace: Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+$iui-sans: 'Open Sans', BlinkMacSystemFont, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', sans-serif;
+$iui-monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
```
mayank99, line 9: - [SFMono](https://github.com/supercomputra/SF-Mono-Font) says "Please note that it is forbidden to use this typeface for your commercial project."
- Menlo seems to have been phased out completely. Monaco is even older.
And I don't think any of the above 3 fonts are even allowed to be used in non-Apple contexts. Likewise, Consolas is Windows-specific, though it has fewer restrictions.

Shouldn't we use a better _default_ font that looks the same on all operating systems?

src/style/typography.scss
```diff
@@ -5,8 +5,8 @@
 
 /// Constants ------------------------------------------------------------------
 
-$iui-sans: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
-$iui-monospace: Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+$iui-sans: 'Open Sans', BlinkMacSystemFont, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', sans-serif;
```
mayank99, line 8: Are `'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'` really necessary?

FlyersPh9: I'll remove them.  CSS-Tricks recommended the stack, I mostly just copy / pasted.  I will say, I have not tested emojis in our pages at all.

src/style/typography.scss
FlyersPh9: I don't understand how [Putra Z.](https://github.com/supercomputra) owns SFMono because [San Francisco Mono is an Apple built font](https://developer.apple.com/fonts/). 🤔

For the monospaced font stack, maybe the first font should be a web font similar to Open Sans so it will be consistent across all OSes.  I like [Inconsolata](https://fonts.google.com/specimen/Inconsolata?category=Monospace).

mayank99 (approved):

PR #188, opened 2021-07-14
Fix slider touch-action

mayank99:
In response to https://github.com/iTwin/iTwinUI-react/pull/197#discussion_r669107981


FlyersPh9 (approved):

src/slider/slider.scss
```diff
@@ -40,6 +40,7 @@ $tick-height: $iui-baseline; // 11px
   position: relative;
   flex-grow: 1;
   cursor: pointer;
+  touch-action: pan-y;
```
veekeys, line 43: Thats interesting.. I thought y is vertical, while the slider is horizontal....

mayank99: pan-y means the browser will preserve the vertical panning behavior (and disable default handling of pan-x, pinch-zoom, and other touch actions). This allows us to use the slider without the browser touch actions getting in the way.

I could have also gone with `touch-action: none` and it would work but that would be more restrictive.

veekeys: Well, maybe I am not that good reader of the docs..
image

mayank99: Maybe the docs are not clear, but `pan-y` means "browser should handle vertical panning, user should handle rest".

PR #192, opened 2021-07-15
Add iPhone notch support

FlyersPh9:
Before:
Screen Shot 2021-07-15 at 12 41 37 PM

After:
*The expander arrow appears off center, that has been fixed since I took these screenshots.*
Screen Shot 2021-07-15 at 12 39 50 PM
Screen Shot 2021-07-15 at 12 39 20 PM


src/header/header.scss
```diff
@@ -12,6 +12,9 @@
   width: 100%;
   height: $iui-baseline * 5;
   user-select: none;
+  box-sizing: border-box;
```
FlyersPh9, line 15: This was necessary, otherwise the `width: 100%` + padding made the header display at something like 110% screen width.  This also made the header lose a little bit in height causing the header image tests to change.

[Sanitize.scss](https://github.com/csstools/sanitize.css#box-sizing-defaults-to-border-box) suggests applying this on everything.  Maybe worth adding to all classes with `.iui-`?

mayank99: Applying it on everything will likely cause hundreds of images to change. Maybe you can combine it with the rem change, and convert this PR to draft until then.

FlyersPh9: Looks to be 98 tests failing if we set `[class*='iui-'] { box-sizing: border-box; }`.

veekeys: I think we talked that in most of the cases we should have border-box.. that way, we can ensure out sizes are actually the ones we set in rules...
But its a huge change, again...

mayank99 (approved):
Looks good, with two conditions:

1. Need to not forget to remove the `+ 2px` height when switching everything to border-box.
2. Need to remember to remove the `unquote` hack for `min`/`max` when switching from node-sass to dart-sass. (Hack was found here: https://github.com/sass/sass/issues/2378#issuecomment-367490840)

bentleyvk (approved):
I hope it works 😃 Need to ask Bentley for an iPhone to double check.

PR #195, opened 2021-07-16
Remove svg selector from tabs

mayank99:
After https://github.com/iTwin/iTwinUI-react/pull/181, there is no need for the `svg` selector, as we are forcing users to use our component (which applies the class automatically).


veekeys (approved):
Just I guess it is important to deliver this in react with Tabs changes  (do not publish new css without new Tab component).

FlyersPh9 (approved):

PR #204, opened 2021-07-22
Remove margin from typography classes

FlyersPh9:
- Removed margin from typography classes.
- Added `.iui-text-spacing` modifier to re-apply the margin as needed.
- `.iui-text-muted` is now using the `$iui-text-color-muted` variable.
- Non-ideal-state requires `@include iui-text-block-spacing;` to avoid test images from changing.


mayank99 (approved):
Looks good!

Will later need to decide the approach in React.

src/text/mixins.scss
```diff
@@ -8,6 +8,8 @@
 }
```
bentleyvk, line 9: There also is `caption`. Are we using it anywhere?

src/text/classes.scss
```diff
@@ -1,35 +1,40 @@
...
-  @include iui-text-title;
+.iui-text-muted {
+  @include iui-text-muted;
 }
 
 .iui-skeleton {
   @include iui-text-skeleton;
 }
+
+.iui-text-spacing {
```
bentleyvk, line 38: I really like this cleanup! But I am afraid that we will break UIs for our users because when I used typography components, I haven't added any additional margin as they had it. Also, we now are adding spacing class to typography examples, so is it worth having spacing separate?

mayank99: So here's what I was thinking:
1. `.iui-text-spacing` would be applied to our current Typography components in React.
2. A new component called `Text` would not have spacing and support any kind of styling on any kind of element (e.g. h1 styling on h2).

src/text/mixins.scss
FlyersPh9: It appears that caption & small have the same properties.  I'm referencing the `iui-text(small)` mixin within `classes.scss` but no mention of `iui-text(caption)`.

saaaaaally (approved):

src/text/mixins.scss
bentleyvk: So we need to remove it, no?

bentleyvk (approved):

PR #209, opened 2021-07-27
Clean up breadcrumb-item styles

mayank99:
- Made anchor links equal the same height as buttons.
- Removed need for explicitly specifying `iui-anchor` class. All `a` elements are now styled automatically.
- Applied margin on breadcrumb item instead of the element inside.
- Removed align-items: center as it was doing nothing.
- Made widths of all kinds of items consistent.


src/breadcrumbs/breadcrumbs.scss
```diff
@@ -20,28 +20,27 @@
...
   }
 
-  > .iui-anchor,
-  > .iui-button > .iui-label {
+  > *,
+  .iui-label {
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
-    max-width: $iui-3xl * 2;
```
bentleyvk, line 44: Why removed?

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -20,28 +20,27 @@
...
+  height: $iui-component-height;
+  max-width: $iui-3xl * 2;
+  margin: 0 $iui-sm;
 
-  > .iui-anchor,
-  > span,
-  .iui-ellipsis {
-    margin: 0 $iui-sm;
+  > .iui-button {
+    margin: 0 (-$iui-s);
```
bentleyvk, line 29: Why negative margin?

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -20,28 +20,27 @@
...
+  &:last-child {
+    &,
+    > .iui-button {
+      margin-right: 0;
+    }
   }
 
-  > .iui-anchor,
-  > .iui-button > .iui-label {
+  > *,
```
bentleyvk, line 39: I don't like targeting everything. I would say to target `a` or bring back `iui-anchor`.

src/breadcrumbs/breadcrumbs.scss
mayank99: I placed it at the root of breadcrumb-item

src/breadcrumbs/breadcrumbs.scss
mayank99: Button has its own padding, so I'm counteracting the margin that I set above in the breadcrumb-item

src/breadcrumbs/breadcrumbs.scss
mayank99: This change is to allow targeting `a`, `span` and anything else like a custom `div`.

It's only first level `*`.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -20,28 +20,27 @@
...
+  > .iui-button {
+    margin: 0 (-$iui-s);
   }
 
-  > .iui-button {
-    margin: 0 $iui-xs;
+  &:last-child {
+    &,
+    > .iui-button {
+      margin-right: 0;
```
veekeys, line 35: Does it make any difference?

mayank99: It increases the width, but looks same visually. I could remove

veekeys: I dont see the width change with or without it

mayank99: well yeah, i said "looks same visually" 😁

I've removed and pushed. See the change in test images

veekeys (approved):

PR #210, opened 2021-07-28
Add split header button

gretanausedaite:
Adding split menu button to header
Slim:
image
image
image

Default:
image
image

Dark mode:
image


src/header/header.scss
```diff
@@ -266,3 +271,70 @@
     }
   }
 }
+
+.iui-split-menu-header-button {
```
mayank99, line 275: Why do we need a completely separate new class for this? I thought we were making all the header buttons split.

mayank99 (commented):
I see some problems with specificity but we should first decide on this whole approach.

src/header/header.scss
gretanausedaite: I don't think it's a good idea to make ALL header buttons split. 
I can remove separate class and move functionality the existing one.

veekeys: I think we should support both.. Both are valid.
@FlyersPh9

FlyersPh9: I think most workflows will use the split menu button.  I can't think of when they'd want the single button, but maybe we support it just in case?

mayank99: In either case, there can be a good bit of code reuse (by making this class a _modifier_ on the normal header button). I think will help fix the specificity problems too.

backstop/tests/header.html
```diff
@@ -752,6 +752,362 @@ <h2>Slim</h2>
...
+                <span class="iui-label">
+                  <div>Project Alpha</div>
+                  <div class="iui-description">0x0123456789</div>
+                </span>
+                <!-- Use caret-down-small.svg from @itwin/itwinui-icons package. -->
+
+                <span
+                  id='test-button-7'
+                  class="iui-button iui-borderless iui-header-button iui-large iui-split-menu"
+                >
```
FlyersPh9, line 804: Can you add `tabindex="0"` to the split menu buttons so that we can test for `:focus`?
Screen Shot 2021-07-28 at 5 16 23 PM

src/header/header.scss
```diff
@@ -266,3 +271,70 @@
...
+    &.iui-active {
+      svg {
+        fill: var(--iui-icons-color-primary);
+      }
+    }
+
+    svg {
+      border-radius: 50%;
+      padding: $iui-xs;
+    }
```
FlyersPh9, line 313: Why does the circular shape get cutoff at the top & bottom within the slim header?
Screen Shot 2021-07-28 at 5 19 01 PM

mayank99: It needs a different radius for slim size

src/header/header.scss
gretanausedaite: Moved css class.

backstop/tests/header.html
gretanausedaite: Added tabIndex

src/header/header.scss
gretanausedaite: I think this fixed itself when css class was moved.

backstop/tests/header.html
```diff
@@ -752,6 +752,357 @@ <h2>Slim</h2>
...
+              aria-expanded="false"
+              id="test-button-5"
+            >
+              <div class="iui-icon iui-header-button-icon demo-picture"></div>
+              <span class="iui-label">
+                <div>Project Alpha</div>
+                <div class="iui-description">0x0123456789</div>
+              </span>
+              <!-- Use caret-down-small.svg from @itwin/itwinui-icons package. -->
+              <span
```
mayank99, line 799: Why use span here?

These are meant to be two separate buttons grouped together. Check our regular split button for example: https://github.com/iTwin/iTwinUI/blob/main/backstop/tests/button.html#L1726

backstop/tests/header.html
mayank99: tabindex wont be necessary if we use button instead of span

src/header/header.scss
```diff
@@ -206,6 +207,29 @@
     max-width: 25vw;
     overflow: hidden;
 
+    // Sizes
```
mayank99, line 210: I don't think sizes are used anywhere. We default to large size, and automatically reduce height for slim mode

src/header/header.scss
```diff
@@ -264,5 +292,54 @@
...
+        }
+      }
+
+      svg {
+        border-radius: 50%;
+        padding: $iui-xxs;
+      }
+
+      &:hover,
+      &:focus-visible {
```
mayank99, line 320: Three problems here:
1. This selector will be invalid in IE/Safari and break hover styling too (see [one-invalid-pseudo-selector-equals-an-entire-ignored-selector](https://css-tricks.com/one-invalid-pseudo-selector-equals-an-entire-ignored-selector/))
2. Declaring focus-visible styles is the wrong approach. The "proper" way to do it is to just add normal focus styles, and negate non-visible styles like this:
```scss
:focus {
  ...
}
:focus:not(:focus-visible) {
  // undo styles from above
}
```
3. Using background is probably not the best focus style. Outline/border/box-shadow would make more sense.

src/header/header.scss
```diff
@@ -264,5 +292,54 @@
...
+        svg {
+          @include themed {
+            background-color: rgba(
+              var(--iui-color-foreground-body-rgb),
+              var(--iui-opacity-5)
+            );
+          }
```
mayank99, line 327: We don't usually use css variables (i.e. `var` in our components) because they are not supported in IE. You are already inside the themed mixin, so just use the `t` syntax and remove `--`. See example:

https://github.com/iTwin/iTwinUI/blob/4b69a213920d4af9f1acee63d2f3f79840c06e68/src/style/theme.scss#L425-L429

src/header/header.scss
```diff
@@ -264,5 +292,54 @@
...
+    .iui-split-menu {
+      display: inline-flex;
+      justify-content: center;
+      @include iui-focus;
+
+      svg {
+        width: 16px;
+        @include themed {
+          fill: var(--iui-icons-color-actionable);
+        }
```
mayank99, line 305: You can include icons mixin instead.
```suggestion
        @include iui-icons-default;
```

mayank99 (commented):
I think there is still a problem with the approach.

I would have expected:
```html
<span>
  <button> ...
  <button-with-dropdown-icon> ...
</span>
```

Instead we have:
```html
<button>
 <regular-span>
 <span-with-icon>
</button>
```

backstop/tests/header.html
FlyersPh9: [Nesting a button within a button doesn't seem valid.](https://stackoverflow.com/a/39386695)

FlyersPh9: What if we do a button within an anchor link?  Clicking on the anchor link takes you to a page about the project / model, clicking on the nested button opens the dropdown menu.

src/header/header.scss
FlyersPh9: Can confirm, I'm not seeing the issue anymore.

src/header/header.scss
FlyersPh9: `@include iui-icons-default;` uses `@include iui-icons-color();` which is equal to `$iui-icons-color`, not `$iui-icons-color-actionable`.

mayank99: Ah, they have different opacities 😕 

Didn't we remove actionable icons?

backstop/scenarios/header.js
```diff
@@ -21,4 +21,14 @@ module.exports = [
     actions: [hover('#test-button-3'), focus('#test-button-4')],
     selectors: ['#demo-slim'],
   }),
+
+  //// Split
+  scenario('Type Default State hover & focus', {
```
veekeys, line 26: Wait, why our different scenario names are same???

backstop/tests/header.html
```diff
@@ -752,6 +752,365 @@ <h2>Slim</h2>
...
+            <span class="iui-label">App name</span>
+          </div>
+
+          <div class="iui-divider"></div>
+
+          <!-- Project Button -->
+          <nav aria-label="breadcrumbs">
+            <span class="iui-button-split-menu">
+              <button
+                class="iui-button  iui-borderless iui-header-button iui-large"
```
veekeys, line 787: extra space in class not needed

backstop/tests/header.html
```diff
@@ -752,6 +752,365 @@ <h2>Slim</h2>
...
+              >
+                <div class="iui-icon iui-header-button-icon demo-picture"></div>
+                <span class="iui-label">
+                  <div>Project Alpha</div>
+                  <div class="iui-description">0x0123456789</div>
+                </span>
+              </button>
+              <!-- Use caret-down-small.svg from @itwin/itwinui-icons package. -->
+              <button
+                class="iui-button iui-borderless iui-header-button iui-split-menu iui-large"
```
veekeys, line 801: `iui-split-menu` class needed here?

src/header/header.scss
```diff
@@ -245,15 +268,104 @@
...
+      &:focus:not(:focus-visible) {
+        @include themed {
+          box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+        }
+      }
+    }
+  }
+
+  .iui-button-split-menu {
+    height: 100%;
```
veekeys, line 297: image
found this generated...not sure where duplication comes from..but it is in our split button. Will need to investigate

src/header/header.scss
```diff
@@ -245,15 +268,104 @@
...
+          @include themed {
+            background-color: rgba(
+              t(iui-color-foreground-primary-rgb),
+              t(iui-opacity-6)
+            );
+          }
+        }
+      }
+
       &:focus,
```
veekeys, line 364: Focus looks bad..
image
It should take same width as on hover (or active)

src/header/header.scss
```diff
@@ -206,6 +207,28 @@
     max-width: 25vw;
     overflow: hidden;
 
+    &.iui-button {
+      .iui-split-menu {
+        svg {
+          margin-left: $iui-xs;
```
veekeys, line 213: This is really needed?

backstop/tests/header.html
veekeys: I see there are some specific rules..
Maybe some more descriptive class or just last-child could be ok for overrides...

src/header/header.scss
```diff
@@ -245,15 +268,104 @@
...
+      .iui-icon {
+        color: t(iui-color-foreground-primary);
+        fill: t(iui-icons-color-primary);
+      }
+
       
+      
+      &,
+      &:hover {
+        background-color: rgba(
```
veekeys, line 349: whenever you use color, you need to use @themed mixin...

src/header/header.scss
```diff
@@ -245,15 +268,104 @@
...
+        @include themed {
+          box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+        }
+      }
+    }
+  }
+
+  .iui-button-split-menu {
+    height: 100%;
+    padding: 0 4px 0 12px;
```
veekeys, line 298: constants please

src/header/header.scss
```diff
@@ -263,6 +375,20 @@
           box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
         }
       }
+
+      .iui-split-menu:hover {
```
veekeys, line 379: You have above also rule for hover and svg... Not sure which should be when now, but seems like repeating...
image

src/header/header.scss
```diff
@@ -245,15 +268,104 @@
...
+      &,
+      &:focus:not(:focus-visible) {
+        @include themed {
+          box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+        }
+      }
+    }
+  }
+
+  .iui-button-split-menu {
```
bentleyvk, line 296: It is not handling overflow like basic header button:
image

src/header/header.scss
```diff
@@ -206,6 +207,28 @@
...
 
+    &.iui-button {
+      .iui-split-menu {
+        svg {
+          margin-left: $iui-xs;
+        }
+      }
+    }
+
+    &.iui-small {
```
bentleyvk, line 218: Are all of these stylings used? Because I see we only use `iui-large`.

src/header/header.scss
```diff
@@ -292,54 +291,104 @@
...
+            0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+        }
+      }
+
+      &,
+      &:focus:not(:focus-visible) {
+        @include themed {
+          box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+        }
+      }
```
mayank99: Did you mean to do this?

```suggestion
      &:focus {
        @include themed {
          box-shadow: t(iui-focus-box-shadow),
            0 $iui-xxs 0 0 t(iui-color-foreground-primary);
        }
      }

      &:focus:not(:focus-visible) {
        @include themed {
          box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
        }
      }
```

Because that first selector (`&:focus, &:focus-visible`) will break all focus styling in Safari/IE.

And the second selector (`&`) also applies those styles when focus is not applied, which is wrong. If unsure which selectors to use, you can look at the compiled css to see what the `&` is doing.

src/header/header.scss
```diff
@@ -292,54 +291,104 @@
...
+              t(iui-color-foreground-primary-rgb),
+              t(iui-opacity-5)
+            );
           }
         }
       }
     }
+
+    
+
```
mayank99, line 392: Extra lines?

backstop/tests/header.html
```diff
@@ -966,17 +967,24 @@ <h2>Split Menu Buttons</h2>
...
-              type="button"
-              aria-label="Project picker"
-              aria-expanded="false"
-            >
-              <span class="iui-label">
-                <div class="iui-name">Project Alpha</div>
-              </span>
+            <span class="iui-button-split-menu">
+              <button
+                class="iui-button  iui-borderless iui-header-button iui-large"
```
mayank99, line 972: Extra space between iui-button and iui-borderless.

Weird it didn't auto format

backstop/tests/header.html
```diff
@@ -966,17 +967,24 @@ <h2>Split Menu Buttons</h2>
...
+                <span class="iui-label">
+                  <div>Project Alpha</div>
+                </span>
+              </button>
               <!-- Use caret-down-small.svg from @itwin/itwinui-icons package. -->
-              <span class="iui-split-menu">
+              <button
+                class="iui-button iui-borderless iui-header-button iui-split-menu iui-large"
+                id='test-button-6'
+                tabindex="0"
```
mayank99, line 986: Buttons should not need tabindex anymore

backstop/tests/header.html
```diff
@@ -966,17 +967,24 @@ <h2>Split Menu Buttons</h2>
...
+                type="button"
+                aria-label="Project picker"
+                aria-expanded="false"
+                id="test-button-5"
+              >
+                <span class="iui-label">
+                  <div>Project Alpha</div>
+                </span>
+              </button>
               <!-- Use caret-down-small.svg from @itwin/itwinui-icons package. -->
```
mayank99, line 982: Can you move these comments down right where the svg is actually used? Saw it in a couple more places (like line 799).

Like this:

```html
<!-- Use caret-down-small.svg ...>
<svg ...>
```

src/header/header.scss
mayank99: It might be hard to make focus take full width unless we absolutely position the dropdown icon.

Maybe it should only extend to the left edge, not right

src/header/header.scss
mayank99: I'm talking about the actual size rules below, but you have just removed the comment 😆

src/header/header.scss
```diff
@@ -265,4 +292,124 @@
       }
     }
   }
+
+  .iui-split-header-button {
```
mayank99, line 296: Why is this class needed? It's not even a modifier on base split-button so you have to reinvent the wheel

mayank99 (commented):
I think the approach is still wrong and there is a lot of duplicated or dead code.

I would have expected that you just add borderless hover states to base split button css first (it's only 2 lines of css), and then check what's required extra for header.

src/header/header.scss
bentleyvk: It is not really worth reusing split button because you need to override a lot of stuff. Previously I thought the same but then I tried to do that myself 😃

mayank99: The main reason it became complicated is that we wanted the split button to not act split, but more like a secondary button between a primary button.

It is much much simpler if we just change the intended hover effect to this:
| hover left | hover right |
| --- | --- |
| image | image |

@FlyersPh9 Is there anything wrong with this hover effect? I think it looks better and is significantly easier to implement (2 lines of css vs 100+ lines).

src/header/header.scss
```diff
@@ -265,4 +292,13 @@
...
   }
+
+  .iui-button-split-menu {
+    height: 100%;
+    padding: 0 0 0 $iui-xs;
+
+    .iui-button {
+      padding: 0 $iui-xs;
+    }
+  }
```
bentleyvk, line 303: Much cleaner code right now 👍 Need to decide about active state button.

src/header/header.scss
```diff
@@ -206,6 +207,28 @@
...
+      }
+    }
+
+    &.iui-large {
+      .iui-split-menu {
+        svg {
+          margin-left: $iui-s;
+        }
+      }
+    }
```
mayank99, line 230: Can you remove all this dead code?

src/header/header.scss
```diff
@@ -265,4 +292,51 @@
...
+
+    &.iui-active {
+      // Basic active styling - blue background and bottom border
+      @include themed {
+        box-shadow: 0 2px 0 0 t(iui-color-foreground-primary);
+      }
+
+      &,
+      &:hover {
```
mayank99, line 314: So active hover state does not change background?

src/header/header.scss
```diff
@@ -265,4 +292,51 @@
...
+    max-width: 25vw;
+
+    .iui-button {
+      padding: 0 $iui-xs;
+    }
+
+    &.iui-active {
+      // Basic active styling - blue background and bottom border
+      @include themed {
+        box-shadow: 0 2px 0 0 t(iui-color-foreground-primary);
```
mayank99, line 310: Can you use variable for 2px? `$iui-xxs`

src/header/header.scss
```diff
@@ -265,4 +292,51 @@
...
+          background-color: transparent;
+        }
+      }
+
+      .iui-button:last-child:hover {
+        @include themed {
+          background-color: rgba(
+            t(iui-color-foreground-primary-rgb),
+            t(iui-opacity-6)
+          ) !important;
```
mayank99, line 337: Lets try to remove these `!important`s.

What rule do you need to override and what is the specificity of it? We should be able to boost the specificity in header (at least temporarily).

mayank99 (commented):
Looking much better! Did anything change visually? Updated screenshots would be nice.

src/header/header.scss
gretanausedaite: With current implementation, none of active header buttons change on hover
hoveraction

mayank99: Makes sense to preserve existing behavior I suppose.

I think when we decide what to do with active split button, we should also decide if this hover state is what we want.
@FlyersPh9

src/header/header.scss
gretanausedaite: iui-active has at least 4 different background-color rules 
image

!important overrides all of them. This should be fixed when reduce specificity task is completed.

mayank99: looks like a specificity of 0-6-0 for active state, and 0-7-0 for active hover state.

So we need at least 0-6-1 and 0-7-1 respectively. Right now those two rules have 0-4-0 and 0-6-0.

One easy thing you can do is add a modifier on `.iui-button-split-menu`.
Another thing you can do is add `.iui-borderless` to the `.iui-button` selectors.
And then, you can also add `button` to boost the last part.

gretanausedaite: Removed !important rules by adding borderless selector.

mayank99: That's great!

Is `::after` `one really needed? If yes, you could combine with the other selectors above.

Also I still think we should have a modifier class on `.iui-button-split-menu` (call it `.iui-header-split-menu` or something). Right now our selector is tackling all split buttons at all levels inside the header, and it could cause unnecessary problems.

src/header/header.scss
```diff
@@ -265,4 +270,56 @@
       }
     }
   }
+
+  .iui-button-split-menu {
+    height: 100%;
+    padding: 0 0 0 $iui-xs;
```
bentleyvk, line 276: Why not just `padding-left`?

src/header/header.scss
```diff
@@ -265,4 +270,56 @@
...
+      // Removing button hovers and backgrounds and adding last child hover
+      .iui-button {
+        &.iui-borderless {
+          &,
+          &:hover {
+            background-color: transparent;
+          }
+
+          &::after {
+            background-color: transparent;
```
bentleyvk, line 310: You could set `display: none` instead, then element won't be rendered at all.

bentleyvk (approved):

backstop/tests/header.html
gretanausedaite: I think this comment can be resolved as we are using button

src/header/header.scss
gretanausedaite: Styling is changed. Not sure if still relevant

backstop/scenarios/header.js
gretanausedaite: Changed scenario names

src/header/header.scss
gretanausedaite: Focus styling was changed.

src/header/header.scss
gretanausedaite: Styling was changed.

backstop/tests/header.html
gretanausedaite: Removed extra space

backstop/tests/header.html
gretanausedaite: Removed tabindex

src/header/header.scss
gretanausedaite: I added active split button

src/header/header.scss
gretanausedaite: ::after is needed to remove top border. Changed styling to display: none
Added iui-header-split-menu modifier class

src/header/header.scss
mayank99: This comes from `buttons/classes.scss`.
image

Need to remove line 35 because we already include button-group mixin in split-menu mixin.

Can tackle in #225

backstop/scenarios/header.js
mayank99: Looks like the images changed file names because of that. Can you delete old images?

One way to do that is to run `yarn test:ci` (without any filter) and then `yarn clean:images`.

src/header/header.scss
FlyersPh9: How you have it working in the animation up above is good & what I was expecting. 😊

mayank99: So it's valid to have an active split button in the first place?

FlyersPh9 (approved):

FlyersPh9:
@gretanausedaite 1 last change before we merge this into main.  Currently the 2 different type of buttons have slightly different spacing:
Screen Shot 2021-08-06 at 10 57 25 AM

If you apply these overrides, everything should line up perfectly:
```
.iui-button-split-menu { padding-left: 8px; }
.iui-header-button { padding-right: 8px; }
```
MicrosoftTeams-image (5)
MicrosoftTeams-image (4)

mayank99 (approved):
Good job!

PR #224, opened 2021-08-04
Rework badge colors to use softer backgrounds and statuses

### LINKED ISSUE #216: Badge color system could use a rework
Documenting 3 issues with current badges:

1. There is no default color applied when doing this:
```html
<span class="iui-badge">Label</span>
```

2. There is no proper styling for status colors (e.g. success, failure, etc)

3. Most of the current data-viz colors (recommended for badge) are inaccessible due to low contrast.

mayank99:
Closes #216.

#### Changes

After a lot of playing around, I decided to create new color variables that are 10-20% lighter than the data-viz colors (which I also converted from hex to hsl). After that, I just changed the text color to `#000` at 0.6 opacity. These two changes combined shot up the accessibility score to above 4.5 for each variant.

Also added status colors: default, positive, negative, warning, primary.

Also updated font weight to semibold (600) and increased inline padding from 4px to 6px. These changes improve the visuals considerably. Expand comparison details below.

<details>

| 4px padding, 400 font weight, 0.8 font opacity | 6px padding, 600 font weight, 0.6 font opacity |
| --- | --- |
|  | image |
|  | image |
</details>

#### Screenshots
image
image

#### Tests

Added missing test scenario and images for badge.

Didn't update badge colors in Tile, and temporarily forced styles to reset to those prior to this PR, because it was otherwise changing 50 images.


backstop/scenarios/badge.js
```diff
@@ -0,0 +1,5 @@
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
+  scenario('basic', { selectors: ['#demo'], hideSelectors: ['h2'] }),
```
FlyersPh9, line 4: Can we rename the scenario to "Type all" to better match the naming structure I'm trying to get passed in #155?

mayank99: Changed

FlyersPh9 (approved):

PR #229, opened 2021-08-06
Remove iui-font-family from all components

mayank99:
### Changes

1. Removed iui-font-family from all components. Had to add `font-family: inherit` to inputs because it doesn't automatically use fonts from body.

2. Added `.iui-no-webfonts` modifier on `.iui-body` to only load fallback fonts.

### Rationale

These two changes combined will give the user more control over when the Open Sans webfont is set and what to show while it's loading.

### Tests

Removed font-family from body in demo.css to be able to test without fonts. Working fine in my testing.

### DIFF
--- a/backstop/assets/demo.css
+++ b/backstop/assets/demo.css
@@ -1,4 +1,3 @@
-body,
 h1,
 h2,
 h3,
@@ -15,6 +14,7 @@ p {
 }
 
 body {
+  margin: 0;
   padding: 24px;
 }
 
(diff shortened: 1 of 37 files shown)


backstop/assets/demo.css
```diff
@@ -15,6 +14,7 @@ p {
 }
 
 body {
+  margin: 0;
```
bentleyvk, line 17: How this doesn't break test images?

bentleyvk (approved):
I really lack knowledge in fonts so I guess it is ok. We will need to update our wiki about loading the fonts.

backstop/assets/demo.css
mayank99: body was in the block above this, I just moved it down.

PR #233, opened 2021-08-09
feat(Tabs): Add vertical modifier to default and borderless tabs

### LINKED ISSUE #218: Add vertical tabs modifier
#### Feature
Add a modifier class to `.iui-tabs` enabling vertical tab layout.

#### Examples
Screen Shot 2021-08-03 at 12 19 46 PM
Screen Shot 2021-08-03 at 12 20 04 PM
*Example above is missing the blue stripe.*

FlyersPh9:
Added `.iui-vertical` modifier to the default and borderless tabs. Closes #218.

Screen Shot 2021-08-09 at 5 39 32 PM
Screen Shot 2021-08-18 at 11 05 45 AM


FlyersPh9:
I am unsure how to get the animated stripe to appear correctly on the borderless vertical tabs. Any suggestions?

mayank99:
> I am unsure how to get the animated stripe to appear correctly on the borderless vertical tabs. Any suggestions?

Is it even valid to have borderless/pill variants for vertical tabs?

If yes, what is the problem? animating? positioning? sizing?

src/tabs/borderless.scss
```diff
@@ -28,6 +28,30 @@
     }
   }
 
+  // #region Horizontal tabs
```
bentleyvk, line 31: Is it worth having such a small regions?

src/tabs/borderless.scss
```diff
@@ -28,6 +28,30 @@
...
+  &:not(.iui-vertical) {
```
bentleyvk, line 32: Maybe we need to have both and `iui-horizontal` and `iui-vertical` just to be more explicit.

src/tabs/tabs.scss
```diff
@@ -151,24 +151,57 @@
...
+    }
+
+    &.iui-large ~ .iui-tab-stripe {
+      top: $iui-component-height-large - $iui-xxs; // Also why not bottom 0?
+    }
   }
+  // #endregion
+
+  // #region Vertical tabs
+  &:not(.iui-vertical) {
```
bentleyvk, line 177: Huh? Both not vertical?

bentleyvk:
> > I am unsure how to get the animated stripe to appear correctly on the borderless vertical tabs. Any suggestions?
> 
> Is it even valid to have borderless/pill variants for vertical tabs?
> 
> If yes, what is the problem? animating? positioning? sizing?

Firstly you need to fix this in `tabs.scss`:
image

Then stripe appears on the right side:
image

Last time I suggested moving stripe under `ul.iui-tabs`. It seems that here is the case where it will help to position stripe on the right side of tabs.

src/tabs/borderless.scss
FlyersPh9: I'll remove them in the next push.

src/tabs/borderless.scss
FlyersPh9: Added `.iui-horizontal` in next push.

src/tabs/tabs.scss
FlyersPh9: Sorry, that was a mistake.  It's been fixed in the next push.

mayank99:
> Last time I suggested moving stripe under `ul.iui-tabs`. It seems that here is the case where it will help to position stripe on the right side of tabs.

Moving non-`li` elements under `ul` is not valid.

I've fixed the stripe positioning by using JS to set `left` equal to the width of the tabs.

backstop/tests/tabs.html
```diff
@@ -696,11 +1242,11 @@ <h2>Borderless</h2>
       <br />
 
       <div
-        class="iui-tabs-wrapper"
+        class="iui-tabs-wrapper iui-vertical iui-not-animated"
```
bentleyvk, line 1245: Huh?
vertiacl-tabs-glitch

src/tabs/tabs.scss
```diff
@@ -121,24 +146,16 @@
     &::after {
       content: '';
       position: absolute;
-      left: 0;
-      bottom: 0;
-      width: 100%;
-      height: 0;
       @media (prefers-reduced-motion: no-preference) {
-        transition: height $iui-speed-fast ease-out;
+        transition: width $iui-speed-fast ease-out, height $iui-speed-fast ease-out;
```
bentleyvk, line 150: I think this one is causing strange switch animation.
vertiacl-tabs-glitch2

src/tabs/borderless.scss
```diff
@@ -9,6 +9,11 @@
     padding: 0 ($iui-component-padding-horizontal * 2);
     background-color: transparent;
 
+    &::after {
```
bentleyvk, line 12: Is this part needed?

src/tabs/tabs.scss
```diff
@@ -112,6 +111,32 @@
...
+
+    ~ .iui-tabs-content {
+      flex-grow: 1;
+    }
+  }
+}
+
 @mixin iui-tab-not-animated {
   .iui-tab {
     @media (prefers-reduced-motion: no-preference) {
```
bentleyvk, line 142: It seems like that one rule is overriding another:
image

src/tabs/tabs.scss
```diff
@@ -112,6 +111,32 @@
...
+    display: inline-flex;
+    flex-direction: column;
+
+    li,
+    .iui-tab {
+      width: 100%;
+    }
+
+    ~ .iui-tabs-content {
+      flex-grow: 1;
```
bentleyvk, line 135: Is it needed?

src/tabs/tabs.scss
```diff
@@ -112,6 +111,32 @@
...
+  .iui-tabs {
+    display: flex;
+  }
+}
+
+@mixin iui-tabs-vertical {
+  flex-direction: row;
+
+  .iui-tabs {
+    display: inline-flex;
```
bentleyvk, line 126: It seems like these two lines can be omited and it will look the same.

src/tabs/tabs.scss
```diff
@@ -150,25 +167,39 @@
...
+    .iui-tab-stripe {
+      height: $iui-xxs;
+      top: $iui-component-height - $iui-xxs;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: width $iui-speed-fast ease-out, left $iui-speed-fast ease-out;
+      }
+    }
+
+    .iui-large ~ .iui-tab-stripe {
+      top: $iui-component-height-large - $iui-xxs;
```
bentleyvk, line 193: @mayank99 Maybe we should also set this in JS like `left` for vertical tabs.

src/tabs/tabs.scss
bentleyvk: Horizontal Pill tabs also look strange with `iui-not-animated`. Maybe we don't need any transition effect then.

src/tabs/tabs.scss
mayank99: What would be the benefit (other than consistency)? It doesn't change so it made sense to keep it in CSS

backstop/tests/tabs.html
FlyersPh9: I guess the classname `iui-not-animated` isn't exactly true 🤔.  Maybe it should be named `.iui-less-animated` but that doesn't sound great.

mayank99: Bring back `iui-idk`

src/tabs/tabs.scss
FlyersPh9: I changed the width and height transitions to use `ease` instead of `ease-out` which caused a laggy visual effect.

src/tabs/borderless.scss
FlyersPh9: Without that CSS I'm getting this incorrect placement
Screen Shot 2021-08-17 at 12 51 32 PM

src/tabs/tabs.scss
FlyersPh9: Without this, if there is little within the content area it does not take the full parent's width.  To demonstrate I've turned the content area pink.

Without this CSS:
Screen Shot 2021-08-17 at 12 53 32 PM

With this CSS:
Screen Shot 2021-08-17 at 12 55 03 PM

src/tabs/tabs.scss
FlyersPh9: Nice catch, you are correct.  Removed.

src/tabs/tabs.scss
FlyersPh9: I think this is fixed now, the box-shadow transition was leftover code.

src/tabs/tabs.scss
mayank99: `ease` fixed it in vertical+borderless, but the animation still looks strange in vertical+default and all horizontal ones

src/tabs/tabs.scss
bentleyvk: I am not a fan of calculating stuff in CSS 😃 If one part changes, then another can break. But we can leave it as it is.

src/tabs/tabs.scss
bentleyvk: For me it also still looks strange. Do we really need it? We have normal and nicely looking animation, for non-animated we could just turn of animation at all.

mayank99: We certainly need it for default variant because it doesn't have the JS animated stripe.

bentleyvk: What if we move this transition rule to active state? Then previous active mark would disappear immediately and new one would have animation.

FlyersPh9: I would be ok with moving the animation only to the active tab.

FlyersPh9: I thought this would work, but it seems to have the same effect as before:
```
.iui-active::after {
      transition: width $iui-speed-fast ease, height $iui-speed-fast ease;
    }
}
```

bentleyvk: I see there is several places where `transition` is set. Can you check whether they are needed?

FlyersPh9: Bingo!  You are correct, it's been fixed and I'm pushing the change now.

mayank99 (approved):

bentleyvk (approved):

Issue #225, opened 2021-08-04
Reducing specificity of button selectors


mayank99:
Currently, the long button selectors make it really hard to add custom styles.

Some requirements:
- button icon should have a [specificity](https://web.dev/learn/css/specificity/#visualizing-specificity) of 0-1-0 (by adding a flat level class).
- there should be a modifier for the default button (e.g. `.iui-button.iui-default`) so we can reference it directly without `:not(.iui-cta):not(.iui-borderless):not(.iui-high-visibility)`.
- hover and focus states should have lower specificity across the board, to allow easier overrides for component-specific use cases (e.g. in split header button) as well as user's custom use cases (e.g. different hover color).

mayank99:
Found another related problem (https://github.com/iTwin/iTwinUI/pull/253#discussion_r699658717): Currently, it's very hard to override size/padding of any button. For example, a small borderless button has a specificity of 0-3-0 (`.iui-button.iui-borderless.iui-small`).

Unsure how this could be solved though. Maybe the user should not use `small` if they want to override styles. And then we could change the `.iui-button.iui-borderless` selector to just `button.iui-borderless` (0-1-1, easier to override).

Issue #199, opened 2021-07-19
Fix clipped focus styling


FlyersPh9:
The focus styling can appear cut off on elements that are side by side (grouped buttons, tabs).  Proposed solution is to use an inset outline rather than box shadow.

PR #234, opened 2021-08-10
Update split header button active/disabled state styles

gretanausedaite:
- Removed .iui-active from child buttons in split header button
- Added style overrides to achieve same look
- Added disabled state styling using .iui-disabled and :disabled
- Removed some `svg.` selectors
- Increased mismatch threshold to fix random test failures


src/header/header.scss
```diff
@@ -13,7 +13,8 @@
   user-select: none;
   box-sizing: border-box;
   // iPhone notch support
-  padding: env(safe-area-inset-top) env(safe-area-inset-right) 0 env(safe-area-inset-left);
+  padding: env(safe-area-inset-top) env(safe-area-inset-right) 0
```
bentleyvk, line 16: I wonder why Greta PRs always fix formatting? @FlyersPh9

src/header/header.scss
```diff
@@ -319,7 +341,6 @@
       }
     }
 
-    // Removing button hovers and backgrounds and adding last child hover
```
bentleyvk, line 322: These comments was kind of useful. At least for me to read the code.

src/header/header.scss
```diff
@@ -304,7 +305,28 @@
   }
 
   &.iui-active {
```
bentleyvk, line 307: @FlyersPh9 @mayank99 We either make these changes or add active state to split button.
For me these changes seem better as they are small and easy to understand.

src/header/header.scss
mayank99: These formatting fixes are from Prettier and need to be done manually because we disabled autoformat (it was messing with Stylelint).

So Jon should always format before pushing.

src/header/header.scss
mayank99: Isn't there some mixin we can use?

mayank99: Oh, there isn't. In that case, I think we should extract out the active state logic (in `button/default.scss`) into a mixin that we can just include here.

After that, we can override anything if needed.

src/header/header.scss
```diff
@@ -291,6 +267,55 @@
   }
 }
 
+@mixin iui-header-button-active {
```
gretanausedaite, line 270: @mayank99 Is this good enough? This doesn't break any other button styles and works for split button.

mayank99: Can you move this mixin below split button? I can't even see the diff properly right now

mayank99: Are you sure we need all these extra rules? I thought we would need fewer overrides after using the mixin.

mayank99: Yeah, this is better. I did notice two different problems:

1. Focus styling for left button looks wrong (it should extend all the way to the left edge).
image

2. Disabled buttons still have hover effect and blue stripe.
image

src/header/header.scss
```diff
@@ -341,5 +307,96 @@
         }
       }
     }
+
+    &.iui-disabled {
+      .iui-button {
```
mayank99, line 312: I would suggest adding `.iui-header-button-icon { filter: grayscale(1) }` here

src/header/header.scss
```diff
@@ -341,5 +307,96 @@
...
+    // Blue background and darker blue bottom line
+    @include themed {
+      box-shadow: 0 $iui-xxs 0 0 t(iui-color-foreground-primary);
+      background-color: rgba(
+        t(iui-color-foreground-primary-rgb),
+        t(iui-opacity-6)
+      );
+    }
+
+    svg.iui-header-button-icon {
```
mayank99, line 342: `svg` should not be needed here

src/header/header.scss
```diff
@@ -341,5 +307,96 @@
...
+      }
+    }
+
+    .iui-icon {
+      @include themed {
+        fill: t(iui-icons-color-actionable-disabled);
+      }
+    }
+
+    svg.iui-header-button-icon {
```
mayank99, line 395: `svg` not needed here too

src/header/header.scss
```diff
@@ -341,5 +307,96 @@
...
+      @include themed {
+        fill: t(iui-icons-color-actionable-disabled);
```
mayank99, line 397: This can be removed as `.iui-icon` above does the same

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Unreleased
+
+`Date`
+
+### What's new
+
+### Fixes
+
+- **Fixed Split Header button active state styling** to require less class names.
```
mayank99, line 11: Grammar
```suggestion
- **Fixed Split Header button active state styling** to require fewer class names.
```

src/header/header.scss
mayank99: Actually, it should also be applied on normal header-button 🤔

backstop/tests/header.html
```diff
@@ -752,7 +752,21 @@ <h2>Slim</h2>
 
     <hr />
 
-    <h2>Split Menu Buttons</h2>
+    <h2>
+      Split Menu Buttons
+      <label
```
mayank99, line 757: This toggle switch will allow us to look at disabled state easily

src/header/header.scss
```diff
@@ -341,5 +313,95 @@
...
+    }
+  }
+
+  // Disabled state
+  &.iui-disabled,
+  &:disabled {
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-body-rgb), 0.05);
+      box-shadow: 0 $iui-xxs 0 0
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
```
mayank99, line 386: These values are very random and may not be "correct". Lets decide the visuals once Jon gets back. I will create a discussion.

mayank99 (approved):
I removed some extra code and finished the PR so we can release and use this in React. Should be good to go now.

There is still a slight problem with focus styling, but it should be fixed in a future PR (or in #199).
image

Also we should revisit header button styling after #225. The nesting/specificity in here is a mess.

src/button/split-menu.scss
```diff
@@ -44,4 +44,10 @@
       );
     }
   }
+
+  &.iui-disabled {
```
bentleyvk, line 48: What it gives?

backstop/scenarios/header.js
```diff
@@ -4,27 +4,32 @@ module.exports = [
...
     actions: [hover('#test-button-3'), focus('#test-button-4')],
     selectors: ['#demo-slim'],
+    misMatchThreshold: 0.02,
   }),
 
   //// Split
   scenario('Type Split State hover & focus', {
     actions: [hover('#test-button-6'), focus('#test-button-6')],
     selectors: ['#demo-split'],
+    misMatchThreshold: 0.05,
```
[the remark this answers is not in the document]
bentleyvk: I think this will be Chrome again rendering images a bit different every time. Because previously we had the same problem with `Tile` on hover where images are scaled up. I then made screenshot, refreshed, made another screenshot and compared them with image comparer. Turns out images always differs a bit in few pixels.

src/button/default.scss
```diff
@@ -49,25 +49,29 @@
...
+    }
+  }
+}
 
-      > .iui-label {
-        @include themed {
-          color: t(iui-color-foreground-primary);
-        }
-      }
+@mixin iui-button-active {
```
bentleyvk, line 57: I thought we will reuse this mixin inside header button but as I understand it really can't be reused there. So maybe undo this change.

bentleyvk (approved):

PR #247, opened 2021-08-18
fix(ButtonGroup): Use flex instead of inline-flex

mayank99:
Changed to display: flex becuase inline-flex messes up responsive width.

Added back inline-flex to split-menu because other buttons are all inline.

Set requireSameDimensions to false for button group tests because there was random space at the bottom.


src/button/button-group.scss
```diff
@@ -3,7 +3,7 @@
 @import '../style/index';
 
 @mixin iui-button-group {
-  display: inline-flex;
+  display: flex;
```
bentleyvk, line 6: Ugh, I think `inline-flex` was intentional that it would take as much place as needed.

backstop/scenarios/button.js
```diff
@@ -15,9 +15,11 @@ module.exports = [
   }),
   scenario('Grouped Default', {
     selectors: ['#demo-grouped-default'],
+    requireSameDimensions: false,
```
bentleyvk, line 18: Why not just replace test images?

src/button/button-group.scss
mayank99: inline sucks though

backstop/scenarios/button.js
mayank99: hmm, good point. I don't like replacing test images in general

src/button/button-group.scss
bentleyvk: In your case, yes 😀 But for the user it doesn't take whole width, only as much as needed.

mayank99:
Closing this PR as it might break layouts.

Will now conditionally apply `width: 100%` for overflow cases.

PR #248, opened 2021-08-18
ci: Combine publish.yml into release-please.yml

mayank99:
Since publish workflow cannot be automatically triggered when release is created, I made it part of the same workflow.


.github/workflows/publish.yml
```diff
@@ -1,50 +0,0 @@
...
-    runs-on: ubuntu-latest
-    steps:
-      - name: Update release title to match tag name
-        uses: actions/github-script@v4
-        with:
-          script: |
-            const { data: release } = await github.repos.getReleaseByTag({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              tag: ${{ github.ref }}
```
bentleyvk, line 41: Why changed this part?

mayank99: previously this was meant to be triggered by a release, whose GITHUB_REF is the tag

now it will be triggered by `push` to main

bentleyvk: Is it possible to have wrong latest tag?

mayank99: I doubt. This job is guarded by the `if` condition that uses `release_created`.

bentleyvk (approved):

PR #251, opened 2021-08-19
feat(InformationPanel): Add Information Panel

### LINKED ISSUE #226: Add InfoPanel
#### Feature
Add InfoPanel component that exists in the bwc-react library as we can't find any alternative to it inside the itwinui-react.
**bwc-react/innersource/InfoPanel**

image

FlyersPh9:
Introducing the **Information Panel**, closes #226.
information-panel

Includes editable state:
Screen Shot 2021-08-19 at 10 37 20 AM

Includes option to change from docked right to docked bottom as seen [here](https://irtgws.axshare.com/#g=1&p=navigate_inside_properties_panel).


src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,140 @@
...
+    overflow-y: overlay;
+
+    > hr {
+      @include themed {
+        border-top: 1px solid t(iui-color-background-4);
+      }
+    }
+  }
+
+  .iui-information-table {
```
FlyersPh9, line 105: Unsure if this belongs in the information panel SCSS or if it belongs in the demo HTML `<style>`.

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,140 @@
...
+    right: 0;
+    opacity: 1;
+  }
+
+  > .iui-resizer {
+    height: 100%;
+    width: $iui-m;
+    position: absolute;
+    top: 0;
+    left: -17px; // Why?
```
FlyersPh9, line 38: Unsure why this needs to be -17px, I would think -16px would work but it does not.

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,140 @@
...
+  opacity: 0;
+  display: flex;
+  flex-direction: column;
+  width: $iui-3xl * 4;
+  min-width: $iui-3xl * 2;
+  height: 100%;
+  box-sizing: border-box;
+  box-shadow: -1px 0 14px rgba(0, 0, 0, 0.25);
+  transition: right $iui-speed-fast ease-out, opacity $iui-speed-fast ease;
+  z-index: 2; // Needs to be higher than table's resizers.
```
FlyersPh9, line 23: This was needed because otherwise the resizer from the table header appears above the information panel.

FlyersPh9:
The standard documentation suggests that the information panel should push the table content over rather than overlay on top of the table content.  I am unsure how to do this & unsure if that is actually the right decision.

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+
+  .iui-information-header,
+  .iui-information-body {
+    padding-left: $iui-sm;
+    padding-right: $iui-sm;
+    box-sizing: border-box;
+  }
+
+  .iui-information-header {
+    width: 100%;
```
veekeys, line 77: seems like not needed

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+    height: $iui-baseline * 6;
+    display: flex;
+    align-items: center;
+    @include themed {
+      background-color: t(iui-color-background-3);
+    }
+
+    .iui-selection-icon {
```
veekeys, line 85: not sure about naming...

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+      width: 24px;
```
veekeys, line 86: variables

src/information-panel/information-panel.scss
veekeys: I guess, I would have just label div with icon and text inside. Then we could use `space-between` on flex container and get rid of `margin-right: auto` on `iui-selection`

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  .iui-information-body {
+    padding-top: $iui-baseline;
+    padding-bottom: $iui-baseline;
+    height: 100%;
+    overflow-x: hidden;
+    overflow-y: overlay;
+
+    > hr {
```
veekeys, line 103: wow..not sure.. maybe lets just have `iui-separator` on simple div...?

src/information-panel/information-panel.scss
veekeys: Hmm.. for sure need to move out of it..Need to think, how we could provide these (if we do that at all)

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+
+  .iui-information-table {
+    table-layout: fixed;
+    width: 100%;
+
+    tr {
+      height: $iui-baseline * 3;
+
+      td {
+        vertical-align: middle;
```
veekeys, line 118: not sure it does anything..

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+
+    .iui-information-table-center {
+      display: flex;
+      align-items: center;
+      gap: $iui-s;
+    }
+
+    label {
```
veekeys, line 141: expectation is to use our labeled input here, so we dont need to do anything with them..

src/information-panel/information-panel.scss
veekeys: After removing translate, left is -8px (which is half the resizer div size, cause the line is in the center of it...)

src/information-panel/information-panel.scss
veekeys: fixed

src/information-panel/information-panel.scss
FlyersPh9: Removed.

src/information-panel/information-panel.scss
FlyersPh9: You are correct, removed the translate and now using -8px.

src/information-panel/information-panel.scss
FlyersPh9: Good suggestion, done this in my next push.

src/information-panel/information-panel.scss
FlyersPh9: I'm reading that [an `<hr>` is better for screen readers than a `<div class="iui-separator" />`](https://stackoverflow.com/questions/41315990/hr-element-vs-1px-high-div-vs-1px-border).

src/information-panel/information-panel.scss
FlyersPh9: Removed.

src/information-panel/information-panel.scss
veekeys: not sure this push ever happened though..

FlyersPh9: You're right, I forgot to push before the weekend.  It'll get pushed today.

src/information-panel/information-panel.scss
```diff
@@ -76,18 +76,34 @@
...
-    .iui-selection-icon {
-      width: $iui-l;
-      height: $iui-l;
-      margin-right: $iui-s;
+    .iui-header-label {
+      display: flex;
+      align-items: center;
+      overflow: hidden;
+
+      span {
```
veekeys, line 69: not sure we should tackle just span... Dont like, but maybe just `*` in here for handling overflow...?

src/information-panel/information-panel.scss
FlyersPh9: Moved this CSS into the HTML.

src/information-panel/information-panel.scss
FlyersPh9: I tried this after reading your comment.  I changed the code from this…
```
<td>
  <label for="year">Year recorded</label>
</td>
<td>
  <input id="year" value="2021" class="iui-input iui-small" />
</td>
```
…to this…
```
<td>
  <label class="iui-input-container" for="year">
    <div class="iui-label">Year recorded</div>
  </label>
</td>
<td>
  <input id="year" value="2021" class="iui-input iui-small" />
</td>
```
…and this was the end result. Musical Genre is unchanged, only Year Recorded was changed so we can visually compare the two.
Screen Shot 2021-09-14 at 3 27 20 PM

veekeys (approved):
We can complete it and have panel elements in separate PR, IMO.
#292

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,189 @@
...
+  .iui-information-header {
+    height: $iui-baseline * 6;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    @include themed {
+      background-color: t(iui-color-background-3);
+    }
+
+    .iui-header-label {
```
veekeys, line 64: we need to rename this
`iui-information-header-label`

src/information-panel/information-panel.scss
```diff
@@ -0,0 +1,189 @@
...
+        width: $iui-icons-large;
+        height: $iui-icons-large;
+        margin-right: $iui-s;
+        @include themed {
+          fill: t(iui-icons-color);
+        }
+      }
+    }
+
+    .iui-header-buttons {
```
veekeys, line 87: rename to `iui-information-header-actions`

src/information-panel/information-panel.scss
FlyersPh9: Done. We should copy this naming format in #264 as well.

src/information-panel/information-panel.scss
FlyersPh9: Done. We should copy this naming format in #264 as well.

PR #253, opened 2021-08-19
feat: Add Color Picker

saaaaaally:
Added basic and advanced color picker.

image


mayank99:
3 major things from my perfunctory look before I do any "review":
- use variables, not magic numbers
- read about CSS/Sass specificity and nesting
- read about focus vs focus-visible (browser support/fallback, respecting input modality, etc)

1 minor thing:
- run prettier ("Format Document" Alt+Shift+F) before pushing

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+@import '../style/index';
+
+@mixin iui-color-picker {
+  width: 250px;
+  border-radius: 6px;
+  box-shadow: $iui-elevation-2;
+  display: flex;
+  flex-wrap: wrap;
+  padding-left: 10px;
+  padding-top: 10px;
```
mayank99, line 12: Use the variables from [`space.scss`](https://github.com/iTwin/iTwinUI/blob/main/src/style/space.scss).

https://github.com/iTwin/iTwinUI/blob/5b9c5f42dcf42cc449a191459e8824b3de89953f/CONTRIBUTING.md#L69

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+  @include themed {  background: rgb(255, 255, 255);}
+
+
+  .iui-color-option {
+    height: 20px;
+    width: 20px;
+    margin-right: 10px;
+    margin-bottom: 10px;
```
mayank99, line 20: Variables here too.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+
+@mixin iui-color-picker {
+  width: 250px;
+  border-radius: 6px;
+  box-shadow: $iui-elevation-2;
+  display: flex;
+  flex-wrap: wrap;
+  padding-left: 10px;
+  padding-top: 10px;
+  @include themed {  background: rgb(255, 255, 255);}
```
mayank99, line 13: What were you trying to achieve here? This looks like you started something and abandoned it.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+    height: 20px;
+    width: 20px;
+    margin-right: 10px;
+    margin-bottom: 10px;
+    cursor: pointer;
+    border-radius: 4px;
+
+    &.iui-light {
+      box-shadow: $iui-elevation-2;
+    }
```
mayank99, line 26: Not a fan of this class. Also not a fan of elevation. In this case, other colors have no elevation, giving a very inconsistent look.
image

If you apply an inset box-shadow instead, you won't need special handling of light colors, as it will always look good.

Similar idea as this: https://twitter.com/steveschoger/status/1064541476615593984

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+
+    &.iui-active {
+      @include themed {
+        box-shadow: 0 0 0 2pt rgb(100, 149, 237);
```
mayank99, line 30: Mixing pt and px is usually a bad idea.

And the color itself is a bit random.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    &:hover { 
+      @include themed {
+        box-shadow: 0 0 0 2pt rgb(170, 170, 170);
+      }
+    }
+
+    &:focus,
+    &:focus-visible {
```
mayank99, line 41: This will break in Safari (and IE). And even in other browsers, it will show focus effect on click (which is wrong).

Recommend reading this: [https://www.tpgi.com/focus-visible-and-backwards-compatibility](https://www.tpgi.com/focus-visible-and-backwards-compatibility/#:~:text=the%20most%20viable%20(though%20still%20not%20particularly%20elegant)%20solution%20may%20be%20to%20use%20the%20%3Anot()%20negation%20pseudo-class%2C%20and%20to%20(paradoxically)%20define%20styles%20not%20for%20%3Afocus-visible%2C%20but%20to%20undo%20%3Afocus%20styles%20when%20it%20is%20absent.)

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+      @include themed {
+        box-shadow: 0 0 2pt 2pt rgb(170, 170, 170);
+      }
+    }
```
mayank99, line 45: Looking at the bigger picture, you might just be able to use the `iui-focus` mixin here.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-color-picker {
+  width: 250px;
+  border-radius: 6px;
+  box-shadow: $iui-elevation-2;
```
mayank99, line 8: You could use the `iui-dropdown` mixin for the elevated popover look.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,47 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  padding-left: 10px;
+  padding-top: 10px;
+  @include themed {  background: rgb(255, 255, 255);}
+
+
+  .iui-color-option {
```
mayank99, line 16: I think this should be its own mixin. There is no reason to nest this inside iui-color-picker.

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,190 @@
...
+  lang="en-US"
+  id="theme"
+>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1"
+    >
+    <title>color-picker | iTwinUI</title>
```
mayank99, line 16: Why lower case?

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,190 @@
...
+    <h2>Basic</h2>
+    <section id="demo-default">
+      <div class="iui-color-picker">
+        <div
+          style="background: #FFF;"
+          class="iui-color-option iui-light"
+          tabindex="0"
+        ></div>
+        <div
+          style="background: #00121D;"
```
mayank99, line 41: You could use an inline [CSS variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) instead of setting inline background.

See example in [`badge.html`](https://github.com/iTwin/iTwinUI/blob/main/backstop/tests/badge.html)

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,190 @@
...
+
+    <h2>Basic</h2>
+    <section id="demo-default">
+      <div class="iui-color-picker">
+        <div
+          style="background: #FFF;"
+          class="iui-color-option iui-light"
+          tabindex="0"
+        ></div>
+        <div
```
mayank99, line 40: Maybe use span instead of div?

mayank99 (commented):
Expanding a bit on my points above.

backstop/tests/color-picker.html
saaaaaally: Oh didn't notice that, it was just the title it gave when using yarn createComponent, but I'll fix it

src/color-picker/color-picker.scss
saaaaaally: It was to make the background white, but I think it is not necessary

mayank99: You should be using the `iui-color-background-1` variable for that, as now it's a white background even on dark theme.

But if you use the `iui-dropdown` mixin, you won't need it I think

backstop/tests/color-picker.html
mayank99: Also change in line 29

src/color-picker/color-picker.scss
saaaaaally: So if I use iui-focus, would this be instead of all the focus/focus-visible stuff I had? 

This is what using iui-focus looks like:
image

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../menu/index';
+
+@mixin iui-color-picker {
+  @include iui-dropdown;
+  width: $iui-baseline * 27;
```
mayank99, line 8: baseline should only be used for height/vertical spacing.

And because of how random this is, you have some extra space on the right side.
image

For calculating, I would suggest doing this:
1. Decide how many columns you want to show (lets say 8).
2. Add the widths and spaces around those columns (24\*8 + 12\*7 + 16\*2 = 308)
3. Find a large spacing variable and multiply it to get close to the number above (e.g. 64 * 5 = 320)

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+  border-radius: $iui-border-radius;
+  display: flex;
+  flex-wrap: wrap;
+  padding-left: $iui-sm;
+  padding-top: $iui-sm;
```
mayank99, line 13: why only left/top?

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+    width: $iui-baseline * 24;
+  }
+}
+
+@mixin iui-color-option {
+  @include iui-focus;
+  height: $iui-l;
+  width: $iui-l;
+  margin-right: $iui-sm;
+  margin-bottom: $iui-sm;
```
mayank99, line 25: why only right/bottom?

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,271 @@
...
+  </head>
+  <body class="iui-body">
+    <theme-button></theme-button>
+    <h1>Color Picker</h1>
+    <hr />
+
+    <h2>Basic</h2>
+    <section id="demo-basic">
+      <div class="iui-color-picker">
+        <div class="iui-color-palette">
```
mayank99, line 35: I don't think you need two divs here.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+      $iui-color-system-magenta 83%,
+      $iui-color-system-red 100%
+    );
+  }
+}
+
+@mixin iui-color-selection-wrapper {
+  position: absolute;
+  top: var(--top, 0%);
+  left: var(--left, 0%);
```
mayank99, line 112: We try to avoid absolute positioning unless absolutely necessary 😄

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+
+@mixin iui-color-palette {
+  width: auto;
+  display: flex;
+  flex-wrap: wrap;
+}
+
+@mixin iui-saved-colors {
+  margin-left: $iui-sm;
+}
```
mayank99, line 61: Something ain't right about these two mixins (iui-color-palette and iui-saved-colors). If you're finding yourself creating very small mixins like these (which don't mean much on their own), then maybe you need to refactor

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  &:hover {
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
+    }
+  }
+} ;
```
mayank99, line 144: Please run prettier before pushing to avoid these formatting blunders. If it doesn't fix it, then change it manually
```suggestion
}
```

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+  transform: translate(-3px, -3px);
+  @include themed {
+    box-shadow: rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-1)) 0
+      0 0 $iui-xxs - 1,
+      inset 0 0 0 $iui-xxs - 1
+      rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-6));
+    background-color: var(--selected-color);
+  }
+
+  &.iui-white {
```
mayank99, line 129: Why do you need to handle white differently?

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+@mixin iui-color-selection-wrapper {
+  position: absolute;
+  top: var(--top, 0%);
+  left: var(--left, 0%);
+}
+
+@mixin iui-color-selection {
+  @include iui-focus;
+  width: $iui-s + 2;
+  height: $iui-s + 2;
```
mayank99, line 118: The `+ 2` kinda defeats the purpose of these variables. Do you really need it? If `iui-s` is too small, try `iui-sm`.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+  width: auto;
+  display: flex;
+  flex-wrap: wrap;
+}
+
+@mixin iui-saved-colors {
+  margin-left: $iui-sm;
+}
+
+@mixin iui-color-selector {
```
mayank99, line 63: This one is very similar to the main color picker mixin, so you might need to refactor something.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,144 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  height: $iui-baseline * 16;
+
+}
+
+@mixin iui-color-field {
+  position: relative;
+  cursor: crosshair;
+  width: $iui-baseline * 19;
```
mayank99, line 75: This is very random and very fragile at the same time.

backstop/scenarios/color-picker.js
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
+  scenario('default', {
+    selectors: ['#demo-default'],
```
mayank99, line 9: Test scenarios need to be updated.

mayank99 (commented):
I think the advanced part could use some restructuring/cleanup.

Maybe @FlyersPh9 could provide some guidance on the values/visuals

src/color-picker/color-picker.scss
mayank99: ok I pulled your branch and realized it's the actual "dot" inside gradient, so `absolute` makes sense here, but it's not obvious from the mixin name so maybe rename it to something better.

src/color-picker/color-picker.scss
mayank99: I think this is a strong case for flex `gap` (#178).

For now we could use `gap` inside `@supports` and fallback to right/bottom margins

src/color-picker/color-picker.scss
saaaaaally: I wasn't sure how to name this clearly, but the white one is for the 'dot' to select a color on the slider, 
image
and the regular one is the 'dot' to select a color on the color field.
image
The colored dot shows the exact color that is selected while the white dot in the slider just shows what color will be shown in the color field. Also I am going to put different cursors on them (crosshair and grab)

src/color-picker/color-picker.scss
FlyersPh9: We discussed in a call today, we're going to set a max-width of 10 swatches in the basic color picker.

src/color-picker/color-picker.scss
FlyersPh9: We discussed options to eliminate the left / top specific padding.

src/color-picker/color-picker.scss
FlyersPh9: @bentleyvk [left a comment here](https://github.com/iTwin/iTwinUI/discussions/178#discussioncomment-1225992) and is unsure if this is a good idea or not.

backstop/tests/color-picker.html
FlyersPh9: I think it does make sense because it is 1 part within the advanced color picker.  Might be able to exclude the div in the basic color picker if we use mixins tho.

src/color-picker/color-picker.scss
FlyersPh9: We discussed in the call making both circles to match the same size of the grab handle on the slider component.

FlyersPh9:
We also discussed in a call setting a max-height on the color swatches.  If we have a scenario where we have dozens of rows that we will only display so many rows (undecided on how many) and will apply a vertical scrollbar.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+  overflow-y: overlay;
+
+  &.iui-advanced {
+    max-width: ($iui-l * 7) + ($iui-xxs * 7) + ($iui-s * 6) + ($iui-sm * 2) + ($iui-s);
+    flex-wrap: wrap;
+  }
+}
+
+@mixin iui-color-swatch {
+  @include iui-focus;
```
gretanausedaite, line 30: Seems like grey border is removed when color swatch was focused (focus, not focus-visible). It's really visible with white color on white background and grey color on darker background. 
colorpickerfocus

veekeys: Focusing white color has this background issue...
image
image

Wonder, if we could do something about it, but should not be a showstopper for this PR, I guess..

-----------------

Also, there is no focus style on the active color. For example, same situation on date picker:
active:
image
focused active:
image

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+  &.iui-white {
+    transform: translate(-4px, -9px);
+    cursor: grab;
+    @include themed {
+      box-shadow: $iui-elevation-2;
+      background-color: #FFF;
+    }
+  }
+
+  &:hover {
```
veekeys, line 172: Not sure, but shouldnt color dot get grab cursor instead of keeping crosshair? Crosshair kinda looks weird..

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+    box-shadow: rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-1)) 0
+      0 0 $iui-xxs - 1,
+      inset 0 0 0 $iui-xxs - 1
+      rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-6));
+    background-color: var(--selected-color);
+  }
+
+  &.iui-white {
+    transform: translate(-4px, -9px);
+    cursor: grab;
```
veekeys, line 165: You might want to add also grabbing when it is `:active`. Check slider example (https://itwin.github.io/iTwinUI/backstop/minified/slider.html)

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,267 @@
...
+            <div
+              class="iui-color-dot iui-white"
+              style="--top: 14%;"
+              tabindex="0"
+            ></div>
+          </div>
+        </div>
+
+
+        <div style="width: 100%;">Saved Colors</div>
```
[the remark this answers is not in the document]
veekeys: Yea, agree. I think it should have some bottom margin, it looks too close (added g on purpose there :D):
image
We might need a new class on this one `iui-saved-color-label` or something to add missing rules (margin, overflow, etc..).

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+      box-shadow: 0 0 0 $iui-xxs t(iui-color-foreground-primary);
+    }
+  }
+
+  &:hover {
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5)),
+        inset 0 0 0 $iui-xxs
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
```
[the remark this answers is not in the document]
saaaaaally: This border around the color swatch was done intentionally to match the color picker from the ux standards 
image

But mines looks different because of the inset box shadow. This is what it look like without it, should I just remove it?
image

saaaaaally:
@mayank99 @veekeys @gretanausedaite I think all the comments should be resolved now, please take a look and let me know if there is still something that needs to be fixed

backstop/tests/color-picker.html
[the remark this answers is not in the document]
veekeys: Summarize:
- Double check the need of width. If its needed, screenshot would be useful, so we could double check, why this is appearing.
- Top/bottom margin should use baseline spacing. 
- Overflow still missing.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  &.iui-add-color-button {
+    @include iui-focus;
+    @include iui-button-borderless;
```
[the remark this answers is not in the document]
veekeys: `button` is already there, but some inline styles are there. Need to review them and see, what to do.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  @supports(gap: $iui-s) {
+    gap: $iui-s;
+  }
+}
+
+@mixin iui-color-selection-wrapper {
+  display: flex;
+  flex-wrap: wrap;
```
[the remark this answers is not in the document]
saaaaaally: It wasn't needed, so I removed it

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+  box-sizing: border-box;
+  border-radius: $iui-border-radius;
+  display: flex;
+  padding: $iui-baseline $iui-sm;
+  user-select: none;
+  overflow-y: overlay;
+
+  &.iui-advanced {
+    max-width: ($iui-l * 7) + ($iui-xxs * 7) + ($iui-s * 6) + ($iui-sm * 2) + ($iui-s);
+    flex-wrap: wrap;
```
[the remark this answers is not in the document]
saaaaaally: I did not make this change

src/color-picker/color-picker.scss
saaaaaally: The background issues were due to the inset box shadow getting removed when focusing so I changed the focus so that the inset box shadow remains when focused. 

When focusing on an active color the active box shadow (dark blue) gets replaced by the focus box shadow (light blue), the same thing happens when hovering over an active color (blue shadow gets replaced by the grey). I'm not sure what to do about this or if I should just leave it as is?

src/color-picker/color-picker.scss
saaaaaally: I made color dot have grab cursor

src/color-picker/color-picker.scss
saaaaaally: I added grabbing to :active

src/color-picker/color-picker.scss
mayank99: When you add interactivity (in React), you might want to change the cursor to `grabbing` when the color dot is being dragged.

To do that, you could add a modifier class which you can toggle when drag starts/ends.

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
mayank99: I would say remove it, but it still doesn't look great (definitely better than the gap though).

backstop/tests/color-picker.html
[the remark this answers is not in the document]
saaaaaally: I set width because I want the Saved Colors label have its own row, even when there are only a few colors
image

Without it the colors move up to the same row
image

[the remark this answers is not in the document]
mayank99: Oh I see what you mean now. See this is a situation where my other comment about adding containers would have helped. https://github.com/iTwin/iTwinUI/pull/253#discussion_r697712651

src/color-picker/color-picker.scss
mayank99: > When focusing on an active color the active box shadow (dark blue) gets replaced by the focus box shadow (light blue), the same thing happens when hovering over an active color (blue shadow gets replaced by the grey).

For hovering, I don't think it should replace the blue shadow. That's why I said "active swatch could either use a double stripe or a different hover effect".(https://github.com/iTwin/iTwinUI/pull/253#discussion_r697716908)

The same could apply to focus too.

mayank99: Also you might want to add some basic JS in your HTML file to toggle the `iui-active` class on click.

backstop/tests/color-picker.html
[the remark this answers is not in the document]
mayank99: Yes, you need new mixins/classes to nest your flex layout. You might be able to get away with just one container, but you might need two if the spacing is weird.
```html
<color-picker> // this will have flex-direction: column
  <selection-container> // this will be a new class with display: flex
    <selection ...>
    <slider ...>
  </selection-container>
  <div>Saved colors</div>
  <color-palette ...>
</color-picker>
```

Or you can leave `width: 100%` if that's too much.

Still need to tackle the other parts: margin values and overflow.

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: I added iui-color-palette-button mixin to style button

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: Focus should be blue now

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: Border was removed

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+  }
+}
+
+@mixin iui-color-swatch {
+  @include iui-focus;
+  height: $iui-l;
+  width: $iui-l;
+  cursor: pointer;
+  border-radius: $iui-border-radius-5;
```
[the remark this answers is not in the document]
saaaaaally: Inset and outset borders were changed to a thinner inset border and thicker outset border

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
mayank99: It doesn't really work, the height and padding get overridden by the higher specificity rules
image

So you'll need to find a way to boost the specificity to at least 0-3-0, preferably 0-4-0.

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
mayank99: oh it looks especially terrible in dark theme now, even worse than before
image

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
mayank99: That is not at all what I meant. I was talking about the border-radius, but what you've done now is change the `spread` of the box-shadows, which is looking really bad for the focus styling now.

To achieve the nicer look, you would need to use `:before`/`:after` pseudo-elements (or add new elements in your html) with different border radii. You don't have to do that in this PR, I guess.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
...
+
+  &:hover {
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs + 1
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5)),
+        $iui-color-swatch-box-shadow;
+    }
+  }
+
+  &:focus {
```
mayank99, line 68: If you add another block after this for `focus:not(:focus-visible)`, it will allow you more control over click vs keyboard, which you can use to improve your active state focus.

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../menu/index';
+@import '../button/borderless';
+
+$iui-color-field-size: $iui-baseline * 19;
+$iui-color-swatch-box-shadow: inset 0 0 0 $iui-xxs - 1
```
mayank99, line 8: I'd say leave this inset box-shadow spread as 1px, it looks fine. 

Also weirdly, using a variable doesn't make sense here because it's just 1px

```suggestion
$iui-color-swatch-box-shadow: inset 0 0 0 1px 
```

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
...
+    margin-right: 0;
+  }
+
+  @include themed {
+    box-shadow: $iui-color-swatch-box-shadow;
+  }
+
+  &.iui-active {
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs + 1 t(iui-color-foreground-primary),
```
mayank99, line 55: The extra `+ 1` for outset box-shadows looks really bad for active state, so I think you should revert it. Also, I think you might need to lower the opacity on the color.

Something like:
```suggestion
      box-shadow: 0 0 0 $iui-xxs rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-3))
```

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
...
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5)),
+        $iui-color-swatch-box-shadow;
+    }
+  }
+
+  &:focus {
+    outline: none;
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs + 1
+        rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-5)),
```
mayank99, line 72: I think this opacity might be too low, because it's hard to see focus effect in dark theme. Maybe try a higher value.
```suggestion
        rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-4)),
```

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
...
+      box-shadow: 0 0 0 $iui-xxs + 1
+        rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5)),
+        $iui-color-swatch-box-shadow;
+    }
+  }
+
+  &:focus {
+    outline: none;
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs + 1
```
mayank99, line 71: For focus and hover states, 3px spread is not too bad I guess, mainly because of the low opacity.

But maybe you could also try playing around with the blur-radius (the 3rd value) and share screenshots with results.
```suggestion
      box-shadow: 0 0 1px $iui-xxs + 1
```

src/color-picker/color-picker.scss
veekeys: `:active` here wont help?

mayank99: `:active` will be enough if we are ok with cursor changing when it leaves the selection area

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: I did notice the height was getting overridden but I was able to get the height I wanted by setting max-height. But I'm not sure how to boost the specificity?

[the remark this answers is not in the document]
mayank99: If you look at the html, it has `iui-button` and `iui-borderless` classes. You could repeat those classes in your selector, and also add `button`, to make it `0-3-1`.

If you remove the mixin, you could also add this selector in your main color palette mixin.
```scss
@mixin iui-color-palette {
  ... // other styles

  button.iui-button.iui-borderless {
    ... // height, width, padding
  }
}
```

Alternatively, you could do the [class repeat hack](https://css-tricks.com/a-specificity-battle/#step-7-stacked-classes) like this:
```scss
@mixin iui-color-palette-button {
  button#{&}#{&}#{&} {
    ... // height, width, padding
  }
}
```

I think it also depends on how you expect the user to put this button here, in React.

backstop/tests/color-picker.html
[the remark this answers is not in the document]
saaaaaally: I added flex-direction to iui-advanced, saved-colors wrapper mixin, and margins and overflow to iui-saved-colors

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,186 @@
...
+
+  @include themed {
+    box-shadow: inset 0 0 0 $iui-xxs
+      rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
+    border: 1px solid t(iui-color-background-1);
+  }
+
+  &.iui-active {
+    @include themed {
+      box-shadow: 0 0 0 $iui-xxs t(iui-color-foreground-primary);
```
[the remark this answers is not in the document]
saaaaaally: I added both a blue and grey shadow when hovering over an active color

src/color-picker/color-picker.scss
saaaaaally: I added both a dark blue and light blue shadow when focusing on an active color.

Also you can now click on colors to toggle iui-active

src/color-picker/color-picker.scss
```diff
@@ -0,0 +1,181 @@
...
+}
+
+@mixin iui-color-dot {
+  position: absolute;
+  top: var(--top, 0%);
+  left: var(--left, 0%);
+  width: $iui-m;
+  height: $iui-m;
+  border-radius: 50%;
+  transform: translate(-8px, -8px);
```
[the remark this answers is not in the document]
saaaaaally: Changed these to variables

src/color-picker/color-picker.scss
saaaaaally: Changed to 1px

src/color-picker/color-picker.scss
saaaaaally: Reverted outset active box shadow and changed opacity (see $iui-active-box-shadow)

src/color-picker/color-picker.scss
saaaaaally: Changed opacity for color swatch focus

src/color-picker/color-picker.scss
saaaaaally: Changed blur-radius for focus box shadow 
Before & After:
image
image

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
mayank99: I think you can do `(-$iui-s) (-$iui-s)` instead of `* -1`

backstop/tests/color-picker.html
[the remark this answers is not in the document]
mayank99: Is the `saved-colors-wrapper` really needed? Seems it only applies margin-top, which `saved-colors` was applying anyway.

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: I refactored the borderless button

backstop/tests/color-picker.html
```diff
@@ -0,0 +1,264 @@
...
+        <div class="iui-color-palette">
+          <button class="iui-button iui-borderless iui-small iui-color-palette-button">
+            <!-- Use add-circular.svg from @itwin/itwinui-icons package. -->
+            <svg
+              class="iui-icon"
+              aria-hidden="true"
+              viewBox="0 0 16 16"
+            >
+              <path d="M8,0C3.6,0,0,3.6,0,8s3.6,8,8,8s8-3.6,8-8S12.4,0,8,0z M13,9H9v4H7V9H3V7h4V3h2v4h4V9z" />
+            </svg>
```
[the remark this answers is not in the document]
saaaaaally: I changed it to svg-add-circular

src/color-picker/color-picker.scss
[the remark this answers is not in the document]
saaaaaally: Changed

src/color-picker/color-picker.scss
saaaaaally: I cleaned it up to hopefully make it more clear, but I am trying to make the hover over active color thicker than the hover over non-active color because active color also has the active box shadow on it

src/color-picker/color-picker.scss
saaaaaally: I'm not sure what I need to do here

src/color-picker/color-picker.scss
mayank99: Why not keep 3px spread and 1px blur in all focus/hover states (including active focus)? It's simpler and looks ok in all cases.
focus

backstop/tests/color-picker.html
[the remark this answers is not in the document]
saaaaaally: I'm not really sure but without it if there are a lot of colors the saved colors label gets covered up for some reason 
image

src/color-picker/color-picker.scss
saaaaaally: I changed the focus and hover shadows to 3px and 2px

src/color-picker/color-picker.scss
saaaaaally: Added

mayank99: Surely you see the problem in this code 😄 


See my other comment for suggested fix.

backstop/tests/color-picker.html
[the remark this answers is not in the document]
mayank99: If you add `flex-shrink: 0` to `saved-colors`, you can remove the wrapper.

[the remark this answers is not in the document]
saaaaaally: I added it and removed saved-colors-wrapper

src/color-picker/color-picker.scss
saaaaaally: Done

mayank99: You forgot the `themed` mixin so I doubt it's even working. Also you should put the whole `.iui-active` block after the base hover/focus states to help the cascade.

saaaaaally: I added the themed mixin

mayank99: The only thing missing now is the `iui-focus` mixin on color dot

saaaaaally: Should I remove $iui-focus-box-shadow and $iui-hover-box-shadow from the color dot and have only the iui-focus mixin? Or keep both?

mayank99: Keep hover, replace the `:focus` block with `@include iui-focus`

saaaaaally: Ok added

mayank99: Ah wait, I see you have a custom box shadow on the dot too, so the mixin will not work (it overrides the iui-elevation shadow). You will need to manually define `focus:not(:focus-visible)` style, similar to how you did for swatch.

Or I guess revert to the previous commit, it was not too bad.

saaaaaally: Ok I reverted it

mayank99 (approved):
Looks good 👍 (unless we need some changes for React)

For some reason, the default color picker tests keep failing (despite no visual differences), so try re-running and/or updating images just in case.

PR #254, opened 2021-08-19
feat(Icons): size variables are now exported

### LINKED ISSUE #182: Icons variables are not exported to users
Review icons code, cleanup and export useful variables to the users through our style variables.

FlyersPh9:
Exporting icon variables, closes #182.

### DIFF
--- a/src/icon/mixins.scss
+++ b/src/icon/mixins.scss
@@ -1,6 +1,6 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import './variables';
+@import '../style/index';
 
 @mixin iui-icons-default {
   @include iui-icons-color();
--- a/src/icon/variables.scss
+++ b/src/icon/variables.scss
@@ -1,6 +1,5 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
 
 /// Sizes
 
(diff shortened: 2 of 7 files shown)


mayank99 (approved):

src/icon/mixins.scss
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import './variables';
+@import '../style/index';
```
veekeys, line 3: Well then need to go through all the imports of `/variables` and remove/replace them.

src/icon/variables.scss
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
 
 /// Sizes
```
veekeys, line 5: Issue says review icons code too... We dont want to do anything with the mixins part? Just having an icon size variable is not that useful, cause you will need to write something like `width: $iui-icons-default;
  height: $iui-icons-default;`.
  But if there was a mixin, you could just include it and forget.. So not sure. Just those current mixins include colors too, which might not be that needed. 
  If we remove colors from mixins, then is a little bit bigger refactor (we could replace all instances of ours of width and height with just the mixin).

src/icon/mixins.scss
mayank99: This is now done. Should be good to merge

PR #259, opened 2021-08-20
feat(ExpandableBlock): update animation to use transition-group

### LINKED ISSUE #223: ExpandableBlock: animation should be updated
We got an issue in React https://github.com/iTwin/iTwinUI-react/issues/243 that `Select` is "jumping" in `ExpandableBlock`.
It is caused by having `transform: scaleY(...)`. Need to remove it and just make the same animation as `Table` has for expandable row.

mayank99:
Replaced all transforms with new classes from react-transition-group. Added new mixin iui-transition-group and reused in Table.

Closes #223

expand-animation

Also:
- moved padding into a new child div because it messes up transition height
- added JS in demo page to remove/add expandable-content from DOM
- changed focus styling to use `iui-focus`
- removed `border` from transition


src/expandable-block/block.scss
```diff
@@ -71,34 +70,45 @@
...
       border-bottom: 1px solid t(iui-color-background-4);
       border-right: 1px solid t(iui-color-background-4);
       border-left: 1px solid t(iui-color-background-4);
     }
+
+    > div {
+      padding: $iui-baseline $iui-sm;
+    }
+
+    // #region react-transition-group classes
```
bentleyvk, line 94: When you wrote it as a region, now it really makes sense to have a `mixin` and reuse it in `Table`.

backstop/tests/expandable-block.html
```diff
@@ -23,6 +23,15 @@
       href="../assets/demo.css"
     />
     <script src="../assets/theme.js"></script>
+    <script>
+      customElements.define('expandable-content', class extends HTMLElement {
+        connectedCallback() {
+          const innerHTML = `<div class="iui-expandable-content"><div>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</div></div>`;
```
bentleyvk, line 29: Maybe have just have Lorem ipsum part here and leave `iui-expandable-content` as it was.

src/expandable-block/block.scss
mayank99: Done

backstop/tests/expandable-block.html
mayank99: Done

backstop/tests/expandable-block.html
```diff
@@ -83,11 +85,9 @@ <h2 id="default">Default</h2>
         </div>
       </div>
       <div class="iui-expandable-content">
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
-        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
-        commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
-        velit esse cillum dolore eu fugiat nulla pariatur.
+        <div>
```
bentleyvk, line 88: Is this `div` really needed?

bentleyvk (approved):

PR #261, opened 2021-08-23
Added classes for toaster placement

gretanausedaite:
Added classes for toaster placement:
- bottom
- bottom left/right
- top left/right

### DIFF
--- a/backstop/tests/toast-notification.html
+++ b/backstop/tests/toast-notification.html
@@ -265,5 +265,44 @@
       </div>
 
     </div>
+    <div class="iui-toast-wrapper iui-position-bottom-end">
+
+      <div
+        class="iui-toast iui-positive"
+        id="demo-positive"
+      >
+        <div class="iui-status-area">
+          <!-- Use status-success.svg from @itwin/itwinui-icons package. -->
+          <svg
+            class="iui-icon"
+            aria-hidden="true"
+            viewBox="0 0 16 16"
+          >
+            <path d="m8 0a8 8 0 1 0 8 8 8 8 0 0 0 -8-8zm-1.35 12-3.65-3.41 1.4-1.3 2.36 2.2 4.83-4.49 1.41 1.29z" />
+          </svg>
+        </div>
+        <div class="iui-message">Toaster position bottom end</div>
+        <a
+          class="iui-anchor"
+          href="#"
+        >View the report</a>
+        <button
+          aria-label="Close"
+          type="button"
+          class="iui-button iui-borderless iui-small"
+        >
+          <!-- Use close-small.svg from @itwin/itwinui-icons package. -->
+          <svg
+            class="iui-icon"
+            aria-hidden="true"
+            viewBox="0 0 16 16"
+          >
+            <path
+              d="m12.5 2-4.5 4.5-4.5-4.5-1.5 1.5 4.5 4.5-4.5 4.5 1.5 1.5 4.5-4.5 4.5 4.5 1.5-1.5-4.5-4.5 4.5-4.5z" />
+          </svg>
+        </button>
+      </div>
+    </div>
+
   </body>
 </html>
--- a/src/toast-notification/classes.scss
+++ b/src/toast-notification/classes.scss
@@ -67,6 +67,35 @@
       animation: closeAnimation 120ms cubic-bezier(0.4, 0, 1, 1) both;
     }
   }
+
+  &.iui-placement-top-start {
+    right: unset;
+    left: 0;
+  }
+
+  &.iui-placement-top-end {
+    right: 0;
+    left: unset;
+  }
+
+  &.iui-placement-bottom {
+    top: unset;
+    bottom: 0;
+  }
+
+  &.iui-placement-bottom-start {
+    right: unset;
+    left: 0;
+    top: unset;
+    bottom: 0;
+  }
+
+  &.iui-placement-bottom-end {
+    right: 0;
+    left: unset;
+    top: unset;
+    bottom: 0;
+  }
 }
 
 .iui-toast {


backstop/tests/toast-notification.html
```diff
@@ -265,5 +265,44 @@ <h1>Toast notification</h1>
       </div>
 
     </div>
+    <div class="iui-toast-wrapper iui-placement-bottom-end">
+
+      <div
+        class="iui-toast iui-positive"
+        id="demo-positive"
```
mayank99, line 272: Looks like this id is duplicate so it's producing lots of extra images

src/toast-notification/classes.scss
```diff
@@ -67,6 +67,35 @@
       animation: closeAnimation 120ms cubic-bezier(0.4, 0, 1, 1) both;
     }
   }
+
+  &.iui-placement-top-start {
+    right: unset;
```
mayank99, line 72: Why do you need unsets?

mayank99 (commented):
Before I approve, can you explain where/how these classes are used?

Also maybe create a draft PR in React with a working implementation (using `yarn link`). Because right now, we can't be sure that these changes are enough. And we shouldn't go back and forth if we need more.

mayank99:
Update: I pulled and linked https://github.com/iTwin/iTwinUI-react/pull/276 and have two observations:

1. `iui-placement-top` is unused, but probably shouldn't be. In fact, adding that class might help you get rid of some `unset`s.

2. The positioning itself looks ok, but the enter animation is broken for these new placements. To fix that, I suspect that you will need to take into account the order of the toasts. May want to do that before finalizing this PR.

backstop/tests/toast-notification.html
mayank99: I would just remove the id if you're not using it.

Also there are still duplicate images because this selector appears twice in the DOM:
https://github.com/iTwin/iTwinUI/blob/e448f3fd9bf5bbc017fc2a193010ad32b036055e/backstop/scenarios/toast-notification.js#L6

You could either update the scenario or make sure there is only one wrapper in the html.

src/toast-notification/classes.scss
gretanausedaite: Removed unsets. Moved all placement stuff to appropriate classes.

backstop/tests/toast-notification.html
gretanausedaite: Removed additional toaster-wrapper component, added JS to switch between placements. Added tests for those placements

gretanausedaite:
> 2. The positioning itself looks ok, but the enter animation is broken for these new placements. To fix that, I suspect that you will need to take into account the order of the toasts. May want to do that before finalizing this PR.

Do you have in mind appear animation or sliding animations when new toasts appear?

backstop/tests/toast-notification.html
```diff
@@ -30,13 +30,166 @@
...
+            focusable="false"
+          >
+            <circle
+              cx="8"
+              cy="8"
+              r="6"
+            />
+          </svg>
+        </span>
+        <span class="iui-label">Top Start</span>
```
FlyersPh9, line 64: Can we relabel these so it says Top Left instead of Top Start?  Same goes with Top End to Top Right.

FlyersPh9:
It's not ideal that these radios get covered up but it seems like it might be unavoidable.
Screen Shot 2021-08-24 at 4 02 05 PM

backstop/tests/toast-notification.html
mayank99: On the contrary, I would just change the text to lower-kebab-case, i.e. `top-start`, `top-end`, etc. It better matches the class names and the prop values in react too.

backstop/tests/toast-notification.html
mayank99: I think there are still a couple left over images. Can you verify? Run `yarn test` followed by `yarn clean:images`

mayank99:
> Do you have in mind appear animation or sliding animations when new toasts appear?

This is the current default animation. Notice how the toasts below the most recent one are "pushed" down.
toast-animation

You'll have to flip this animation for all classnames that start with `iui-bottom-`.

src/toast-notification/classes.scss
```diff
@@ -60,13 +55,50 @@
...
       animation: closeAnimation 120ms cubic-bezier(0.4, 0, 1, 1) both;
     }
   }
+
+  &.iui-placement-top {
+    margin-left: auto;
+    margin-right: auto;
+    left: 0;
+    top: 0;
+    right: 0;
```
mayank99, line 72: Do you really need left:0 and right:0 here?

backstop/tests/toast-notification.html
```diff
@@ -30,13 +30,166 @@
...
+        class="iui-radio"
+        id='top-start-option'
+      >
+        <input
+          type="radio"
+          name="radio-demo-6"
+          value="top-start"
+        />
+        <span class="iui-radio-dot">
+          <svg
```
mayank99, line 52: I wonder if it'd be better to just use native radio. The svg stuff is too verbose.

src/toast-notification/classes.scss
```diff
@@ -60,13 +55,50 @@
 
     .iui-toast-entering,
     .iui-toast-entered {
-      animation: openAnimation 240ms cubic-bezier(0.175, 0.885, 0.32, 1.175) both;
+      animation: openAnimation 240ms cubic-bezier(0.175, 0.885, 0.32, 1.175)
+        both;
```
mayank99, line 59: I would revert this. Looks weird with `both` in new line (even if prettier disagrees).

src/toast-notification/classes.scss
mayank99: Same thing in iui-placement-bottom

src/toast-notification/classes.scss
```diff
@@ -60,13 +55,50 @@
...
+    left: 0;
+    right: 0;
+  }
+
+  &.iui-placement-bottom-start {
+    left: 0;
+    bottom: 0;
+  }
+
+  &.iui-placement-bottom-end {
```
mayank99, line 98: Wonder if this is how it should look or if we want to push everything to the edge.

src/toast-notification/classes.scss
gretanausedaite: Without these positions toaster is placed on the left side. We need left:0, right:0.

src/toast-notification/classes.scss
gretanausedaite: Added align-items: flex-start/flex-end; Now it looks much better

backstop/tests/toast-notification.html
gretanausedaite: Simplified radio buttons.

PR #264, opened 2021-08-24
feat(SideNavigation): Add submenu

### LINKED ISSUE #227: SideNav: support sub-menu
SideNav needs to support sub-menus as it seems at least few projects need that for switching to our SideNav.
image

FlyersPh9:
Adding side navigation submenu. Closes #227.
- ~~Activate the submenu by clicking on the Documents button within the side navigation. Click on home button to collapse the submenu.~~ Expansion and animation will be handled using JS/React.
- Submenu is vertically scrollable.
- Submenu can be horizontally resized.
- Submenu header supports a single line truncated title, right aligned action buttons (settings, etc), and a left aligned back button _as needed_.

Screen Shot 2021-09-14 at 2 50 44 PM
Screen Shot 2021-09-14 at 2 51 02 PM


backstop/tests/side-navigation.html
```diff
@@ -67,165 +67,195 @@ <h1>Side navigation</h1>
...
-              style="margin-bottom: 10px;"
-            >Administration</span>
+              <path d="m4.7 0-1.4 1.4 6.6 6.6-6.6 6.6 1.4 1.4 8-8z" />
+            </svg>
+          </button>
+        </div>
+
+        <div class="iui-side-navigation-submenu">
+          <div class="iui-resizer">
+            <div class="iui-resizer-bar"></div>
```
FlyersPh9, line 236: The resizer bar is conflicting with the vertical scroll bar.  I was thinking 1 solution might be to wrap the submenu content and have a 8px left/right margin so that the scrollbar appears 8px indented from the edge of the submenu.
conflict

FlyersPh9 (commented):
In the standard if a submenu is expanded we have the right border of the tab disappear.  However, we changed the visual appearance of the active tab to show a light blue background color.  We can't have the submenu with a light blue background color, that would look weird.  Maybe we just ignore hiding the right border?  Maybe we cover up the right border with a light blue to white gradient? I am unsure and open to discussion.

Screen Shot 2021-08-24 at 11 14 16 AM

backstop/tests/side-navigation.html
bentleyvk: Is it mandatory to have resizer now?

mayank99: Maybe we could deprioritize resizer and leave it as a CSS implementation using [`resize: horizontal`](https://developer.mozilla.org/docs/Web/CSS/resize)?

image 

The resize handle shows below the scrollbar so at least that's nice?
image

bentleyvk:
> In the standard if a submenu is expanded we have the right border of the tab disappear. However, we changed the visual appearance of the active tab to show a light blue background color. We can't have the submenu with a light blue background color, that would look weird. Maybe we just ignore hiding the right border? Maybe we cover up the right border with a light blue to white gradient? I am unsure and open to discussion.
> 
> Screen Shot 2021-08-24 at 11 14 16 AM


I tried hiding border. I kind of looks odd to me:
image

src/side-navigation/side-navigation.scss
```diff
@@ -2,9 +2,10 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 @import '../icon/index';
+@import '../text/index';
 
 @mixin iui-side-navigation {
-  display: inline-flex;
+  display: flex;
```
bentleyvk, line 8: Why changed?

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,41 @@
...
+  height: 100%;
+  min-width: $iui-3xl * 2;
+  width: $iui-3xl * 4;
+  max-width: 50vw;
+  box-sizing: border-box;
+  padding: $iui-baseline $iui-sm;
+  overflow-x: hidden;
+  overflow-y: overlay;
+  resize: horizontal;
+  @media (prefers-reduced-motion: no-preference) {
```
bentleyvk, line 188: This animation looks odd.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,41 @@
     }
   }
 }
+
+@mixin iui-side-navigation-submenu {
+  display: flex;
+  flex: 0 0 auto;
```
bentleyvk, line 177: Is it needed?

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,41 @@
...
+  .iui-side-navigation-title {
+    @include iui-text(title);
+    margin-bottom: $iui-baseline;
+    user-select: none;
+  }
+}
+
+@mixin iui-side-navigation-wrapper {
+  display: flex;
+  position: relative;
```
bentleyvk, line 205: Not needed

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,41 @@
...
+  resize: horizontal;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: max-width $iui-speed-fast ease-out;
+  }
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-right: 1px solid t(iui-color-background-5);
+  }
+
+  .iui-side-navigation-title {
```
bentleyvk, line 196: I am not sure about this class and whether we need to hide it on expand. I thought users will use our typography `Title` component. Also, here is even an example with more than just a title:
image

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,41 @@
     }
   }
 }
+
+@mixin iui-side-navigation-submenu {
```
bentleyvk, line 175: How it should be minimized/maximized? In UX site an expander is above everything (sidenav + submenu): 
image

src/side-navigation/classes.scss
```diff
@@ -5,3 +5,11 @@
 .iui-side-navigation {
   @include iui-side-navigation;
 }
+
+.iui-side-navigation-submenu {
```
veekeys, line 9: forgot export in barrel index and classes

src/side-navigation/side-navigation.scss
veekeys: When it even works? We plan to change `max-width`?

src/side-navigation/side-navigation.scss
veekeys: With current html demo, there is no need of flex at all in here...

src/side-navigation/side-navigation.scss
bentleyvk: When you resize the whole window it has a delay.

veekeys: oh wow :O that is interesting

src/side-navigation/side-navigation.scss
veekeys: I guess we can have it.. if any absolute elements appear in children, they could be relative to the side nav...?

src/side-navigation/side-navigation.scss
FlyersPh9: I was having issues placing the sub-menu correctly, but I must have fixed that elsewhere and this change was unneeded.  Reverting.

src/side-navigation/side-navigation.scss
FlyersPh9: Removed the flex parts.

src/side-navigation/side-navigation.scss
FlyersPh9: I've removed it although my demo does not currently test / demonstrate the sub-menu being activated where this transition on max-width _might_ be necessary.

src/side-navigation/side-navigation.scss
bentleyvk: Forgot to push? 😁

FlyersPh9: Yeah, sorry, having a rough start back from moving and forgetting things. 😅 I'm going to stop commenting "fixed in the next push" and only comment after it is pushed.

src/side-navigation/side-navigation.scss
FlyersPh9: I had included the transition on width to handle the submenu being activated when not present but it seems to be negatively effecting the resizing.  I've removed that and now included a region of CSS that copies the animation in/out classes we used on expandable block but I have not tested this yet.  Mayank had a cool way of testing this by linking the CSS to Storybook but I don't remember how to do that.

src/side-navigation/side-navigation.scss
FlyersPh9: I'd like to remove the title because it is repeated in the expanded side navigation and can free up some space.  However, it makes it very challenging when figuring out where to place the settings button.  Because of this I have removed CSS that hides the submenu title when the side navigation is expanded.

src/side-navigation/side-navigation.scss
FlyersPh9: That's a good point, I kind of overlooked that. 🤔 I'm kind of liking the min/max button only expanding the width of the side navigation because it's not really expanding/collapsing anything within the submenu.  I'm going to pass this idea around the UX team and see if there are any thoughts.

backstop/scenarios/side-navigation.js
```diff
@@ -9,18 +9,18 @@ module.exports = [
...
     actions: [hover('#test-tab-2'), focus('#test-tab-1')],
     selectors: ['#demo-side-navigation'],
   }),
   scenario('State expanded hover and focus', {
-    actions: [
-      click('#test-expand'),
-      hover('#test-tab-2'),
-      focus('#test-tab-1'),
-    ],
+    actions: [click('#test-expand'), hover('#test-tab-2'), focus('#test-tab-1')],
```
bentleyvk, line 23: Is focus visible on dark theme? Because I can't see it.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,93 @@
...
+  padding: 0 $iui-sm;
+  overflow-x: hidden;
+  overflow-y: overlay;
+  resize: horizontal;
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-right: 1px solid t(iui-color-background-5);
+  }
+
+  // #region Animating expanding / collapsing of the submenu
```
bentleyvk, line 187: Better use `@mixin iui-transition-group` and leave animation to React. You only need to add `width` to mixin `transition` or override it.

src/side-navigation/side-navigation.scss
FlyersPh9: I spoke with a couple of UXers about the placement of the expander button.  They are ok with restricting it to only the side navigation.  In fact, some of their designs in Axure already have it this way:
Screen Shot 2021-09-15 at 11 52 24 AM

backstop/scenarios/side-navigation.js
FlyersPh9: Hmm… I am noticing that now too in both light and dark theme.  The non-expanded side navigation tests are working correctly tho.  Is it possible the `click` action is being applied last so it is stealing focus away from the calendar button?

src/side-navigation/side-navigation.scss
FlyersPh9: Ok, I think I understand what you are suggesting.  I've made the change in `mixins.scss` and `side-navigation.scss`.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+  padding: 0 $iui-sm;
+  overflow-x: hidden;
+  overflow-y: overlay;
+  resize: horizontal;
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-right: 1px solid t(iui-color-background-5);
+  }
+
+  &.iui-side-navigation-submenu-visible {
```
bentleyvk, line 191: I think this class is not needed and some rules can be moved to the submenu class and others moved to HTML.

backstop/tests/side-navigation.html
```diff
@@ -95,131 +103,261 @@ <h1>Side navigation</h1>
     <hr />
 
     <section id="demo-side-navigation">
-      <div
```
bentleyvk, line 98: Why you moved expander to the bottom?

backstop/scenarios/side-navigation.js
```diff
@@ -9,18 +9,18 @@ module.exports = [
     actions: [click('#test-expand')],
     selectors: ['#demo-side-navigation'],
   }),
+  scenario('Type submenu', {
+    actions: [click('#myDocumentsButton'), hover('#demo-take-hover')],
```
bentleyvk, line 13: What hovering `demo-take-hover` does? Because it is a simple `span`.

backstop/tests/side-navigation.html
```diff
@@ -95,131 +103,261 @@ <h1>Side navigation</h1>
...
+        >
+          <div class="iui-submenu-header">
+            <div class="iui-submenu-header-label">
+              <button class="iui-button iui-borderless">
+                <svg-chevron-left
+                  class="iui-icon"
+                  aria-hidden="true"
+                ></svg-chevron-left>
+              </button>
+              <span id="demo-take-hover">Documents</span>
```
bentleyvk, line 254: Can we have here a longer string? Just to make sure that ellipsis is working.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+    align-items: center;
+    justify-content: space-between;
+
+    .iui-submenu-header-label {
+      @include iui-text(title);
+      display: flex;
+      align-items: center;
+      overflow: hidden;
+
+      button {
```
bentleyvk, line 214: I am not sure about this one. Maybe move to HTML. What do you think @veekeys

veekeys: is it even needed? it is for the button which gonna be child of the `iui-header-buttons`, which already has flex-shrink... 
btw, need to rename those classes as in info panel.

bentleyvk: No no, it is for the buttons inside label:
image

backstop/scenarios/side-navigation.js
bentleyvk: Yeah, I see that there is no focus. I am looking into that.

bentleyvk: Made a fix https://github.com/iTwin/iTwinUI/pull/295

backstop/scenarios/side-navigation.js
FlyersPh9: `demo-take-hover` steals the hover effect after the click action takes place.  Otherwise the test image has a tooltip displaying over top of the submenu.

bentleyvk: In that case you can just hover/focus `body`, then no need to use additional IDs.

FlyersPh9: Ahh yes, that is a much better idea.  Change made!

src/side-navigation/side-navigation.scss
veekeys: ohh... damn.. and it is shrinking while resizing...?

bentleyvk: Without that rule yes

backstop/tests/side-navigation.html
FlyersPh9: I placed the expander on the bottom because the header content of the submenu (back button, title, action buttons) are vertically aligned with the first button:
Screen Shot 2021-09-21 at 9 21 49 AM

When the expander is at the top, the alignment doesn't look as nice:
Screen Shot 2021-09-21 at 9 23 22 AM

backstop/tests/side-navigation.html
FlyersPh9: Yes sir.
Screen Shot 2021-09-21 at 9 25 04 AM

src/side-navigation/side-navigation.scss
mayank99: I was thinking the same thing

mayank99: I think it would be better to toggle the HTML itself (i.e. remove from DOM when collapsed) than toggle visibility through CSS.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+@mixin iui-side-navigation-submenu {
+  display: none;
+  visibility: hidden;
+  opacity: 0;
+  width: 0;
+  height: 100%;
+  box-sizing: border-box;
+  padding: 0 $iui-sm;
+  overflow-x: hidden;
+  overflow-y: overlay;
```
mayank99, line 184: Add `overflow-y: auto` before this rule to support all browsers.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+
+      button {
+        flex-shrink: 0;
+      }
+
+      > * {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        user-select: none;
```
mayank99, line 222: Are you sure about disabling text selection here?

backstop/scenarios/side-navigation.js
mayank99: Any chance that PR might have caused this change? Looks like old one had hover effect leftover in expander.
image

This is the only change I can see in all test images currently.

bentleyvk: I think so. Previously `click` would leave hover on the clicked element but sometimes it wouldn't leave hover effect.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+
+  @include iui-transition-group;
+
+  .iui-submenu-header {
+    height: $iui-baseline * 5;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+
+    .iui-submenu-header-label {
```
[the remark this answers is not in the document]
FlyersPh9: Oh wow, this is a game changer for me.  I do really like the organizational aspect of nesting.  I found myself often nesting on my first implementation & then going back to un-nest everything to reduce specificity.  I will definitely use this now & in future PRs.

src/side-navigation/side-navigation.scss
FlyersPh9: Yeah, this one might be too aggressive, I'll remove it.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,71 @@
...
+      }
+
+      > * {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        user-select: none;
+      }
+
+      .iui-submenu-header-actions {
```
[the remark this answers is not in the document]
FlyersPh9: This class has the same CSS similar to the `.iui-information-header-actions` class within [information-panel.scss](https://github.com/iTwin/iTwinUI/blob/f1730b28bcdf175dcfeee8798d1535de2d549f2d/src/information-panel/information-panel.scss#L87).  If we can avoid it here, we should probably make the same change to info panel.

src/side-navigation/side-navigation.scss
FlyersPh9: Done.

mayank99: I meant like this:
```scss
overflow-y: auto;
overflow-y: overlay;
```

This will make sure that `overlay` is used when supported, and `auto` is used as fallback.

FlyersPh9: Ah, ok, I misunderstood.  Implemented like you suggested now.

src/side-navigation/side-navigation.scss
FlyersPh9: Ok, I've moved this out of the CSS and into the HTML.  This class handles making it visible (display, visibility, & opacity) but also included some width, min, & max.  Will those also be handled in React?

src/side-navigation/side-navigation.scss
veekeys: It can be just simple image in there too, I assume..?Or this is always gonna be there a back button nothing else? Then we will need to hardcode that in react.
Then we could just have a separate class for the back-button or something..?

FlyersPh9: `button` has been replaced with `.iui-button`.

src/side-navigation/side-navigation.scss
mayank99: The width stuff can moved into the general submenu mixin. And the visibility part can be handled in React using transition-group classes.

mayank99: Oh you moved the CSS verbatim, and I see we also have some more rules
image

Those will all need to be removed. And instead, we'll need to add some JS for show/hide (similar to `expandable-block.html`). The animation will not work in the demo html, but that's ok.

The width/min/max stuff can moved into the general submenu mixin. And the enter/exit animation can be handled in React using transition-group classes.

FlyersPh9: The submenu demo has the submenu stuck in the open state, so none of the display, visibility, opacity are present anymore.

mayank99 (approved):

src/side-navigation/side-navigation.scss
FlyersPh9: @bentleyvk @veekeys If this discussion is resolved, feel free to merge this PR.  @mayank99 has approved the PR but wanted to make sure this discussion was resolved prior to merging.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,61 @@
...
+  overflow-y: overlay;
+  resize: horizontal;
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-right: 1px solid t(iui-color-background-5);
+  }
+
+  @include iui-transition-group;
+
+  &-header {
```
bentleyvk, line 193: I am not sure about using this kind of naming. It is a nice feature but it is really hard when you want to find specific class.

src/side-navigation/side-navigation.scss
```diff
@@ -170,3 +171,61 @@
...
+      }
+
+      &-actions {
+        flex-shrink: 0;
+        margin-left: $iui-s;
+      }
+    }
+  }
+
+  &-body {
```
bentleyvk, line 222: Is this class really needed? Maybe just add `padding` to all directions in `iui-side-navigation-submenu`.

src/side-navigation/side-navigation.scss
bentleyvk: Fine by me.

src/side-navigation/side-navigation.scss
mayank99: It's not really hard because the component name is in front of it, so you know where to look.

bentleyvk: But you can't just copy class name from HTML and search. You will need to first thought why no results are found, then open the SCSS file and search manually line by line.

mayank99: You could just open the correct file and search for the suffix.

I think it's a price worth paying to reduce specificity, which is more important.

src/side-navigation/side-navigation.scss
FlyersPh9: If I add `padding-top` to `.iui-side-navigation-submenu` I am unable to align the `.iui-side-navigation-submenu-header` with the first button in the side navigation.

I've removed the `.iui-side-navigation-submenu-body` class and instead added a bottom padding to `.iui-side-navigation-submenu`.

Only thing we should be cautious of is if a user does _not_ use the `.iui-side-navigation-submenu-header` class we will not have any top padding present.  This isn't standard, so hopefully shouldn't be an issue.

mayank99: Do you still need the extra div?
image

bentleyvk: Seems like not needed, removed.

PR #267, opened 2021-08-24
refactor: Use icon web components

FlyersPh9:
Implementing native icon web components everywhere an SVG is used.


FlyersPh9:
Icon web components are working without any SCSS changes in the demo pages updated in the initial push.  Other demo pages that were skipped either do not work without SCSS changes or I did not try it yet because there are so many dang SVGs (table).

PR #268, opened 2021-08-25
fix(Table): Handling very long strings without spaces

bentleyvk:
Before:
image

After:
image


mayank99:
Could you revert the changes in html and test images? Looks terrible

bentleyvk:
> Could you revert the changes in html and test images? Looks terrible

Nope 😀 It is a test file so we need to have such edge cases. It is not the first time we failed to handle very long strings.

mayank99 (approved):

PR #270, opened 2021-08-25
refactor(ExpandableBlock): Use flex for better positioning

FlyersPh9:
Refactored Expandable block to do the following:
- Use flex for better positioning of icons & labels (this was needed to get icon web components to display correctly).
- Added support for optional status icon, `.iui-status-icon`.
- `.iui-title` and `.iui-caption` are now wrapped with `.iui-expandable-block-label` to enable truncation within a flex.
- Added `<section>`s to get better test results, copied from #155.

Screen Shot 2021-08-25 at 1 14 11 PM


mayank99 (approved):
Don't like the discrepancy between new and old class names, but this is not a full refactor so I guess it's ok

PR #276, opened 2021-08-27
fix(InputGroup): Remove unused styles and align labels

mayank99:
This PR is the first step towards input refactor. It does 3 things:
- Remove random unused styles.
- Remove `:enabled` from `:hover` selector by moving it above `[disabled]`.
- Align labels for required and non-required inputs.

*The last point changes all test images, but the visuals were approved by @FlyersPh9* so I wanted to create this PR before I do the bigger refactor, to make reviews easier.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,5 +3,5 @@ end_of_line = lf
 insert_final_newline = true
 indent_size = 2
 indent_style = space
-max_line_length = 80
+max_line_length = 120
 quote_type = single
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -12,6 +12,7 @@
       "border-bottom": ["/^0/"],
       "border-left": ["/^0/"]
     },
+    "no-eol-whitespace": true,
     "function-parentheses-space-inside": "never-single-line",
     "max-nesting-depth": 6,
     "order/order": null,
--- a/backstop/tests/checkbox.html
+++ b/backstop/tests/checkbox.html
@@ -1159,8 +1159,7 @@
           width: fit-content;
           display: flex;
           flex-direction: column;
-          align-items: flex-start;
-        "
+          align-items: flex-start;"
       >
         <label class="iui-checkbox">
           <input
@@ -1388,8 +1387,7 @@
           width: fit-content;
           display: flex;
           flex-direction: column;
-          align-items: flex-end;
-        "
+          align-items: flex-end;"
       >
         <label class="iui-checkbox">
           <input
--- a/backstop/tests/file-upload.html
+++ b/backstop/tests/file-upload.html
@@ -31,7 +31,7 @@
 
       .demo-right-align {
         justify-content: flex-end;
-      } 
+      }
     </style>
 
     <script
--- a/backstop/tests/footer.html
+++ b/backstop/tests/footer.html
@@ -27,7 +27,7 @@
       body { padding: 0; }
       h1 { padding: 24px 24px 0; }
 
-      #demo-content { 
+      #demo-content {
         height: 1200px;
         background-color: var(--iui-color-background-5);
       }
--- a/backstop/tests/progress-indicator.html
+++ b/backstop/tests/progress-indicator.html
@@ -31,7 +31,7 @@
       section .iui-progress-indicator-radial {
         margin-right: 16px;
       }
-      
+
       #demo-overlay-content {
         padding: 11px 12px;
         border: 1px solid var(--iui-color-background-5);
--- a/backstop/tests/select.html
+++ b/backstop/tests/select.html
@@ -33,7 +33,7 @@
       #demo-default-opened {
         height: 420px;
       }
-      
+
       #demo-default-singular-opened {
         height: 210px;
       }
--- a/backstop/tests/toggle-switch.html
+++ b/backstop/tests/toggle-switch.html
@@ -154,8 +154,7 @@
           width: fit-content;
           display: flex;
           flex-direction: column;
-          align-items: flex-start;
-        "
+          align-items: flex-start;"
       >
         <label class="iui-toggle-switch">
           <input
@@ -204,8 +203,7 @@
           width: fit-content;
           display: flex;
           flex-direction: column;
-          align-items: flex-end;
-        "
+          align-items: flex-end;"
       >
         <label class="iui-toggle-switch">
           <input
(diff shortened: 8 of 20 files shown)


src/inputs/labeled-inputs.scss
```diff
@@ -4,45 +4,36 @@
...
+    }
+  }
 
-      &.iui-disabled {
-        cursor: not-allowed;
-      }
+  &.iui-disabled {
+    &,
+    label {
+      cursor: not-allowed;
```
veekeys, line 25: I think this one introduced regression...
Now this inline container gets `not-allowed` cursor on whole line (versus only on checkboxes and their labels).
image

src/inputs/labeled-inputs.scss
```diff
@@ -4,45 +4,36 @@
...
-    line-height: $iui-line-height;
-    text-align: left;
-    @include themed {
-      color: t(iui-text-color);
-    }
 
     &.iui-required {
       &::after {
         content: ' *';
+        margin-right: 0;
```
veekeys, line 36: Why this was needed?

src/inputs/labeled-inputs.scss
```diff
@@ -87,17 +78,17 @@
...
-  }
-
   &.iui-inline {
     flex-direction: row;
     align-items: center;
 
     > .iui-label {
       flex-shrink: 0;
-      margin: 0 $iui-sm 0 0;
+      margin: 0 ($iui-m + 2) 0 0;
```
veekeys, line 87: It looks more centered with margin bottom 3px than not, to be honest :D

These +2 come from...?

src/inputs/labeled-inputs.scss
```diff
@@ -4,45 +4,36 @@
...
+      cursor: not-allowed;
     }
   }
 
   > .iui-label {
-    padding: 0;
     margin-bottom: $iui-component-offset;
-    border: none;
-    display: block;
-    width: fit-content;
```
veekeys, line 34: Sure this is not needed? NOt label takes full width.. or that was bringin some issues somewhere?

src/inputs/labeled-inputs.scss
veekeys: Is that the least margin we could have? Required is not that often compared how many places it influenced, and in general, I think endresult is too wide (looking at filters and other places, where there is no required...)

src/inputs/labeled-inputs.scss
mayank99: Is that a regression or a bug fix?

src/inputs/labeled-inputs.scss
mayank99: This counteracts the extra margin set on non-required label

mayank99: Oh actually this is leftover, I think. Will remove

src/inputs/labeled-inputs.scss
mayank99: what does margin-bottom have to do with anything?

the `+ 2` is just the value I found which aligned this label with the required one.

I could try making it even tighter, but the visuals were approved by Jon before he left.

src/inputs/labeled-inputs.scss
mayank99: The whole component takes full width anyway, how does it change anything?

src/inputs/labeled-inputs.scss
veekeys: regression

src/inputs/labeled-inputs.scss
veekeys: Margin-bottom is overriden with this rule, so just commented, what I was playnig around. Not sure, if that was influenced by your changes.
I just checked the images, thought that we expanded in width too much, compared, the content did not really change. Especially, when we talk a lot about condensed and stuff..
But I guess we can leave it, if thats ok for everyone.

Then +2 look kinda random.

src/inputs/labeled-inputs.scss
veekeys: I assumed, that rule was there for a reason. So now I am just commenting the removal of it.
We never look at the parent width, because at the end, everyones parent is body which takes it full.
It can have influence for random elements in that label, but idk...

src/inputs/labeled-inputs.scss
mayank99: The margin-bottom was not influenced by my changes, we always override it with 0 for inline label.

mayank99: Reduced overall margin-right by 2px, so the `+ 2` is gone now.

The value is now 6px on required, and 16px on non-required.
image
image

I also tried changing the ::after `content` from `' *'` to `'*'` and adding margin-left: 2px, but I can't get the checkboxes to align so I didn't push.
image
image

src/inputs/labeled-inputs.scss
mayank99: I disagree. It is now consistent with how our other labeled inputs behave: https://itwin.github.io/iTwinUI/backstop/minified/text-field.html

The style was previously coming from this line (all I did is remove `>`).
https://github.com/iTwin/iTwinUI/blob/21c5d886609d3c4cf8648ff909558d2f121fdc6d/src/inputs/labeled-inputs.scss#L96-L97

It also fixes the weird bug where moving your cursor in a disabled input group inside fieldset causes cursor to switch between `not-allowed` and default.
fieldset

If there is a need to allow default cursor on disabled input group's label, then it could still be achieved by only setting `disabled` in the elements, not the group.

veekeys: It is different between situations. Input with Label works well (when you click on label, input is focused, so cursor should be on both of them).
Now here label is for checkboxes/radios group. Those have their own labels, so interaction is happening there, as a result disabled status should be there too. Disabled on the label for the group of those mean what?
I would have assumed we change cursor to show the interaction, not just to change it.

src/inputs/labeled-inputs.scss
veekeys: still not removed

mayank99: ? 
image

veekeys: amm.. ok. I was schecking, and it was still there.. Now i see its not.. weird.

veekeys (approved):

PR #277, opened 2021-08-30
feat(Table): Editable cell styling

bentleyvk:
- I chose `outline` as it is the default styling also it does not affect sizing of cell. Also, with borders there will be much more work as I need to add invisible border to avoid content shift.
 - Focus cell has double border but showed it to Jon and he said it's ok.
   - I tried to fix that by moving all border logic to cells instead of row but then small spaces are visible:
  image

For some reason longer text is a bit more longer after some point with `contenteditable` and I have no idea why:
image


src/table/table.scss
```diff
@@ -332,6 +332,23 @@
   &.iui-negative {
     @include iui-table-cell-status($status: negative);
   }
+
+  &[contenteditable] {
+    outline-offset: -1px;
```
bentleyvk, line 337: Without this, outline goes outside cell. Vertically it is ok as it merges with the borders but horizontally it goes outside the bound.

mayank99 (approved):

PR #280, opened 2021-08-30
style: Increase line length to 120 and fix whitespace

mayank99:
1. Because CSS values often get very long, it makes sense to have a longer line length. 80 is too limiting.
*This change only affects manual prettier formatting. Simply saving the file uses stylelint, not prettier, so it's unaffected by this change.*

2. Added stylelint rule [no-eol-whitespace](https://stylelint.io/user-guide/rules/list/no-eol-whitespace/) to make sure there are no trailing white spaces when a file is saved. *Fixed all violations reported by this rule.*


backstop/tests/checkbox.html
```diff
@@ -1159,8 +1159,7 @@ <h3>Checkbox on left</h3>
           width: fit-content;
           display: flex;
           flex-direction: column;
-          align-items: flex-start;
-        "
+          align-items: flex-start;"
```
veekeys, line 1162: double quotes in new line looked better in this situation..
If you put it all in new line and then press save, will be same formatting?

mayank99: When I ran lint, it removed leading whitespace like this:

```diff
-         "
+ "
```

veekeys (approved):

PR #282, opened 2021-09-01
feat(Table): Added transition rules to row

veekeys:
Moved out transition rules from expanded content to row, so we could animate subrows expansion too.

subrows3

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -142,6 +142,8 @@
       border-bottom-color: t(iui-color-background-4);
     }
 
+    @include iui-transition-group;
+
     > .iui-slot > .iui-more-options {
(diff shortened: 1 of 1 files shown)


src/table/table.scss
```diff
@@ -191,9 +191,10 @@
       }
     }
 
+    @include iui-transition-group;
```
bentleyvk, line 194: Maybe move to the top that it would be easier to see.

veekeys: moved somewhere

mayank99 (approved):

bentleyvk (approved):

veekeys:
Buggy, maybe next time...

PR #287, opened 2021-09-02
feat(labeled-inputs): Add new layout options using grid

mayank99:
Refactored the entire input-container logic.
- Flattened html structure:
  - icon is no longer inside message div, and has a class: `iui-input-icon`.
- Used css grid for all layouts.
  - Grid allows using same markup for different layouts.
  - Replaced `position: absolute` by using same grid-area for overlap.
  - New code still produces pixel-identical layouts to old code.
  - Only changes in test images are from new additions for hybrid layouts.
  - Enabled IE grid in autoprefixer.
- Supported layouts using modifiers on `iui-input-container` (can be combined to form hybrid layouts): 
  - `iui-inline-label`
  - `iui-inline-icon`
  - `iui-with-message` 
- Radio, checkbox, slider, etc need to be wrapped in `iui-input-group`.

#### Screenshots

image
image


src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +40,163 @@
...
-  // #region Statuses
-  &.iui-positive {
-    @include iui-inputs-status(
-      iui-color-foreground-positive,
-      iui-color-foreground-positive-overlay,
-      iui-color-foreground-positive-rgb
-    );
+/// Text label for an input.
+/// Supports .iui-required modifier to show red asterisk.
+@mixin iui-input-label($displayStyle: default) {
```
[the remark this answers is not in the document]
mayank99: Good catch, this is leftover. Will remove

saaaaaally (approved):

postcss.config.js
```diff
@@ -0,0 +1,7 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+module.exports = {
+  plugins: [require('autoprefixer')({ grid: 'autoplace' }), require('postcss-discard-comments')],
```
veekeys, line 6: Getting this on build:
image

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
+  }
+}
 
-        svg:not(.iui-icon) {
-          margin-right: $iui-sm;
-        }
-      }
+/// Message shown below input
+@mixin iui-input-message {
+  font-size: $iui-font-size-small;
```
veekeys, line 124: `line-height` needs to be changed.
image

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +40,163 @@
...
+  justify-self: end;
+  margin-right: $iui-sm;
+  position: relative;
+
+  &.iui-button {
+    height: 100%;
+    width: fit-content;
+    margin-right: 0;
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
```
veekeys, line 169: Why this was needed? Could not find the change in Chrome if removed

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +40,163 @@
...
+  -ms-grid-column: 2;
+  -ms-grid-column-align: end;
+  // stylelint-enable
+
+  grid-area: inputs;
+  justify-self: end;
+  margin-right: $iui-sm;
+  position: relative;
+
+  &.iui-button {
```
veekeys, line 164: Would be nice to have demo for this one, to verify

veekeys: a, ok. There is one.

veekeys: Need to handle other sizes too, I guess...
image

If applying iui-small class on button too, need to arrange spacing..
image

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +40,163 @@
...
+  a {
+    user-select: none;
   }
-  // #endregion
 }
 
-@mixin iui-inputs-status($foregroundColor, $foregroundOverlay, $foregroundRgb) {
+/// Icon to be shown near an input.
+@mixin iui-input-icon {
+  display: flex;
```
veekeys, line 137: in which scenario anchor is used? we do not really have demo for that, do we?

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
-    @include iui-inputs-status(
-      iui-color-foreground-positive,
-      iui-color-foreground-positive-overlay,
-      iui-color-foreground-positive-rgb
-    );
+/// Icon to be shown near an input.
+@mixin iui-input-icon {
+  display: flex;
+  grid-area: icon;
+  width: $iui-icons-default;
```
veekeys, line 139: should be able to reuse icons mixin

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
   }
-  // #endregion
 }
 
-@mixin iui-inputs-status($foregroundColor, $foregroundOverlay, $foregroundRgb) {
+/// Mixin for applying status colors to input container.
+/// @arg status - must be one of: positive, negative, warning
+/// @arg iconSelector - selector to apply status fill on. Defaults to .iui-input-icon
+/// @arg textSelector - selector to apply text color on. Defaults to .iui-message
+@mixin iui-input-status($status, $iconSelector: '.iui-input-icon', $textSelector: '.iui-message') {
```
veekeys, line 177: for now, I dont see this being reused, not sure we need `iconSelector` and `textSelector` args.

src/inputs/labeled-inputs.scss
```diff
@@ -212,13 +225,13 @@
     }
```
veekeys, line 225: need to check these paddings.. and selector. Maybe we dont need to duplicate native textarea and input? would be enough our classes..

backstop/tests/text-field.html
```diff
@@ -146,7 +159,7 @@ <h2>Default</h2>
 
     <h2>Inline</h2>
     <section id="demo-inline">
-      <label class="iui-input-container iui-inline">
+      <label class="iui-input-container iui-with-message iui-inline-label">
```
veekeys, line 162: I assume `iui-with-message` is not needed here, cause we do not display any message...

backstop/tests/text-field.html
```diff
@@ -238,81 +302,84 @@ <h2>Inline</h2>
 
     <h2>Without label</h2>
     <section id="demo-no-label">
-      <label class="iui-input-container iui-inline">
+      <label class="iui-input-container iui-with-message iui-inline-label">
```
veekeys, line 305: and here no need to have `iui-with-message` and `iui-inline-label`..

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
+  grid-area: message;
+  @include themed {
+    color: t(iui-text-color-muted);
+  }
 
-      &:last-child {
-        padding-right: $iui-sm;
-      }
-    }
+  a {
```
bentleyvk, line 130: Where it is used?

src/inputs/labeled-inputs.scss
```diff
@@ -3,16 +3,33 @@
...
-  text-align: left;
-  font-weight: $iui-font-weight-normal;
+  display: grid;
+  grid-template:
+    'label label'
+    'inputs inputs' / auto 1fr;
+  grid-row-gap: $iui-component-offset;
+
+  &.iui-with-message {
+    grid-template:
```
bentleyvk, line 17: Might be made as a default template.

bentleyvk: Ok, tests are failing with this as a default template 😀

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
-
-    &.iui-required {
-      &::after {
-        content: ' *';
-        @include themed {
-          color: t(iui-color-foreground-negative);
-        }
-      }
-    }
+  // Appropriate spacing even if no text label is given
```
bentleyvk, line 50: When these rules are needed? I removed them and nothing changed, tests are also passing.

src/inputs/labeled-inputs.scss
```diff
@@ -28,169 +45,165 @@
...
+      display: flex;
+      gap: $iui-m;
 
-      svg {
-        @include themed {
-          fill: t(iui-icons-color-actionable);
+      > .iui-checkbox,
+      > .iui-radio,
+      > .iui-toggle-switch {
+        margin-right: $iui-m;
```
bentleyvk, line 73: Shouldn't this be for non-last children?

src/inputs/labeled-inputs.scss
veekeys: or ok.. lets do that after icons refactor..

src/inputs/labeled-inputs.scss
veekeys: Selector changed

src/inputs/labeled-inputs.scss
veekeys: @FlyersPh9 I see we use same line-height for smaller font-size (12px). @bentleyvk said it looks better like this, but I tohught it should be a little less. As in the screenshot you can see, the spacing below input and above input looks different, because font size of message is smaller.

src/inputs/labeled-inputs.scss
```diff
@@ -28,177 +66,214 @@
...
     }
 
-    > .iui-checkbox:not(:last-child),
-    > .iui-radio:not(:last-child),
-    > .iui-toggle-switch:not(:last-child) {
-      margin-right: $iui-m;
-    }
+    > .iui-input-group {
+      display: flex;
+      gap: $iui-m;
```
FlyersPh9, line 128: Are we ok with using `gap`?

FlyersPh9: Ah wait, I see `@supports (gap: $iui-m)` a couple lines down which I'm assuming is the IE solution.

FlyersPh9: Oh wait, I think I see what is happening down below:
1. It tries to apply gap.
2. If a browser doesn't support gap, it also applies margin-right to try to replicate gap.
3. If a browser _does_ support gap, using `@supports (gap)` we then remove the margin-right & only rely on gap.
Clever.

src/inputs/labeled-inputs.scss
FlyersPh9: I think you are correct.  I commented out the `gap` CSS to see what it would look like if only relying on `margin-right`. Notice the extra space between the last checkbox & status icon.

As is:
Screen Shot 2021-09-14 at 4 00 56 PM

After removing `gap`:
Screen Shot 2021-09-14 at 4 01 27 PM

bentleyvk: I see there is some issues with IE so let's leave it as it is.

src/inputs/labeled-inputs.scss
```diff
@@ -3,16 +3,54 @@
...
-  position: relative;
-  text-align: left;
-  font-weight: $iui-font-weight-normal;
+  display: grid;
+  grid-template:
+    'label label'
+    'inputs inputs' / auto 1fr;
+
+  // stylelint-disable -- fix position in IE
+  -ms-grid-rows: auto auto;
```
bentleyvk, line 16: Seems like it is not needed.

bentleyvk (approved):

PR #285, opened 2021-09-02
test: Stopped animations in tests in order to test animated components

bentleyvk:

PR #288, opened 2021-09-03
ci: Fix index.html getting removed on failed builds

mayank99:
Disabled `Host test results on gh pages` to prevent hosting test results for failed builds on main branch (because build should never fail for legit reasons on main branch).

The problem was happening because `github.event.number` is undefined on main branch and it was putting the test results in the root folder. If we really want to host the test results for main branch, we could change target folder.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Host test results on gh pages
         continue-on-error: true
-        if: failure()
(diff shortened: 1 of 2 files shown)


mayank99:
@bentleyvk The modal test keeps failing randomly in CI and locally too. I think it has to do with #285 because I haven't seen that before.

bentleyvk (approved):

PR #293, opened 2021-09-15
feat(NotificationMarker): Add notification marker

FlyersPh9:
- Add notification marker class that can be applied to pretty much anything but mostly intended for icons within default / borderless buttons.
- Supports all 4 status colors + accessory color in case the notification marker is above a blue / green button.
- Has `.iui-urgent` modifier to include an animation.

Animated and non-animated notification within the side navigation buttons
Animated notification within header notification button


src/button/notification.scss
```diff
@@ -0,0 +1,47 @@
...
+    @media (prefers-reduced-motion: no-preference) {
+      animation: pulse 2s infinite;
+    }
+  }
+
+  @media (prefers-reduced-motion: no-preference) {
+    @keyframes pulse {
+      0% {
+        @include themed {
+          box-shadow: 0 0 0 0 rgba(t(iui-color-foreground-#{$status}-rgb), t(iui-opacity-1));
```
FlyersPh9, line 30: The only issue remaining is no matter what status is applied the `.iui-urgent` class is always displaying in red.  I am unsure why.

bentleyvk: I think the issue is that all keyframes has the same name: "pulse" and one status styling is overriding another. Try renaming to `pulse#{$status}` and using that name in `animation`.

FlyersPh9: Amazing, that fixed it!  Thank you!

backstop/scenarios/header.js
```diff
@@ -4,6 +4,11 @@ module.exports = [
   // Header types
   scenario('Type Default', {
     selectors: ['#demo-default'],
+    hideSelectors: ['.iui-urgent::before'],
```
bentleyvk, line 7: You shouldn't need to hide animated parts anymore as all animations are stopped in tests.

src/button/notification.scss
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-button-notification($status: primary) {
```
bentleyvk, line 5: What if we apply this class not on a button but on an icon?

src/button/notification.scss
```diff
@@ -0,0 +1,58 @@
...
+    @include themed {
+      background-color: t(iui-color-foreground-#{$status});
+      border: 2px solid t(iui-color-background-1);
+    }
+    @media (prefers-reduced-motion: no-preference) {
+      transition: border-color $iui-speed-fast ease-out;
+    }
+  }
+
+  &:enabled:hover::before,
```
bentleyvk, line 23: What it does?

src/button/notification.scss
```diff
@@ -0,0 +1,58 @@
...
+    position: absolute;
+    width: $iui-s;
+    height: $iui-s;
+    border-radius: 100%;
+    @include themed {
+      background-color: t(iui-color-foreground-#{$status});
+      border: 2px solid t(iui-color-background-1);
+    }
+    @media (prefers-reduced-motion: no-preference) {
+      transition: border-color $iui-speed-fast ease-out;
```
bentleyvk, line 19: Isn't transition also needed for background color?

src/button/notification.scss
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-button-notification($status: primary) {
+  position: relative;
+
+  &::before {
+    content: ' ';
```
bentleyvk, line 9: It can also be an empty string.

src/button/classes.scss
```diff
@@ -25,6 +25,22 @@
   &[disabled] {
     @include iui-button-disabled;
   }
+
+  &.iui-notification-primary {
```
saaaaaally, line 29: Not sure but maybe instead of offering only 4 types of notifications maybe it can be only 1 iui-notification but the user can pass in the color they want it to be so that way they aren't limited to only 4 colors. We can also offer a list of the default colors in react (primary, positive, warning, negative) but also let them put in whatever color they want.

src/button/notification.scss
FlyersPh9: Without this, on hover of the button the border of the notification dot does not match the button background color. Screenshot without that CSS: Screen Shot 2021-09-20 at 3 35 05 PM

backstop/scenarios/header.js
FlyersPh9: Ah, you are correct.  Removed.

src/button/notification.scss
FlyersPh9: I guess if we had the dot change from green to red then yes, we would want the background-color to transition as well.  I can add that.

src/button/classes.scss
FlyersPh9: Would it be possible for a user to define a custom color with the way the CSS is setup currently?  They'd need to override `&::before { background-color: t(iui-color-foreground-#{$status}); }` & the `box-shadow` color within the pulse animation.  But I think their override would be tied to a status name (primary, positive, warning, negative), right?

src/button/notification.scss
FlyersPh9: Hmm… that might be possible.  I'll investigate later this week if I have time before I leave (I'm off Thurs-Fri).

src/button/notification.scss
bentleyvk: Hm, so notification marker can only be used on borderless button? Or can we use transparent border?

src/button/classes.scss
bentleyvk: But do we want to offer custom colors? I think this can be achieved using CSS vars like in `Badge`.

src/button/notification.scss
FlyersPh9: Right now the only places that the UX team knows that they want these indicators is the buttons within the side navigation (per service) & the notification bell icon within the header.  I'm sure some other team will find another location where it may be valid, but we aren't sure yet.

bentleyvk: I was thinking maybe we can even have it as a util class and users could add it on any component they like.

FlyersPh9: One component I was thinking I could see a team wanting to use this within is the tabs. So you may be right.

src/button/classes.scss
saaaaaally: If it could be useful to make notifications of any color, I was thinking the notifications would not be tied to status at all. 
Instead of choosing between iui-notification-primary, iui-notification-positive, iui-notification-warning, or iui-notification-negative, we only offer 1 option iui-notification and have a variable passed in --color (green, orange, red, etc.) that sets the color of the notification

bentleyvk:
I saw PW365 has kind of the same notification dot but with number inside it:
image
Is it a valid case and we should consider supporting it too?

src/button/notification.scss
FlyersPh9: Chatted with @mayank99 about this a little bit, we are going to only focus on the situations where we know this will be used.  It may be used in new locations in the future that will require some changes (like placed with a blue / green button).

FlyersPh9:
> I saw PW365 has kind of the same notification dot but with number inside it: image Is it a valid case and we should consider supporting it too?

@VirginieGaudreault raised this same concern with me.  I did some research on other apps (Microsoft Teams, Facebook, Twitter) and all of them do include a number. 😕 So _maybe_ we should?

Concerns if we do add a number:
1. How do we handle very large numbers. 999 won't fit in a circle.
2. The circle will need to be significantly larger.  That will be challenging, especially within the slim header.

I'd almost rather convince the app team that the number is not needed.

src/button/classes.scss
FlyersPh9: @bentleyvk or @saaaaaally could you take a look at the code I pushed just now & suggest how I could use CSS vars like in badge.  The notification dot color does change with the theme, so it might not be as easy as it was in badge because the badge color doesn't change with the theme.

VirginieGaudreault:
> > I saw PW365 has kind of the same notification dot but with number inside it: image Is it a valid case and we should consider supporting it too?
> 
> @VirginieGaudreault raised this same concern with me. I did some research on other apps (Microsoft Teams, Facebook, Twitter) and all of them do include a number. 😕 So _maybe_ we should?
> 
> Concerns if we do add a number:
> 
> 1. How do we handle very large numbers. 999 won't fit in a circle.
> 2. The circle will need to be significantly larger.  That will be challenging, especially within the slim header.
> 
> I'd almost rather convince the app team that the number is not needed.

When apps display a notification badge with a number included, they'll usually not go over 9. If there are 10 or more notifications, it will display as 9+. I have seen occurrences of notification counts on a badge going up to 99+ but it's rare.

If they must display the real quantity of notifications, then the badge's width would adapt to the number. See below

image

Though, a badge this wide isn't very sexy on top of a circle and is better fit for an app icon much bigger than it is.

src/notification-marker/notification-marker.scss
```diff
@@ -56,3 +50,14 @@
     }
   }
 }
+
+@mixin iui-notification-marker-hover {
```
bentleyvk, line 54: Why mixin instead of having stylings inside of the main mixin?

bentleyvk:
Tests are failing.

src/button/classes.scss
bentleyvk: If we really really want to support custom colors, then you do like this:
image
image

But I think we might need to have separate classes for the colors like `iui-primary` and it would set the variable. I guess something like this:
```scss
.iui-primary {
  @include themed {
    --marker-color: t(iui-color-foreground-primary);
  }
}
```

I haven't really used CSS vars so need to read more and play around.

src/notification-marker/notification-marker.scss
FlyersPh9: The main mixin is only used by `/notification-marker/classes.scss` but `@mixin iui-notification-marker-hover` is used within `/button/default.scss`.  Button targets `&:enabled:hover, &:enabled:active, &:focus-visible`, but if we wanted to use the notification marker within a different component, that component may not target `:enabled:hover` for example.

src/button/classes.scss
FlyersPh9: I vote we save custom colors for a future PR upon request.

bentleyvk (approved):

PR #295, opened 2021-09-16
test: Changed click and focus to native JS equivalents

bentleyvk:
Puppeteer click and focus actions didn't work properly when used together. I couldn't find any issues so changed to use native JS equivalents that are much more reliable.
Saw this issue in https://github.com/iTwin/iTwinUI/pull/264#discussion_r709340014


FlyersPh9:
So I would use native JS _**only**_ when focus & click are used in the same scenario, but I can use puppeteer in all other scenarios, correct?

bentleyvk:
> So I would use native JS _**only**_ when focus & click are used in the same scenario, but I can use puppeteer in all other scenarios, correct?

The same actions are used for every scenario, so you don't need to worry when to use what because now click and focus will be used from JS.

FlyersPh9:
> The same actions are used for every scenario, so you don't need to worry when to use what because now click and focus will be used from JS.

Ohh, I see now.  I still write the tests the same, you changed how it works behind the scenes.  Got it, sounds good!

FlyersPh9 (approved):

veekeys (approved):

PR #312, opened 2021-09-29
feat(Menu): Add non-selectable menu item

### LINKED ISSUE #201: Add support for custom content in menu item (non-interactive)
Similar to the column filter menus that we offer, we should offer a standard user profile menu accessed from the header.
Screen Shot 2021-07-19 at 12 27 51 PM

FlyersPh9:
- Add `.iui-menu-content` that allows for non-selectable row.
- Improved menu scenario test to more closely match #155.
- Added some `aria` and `role`, tried to follow [this article](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/) for best practices.

Screen Shot 2021-09-29 at 5 00 18 PM

Closes #201.


src/menu/menu.scss
```diff
@@ -162,6 +162,10 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
   }
 }
 
+@mixin iui-menu-content {
+  padding: $iui-baseline ($iui-component-padding-horizontal + 1px);
```
bentleyvk, line 166: Why +1?

backstop/tests/menu.html
```diff
@@ -320,6 +320,53 @@ <h2>Default</h2>
...
+              >
+                <span class="iui-content">Administrator</span>
+              </div>
+            </div>
+          </label>
+
+        </li>
+        <li
+          class="iui-menu-divider"
+          aria-hidden="true"
```
bentleyvk, line 346: Can you change `aria-hidden="true"` to `role="separator"` in every divider? Because it is the preferable way from accessibility point.

backstop/scenarios/menu.js
```diff
@@ -4,15 +4,18 @@ module.exports = [
   // Menu types
   scenario('Type Default', {
     selectors: ['#demo-default'],
+    viewports: [{ width: 800, height: 600 }],
```
bentleyvk, line 7: Are these newly generated images or or just changed the old ones? Because github is not showing diff button so if they are new ones then need to delete old ones (if still present).

bentleyvk:
Visually looks good! Left a few comments.

src/menu/menu.scss
FlyersPh9: It's the same horizontal padding used within `@mixin iui-menu-item` on [line 44](https://github.com/iTwin/iTwinUI/blob/d2d3e42568500dec2a60a174909867ebacd96924/src/menu/menu.scss#L44).  I think the +1 was used to accommodate for the border?

backstop/tests/menu.html
FlyersPh9: Yes, I can do that, good suggestion.  @mayank99 was suggesting that we also apply some sort of aria on the new `.iui-menu-content` since it's not selectable like the rest of the menu items.  Any suggestions?

mayank99: Best I can think of is `role='presentation'` but I'm not sure

backstop/scenarios/menu.js
mayank99: I think `desktop` is added when there's no viewport size provided. So you'll have to delete those old images

backstop/tests/menu.html
FlyersPh9: I've added `role="presentation"` as well as some other roles and aria attributes found in [this article](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/).

backstop/scenarios/menu.js
FlyersPh9: Ah, ok.  I don't remember that being the case in #155.  I've removed the images that don't have `_desktop` in the file name.

FlyersPh9: Tests are failing now.  I ran the tests again and it wants to add the non `_desktop` images back in.  I am confused what's going on here.

mayank99: Seems you removed the wrong set of images

FlyersPh9: That fixed it.  Let's just pretend this mistake never took place. 😂

mayank99 (approved):

bentleyvk (approved):

PR #313, opened 2021-09-30
feat(labeled-inputs): Add arrow icon with ripple effect

mayank99:
Added new `.iui-actionable` modifier on `iui-input-icon` from input container. Will be used for arrow icon in combobox.
- adds cursor pointer and a 40x34 touch area using padding.
  - needed to reduce height to 90% and add 1px right margin to avoid overlapping with input outline.
- has ripple effect using new `iui-ripple` mixin.
- supports `.iui-open` modifier to do a 180° transform on svg (with transition).

ripple


src/inputs/labeled-inputs.scss
```diff
@@ -229,6 +230,31 @@
     border-top-left-radius: 0;
     border-bottom-left-radius: 0;
   }
+
+  &.iui-actionable {
+    align-items: center;
+    height: 90%;
+    width: fit-content;
```
FlyersPh9, line 237: `width: fit-content` does not work in IE.  I know we are planning to phase out IE support eventually.  Does this have any negative effect on IE or will they just not see the animation?

mayank99: It is surprisingly functional in IE 😮
IE

FlyersPh9:
I wanted to test out the ripple effect so I tested it in `text-field.html`. The SVG becomes vertically off center. This does not seem to effect React because React doesn't use SVG web components.  The fix is to apply `display: flex;` to a class name that is applied to the web component.  The web component currently doesn't have a class.  Since this effect is only used on the combo box, we can ignore this for now.
Screen Shot 2021-09-30 at 5 04 19 PM

src/inputs/labeled-inputs.scss
mayank99: Turns out that's because `width: fit-content` is not even needed. So I've removed it

mayank99:
> The fix is to apply `display: flex;` to a class name that is applied to the web component. The web component currently doesn't have a class.

If you think about it, all our icon webcomponents should always have `inline-flex` or `flex`, so maybe we should just mention it in the README of iTwinUI-icons-elements repo, so that users can do it on their end just to be safe.

And when we do end up using the `iui-actionable` class in this css repo, we can find a better solution.

src/inputs/labeled-inputs.scss
```diff
@@ -229,6 +230,30 @@
...
+
+  &.iui-actionable {
+    align-items: center;
+    height: 90%;
+    margin-right: 1px;
+    padding: 0 $iui-sm;
+    cursor: pointer;
+    border-radius: 50%;
+
+    @include iui-ripple(var(--iui-color-background-1), var(--iui-color-background-2));
```
bentleyvk, line 242: If I understand correctly it currently always does ripple effect from the center not from the click position. Maybe we could create ripple effect like Material has by adding additional ripple element to do the animation:
material-ripple

mayank99: That requires JavaScript, so someday

src/inputs/labeled-inputs.scss
```diff
@@ -229,6 +230,29 @@
     border-top-left-radius: 0;
     border-bottom-left-radius: 0;
   }
+
+  &.iui-actionable {
+    align-items: center;
+    height: calc(100% - 2px);
```
FlyersPh9, line 236: The `-2px` may need to change if a status line is applied on the bottom:
Screen Shot 2021-10-01 at 2 09 23 PM

FlyersPh9 (approved):

src/inputs/labeled-inputs.scss
mayank99: I'm reverting height to 90%. @bentleyvk

PR #318, opened 2021-10-05
refactor(Alert): Reduce specificity and cleanup

### LINKED ISSUE #317: Alert: anchors within message get extra margin
our anchor rules add margin on all anchor elements in text.. then we get this:
image

_Originally posted by @veekeys in https://github.com/iTwin/iTwinUI-react/issues/334#issuecomment-926514924_

mayank99:
Reduced specificity by generating longer class names using sass suffix selector. Also removed redundant code for statuses by creating a parameterized mixin.

Other notes:
- Added `cursor: pointer` to link so that we don't need to inline it when href is not provided.
- Removed unintentional margin applied to all links within alert message. Closes #317


src/alert/alert.scss
```diff
@@ -15,28 +15,29 @@
...
-      &:hover {
-        text-decoration: none;
-      }
+    &:hover {
+      text-decoration: none;
     }
   }
 
-  > .iui-button {
+  .iui-button {
```
veekeys, line 40: So what about this then? Maybe also `alert-close-button` or something?

src/alert/alert.scss
```diff
@@ -49,3 +50,48 @@
     width: 100vw;
   }
 }
+
+/// Modifier mixin to categorize the alert
+/// @arg $statusColor Can be one of: positive, negative, warning, primary
+/// @arg $rootSelector Selector of the root alert element. Defaults to .iui-alert
+@mixin iui-alert-category($statusColor, $rootSelector: '.iui-alert') {
```
veekeys, line 57: Do you use somewhere this with the differernt rootSelector? I dont think it is needed then here as an arg..

src/alert/alert.scss
```diff
@@ -15,28 +15,29 @@
...
+  }
 
-    > a {
-      margin-left: $iui-s;
-      text-decoration: underline;
-      user-select: none;
-      white-space: nowrap;
+  &-link {
+    cursor: pointer;
+    margin-left: $iui-s;
```
veekeys, line 30: You said you removed margin, but its still here... ?

src/alert/alert.scss
mayank99: I like it this way because it makes it easy to change it if needed, and does not closely tie the mixin to the class name.

src/alert/alert.scss
veekeys: we are adding margin. Isnt that same situation as with anchor? Just difference that its our iui-button class versus anchor element in there...

src/alert/alert.scss
veekeys: Could just had a variable with same outcome,  I guess.. but its minor. When somewhere previously you had mixin with 4-5 args for the same purpose, thats too many, IMO.

src/alert/alert.scss
mayank99: For anchor we are actually styling it by setting `color`, `text-decoration`, `cursor` and some other stuff. For button we are just positioning it using margin. But I can change the class name if you really want.

veekeys (approved):

src/alert/alert.scss
mayank99: ah ok, added back `>` in that case.

If I change to a class name, then it will have a specificity of 0-1-0 (same as `.iui-button` which resets margin) so it would have to rely on order.

src/alert/alert.scss
```diff
@@ -15,28 +15,29 @@
     color: t(iui-text-color);
```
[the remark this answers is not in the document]
mayank99: Done but link looks bad in dark theme
image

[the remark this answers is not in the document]
mayank99: We also don't support non-status alerts in React, so that mockup would need to be constructed through plain html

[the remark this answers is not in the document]
FlyersPh9: The hyperlink part I agree does look bad, maybe we should also color that to our typical hyperlink color?

Would we be able to offer this non-status alert in React?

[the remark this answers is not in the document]
mayank99: I'll create a separate PR for default link styling.

We can offer non-status Alerts if we want, but we do set the "default" to `informational`, so it would be kinda weird to have to set status to `none` or something.

Issue #305, opened 2021-09-21
Tree component


FlyersPh9:
A tree component has been requested by a few people over the past year.  iTwin.js already has some trees implemented (see Tree section [here](https://developer.bentley.com/samples/)), maybe we can start by copying their base HTML?

danleydmello:
For Reliability 4.0, we would like to show hierarchies of objects such as Assets, Indicators etc. 
The Tree component will be the best fit for this. This is going to be read-only UI, so we just need a basic Tree component that shows icon and text for each node in the tree.  And of course expandable and collapsable nodes. Thanks

mayank99:
This is from our Tree standard:
<p>
<img src="https://user-images.githubusercontent.com/9084735/134363583-557da78a-7163-4857-a3a1-b9d5b98c313d.png" width="35%" />
<img src="https://user-images.githubusercontent.com/9084735/134363834-5287a974-db95-494c-ad33-e9c9fbe87d3c.png" width="40%" />
</p>

And this is what our extra-condensed Table looks like with expandable sub-rows:
![image](https://user-images.githubusercontent.com/9084735/134363292-9ae490f9-565c-4c6e-9619-f03b400a4aad.png)

The way I see it, they are pretty close. Just need to work on a few small issues:
- Customizable checkbox
- Spacing and sizing for extra-condensed expander (may be addressed in #252)
- Spacing and sizing for cell content (maybe a custom cell with icon, sublabel, etc)

danleydmello:
@mayank99 , this is interesting. To add to what you addressed about using condensed-table:
1. can row lines be hidden
2. is it possible to hide checkboxes ( for use case its going to be Readonly so no need for checkbox , i think :))

mayank99:
Ah yes, row lines are another issue we'll need to address. However, it is an easy CSS override using `border-color: transparent`.

The checkboxes can easily be turned off in the Table (using `isSelectable` in React, or just excluding that column in HTML), so that's not an issue.

PR #349, opened 2021-10-15
feat(Checkbox): Add visibility checkbox styling

FlyersPh9:
- Add ability to replace checkbox checkmark with eyeball glyph.
- Because an icon is present in the unchecked state, we remove the border of the checkbox.

Is a step towards completing #305.

Screen Shot 2021-10-15 at 10 27 01 AM


src/inputs/checkbox.scss
```diff
@@ -26,4 +26,35 @@
       cursor: wait;
     }
   }
+
+  &.iui-visibility {
```
bentleyvk, line 30: Why it is now needed?

PR #351, opened 2021-10-15
feat(Label): Add non-wrapping label component

FlyersPh9:
Originally an input such as a text field, text area, etc. needed to be wrapped with a `<label>`. This caused issues when trying to right align multiple labels.

This PR introduces the option of not wrapping inputs with labels.
Screen Shot 2021-10-18 at 4 40 34 PM

- `@mixin iui-input-label` becomes `@mixin iui-input-container-label`.
- Introducing new `@mixin iui-input-label` and its class `.iui-input-label`.
- Refactored common styles into `@mixin iui-input-label-styling` and `@mixin iui-input-label-cursor`.


src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,45 @@
   }
 }
 
+/// Wrapping
+@mixin iui-input-label-wrapper {
```
mayank99, line 171: maybe call this `iui-input-container-label` and add more info the comment above

src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,45 @@
...
+
+  @at-root {
+    label#{&} {
+      cursor: pointer;
+
+      &.iui-disabled {
+        cursor: not-allowed;
+      }
+    }
+  }
```
mayank99, line 192: Do we ever want to show a `default` cursor? I think this class should always be applied on a `label` element, so you could simplify the rules
```suggestion
  cursor: pointer;

  &.iui-disabled {
    cursor: not-allowed;
  }
```

src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,45 @@
...
 }
 
+/// Wrapping
+@mixin iui-input-label-wrapper {
+  @include iui-input-label-styling;
+  grid-area: label;
+  align-self: center;
+}
+
+/// Non-wrapping
```
mayank99, line 177: Would be nice to document what this mixin is for, what the modifiers are, and what's the difference between default and inline.

src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,45 @@
...
+@mixin iui-input-label {
+  @include iui-input-label-styling;
+  display: block;
+  width: fit-content;
```
mayank99, line 181: Do you really need these?

src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,45 @@
...
+    margin-right: $iui-m;
+    display: inline-block;
+
+    &.iui-required {
+      margin-right: $iui-xs * 1.5;
+    }
+
+    ~ .iui-input-container {
+      display: inline-grid;
+    }
```
mayank99, line 205: not sure about this inline-grid

src/inputs/labeled-inputs.scss
FlyersPh9: The default cursor is used if applied to a `<div>` and not a `<label>` like when we show a label for a group of checkboxes.

src/inputs/labeled-inputs.scss
FlyersPh9: `display: block` and further down `display: inline-block` can be safely removed now that we have the grid.

`width: fit-content` however would be nice to keep. Without it you get the pointer cursor when hovering anywhere in this highlighted area.
Screen Shot 2021-10-18 at 5 39 36 PM

src/inputs/labeled-inputs.scss
FlyersPh9: I updated the descriptions a little bit but probably not descriptive enough.

src/inputs/labeled-inputs.scss
FlyersPh9: I'm not so sure, because if I place the inline checkbox / radio as it is currently we end up with something like this:
Screen Shot 2021-10-19 at 5 06 00 PM

We would want something more like this:
Screen Shot 2021-10-19 at 5 07 32 PM

FlyersPh9: And if I place the radio component outside of the grid, the labels don't align:
Screen Shot 2021-10-19 at 5 08 36 PM

src/inputs/labeled-inputs.scss
```diff
@@ -169,6 +167,39 @@
...
+  }
+
+  &.iui-inline {
+    margin-bottom: 0;
+    margin-right: $iui-m;
+
+    &.iui-required {
+      margin-right: $iui-xs * 1.5;
+    }
+  }
```
mayank99, line 200: You might be able to remove these styles and reuse the `iui-input-label-inline` mixin

backstop/tests/text-field.html
```diff
@@ -29,6 +29,17 @@
         flex-direction: column;
         gap: 11px;
       }
+
+      #block-labels { display: grid; }
```
mayank99, line 33: Do you plan to clean up these head styles? (could remove `display: grid` from #block-labels and could add `:not(:last-child)` to the margin-bottom in the next line, though you might need to add a `<br />` to separate the block and inline labels)

I only added these rules to demonstrate what was possible. In this HTML page, the emphasis should be on the styling more than the layout. And then you can have proper grid layouts in your info panel PR.

src/inputs/labeled-inputs.scss
mayank99: > Without it you get the pointer cursor when hovering anywhere in this highlighted area.

It seems this is only happening because of `display: grid` (which we don't need for block layout) so it could be removed

mayank99: > display: block and further down display: inline-block can be safely removed now that we have the grid.

I think people should be allowed to use this component without grid, so maybe we should add back `display: inline-block` (without it, label defaults to `display: inline`).

FlyersPh9: Removed `width: fit-content` and replaced with `display: inline-block`.

backstop/tests/text-field.html
```diff
@@ -29,6 +29,16 @@
         flex-direction: column;
         gap: 11px;
       }
+
+      #block-labels > .iui-input-container { margin-bottom: 11px; }
+
+      #inline-labels {
+        display: grid;
+        row-gap: 11px;
```
saaaaaally, line 37: why not use iui-baseline?

mayank99: it's in the `head` of the html file, you can't use Sass variables here and it also doesn't matter because it's not part of the component

saaaaaally (approved):

backstop/tests/text-field.html
```diff
@@ -29,6 +29,16 @@
         flex-direction: column;
         gap: 11px;
       }
+
+      #block-labels > .iui-input-container { margin-bottom: 11px; }
```
bentleyvk, line 33: Move CSS rules to a new line. Same for bellow one.

backstop/tests/text-field.html
```diff
@@ -302,37 +312,37 @@ <h2>Hybrid layouts</h2>
 
     <h2>Without label</h2>
     <section id="demo-no-label">
-      <label class="iui-input-container">
+      <div class="iui-input-container">
```
bentleyvk, line 315: Why changed to `div`?

src/inputs/classes.scss
```diff
@@ -36,6 +36,14 @@
   }
 }
 
+.iui-input-label {
```
bentleyvk, line 39: Hm, should it have `input` in the name? We/Users might want to use label in other places just to have the same styling.

backstop/tests/text-field.html
mayank99: image

Doesn't make sense to use `label` when there's no label

backstop/tests/text-field.html
mayank99: Done

src/inputs/classes.scss
bentleyvk: `iui-component-label`? 😁

bentleyvk (approved):

PR #355, opened 2021-10-20
fix(ColorPicker): Add spacing between color field & adjacent slider

FlyersPh9:
Previously the color field dot would come into contact with the slider grab handle:
Screen Shot 2021-10-20 at 12 38 51 PM

Fixed:
Screen Shot 2021-10-20 at 12 53 20 PM


src/color-picker/color-picker.scss
```diff
@@ -165,6 +165,10 @@ $iui-focus-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-primary-
     background-image: linear-gradient(0deg, rgb(0, 0, 0), transparent),
       linear-gradient(90deg, rgb(255, 255, 255), var(--hue));
   }
+
+  + .iui-slider-component-container {
```
mayank99, line 169: Where does this class come from?

saaaaaally: I added margin-bottom to iui-color-field

src/color-picker/color-picker.scss
```diff
@@ -122,11 +122,22 @@ $iui-focus-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-primary-
...
 
+    @supports (gap: $iui-xxs) {
+      > *:not(:last-child) {
+        margin-right: 0;
+      }
+    }
+
+    > * {
+      flex-grow: 1;
+    }
```
mayank99, line 139: Can make it more readable
```suggestion
    > * {
      flex-grow: 1;

      &:not(:last-child) {
        margin-right: $iui-xxs;
        @supports (gap: $iui-xxs) {
          margin-right: 0;
        }
      }
    }
```

mayank99 (approved):

PR #362, opened 2021-10-26
feat(labeled-inputs): Add css variable for inline padding

mayank99:
Added `--inline-icon-width` variable to be set in JS (see https://github.com/iTwin/iTwinUI-react/pull/403). Allows putting any arbitrary content as the inline icon, instead of limiting to 40px icon buttons. Falls back to 40px if not specified.

Before:
image
image

After:
image
image

image

### DIFF
--- a/src/inputs/labeled-inputs.scss
+++ b/src/inputs/labeled-inputs.scss
@@ -23,7 +23,8 @@
   &.iui-inline-icon {
     > .iui-input,
     > .iui-textarea {
-      padding-right: $iui-icons-default + $iui-l;
+      padding-right: ($iui-icons-default + $iui-l);
+      padding-right: var(--inline-icon-width, ($iui-icons-default + $iui-l));
 
       &:last-child {
         padding-right: $iui-sm;


src/inputs/labeled-inputs.scss
```diff
@@ -23,7 +23,8 @@
   &.iui-inline-icon {
     > .iui-input,
     > .iui-textarea {
-      padding-right: $iui-icons-default + $iui-l;
+      padding-right: ($iui-icons-default + $iui-l);
```
bentleyvk, line 26: So it won't work on IE. Is it worth having this when you started working on this PR https://github.com/iTwin/iTwinUI/pull/365?

mayank99: I am not sure if it's worth it. That's why I didn't merge it

FlyersPh9 (approved):

src/inputs/labeled-inputs.scss
mayank99: #365 is kind of in a bad state and I might rewrite the code in there.

So maybe we should merge this? It can't hurt (IE will still use 40px), and might even help in cases where the borderless button width is not 40px.

veekeys: Before merge, can we come up with a standard to our variables namings? Something like iui-{component}-{element}-{value}

mayank99: So in this case it would be something like `--iui-input-icon-inline-padding`?

veekeys: amm..something, even though it sounds like it is padding for icon, whereas actually it is padding of input itself

veekeys: cant it be just `--iui-input-right-padding` ?
Yes, it is now used only in inlice icon context, but.. maybe it could be used in other cases, just to set right padding dynamically???

bentleyvk: I looked at React implementation and it is a bit ugly 😕 You need to handle all of the sizing when in reality correct CSS could do that easily.
image

mayank99: > You need to handle all of the sizing when in reality correct CSS could do that easily.

what do you mean by "correct CSS"?

bentleyvk: Having button next to input not inside 😀

mayank99:
Closing this as there won't be any valid use cases for this once #365 is available.

PR #363, opened 2021-10-27
feat(Buttons): Reduce nesting and specificity

gretanausedaite:
- Renamed `iui-label` to `iui-button-label` for buttons
- Renamed `iui-icon` to `iui-button-icon` and created mixin for unified styles
- Removed `:enabled` tag from selector
- Removed specificity in multiple mixins
- Refactored base, hover and focus styles in multiple buttons mixins
- Changed `iui-button-size` mixin to take parameters for size (small, medium, large) and style (default or borderless)
- Changed `margin-left` to `gap` in `iui-button-size` mixin
- Improved `disabled` selector
- Changed other components to reflect changes made to buttons
- Updated HTML code to reflect changes made to buttons


src/button/button.scss
```diff
@@ -7,6 +7,44 @@ $iui-button-padding: $iui-xs * 4;
...
+    }
+  }
+
+  > .iui-icon {
+    @include themed {
+      fill: t(iui-icons-color-actionable);
+    }
+  }
+
+  &:enabled:hover,
```
FlyersPh9, line 25: Is it possible to remove `:enabled` from all of the button's CSS?

mayank99: If the specificity of the hover selector stays low, then disabled styling will take preference if it comes later:
```scss
&:hover { ... }
&:disabled { ... }
```

src/button/disabled.scss
```diff
@@ -16,9 +16,29 @@
...
 
-  > .iui-icon {
+  > .iui-button-icon {
     @include themed {
       fill: t(iui-icons-color-actionable-disabled);
     }
   }
+
+  &:hover,
+  &:active {
```
mayank99, line 26: No reason to repeat all these rules. 

Ideally you won't even need these selectors but worst case, you can use `&, &:hover, &:active` if it doesn't work

src/button/button.scss
```diff
@@ -28,17 +29,48 @@ $iui-button-padding-large: $iui-xs * 6;
...
+        fill: t(iui-icons-color-primary);
+      }
+    }
+  }
+
+  &:hover {
+    text-decoration: none;
   }
 
   &:enabled:hover {
```
mayank99, line 76: This is repeated now

src/button/button-icon.scss
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+@mixin iui-button-icon {
+  &:not(.iui-user-icon) {
```
mayank99, line 7: no need for `:not(iui-user-icon)` anymore. This was only added because the specificity of `iui-button > iui-icon` was too high.

src/button/button-icon.scss
```diff
@@ -0,0 +1,14 @@
...
+    width: $iui-icons-default;
+    height: $iui-icons-default;
+    @media (prefers-reduced-motion: no-preference) {
```
mayank99, line 10: a fill transition does not really count as "motion" so I think this media query could be removed

src/button/default.scss
```diff
@@ -27,13 +36,34 @@
...
+      @include themed {
+        fill: t(iui-icons-color-actionable-disabled);
+      }
+    }
+
+    &.iui-active {
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), 0.05);
+      }
+    }
```
mayank99, line 64: I think all of these rules could moved into the `iui-button-disabled` mixin.

I see that these rules should not apply to cta/hv buttons so maybe the mixin could have a parameter.

src/button/default.scss
```diff
@@ -27,13 +36,34 @@
...
       @include themed {
         fill: t(iui-icons-color-actionable-hover);
       }
     }
   }
 
+  &[disabled],
+  &[disabled]:hover,
+  &[disabled]:active,
+  &[disabled]:focus {
```
mayank99, line 49: This selector doesn't need all these states. Also consider adding `&:disabled` because `[disabled] not work inside a fieldset. Need both because we are adding support for links as buttons, and `:disabled` will not work on links.

```suggestion
  &[disabled],
  &:disabled {
```

src/button/high-visibility.scss
```diff
@@ -10,22 +11,26 @@
     color: t(iui-color-foreground-accessory);
   }
 
-  > .iui-icon {
+  &[disabled] {
```
mayank99, line 14: disabled selector should come *after* all the hover/focus/active states (a later rule takes preference if they have the same specificity).

Should be fixed in all files

src/button/high-visibility.scss
```diff
@@ -10,22 +11,26 @@
...
     @include themed {
       fill: t(iui-color-foreground-accessory);
     }
   }
 
-  &:enabled:hover,
-  &:enabled:active,
+  &:hover,
+  &:active,
   &:focus-visible {
```
mayank99, line 26: unrelated to your changes focus-visible should be removed from this list to fix Safari/IE bug. https://css-tricks.com/one-invalid-pseudo-selector-equals-an-entire-ignored-selector/

Suggest changing it to just `:focus` and then adding a separate `:focus:not(:focus-visible)` to negate those styles.

src/button/button.scss
```diff
@@ -28,17 +29,48 @@ $iui-button-padding-large: $iui-xs * 6;
...
+    @include iui-button-disabled;
+  }
+
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+    color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+  }
+
+  &.iui-active {
```
mayank99, line 46: Why moved from `default.scss`?

src/button/button.scss
```diff
@@ -28,17 +29,48 @@ $iui-button-padding-large: $iui-xs * 6;
...
-      transition: fill $iui-speed-fast ease-out;
+  &[disabled] {
+    @include iui-button-disabled;
+  }
+
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+    color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+  }
```
mayank99, line 44: Dont think these are needed in this mixin.

backstop/tests/button.html
```diff
@@ -110,7 +110,7 @@ <h3>High visibility</h3>
 
         <button class="iui-button iui-high-visibility iui-small">
           <svg-placeholder
-            class="iui-icon"
+            class="iui-button-icon "
```
mayank99, line 113: There's extra spaces everywhere. Recommend manually formatting with prettier and then saving (which will format again with vscode formatter).

src/button/borderless.scss
```diff
@@ -17,14 +18,18 @@
     padding: 0 $iui-sm;
   }
 
+  &[disabled] {
+    @include iui-button-disabled;
+  }
+
   // If icon is by itself, button takes 1:1 size ratio
-  > .iui-icon:only-child {
+  > .iui-button-icon:only-child {
```
mayank99, line 26: Unrelated to this PR, but I don't think this comment is correct. 40x38 is not 1:1

@FlyersPh9

src/button/cta.scss
```diff
@@ -2,33 +2,41 @@
...
 
-    > .iui-icon {
+    > .iui-button-icon {
       @include themed {
         fill: t(iui-color-foreground-accessory);
       }
     }
   }
+
+  &[disabled] {
```
FlyersPh9, line 39: Can we add a class name here in addition to this?  Something like `&[disabled], &.iui-disabled`.

Reason I'm asking is when a button is disabled, it can no longer show a tooltip on hover.  I think if we offer a class name that makes a button appear disabled, we should be able to show a tooltip on hover / focus.
MicrosoftTeams-image

mayank99: That's a bad idea.

Disabled buttons cannot be interacted with in any way (i.e. no click, no focus, AND **no focus** which is why https://github.com/iTwin/iTwinUI-react/issues/30 is a known bug). If we were to use a class name instead of the `disabled` attribute, it would require manually disabling the click handler and removing the button from tabbing sequence and adding aria labels for screen readers, and who knows what else.

The workaround for tooltips is to add a `div` around the disabled button, but that also breaks our button-group styling.

src/button/borderless.scss
FlyersPh9: Oops, you are right.  We can remove / reword that comment.

mayank99: I was expecting you'd say we should fix the size to be 38x38 😆

src/button/button.scss
```diff
@@ -28,44 +29,45 @@ $iui-button-padding-large: $iui-xs * 6;
...
 
   @include iui-focus;
   @include iui-button-sizes;
+
+  &[disabled] {
+    @include iui-button-disabled;
+  }
 }
 
 @mixin iui-button-sizes {
```
mayank99, line 54: I think it would be nice to refactor this mixin to not require classes.

e.g. you could add `@include iui-button-sizes(large)` to header-button, and then you can use it like this:
```diff
- <button class="iui-button iui-large iui-header-button" ... />
+ <button class="iui-button iui-header-button" ... />
```

And then we can reduce the specificity of overrides done in header.scss.

src/button/cta.scss
FlyersPh9: Here's an article I bookmarked a while back that should help here: https://css-tricks.com/making-disabled-buttons-more-inclusive/

mayank99: Yeah so it looks like the discussion should be centered around these 5 points from the article. And we can narrow it down to the top 3 points (because semantics and styles are covered with CSS and aria-disabled):
image

If we choose to make disabled buttons focusable and hoverable, then it will introduce an inconsistency with other disabled elements, such as tabs and menu items. So this decision needs to be very carefully finalized.

src/button/button.scss
```diff
@@ -28,57 +29,79 @@ $iui-button-padding-large: $iui-xs * 6;
...
   padding: 0 $iui-button-padding;
   height: $iui-component-height;
 
-  > .iui-icon + .iui-label,
-  > .iui-label + .iui-icon,
-  > .iui-icon + .iui-icon {
+  > .iui-button-icon + .iui-label,
+  > .iui-label + .iui-button-icon,
+  > .iui-button-icon + .iui-button-icon {
     margin-left: $iui-s;
```
mayank99, line 82: Can't we just use `gap` for everything here?

src/button/cta.scss
```diff
@@ -2,33 +2,41 @@
...
 @mixin iui-button-cta {
-  @include themed {
-    background-color: t(iui-color-background-positive);
-    border-color: t(iui-color-background-positive);
-    color: t(iui-color-foreground-accessory);
-  }
-
-  > .iui-icon {
+  &,
+  &:focus:not(:focus-visible) {
```
mayank99, line 9: focus-visible should NOT appear alongside any other selectors. If there is a comma in your selector, dont use focus-visible

And why are you putting it around the base styles anyway?

src/button/button.scss
```diff
@@ -28,57 +29,79 @@ $iui-button-padding-large: $iui-xs * 6;
...
+    @include iui-button-sizes(small);
+  }
+
+  &.iui-large {
+    @include iui-button-sizes(large);
+  }
+
   @include iui-button-sizes;
+
+  &[disabled] {
```
mayank99, line 70: This selector needs to be fixed (add `:disabled` and maybe even `[aria-disabled]='true'` in all places)

src/button/borderless.scss
```diff
@@ -18,55 +19,50 @@
...
       @include themed {
-        background-color: rgba(
-          t(iui-color-foreground-primary-rgb),
-          t(iui-opacity-6)
-        );
+        background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
       }
       border-color: transparent;
+
+      > .iui-label {
```
mayank99, line 44: I'm thinking iui-label also needs to be iui-button-label

src/button/button.scss
```diff
@@ -28,57 +29,79 @@ $iui-button-padding-large: $iui-xs * 6;
...
-    height: $iui-icons-default;
-    @media (prefers-reduced-motion: no-preference) {
-      transition: fill $iui-speed-fast ease-out;
+  @include themed {
+    color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+  }
+
+  > .iui-button-icon {
+    @include themed {
+      fill: t(iui-icons-color-actionable);
```
mayank99, line 42: If we need to target this class inside all our button variants, then the specificity still remains high.

I'm thinking maybe the `iui-button-icon` mixin can take parameters to set fill

src/button/button.scss
```diff
@@ -56,14 +56,23 @@ $iui-button-padding-large: $iui-xs * 6;
...
+
   @include iui-button-sizes;
 
   &[disabled] {
     @include iui-button-disabled;
   }
 }
 
-@mixin iui-button-sizes {
+@mixin iui-button-sizes($size: medium) {
```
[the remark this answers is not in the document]
mayank99: Also maybe it should be called `iui-button-size` singular

src/button/button.scss
gretanausedaite: Removed all `:enabled` from buttons.

src/button/disabled.scss
gretanausedaite: Deleted repeating code.

src/button/high-visibility.scss
gretanausedaite: Fixed focus, not-focus-visible code.

src/button/button.scss
gretanausedaite: Moved back.

src/button/button.scss
gretanausedaite: You're right.

src/button/cta.scss
gretanausedaite: So should I add `.iui-disabled` to disabled styles?

src/button/button.scss
gretanausedaite: Changed this mixin

src/button/button.scss
```diff
@@ -28,57 +29,79 @@ $iui-button-padding-large: $iui-xs * 6;
...
+
+  > .iui-button-icon {
+    @include themed {
+      fill: t(iui-icons-color-actionable);
     }
   }
 
-  &:enabled:hover {
+  &:hover,
+  &:focus-visible {
```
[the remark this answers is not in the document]
gretanausedaite: fixed all focus - focus-visible

src/button/button.scss
gretanausedaite: Using gaps now.

src/button/default.scss
```diff
@@ -47,10 +50,9 @@
     }
   }
 
-  &.iui-active:not(.iui-high-visibility):not(.iui-cta) {
-    &:enabled,
-    &:enabled:hover,
-    &:enabled:active {
+  &.iui-active {
+    &,
+    &:hover {
```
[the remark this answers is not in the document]
gretanausedaite: Added back `:active`

src/button/button.scss
gretanausedaite: Added `:disabled`

src/button/borderless.scss
gretanausedaite: Changed to iui-button-label

src/button/button.scss
gretanausedaite: Added parameter to fill icons.

src/button/borderless.scss
gretanausedaite: Removed comment

src/header/header.scss
```diff
@@ -257,10 +260,16 @@
...
 
-  > .iui-icon {
+  > .iui-button-icon {
     flex-shrink: 0;
   }
 
+  > .iui-button-icon + .iui-button-label,
+  > .iui-button-label + .iui-button-icon,
+  > .iui-button-icon + .iui-button-icon {
+    margin-left: $iui-sm;
```
mayank99, line 270: Why add this back? / why unset the gap?

src/header/header.scss
```diff
@@ -172,7 +173,8 @@
   @at-root {
     a#{&}, // rendered as anchor element
     button#{&}, // rendered as button element
-    &[role='button'] { // rendered as div with role=button
+    &[role='button'] {
+      // rendered as div with role=button
```
mayank99, line 177: I'd say remove this comment as it does not make sense on the next line, and prettier won't allow it on the same line

src/header/header.scss
```diff
@@ -154,7 +154,8 @@
     @include iui-header-button-icon;
   }
 
-  .iui-header-button.iui-header-button { // boost specificity
+  .iui-header-button.iui-header-button {
+    // boost specificity
```
mayank99, line 158: Remove or move to previous line

src/code/codeblock.scss
```diff
@@ -26,6 +26,10 @@
     }
   }
 
+  .iui-button {
+    gap: unset;
```
mayank99, line 30: Why unset?

src/button/default.scss
```diff
@@ -59,17 +53,34 @@
...
+@mixin iui-button-default-base {
+  @include themed {
+    background-color: t(iui-color-background-1);
+    border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+    color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+  }
+
+  &.iui-active {
+    > .iui-button-icon {
+      @include iui-button-icon(t(iui-icons-color-primary));
```
mayank99, line 83: There are still two problems in this approach:
1. Including this mixin again will repeat all those rules
2. Plus all the nesting again makes it hard to override icon styles (i think this one is some 0-4-0)

I wonder if we should just change the value of fill CSS variable, instead of setting all those styles again. It won't work in IE but it will help

src/header/header.scss
```diff
@@ -319,6 +328,10 @@
   &.iui-active {
     @include iui-header-button-active;
 
+    .iui-button-label {
+      color: t(iui-color-foreground-primary);
```
mayank99, line 332: Why is this needed?

src/side-navigation/side-navigation.scss
```diff
@@ -100,20 +101,20 @@
       }
     }
 
-    > .iui-icon:not(.iui-user-icon) {
+    > .iui-button-icon:not(.iui-user-icon) {
```
mayank99, line 104: `:not(.iui-user-icon)` can be removed now (also search for other instances in the project)

src/button/borderless.scss
```diff
@@ -2,65 +2,77 @@
...
+  &[disabled] {
+    @include iui-button-disabled;
+  }
+
   &[disabled],
-  &[disabled]:hover,
-  &[disabled]:active,
-  &[disabled]:focus {
+  &:disabled,
+  &:disabled:hover {
```
mayank99, line 54: This selector is messed up.

1. Why have `[disabled]` separately for including the mixin?
2. Why include `hover` here?

src/button/button.scss
mayank99: Then we should improve some selectors in header.scss

src/header/header.scss
bentleyvk: Wasn't prettier that made these changes? Maybe need to fix rules there?

src/button/default.scss
bentleyvk: Button not fully working in IE? Big no no no!

src/header/header.scss
mayank99: well, prettier is not configurable so we dont have a choice

src/button/default.scss
mayank99: it already doesn't fully work in both Safari & IE and no one complained 😆

src/header/header.scss
bentleyvk: What you mean not configurable? Also, @gretanausedaite have you ran `yarn install` before working on buttons? I am wondering was it prettier that made the changes or Greta.

src/button/default.scss
bentleyvk: What? What is not working there?

src/header/header.scss
mayank99: prettier is not configurable! There is no way to tell prettier to keep that comment on the same line as it does not like anything after `{`.

image

bentleyvk: Ok, then I don't understand why prettier changed comment placement here and not on the PR when you added prettier to the pre-commit. And if it was the prettier that changed it then no one can blame Greta for this comment movement 🤷‍♂️

mayank99: I think it's because this file was not staged in my PR? Not sure.

And no one is blaming Greta 😆 but we do need to fix the formatting, as this comment does not make sense on the next line

bentleyvk: I thought you ran prettier through all the files... Then there can be more random changes like this.

gretanausedaite: Can't we just add comments before statements and not after them if prettier just keeps "fixing" them? 
This doesn't look bad:

// boost specificity
.iui-header-button.iui-header-button {

mayank99: that's something I also suggested for this
image

src/button/default.scss
mayank99: hover and focus states are broken in many of the button components (or derived ones like sidenav-button/header-button), because we use `&:hover, &:focus-visible`  in many places.

src/header/header.scss
gretanausedaite: Got `husky > pre-commit` error while trying to push code with fix. 
`--no-verify` command worked, so should be same line now.

src/side-navigation/side-navigation.scss
mayank99: Wait I think tests are failing because of the removal, so maybe not.

src/header/header.scss
gretanausedaite: Because left gap does not disappear when button is getting smaller on smaller screens:
image
image

src/code/codeblock.scss
gretanausedaite: Because gap is not needed in this.
No gap:
image
Gap: 
image

src/button/borderless.scss
```diff
@@ -48,8 +48,15 @@
   &[disabled],
   &:disabled {
     @include iui-button-disabled;
-    background-color: transparent;
-    border-color: transparent;
+  }
+
+  &[disabled],
```
gretanausedaite, line 53: If I add both @include and transparent stuff into same selector, I get this (disabled background is not transparent and also hoverable): 
image
Would appreciate some help how to make this work.

src/header/header.scss
mayank99: In that case, can we at least change the selector? We don't need those specific classes.
```suggestion
  > * + * {
    margin-left: $iui-sm;
```

src/side-navigation/side-navigation.scss
```diff
@@ -99,18 +99,19 @@
...
         border-color: t(iui-color-background-5);
       }
-    }
 
-    > .iui-button-icon {
-      width: $iui-icons-large;
-      height: $iui-icons-large;
-      flex-shrink: 0;
+      //boost specificity
+      > .iui-button-icon.iui-button-icon {
```
mayank99, line 104: This is now boosting the specificity to 0-6-0 (and the label has 0-7-0), which hardly seems necessary. Why do you need it?

src/side-navigation/side-navigation.scss
```diff
@@ -99,18 +99,19 @@
...
-        margin-left: $iui-icons-large;
-        white-space: nowrap;
-        overflow: hidden;
-        text-overflow: ellipsis;
+        + .iui-button-label {
+          margin-left: $iui-icons-large;
+          white-space: nowrap;
+          overflow: hidden;
+          text-overflow: ellipsis;
+        }
```
mayank99, line 114: Will gap not work here? I see you unset it at the top, but you could just change it to `$iui-icons-large`.

If not, we could still remove the dependency on iui-button-label
```suggestion
        &:not(:last-child) {
          margin-right: $iui-icons-large;
        }
```

And those overflow rules don't need to be nested this deep so you could add those somewhere above.

src/button/borderless.scss
mayank99: Previously disabled borderless button had `background-color: rgba(var(--iui-color-foreground-body-rgb), 0.05)`. Where did that go?

gretanausedaite: Borderless disabled buttons have background transparent. 
Borderless disabled active buttons have this bacground - I'll add it back, it's lost somewhere

src/button/default.scss
mayank99: I guess lets tackle CSS variables in another PR in 2022. Created issue #390 

And focus-visible should be fixed in this PR hopefully

src/button/cta.scss
mayank99: Lets save this for another PR. Created issue #391

src/side-navigation/side-navigation.scss
mayank99: I see you went with the suggested change. Did `gap: $iui-icons-large` not work?

src/header/header.scss
```diff
@@ -290,7 +302,7 @@
             padding: $iui-xxs;
           }
 
-          .iui-label {
+          .iui-button-label {
```
mayank99, line 305: This doesn't need to be nested inside .iui-button anymore.

src/button/classes.scss
```diff
@@ -24,14 +24,25 @@
 
   &.iui-borderless {
     @include iui-button-borderless;
+    @include iui-button-size(medium, borderless);
```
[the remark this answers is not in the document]
mayank99: I see you also added back `.iui-small` and `.iui-large` back. Unless there is a specific reason for it, I think those classes should stay in `classes.scss`. If there is a need to use `iui-button-borderless()` with small/large size, then `iui-button-size()` can be included separately.

src/side-navigation/side-navigation.scss
gretanausedaite: `gap` did not work.

src/button/classes.scss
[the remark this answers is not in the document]
gretanausedaite: Tiles borderless buttons doesn't work when .iui-small and .iui-large is added to classes.

src/header/header.scss
gretanausedaite: removed nesting

src/button/classes.scss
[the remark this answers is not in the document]
mayank99: What doesn't wo rk?

src/button/default.scss
```diff
@@ -47,28 +41,39 @@
...
+    }
+
+    .iui-button-label {
       @include themed {
-        background: linear-gradient(
-          rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6)),
-          rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6))
-        ),
-          linear-gradient(t(iui-color-background-1), t(iui-color-background-1));
+        color: t(iui-color-foreground-primary);
```
mayank99, line 55: This label color can be set one level above because it gets inherited.

mayank99: Same applies in other places (e.g. `borderless.scss` and `disabled.scss`)

src/button/default.scss
```diff
@@ -2,38 +2,32 @@
...
       color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-1));
     }
     @include iui-notification-marker-hover;
 
-    > .iui-icon {
-      @include themed {
-        fill: t(iui-icons-color-actionable-hover);
-      }
+    .iui-button-icon {
+      @include iui-button-icon(t(iui-icons-color-actionable-hover));
```
mayank99, line 23: This one should also not include the full mixin.

src/button/default.scss
```diff
@@ -2,38 +2,32 @@
...
     }
   }
 
+  &:focus:not(:focus-visible) {
+    @include iui-button-default-base;
+  }
+
   &.iui-dropdown {
```
mayank99, line 31: Nit: this could be moved into a mixin similar to `iui-button-size` and included in `classes.scss` for consistency

src/button/button-icon.scss
```diff
@@ -0,0 +1,16 @@
...
+
+/// Mixin for button icons
+/// @arg fillColor - color to fill in icon
+@mixin iui-button-icon($fillColor: t(iui-icons-color-actionable)) {
+  width: $iui-icons-default;
+  height: $iui-icons-default;
+  transition: fill $iui-speed-fast ease-out;
+
+  @include themed {
+    fill: $fillColor;
```
mayank99, line 14: So as I was reviewing the visuals, I realized that the icon color is always the same as text color. So I think we'll be able to reduce complexity a lot just by using `fill: currentColor`, provided the `color` is set on the parent (which it should). We won't need to override the fill in various places.

src/side-navigation/side-navigation.scss
mayank99: Are you sure? It seems to work on my end. All I did is remove `margin-right: $iui-icons-large` and set `gap: $iui-icons-large` on the parent.

gretanausedaite: Fixed it. Your suggestion worked.

src/button/classes.scss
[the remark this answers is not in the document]
gretanausedaite: Nvm. Fixed it.

src/button/button-icon.scss
gretanausedaite: Please check code if I understood you correctly. Made some changes.

mayank99: I just approved the failing tests in [`584909a`](https://github.com/iTwin/iTwinUI/pull/363/commits/584909a0d034de52be998d406b54bdb8615fe2d2). It is a slight change in opacity of icons, but I think it might actually have a better outcome

mayank99 (approved):

PR #365, opened 2021-11-02
feat(ButtonGroup): handle input + button combo

### LINKED ISSUE #243: Combo input + button
image
image

Similar to #93

mayank99:
Modified iui-button-group to handle inputs. Biggest hurdle was handling borders and focus styling. Used negative margin and added z-index to prevent clipping.

Each item inside the group is now wrapped by a div to allow using first/last-child selectors while also supporting nested elements and popovers. Also affects split-menu which is built on top of button-group.

image

Closes #243


src/button/button-group.scss
```diff
@@ -6,45 +6,53 @@
   display: inline-flex;
   align-items: center;
 
-  > .iui-button {
+  $childSelector: '&, input, button'; // target nested inputs and buttons
```
mayank99, line 9: nested inputs need to be targeted because inputs can be inside `iui-input-container`, and disabled buttons can be inside a wrapper.

src/button/button-group.scss
```diff
@@ -6,45 +6,53 @@
...
+
+  > * {
     &:not(.iui-borderless) {
-      border-radius: 0;
-    }
+      #{$childSelector} {
+        border-radius: 0;
+      }
+
+      &:first-child {
```
mayank99, line 17: I remember we changed to first-of-type/last-of-type because split-menu has a tippy as last-child.

But we can't use first-of-type/last-of-type anymore because of the mixed nature of these combo groups. So we might need to move the menu into a separate div.

src/button/button-group.scss
```diff
@@ -6,45 +6,53 @@
...
+    &:not(:last-child) {
+      margin-right: -1px;
     }
 
-    &:not(.iui-borderless):not(.iui-cta):not(.iui-high-visibility):not(:last-of-type) {
-      border-right: none;
+    &:hover,
+    &:focus,
+    &:focus-within {
+      z-index: 2;
```
mayank99, line 39: I'm afraid this z-index could cause unexpected issues in some places but I can't think of any

src/button/high-visibility.scss
```diff
@@ -31,4 +30,10 @@
       }
     }
   }
+
+  &:focus {
+    @include themed {
+      outline-color: t(iui-color-foreground-accessory);
```
mayank99, line 36: @FlyersPh9 What do you think of only changing (in #166) the outline-color and leaving the offset as -1px for hv/cta buttons? Looks more consistent imo. But we could play with the opacity ofc
image
image

mayank99: ah wait, it is not visible in light theme so we'd need a different color

FlyersPh9: 1 option would be to use the same color as body text, but I think a black border on a blue button in light theme won't look so great.  Maybe if we combine the body text color with opacity so some of the blue / green button comes through, maybe that will look better?

mayank99: I reverted the focus styling for now. Lets handle it in yours.

src/button/button-group.scss
bentleyvk: What if users have Tooltips on buttons inside button group? Then it won't work too and we don't control that.

src/button/button-group.scss
bentleyvk: @veekeys you seen some `z-index` related bugs in APIM. Any thoughts?

veekeys: Yeah.... Is there any different solution than z-index? I would urge not to use it if possible. Usually z-index issues appear when using multiple z-index based components in single place (in APIM there was with the popover based elements and overlay).

src/button/button-group.scss
veekeys: we want to target native html elements or only itwinui ones?

src/button/button-group.scss
FlyersPh9: As I understand it, the `z-index` was suggested so that the outer glow effect on focus would not get clipped by the element directly to the right of it.  This is a known issue in many other components, but will be fixed when #166 passes.  I vote we remove the `z-index` and approve this PR knowing that the issue is wide spread but will be fixed soon.

mayank99: It will NOT be automatically fixed when #166 passes, but it could maybe be worked around with some nasty border hacks.

> in APIM there was with the popover based elements and overlay

would like to see an example

src/button/button-group.scss
mayank99: I think the users will just need to move the tooltip outside the div 🤷‍♀️

src/button/button-group.scss
mayank99: > we want to target native html elements or only itwinui ones?

Not sure what you mean. itwinui components use the native elements. I am targeting `input`/`button` instead of classes to keep the specificity low

src/button/button-group.scss
bentleyvk: But it will be kind of a breaking change then and not really obvious one.

src/button/button-group.scss
mayank99: I created a new stacking context using `isolation: isolate` so the z-index should be sealed off from outside influence.

src/button/button-group.scss
mayank99: I added wrapper divs to each child. So now if there is a need to put tooltips on a disabled button, it will work like this:
```jsx
<div.button-group>
  <div> // This div will be added to all children regardless
    <div> // Tooltip component will wrap this element, but the DOM will be appended as a sibling
      <button disabled />
    </div>
  </div>
   ...
</div>

File: backstop/tests/header.html [hunk 5]
```diff
@@ -641,28 +641,32 @@ <h2>
...
-                <button
-                  class="iui-button iui-borderless"
-                  id='test-button-6'
-                >
-                  <svg-caret-down-small
-                    class="iui-button-icon"
-                    aria-hidden="true"
-                  ></svg-caret-down-small>
-                </button>
+                <div>
```
bentleyvk, line 644: Why wrapper here is needed?

src/button/button-group.scss
```diff
@@ -5,72 +5,79 @@
...
+        position: relative;
+        z-index: -1;
+      }
 
-    &.iui-default {
-      // Border trick for grouped default & disabled buttons
-      &:not(:disabled) {
-        + :disabled {
+      // Adds stripe above active button
+      &.iui-active {
```
bentleyvk, line 27: So this allows to have active input? 😁

src/button/button-group.scss
```diff
@@ -5,72 +5,79 @@
 @mixin iui-button-group {
   display: inline-flex;
   align-items: center;
+  isolation: isolate;
```
bentleyvk, line 8: Any alternatives to support IE?

src/button/button-group.scss
```diff
@@ -5,72 +5,79 @@
...
         }
       }
 
-      // Border between 2 side-by-side disabled buttons
-      &:disabled + .iui-button:disabled {
-        @include themed {
-          border-left-color: t(iui-color-background-4);
-        }
+      // Reset border radius only for non-borderless items
+      &:not(.iui-borderless) {
```
bentleyvk, line 51: Why not omiting borderless?

bentleyvk:
Why test images width changed?

backstop/tests/header.html
mayank99: split-menu uses button-group

src/button/button-group.scss
mayank99: "Supporting IE" does not mean everything has to work perfectly. This property is just to seal off the z-index. If the z-index is leaking in IE, I don't think it's a big deal as it's just on hover/focus.

src/button/button-group.scss
mayank99: 🤷‍♀️

src/button/button-group.scss
mayank99: borderless buttons should keep using their normal radius.

It's the same as before
image

mayank99:
> Why test images width changed?

It's because of the negative margin. I think that's fine
image

backstop/scenarios/button.js
```diff
@@ -21,8 +21,13 @@ module.exports = [
   }),
   scenario('Type Split Menu', {
     selectors: ['#demo-split-menu'],
+    misMatchThreshold: 0.03,
```
bentleyvk, line 24: Why these are needed here?

src/header/header.scss
```diff
@@ -314,18 +314,24 @@
 }
 
 @mixin iui-header-split-button {
```
bentleyvk, line 316: Ellipsis is not working.
image

bentleyvk: What the hell, before building I pulled latest changes, it showed that my sources are in sync and it had this image with broken ellipsis, then pulled again and it pulled newer changes with everything working... And I did everything this morning (in Lithuania time morning)...

backstop/scenarios/button.js
mayank99: The tests were failing without it, even with no visible change. I removed threshold and updated images, as the images were already changing.

For the ideas button I left it as I didn't want to update image

bentleyvk: Maybe update ideas button too and we can just remove increased threshold from there.

mayank99: ok

bentleyvk (approved):

src/button/button-group.scss
```diff
@@ -5,72 +5,79 @@
...
+        position: relative;
+        z-index: 1;
+      }
 
-    &.iui-default:not(:last-of-type) {
-      border-right: none;
-    }
+      &:disabled {
+        position: relative;
+        z-index: -1;
```
FlyersPh9, line 23: Is this `z-index` here & above so `:focus` appears above the disabled adjacent element?

mayank99: This is the thought train:

0. negative margins are being used, so by default an element does not have its right border visible
1. focus should be above siblings to show border/focus styles correctly despite negative margin
2. disabled elements should always be below non-disabled siblings regardless of focus, because non-disabled borders should take preference

FlyersPh9 (approved):

PR #367, opened 2021-11-04
feat(InformationPanel): Style inner content using grid

### LINKED ISSUE #292: Add css for info panel rows
We need to provide classes for info panel elements. Either, some container class to style and align our inputs or something more specific.

mayank99:
Added new iui-information-body-content class with iui-inline modifier. Uses css grid to allow toggling between stacked and two-column layouts for the label+input/text groups.

Closes #292

Also fixed some random bugs: added flex-shrink 0 on header icons, and max-width/max-height 100% on the panel to avoid resizing beyond parent.


src/information-panel/information-panel.scss
```diff
@@ -187,3 +189,34 @@
...
+    > *:nth-child(even):not(:last-child) {
+      margin-bottom: 11px;
+    }
+  }
+
+  &.iui-inline {
+    display: grid;
+    row-gap: 11px;
+    // grid-template-columns: auto 1fr;
+    grid-template-columns: auto auto;
```
mayank99, line 213: I am struggling with this part.

`auto 1fr` will make the input take up the entire remaining space.
image

`auto auto` will only make it take the least amount of space needed (even if the text inside the input is overflow).
image

@FlyersPh9 Previously you had this. Is there any logic or did you just find a random width that "looks nice"? 
image

FlyersPh9: Pretty sure I picked a value that worked well with the label lengths.

I like your first image as long as the 1st column is wide enough so none of the labels wrap to a 2nd line.

mayank99: No wrapping when there is space:
image

There is something going on in narrow widths though:
image

bentleyvk: Aren't labels too wide in screenshot from description? It looks like it takes 60-70% of width. Is it really bad if some of them are multiline?
image

FlyersPh9: > No wrapping when there is space:
> There is some (w)rapping going on in narrow widths though:

I just checked this locally, what you have now looks good for the editable demo.  The labels should be on 1 line / not wrapping in our default width.  But if the user sets the width extremely narrow, it makes sense to wrap the labels.

src/information-panel/information-panel.scss
```diff
@@ -187,3 +190,34 @@
...
+    }
+  }
+
+  &.iui-inline {
+    display: grid;
+    row-gap: 11px;
+    grid-template-columns: auto 1fr;
+    align-items: center;
+    place-content: end;
+    text-align: end;
```
FlyersPh9, line 216: This appears to be causing alignment issues in the non-editable information panel:
Screen Shot 2021-11-09 at 1 00 21 PM

src/information-panel/information-panel.scss
```diff
@@ -74,6 +75,7 @@
         width: $iui-icons-large;
         height: $iui-icons-large;
         margin-right: $iui-s;
+        flex-shrink: 0;
```
FlyersPh9, line 78: Line 73 is targeting SVG.  On a local copy I changed it to target `&-icon` and then added the class `.iui-information-header-label-icon` to the Icon in the header to fix `flex-shrink` not being applied correctly.  Screenshot prior to my change when the info panel is very narrow:
Screen Shot 2021-11-09 at 1 11 39 PM

FlyersPh9:
Can we set a `max-width` on the information panel?  Currently I can make the info panel wider than the parent element.  If I do that, I am unable to grab the resizer to fix it without refreshing.
Screen Shot 2021-11-09 at 1 16 24 PM

src/information-panel/information-panel.scss
mayank99: Adding a new class is not a proper fix, because this icon is passed by the user.

But this is because of webcomponents so we don't have this issue in React.

backstop/tests/information-panel.html
```diff
@@ -725,98 +690,118 @@ <h1>Information panel</h1>
...
+              <span class="iui-input-label">
+                Kind
+              </span>
+              <span>MP3 audio</span>
 
-            </table>
+              <span class="iui-input-label">
+                Size
+              </span>
+              <span>>9.7 MB</span>
```
FlyersPh9, line 702: We have an extra `>` in here.
```suggestion
              <span>9.7 MB</span>
```

src/information-panel/information-panel.scss
```diff
@@ -187,3 +190,38 @@
...
+  &:not(.iui-inline) {
+    // add bottom margin between pairs of label+input
+    > *:nth-child(even):not(:last-child) {
+      margin-bottom: 11px;
+    }
+  }
+
+  &.iui-inline {
+    display: grid;
+    row-gap: 11px;
```
FlyersPh9, line 212: Variable should work here:
```suggestion
    row-gap: $iui-baseline;
```

src/information-panel/information-panel.scss
```diff
@@ -187,3 +190,38 @@
...
+    font-weight: normal;
+    @include themed {
+      color: t(iui-text-color-muted);
+    }
+  }
+
+  &:not(.iui-inline) {
+    // add bottom margin between pairs of label+input
+    > *:nth-child(even):not(:last-child) {
+      margin-bottom: 11px;
```
FlyersPh9: Variable should work here:
```suggestion
      margin-bottom: $iui-baseline;
```

src/information-panel/information-panel.scss
mayank99: Changed

backstop/tests/information-panel.html
mayank99: You didn't see that 🤫

src/information-panel/information-panel.scss
FlyersPh9: Ah, ok.  Should we add the `flex-shrink` in the HTML's head to fix it for the demo page?  Or leave it as is?

mayank99:
> Can we set a `max-width` on the information panel? Currently I can make the info panel wider than the parent element. If I do that, I am unable to grab the resizer to fix it without refreshing.

Added `max-width: 100%`

FlyersPh9:
I am trying to vertically center align the user icon & text but failing to do so.  Here's what I tried:
```
&.iui-inline {
	…
	> *:nth-child(even) {
	      align-items: center;
	      vertical-align: middle;
	      display: table-cell;
	}
}
```
Screen Shot 2021-11-09 at 1 36 00 PM
I applied colors to see if the wrapping element was aligned, its only the content inside (user icon + text).

src/information-panel/information-panel.scss
mayank99: Fixed already

mayank99:
> I am trying to vertically center align the user icon & text but failing to do so. Here's what I tried:

I already fixed the alignment by adding some rules in the demo html. I don't think it should be part of the component css

src/information-panel/information-panel.scss
mayank99: I'd say leave it as is, it's not a big deal

FlyersPh9:
Open discussion on this part:  what are your thoughts on using the `<span>` instead of `<input readonly>`?  Visually it is easier to tell which text fields are editable VS not.  Another option might be removing the border from `<input readonly>`.

Screen Shot 2021-11-09 at 1 47 01 PM

👆 Bad example because you shouldn't be able to edit the Size, but you get the idea.

mayank99:
> Open discussion on this part: what are your thoughts on using the `<span>` instead of `<input readonly>`? Visually it is easier to tell which text fields are editable VS not. Another option might be removing the border from `<input readonly>`.

That does not change the component styling. It will work with literally anything. We just provide the layout.

FlyersPh9 (approved):

backstop/tests/information-panel.html
```diff
@@ -139,40 +143,28 @@
...
+        align-items: center;
+        gap: 4px;
       }
 
-      td {
-        white-space: nowrap;
-        overflow: hidden;
-        text-overflow: ellipsis;
+      /* Removes border and padding from readonly text fields */
+      input[readonly] {
```
bentleyvk, line 153: What if we handle that in CSS? Or it might be unwanted styling for the users?

src/information-panel/information-panel.scss
```diff
@@ -187,3 +192,34 @@
     top: $iui-m * -0.5;
   }
 }
+
+/// Labels and inputs aligned using css grid
+@mixin iui-information-body-content {
+  .iui-input-label {
+    font-weight: normal;
```
bentleyvk, line 199: Don't we have a variable for that? Because regular is `400`.

src/information-panel/information-panel.scss
```diff
@@ -187,3 +192,34 @@
...
+    @include themed {
+      color: t(iui-text-color-muted);
+    }
+  }
+
+  &:not(.iui-inline) {
```
bentleyvk, line 205: When it is applied? I can't find it in demo page.

mayank99: Ah good catch. I was going to toggle this class using JS

backstop/tests/information-panel.html
mayank99: Yeah I would like to leave this component to be mostly layout

src/information-panel/information-panel.scss
mayank99: what's the benefit?

bentleyvk: In case we decide to change it.

FlyersPh9: I agree with Vilius here.

Also, I read somewhere (can't find the article again) that you should use numbers instead of values like `bolder` because `bolder` may be different across devices / browsers.

mayank99: > Also, I read somewhere (can't find the article again) that you should use numbers instead of values like bolder because bolder may be different across devices / browsers.

I have not been able to find a source for that.

But changed

src/information-panel/information-panel.scss
mayank99: Added JS

PR #370, opened 2021-11-08
fix(Tabs): use `::after` pseudo-element for tab stripe

mayank99:
Replaced `iui-tab-stripe` with `::after` pseudo-element applied on `ul.iui-tabs`.

Using a pseudo-element ties the positioning to the main element, and fixes random bugs that break stripe positioning in scenarios like scrolling (e.g. iTwin/iTwinUI-react#421).

Before:
image

After:
image


src/tabs/tabs.scss
```diff
@@ -177,38 +177,41 @@
...
       }
     }
   }
 
   @at-root {
-    .iui-vertical & ~ .iui-tab-stripe {
+    .iui-vertical &::after {
+      left: calc(100% - #{$iui-xxs});
       width: $iui-xxs;
+      top: var(--stripe-top);
```
bentleyvk, line 213: And it won't work on IE.

mayank99: Yes. For IE, we have `iui-not-animated`

bentleyvk: We will check whether CSS variables are supported and set that class then?

mayank99: In React? Yeah we can do that. Or we could leave it upto the developer

veekeys: Sure we can leave, but then.. I do feel we need to start thinking about either readme for every component or something else.. Cause going through source codes and trying to figure this out is really not a nice experience.

FlyersPh9: [Only 7 months of life support left on IE11.](https://death-to-ie11.com/)  But seems like we have the not animated class as an alternative solution.

FlyersPh9:
The stripe animation doesn't appear to be working for me.  I'm using the [latest version of Brave](https://brave.com/latest/).
Screen Recording 2021-11-09 at 12 33 02 PM

mayank99:
> The stripe animation doesn't appear to be working for me. I'm using the [latest version of Brave]

Fixed

FlyersPh9 (approved):

PR #372, opened 2021-11-08
style: Add prettier to lint-staged hook

mayank99:
Because we have lint-staged in this repo, we can control the order in which prettier and stylelint are run.

I updated lint-staged to the latest version and added prettier-cli. After that I ran linter on all files to fix the remaining formatting issues.

From this point on, we won't have any formatting discrepencies in scss files, as every commit will run prettier and stylelint (in order).


bentleyvk:
Will Prettier extension in VS Code properly format code on save?

mayank99:
> Will Prettier extension in VS Code properly format code on save?

No. And there is no way around it, from what I can tell. Adding format on save to vscode overrides stylelint.

FlyersPh9 (approved):

veekeys (approved):

PR #373, opened 2021-11-08
fix(Header): Improve styling of header buttons

### LINKED ISSUE #376: Active Header action has no underline
Both HeaderButton & IconButton components have no underline when `isActive={true}` is set. 

#### Expected behavior
Active item should be underlined same as it would be underlined using HeaderButton in breadcrumbs.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/itwinui-react-active-header-element-highlight-issue-1024k

##### Steps to reproduce
1. Use HeaderButton/IconButton as Header actions.
1. Set any of them as active.

#### Additional information
Used this as a temp workaround:
```
.iui-button.iui-active,
.iui-header-button.iui-active {
  box-shadow: 0 2px 0 0 var(--iui-color-foreground-primary);
}
```

mayank99:
1. Moved `iui-header-buttons` mixin out from `iui-left`. This allows header buttons to be used in `iui-right` with the active stripe (fixes #376)
2. Added ability to render header-logo as `button` or `a` element.
3. Added explicit `content-box` sizing to `iui-left` so that it doesn't get reset by user's stylesheet.
4. Fixed random bugs in split header button re: responsive design and disabled icon fill

Added scenarios and images for visually testing responsive design.


src/header/header.scss
```diff
@@ -13,8 +13,7 @@
   user-select: none;
```
bentleyvk, line 13: Could you undo formatting because it is hard to review?

mayank99: That's why I created #372

mayank99: Done. Also I suggest hiding whitespace changes in git diff

src/header/header.scss
```diff
@@ -161,45 +162,62 @@
     @include iui-header-button-icon;
   }
 
-  .iui-header-button {
+  .iui-header-button.iui-header-button { // boost specificity
```
bentleyvk, line 165: Why this change is needed?

mayank99: Because I moved the class out of `iui-left`, the specificity was lowered and was breaking styling, so I had to boost it back up

src/header/header.scss
```diff
@@ -119,9 +120,9 @@
 
   @media (max-width: 768px) {
     &,
-    &.iui-slim.iui-slim {
+    &.iui-slim {
       .iui-left {
-        .iui-header-button {
+        .iui-button {
```
bentleyvk, line 125: Why changed to target all buttons?

mayank99: I noticed it was not working for split header button (which does not use `iui-header-button` class)

bentleyvk: Maybe add that to test HTML.

mayank99: huh? it's already there

bentleyvk: Then why I don't see any difference? 😁

mayank99: this is inside a media query, it only happens on small screens

src/header/header.scss
FlyersPh9: Will this boost in specificity be needed after #363 passes?

FlyersPh9:
Glad you were able to fix the disabled dropdown arrow bug, I tried fixing that a couple weeks back and was unsuccessful.

FlyersPh9 (approved):

src/header/header.scss
mayank99: Hopefully not, but we'll see

PR #375, opened 2021-11-09
feat: Add shadows on scrollable areas

saaaaaally:
Created `iui-scroll-shadow` mixin that can be applied to show a shadow over scrollable areas.

- Scroll shadow should not be visible if there is not enough content to scroll/scrollbar is not visible. 
- Scroll shadow should be hidden if you are at the top/bottom since there is no more content to scroll to.
- Use `$background-color` to ensure that scroll shadow shows correctly in dark mode, on disabled content, or if content has a different background color than the default. 

shadowOverTable
menuScroll


backstop/tests/information-panel.html
```diff
@@ -1261,127 +1261,148 @@ <h1>Information panel</h1>
...
+            style='overflow: hidden; position: relative;'
+          >
+            <div
+              class="iui-scroll-shadow iui-shadow-top"
+              style="position: absolute;
+    width: 100%;
+    height: 11px;
+    top: 0;
+    background: linear-gradient(180deg, rgba(0, 0, 0, 0.35), transparent);
+    opacity: 0;"
```
saaaaaally, line 1275: @FlyersPh9 Not sure how to get the `iui-scroll-shadow iui-shadow-top` mixin to work here

saaaaaally:
Not sure why tests are failing, `yarn test` ran successfully for me

src/style/mixins.scss
```diff
@@ -171,3 +171,28 @@
...
+        transparent calc(100% - #{$iui-s})
+      ),
+      // Cover top shadow
+      linear-gradient($background-color 0% 100%) center top,
+      // Cover bottom shadow
+      linear-gradient($background-color 0% 100%) center bottom,
+      // Top shadow
+      linear-gradient(180deg, rgba(0, 0, 0, t(iui-opacity-6)) 20%, $background-color) center top,
+      // Bottom shadow
+      linear-gradient(0deg, rgba(0, 0, 0, t(iui-opacity-6)) 20%, $background-color) center bottom;
```
saaaaaally, line 192: The top/bottom shadow would look much better with radial-gradient like in https://codepen.io/chriscoyier/pen/YzXBYvL?editors=1100 but I couldn't figure out how to get the first `linear-gradient`  (used to maintain the background color on the whole component) to work correctly with this. Since it covers everything except the bottom/top 8px (where the scroll shadow is shown), with the radial gradient it doesn't go in the corners so the background color gets messed up in the bottom/top corners. 

image

mayank99: What have you tried?

I was able to change the body background to `grey` in that codepen and had to make the same change in the linear gradient
image
image

mayank99: ok I see the problem, it's because the background color and the gradients are applied on the same element. This can be solved by a wrapper div or maybe even a pseudo element

FlyersPh9: I started exploring our alternative options.  I'm thinking we might want to take [a JS based approach](https://javascript.plainenglish.io/how-to-create-horizontal-vertical-scroll-shadows-26aa7a7f00a0#4203).  The reason I'm thinking the CSS only approach won't work is because the scroll shadow appears underneath UI elements.  
_In the screenshots below I turned the shadow opacity up very high so you can better see the issues._
Screen Shot 2021-12-07 at 12 03 07 PM

What I'm thinking might be better is to use `::before` & `::after` to create a block at the top and the bottom of the scrollable area.  I have started exploring this option, but I can't get the blocks to stay at the top and bottom, they want to scroll.
Screen Recording 2021-12-07 at 12 09 23 PM
Here's the SCSS I have right now:
```SCSS
@mixin iui-scroll-shadow($background-color: t(iui-color-background-1)) {
  position: relative;

  &::before,
  &::after {
    content: '';
    position: absolute;
    left: 0;
    right: 0;
    width: 100%;
    height: $iui-baseline;
    pointer-events: none;
    background-repeat: no-repeat;
    background-size: 100% $iui-s;
  }

  &::before {
    top: 0;
    bottom: auto;
    @include themed {
      background: radial-gradient(farthest-side at 50% 0, rgba(0, 0, 0, t(iui-opacity-5)), rgba(0, 0, 0, 0)) center top;
    }
  }

  &::after {
    top: auto;
    bottom: 0;
    @include themed {
      background: radial-gradient(farthest-side at 50% 100%, rgba(0, 0, 0, t(iui-opacity-5)), rgba(0, 0, 0, 0)) center bottom;
    }
  }
}
```

I'm thinking the JS method might be best if we turn on / off the `::before` & `::after` depending on where the scroll bar is.  Here's [a demo I've been looking at](https://docs.telerik.com/kendo-ui/knowledge-base/grid-add-scroll-shadow) that seems to be working.

mayank99: I think with the current wrapper div approach, the scroll shadow still appears underneath UI elements like inputs. Using a pseudo-element would allow placing it on top of all content

saaaaaally: I just pushed my attempt at adding the scroll shadows with a wrapper div using `::before` & `::after` like in the link Jon posted, but it is not a one size fits all solution. 
It works in information panel and menu:
scrollOnInfoPanel
scrollOnMenu
But I was not able to get it to work on Table or textarea.
In table the shadows are added and removed correctly but they do not stay at the top/bottom. 
scrollOnTable
In textarea the layout gets messed up and the bottom shadow doesn't work.
scrollOnTextarea

FlyersPh9:
We're failing to achieve the goal of this PR in an easy to implement way so we will be closing this with plans to revisit if we have a slow day.

PR #377, opened 2021-11-10
feat(Carousel): Add carousel component

### LINKED ISSUE #374: Dots navigation component
#### Feature
Dot navigation component need to operate slideshows or to scroll to different page sections
This component used in ConstructSim project

#### Examples
dotsNavigation

FlyersPh9:
- Adding carousel component.
- Dots can shrink in size when there are many slides present.  _Not shown on the demo page._
- Slides allow for custom interior content.
- Following w3 recommendations for accessibility.  Inspiration came from [here](https://www.w3.org/TR/wai-aria-practices/examples/carousel/carousel-1.html) & [here](https://www.w3.org/WAI/tutorials/carousels/working-example/).
- Left & right buttons are working in this demo.  Dots will change color to match the active slide but are not clickable in the demo.

Closes #374.  Dots are based off [this provided code](https://github.com/iTwin/iTwinUI/issues/374#issuecomment-965345123).
iTwinUI_carousel_Type_default_0_demo-default_0_
iTwinUI_dark_carousel_Type_default_0_demo-default_0_


FlyersPh9:
Unsure of what we call this component.  Currently I have it in `/carousel/` but I don't know.

src/carousel/carousel.scss
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-carousel {
+  @include iui-reset;
+  display: flex;
+
+  .bnt-hc-scroll-container {
```
veekeys, line 9: What is this class name? :D

FlyersPh9: It's the classnames Alexandr was using.  I didn't bother changing the class names until we figure out what we want to call this component.  Any suggestions?  Carousel seems like the whole component (dots, arrow buttons, sliding part).  Maybe we call it `.iui-carousel-navigation`?  I could maybe use the `<nav>` then?

veekeys: I would have expected to have full carousel component with sliding animation of the elements. I do not really get only the dots part. Especially, when we have that document already in ux standards (welcome page).

mayank99: the dots are used in both loading page and welcome page, which look different.

| loading | welcome |
| --- | --- |
|  |  |

I think it should be extracted out into something reusable, since it could be used in other places and it has so much logic in it anyway (both CSS and JS).

veekeys: I find both loading and welcome page carousels same component.

bentleyvk: I wanted to say that loading, welcome and Alexandr's gallery components are not the same. But thought that using composition in React for bottom content, I think it is possible to achieve all of them.

FlyersPh9:
One issue I am aware of is the tabindex.  If I use the tab key the focus will go to slide 2 before it goes down to the left / dots / right navigation area.  Unsure if this is something that can be fixed with JS or not.
focus-in-carousel

FlyersPh9:
I am also unsure how to make the left / right arrows work, I'm thinking it will require JS.

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+  flex-direction: column;
+  width: 100%;
+}
+
+@mixin iui-carousel-slider {
+  position: relative;
+  display: flex;
+  width: 100%;
+  @supports (aspect-ratio: 16 / 9) {
+    aspect-ratio: 16 / 9;
```
FlyersPh9, line 18: When I run tests on my computer, `aspect-ratio` works, but when tested on GitHub the `aspect-ratio` doesn't work and test images fail.  How should I proceed, set a fixed width & height in the demo HTML?  I dont want to set a fixed width / height, I'd like for the user to set that as needed.

src/classes.scss
```diff
@@ -39,3 +39,4 @@
 @import './tree/classes';
 @import './user-icon/classes';
 @import './wizard/classes';
+@import './carousel/classes';
```
saaaaaally, line 42: This needs to be alphabetized. Same in index.scss

FlyersPh9: Alphabetized both classes and index in the next push.

src/carousel/carousel.scss
veekeys: Locally, i have same failures with tests.
I have tried to update backstopJs version (with the idea, that there is gonna be newer Chrome version), and seems that tests started to pass. So I think issue is of Chrome version, where aspect-ratio is experimental feature.
Just after udpating versino of backstop, I got over 70 failures on other components :D So not sure...
Let wait until focus PR is completed, if that could be done pretty soon ( I think there version of backstop is updated..)

backstop/tests/carousel.html
```diff
@@ -0,0 +1,355 @@
...
+    <script src="../assets/theme.js"></script>
+  </head>
+  <body class="iui-body">
+    <theme-button></theme-button>
+    <h1>Carousel</h1>
+    <hr />
+
+    <h2>Default</h2>
+    <section id="demo-default">
+      <div
```
veekeys, line 71: I would suggest wrapping carousel in some smaller div.. Cause now seeing it full screen is a little bit too big and navigation is not even seenable

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../text/index';
+
+@mixin iui-carousel {
+  @include iui-reset;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
```
veekeys, line 10: Is it required?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+}
+
+@mixin iui-carousel-slider {
+  position: relative;
+  display: flex;
+  width: 100%;
```
veekeys, line 16: this needed?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+
+@mixin iui-carousel {
+  @include iui-reset;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+}
+
+@mixin iui-carousel-slider {
+  position: relative;
```
veekeys, line 14: Any reason you remember for adding this?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+          background-color: rgba(0, 0, 0, t(iui-opacity-3));
+        }
+      }
+    }
+  }
+}
+
+@mixin iui-carousel-navigation {
+  display: flex;
+  justify-content: space-between;
```
veekeys, line 86: If you are using flex-grow on the container, this might be not needed.

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+    flex-grow: 1;
+    white-space: nowrap;
+    overflow: hidden;
+    justify-content: center;
+  }
+
+  &-dots {
+    @include iui-reset;
+    display: flex;
+    align-items: center;
```
veekeys, line 101: I think does nothing

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+    list-style: none;
+  }
+
+  &-dot {
+    @include iui-reset;
```
veekeys, line 106: does nothing on span and anchor elements

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+    width: $iui-baseline * 2;
+    height: $iui-baseline * 2;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
veekeys, line 111: shouldnt dot always have pointer cursor??

src/carousel/carousel.scss
```diff
@@ -0,0 +1,193 @@
...
+
+    &::after {
+      content: '';
+      display: block;
+      width: $iui-icons-small;
+      height: $iui-icons-small;
+      border-radius: 50%;
+      border: 1px solid transparent;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: background-color $iui-speed-fast ease, border-color $iui-speed-fast ease, width $iui-speed-fast ease,
```
veekeys, line 152: Where do you change border-color?

backstop/tests/carousel.html
```diff
@@ -0,0 +1,355 @@
...
+          >
+            <svg-chevron-left
+              class="iui-button-icon"
+              aria-hidden="true"
+            ></svg-chevron-left>
+          </button>
+          <div class="iui-carousel-navigation-container">
+            <div class="iui-carousel-navigation-counter">
+              <span class="iui-carousel-navigation-counter-current">4</span>
+              <span class="iui-carousel-navigation-counter-total">7</span>
```
veekeys, line 327: dont have rules for this class.. so maybe not needed? Or we can tackle first-child instead of having current class? Not sure..

backstop/tests/carousel.html
```diff
@@ -0,0 +1,355 @@
...
+                <span class="iui-carousel-navigation-dot iui-second"></span>
+              </li>
+              <li>
+                <span class="iui-carousel-navigation-dot"></span>
+              </li>
+              <li>
+                <span class="iui-carousel-navigation-dot"></span>
+              </li>
+              <li>
+                <span class="iui-carousel-navigation-dot iui-active"></span>
```
veekeys, line 249: these are not focusable, have no cursor pointer.. expected?
also clicking gets messy

image

FlyersPh9: The first carousel is using `<a>` so they have the hand cursor and are in the tabindex.  The second carousel is demonstrating that you can replace the `<a>` with a `<span>` if the user doesn't want the dots to be clickable.  The JS to change what dot is active is incorrectly applying to the `<span>`s as well, so that could be fixed.  I do have a warning on the demo page that says "The dots are not functional in this demo."

veekeys: But does it make sense to have dot not clickable?

veekeys: also.. off topic, but I just thought, that now your sliding in carousel is working through scroll, if I understood correctly..
I am just wondering, that to sync this together with the dots below using js (if you use touch to slide), might be more complicated...

src/carousel/carousel.scss
FlyersPh9: Removed.

src/carousel/carousel.scss
FlyersPh9: If the user wanted the dots to be non-interactable, the CSS only applys the pointer cursor when the dot is an `<a>`.

src/carousel/carousel.scss
FlyersPh9: I think that may be leftover from my initial attempt at building this.  I have removed border-color from the transition.

backstop/tests/carousel.html
FlyersPh9: Removed `.iui-carousel-navigation-counter-total` but left `.iui-carousel-navigation-counter-total` rather than looking for first child incase the user places anything else within the `.iui-carousel-navigation-counter`.

backstop/tests/carousel.html
FlyersPh9: I applied `.iui-carousel { max-width: 600px; }` in the demo's `<head>` styling, is that ok?

src/carousel/carousel.scss
FlyersPh9: Seems you are correct, removed.

src/carousel/carousel.scss
FlyersPh9: I removed it and it doesn't seem to be changing anything, so it is gone.

src/carousel/carousel.scss
FlyersPh9: Seems like it is not needed, removed.

src/carousel/carousel.scss
FlyersPh9: Seems like it is not needed, removed.

src/carousel/carousel.scss
veekeys: not sure, but anchor is not intuitive to me.. Usually anchor is used when you have a link.. Have you found this approach in some examples maybe? w3 shows only buttons as examples for navigation
https://www.w3.org/WAI/tutorials/carousels/functionality/

src/carousel/carousel.scss
FlyersPh9: On second thought, I removed the `aspect-ratio` completely.  I'd like to suggest that 16.9 aspect ratio be the suggested default, but if needed the user should be able to override that and some users may be confused how to remove the 16.9 ratio.

src/carousel/carousel.scss
FlyersPh9: Originally I was using `<button>` but I wasn't sure how to have it target an anchor like `#slide-1`.

FlyersPh9: I am not familiar with the `data-slide` in the W3 link you shared.  I can look into that when I get back in 2022.

backstop/tests/carousel.html
FlyersPh9: I am unsure if we want to have non-actionable dots or not.  We can force them to always be actionable if you'd like.

FlyersPh9: And to your second thought, I am very iffy about the overflow scroll implementation.  I discovered that if you click on the dots, the whole page scrolls down to the top of the carousel which probably isnt ideal.
carousel

FlyersPh9:
Also noticing that some of the tests are failing because it's lacking the puppeteer change in #166. 
Screen Shot 2021-12-30 at 11 33 03 AM

However, some are failing because of whatever this little stripe is.  I think this might be fixed if I adjust the threshold.
Screen Shot 2021-12-30 at 11 33 34 AM

backstop/tests/carousel.html
veekeys: looks much better.

veekeys: Just maybe isntead we could have a custom class or even apply through style not to confuse, which part is coming from our css, which is just for demo purposes?

backstop/tests/carousel.html
```diff
@@ -0,0 +1,270 @@
...
+      type="module"
+      src="https://cdn.skypack.dev/@itwin/itwinui-icons-elements/chevron-right"
+    ></script>
+    <script
+      type="module"
+      src="https://cdn.skypack.dev/@itwin/itwinui-icons-elements/pause"
+    ></script>
+    <script
+      type="module"
+      src="https://cdn.skypack.dev/@itwin/itwinui-icons-elements/play"
```
veekeys, line 77: Not used

backstop/tests/carousel.html
```diff
@@ -0,0 +1,270 @@
...
+                class="iui-carousel-navigation-dot"
+                data-slide="2"
+                aria-label="Slide 3"
+                id="test-2"
+              ></button>
+            </li>
+          </ol>
+
+          <!-- Replicates the width of the play/pause button to center the dots -->
+          <div style="width: 28px;"></div>
```
veekeys, line 211: huh, what looks hacky :/ So what is the idea if ours? We plan to allow users to pass anything in the carousel "footer"? If so, shouldnt we do something like we did with some components with 3 sections a.k.a left, center and right? I am not fan of that approach either, but I think it should be better than empty element to center things

backstop/tests/carousel.html
```diff
@@ -0,0 +1,270 @@
...
+      </div>
+
+    </section>
+
+    <script>
+      document.querySelectorAll('.iui-carousel').forEach((carousel) => {
+        const slides = [...carousel.querySelectorAll('.iui-carousel-slider-item')];
+        const [prev, next] = [...carousel.querySelectorAll('.iui-carousel-navigation-button')];
+
+        const updateSlideClasses = (current) => {
```
veekeys, line 234: Maybe do the same with the dots do? Not that much of the code (basically copy paste) and we could have fully functional carousel.

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../text/index';
+@import '../button/button';
+
+@mixin iui-carousel {
+  @include iui-reset;
+  display: flex;
```
veekeys, line 9: Does flex do here anything?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+  flex-direction: column;
+  position: relative;
```
veekeys, line 11: Why was needed?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+}
+
+@mixin iui-carousel-slider {
+  display: flex;
```
veekeys, line 15: Is it needed?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+  position: relative;
+  overflow: hidden;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+
+  &-item {
+    position: absolute;
+    margin-left: auto;
```
veekeys, line 24: why margins are needed if you already set the position and set its width?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+
+    &.iui-next {
+      transform: translateX(100%);
+    }
+
+    &.iui-current {
+      transform: translateX(0);
+    }
+
+    &-image {
```
veekeys, line 47: hm.. Sure we want to have classes for inner content like image and title? We could just leave content up to the user so he puts how ever he wants. Less code, easier react implementation. Opinions?
Cause they could decide to put image not as background, but as  and in title part might be something else, not necessarily as a title...

backstop/tests/carousel.html
```diff
@@ -0,0 +1,270 @@
...
+              style="background-image: url('../assets/image-test.jpg');"
+            ></div>
+            <div class="iui-carousel-slider-item-title">Bentley System headquarters</div>
+          </li>
+
+        </ol>
+
+        <nav class="iui-carousel-navigation">
+          <button
+            class="iui-button iui-borderless iui-small iui-carousel-navigation-button iui-carousel-navigation-previous"
```
veekeys, line 162: So specific nagivation button classes are not used anymore, can be removed.

backstop/tests/carousel.html
```diff
@@ -0,0 +1,270 @@
...
+            data-action="pause"
+          >
+            <svg-pause
+              class="iui-button-icon"
+              aria-hidden="true"
+            ></svg-pause>
+          </button>
+
+          <ol class="iui-carousel-navigation-dots">
+            <li>
```
veekeys, line 185: Lets have an example of 4 slides in carousel, just to be sure that everything is ok (as we have 3 classes for slides previous, current and next)

src/carousel/carousel.scss
veekeys: So needed?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+    list-style: none;
+  }
+
+  &-dot {
+    @include iui-button;
+    @include iui-button-size($size: small);
+    background-color: transparent;
+    outline: none;
+
+    &:hover:not(.iui-invisible):not(.iui-active)::after {
```
veekeys, line 104: This is massive.. sure cannot be simplified?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-3));
```
veekeys, line 106: Also, is it ok, that mouse does not need to come to dot and hover works? I know it wokrs like that in our buttons, but they have background color which appears and here it seems visually weird..

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
+
+    &.iui-invisible {
```
veekeys, line 110: We dont use it anymore.. Maybe just remove?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+    justify-content: center;
+    align-items: center;
+    list-style: none;
+  }
+
+  &-dot {
+    @include iui-button;
+    @include iui-button-size($size: small);
+    background-color: transparent;
+    outline: none;
```
veekeys, line 102: not needed

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+    flex-grow: 1;
+    white-space: nowrap;
+    overflow: hidden;
+    justify-content: center;
+    align-items: center;
+    list-style: none;
+  }
+
+  &-dot {
+    @include iui-button;
```
veekeys, line 99: Sure we need all of the rules of button if everything seems to be set with the `:after` rule?

src/carousel/carousel.scss
veekeys: cause iui-button mixin has it too

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+      transition: background-color $iui-speed-fast ease;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: background-color $iui-speed-fast ease, transform $iui-speed-fast ease;
+      }
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+      }
+    }
+
+    &.iui-first::after {
```
veekeys, line 146: all these not used too in demo, so for now remove?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+
+    &::after {
+      content: '';
+      display: block;
+      width: $iui-icons-small;
+      height: $iui-icons-small;
+      border-radius: 50%;
+      transition: background-color $iui-speed-fast ease;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: background-color $iui-speed-fast ease, transform $iui-speed-fast ease;
```
veekeys, line 139: why transition repeating...?

src/carousel/carousel.scss
mayank99: I agree, the CSS should be content agnostic

src/carousel/carousel.scss
```diff
@@ -0,0 +1,167 @@
...
+      transform: translateX(-100%);
+    }
+
+    &.iui-next {
+      transform: translateX(100%);
+    }
+
+    &.iui-current {
+      transform: translateX(0);
+    }
```
mayank99, line 45: You know, I'm thinking we might be able to improve this approach by applying transform the parent instead. That way we won't even need to use `position: absolute` and not need to modify every single item.

src/carousel/carousel.scss
mayank99: I don't think we should ever be using `outline: none` outside focus styles

we removed it from `iui-button` too

src/carousel/carousel.scss
FlyersPh9: Line 137 is the transition that is applied if the user has reduced motion set while line 138-140 is the transition that is applied if the user has motion enabled.  

Since changing the background color isn't motion, we thought it'd make sense to allow it to show for reduced motion users.  I know in other parts of iTwinUI we do have color changes within the reduced motion media query…we should probably fix that someday.

src/carousel/carousel.scss
FlyersPh9: `.iui-first`, `.iui-second`, and `.iui-invisible` would allow us to "truncate" the dots if more than X amount of slide are present.  #374 shows the truncation at the start and end of the dots.  The image below shows what it would look like if you were on slide 1 of 6 or more.
Screen Shot 2022-01-18 at 4 09 19 PM

The only reason they are unused in the demo is I don't know how to actually make them react to what slide is current.  I could include a demo to showcase how it's used, but you wouldn't be able to interact with it.

src/carousel/carousel.scss
FlyersPh9: I used the button mixin so that the dots width & height would match the left / right / pause buttons.  This way if we ever decided to change the button sizing, the dots would automatically change to match.  The `::after` is only used to place the dot.

src/carousel/carousel.scss
FlyersPh9: [See this comment.](https://github.com/iTwin/iTwinUI/pull/377#discussion_r787153663)

src/carousel/carousel.scss
FlyersPh9: The dots are 16x16 px which would be a pretty small touch target size.  It might be a little weird that you get the hover state when hovering a few pixels away from the dot, but i think the pros outweigh the cons.

src/carousel/carousel.scss
FlyersPh9: I believe I can remove `:not(.iui-invisible)` since `.iui-invisible`'s scale is at 0%.

src/carousel/carousel.scss
mayank99: that's correct, except we want to translate in the negative direction. These are the changes I had to make:
image
image

backstop/tests/carousel.html
FlyersPh9: Can do.

src/carousel/carousel.scss
mayank99: > Line 137 is the transition that is applied if the user has reduced motion set while line 138-140 is the transition that is applied if the user has motion enabled.

Technically, they are both applied when motion is enabled, the latter one just takes preference. But yeah that's the idea, and we should document this as a "best practice" somewhere, for future contributors

src/carousel/carousel.scss
mayank99: > Line 137 is the transition that is applied if the user has reduced motion set while line 138-140 is the transition that is applied if the user has motion enabled.

Technically, they are both applied when motion is enabled, the latter one just takes preference. But yeah that's the idea, and we should document this as a "best practice" somewhere, for future contributors

backstop/tests/carousel.html
mayank99: I would even suggest lets have 10 or 20 slides (we could generate them using JS, and even remove the weird Bentley-specific hardcoded content). It would help see the first/second/invisible dots in action too.

backstop/tests/carousel.html
FlyersPh9: I have removed the empty DIV and am instead doing:
```scss
.iui-carousel-pause-button ~ .iui-carousel-navigation-button {
	margin-left: $iui-m + ($iui-xxs * 6);
}
```

Math is weird because we need it to equal 28px (same as the left / right / pause buttons).

src/carousel/carousel.scss
FlyersPh9: I thought it was necessary for `flex-direction: column;` to work, but it would appear it is not.  Removed.

src/carousel/carousel.scss
FlyersPh9: It is not, removed.

src/carousel/carousel.scss
FlyersPh9: Seems like it was not needed, removed.

src/carousel/carousel.scss
FlyersPh9: Seems it was not needed, removed.

backstop/tests/carousel.html
mayank99: I don't think this is necessarily "hacky". The idea of "spacer" components is quite popular: https://www.joshwcomeau.com/react/modern-spacer-gif/

In React, I would expect a structure like this:
```tsx
<CarouselNav>
  <CarouselPrev />
  <CarouselDots />
  <CarouselNext />
</CarouselNav>
```

The user could easily add stuff in between those children.

src/carousel/carousel.scss
FlyersPh9: Removed `&-image` and `&-title`, replaced the slide content with gradients from https://uigradients.com.

backstop/tests/carousel.html
FlyersPh9: Was able to remove `.iui-carousel-navigation-previous` & `.iui-carousel-navigation-next`.

src/carousel/carousel.scss
FlyersPh9: I've removed this line.

src/carousel/carousel.scss
FlyersPh9: It only becomes necessary if someone forgets to apply `.iui-small` on the next / previous / pause buttons or if a user places custom content inside that is taller.

Screen Shot 2022-01-18 at 5 15 25 PM

backstop/tests/carousel.html
FlyersPh9: It is now. 😆 The play / pause button will toggle when clicked (although it appears the button width is being effected because technically there are 2 elements within the icon, right click and inspect the margin on the icon compared to the next / previous buttons).

backstop/tests/carousel.html
FlyersPh9: Dots now update to show which slide is active.  I haven't tried making the dots clickable yet.

backstop/tests/carousel.html
FlyersPh9: If we agree to do that, I would definitely need help implementing something like that.

src/carousel/carousel.scss
FlyersPh9: Implemented the code & got the dots to change active state, check it out.

src/carousel/carousel.scss
FlyersPh9: Had to add it back in to get [the new transition method](https://github.com/iTwin/iTwinUI/pull/377#discussion_r787163634) to work properly.

backstop/tests/carousel.html
mayank99: Made the dots clickable in [`39ea6ce`](https://github.com/iTwin/iTwinUI/pull/377/commits/39ea6ce28f838987cc0bedf9718244926f178d31)

backstop/tests/carousel.html
mayank99: Oh wow, I didn't realize we were using `:only-child` in our CSS.

Updated code in [`bbccbfd`](https://github.com/iTwin/iTwinUI/pull/377/commits/bbccbfdf596c95c89f3e05210e4fce42873c2bb7) to only have one SVG in the button.

src/carousel/carousel.scss
mayank99: Pushed some JS in 
[`cd5c898`](https://github.com/iTwin/iTwinUI/pull/377/commits/cd5c898c487bdb241a5ea8f3f850c916ec906bc9) to handle dots truncation. It seems to "work" for odd numbers, but looks weird when the selection stays in the middle.

Also temporarily added `display: none` until we figure out the focus situation with invisible dots. W3's example shows that selection follows focus (using left/right arrow keys), so maybe we'll never have to focus invisible dots.
https://www.w3.org/TR/wai-aria-practices-1.2/examples/carousel/carousel-2-tablist.html

backstop/tests/carousel.html
```diff
@@ -215,95 +98,145 @@ <h2>Default</h2>
 
           <button
             class="iui-button iui-borderless iui-small"
+            type="button"
             aria-label="Pause animation"
-            data-action="pause"
```
gretanausedaite: This change breaks the script bellow. Change data-action back or update script.

FlyersPh9: Thank you, this is fixed now.

src/carousel/carousel.scss
veekeys: I was able to remove .iui-active too and reduce specificity from 0.2.1 to 0.1.1
Try it at your end. If it works same, commit that change.
It works, because we have iui-active::after rule below, which has same specificity.

src/carousel/carousel.scss
veekeys: So if we care only about sizing, shouldnt mixin `iui-button-size($size: small)` be enough?

src/carousel/carousel.scss
veekeys: For me navigation is buggy.. Seems active class is jumping between dots...

backstop/tests/carousel.html
veekeys: Im going to resolve this as now we have more than 3... 
But js is bugged and now it works weirdly, so keeping comment https://github.com/iTwin/iTwinUI/pull/377#discussion_r789997143 open

src/carousel/carousel.scss
veekeys: So now instead, you have changed HTML layout and putting nav first and then slides. For that, you are using flex and column-reverse direction. I am never a fan of inverting the html order unless there is a reason.
So what was the reason to do that?

backstop/tests/carousel.html
veekeys: and who would align all of the custom content in there. We or user? In the example I am not sure I understand where the spacer would go and how we would even make reusable (for different size custom elements) ?

backstop/tests/carousel.html
veekeys: amm, yes, clicking the button toggles between play and pause, but technically it does nothing. What are we trying to show here?
Id say just leave single element if we want to show that any element can be put additionaly :/

src/carousel/carousel.scss
```diff
@@ -0,0 +1,117 @@
...
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+      }
+    }
+
+    &.iui-first::after {
+      transform: scale(0.5);
+    }
+
+    &.iui-second::after {
```
veekeys, line 100: Do you think we could rename first and second into something else? Cause these dots are at the end, they are inverted order :D all they differ is in size, so maybe could figure some fancy wording? 
If not, its ok, minor

src/carousel/carousel.scss
```diff
@@ -0,0 +1,117 @@
...
+    background-color: transparent;
+
+    &:hover:not(.iui-active)::after {
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-3));
+      }
+    }
+
+    &.iui-invisible {
+      cursor: default;
```
veekeys, line 60: Just curiuos, but does it do anything?

src/carousel/carousel.scss
FlyersPh9: [w3 documentation](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-4) says:
> If present, the rotation control is the first element in the Tab sequence inside the carousel. It is essential that it precede the rotating content so it can be easily located.
By reordering the HTML structure, if you hit tab it will target the back arrow, pause, dots, right arrow before it targets anything within the slides themselves.

nigtas: Thats nice, but.. Rotation control is not the same as Slide picker controls.
image

So in our case, this should be only pause/start button, which is fully custom and provided by user.

veekeys: Thats nice, but.. Rotation control is not the same as Slide picker controls.
image

So in our case, this should be only pause/start button, which is fully custom and provided by user.

mayank99: I believe this should be the order:
1. rotation control ("play"/"pause")
2. slide controls ("prev"/"next"/"dots")
3. slides (i.e. focusable elements inside the slides)

src/carousel/carousel.scss
mayank99: what do you mean "jumping between dots"?

src/carousel/carousel.scss
FlyersPh9: Probably not with `display: none` right underneath it.  Removed.

src/carousel/carousel.scss
FlyersPh9: If I remove `@include iui-button;` we get something like this:
Screen Shot 2022-01-24 at 2 48 38 PM

So I need to add:
```scss
border: none;
cursor: pointer;
```

The button mixin included a 1px transparent border which serves us no purpose here, so test images are changing ever so slightly on my last push.

src/carousel/carousel.scss
FlyersPh9: You are correct, pushed the change.

src/carousel/carousel.scss
veekeys: Well, we do not provide special case for play/pause button. That one is user provided, if needed. I guess we could have documentation about it (that would be really nice).
Even though w3 does not say anything about controls and order, I guess from this point of view it might make sense to have those first. I will resolve this.

src/carousel/carousel.scss
veekeys: Notice this weird flickering when dots get smaller around
carousel

src/carousel/carousel.scss
FlyersPh9: I proposed `.iui-small`, `.iui-smaller`, & `.iui-smallest` instead of `.iui-second`, `.iui-first`, & `.iui-invisible` but it wasn't received well.

Another idea is applying `.iui-truncate-start` and / or `.iui-truncate-end` on the `<ol>` to target the nth child, but the hidden `<li>` are still in the DOM so that would require some re-working.

src/carousel/carousel.scss
mayank99: that's not a bug, it's designed like that and that's exactly what I was talked about when I said "looks weird when the selection stays in the middle".

If there was absolutely no motion, then it would look even more weird because the slide would change but there would be no feedback on the clicked element.

I think we need some animation (maybe using react-transition-group) to show the dots moving

backstop/tests/carousel.html
FlyersPh9: Reworked the navigation area so the dots are always centered no matter how lopsided the exterior content is, no spacer `<div>` or margin required.

src/carousel/carousel.scss
FlyersPh9: Latest push now targets play/pause first.

src/carousel/carousel.scss
veekeys: @FlyersPh9 had this short video of nice animation, so the only question is, if there is something more/less we need from css to achieve that in react.
But I am resolving this one then..

backstop/tests/carousel.html
veekeys: Looks good, but now we have a special class for animation button.. Which has single rule for order.. That looks specific and means we want that button to be only there, nowhere else.. ANd that means, we need to have react api for that button?
idk, dont like only this part :D

src/carousel/carousel.scss
```diff
@@ -0,0 +1,129 @@
...
+
+    &::after {
+      content: '';
+      display: block;
+      width: $iui-icons-small;
+      height: $iui-icons-small;
+      border-radius: 50%;
+      transition: background-color $iui-speed-fast ease;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: background-color $iui-speed-fast ease, transform $iui-speed-fast ease;
```
mayank99, line 93: I think this second transition (`transform`) needs to be applied on `::after` because that's where the scaling is happening

src/carousel/carousel.scss
veekeys: amm. I would stick with having classes better.. looks much simpler...But cannot propose anything better than `small` namings...
Another question then, do all dots need to be in the dom, if they are invisible? Shouldnt be enough to render only those visible? Is there something about that in w3? Cause then maybe we could simplify css selectors even without classes...

src/carousel/carousel.scss
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  &-item {
+    width: 100%;
+    flex-shrink: 0;
+    box-sizing: border-box;
+  }
+}
+
+@mixin iui-carousel-navigation {
```
mayank99, line 29: I think this whole navigation needs some focus styling (and tabindex).

I imagine a keyboard user will want to tab into the carousel in the following sequence:
1. get to the play/pause button first so that autoplay can be paused (using enter key)
2. tab to the slide control and press left/right arrow keys to change slides. They will _not_ want to focus the actual dots or the prev/next buttons if they can just use arrow keys.
3. tab to the interactable content in the current slide.

src/carousel/carousel.scss
```diff
@@ -0,0 +1,129 @@
...
+  display: flex;
+  align-items: center;
+  height: $iui-baseline * 3;
```
mayank99, line 32: Doesn't this need border-box for the height to be accurate?

backstop/tests/carousel.html
mayank99: Maybe it makes sense to expose a React API for it, since the autoplay feature would require knowledge of our transforms.

src/carousel/carousel.scss
veekeys: I dont know.. I would say, if I tab into Carousel, Enter/Space should trigger play/pause. Then left/right arrows should switch slides. And then additional tab to get into carousel content. That intuitive to me, but need to read w3 recommendation on this one, if they have it :/

mayank99: W3C does not even recommend using prev/next buttons together with dots, so our situation is a bit weird. But they do recommend the arrow keys https://www.w3.org/TR/wai-aria-practices-1.2/examples/carousel/carousel-2-tablist.html

veekeys: ahh.. getting weird..
But current solution, when focus is on play button and pressing tab moves you to the prev arrow feels wrong.

src/carousel/carousel.scss
mayank99: I think the accessibility benefit here would be that the screen reader would announce how many dots are in the list.

It might also help to keep them in the DOM for animation purposes (if we want to add animation)

backstop/tests/carousel.html
veekeys: Why? I thought users just gonna switch the `activeSlide` prop and we would have animation, no??

veekeys: But I start to think us having implemented auto play function is better.. Just I really thuoght we could give user a solution to put that button (or what ever else) whenever he needs.. But maybe we can just have a flag for it and have a specific place, not sure anymore :/

src/carousel/carousel.scss
veekeys: Yeah, and then someone comes in with a lot of images and we render lots of empty items :)
I am not that sure about it.. Maybe there are some attributes to say the amount instead..

mayank99: we would still have lots of slides in the DOM, which I think is a bigger problem 😆

veekeys: I assumed we gonna have just 3 slides and switch content inside :D

src/carousel/carousel.scss
FlyersPh9: Maybe we put `tabindex="-1"` on the left / right buttons since the dots are going to use the left / right arrow keys similar to paginator.  At that point, the dots & left / right buttons would be doing the same thing pretty much, right?

veekeys: One of the options, I guess.. Just paginator left/right buttons are also focusable by tabbing into them..

src/carousel/carousel.scss
FlyersPh9: > I assumed we gonna have just 3 slides and switch content inside :D

Wouldn't that be an issue if you are at the last slide & click the next button which sends you back to the very first slide (animating back past all of the previous slides)?

veekeys: oh, I thought you cannot go beyond last slide :O Is this recommended UX?

FlyersPh9: We don't have clear guidance on if looping back from the last slide to the first slide is good UX or not.  w3 does do this though.

I could see home page carousels / image carousels finding this useful, but I could also find this being bad for a welcome / onboarding tutorial.  Maybe we give this as an option within React?

src/carousel/carousel.scss
FlyersPh9: Good catch, added.

backstop/tests/carousel.html
```diff
@@ -0,0 +1,471 @@
...
+        aria-live="off"
+        aria-atomic="true"
+      >
+        <nav class="iui-carousel-navigation">
+          <div class="iui-carousel-navigation-left">
+            <button
+              class="iui-button iui-borderless iui-small iui-carousel-navigation-animation-button"
+              type="button"
+              aria-label="Pause animation"
+              aria-pressed="true"
```
mayank99, line 92: I think you don't need aria-pressed here because you have aria-label.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed#:~:text=don't%20use%20aria-pressed

backstop/tests/carousel.html
```diff
@@ -0,0 +1,471 @@
...
+            >
+              <svg-pause
+                class="iui-button-icon"
+                aria-hidden="true"
+              ></svg-pause>
+            </button>
+            <button
+              class="iui-button iui-borderless iui-small iui-carousel-navigation-button"
+              type="button"
+              aria-controls="myCarousel-items"
```
mayank99, line 103: I think this `aria-controls` is wrong. My understanding is that `aria-controls` should have a 1:1 relationship between the toggle and the element (e.g. dot-1 controls slide-1).

backstop/tests/carousel.html
```diff
@@ -0,0 +1,471 @@
...
+                class="iui-button-icon"
+                aria-hidden="true"
+              ></svg-chevron-left>
+            </button>
+          </div>
+
+          <ol
+            class="iui-carousel-navigation-dots"
+            role="tablist"
+            aria-controls="myCarousel-items"
```
mayank99, line 117: This `aria-controls` also looks wrong because the list itself is not controlling anything. Though I'm less sure about it because we made it focusable and decided to add keyboard handling

mayank99 (commented):
Not sure about some of these aria attributes

backstop/tests/carousel.html
mayank99: I removed it

src/carousel/carousel.scss
mayank99: > I would say, if I tab into Carousel, Enter/Space should trigger play/pause. Then left/right arrows should switch slides. And then additional tab to get into carousel content.

I implemented this approach so the user can now focus the whole carousel instead of having to tab into those individual pieces (which all have tabindex=-1 now). See [`0be5873`](https://github.com/iTwin/iTwinUI/pull/377/commits/0be587392c44815404a5233d50743a5b94733f11)

mayank99: @FlyersPh9 I think we could change the ordering back to normal now?

backstop/tests/carousel.html
FlyersPh9: [This w3 example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/carousel/carousel-1-prev-next.html#sc1_label) does have left / right buttons with `aria-controls` equal to the slides as a whole.

backstop/tests/carousel.html
FlyersPh9: This one I am unsure about.  [w3](https://www.w3.org/TR/wai-aria-practices-1.2/examples/carousel/carousel-2-tablist.html) does not have it present on theirs, so I guess I'll remove it.

src/carousel/carousel.scss
FlyersPh9: HTML structure is back to the way it was previously.

FlyersPh9:
Added padding to the `<section>` so tests can see the offset outline during `:focus`.

src/carousel/carousel.scss
```diff
@@ -26,24 +24,28 @@
     flex-shrink: 0;
     box-sizing: border-box;
   }
+
+  ~ .iui-carousel-navigation {
```
mayank99, line 28: Why is the slider controlling the styles of the nav? Why not leave it in the navigation mixin? This is a bad pattern which kinda breaks the whole point of using mixins by tightly coupling the styles to the class names.

Also do we even need this border? Looks really weird to me

src/carousel/carousel.scss
```diff
@@ -112,10 +114,8 @@
     }
   }
 
-  &-animation-button {
-    order: 1;
-
-    &[data-pressed="true"] {
+  &-animation-button[data-pressed='true'] {
+    @include themed {
       outline: 1px solid var(--iui-color-foreground-primary);
```
mayank99, line 119: If you're using the themed mixin, then change this CSS variable to use `t`

mayank99: Maybe also change the other instance in line 49

FlyersPh9: Oh dang, I didn't even notice that was a CSS variable and not one of our `t()` variables, my mistake.   Both 119 & 49 are using the `t()` now.

src/carousel/carousel.scss
FlyersPh9: The thought was if someone only used the slides or only the navigation, the divider line wouldn't be present.  When would this happen, I have no idea.  I'll move this out of the slider mixin and put it back in the navigation mixin.
Screen Shot 2022-01-26 at 5 11 05 PM

mayank99 (approved):

backstop/tests/carousel.html
```diff
@@ -0,0 +1,529 @@
...
+                class="iui-button-icon"
+                aria-hidden="true"
+              ></svg-chevron-left>
+            </button>
+            <button
+              class="iui-button iui-borderless iui-small iui-carousel-navigation-animation-button"
+              type="button"
+              aria-label="Pause animation"
+              data-action="pause"
+              tabindex="-1"
```
veekeys, line 230: Is this supposed to be not focusable now?
If no, we have those rules with target of `data-pressed`. Dont need for animation button?

src/carousel/carousel.scss
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  &-dots,
+  &-left,
+  &-right {
+    display: flex;
+    align-items: center;
+    flex: 1;
+
+    button[data-pressed='true'] {
```
veekeys, line 45: How this `data-pressed` works? Is it getting applied on all the buttons? Not sure we wanna override all the buttons (including user ones)

src/carousel/carousel.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    &.iui-active::after {
+      @include themed {
+        background-color: t(iui-color-foreground-primary);
+      }
+    }
+  }
+
+  &-animation-button[data-pressed='true'] {
```
saaaaaally, line 116: Is this class needed? It doesn't seem to change anything if removed since the same button outline is being set in line 48 or am I missing something?

backstop/tests/carousel.html
FlyersPh9: Instead of focus going to each individual button we now have focus going on the entire carousel.  From there the user can use the left / right arrow keys or the spacebar / enter key to play / pause.  Unsure about the `data-pressed` thing.

src/carousel/carousel.scss
mayank99: Well, the user will probably not be using `data-pressed` on their buttons.

src/carousel/carousel.scss
mayank99: You're right

FlyersPh9: Good catch, removed.

saaaaaally: Don't forget to remove `iui-carousel-navigation-animation-button` in html too, line 226

PR #380, opened 2021-11-10
feat(Table): Add scroll snapping

### LINKED ISSUE #378: Add scroll snapping to table
Add [scroll snapping](https://codepen.io/scottjehl/pen/abJrPOP).

FlyersPh9:
Adds `iui-scroll-snapping` mixin to `.iui-table-body` to enable Y-axis scroll snapping. Closes #378.

### DIFF
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -160,7 +160,7 @@
   }
 }
 
-/// Adds the ability to toggle vertical scroll snapping by setting `.iui-scroll-snapping` as a modifier or at root level if used outside a class.
+/// Adds the ability to toggle vertical scroll snapping by setting `.iui-scroll-snapping` as a modifier (if used inside a class) or at root level.
 /// @arg $selector - selector to apply `scroll-snap-align: start` on. Defaults to '> *'
 @mixin iui-scroll-snapping($selector: '> *') {
   #{if(&, '&.iui-scroll-snapping', '.iui-scroll-snapping')} {
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -166,3 +166,9 @@
     }
   }
 }
+
+@mixin iui-scroll-snapping {
+  .iui-scroll-snapping {
+    scroll-snap-type: y mandatory;
+  }
+}
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -159,7 +159,7 @@
     }
   }
 }
-
+/// Mixin allows scroll snapping to a specific class if .iui-scroll-snapping is applied to the parent
 @mixin iui-scroll-snapping($selector: '> *') {
   .iui-scroll-snapping {
     scroll-snap-type: y mandatory;
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -159,7 +159,8 @@
     }
   }
 }
-/// Mixin allows scroll snapping to a specific class if .iui-scroll-snapping is applied to the parent
+/// Adds the ability to toggle vertical scroll snapping using .iui-scroll-snapping class.
+/// @arg $selector - selector to apply `scroll-snap-align: start` on. Defaults to '> *'
 @mixin iui-scroll-snapping($selector: '> *') {
   .iui-scroll-snapping {
     scroll-snap-type: y mandatory;
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -167,8 +167,12 @@
   }
 }
 
-@mixin iui-scroll-snapping {
+@mixin iui-scroll-snapping($selector: '> *') {
   .iui-scroll-snapping {
     scroll-snap-type: y mandatory;
+
+    #{$selector} {
+      scroll-snap-align: start;
+    }
   }
 }
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -159,10 +159,11 @@
     }
   }
 }
-/// Adds the ability to toggle vertical scroll snapping using .iui-scroll-snapping class.
+
+/// Adds the ability to toggle vertical scroll snapping by setting `.iui-scroll-snapping` as a modifier or at root level if used outside a class.
 /// @arg $selector - selector to apply `scroll-snap-align: start` on. Defaults to '> *'
 @mixin iui-scroll-snapping($selector: '> *') {
-  .iui-scroll-snapping {
+  #{if(&, '&.iui-scroll-snapping', '.iui-scroll-snapping')} {
     scroll-snap-type: y mandatory;
 
     #{$selector} {
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -126,6 +126,7 @@
   overflow-x: hidden;
   overflow-y: scroll;
   overflow-y: overlay;
+  scroll-snap-type: y mandatory;
   display: flex;
   flex-direction: column;
   flex-grow: 1;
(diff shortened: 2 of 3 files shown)


src/table/table.scss
```diff
@@ -129,11 +129,18 @@
...
-
   @include themed {
     background-color: t(iui-color-background-1);
   }
 
+  &.iui-scroll-snapping {
+    scroll-snap-type: y mandatory;
+
+    .iui-row {
+      scroll-snap-align: start;
```
mayank99, line 140: It might be ok to leave this rule in the general `.iui-row` block. It shouldn't have any effect when scroll-snapping is turned off

src/table/table.scss
```diff
@@ -129,11 +129,18 @@
   display: flex;
   flex-direction: column;
   flex-grow: 1;
-
   @include themed {
     background-color: t(iui-color-background-1);
   }
 
+  &.iui-scroll-snapping {
```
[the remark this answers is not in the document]
FlyersPh9: How about I put this in `/style/global.scss`:
```
.iui-scroll-snapping {
  scroll-snap-type: y mandatory;
}
```
And then leave `scroll-snap-align: start;` within the general `.iui-row` block?  This way the `.iui-scroll-snapping` class could be used in other places (maybe [menu](https://itwin.github.io/iTwinUI/backstop/minified/menu.html) & [codeblock](https://itwin.github.io/iTwinUI/backstop/minified/code.html)).  Those components would need to add the `scroll-snap-align: start;` part though.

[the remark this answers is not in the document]
mayank99: maybe in `mixins.scss` not `global.scss`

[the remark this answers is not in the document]
mayank99: or both actually. create a mixin and use it in a class in `global.scss`

similar to iui-anchor
image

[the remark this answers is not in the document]
mayank99: The problem is that global css is included in every single component/file, so we don't want to add unnecessary stuff there.

Maybe @veekeys has a different suggestion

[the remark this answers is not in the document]
veekeys: Global should not consist of just plain mixins, IMO. There should be left minimal needed rules for itwinui to look better (so global css reset, scroll, etc..).
If we want just to have some helper mixins, this is also good, but lets put them in separate file, which would be included only in demand. Also, if it gets big, we might want to put it in chunks even more.
I would also suggest to start thinking about expanding our documentation, havnig more readme or w/e, because we already have a bunch of really good stuff, but it is hard to find or even remember (personally).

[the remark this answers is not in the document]
veekeys: Just checked the implementation now..
One question, without `scroll-snap-align: start` snapping wont work at all?? Like it is mandatory?
Also, I am not sure about mixin now :D That would require including it explicitly in every component, where you actually need. Instead, I would think it should be enough to include css with that class in your app and you can use that class wherever you need (wouldnt that be more comfortable??). I am talking here from users perspective. But it still should not be in global (forced include).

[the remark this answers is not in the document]
mayank99: > One question, without scroll-snap-align: start snapping wont work at all?

Yes, because the browser doesn't know where to snap.

> That would require including it explicitly in every component, where you actually need.

Not a bad thing imo. It gives us the control to only provide snapping where we approve.

> I would think it should be enough to include css with that class in your app and you can use that class wherever you need

I was thinking the user would just toggle a react prop in the components we add this to. We can't expect them to find this class name, and it's not something difficult to write on your own.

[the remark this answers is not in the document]
veekeys: In this case, mixin of a single rule looks kinda useless...

> I was thinking the user would just toggle a react prop in the components we add this to. We can't expect them to find this class name, and it's not something difficult to write on your own.

You are talking here about react users, whereas I had in mind css users and the documentation part too. I do agree about props and our components, but I do feel scroll snapping set of rules might be something we could advertise and provide from css package as well.

gretanausedaite (commented):
Seems simple and good solution for this problem, I have no comments/issues with this code. This branch needs an update, otherwise approved.

src/table/table.scss
```diff
@@ -20,7 +20,7 @@
   }
 
   @include iui-table-cell-icon;
-  @include iui-scroll-snapping;
+  @include iui-scroll-snapping($selector: '.iui-row');
```
mayank99, line 23: Does this not work?

```suggestion
  @include iui-scroll-snapping(.iui-row);
```

src/style/mixins.scss
```diff
@@ -167,8 +167,12 @@
   }
 }
 
-@mixin iui-scroll-snapping {
+@mixin iui-scroll-snapping($selector: '> *') {
```
mayank99, line 163: Can you add some comments here? It's not obvious what exactly this mixin does without looking at the code

You can use the mixins in `labeled-inputs.scss` as example of how to format

src/table/table.scss
FlyersPh9: This works: `@include iui-scroll-snapping('.iui-row');`. I'll make the change.

src/style/mixins.scss
FlyersPh9: Added `/// Mixin allows scroll snapping to individual table rows if .iui-scroll-snapping is applied to the table`

mayank99: uhhh that's not what I meant 😆 The comment needs to document the mixin itself (in `mixins.scss`), not its reference

The tripe slash comment format serves as ["SassDoc"](http://sassdoc.com/) (we don't do anything with it currently, but it's nice).

https://github.com/iTwin/iTwinUI/blob/2bb38fb392dd459707e02500436f2f98166a573b/src/inputs/labeled-inputs.scss#L275-L279

mayank99: You can just commit this if it's easier 😄 

```suggestion
/// Adds the ability to toggle vertical scroll snapping using .iui-scroll-snapping.
/// @arg $selector - selector to apply `scroll-snap-align: start` on. Defaults to '> *'
@mixin iui-scroll-snapping($selector: '> *') {
```

src/style/mixins.scss
```diff
@@ -159,3 +159,13 @@
     }
   }
 }
+/// Mixin allows scroll snapping to a specific class if .iui-scroll-snapping is applied to the parent
+@mixin iui-scroll-snapping($selector: '> *') {
+  .iui-scroll-snapping {
```
mayank99, line 164: I think it would be more useful if this class was applied on the same level instead of as a child, i.e.
```suggestion
&.iui-scroll-snapping {
```


However, that would make this mixin unusable at the root level, so we can take it one step further and make it conditional. The suggestion below will apply it as a modifier if it's used inside another class, otherwise it will leave it at root level.
```suggestion
#{if(&, '&.iui-scroll-snapping', '.iui-scroll-snapping')} {
```

src/style/mixins.scss
FlyersPh9: Ah, yes I did misunderstand.  I like your suggestion, added.

mayank99 (approved):

mayank99:
@veekeys Please take another look at the new changes. The approach is changed quite a bit but it's more versatile now.

src/style/mixins.scss
FlyersPh9: Applied these changes, good suggestions.

veekeys:
> @veekeys Please take another look at the new changes. The approach is changed quite a bit but it's more versatile now.

Wouldnt it be you if from simple thing wouldnt end up with some sass magic :D 
I guess the only question would have what we would do if wanted to have also X snapping. But should be able to still reuse same mixin with some more params.
In general, you can merge this one. Should be fine..

PR #381, opened 2021-11-11
feat(Table): Add label to paginator row count dropdown button

FlyersPh9:
- Add "Rows per page" left of paginator row count dropdown button.  Responsive design will be handled in React.

Rows per page label


backstop/tests/table.html
```diff
@@ -2412,6 +2412,7 @@ <h2>Additional options</h2>
           </div>
 
           <div class="iui-right">
+            <span class="iui-text-muted">Rows per page</span>
```
mayank99, line 2415: Shouldn't these also include `iui-text-block`?

FlyersPh9: Applying that class doesn't appear to change anything visually.

FlyersPh9:
We do have responsive issues, but we have responsive issues in other areas throughout the table.  It might make sense to have this text right aligned.  Ideally we could hide the label when the table becomes too narrow, but we might want to wait for [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) before fixing that.
Screen Shot 2021-11-11 at 3 12 13 PM

mayank99:
> We do have responsive issues, but we have responsive issues in other areas throughout the table. It might make sense to have this text right aligned. Ideally we could hide the label when the table becomes too narrow, but we might want to wait for [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) before fixing that.

Container queries are not gonna be widely available for a while, so lets use media queries.

FlyersPh9:
> Container queries are not gonna be widely available for a while, so lets use media queries.

Using a media query to hide the label at 1024px.  Tested it in Storybook, should work well since the paginator reduces the buttons when elements start to conflict with one another.

bentleyvk:
Maybe we can use `ResizeObserver` to hide it? Because if media query is not ideal as Table might not take the full width of the screen e.g. when there is expanded SideNav.

gretanausedaite (commented):
Design looks good. 
Are we concerned about mobile screens? 
image

It seems like all of responsive issues would be solved in React.

FlyersPh9:
> Are we concerned about mobile screens?

I was thinking on extremely narrow screens (phone) when the rows per page & paginator can't fit on the same row that we could adjust the layout to something like this:
paginator-layout

I kind of had this working yesterday by creating a media query and setting:
```
.iui-center {align-self: flex start;}
.iui-left, .iui-right {align-self: flex-end;}
```

FlyersPh9:
I've added a `.iui-paginator-responsive` that changes the `flex-direction` to `column`, reorders left, center, right, and some other layout things.
Screen Shot 2021-11-12 at 9 25 09 AM

mayank99:
> I've added a `.iui-paginator-responsive` that changes the `flex-direction` to `column`, reorders left, center, right, and some other layout things.
but why? This is way too much code for near-zero benefit. And it's changing order which is confusing. And it doent look great.

Now I'm kinda with @bentleyvk on this. If we use ResizeObserver, we probably dont even need this PR (we can just reuse our muted typography)

FlyersPh9:
> The paginator collapses in React, so this is all unnecessary image

I would argue that this layout is more usable because we can display more page numbers with the full width being available.
Screen Shot 2021-11-12 at 9 40 21 AM

Is it possible for our users to completely remove `.iui-left` & `.iui-right` on mobile so they could achieve this if they wanted?
Screen Shot 2021-11-12 at 9 41 48 AM

mayank99:
> I would argue that this layout is more usable because we can display more page numbers with the full width being available.
Screen Shot 2021-11-12 at 9 40 21 AM

This is at the cost of reduced touch target size and gap, which increases the chance of misclicks.

And it's also a bad look for medium-size screens. And it will require changing the whole logic in React.

> Is it possible for our users to completely remove `.iui-left` & `.iui-right` on mobile so they could achieve this if they wanted?

"users" = React and Blazor, so yes. No one is using this table directly as it requires JS.

FlyersPh9:
> This is at the cost of reduced touch target size and gap, which increases the chance of misclicks.
Worth pointing out that screenshot is showing the smaller size buttons, but it works on all size table options.

bentleyvk:
> Now I'm kinda with @bentleyvk on this. If we use ResizeObserver, we probably dont even need this PR (we can just reuse our muted typography)
I am not saying that this PR is not needed 😀 I just think that media query is not needed. But I think that increasing vertical gap between pages and page size selector it will be ok. Because we don't want to reduce functionality for our mobile users.

mayank99:
> > Now I'm kinda with @bentleyvk on this. If we use ResizeObserver, we probably dont even need this PR (we can just reuse our muted typography)
> 
> I am not saying that this PR is not needed 😀 I just think that media query is not needed. But I think that increasing vertical gap between pages and page size selector it will be ok. Because we don't want to reduce functionality for our mobile users.

Have you seen this? 😆 https://github.com/iTwin/iTwinUI-react/pull/427

bentleyvk:
> > I am not saying that this PR is not needed 😀 I just think that media query is not needed. But I think that increasing vertical gap between pages and page size selector it will be ok. Because we don't want to reduce functionality for our mobile users.
> 
> Have you seen this? 😆 [iTwin/iTwinUI-react#427](https://github.com/iTwin/iTwinUI-react/pull/427)

I saw it but I still think this Jon's PR is needed.

src/table/paginator.scss
```diff
@@ -3,12 +3,13 @@
...
 @import '../icon/index';
 @import '../button/index';
+@import '../text/index';
 @import './variables';
 
 @mixin iui-paginator {
   @include iui-reset;
   display: flex;
-  justify-content: space-around;
+  justify-content: space-between;
```
mayank99, line 12: Why was this changed?

src/table/paginator.scss
```diff
@@ -70,4 +71,9 @@
   > .iui-right {
     justify-content: flex-end;
   }
+
+  &-page-size-label {
+    @include iui-text-muted;
+    white-space: nowrap;
```
mayank99, line 77: Doesn't this also need `overflow: hidden` and `text-overflow: ellipsis`?

src/table/extra-condensed.scss
```diff
@@ -6,7 +6,7 @@
 @mixin iui-table-extra-condensed {
   .iui-table-header .iui-cell,
   .iui-row .iui-cell,
-  .iui-paginator {
+  .iui-paginator:not(.iui-paginator-responsive) {
```
mayank99, line 9: This is leftover

src/table/condensed.scss
```diff
@@ -6,7 +6,7 @@
 @mixin iui-table-condensed {
   .iui-table-header .iui-cell,
   .iui-row .iui-cell,
-  .iui-paginator {
+  .iui-paginator:not(.iui-paginator-responsive) {
```
mayank99, line 9: Leftover

src/table/paginator.scss
FlyersPh9: I don't recall, it may have had something to do with the responsive design I was trying out.  But looking at [this image from CSS-Tricks' article](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#justify-content), I believe `space-between` is what we actually want.
image

src/table/paginator.scss
FlyersPh9: I thought we were planning to use React to auto-hide the label when the width gets too narrow?  If that's the case, truncation should never happen.

mayank99: Correct, we're hiding the label at a certain width (currently 1024px but maybe needs to be more narrow). But it can still overflow. See https://github.com/iTwin/iTwinUI-react/pull/427#discussion_r752500421

FlyersPh9: Ok, makes sense.  I've added this.

mayank99 (approved):

PR #388, opened 2021-11-18
feat: Add tree component

saaaaaally:
Add Tree Component

image


backstop/scenarios/tree.js
```diff
@@ -0,0 +1,23 @@
...
+  }),
+  scenario('With Checkbox', {
+    selectors: ['#demo-checkbox'],
+  }),
+  scenario('Disabled', {
+    selectors: ['#demo-disabled'],
+  }),
+  scenario('With Icons', {
+    selectors: ['#demo-icons'],
+  }),
```
saaaaaally, line 22: Not sure why tests are failing, `yarn test --filter=tree` ran successfully

bentleyvk:
Your build is failing. You should have seen the same error in your own machine:
image

src/tree/tree.scss
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  &.iui-disabled {
+    &,
+    &:hover {
+      cursor: not-allowed;
+      outline: none;
+      @include themed {
+        background-color: t(iui-color-background-1);
+        color: t(iui-text-color-placeholder);
```
FlyersPh9, line 108: In the tree we support a secondary line of text that uses the muted text color.  Because of this, our disabled row needs to use an even lighter text color, placeholder:
Screen Shot 2021-11-22 at 11 13 53 AM

Our table does _not_ support a secondary line of text that uses the muted color.  We should update the table's disabled rows text color to align with the tree.

FlyersPh9:
I noticed while doing a visual review of the tree that I incorrectly implemented the eyeball checkboxes.  The unchecked eyeball should be using the actionable icon color but it is currently using our primary color.

How it is currently:
Screen Shot 2021-11-22 at 11 18 11 AM

How it should be:
Screen Shot 2021-11-22 at 11 17 40 AM

I can fix this in a separate PR with this line added to `checkbox.scss`:
```scss
input:enabled ~ .iui-checkbox-checkmark > svg > .iui-uncheck {
  @include themed {
    fill: t(iui-icons-color-actionable);
  }
}
```

src/tree/tree.scss
mayank99: That will be an accessibility concern, especially in light theme.

In Table, we make the disabled rows have italic text. So it might be fine to leave it like that. But italic might not look good in Tree.

How often do we expect the disabled node pattern to be used?

FlyersPh9: The italics within the table is also slightly incorrect on my behalf as well.  My intent was for a disabled row to look like this without italics:
Screen Shot 2021-11-22 at 12 38 51 PM
 
The italics was meant to symbolize that the file is processing / uploading:
Screen Shot 2021-11-22 at 12 39 40 PM

I think this could be accomplished by creating an additional class like `.iui-processing` to add the italics & decouple the italics from the `.iui-disabled`.


And to answer you last question, I have no gauge on how often we expect tree nodes to be disabled.

backstop/tests/tree.html
```diff
@@ -0,0 +1,1305 @@
...
+      rel="stylesheet"
+      href="../../lib/css/all.css"
+    />
+    <link
+      rel="stylesheet"
+      href="../assets/demo.css"
+    />
+    <style>
+      section[id^='demo'] {
+        width: auto;
```
bentleyvk, line 27: This change makes test images bigger. I would say either remove this or you can add viewport like this `viewports: [{ width: 800, height: 600 }],` (check `expandable-block.js`).

backstop/tests/tree.html
```diff
@@ -0,0 +1,1305 @@
...
+
+    <hr />
+    <h2>With Checkbox</h2>
+    <section id="demo-checkbox">
+      <div class="iui-tree">
+        <div
+          class="iui-tree-node iui-expanded"
+          tabindex="0"
+        >
+          <label class="iui-checkbox iui-checkbox-visibility">
```
bentleyvk, line 308: Why eye background is not transparent?

image

backstop/tests/tree.html
```diff
@@ -0,0 +1,1305 @@
...
+            <div class="iui-caption">A basic description.</div>
+          </span>
+        </div>
+        <div
+          class="iui-tree-node iui-disabled"
+          style='--level: 1;'
+          tabindex="0"
+        >
+          <label class="iui-checkbox iui-checkbox-visibility iui-disabled">
+            <input
```
bentleyvk, line 709: Add `disabled` also in other elements.

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+        color: t(iui-color-foreground-body);
+      }
+      @media (prefers-reduced-motion: no-preference) {
+        transition: color $iui-speed-fast ease;
+      }
+    }
+  }
+
+  &.iui-expanded {
+    > .iui-button > .iui-icon {
```
bentleyvk, line 144: Maybe we should have a separate class for expander button? I am afraid that if user for some reason will have button inside node, it might rotate its icon, too.

backstop/tests/tree.html
```diff
@@ -0,0 +1,1305 @@
...
+    ></script>
+  </head>
+  <body class="iui-body">
+    <theme-button></theme-button>
+    <h1>Tree</h1>
+    <hr />
+
+    <h2>Default Tree</h2>
+    <section id="demo-default">
+      <div class="iui-tree">
```
bentleyvk, line 47: We should use `ul` and `li` tags with corresponding roles and `aria-*` attributes for the tree as W3 has in their documantation: https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+  @include iui-focus;
+  display: flex;
+  width: 100%;
+  align-items: center;
+  padding: $iui-xxs;
+  cursor: pointer;
+  box-sizing: border-box;
+  user-select: none;
+  margin-right: $iui-sm;
+  //min-height: $iui-component-height-small + (2 * $iui-xxs); //button height + padding
```
bentleyvk, line 25: Remove comments.

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+@mixin iui-tree {
+  @include themed {
+    background-color: t(iui-color-background-1);
+  }
+}
+
+@mixin iui-tree-node {
+  @include iui-focus;
+  display: flex;
+  width: 100%;
```
bentleyvk, line 18: Is it needed?

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+  align-items: center;
+  padding: $iui-xxs;
+  cursor: pointer;
+  box-sizing: border-box;
+  user-select: none;
```
bentleyvk, line 23: Do we really want user to prevent from selecting? What if there is a file name and user can't even copy the name.

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+  margin-right: $iui-sm;
```
bentleyvk, line 24: Why you need margin on the right then there isn't anything there?

backstop/tests/tree.html
```diff
@@ -0,0 +1,1305 @@
...
+          <span class="iui-node-label">
+            <div class="iui-title">Node 3</div>
+          </span>
+        </div>
+      </div>
+    </section>
+
+    <hr />
+
+    <h2>With Captions</h2>
```
bentleyvk, line 176: Maybe add icons to some nodes here.

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+  //min-height: $iui-component-height-small + (2 * $iui-xxs); //button height + padding
+  min-height: $iui-baseline * 3;
+
+  > .iui-icon {
+    @include iui-icons-default;
+    margin-right: $iui-s;
+    margin-left: $iui-sm + 1;
```
bentleyvk, line 31: Why you need margin left here? Especially why +1?

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  .iui-button + .iui-icon {
```
bentleyvk, line 34: Maybe you should have explicit names for the classes because for me it is really hard to read CSS when there is generic buttons and icons.

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+    min-width: 0;
+
+    &:only-child {
+      margin-left: $iui-child-node-spacing;
+    }
+  }
+
+  .iui-title,
+  .iui-caption {
+    white-space: nowrap;
```
bentleyvk, line 72: Can you add a very long string to HTML so we could really see that ellipsis is shown correctly?

src/tree/tree.scss
```diff
@@ -0,0 +1,148 @@
...
+
+  .iui-checkbox + .iui-node-label {
+    margin-left: $iui-s;
+  }
+
+  .iui-node-label {
+    min-width: 0;
+
+    &:only-child {
+      margin-left: $iui-child-node-spacing;
```
bentleyvk, line 66: These all margins look weird e.g. this one adds margin to the child but parent label also has margin.

backstop/tests/tree.html
saaaaaally: which elements?

backstop/tests/tree.html
FlyersPh9: I vote adding `viewports: [{ width: 800, height: 600 }],` to the JS.

backstop/tests/tree.html
bentleyvk: Checkboxes and buttons but only if node is disabled.

backstop/tests/tree.html
FlyersPh9: Fixed in #392.

backstop/tests/tree.html
saaaaaally: Added

src/tree/tree.scss
saaaaaally: Added iui-expander class

FlyersPh9: Oh that's a good catch.  I think I agree with Vilius on this one.

backstop/tests/tree.html
FlyersPh9: Using `<ul>` makes sense to me.

@VirginieGaudreault, tagging you so you can follow / suggest `aria-*` tags for accessibility.

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: Removed

backstop/tests/tree.html
saaaaaally: Added icons

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: Added iui-expander class

src/tree/tree.scss
saaaaaally: Added to With Caption

src/tree/tree.scss
saaaaaally: This one is needed to ensure the node is at the correct depth when the node label is the only element inside the node, like Node 2.2.1.1. In the nodes above the margin is added to the button but since this node doesn't have a button it needs to be added to the label instead.

image

src/tree/tree.scss
bentleyvk: It was more a question then asking to remove 😀 I think layout is broken now because of that and tests are failing:
image

src/tree/tree.scss
bentleyvk: Maybe we need to add padding on nodes or on all the tree itself because for me it doesn't look nice when text ends on the very beginning.
image

src/tree/tree.scss
saaaaaally: I added iui-node-content class to simplify margins and this should be fixed now

src/tree/tree.scss
saaaaaally: I added right margin to iui-node-label

src/tree/tree.scss
saaaaaally: I added iui-node-content class to simplify these margins

backstop/tests/tree.html
saaaaaally: Added

backstop/tests/tree.html
mayank99: Side note: That link is out of date. WAI ARIA Practices 1.2 should be preferred over 1.1. https://www.w3.org/TR/wai-aria-practices-1.2/#TreeView

Luckily in this case there is no difference between 1.1 and 1.2, but sometimes there are huge changes.

When in doubt, you can also just look at the bleeding edge version at https://w3c.github.io/aria-practices/

bentleyvk: I see, Saally, you are using ul and li elements now but if I understand recommendations correctly you need to have sub-items inside parent node element (like in the examples Mayank linked). It also should help to avoid having `--level` variable in CSS and it will let us to add animation in React.

src/tree/tree.scss
bentleyvk: Yeah but I still don't understand why you need right and left margins on different elements. I think appropriate `margin-right` on each element should be enough.

src/tree/tree.scss
bentleyvk: Better use padding for these kind of things.

src/tree/tree.scss
```diff
@@ -0,0 +1,146 @@
...
+
+  > .iui-checkbox {
+    margin-left: $iui-s;
+    margin-right: $iui-s;
+  }
+
+  > .iui-node-content {
+    display: flex;
+    align-items: center;
+    padding: $iui-xxs;
```
bentleyvk, line 29: Why you need this padding?

src/tree/tree.scss
```diff
@@ -0,0 +1,146 @@
...
+    box-sizing: border-box;
+    min-height: $iui-baseline * 3;
+    margin-left: calc(36px * (var(--level, 0)));
+    overflow: hidden;
+
+    > .iui-icon {
```
bentleyvk, line 35: Have you checked test images?
image
You might need to add `flex-shrink: 0`

backstop/scenarios/tree.js
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
```
bentleyvk, line 7: Need tests with hover and focus.

backstop/tests/tree.html
saaaaaally: I added ul sub tree and got rid of the `--level` variable but this caused a problem with keeping the checkboxes left aligned. `ul` causes the entire `iui-tree-node` to shift, where as before I was only indenting `iui-node-content`. Any ideas how to get around this to shift only the node content, not the entire node or should I not nest the subtrees and keep the flat view? 
tree

saaaaaally: I fixed the checkboxes to make them align on the left. I had to keep the `--level` variable to set the spacing between the checkbox and `iui-node-content`. Not sure if there would be a good way to automatically set the margin by detecting how deep it is (ul li = level 0, ui li ui li = level 2) because then we have to set a max depth, with `--level` the max depth is up to the user

src/tree/tree.scss
saaaaaally: Additional margins are needed on some elements to help align with other elements above and below. 

For example `.iui-icon` has left margin so that it aligns with the icon inside the expander button, without this it aligns with the button but not the icon inside the button
image

src/tree/tree.scss
```diff
@@ -0,0 +1,151 @@
...
+@import '../icon/index';
+
+$iui-active-outline: thin solid t(iui-color-foreground-primary);
+
+@mixin iui-tree {
+  @include themed {
+    background-color: t(iui-color-background-1);
+  }
+  list-style: none;
+  padding-left: 0;
```
bentleyvk, line 13: Can you also reset margin?

backstop/tests/tree.html
```diff
@@ -0,0 +1,1613 @@
...
+                </div>
+              </span>
+            </div>
+          </div>
+        </li>
+      </ul>
+    </section>
+
+    <hr />
+    <h2>With Checkbox</h2>
```
bentleyvk, line 495: Shouldn't first "Subcategory" label align with the second one's label?
image

saaaaaally: Not sure, what do you think @FlyersPh9

src/tree/tree.scss
saaaaaally: Updated this margin

src/tree/tree.scss
saaaaaally: Added

backstop/tests/tree.html
FlyersPh9: image
Good catch Vilius.  If a subcategory can not be expanded we should have empty space where the expand arrow would be.  Otherwise, if we had a mix of expandable and non-expandable subcategories the text would not align.

saaaaaally: Added left margin to `iui-node-label` when expander button is not present

bentleyvk: Text after icon is still misaligned:
image

Now I see that the icon is also misaligned with previous node text.

bentleyvk:
@FlyersPh9 Just started to think that our expander icons are a bit too big which makes 3rd level to be far from left:
image
Comparison with UX standards page (and I guess iModel.js):
image

backstop/tests/tree.html
bentleyvk: Can you also add required roles from W3?

src/tree/tree.scss
```diff
@@ -0,0 +1,160 @@
...
+
+    > .iui-icon {
+      @include iui-icons-default;
+      margin-right: $iui-s;
+      margin-left: $iui-xs + 2; //Help align with icon inside expander button 2px (button left padding) + 1px (button left border) + 3px (button icon left margin)
+      flex-shrink: 0;
+
+      &:first-child {
+        margin-left: 42px; // Align with other icons when expander button is not present
+        // (icon margin 6px) + (icon width 16px) + (button padding 4px) + (button border 2px) + (button right margin 8px) + (icon align 6px)
```
bentleyvk, line 47: Maybe better write whole comment above this margin.

backstop/tests/tree.html
bentleyvk: Ok, it is weird situation, it seems we either align icon with expander or icon with label text.

saaaaaally: Fixed alignment

backstop/tests/tree.html
```diff
@@ -0,0 +1,1611 @@
...
+                </div>
+              </div>
+            </li>
+            <li>
+              <div
+                class="iui-tree-node"
+                style='--level: 1;'
+                tabindex="0"
+                aria-expanded="false"
+                id="test-node"
```
FlyersPh9, line 310: Can you move this `id="test-node"` down 2 rows?  That way we're testing on a row that has the secondary line of text.
Screen Shot 2021-12-01 at 4 23 44 PM

saaaaaally: For some reason test was not working on that node, but I moved test node to the first node that also has a caption

backstop/tests/tree.html
saaaaaally: Added roles

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+$iui-active-outline: thin solid t(iui-color-foreground-primary);
+
+// Spacing for when expander button is not present: (icon margin 6px) + (icon width 16px) + (button padding 4px) + (button border 2px)
+$iui-expander-button-width: 28px;
```
mayank99, line 9:
```suggestion
// 28px = (icon margin) + (icon width) + (button padding) + (button border 2px)
$iui-expander-button-width: ($iui-component-offset * 2) + ($iui-icons-default) + ($iui-xxs * 2) + 2px;
```

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+    align-items: center;
+    box-sizing: border-box;
+    min-height: $iui-baseline * 3;
+    margin-left: calc(28px * (var(--level, 0)));
+    overflow: hidden;
+    padding-left: $iui-xxs;
+
+    &-icon {
+      @include iui-icons-default;
+      padding: 0 6px;
```
mayank99, line 44:
```suggestion
      padding: 0 ($iui-component-offset * 2);
```

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+  &-checkbox {
+    margin-right: $iui-s;
+  }
+
+  &-content {
+    display: flex;
+    align-items: center;
+    box-sizing: border-box;
+    min-height: $iui-baseline * 3;
+    margin-left: calc(28px * (var(--level, 0)));
```
mayank99, line 38:
```suggestion
    margin-left: calc(#{$iui-expander-button-width} * (var(--level, 0)));
```

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+
+      .iui-tree-node-content-caption {
+        @include themed {
+          color: t(iui-text-color-placeholder);
+        }
+      }
+    }
+  }
+
+  &:hover {
```
mayank99, line 129: This hover block should be moved before `&.iui-active` and `&.iui-disabled`

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
+    }
+
+    .iui-tree-node-content-caption {
+      @include themed {
+        color: t(iui-color-foreground-body);
+      }
+      @media (prefers-reduced-motion: no-preference) {
+        transition: color $iui-speed-fast ease;
```
mayank99, line 139: color transition is not "motion" so this media query is not needed

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+      }
+    }
+  }
+
+  &.iui-expanded {
+    > .iui-tree-node-content > .iui-tree-node-content-expander > .iui-button-icon {
```
mayank99, line 145: This long selector can be fixed by giving a modifier to `iui-button-icon`

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+
+    &:focus {
+      outline-width: $iui-xxs;
+      outline-offset: -2px;
+    }
+  }
+
+  &.iui-disabled {
+    &,
+    &:hover {
```
mayank99, line 113: This `&:hover` selector should be removed from here and from iui-active as well.

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+
+    &-caption {
+      font-size: $iui-font-size-small;
+      @include themed {
+        color: t(iui-text-color-muted);
+      }
+    }
+  }
+
+  &:focus {
```
mayank99, line 88: `:focus` will not work. When you add interaction, the focus will need to be handled manually (probably using `aria-activedescandant`), so you need a custom class name to emulate focus styling.

You can find all the info you need on W3C. [https://w3c.github.io/aria-practices/#:~:text=the%20tree%20role%20supports%20the%20aria-activedescendant%20property](https://w3c.github.io/aria-practices/#TreeView)

src/tree/tree.scss
```diff
@@ -0,0 +1,149 @@
...
+
+// Spacing for when expander button is not present: (icon margin 6px) + (icon width 16px) + (button padding 4px) + (button border 2px)
+$iui-expander-button-width: 28px;
+
+@mixin iui-tree {
+  @include themed {
+    background-color: t(iui-color-background-1);
+  }
+  list-style: none;
+  @include iui-reset;
```
mayank99, line 16: nit: reset should always come before everything else

src/tree/tree.scss
mayank99: You can also add a second variable to make things more readable/usable.
```suggestion
$iui-expander-inline-padding: $iui-component-offset + $iui-xxs + 1px; // margin + padding + border
$iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding * 2); // icon width + inline padding
```

saaaaaally: Changed

src/tree/tree.scss
saaaaaally: Changed

src/tree/tree.scss
saaaaaally: Changed

src/tree/tree.scss
saaaaaally: Moved

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: changed to `iui-expander-icon`

src/tree/tree.scss
saaaaaally: Removed

src/tree/tree.scss
saaaaaally: Moved

src/tree/tree.scss
mayank99: That's not really helping, because now that modifier is applied on all expanders. What you need is a class that will only be used to rotate the icon.

saaaaaally: What do you mean? Should it be something like iui-tree-node-content-expander-icon?

mayank99: what I mean is that selector still has 0-4-0 specificity (previously it had 0-5-0 but this decrement is not because off the new classname).

Yes `iui-tree-node-content-expander-icon` might help, because then you won't need to nest it inside `iui-expanded`

saaaaaally: Changed it to use only `iui-tree-node-content-expander-icon`

src/tree/tree.scss
bentleyvk: Why we will need that attribute to handle focus?
image

bentleyvk (approved):

src/tree/tree.scss
mayank99: i thought we would need it because we have extra focusable elements checkboxes and buttons for every node, but now I'm thinking we could just set disable tabindex in those too

src/tree/tree.scss
mayank99: Maybe I'm not being clear. I'm trying to say you don't need `iui-expanded` at all. It's adding unnecessary nesting and not doing anything else. So you could just have a class whose sole purpose is to rotate the icon. That way its specificity will remain low.

saaaaaally: If I remove iui-expanded how/where would I do the rotation?

mayank99: You can do the rotation by toggling a class on the icon itself.

Example:
image

This does make the class name super long but the specificity stays at 0-1-0. You can also do `&.iui-expanded` instead of `&-expanded` if you don't like the long class; it does increase the specificity to 0-2-0 but it's better than current situation.

Also unrelated but I noticed an extra (unused) class in your HTML which you may want to remove
image

src/tree/tree.scss
mayank99: The focus management will get a little complex here when we add this in React. W3C only has recommendations for expansion (using left/right arrow keys) and selection (using space/enter). Since we have checkboxes too, we will need to innovate here.

We can leave `:focus` here for now, but we should also add `:focus:not(:focus-visible)` at least. If we later find the need for a class like `.iui-focus`, we can come back and change it.

src/tree/tree.scss
saaaaaally: Removed `iui-expanded` and `iui-tree-node-content-expander`

src/tree/tree.scss
saaaaaally: Added `:focus:not(:focus-visible)`

src/tree/tree.scss
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+$iui-active-outline: thin solid t(iui-color-foreground-primary);
```
FlyersPh9, line 6: I've never used `thin` before, I typically use a variable or a hard coded value.  Does `thin` always equal 1px across different browsers / OSes?

src/tree/tree.scss
```diff
@@ -0,0 +1,137 @@
...
+    &:focus {
+      outline-width: $iui-xxs;
+      outline-offset: -2px;
+    }
+  }
+
+  &.iui-disabled {
+    cursor: not-allowed;
+    outline: none;
+    background-color: transparent;
```
FlyersPh9, line 126: I removed `background-color` from the tree SCSS and placed it in the `<head><style>`.  To avoid `.iui-disabled` from getting a hover state, I changed the `background-color` to `transparent`.  An alternative would be `&:hover:not(.iui-disabled)` but I imagine that has specificity issues.

FlyersPh9 (approved):

PR #392, opened 2021-11-23
fix(Checkbox): Changing visibility hover background color & unchecked fill color.

FlyersPh9:
- Changes visibility checkbox hover background color to allow background color to come through, matches borderless button's background fill color.  [Solves this issue in tree](https://github.com/iTwin/iTwinUI/pull/388#discussion_r754852780).
- Changes the fill color of the unchecked visibility checkbox glyph to better align with the standards.  [Discovered this issue in tree](https://github.com/iTwin/iTwinUI/pull/388#issuecomment-975690366).

iTwinUI_checkbox_Type_specialty_0_demo-special_0_desktopiTwinUI_dark_checkbox_Type_specialty_0_demo-special_0_desktop


src/inputs/checkbox.scss
```diff
@@ -31,7 +31,7 @@
     // Adds hover to everything that is not disabled
     &:hover > input:enabled ~ .iui-checkbox-checkmark {
       @include themed {
```
bentleyvk, line 33: So for disabled checkmark, background is still there?

FlyersPh9: This is only for hover on enabled visibility checkboxes.  All disabled checkboxes have the same background color (hover or no hover) across all checkboxes.

src/inputs/checkbox.scss
```diff
@@ -58,5 +58,11 @@
       opacity: 1;
       transition: opacity $iui-speed-fast ease;
     }
+
+    input:enabled ~ .iui-checkbox-checkmark > svg > .iui-uncheck {
```
mayank99, line 62: You don't need `:enabled` here, as disabled checkboxes already have a high specificity which should override this if you remove `> svg`.

image

But I did have to reduce specificity in one other place in `checkbox-radio`.
image

src/inputs/checkbox.scss
mayank99: Then it won't really solve this issue.
image

Unless that is the intended look

FlyersPh9: I believe Vilius' screenshot is when hovered.  Here is a screenshot of this PR's checkboxes with a pink background.  Top checkbox is hovered (allows some of the pink bg to come thru due to opacity) and bottom checkbox has no hover.
Screen Shot 2021-11-23 at 10 22 37 AM

src/inputs/checkbox.scss
FlyersPh9: Great suggestion, change is made!

mayank99 (approved):

bentleyvk (approved):

PR #394, opened 2021-11-24
fix(UserIcon): Apply background color to transparent images

FlyersPh9:
If a PNG with transparency is used for the user icon photo the circle's color and initials / SVG display through:
Screen Shot 2021-11-24 at 3 29 16 PM Screen Shot 2021-11-24 at 3 49 16 PM

This PR applies background-1 to the `` itself to hide any content underneath:
Screen Shot 2021-11-24 at 3 29 59 PM


src/user-icon/user-icon.scss
```diff
@@ -70,6 +70,7 @@
     height: 100%;
     border-width: 0;
     border-radius: 50%;
+    background-color: #FFF;
```
mayank99, line 73: While I would have liked to see a dark background on dark theme, I think if someone is using an img it likely is designed for light theme only, so I guess it's ok.
---
Why `#FFF` over `white` tho?

mayank99: we already clear the background in react btw, so most users will never experience this
image

FlyersPh9: Updated to use background-1.  This may still introduce issues if the transparent PNG is mostly black, but the same could be said if the transparent PNG was mostly white.  Dev teams can override as needed.

mayank99: This is better, but I wonder if we should just leave it up to the devs to set the background. Currently because of the specificity, it'd be hard to override.

FlyersPh9: It's only `.iui-user-icon > img { background-color: red; }`, right?  My concern is if we don't provide anything, they'll remove the color entirely from the user icon (like they did on the OIDC screen).  If they remove the color & if no picture is provided, we end up with this for initials.
Screen Shot 2021-12-01 at 11 58 13 AMScreen Shot 2021-12-01 at 11 58 07 AM

mayank99 (approved):

PR #404, opened 2021-12-06
fix(Button): Fixed hover effects after button is clicked

gretanausedaite:
Fixed a bug where button hover effects are not applied after button is clicked:
- Added `:not(:hover)` to `focus:not:(focus-visible)` selector.

### DIFF
--- a/src/button/cta.scss
+++ b/src/button/cta.scss
@@ -18,7 +18,8 @@
     }
   }
 
-  &:focus:not(:focus-visible) {
+  // `not(:hover)` is needed for focus styles after button is clicked.
+  &:focus:not(:focus-visible):not(:hover) {
     @include iui-button-cta-base;
   }
 
--- a/src/button/cta.scss
+++ b/src/button/cta.scss
@@ -8,24 +8,30 @@
 @mixin iui-button-cta {
   @include iui-button-cta-base;
 
-  &:hover,
-  &:active,
   &:focus {
-    @include themed {
-      background-color: t(iui-color-background-positive-overlay);
-      border-color: t(iui-color-background-positive-overlay);
-      color: t(iui-color-foreground-accessory);
+    @include iui-button-cta-hover-focus;
+
+    &:focus:not(:focus-visible):not(:hover) {
+      @include iui-button-cta-base;
     }
   }
 
-  // `not(:hover)` is needed for focus styles after button is clicked.
-  &:focus:not(:focus-visible):not(:hover) {
-    @include iui-button-cta-base;
+  &:hover,
+  &:active {
+    @include iui-button-cta-hover-focus;
   }
 
   &[disabled],
   &:disabled {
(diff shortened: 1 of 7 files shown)


src/button/default.scss
```diff
@@ -20,7 +20,8 @@
     @include iui-notification-marker-hover;
   }
 
-  &:focus:not(:focus-visible) {
+  // `not(:hover)` is needed for focus styles after button is clicked.
+  &:focus:not(:focus-visible):not(:hover) {
```
mayank99, line 24: I took the time to think about this some more. This is a bandaid fix which I'm afraid can cause other problems as it increases specificity even more.

One way to fix this is by changing all instances of `:focus:not(:focus-visible)` to `:focus:where(:not(:focus-visible))`. The `:where` selector doesn't add specificity so it will lower it.

The [`:where`](https://caniuse.com/mdn-css_selectors_where) selector is new but not much newer than [`:focus-visible`] itself, so I think it's fine to use here.

src/side-navigation/side-navigation.scss
```diff
@@ -141,6 +141,7 @@
       &:focus:not(:focus-visible) {
         @include themed {
           box-shadow: inset $iui-xxs 0 0 0 t(iui-icons-color-primary);
+          color: t(iui-color-foreground-primary);
```
mayank99, line 144: This is still not completely fixing the problem. If you try selecting a side nav button, it has a leftover top border now.
image

Maybe you won't need any of these overrides once you add `:where`.

src/button/default.scss
mayank99: With this approach, just remember to put the focus rules _before_ the hover ones so that the cascade knows which one should take preference.

If the focus and hover styles were previously grouped, a new mixin may be required.
image
image

src/side-navigation/side-navigation.scss
gretanausedaite: Yes, it's not needed when added `:where`. Removed.

src/button/cta.scss
```diff
@@ -8,18 +8,17 @@
...
-  &:hover,
-  &:active,
   &:focus {
-    @include themed {
-      background-color: t(iui-color-background-positive-overlay);
-      border-color: t(iui-color-background-positive-overlay);
-      color: t(iui-color-foreground-accessory);
+    @include iui-button-cta-hover-focus;
+
+    &:focus:not(:focus-visible):not(:hover) {
```
FlyersPh9, line 14: This line should also follow the `&:where(:not(:focus-visible)) {` structure I believe.

src/button/default.scss
gretanausedaite: oops :(

src/button/cta.scss
gretanausedaite: Fixed.

FlyersPh9 (approved):

PR #409, opened 2021-12-09
fix(Inputs): Icon alignment fix in input container

### LINKED ISSUE #408: InputContainer status does not work with combobox
Our input container rules are applied on other input containers.
This is the issue with iui-icon:

image

What happens:
image

As we have in this case 2 `iui-input-container` instances, rules of the first one are being applied to the second ones icon. This fixes the issue:
image

This is the full html:
image

Not sure if proposed fix will not brake anything as we heavily rely on css structure with grid layout.

veekeys:
Closes #408

### DIFF
--- a/src/inputs/classes.scss
+++ b/src/inputs/classes.scss
@@ -9,7 +9,7 @@
     @include iui-input-container-label;
   }
 
-  .iui-input-icon {
+  > .iui-input-icon {
     @include iui-input-icon;
   }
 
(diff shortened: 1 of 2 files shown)


mayank99 (approved):
At first I thought adding `>` will break the ability to wrap icons (for things like tooltips). But the `iui-input-icon` class can just be applied on the wrapper, so it should be fine. We handle it correctly in react too.
image

FlyersPh9 (approved):

PR #414, opened 2021-12-10
fix(Header): Correct label color in disabled header button

FlyersPh9:
Disabled button in the header was not getting the correct color.  Is now fixed.

Before:
Screen Shot 2021-12-10 at 4 03 52 PM

After:
Screen Shot 2021-12-10 at 4 03 36 PM


src/header/header.scss
```diff
@@ -415,7 +415,7 @@
       box-shadow: 0 $iui-xxs 0 0 rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
     }
 
-    .iui-label {
+    .iui-button-label {
```
mayank99, line 418: Can you also change the class below from iui-icon to iui-button-icon?
image

gretanausedaite (commented):
Sorry for the bug :( I think we should add item for (or just add) tests for disabled state. This could be prevented.

src/header/header.scss
FlyersPh9: Done.

FlyersPh9:
> Sorry for the bug :( I think we should add item for (or just add) tests for disabled state. This could be prevented.

No worries, it happens.  I've added a test for the disabled header buttons.

mayank99 (approved):

PR #417, opened 2021-12-15
feat(Table): Added proper support for horizontal scroll

### LINKED ISSUE #410: Table: need to support horizontal scrolling
Need to support horizontal scrolling when total columns width is wider than Table itself.
Here is the codesandbox with fix for static column widths but it breaks regular columns:
https://codesandbox.io/s/itwinui-react-example-forked-8fn5v?file=/src/App.tsx
There `overflow-x: auto` and `align-items: flex-start` are applied to the Table. Link to the explanation: https://stackoverflow.com/questions/40141163/make-flex-items-take-content-width-not-width-of-parent-container/40141200#40141200

image

bentleyvk:
table-scroll-hack

Closes #410 
I couldn't find a proper way to have both scrollbars on table body and not on header. So I am syncing body scroll position with header using JS.
I tried to look into Antd Table but I couldn't understand what they are doing there 😁 but they I think they are doing something similar.
https://ant.design/components/table/#components-table-demo-fixed-columns-header


mayank99:
The easy solution for this is to move paginator in a separate div so that header and body can scroll together

bentleyvk:
> The easy solution for this is to move paginator in a separate div so that header and body can scroll together

I don't want us to have this:
image

mayank99:
If we made the table header use `position: sticky`, that scrollbar displaying over the header might not be too bad, considering the amount of code it saves.

bentleyvk:
> If we made the table header use `position: sticky`, that scrollbar displaying over the header might not be too bad, considering the amount of code it saves (no JS!)

I really don't like scroll on header, it looks ugly too me 😁 Also, it will be kind of regression for our users that doesn't have horizontal scrolling because vertical scroll will appear on header.

gretanausedaite (commented):
I don't think having this much JS is bad. Maybe we could document it somewhere so the users would know to add it to their projects when using only iTwinUI-CSS package.
Do we need to test only horizontal scroll (without vertical)?
Side topic: 
Could we add some kind of scroll track on hover? This looks weird not going to the end:
image

bentleyvk:
> I don't think having this much JS is bad. Maybe we could document it somewhere so the users would know to add it to their projects when using only iTwinUI-CSS package. Do we need to test only horizontal scroll (without vertical)? Side topic: Could we add some kind of scroll track on hover? This looks weird not going to the end: 

What you mean by "testings only horizontal"?

Would it help having e.g. white background when hovered over scroll? Because it kind of looks the same 😀
image

gretanausedaite:
> What you mean by "testings only horizontal"?

Without vertical scrollbar - have 4 (or some amount of) rows in table that fit without scroll.
 
> Would it help having e.g. white background when hovered over scroll? Because it kind of looks the same 😀 image

I was thinking transparent gray. But I can be wrong, maybe we don't need track color :)

src/table/table.scss
```diff
@@ -117,12 +119,15 @@
 }
 
 @mixin iui-table-body {
-  overflow-x: hidden;
+  overflow-x: auto;
+  overflow-x: overlay;
   overflow-y: scroll;
   overflow-y: overlay;
```
mayank99, line 125: These can be combined into `overflow: auto; overflow: overlay;`

src/table/table.scss
```diff
@@ -117,12 +119,15 @@
...
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  // Allows columns with static width to take more space than the table width
+  align-items: flex-start;
```
mayank99, line 130: If you change this to `center`, you won't need to specify it again for empty table

FlyersPh9:
> I really don't like scroll on header, it looks ugly too me

I agree that the scrollbar should not be present over the table header unless the table header is also being scrolled like the body of the table.

src/table/table.scss
bentleyvk: Changed

bentleyvk:
> > What you mean by "testings only horizontal"?
> 
> Without vertical scrollbar - have 4 (or some amount of) rows in table that fit without scroll.

Added

src/table/table.scss
```diff
@@ -18,6 +18,8 @@
 
 @mixin iui-table-header {
   user-select: none;
+  overflow: hidden;
```
mayank99, line 21: We could improve the experience for Firefox.

This is what it currently looks like. Notice how the header and body cells are not vertically aligned because the scrollbar takes up space.
image

If we make the scroll track visible on the header, it will line up.
```suggestion
  overflow: hidden;
  @supports not (overflow: overlay) {
    overflow-y: scroll;
  }
```

Would need to also change body to use `overflow: scroll` instead of `overflow: auto`

bentleyvk: Added

PR #418, opened 2021-12-17
fix(ButtonGroup): Use flex instead of position to set z-index

mayank99:
Changed z-index approach to rely on `display: flex` instead of `position: relative` to fix this weird bug:
zindex

Also changed logic a little to give focus priority over hover, and added correct focus-visible behavior.

### DIFF
--- a/src/button/button-group.scss
+++ b/src/button/button-group.scss
@@ -10,16 +10,22 @@
   $childSelector: 'input, button'; // target nested inputs and buttons
 
   > * {
+    display: flex;
+
     #{$childSelector} {
-      &:hover,
-      &:focus,
-      &:focus-within {
-        position: relative;
+      &:hover {
         z-index: 1;
       }
 
+      &:focus {
+        z-index: 2;
+
+        &:where(:not(:focus-visible):not(:hover)) {
+          z-index: revert;
+        }
+      }
+
       &:disabled {
-        position: relative;
         z-index: -1;
       }
 
--- a/src/table/paginator.scss
+++ b/src/table/paginator.scss
@@ -40,7 +40,6 @@
 
       > * {
         margin-right: 0; // unset negative margin from button-group
-        display: flex; // for vertically aligning svgs
 
         > .iui-button,
         > .iui-ellipsis {


FlyersPh9 (approved):

src/button/button-group.scss
```diff
@@ -10,16 +10,22 @@
   $childSelector: 'input, button'; // target nested inputs and buttons
 
   > * {
+    display: flex;
```
bentleyvk, line 13: Why does this work? Maybe you have an article explaining `display: flex` vs `position: relative`?

src/button/button-group.scss
```diff
@@ -10,16 +10,22 @@
...
-      &:focus-within {
-        position: relative;
+      &:hover {
         z-index: 1;
       }
 
+      &:focus {
+        z-index: 2;
+
+        &:where(:not(:focus-visible):not(:hover)) {
```
bentleyvk, line 23: Is `where` needed?

mayank99: Yes, we want to keep the specificity low and use the cascade. `where` has similar browser support as `focus-visible` so it's not an issue.

bentleyvk (approved):

src/button/button-group.scss
veekeys, line 13: why not to child (child selector) ?

veekeys: so if it as default position, its considered as "not positioned"?

mayank99: "positioned" = anything other than `static` (default)

mayank99: z-index is applied to "flex items" which means children of the element that has `display: flex`

veekeys: hopefully this will not break anything....
Did you figure why in css it was not reproduced?

mayank99: It will not break anything because this div only has one child, so `display: flex` does not hurt (it actually helps in most cases, like vertical alignment)

I could not find out exactly why, but I think it might have to do with Storybook's custom styling. If I popped the story out in new tab, it was working fine. It's possible that user's app could have similar styling which might break it, so it's better to fix it anyway

mayank99: Actually I thought about it some more and found three situations:
- split-menu/tooltips are put in the same div by default, so it doesn't always have "only one child". But those are using absolute/fixed positioning so it should not be affected
- input-container has input nested inside a div, so it's not a direct child but it's still a grid item, so z-index works.
- there is one situation it could "break", which is if the child is wrapped (e.g. a div around a disabled button). In that case, the user would need to manually add `display: flex` or `position: relative`

saaaaaally (approved):

PR #423, opened 2021-12-20
fix(Modal): Modal fits within landscape phone display

FlyersPh9:
Before:
Screen Shot 2021-12-20 at 3 36 15 PM

After:
Screen Shot 2021-12-20 at 3 35 39 PM

### DIFF
--- a/src/modal/modal.scss
+++ b/src/modal/modal.scss
@@ -28,6 +28,7 @@
     z-index: 1000;
     max-width: 50%;
     min-width: 400px;
+    max-height: 100%;
     border-radius: $iui-border-radius;
     box-shadow: $iui-elevation-24;
     padding: $iui-baseline $iui-m;


src/modal/modal.scss
```diff
@@ -28,6 +28,7 @@
     z-index: 1000;
     max-width: 50%;
     min-width: 400px;
+    max-height: 100%;
```
mayank99, line 31: Three problems:
- modal uses content box, so this 100% does not include padding
- for portrait orientation, this will not consider the notch, though that's an extremely rare case I guess
- scrolling is not enabled

saaaaaally, line 31: Just curious, how do you test that this works?

FlyersPh9: I'm going to close this PR and we can revisit another time.

FlyersPh9: I first encountered the bug when viewing the modal demo on my phone in landscape.  To test, I inspected the page & then:

1. Press the button in devtools
2. Select a device with a small screen size, I picked iPhone 5.
3. Press the rotate button.

Screen Shot 2021-12-22 at 4 48 14 PM
 
You may need to enable **Show device frame** from this menu:

Screen Shot 2021-12-22 at 4 49 36 PM

PR #426, opened 2021-12-28
fix(Inputs): Fix radio button dot and checkbox checkmark centering

### LINKED ISSUE #406: Circles in radio buttons and handles of toggles are offset
We noticed that in one of the Design Review modal dialogs that is using iTwin UI, circles in radio buttons and handles of toggles are offset a bit and not aligned properly (notice unit system radio buttons and issues) toggle:

image

Seems like it is affected by the browser window size, as resizing it changes where circles/handles are being placed.

Easiest to reproduce in Firefox, but same thing happens in Chrome to a lesser extent.

FlyersPh9:
- Checkbox and radio are placed using `::after` which fixes half of #406.
- Checkmark / radio dot / visibility eye are filled on the SVG level rather than on the path, g, circle level.
- SVGs are displayed at 100% width / height, & SVGs have been updated. Explanation [here](https://github.com/iTwin/iTwinUI/pull/426#discussion_r780472977).
- Adding `name="radio-demo-14"` to the radio button group that has SVG smiley's next to them which fixes a bug on the demo page.


mayank99:
You won't need to change the svg size to 14x14 if you use `outline` instead of `border` here:
image

FlyersPh9:
> One thing I've tried that might fix it is using a pseudo element

I do remember playing around with this but didn't run with it because it would only work for radio button dots.  Unless there is a way to use `content` to display an SVG for the checkbox that I am unaware of.

src/inputs/radio.scss
```diff
@@ -5,7 +5,7 @@
 @mixin iui-radio {
   @include iui-inputs-checkbox-radio;
 
-  .iui-radio-dot {
+  .iui-radio-dot::after {
```
saaaaaally, line 8: Not sure if this is supposed to be part of the fix in this pr but I noticed when zooming in a lot sometimes the radio dot becomes an oval. Does this look right? 

image

image

FlyersPh9: Yikes, that's not good.  Investigating, i think it probably has something to do with `flex-shrink`.

FlyersPh9: Yep, it was `flex-shrink`, change is pushed.  Nice catch.

src/inputs/checkbox-radio.scss
```diff
@@ -114,8 +121,8 @@
   // Adds focus to enabled checkbox & radio
   input:enabled {
     &:focus {
-      ~ .iui-checkbox-checkmark,
-      ~ .iui-radio-dot {
+      ~ .iui-checkbox-checkmark::after,
+      ~ .iui-radio-dot::after {
```
mayank99, line 125: `::after` should not be needed for focus outline

src/inputs/checkbox-radio.scss
```diff
@@ -30,31 +29,42 @@
...
+      @media (prefers-reduced-motion: no-preference) {
+        transition: border-color $iui-speed-fast ease-out;
+      }
     }
 
     svg:not(.iui-radial) {
-      width: $iui-sm;
-      height: $iui-sm;
+      width: 100%;
+      height: 100%;
```
mayank99, line 58: Maybe it makes sense to keep the explicit size?

src/inputs/checkbox-radio.scss
```diff
@@ -30,31 +29,42 @@
...
+      content: '';
+      position: absolute;
+      inset: 0;
+      z-index: -1;
+      @include themed {
+        border: 1px solid rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
+        background-color: t(iui-color-background-1);
+      }
+      @media (prefers-reduced-motion: no-preference) {
+        transition: border-color $iui-speed-fast ease-out;
```
mayank99, line 52: color transition doesn't count as "motion" so this media query is not needed

src/inputs/checkbox-radio.scss
```diff
@@ -30,31 +29,42 @@
...
-    @include themed {
-      border: 1px solid rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
-      background-color: t(iui-color-background-1);
+    position: relative;
+    isolation: isolate;
+
+    &::after {
+      content: '';
+      position: absolute;
+      inset: 0;
```
mayank99, line 45: You could inherit `border-color` and `border-radius` in here, so that you can set it on the parent for those different variants. This means you will be able to remove `::after` from the selector everywhere else (without even using CSS variables).

```scss
&::after {
  ...
  border-radius: inherit;
  border-color: inherit;
```

backstop/tests/checkbox.html
```diff
@@ -83,11 +83,11 @@ <h2>Default</h2>
               >
                 <path
                   class="iui-check"
-                  d="M6,14L0,8l2-2l4,4l8-8l2,2L6,14z"
+                  d="m6.5 12.5-4.5-4.5 1.5-1.5 3 3 6-6 1.5 1.5z"
                 />
                 <path
                   class="iui-check-partial"
-                  d="m1 6.5h14v3h-14z"
+                  d="m2.75 6.875h10.5v2.25h-10.5z"
```
mayank99, line 90: Why are all the svgs changed? I see no mention in the PR description

backstop/tests/radio.html
```diff
@@ -106,7 +106,7 @@ <h2>Default</h2>
                 <circle
                   cx="8"
                   cy="8"
-                  r="6"
+                  r="4"
```
mayank99, line 109: Do we really need this radius change?

src/inputs/checkbox-radio.scss
FlyersPh9: Without `::after` the focus does not have rounded corners:
Screen Shot 2022-01-07 at 1 51 10 PM

However, I can fix that by doing this:
```scss
.iui-checkbox-checkmark,
.iui-checkbox-checkmark::after {
}
```
&
```scss
.iui-radio-dot,
.iui-radio-dot::after {
  border-radius: 50%;
}
```

src/inputs/checkbox-radio.scss
FlyersPh9: The only benefit I can see for going with 100% is if the user changes the size of the checkbox / radio button, the interior will scale accordingly.  

With 100%:
Screen Shot 2022-01-07 at 1 55 44 PM

With explicit size:
Screen Shot 2022-01-07 at 1 57 35 PM

backstop/tests/checkbox.html
FlyersPh9: The image below shows the new SVGs on the left & the old SVGs on the right.
Screen Shot 2022-01-07 at 2 11 35 PM

We used to take those 16px SVGs and display them at `$iui-sm` (12px).  With the new method, we display the SVG at 100% (which happens to be 16px) so any of the empty space from the checkbox / radio button border need to be handled within the SVG itself.

backstop/tests/radio.html
FlyersPh9: Our old radio dots had a 16px canvas, but the dots were actually 12px.  Then we displayed that 16px artboard at 12px which decreased the dot size to 9px.

With this new method, the 16px canvas is displayed at 100% (which happens to be 16px), so now whatever size we make the radius, it will actually display at that size.  I thought about using `r="4.5"` so that our test images wouldn't change, but decided to go with `r="4"` so that the dot was an even number (8px) instead of odd (9px).

src/inputs/checkbox-radio.scss
FlyersPh9: Your suggestion below solves this, is working now.

src/inputs/checkbox-radio.scss
FlyersPh9: Removed `@media (prefers-reduced-motion: no-preference)`.

src/inputs/checkbox-radio.scss
FlyersPh9: I thought that this might work `border: 1px solid inherit;` but it did not so I did:
```scss
&::after {
  …
  border-style: solid;
  border-width: 1px;
  border-color: inherit;
  … 
}
```

The only one I didn't change was this portion because it also handles `background-color`:
```scss
input:disabled ~ .iui-checkbox-checkmark::after,
input:disabled ~ .iui-radio-dot::after {
  cursor: not-allowed;
  @include themed {
    background-color: t(iui-color-background-disabled);
    border-color: t(iui-color-background-disabled);
  }
}
```

mayank99: background-color doesn't affect the border, so is that really an issue? You might be able to remove `::after` there too.

FlyersPh9: If I remove `::after` the disabled inputs get the background-color applied on active state.
Screen Shot 2022-01-07 at 3 12 57 PM
Screen Shot 2022-01-07 at 3 13 10 PM

FlyersPh9: But it looks like if I move the background-color out of `::after` and into the parent it does work.

FlyersPh9: The only area I am struggling to remove `::after` from at this point is lines 38-41 in `checkbox.scss`.  It must have something to do with specificity because if I remove `::after` and apply `!important` it does work.

mayank99: You can artificially inflate specificity by adding `&:checked, &:indeterminate`

mayank99 (approved):
Looks good. Leaving open the comments about the html changes, so that it's clear what needs to be changed and why

PR #427, opened 2021-12-30
feat(DatePicker): Reduce nesting and specificity

veekeys:
-

### DIFF
--- a/src/date-picker/classes.scss
+++ b/src/date-picker/classes.scss
@@ -5,3 +5,15 @@
 .iui-date-picker {
   @include iui-date-picker;
 }
+
+.iui-calendar-month-year {
+  @include iui-calendar-month-year;
+}
+
+.iui-calender-weekdays {
+  @include iui-calendar-weekdays;
+}
+
+.iui-calendar-week {
+  @include iui-calendar-week;
+}
--- a/src/date-picker/date-picker.scss
+++ b/src/date-picker/date-picker.scss
@@ -14,117 +14,104 @@
   @include themed {
     background-color: t(iui-color-background-1);
   }
-
-  @include iui-calendar;
 }
 
-@mixin iui-calendar {
-  > .iui-calendar {
-    > .iui-header {
-      border: none;
-      padding: 0;
-      position: relative;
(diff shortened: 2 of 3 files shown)


FlyersPh9 (approved):
I pulled and built locally, I can't see any visual changes which is a good thing.  There is still some nesting in the CSS, but most of it seems necessary.  You could maybe unnest `.iui-month` & `.iui-date`.  Otherwise changes seem good to me.

src/date-picker/classes.scss
```diff
@@ -5,3 +5,15 @@
 .iui-date-picker {
   @include iui-date-picker;
 }
+
+.iui-calendar-month-year {
+  @include iui-calendar-month-year;
+}
+
+.iui-calender-weekdays {
```
mayank99, line 13:
```suggestion
.iui-calendar-weekdays {
```

src/date-picker/date-picker.scss
```diff
@@ -14,117 +14,104 @@
...
-          &.iui-today {
-            font-weight: $iui-font-weight-semibold;
-            border-radius: 50%;
-            @include themed {
-              background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-6));
-            }
-          }
+    &:hover {
+      border-radius: $iui-border-radius;
+      font-weight: $iui-font-weight-semibold;
```
mayank99, line 95: @FlyersPh9 We could take this opportunity to rethink/improve hover styles.

The current styling is weird for two reasons:
- current date changes shape for some reason
- other dates have content shift (because of font-weight change)

src/date-picker/classes.scss
veekeys: oh wow, nice catch...changed.
Why my spell checker didnt catch this one :/

src/date-picker/date-picker.scss
veekeys: Would you mind doing this with separate PR? :D

saaaaaally (approved):

mayank99 (approved):

PR #429, opened 2021-12-30
feat(Paginator): Reduce nesting and specificity

veekeys:
- reduced nesting to level of 2
- Added new classes on paginator button group, page button instead of tackling the default ones
- Renamed `iui-ellipsis` to `iui-paginator-ellipsis`
- Ellipsis class could go on the root level as there are no other rules, but `iui-paginator-pages-group` still needs to go under paginator class, because of button-group rules specificity.. But I think thats fine as paginator is pretty specific and I dont think we expect lots of overrides there


src/table/classes.scss
```diff
@@ -28,6 +28,14 @@
 
 .iui-paginator {
   @include iui-paginator;
+
+  .iui-paginator-ellipsis {
+    @include iui-paginator-ellipsis;
+  }
+
+  .iui-paginator-pages-group {
```
mayank99, line 36: > iui-paginator-pages-group still needs to go under paginator class, because of button-group rules specificity

Which rules were causing problems? Button group was freshly refactored so I don't like that this is happening

src/table/classes.scss
```diff
@@ -28,6 +28,14 @@
 
 .iui-paginator {
   @include iui-paginator;
+
+  .iui-paginator-ellipsis {
```
mayank99, line 32: > Ellipsis class could go on the root level as there are no other rules

So move it to root?

> paginator is pretty specific and I dont think we expect lots of overrides there

Keeping the specificity low always helps. We could in the future allow using this paginator outside the table

src/table/paginator.scss
```diff
@@ -78,3 +47,38 @@
...
+
+  > * {
+    margin-right: 0; // unset negative margin from button-group
+  }
+
+  .iui-paginator-page-button {
+    width: $iui-component-height;
+
+    &.iui-small {
+      width: $iui-component-height-small;
```
mayank99, line 76: This approach ties the page-button styles to our base button classes (which might change in the future). It might be better to add a new modifier here (and include the button-sizes mixin to inherit any base styles).

backstop/tests/breadcrumbs.html
```diff
@@ -84,7 +84,7 @@ <h1>Breadcrumbs</h1>
               <svg-chevron-right aria-hidden="true"></svg-chevron-right>
             </li>
             <li class="iui-breadcrumbs-item">
-              <span class="iui-ellipsis">&mldr;</span>
+              <span>&mldr;</span>
```
mayank99, line 87: This class was not doing anything?

veekeys: There are no rules for it, so I guess not /

src/table/classes.scss
veekeys: For example this one
image

I was not aware someone refactored button group.  I thought only Button was taken care of..?

src/table/paginator.scss
veekeys: amm.. so you want me to do same for small and active modifiers? Go like `iui-paginator-button-small` or something like that?

src/table/classes.scss
mayank99: I guess it was only partly refactored (in the Button PR).

For this issue, you could `@include iui-button-group` and target the same selector (`> *:not(:last-child)`). It will lower the specificity for the rest of the rule

src/table/paginator.scss
mayank99: Yeah something like that. It might even make sense to make the `-small` a modifier on the paginator root, because it would allow decoupling the paginator height from the table density

src/table/classes.scss
veekeys: Done something

src/table/classes.scss
veekeys: moved to the root

src/table/paginator.scss
veekeys: done

src/table/paginator.scss
```diff
@@ -57,28 +57,34 @@
...
   > * {
-    margin-right: 0; // unset negative margin from button-group
-  }
-
-  .iui-paginator-page-button {
-    width: $iui-component-height;
-
-    &.iui-small {
-      width: $iui-component-height-small;
+    &:not(:last-child) {
```
mayank99, line 70: I will create a separate PR to lower specificity here after this is merged

image
image

mayank99 (approved):

PR #431, opened 2021-12-31
feat(Tile): Reduce nesting

veekeys:
- Reduced nesting  to max level 3 in Tile by creating separate mixins for sections.
- Renamed classes to include a `.iui-tile-` prefix:
	- `.iui-tile-more-options`
	- `.iui-tile-metadata`
	- `.iui-tile-metadata-icon`
	- `.iui-tile-name`
	- `.iui-tile-description`


src/tile/tile.scss
```diff
@@ -17,303 +17,332 @@
...
-  .iui-thumbnail {
-    width: 100%;
-    height: $iui-baseline * 14;
-    flex-shrink: 0;
-    overflow: hidden;
-    position: relative;
-    display: flex;
-    align-items: center;
-    justify-content: center;
+  .iui-tile-thumbnail-type-indicator {
```
FlyersPh9, line 20: You could replace `.iui-tile-thumbnail-type-indicator` with `&-thumbnail-type-indicator`.

However I'm used to seeing all these mixins referenced within `classes.scss` rather than in the component's SCSS file.  Here's [an example from date-picker](https://github.com/iTwin/iTwinUI/blob/main/src/date-picker/classes.scss).

FlyersPh9: I've made these changes.

veekeys: But forgot to push? :D

FlyersPh9: I ran into a test issue last minute and held off on pushing.  Fixed it today so the changes should be available now.

src/tile/tile.scss
```diff
@@ -11,309 +11,304 @@
...
+    margin: 0;
+    border: initial;
+    border-radius: 0;
+
+    &,
+    &:hover {
       @include themed() {
-        border-right: 1px solid t(iui-color-background-4);
+        border-top: 1px solid t(iui-color-background-4);
+        border-color: t(iui-color-background-4);
```
FlyersPh9, line 304: This was the only CSS that I had to add.  Without this line of CSS, the middle dividing line color was incorrect like so:
Screen Shot 2022-01-26 at 4 02 58 PM

bentleyvk (approved):

bentleyvk:
It would be great to change classes in React ASAP when released.

PR #433, opened 2021-12-31
fix(checkbox): `iui-label` content is centered

gretanausedaite:
image


backstop/tests/checkbox.html
```diff
@@ -599,7 +599,14 @@ <h2>Default</h2>
               </svg>
             </span>
             <span class="iui-label">
-              <svg-smiley-happy-very aria-hidden="true"></svg-smiley-happy-very>
+              <svg
```
gretanausedaite, line 602: Web components acting weird - svg becomes 18-22px height (depends on screen zoom) and can not be centered without changing `height`

veekeys: We need to investigate here more to understand, what is the exact issue. Apart from that, solution looks ok.

mayank99: Have you tried adding `display: flex` to the web components?

gretanausedaite: Yeah, tried that today. `display: flex` on component or `display: block` on svg in component works. If added, we won't need these css changes. Still have to figure out how to add that to build script in components package

mayank99: I'm not talking about adding changes to the package. That's a different (but important) conversation.

What I'm suggesting is adding custom styles to this html file, because this content is not part of checkbox

gretanausedaite: Ok, that's easy.
But it's the same issue in React package (InputGroup radio example). My css changes works without code change on React. Do you suggest adding `display:flex` on SVG in React example?

mayank99: React is not using webcomponents, so it should not be a problem there

gretanausedaite: There is the same problem in React: 
image

I guess web components display style are not the cause of this issue, just one way to solve that. I can change back svg to web components but I'd like to push css changes too.

gretanausedaite: Added `display: flex` to web components

mayank99 (approved):

PR #441, opened 2022-01-11
chore: Updated dependencies to use PostCSS 8

bentleyvk:
Fix for https://github.com/iTwin/iTwinUI/security/dependabot/yarn.lock/postcss/open
image

Ok, I saw this and I don't know what to do...
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,13 +69,15 @@
     "husky": "=4",
     "lint-staged": "^11.2.6",
     "node-sass": "^5.0.0",
-    "postcss": "^8.2.15",
+    "postcss": "^8.4.5",
     "postcss-cli": "^8.3.1",
-    "postcss-discard-comments": "^4.0.2",
+    "postcss-discard-comments": "^5.0.1",
+    "postcss-html": "^1.3.0",
     "prettier": "^2.4.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -69,13 +69,15 @@
     "husky": "=4",
     "lint-staged": "^11.2.6",
     "node-sass": "^5.0.0",
-    "postcss": "^8.2.15",
+    "postcss": "^8.4.5",
     "postcss-cli": "^8.3.1",
-    "postcss-discard-comments": "^4.0.2",
+    "postcss-discard-comments": "^5.0.1",
+    "postcss-html": "^1.3.0",
```
bentleyvk, line 75: image
That why we need `postcss-html` and `stylelint-config-html`.

mayank99: Do we really need postcss-html? Stylelint doesn't help in html files

bentleyvk: So why we have this? 😀
image

mayank99: I dont know, I'm saying maybe we should remove it instead of adding a new dependency

bentleyvk: Removed

mayank99 (approved):

PR #444, opened 2022-01-12
feat(Select): Dropdown icon is an icon

gretanausedaite:
image
Also solves issue from React [#465](https://github.com/iTwin/iTwinUI-react/issues/465)

- `iui-select` is removed and replaced by `iui-input-with-icon`
- `::after` icon is removed from `iui-select-button` and replaced by `iui-end-icon`
- removed border on hover from disabled `iui-input`
- changed `iui-input-icon-inline` margin right to 14px
- added disabled state to `iui-input-icon-inline`
- new bottom border for status in `iui-select-button`


mayank99:
@gretanausedaite Can you link which issues this will fix and mention which ones it wont?

src/inputs/labeled-inputs.scss
```diff
@@ -237,7 +237,19 @@
...
+  width: $iui-icons-default;
+  height: $iui-icons-default;
+  align-self: center;
+
+  @include themed {
+    fill: t(iui-icons-color);
+  }
+
+  &:not(:last-child) {
+    margin-right: $iui-xs;
```
mayank99, line 251: What does this do?

backstop/tests/select.html
```diff
@@ -201,31 +284,71 @@ <h2>Default</h2>
...
+        </span>
       </div>
 
-      <div class="iui-select iui-large">
-        <div
-          tabindex="0"
-          class="iui-select-button"
-        >
-          <span class="iui-content">Large</span>
+      <div class="iui-input-container">
```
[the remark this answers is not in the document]
mayank99: This also creates a weird situation where `LabeledSelect.tsx` is no longer a wrapper around `Select.tsx`

src/inputs/select.scss
```diff
@@ -41,21 +42,6 @@ $iui-after-border: 5px solid;
       }
     }
 
-    &::after {
-      content: '';
```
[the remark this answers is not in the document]
FlyersPh9: Are you suggesting displaying the SVG using `::after`?  I was exploring adding an SVG to the end of a hyperlink using `::after`.  I based the code off of [this article](https://christianoliff.com/blog/styling-external-links-with-an-icon-in-css/):
```scss
.iui-anchor {
  @include iui-anchor;

  &.iui-anchor-popout::after {
    content: '';
    width: $iui-icons-small;
    height: $iui-icons-small;
    margin-left: $iui-xs;
    background-position: center;
    background-repeat: no-repeat;
    background-size: contain;
    display: inline-block;
    background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z'/%3E%3C/svg%3E");
  }
}
```

But I haven't figured out how to pass our color variables through to the icon.  [This link](https://stackoverflow.com/a/27827090) says it is doable in SCSS but we use those weird `t()` things, so I'm not sure how to do so.
Screen Shot 2022-01-12 at 12 35 06 PM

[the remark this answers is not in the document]
mayank99: You might be able to just use `currentColor` to get the value from the text without messing with the `themed()` mixin.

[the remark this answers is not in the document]
FlyersPh9: Unfortunately it appears that does not work, it gives me the same results as above.
```scss
background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='currentColor' d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z'/%3E%3C/svg%3E");
```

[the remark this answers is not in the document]
FlyersPh9: Also just tried applying `color: red` right above the `background-image` CSS within the `::after`, still no luck.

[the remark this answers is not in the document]
FlyersPh9: Found this article just now & playing around with it: [Coloring SVGs in CSS Background Images](https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images)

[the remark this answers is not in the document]
FlyersPh9: I've got this working, but [browser support is poor](https://caniuse.com/css-masks).
```scss
@supports (-webkit-mask-image: url("https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg") or (mask-image: url("https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg"))) {
  -webkit-mask-image: url("https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg");
  mask-image: url("https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg");
  @include themed {
  }
}
```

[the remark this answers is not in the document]
mayank99: Does this work?

```scss
@include themed {
  content: url("data:image/svg+xml<svg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'><path fill=#{t(iui-icons-color-actionable)} d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z'/ ></svg>");
}
```

I'm guessing it will work in light theme but fail in dark theme because it uses CSS variables.

If there is no other solution, lets just go abandon this idea and move on with keeping the svg in html.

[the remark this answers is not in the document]
mayank99: > I've got this working, but [browser support is poor](https://caniuse.com/css-masks).

The prefixed version does work fine across Chrome/Firefox/Safari for our use case, right? Maybe we could use it with my solution as fallback?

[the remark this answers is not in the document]
FlyersPh9: The icon is taking up space but there is no color for light or dark theme.
Screen Shot 2022-01-12 at 1 30 29 PM

backstop/tests/select.html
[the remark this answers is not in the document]
gretanausedaite: So it would be better to add icon in `Select.tsx` to not have label wrapper all the time?

src/inputs/select.scss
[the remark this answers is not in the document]
FlyersPh9: I have the below working.  It uses `content` to display the icon in #008BE1, then within the `@supports` we remove the `content` and display the icon instead using our color variable & `mask-image`.
```scss
&::after {
  content: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill='%23008BE1' d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z'/%3E%3C/svg%3E");
  display: inline-block;
  width: $iui-icons-small;
  height: $iui-icons-small;
  margin-left: $iui-xs;
  background-position: center;
  background-repeat: no-repeat;
  background-size: contain;

  @supports (
    -webkit-mask-image: url('https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg') or (mask-image:
    url('https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg'))
  ) {
    content: '';
    mask-image: url('https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg');
    mask-image: url('https://unpkg.com/@itwin/itwinui-icons/icons/window-popout.svg');
    mask-repeat: no-repeat;
    mask-repeat: no-repeat;
    background-color: var(--iui-color-foreground-primary);
    height: $iui-icons-small + $iui-xxs;
    vertical-align: middle;
  }
}
```

gretanausedaite:
Added `iui-with-icon` mixin. @mayank99 something similar you had in mind? p.s. visuals are not completely working yet.

mayank99:
> Added `iui-with-icon` mixin. @mayank99 something similar you had in mind? p.s. visuals are not completely working yet.

It's so many changes, I don't know what to say if it doesn't even work 😆

Here's what I had in mind:
```html
<input-with-icon>
  <input />
  <icon />
</input-with-icon>
```

But I see you have:
- added a modifier on the `<div class="iui-select">`
- included the `iui-input` mixin in `<div class="iui-select">`

And your solution is using flex, which does not really allow placing elements on top of each other (you need grid or position: absolute).

And it doesn't fix any of the issues with combobox.

I wonder if it'd be better to just use an input instead of div for select. That way, we'd be able to reuse all our CSS and get rid of the overrides. And it will also work with form validation and autocomplete.

src/inputs/select.scss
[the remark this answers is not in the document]
mayank99: Worth mentioning: If we change Select to use `input` instead of `div`, we won't be able to use a psuedo-element (CSS limitation). Same problem in ComboBox. So I guess this discussion is not productive.

backstop/tests/select.html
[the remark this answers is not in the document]
gretanausedaite: Removed select classes and using only `input-with-icon` now

src/inputs/classes.scss
```diff
@@ -13,6 +13,10 @@
     @include iui-input-icon;
   }
 
+  > .iui-message-icon {
```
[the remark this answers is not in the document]
gretanausedaite: removed

src/inputs/labeled-inputs.scss
```diff
@@ -284,7 +296,7 @@
 /// @arg status - must be one of: positive, negative, warning
 /// @arg iconSelector - selector to apply status fill on. Defaults to .iui-input-icon
```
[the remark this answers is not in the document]
gretanausedaite: updated

gretanausedaite:
Select updates:
- Created new `iui-input-with-icon` component
- Created new `iui-end-icon` (can be renamed) component for end icon to component above
- Added `iui-input-with-icon` to `labeled-input` grid
- Updated HTML and images

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+// Mixin to have inputs combined with icon (eg. select)
+@mixin iui-input-with-icon {
+  @include iui-input;
+  display: grid;
+  align-items: center;
+  grid-template: 'input icon' / 1fr auto;
+  margin: 0;
+  padding: 0;
+  vertical-align: baseline;
+  position: relative;
```
mayank99, line 15: Don't think this does anything

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+@import './input';
+
+// Mixin to have inputs combined with icon (eg. select)
+@mixin iui-input-with-icon {
+  @include iui-input;
+  display: grid;
+  align-items: center;
+  grid-template: 'input icon' / 1fr auto;
+  margin: 0;
+  padding: 0;
```
mayank99, line 13: Might want to include `iui-reset` instead.

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+  vertical-align: baseline;
+  position: relative;
+  width: 100%;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: border-color $iui-speed-fast ease-out;
```
mayank99, line 18: This doesn't need the motion media query because it's just transitioning a color in-place

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+        @include themed {
+          background: t(iui-color-background-disabled);
+          fill: t(iui-icons-color-actionable-disabled);
+        }
+      }
+    }
+  }
+}
+
+@mixin iui-end-icon {
```
mayank99, line 69: Would it be possible to avoid some of the duplication by including the `iui-input-icon-inline` mixin here?

src/inputs/labeled-inputs.scss
```diff
@@ -319,9 +319,9 @@
     }
   }
 
-  > .iui-input,
-  > .iui-textarea,
-  > .iui-select > .iui-select-button {
+  .iui-input,
+  .iui-input-with-icon,
```
mayank99, line 323: I think these `padding-bottom` declarations are breaking layout of LabeledSelect

src/inputs/input.scss
```diff
@@ -2,6 +2,8 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 
+$iui-after-border: 5px solid;
```
mayank99, line 5: What does this do?

src/inputs/select.scss
```diff
@@ -3,113 +3,38 @@
...
   @include iui-reset;
   display: flex;
-  flex-direction: column;
+  align-items: center;
+  user-select: none;
   position: relative;
-  width: 100%;
-  min-width: 0;
-
-  > .iui-select-button {
```
mayank99, line 16: From what I understand, your new styles are changed like this:
- `iui-select` becomes `iui-input-with-icon`
- `iui-select-button` becomes `iui-select`
- `iui-select-button::after` becomes `iui-end-icon`
- `iui-content` and `iui-icon` are preserved

Does that sound right?

Once you make the overlapping grid change mentioned in my other comment, I think you would need to revert some of that structure back (e.g. `iui-input-with-icon` will have no border).

backstop/tests/menu.html
```diff
@@ -703,298 +717,340 @@ <h3>Select</h3>
...
-              </span>
-            </li>
-          </ul>
+              <path
+                d="m4.8067 6h6.3953a.27961.27961 0 0 1 .24043.44321l-3.1736 3.45707a.33969.33969 0 0 1 -.48085 0l-3.2217-3.45707a.26909.26909 0 0 1 .24042-.44321z"
+              ></path>
+            </svg>
+          </span>
         </div>
+        <ul
```
mayank99, line 746: I think this menu needs to be outside the `label` to fix the responsive grid issue. However, I think that's just because we are not using a popover here in the html demo. In React, it should work fine.

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  &.iui-disabled {
+    @include themed {
+      background-color: t(iui-color-background-disabled);
+      border-color: t(iui-color-background-disabled);
+    }
+
+    .iui-end-icon {
```
mayank99, line 55: It looks like we are not really changing any styles on this icon when it's disabled. So if you just use JS to remove the `iui-actionable` class when disabled, then this CSS will not be needed. See `ComboBox.tsx:316` for how that looks in React.

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+      cursor: not-allowed;
```
mayank99, line 56: I think this cursor styling is in the wrong place. It seems that `iui-input-with-icon` is just missing the entire cursor styling, so if you go in the demo and hover over non-labeled selects, then you get the default cursor.

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+  @media (prefers-reduced-motion: no-preference) {
+    transition: border-color $iui-speed-fast ease-out;
+  }
+
+  &:hover {
+    @include themed {
+      border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+    }
+
+    &::after {
```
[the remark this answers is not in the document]
gretanausedaite: removed `::after`

src/inputs/input-with-icon.scss
gretanausedaite: removed position

src/inputs/input-with-icon.scss
gretanausedaite: removed media query

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,107 @@
...
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import './input';
+
+// Mixin to have inputs combined with icon (eg. select)
+@mixin iui-input-with-icon {
+  @include iui-input;
+  display: grid;
+  align-items: center;
+  grid-template: 'input icon' / 1fr auto;
```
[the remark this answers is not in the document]
gretanausedaite: I thought you wanted to use current labeled-inputs grid for ComboBox,
image
Did I misunderstood this example? (last code part)

src/inputs/labeled-inputs.scss
gretanausedaite: Why do you think that? Example?

src/inputs/input.scss
gretanausedaite: Nothing, removed

backstop/tests/menu.html
gretanausedaite: Fixed

src/inputs/input-with-icon.scss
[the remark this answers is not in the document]
mayank99: Oh, that last example is just for a labeled-input with a button (https://itwin.github.io/iTwinUI-react/?path=/story/input-labeledinput--hybrid-layout-with-button)

The middle example is for combobox. (just sub svg with button too I guess)

There are two reasons we want to use the same solution for select and comobobox (other than just consistency and reducing duplication):
1. allow message icon in combobox
2. fix combobox icon bugs

src/inputs/labeled-inputs.scss
mayank99: Notice how the text and icon are not vertically centered
image

gretanausedaite:
- `iui-input-with-icon` is used for positioning and coloring `disabled` content in it.
- Using `:first-child` for adding `grid-area` positioning to children. Did not want to specify `iui-select` or `iui-input` classes.
- Added `iui-input` styles to `iui-select`
- Added status border coloring to `iui-input-with-icon` first child

gretanausedaite:
Last commit solves:
- change icon and first child grid-area to 1/1
- removed grid areas from input-with-icon
- move sizing into select-button
- rename select to select-button
- remove all code from end-icon
- remove iui-input from select-button
- remove iui-actionable and disabled code from iui-with-input (added disabled colors to inline icon in labeled-inputs.)

still need to work on:
- height (39 px -> 38px)
- cursor stuff
- box sixing on icon

gretanausedaite:
Fixed:
- height (39 px -> 38px)
- cursor stuff (?)
- box sixing on icon

src/inputs/input-with-icon.scss
mayank99: The cursor styling is mostly fixed now, but the icon is now showing the `pointer` cursor even when disabled

src/inputs/labeled-inputs.scss
mayank99: ok so now i'm not sure about this pseudo-element approach again, as it looks really off-center in small and large sizes
image

src/inputs/select.scss
```diff
@@ -3,113 +3,75 @@
 @import '../style/index';
 @import '../menu/index';
 
-$iui-after-border: 5px solid;
+@mixin iui-select-button {
+  @include iui-focus($offset: -2px, $thickness: 2px);
```
mayank99, line 7: Maybe we should move focus styling to the parent and use `focus-within` so that it's consistent with combobox

gretanausedaite: adding focus to parent could be difficult as `iui-select-button` have background color and border which cover `iui-input-with-icon` borders and outline.

src/inputs/input-with-icon.scss
gretanausedaite: Added cursor styling to iui-input-icon-inline disabled state

src/inputs/input-with-icon.scss
gretanausedaite: removed iui-actionable on disabled icons

src/inputs/select.scss
mayank99: `focus-within` on select-button doesn't do anything because the menu will not be inside the button.

You can target `:focus-within` on the parent and apply the actual styles to the child

gretanausedaite: moved focus-within to input with icon

src/inputs/labeled-inputs.scss
```diff
@@ -344,4 +356,30 @@
...
+    ::after {
+      content: '';
+      width: 100%;
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      display: block;
+      height: $iui-xxs;
+      opacity: 1;
+      width: 100%;
```
mayank99, line 372: Can be cleaned up
```suggestion
    &::after {
      content: '';
      width: 100%;
      position: absolute;
      bottom: 0;
      left: 0;
      height: $iui-xxs;
```

src/inputs/labeled-inputs.scss
gretanausedaite: @FlyersPh9 for some advice.
Short update: We removed padding and border to `iui-select-button` and added `::after` to add colored bottom border. All items are flex centered and it appears not centered.

mayank99 (approved):

src/inputs/select.scss
mayank99: I think it wont work with status inputs (or maybe it will show double outline). You might want to link this repo to React, make the necessary changes, and see what happens.

My guess is that you would need to remove `iui-focus` mixin in `select.scss` and remove `:focus` styling for `iui-select-button` in `labeled-inputs.scss` and change it to just outline-color
image

src/inputs/select.scss
```diff
@@ -3,113 +3,83 @@
...
+    text-overflow: ellipsis;
+  }
+
+  &.iui-placeholder {
+    @include themed {
+      color: t(iui-text-color-placeholder);
+    }
+
+    .iui-icon {
+      fill: currentColor;
```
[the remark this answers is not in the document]
gretanausedaite: moved to base styling

src/inputs/select.scss
```diff
@@ -3,113 +3,83 @@
...
     padding-bottom: $iui-component-padding-vertical-large;
     min-height: $iui-component-height-large;
     font-size: $iui-font-size-leading;
   }
+
+  &:hover {
+    @include themed {
+      border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+    }
+    @media (prefers-reduced-motion: no-preference) {
```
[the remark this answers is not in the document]
gretanausedaite: removed

src/inputs/select.scss
```diff
@@ -3,113 +3,83 @@
...
 
-  &.iui-large .iui-select-button {
+  &.iui-large {
     padding-top: $iui-component-padding-vertical-large;
     padding-bottom: $iui-component-padding-vertical-large;
     min-height: $iui-component-height-large;
     font-size: $iui-font-size-leading;
   }
+
+  &:hover {
```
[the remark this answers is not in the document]
gretanausedaite: moved to parent

backstop/tests/column-filter.html
```diff
@@ -43,14 +43,28 @@ <h2 id="default">Text filter</h2>
...
               id="test-select-1"
               class="iui-select-button"
             >
               <span class="iui-content">Starts with</span>
             </div>
+            <span
+              class="iui-end-icon iui-actionable"
+              aria-hidden="true"
+            >
+              <svg
```
FlyersPh9, line 58: I believe we should be able to use a web component icon instead:
```html
<svg-caret-down-small
	class="iui-end-icon iui-actionable"
	aria-hidden="true"
></svg-caret-down-small>
```

You'll need to include this in the `<head>`:
```html
<script
	type="module"
	src="https://cdn.skypack.dev/@itwin/itwinui-icons-elements/caret-down-small"
></script>
```

src/inputs/input-with-icon.scss
```diff
@@ -0,0 +1,35 @@
...
+  &:focus-within {
+    > :first-child {
+      outline: 2px solid var(--_focus-color, var(--iui-color-foreground-primary));
+      outline-offset: -2px;
+    }
+  }
+
+  &:hover {
+    > :first-child {
+      border-color: var(--_hover-color, rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2)));
```
FlyersPh9, line 26: Interesting, I've never seen a CSS variable like this before.  Can you explain what this is doing?

veekeys: Can you check if it even works? we should not use `t()` without `@include themed` ...?

mayank99: It's just using the CSS variable `--_hover-color`. If that variable is not found, then it's using the default value.

We set that variable in select.scss

mayank99: It does work, I think `t()` just converts to `var()` if not inside themed mixin. But you're right, we should change it to `var()`

src/inputs/input-with-icon.scss
mayank99: I see you've added back `iui-disabled` on the icon for the cursor. It can be removed if you do this instead:
1. apply `pointer-events: none` on the base icon itself
2. apply `pointer-events: auto` inside `iui-actionable` modifier

FlyersPh9:
On line 27 can we remove the `width` on the `<section>` and instead set `viewports` within `select.js`?  It looks bad on mobile.
IMG_A0C17977D23A-1

src/inputs/input-with-icon.scss
gretanausedaite: moved disabled to `iui-input-with-icon` because end-icon needs disabled colors. Also added pointer-events: none there for disabled combobox.

backstop/tests/column-filter.html
gretanausedaite: Changed icons to web components.

mayank99 (approved):
Looks good 🍕

PR #446, opened 2022-01-12
feat(Tile): Add focus styling to entire tile

### LINKED ISSUE #442: Tile: focus states
1. Tile is currently missing a custom focus state, so it shows the default browser outline if `tabindex="0"` is added to `div.iui-tile`

2. The "more options" button is hidden unless hovered, so it cannot be tabbed into. This could be fixed using a "visually hidden" that keeps the content accessible to screen readers.

FlyersPh9:
- Add `:focus` to entire tile.
- Increase outline when `:focus` & `.iui-selected` are combined.
- Use [this visually hidden method](https://github.com/iTwin/itwin-synchronization-report-ui/blob/main/src/components/utils/VisuallyHidden.scss) to allow the "more options" button to be focused.
- `.iui-tile-more-options` is now a wrapper div instead of applied on the button directly.
- Removed some unnecessary CSS.
- Replaced map `iframe` with an image since it was visually broken. 

Closes #442.

**Focus on 1st tile, selection on 3rd tile**
Screen Shot 2022-01-12 at 2 07 18 PM

**Focus & selection on 3rd tile**
Screen Shot 2022-01-12 at 2 07 28 PM

**Map image instead of iframe**
Screen Shot 2022-02-09 at 12 33 30 PM


FlyersPh9:
Downgrading this to draft until I can figure out why my test images are failing.

FlyersPh9:
I'm getting an error with my test image results:
```
Puppeteer encountered an error while running scenario "tile_State More Options hover"
Error: Node is either not clickable or not an HTMLElement
```
The ID is there, it should be clickable.

src/tile/tile.scss
```diff
@@ -211,10 +211,24 @@
...
-      visibility: hidden;
+
+      clip-path: inset(50%);
+      overflow: hidden;
+      white-space: nowrap;
+      height: 1px;
+      width: 1px;
+      padding: 0;
+      margin: -1px;
+      border-width: 0;
```
mayank99, line 222: Why not a reusable mixin for this?

src/tile/tile.scss
```diff
@@ -211,10 +211,24 @@
...
 
       &.iui-visible {
```
mayank99, line 224: When is this `iui-visible` class applied? The more-options button should really only be shown when the tile is hovered or has focus within.

veekeys:
Can we do this PR after #431 ?

FlyersPh9:
> Can we do this PR after #431 ?

@veekeys yes, I will move this PR back to draft in the meantime.

veekeys:
> > Can we do this PR after #431 ?
> 
> @veekeys yes, I will move this PR back to draft in the meantime.

#431 is merged, so we can proceed with this one @FlyersPh9

src/tile/tile.scss
FlyersPh9: Created `@mixin iui-visually-hidden`.  Not sure if we can provide a reusable mixin for reverting because we don't know what values they should have for border, margin, padding, etc.

mayank99: The solution to that is apply it on a wrapper div instead.

See https://github.com/iTwin/synchronization-report-react/blob/main/src/components/utils/VisuallyHidden.scss

src/tile/tile.scss
bentleyvk: I think we added this class to show more-options when it is clicked and dropdown is shown. 
image
Without this class:
image

src/tile/tile.scss
bentleyvk: Any source to recommendation for doing this?

gretanausedaite (approved):
I don't have any comments, looks good. Just needs a branch update. 🌵

src/tile/tile.scss
mayank99: - https://webaim.org/techniques/css/invisiblecontent/#:~:text=receive%20keyboard%20focus.-,css%20clip,-%7Bclip%3A%20rect
- https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html
- https://www.a11yproject.com/posts/how-to-hide-content/

src/tile/tile.scss
mayank99: Oh, for some reason I thought this was a new addition. In that case, we should undo the visually hidden styles inside this class

FlyersPh9: `.iui-visible` is back.

src/tile/tile.scss
FlyersPh9: If I did a wrapper div, does something like this make sense within `global.scss`:
```scss
.iui-visually-hidden {
  @mixin iui-visually-hidden;

  &:focus-within {
		@mixin iui-visually-hidden-revert;
	}
}
```

src/style/mixins.scss
```diff
@@ -130,6 +130,19 @@
   padding: 0;
 }
 
+/// Visually hides an element but still allows it to be tabbed to.
+@mixin iui-visually-hidden {
```
bentleyvk, line 134: Shouldn't we have undo mixin? Like in https://github.com/iTwin/synchronization-report-react/blob/main/src/components/utils/VisuallyHidden.scss

src/tile/tile.scss
```diff
@@ -153,13 +161,18 @@
...
     @include iui-button-size(small, borderless);
-    position: absolute;
+    @include iui-visually-hidden;
     bottom: $iui-baseline * 0.5;
     right: $iui-s;
-    visibility: hidden;
 
     &.iui-visible {
-      visibility: visible;
+      clip-path: revert;
```
bentleyvk, line 169: Maybe have this part in mixin and reuse inside Tile css

src/tile/tile.scss
mayank99: Might be able to avoid a wrapper div with something like this:
```scss
.iui-tile-more-options {
  // ... base styles

  &:not(:focus):not(.iui-visible) {
    @include iui-visually hidden;
  }
}
```

FlyersPh9: Ugh…I read this _after_ adding the wrapper. 😓 Check what I have now and tell me if I should revert it.

Worth noting, I can not figure out why the `State More Options hover` scenario test is failing.

src/style/mixins.scss
FlyersPh9: I've added the revert mixin.

src/style/mixins.scss
```diff
@@ -130,6 +130,30 @@
...
+  position: absolute;
+  white-space: nowrap;
+  height: 1px;
+  width: 1px;
+  padding: 0;
+  margin: -1px;
+  border-width: 0;
+}
+
+@mixin iui-visually-hidden-revert {
```
mayank99, line 146: Any reason you added the `iui-` prefix? I would think this should be an internal mixin.

Also maybe add a comment above this mixin explaining what it does and that it is recommended to use on wrappers.

mayank99 (approved):
The behavior look good, but I'm not sure about the mixin name.

src/style/mixins.scss
FlyersPh9: Removed `iui-` prefix & added comments.

Issue #391, opened 2021-11-19
Disabled elements should be interactable for a11y


mayank99:
> when a button is disabled, it can no longer show a tooltip on hover.  I think if we offer a class name that makes a button appear disabled, we should be able to show a tooltip on hover / focus.

_Originally posted by @FlyersPh9 in https://github.com/iTwin/iTwinUI/pull/363#discussion_r744888686_

mayank99:
Another use for `aria-disabled` is when showing a loading spinner inside the button.

https://user-images.githubusercontent.com/9084735/224162889-38720f1f-5c20-46a3-af45-d0dd60ba147a.mov

When loading, the button will become disabled. The focus should stay on the button for better accessibility. This will be useful when the button is used to open a dialog (the focus will return to button when dialog closes) and also if the loading fails, keyboard users can easily press <kbd>Enter</kbd> again.

FlyersPh9:
It looks like [the button CSS already has `aria-disabled`](https://github.com/iTwin/iTwinUI/blob/15c480d7cb321d1dd13b5359581c1c888a733f81/packages/itwinui-css/src/button/button.scss#L61), so do we just need to add support for this in React?

mayank99:
yes, we need to manually prevent clicks when using aria-disabled

PR #448, opened 2022-01-13
fix(ButtonGroup): Don't target `<input>` inside labeled-input to fix z-index issues

### LINKED ISSUE #447: LabeledInput custom icon vanishes
When hosting a LabeledInput inside a ButtonGroup, the inputs custom icon vanishes as soon as the mouse pointer enters the input text area.

#### Expected behavior
Custom icon stays visible inside the control (is drawn on top).

#### Reproduction
https://codesandbox.io/s/itwinui-react-buttongroup-hosting-labeledinput-and-icon-vanished-on-hover-f152x

##### Steps to reproduce
1. Enter the input near the text "Test"
2. "Close" icon no longer displays

#### Additional information
Working in itwin/itwinui-react@1.27

mayank99:
Labeled inputs are wrapped in a div, so setting z-index on the input was causing problems with the icon disappearing. I was able to use `:not` to target inputs that are not inside input-container, and then target input-container separately. This fixes #447

Also, table paginator no longer uses button-group as it doesn't need to handle all the z-indices.

Also, active button stripe is now handled using CSS variables as hooks. The styling is mostly contained in `button.scss` but positioning needs to be specified using `--_iui-button-active-stripe-inset` (from outside). There is no default value for `--_iui-button-active-stripe-inset` because we don't want to show the stripe by default, so I've set it to `initial`.


FlyersPh9 (approved):

mayank99:
This is a tough one. My current solution breaks disabled buttons because now they get `z-index: 1` on hover.

I tried increasing the z-index of the icon when the input is hovered/focused like this:
```scss
.iui-input-container.iui-inline-icon {
  > .iui-input {
    &:hover,
    &:focus {
      + .iui-input-icon {
        z-index: 2;
      }
    }
  }
}
```

But this is a leaky solution which could have side effects (and it has very high specificity). If I add `isolation: isolate`, it will stop working with button group which relies on its own stacking context.

I think the root of the problem lies in this line below, as it tries to reach inside other components and modify them:
https://github.com/iTwin/iTwinUI/blob/08b4aa30fd15ad4d3abd2be01bb25bee48d4fad7/src/button/button-group.scss#L10

This is made worse by the fact that we can't currently query for a descendant state and apply styles on the parent (i.e. `has()` selector).

I'm starting to think we might need to refactor button-group again to remove the use of z-index.

src/button/button-group.scss
```diff
@@ -12,20 +12,18 @@
...
+    }
 
-      &:hover {
-        z-index: 1;
-      }
-
-      &:focus {
-        z-index: 2;
+    &:focus-within {
+      position: relative;
```
veekeys, line 21: Why position is required here?

mayank99:
I've changed the approach a bit to target `.iui-button-group-item` which is applied on the wrapper div. See https://github.com/iTwin/iTwinUI-react/pull/524 for how I envision it being used in React.

`$childSelector` is now only used for changing border radii.

src/button/button-group.scss
mayank99: Dont remember why it was needed but I've changed the approach now.

backstop/tests/header.html
```diff
@@ -641,7 +641,7 @@ <h2>
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
               <span class="iui-button-split-menu iui-header-split-button">
-                <div>
+                <div class="iui-button-group-item">
```
veekeys, line 644: Do we really need here button group? It is split button and we know how its gonna be (no possible customizations). So maybe instead we could just have those few needed rules in that button css and do not depend on button group mixin? 
This class looks out of context now.

backstop/tests/table.html
```diff
@@ -92,46 +92,64 @@
     ></script>
 
     <script>
-      customElements.define('svg-folder', class extends HTMLElement {
-        connectedCallback() {
-          this.innerHTML = `<svg viewBox="0 0 16 16"><path d="M8,3,6.5856,1.5857A2,2,0,0,0,5.1715,1H1A1,1,0,0,0,0,2V5H16V4a1,1,0,0,0-1-1Z" fill="#e7b93a" /><path d="M0,14a1,1,0,0,0,1,1H15a1,1,0,0,0,1-1V5H0Z" fill="#f6d97a" /></svg>`;
+      customElements.define(
+        'svg-folder',
```
veekeys, line 96: oh wow, formatter was not run before? :O

backstop/tests/table.html
```diff
@@ -382,16 +396,13 @@ <h2>Default</h2>
...
-            </div>
+            >Yesterday</div>
             <div class="iui-cell">
               <div class="iui-cell-start-icon">
                 <user-icon
                   size="small"
                   type="1"
-                >
-                </user-icon>
+                > </user-icon>
```
veekeys, line 405: why empty space appeared here :?

backstop/tests/header.html
veekeys: Also noticed, in split menu, we are importing button group mixin twice :O

src/button/button-group.scss
```diff
@@ -9,25 +9,21 @@
...
-          z-index: revert;
-        }
-      }
+    &:focus-within {
+      z-index: 2;
+    }
 
-      &:disabled {
+    > * {
+      &:disabled,
```
veekeys, line 25: So now hover is broken.
image

image

backstop/tests/button.html
```diff
@@ -1392,29 +1396,56 @@ <h3>Input + button combo group</h3>
...
             />
           </div>
         </div>
-        <div><button class="iui-button iui-high-visibility">Copy</button></div>
+        <div class="iui-button-group-item"><button class="iui-button iui-high-visibility">Copy</button></div>
+      </div>
+
+      <br />
+
+      <div class="iui-button-group">
```
veekeys, line 1426: I dont know, maybe I just start to be picky on not related things, but is this focus style supposed to be like that?
image
image
image

Notice different border on the close icon. Is it the expected styling for bordered vs borderless components?
@FlyersPh9

mayank99: Yes it is expected to be like this and has nothing to do with this PR
- Bordered components go from 1px border to 2px outline
- Borderless components go from no border to 1px outline

src/button/button-group.scss
mayank99: Should be fixed now, I moved it to the same level as `disabled`

backstop/tests/header.html
mayank99: we could remove button-group here, but then we should also do that in the base split-button

backstop/tests/table.html
mayank99: I think the vscode formatter is ok with both formats.

I removed the extra div in paginator using regex find and search, and it had a bunch of empty lines which the vscode formatter was not removing, so I had to run prettier which caused all these other changes

backstop/tests/header.html
veekeys: Yeah, so thats what I am asking. Cant we do same with split button as you did with paginator?

backstop/tests/button.html
veekeys: Ok, I know its not related, that why I said in my comment "not related things".
Just wanted to confirm, resolving

src/button/button-group.scss
```diff
@@ -9,25 +9,21 @@
...
-        z-index: 2;
-
-        &:where(:not(:focus-visible):not(:hover)) {
-          z-index: revert;
-        }
-      }
-
-      &:disabled {
+      &:disabled,
+      &[aria-disabled='true'] {
```
veekeys, line 26: also thought, do we everywhere check `aria-disabled` where we check for disabled status? Personally, do not remember this in other places. If not, should we have this recorded to do one day?

mayank99: I added this to be future-proof. I think I should also add `iui-disabled` because we use that in labeled-input

backstop/tests/header.html
mayank99: With paginator it was easy because we don't need any of the custom button-group logic.

For split-button, we need the border radius handling so I would have to duplicate that part. That's not really the point of this PR, so I would rather do it in a separate PR.

src/button/button-group.scss
FlyersPh9: `aria-disabled` is only present in `button-group.scss` as of this PR.  I think it makes sense to include this in other places.  Here is a list of where we are referencing `:disabled` and could probably include it:
- /button/borderless.scss
- /button/button.scss
- /button/cta.scss
- /button/default.scss
- /button/high-visibility.scss
- /button/split-menu.scss
- /header/header.scss
- /inputs/checkbox-radio.scss
- /inputs/checkbox.scss
- /inputs/radio-tile.scss
- /side-navigation/side-navigation.scss
- /toggle-switch/toggle-switch.scss

I also agree with including `.iui-disabled`, that as well as the aria method may help us achieve `:focus` on a disabled button which has been requested in the past.

src/button/button-group.scss
```diff
@@ -61,11 +57,11 @@
...
         }
       }
+    }
 
-      // Reset border radius only for non-borderless items
-      &:not(.iui-borderless) {
-        border-radius: 0;
-      }
+    // Reset border radius only for non-borderless items
+    #{$childSelector}:not(.iui-borderless) {
```
FlyersPh9, line 63: This is interesting.  I see `$childSelector: 'input, button';` above.  Does that mean the SCSS is:
```scss
input, button:not(.iui-borderless) {
```

Or does it mean:
```scss
input:not(.iui-borderless),
button:not(.iui-borderless) {
```

backstop/tests/header.html
veekeys: Well, this PR changes the split button, which is not nice. It can be done with separate PR, but Id prefer this not to be released without split buttons fix.

src/button/button-group.scss
```diff
@@ -61,28 +55,35 @@
...
         }
       }
+    }
 
-      // Reset border radius only for non-borderless items
-      &:not(.iui-borderless) {
-        border-radius: 0;
-      }
+    // #region Borders
+    $childSelector: 'input, .iui-button';
```
bentleyvk, line 61: Why one is targeting base element and other our class?

mayank99: i changed `button` to `.iui-button` because it can be rendered as a link

src/button/button-group.scss
veekeys: Now seems we ended up in inconsistencies.. one place we target disabled and aria-disabled, but on line 50 it is only disabled state...
That part also has too deep nesting for our future refactors..

bentleyvk: Why `aria-disabled` is even needed here?

backstop/tests/button.html
```diff
@@ -1172,12 +1176,12 @@ <h3>Borderless</h3>
     <h3>Split menu</h3>
     <section id="demo-split-menu">
       <span class="iui-button-split-menu">
-        <div>
+        <div class="iui-button-group-item">
```
bentleyvk, line 1179: Looks weird that we are using button group item class here but container is split button and there is nothing else from button group.

src/button/button-group.scss
bentleyvk: So don't we want to be consistent and have CSS selectors here instead?

mayank99: What do you mean by "CSS selectors"?

backstop/tests/button.html
mayank99: image

src/button/button-group.scss
mayank99: > Now seems we ended up in inconsistencies

made the disabled selector consistent by adding `iui-disabled` and storing it in a variable

> Why aria-disabled is even needed here?

See discussion above. It will help with #391

veekeys: Didnt want to check anything about the nesting? Anyway you are changing a bunch of stuff here, wouldnt need to come back

src/button/button-group.scss
bentleyvk: Ups, wanted to say classes

mayank99: Updated the selectors, take a look

mayank99:
I have once again changed the approach in [`b3414b5`](https://github.com/iTwin/iTwinUI/pull/448/commits/b3414b5120f6679c3605245044eddc33f1554e33#diff-537d2d290f6adc5fed41e625013b22f69b851b4280a811526b5c89f34c35ac64). Now we are back from `<div class="iui-button-group-item">` to just `<div>` because it was determined that this new class would be problematic to support in a backwards-compatible way in React.

For making it work, I had to change the selectors a little bit. `.iui-input` is now accompanied by a complex selector which allows us to achieve conditional z-indexes. Here's what this means:
- If `iui-input` is inside a `iui-input-container`, then only the container will get z-indexes. This fixes the linked issue.
- If `iui-input` is by itself, then the z-index will be applied on the actual input
- Border radius handling is not affected by this selector as it is done separately

backstop/tests/button.html
mayank99: This class is removed so it's no longer a concern

backstop/tests/header.html
mayank99: This class is removed so it's no longer a concern

src/button/button-group.scss
```diff
@@ -7,37 +7,29 @@
   align-items: center;
   isolation: isolate;
 
-  $childSelector: 'input, button'; // target nested inputs and buttons
+  $childZindexSelector: '.iui-input-container, .iui-button, .iui-input:where(:not(.iui-input-container .iui-input))';
```
bentleyvk, line 10: Can you explain or paste a link to explaination, how this is working `.iui-input:where(:not(.iui-input-container .iui-input))`? Because I can't find why this magic is working 😀

src/button/button-group.scss
```diff
@@ -7,37 +7,29 @@
...
+  $childBorderSelector: '.iui-input, .iui-button:not(.iui-borderless)';
+  $disabledSelector: '&:disabled, &.iui-disabled, &[aria-disabled="true"]';
```
veekeys, line 12: need examples with disabled elements in button-group

src/button/button-group.scss
```diff
@@ -7,37 +7,29 @@
...
 
   > * {
     display: flex;
 
-    #{$childSelector} {
+    #{$childZindexSelector} {
       position: relative;
```
veekeys, line 18: could you remind again, why this is needed? something with z-index? Cause in certain elements it gets overriden..

src/button/button-group.scss
mayank99: we already have some
image

src/button/button-group.scss
mayank99: what do you mean overridden?

For z-index to work, you either need `display: flex` (or grid) on parent or `position: relative`. In this case, it's needed in case the *direct* parent is not flex (e.g. nested elements)

src/button/button-group.scss
veekeys: I meant examples with mixed group (buttons + inputs)

src/button/button-group.scss
veekeys: Hmm.. I am sure I saw element where that position was overriden from other rules, but now quickly could not find it anymore :/

src/button/button-group.scss
mayank99: > Can you explain or paste a link to explaination

What part is not clear? It is targeting `iui-input` only if it is NOT a descendant of `iui-input-container`. And it is also targeting `iui-input-container` separately.

> specificity of this one is really big

Anything inside [`:where`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) does not count towards specificity, so that part is fine (the resulting specificity is 0-3-0).

The only place where it has huge specificity is for active stripe. I don't know what we can do about that.
image

src/button/button-group.scss
```diff
@@ -7,77 +7,55 @@
...
-        }
       }
 
-      &:disabled {
+      #{$disabledSelector} {
         z-index: -1;
       }
 
+      // Adds stripe above active button
+      --_active-stripe-inset: #{$iui-xxs} #{$iui-xxs} calc(100% - #{$iui-xs});
```
bentleyvk, line 34: So once again we are removing functionality from IE...

bentleyvk, line 34: What is this naming? Any link to this convention? No need for `iui-button-*0?`

src/button/button.scss
```diff
@@ -50,6 +50,23 @@ $iui-button-padding-large: $iui-xs * 6;
...
+  &.iui-active {
+    // Shows active stripe only if --_active-stripe-inset is set (from outside)
+    &::after {
+      content: '';
+      position: absolute;
+      inset: var(--_active-stripe-inset);
+      background-color: var(--iui-color-foreground-primary);
+    }
+
+    &[disabled],
```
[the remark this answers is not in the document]
mayank99: `[disabled]` is used for anchor, `:disabled` is used for fieldset

src/button/button-group.scss
mayank99: It's not an external variable that the user should be using, so I prefer this format. The `_` also indicates that it's private

mayank99: The code and specificity benefit seems worth it to me

src/button/button-group.scss
mayank99: stylelint should pass now

src/button/button-group.scss
veekeys: I dont like this. Why we cannot name everything in our library with the prefix `iui` ? Why initiate some new standards, which we have never talked about.

src/button/button.scss
```diff
@@ -50,6 +50,23 @@ $iui-button-padding-large: $iui-xs * 6;
   &:disabled {
     @include iui-button-disabled;
   }
+
+  &.iui-active {
+    // Shows active stripe only if --_active-stripe-inset is set (from outside)
+    &::after {
+      content: '';
+      position: absolute;
+      inset: var(--_active-stripe-inset);
```
veekeys, line 59: Do you have some resources to read about this approach? Is it a new way of writing css or a new way to hack around complicated selectors?
Just I have the concern that it might start to get complicated to catch all the parts if we gonna introduce this in more places...

src/button/button-group.scss
veekeys: Most of selectors now how 0-3-0 and there is a bunch of 0-4-0
image

Do no remember, what is the specificity we are trying to achieve in most places.

src/button/button-group.scss
mayank99: `iui-` to me sounds like it's a public variable. What benefit does it provide?

src/button/button-group.scss
mayank99: ok yes forgot about that `borderless` selector, so it will be 0-4-0 in a few places. I could add `:where` to make it 0-3-0 if you want, or we could leave it like this

src/button/button-group.scss
mayank99: Added

src/button/button-group.scss
mayank99: added prefix

veekeys (approved):
Lets hope nothing will get worse

gretanausedaite (approved):
Looks good 🌵 
Off topic:
Why borderless active button does not have hover effect but default active button does?

PR #449, opened 2022-01-14
fix(Button): Use `::before` for active background to fix flicker

mayank99:
gradients cannot be animated which causes this flicker:
before

Instead of using gradient to mix two colors, I am now using an absolutely positoned `::before` psuedo element with a background-color. It fixes the flicker:
after


src/button/default.scss
```diff
@@ -38,14 +38,19 @@
...
-        linear-gradient(t(iui-color-background-1), t(iui-color-background-1));
       color: t(iui-color-foreground-primary);
     }
+
+    &::before {
+      content: '';
+      position: absolute;
+      inset: 0;
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
```
mayank99, line 51: Oh, the test is failing and I can't run it locally anymore because I uninstalled docker. 😕 I wonder if the colors are different; thought they looked the same

FlyersPh9 (approved):

src/button/default.scss
```diff
@@ -37,15 +37,20 @@
...
-        rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6))
-      ),
-        linear-gradient(t(iui-color-background-1), t(iui-color-background-1));
       color: t(iui-color-foreground-primary);
     }
+
+    &::before {
+      content: '';
+      position: absolute;
+      inset: 0;
```
bentleyvk, line 49: image
So only the blue stripe is showing that it is active which is barely visible

gretanausedaite: I can't see icons in IE. 
If we are dropping IE support, this is not a problem.
image

mayank99: > blue stripe is showing that it is active which is barely visible

I would not call it "barely visible". It is functional and that's good enough for progressive enhancement

>I can't see icons in IE. 

That's because we are using webcomponents in demo pages. The CSS itself is fine

veekeys: So what is stalling here? Is the fix regression on IE?
Could we wait for official drop time and only then start fixing things which break on IE?
What are other concerns, because this PR is staying here for a very long time. (css selector specificity increased...?)

PR #455, opened 2022-01-24
fix(Inputs): Fix toggle switch centering

FlyersPh9:
- Toggle switch border is placed using `::before` which fixes half of #406. (Similar approach to #426 )
- Toggle switch handle is placed using `::after`, removing `.iui-handle`.
- CSS is more explicit when styling checked VS unchecked.


FlyersPh9:
I adjusted the CSS of the checkbox / radio so if the width / height is changed the inner content scales accordingly.
Screen Shot 2022-01-24 at 1 58 25 PM

Our toggle switch currently does not scale properly:
Screen Shot 2022-01-24 at 1 59 38 PM

Should I include that in this PR as well?

src/toggle-switch/toggle-switch.scss
```diff
@@ -94,47 +110,58 @@
...
+      }
 
-          > .iui-handle {
-            @include themed {
-              opacity: t(iui-opacity-1);
-            }
-          }
+      &:hover:enabled ~ .iui-toggle {
+        @include themed {
+          border-color: t(iui-color-background-primary);
```
veekeys, line 130: Is this needed?

src/toggle-switch/toggle-switch.scss
```diff
@@ -94,47 +110,58 @@
...
         @include themed {
           background-color: t(iui-color-background-disabled);
           border-color: t(iui-color-background-disabled);
         }
 
-        > .iui-handle {
+        &::after {
           @include themed {
             background-color: t(iui-color-foreground-body);
             opacity: t(iui-opacity-5);
```
veekeys, line 167: Was not allowed to comment, but on line `180` cursor not-allowed for iui-label might not be needed, as its inherited.

src/toggle-switch/toggle-switch.scss
```diff
@@ -16,35 +16,32 @@
     color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
   }
 
-  > input {
+  > input[type='checkbox'] {
```
veekeys, line 19: Zoomed on 125% on chrome, this is what I get
image

90% look like that
image


But not sure anymore, if thats my eyes or its not aligned :D

src/toggle-switch/toggle-switch.scss
```diff
@@ -16,35 +16,32 @@
...
     width: 0;
     height: 0;
     appearance: none;
     opacity: 0;
     position: absolute;
 
-    // Unchecked toggle switch
+    // Toggle switch
```
veekeys, line 26: Also, if you open up 2 tabs (one with localhost your PR and one which is deployed), change the zoom and switch between those, you can see slight differences (icon placement, toggle size too).

src/toggle-switch/toggle-switch.scss
```diff
@@ -94,47 +110,58 @@
...
-          }
+      &:hover:enabled ~ .iui-toggle {
+        @include themed {
+          border-color: t(iui-color-background-primary);
         }
       }
     }
 
+    // Focused or hovered toggle switch
+    &:hover:enabled ~ .iui-toggle::after,
```
veekeys, line 136: Focus changed?
current
image


deployed
image

Also, if you press on toggle (from active) and back, you can see white color now stays when you move mouse out.

veekeys:
As you have fully changed the approach here and a bunch of css was rewritten, could you investigate if these are easily changeable for our selectors/specificity improvements?

image

src/toggle-switch/toggle-switch.scss
FlyersPh9: Seems you are correct, removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Does not appear to be needed, removed.

FlyersPh9:
Closing this PR, see #465 for the new PR.

PR #456, opened 2022-01-26
fix(Tree): Fix tree focus

saaaaaally:
Fixed focus on tree nodes, removed focus outline that showed when selecting and unselecting tree nodes.

Before:
treeFocus

After:
treeFocusAfter

### DIFF
--- a/src/tree/tree.scss
+++ b/src/tree/tree.scss
@@ -91,7 +91,7 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
   }
 
   &:focus:not(:focus-visible) {
-    outline-offset: -2px;
+    outline: none;
   }
 
   &:hover {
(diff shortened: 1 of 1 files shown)


src/tree/tree.scss
```diff
@@ -108,10 +108,13 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
   }
 
   &.iui-active {
-    @include themed {
-      background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-5));
-      outline: $iui-active-outline;
-      outline-offset: -1px;
+    &,
+    &:focus:not(:focus-visible) {
```
mayank99, line 112: This will break all active styling in Safari
https://css-tricks.com/one-invalid-pseudo-selector-equals-an-entire-ignored-selector/

And I don't think it's even correct as now there is no way to tell if the active node is focused.

saaaaaally: Not sure how to fix it in Sarai?

But you can still see a thicker outline when focusing with keyboard navigation on an active node
image

mayank99: Oh ok, I missed this part:
image

You can try using the `focus-visible` mixin I created in a recent PR:
image
image

mayank99: Actually you could also use the `iui-focus` mixin directly
image
image

saaaaaally: Ok changed to `iui-focus`

mayank99 (approved):

PR #463, opened 2022-02-02
feat(Header): Reduce nesting

### LINKED ISSUE #462: Reduce nesting of Header

FlyersPh9:
- Reduced nesting to max level 3 in Header by creating separate mixins for sections.
- Renamed classes to include a `.iui-header-` prefix:
	- `.iui-header-divider`
	- `.iui-header-chevron`
	- more listed soon…

Closes #462.

### DIFF
--- a/backstop/tests/header.html
+++ b/backstop/tests/header.html
@@ -108,7 +108,7 @@
       <section id="demo-default">
 
         <header class="iui-page-header">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -124,7 +124,7 @@
               <span class="iui-label">App name</span>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <nav aria-label="breadcrumbs">
               <!-- Project Button -->
@@ -147,7 +147,7 @@
               </button>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -175,7 +175,7 @@
             </nav>
           </div>
 
-          <div class="iui-center">
+          <div class="iui-page-header-center">
             <input
               type="search"
               placeholder="Search within Model Beta…"
@@ -183,7 +183,7 @@
             >
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -239,7 +239,7 @@
         <br />
 
         <header class="iui-page-header">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -254,7 +254,7 @@
               ></svg-imodel-hollow>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
@@ -274,7 +274,7 @@
               </button>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -297,7 +297,7 @@
             </nav>
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -359,7 +359,7 @@
       <section id="demo-slim">
 
         <header class="iui-page-header iui-slim">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -375,7 +375,7 @@
               <span class="iui-label">App name</span>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
@@ -398,7 +398,7 @@
               </button>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -426,7 +426,7 @@
             </nav>
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -485,7 +485,7 @@
         <br />
 
         <header class="iui-page-header iui-slim">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -500,7 +500,7 @@
               ></svg-imodel-hollow>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
@@ -520,7 +520,7 @@
               </button>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -543,7 +543,7 @@
             </nav>
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -620,7 +620,7 @@
       <section id="demo-split">
 
         <header class="iui-page-header iui-slim">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -636,14 +636,14 @@
               <span class="iui-label">App name</span>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
               <span class="iui-button-split-menu iui-header-split-button">
                 <div>
                   <button
-                    class="iui-button iui-borderless"
+                    class="iui-button iui-borderless iui-header-button"
                     type="button"
                     aria-label="Project picker"
                     aria-expanded="false"
@@ -670,7 +670,7 @@
               </span>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -678,7 +678,7 @@
               <span class="iui-button-split-menu iui-header-split-button iui-active">
                 <div>
                   <button
-                    class="iui-button iui-borderless"
+                    class="iui-button iui-borderless iui-header-button"
                     type="button"
                     aria-label="Model picker"
                     aria-expanded="false"
@@ -706,7 +706,7 @@
             </nav>
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -764,7 +764,7 @@
         <br />
 
         <header class="iui-page-header">
-          <div class="iui-left">
+          <div class="iui-page-header-left">
             <!-- Application Button -->
             <div
               class="iui-header-logo"
@@ -779,7 +779,7 @@
               ></svg-imodel-hollow>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
@@ -813,7 +813,7 @@
               </span>
 
               <svg-chevron-right
-                class="iui-chevron"
+                class="iui-header-chevron"
                 aria-hidden="true"
               ></svg-chevron-right>
 
@@ -849,7 +849,7 @@
             </nav>
           </div>
 
-          <div class="iui-right">
+          <div class="iui-page-header-right">
             <!-- Notifications -->
             <button
               class="iui-button iui-borderless"
@@ -920,7 +920,7 @@
       <br />
 
       <header class="iui-page-header">
-        <div class="iui-left">
+        <div class="iui-page-header-left">
           <!-- Application Button -->
           <div
             class="iui-header-logo"
@@ -936,7 +936,7 @@
             <span class="iui-label">App name</span>
           </div>
 
-          <div class="iui-divider"></div>
+          <div class="iui-header-divider"></div>
 
           <!-- Project Button -->
           <nav aria-label="breadcrumbs">
@@ -959,7 +959,7 @@
             </button>
 
             <svg-chevron-right
-              class="iui-chevron"
+              class="iui-header-chevron"
               aria-hidden="true"
             ></svg-chevron-right>
 
@@ -987,7 +987,7 @@
           </nav>
         </div>
 
-        <div class="iui-right">
+        <div class="iui-page-header-right">
           <!-- Notifications -->
           <button
             class="iui-button iui-borderless"
--- a/src/header/classes.scss
+++ b/src/header/classes.scss
@@ -6,6 +6,38 @@
   @include iui-page-header;
 }
 
+.iui-page-header-left {
+  @include iui-page-header-left;
+}
+
+.iui-page-header-center {
+  @include iui-page-header-center;
+}
+
+.iui-page-header-right {
+  @include iui-page-header-right;
+}
+
 .iui-header-logo {
   @include iui-header-logo;
+}
+
+.iui-header-button-icon {
+  @include iui-header-button-icon;
+}
+
+.iui-header-button.iui-header-button {
+  @include iui-header-button;
+}
+
+.iui-header-split-button {
+  @include iui-header-split-button;
+}
+
+.iui-header-divider {
+  @include iui-header-divider;
+}
+
+.iui-header-chevron {
+  @include iui-header-chevron;
 }
--- a/src/header/header.scss
+++ b/src/header/header.scss
@@ -7,7 +7,6 @@
 
 @mixin iui-page-header {
   @include iui-reset;
-
   display: flex;
   justify-content: space-between;
   width: 100%;
@@ -25,118 +24,93 @@
     color: t(iui-text-color);
   }
 
-  > .iui-left,
-  > .iui-center,
-  > .iui-right {
-    display: flex;
-    align-items: center;
-    height: 100%;
-  }
-
-  @include iui-header-buttons;
-
-  > .iui-left {
-    > nav {
-      display: flex;
-      align-items: center;
-      height: 100%;
-    }
-
-    &,
-    > nav {
-      overflow: hidden;
-      box-sizing: content-box;
-      margin-bottom: -$iui-xs;
-      padding-bottom: $iui-xs;
-    }
-
-    .iui-chevron {
-      @include iui-icons-small;
-      flex-shrink: 0;
-      margin: 0 $iui-xs;
-    }
-  }
-
-  > .iui-center {
-    justify-content: center;
-  }
-
-  > .iui-right {
-    justify-content: flex-end;
-
-    @media (prefers-reduced-motion: no-preference) {
-      .iui-user-icon {
-        &,
-        * {
-          transition: all $iui-speed-fast ease-out;
-        }
-      }
-    }
-  }
-
   .iui-button {
     height: 100%;
     border-radius: 0;
   }
 
-  .iui-divider {
-    width: 1px;
-    height: $iui-baseline * 2;
-    @include themed {
-      background-color: t(iui-color-background-4);
-    }
-  }
-
   &.iui-slim {
     height: round($iui-baseline * 2.5) + $iui-xxs; // Remove `+ $iui-xxs` when box-sizing: border-box is applied to all elements.
 
-    > .iui-left {
-      .iui-header-logo {
-        > .iui-label {
-          font-size: $iui-font-size;
-        }
-
-        .iui-header-button-icon {
-          margin: 0 $iui-l + $iui-xs;
-        }
+    .iui-header-logo {
+      > .iui-label {
+        font-size: $iui-font-size;
       }
 
       .iui-header-button-icon {
-        width: $iui-icons-default;
-        height: $iui-icons-default;
-        margin: 0 $iui-xs;
-        display: inline-flex;
+        margin: 0 $iui-l + $iui-xs;
       }
     }
 
-    > .iui-right {
-      .iui-user-icon {
-        @include iui-user-icon-size($size: $iui-icons-large, $font-size: $iui-font-size-small);
-      }
+    .iui-header-button-icon {
+      width: $iui-icons-default;
+      height: $iui-icons-default;
+      margin: 0 $iui-xs;
+      display: inline-flex;
+    }
+
+    .iui-user-icon {
+      @include iui-user-icon-size($size: $iui-icons-large, $font-size: $iui-font-size-small);
     }
   }
 
   @media (max-width: 768px) {
-    &,
-    &.iui-slim {
-      .iui-left {
-        .iui-button {
-          .iui-header-button-icon.iui-header-button-icon {
-            @include iui-display-none;
-          }
+    .iui-header-button {
+      .iui-header-button-icon {
+        display: none;
+      }
 
-          .iui-header-button-icon + .iui-button-label {
-            margin-left: 0;
-          }
-        }
+      .iui-header-button-icon + .iui-button-label {
+        margin-left: 0;
       }
     }
   }
 
   @media (max-width: 1024px) {
-    .iui-left {
-      .iui-header-logo > .iui-label {
-        display: none;
+    .iui-header-logo > .iui-label {
+      display: none;
+    }
+  }
+}
+
+@mixin iui-page-header-zones {
+  display: flex;
+  align-items: center;
+  height: 100%;
+}
+
+@mixin iui-page-header-left {
+  @include iui-page-header-zones;
+
+  > nav {
+    display: flex;
+    align-items: center;
+    height: 100%;
+  }
+
+  &,
+  > nav {
+    overflow: hidden;
+    box-sizing: content-box;
+    margin-bottom: -$iui-xs;
+    padding-bottom: $iui-xs;
+  }
+}
+
+@mixin iui-page-header-center {
+  @include iui-page-header-zones;
+  justify-content: center;
+}
+
+@mixin iui-page-header-right {
+  @include iui-page-header-zones;
+  justify-content: flex-end;
+
+  @media (prefers-reduced-motion: no-preference) {
+    .iui-user-icon {
+      &,
+      * {
+        transition: width $iui-speed-fast ease-out, height $iui-speed-fast ease-out, font-size $iui-speed-fast ease-out;
       }
     }
   }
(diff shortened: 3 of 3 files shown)


src/header/header.scss
```diff
@@ -25,118 +24,88 @@
...
-    &,
-    &.iui-slim {
-      .iui-left {
-        .iui-button {
-          .iui-header-button-icon.iui-header-button-icon {
-            @include iui-display-none;
-          }
+@mixin iui-page-header-center {
+  @include iui-page-header-zones;
+  justify-content: center;
```
veekeys, line 97: dont think its needed.. This would center inner content horizontally, which is not the point :?

src/header/header.scss
```diff
@@ -25,118 +24,88 @@
...
-          .iui-header-button-icon + .iui-button-label {
-            margin-left: 0;
-          }
-        }
-      }
-    }
-  }
+@mixin iui-page-header-right {
+  @include iui-page-header-zones;
+  justify-content: flex-end;
```
veekeys, line 102: Same here as with center above

src/header/header.scss
```diff
@@ -190,11 +144,11 @@
...
-        &:not(:focus-visible):not(:hover) {
-          @include themed {
-            background-color: t(iui-color-background-1);
-          }
+      &:not(:focus-visible):not(:hover) {
+        @include themed {
+          background-color: t(iui-color-background-1);
         }
       }
     }
```
veekeys, line 154: Could not comment there, but naming of `iui-header-button-icon`, which is under `iui-header-logo` seems weird..

src/header/header.scss
```diff
@@ -252,20 +204,21 @@
...
-      @media (prefers-reduced-motion: no-preference) {
-        transition: all $iui-speed-fast ease;
-      }
-      @include themed {
-        opacity: t(iui-opacity-3);
-      }
+  .iui-description {
+    height: $iui-baseline * 2;
+    @media (prefers-reduced-motion: no-preference) {
+      transition: all $iui-speed-fast ease;
```
veekeys, line 212: is `all` really needed here?

src/header/header.scss
```diff
@@ -252,20 +204,21 @@
...
+    }
+    @include themed {
+      opacity: t(iui-opacity-3);
     }
   }
 
   > .iui-button-icon {
     flex-shrink: 0;
+    box-sizing: content-box;
```
veekeys, line 221: this is getting overriden

src/header/header.scss
```diff
@@ -274,61 +227,72 @@
 
   .iui-header-button-icon {
```
veekeys, line 228: whats the difference between this and `iui-button-icon` ?

src/header/header.scss
veekeys: oh, by the same value defined below..

src/header/header.scss
```diff
@@ -274,61 +227,72 @@
...
     padding: $iui-xs;
+    box-sizing: content-box;
     @include iui-header-button-icon($iui-icons-default, $iui-border-radius);
     @include themed {
       background-color: t(iui-color-background-3);
       fill: t(iui-icons-color);
     }
   }
 
+  img.iui-header-button-icon {
```
veekeys, line 238: do we have example with img as button icon?

src/header/header.scss
```diff
@@ -274,61 +227,72 @@
...
+    padding: 0;
+    width: auto;
+    height: $iui-icons-default + ($iui-xs * 2);
+  }
+
   &.iui-active {
```
veekeys, line 244: Maybe we could have a single set of active rules?
image

src/header/header.scss
```diff
@@ -274,61 +227,72 @@
...
+        .iui-description {
+          @include iui-display-none;
         }
       }
     }
   }
 }
 
 @mixin iui-header-split-button {
+  max-width: 25vw;
```
veekeys, line 280: do you remember the story behind this value?

src/header/header.scss
```diff
@@ -337,7 +301,7 @@
     @include iui-header-button-active;
 
     // Removing button hovers and backgrounds and adding last child hover
-    .iui-button {
+    .iui-header-breadcrumbs-button {
       &.iui-borderless {
```
veekeys, line 305: dont think we use now iui-borderless anywhere.. all this might be gone now (unless now smsth is broken..)

src/header/header.scss
```diff
@@ -359,7 +323,7 @@
     }
```
veekeys, line 323: and is this last-child needed?

src/header/header.scss
```diff
@@ -359,7 +323,7 @@
...
 
     &.iui-disabled {
-      .iui-button {
+      .iui-header-breadcrumbs-button {
         &:last-child:hover {
           @include themed {
             background-color: transparent;
```
veekeys, line 329: couldnt find the place where this is working too :/

src/header/header.scss
```diff
@@ -403,7 +367,8 @@
```
veekeys, line 367: COuld you double check those rules on hover above? I think now it is working differently than before (maybe better?). Now split button is hoverable when active, before it was not.
And simple buttons have no effect on hover..

src/header/header.scss
```diff
@@ -429,3 +394,28 @@
     }
```
veekeys, line 394: these 2 classes are really confusing... `iui-button-icon` and `iui-header-button-icon`.
Seems disabled still has cursor pointer (in simple buttons, not split, which we dont have example for)

src/header/header.scss
FlyersPh9: Removed it and visuals didn't change, removed.

src/header/header.scss
FlyersPh9: Changed it to just `height` & `ease-out` (matching the rest of the header transitions).  Now the animation is a bit smoother than before.

src/header/header.scss
```diff
@@ -359,7 +323,7 @@
...
 
     &.iui-disabled {
-      .iui-button {
+      .iui-header-breadcrumbs-button {
         &:last-child:hover {
           @include themed {
```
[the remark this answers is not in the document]
FlyersPh9: Check it out again, I got it to pass! 🎉

src/header/header.scss
FlyersPh9: All mentions of `.iui-borderless` have been removed from the header CSS.

src/header/header.scss
FlyersPh9: Check again, my latest push I believe fixed this.

src/header/header.scss
FlyersPh9: Disabled cursor has been fixed.

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
-  width: $size;
-  height: $size;
-  border-radius: $border-radius;
-  @media (prefers-reduced-motion: no-preference) {
-    transition: width $iui-speed-fast ease-out, height $iui-speed-fast ease-out, margin $iui-speed-fast ease-out;
-  }
+// #region Header buttons
+@mixin iui-header-button-reset {
+  height: 100%;
+  border-radius: 0 !important;
```
veekeys, line 82: nononononono

FlyersPh9: Oh right, I meant to leave a comment on this.  This is the only instance where I was running into issues.  I'm sure we can get the `!important` removed, I'm just not sure how yet and could use some advice.

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
-
-    &,
-    > nav {
-      overflow: hidden;
-      box-sizing: content-box;
-      margin-bottom: -$iui-xs;
-      padding-bottom: $iui-xs;
-    }
+  &.iui-slim {
+    height: round($iui-baseline * 2.5) + $iui-xxs; // Remove `+ $iui-xxs` when box-sizing: border-box is applied to all elements.
```
veekeys, line 29: Mybe we could apply that bordex-box where needed and remove these 'hacks'?

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
+  @media (prefers-reduced-motion: no-preference) {
+    .iui-user-icon {
+      &,
+      * {
+        transition: width $iui-speed-fast ease-out, height $iui-speed-fast ease-out, font-size $iui-speed-fast ease-out;
       }
     }
   }
 }
+// #endregion Selection
```
veekeys, line 77: no region of this.. remove

veekeys: Or rename to the beggining of the region.. if its needed at all

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
+@mixin iui-header-button-reset {
+  height: 100%;
+  border-radius: 0 !important;
 }
 
-@mixin iui-header-buttons {
-  .iui-header-button-icon {
-    @include iui-header-button-icon;
-  }
+@mixin iui-header-brand {
```
veekeys, line 85: Sure about the renaming? I liked logo..

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
     a#{&}, // rendered as anchor element
     button#{&}, // rendered as button element
     &[role='button'] {
-      @include iui-header-button-focus;
+      @include iui-focus;
       cursor: pointer;
+      transition: background-color $iui-speed-fast ease-out;
 
       // #region resets for button
       @include iui-reset;
```
veekeys, line 129: could you double check, if all needed? color for sure not

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
-  .iui-header-split-button {
-    @include iui-header-split-button;
-  }
-}
+  &-label {
+    margin-right: $iui-sm;
+    font-size: $iui-font-size-leading;
+    transition: color $iui-speed-fast ease-out;
+    @media (prefers-reduced-motion: no-preference) {
+      transition: color $iui-speed-fast ease-out, font-size $iui-speed-fast ease-out;
```
veekeys, line 112: when color is changed?

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
-  &:focus:not(:focus-visible) {
-    outline: 0;
-  }
-}
-
-@mixin iui-header-button {
+@mixin iui-header-breadcrumbs-button {
+  @include iui-button;
+  @include iui-button-borderless;
   @include iui-button-size(large, borderless);
```
veekeys, line 159: after all these includes there are so many rules, that its hard to follow.. And there are some which are not used or overriden.. But I guess it would take too long to check one by one and mb write separately. So lets keep them, for now..

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
-      }
-      @include themed {
-        opacity: t(iui-opacity-3);
-      }
+    @media (max-width: 768px) {
+      margin-left: 0;
     }
   }
 
   > .iui-button-icon {
```
veekeys, line 193: dont all breadrcumbs buttons have `iui-header-button-icon` class?

veekeys: a ok.. so its start icon only,, and iui-button-icon is also caret.. hmm..

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
+
+    @media (max-width: 768px) {
+      display: none;
+    }
+  }
+
+  img.iui-header-button-icon {
+    padding: 0;
+    width: auto;
+    height: $iui-icons-default + ($iui-xs * 2);
```
veekeys, line 220: `iui-icons-large` ?

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
     padding: 0 $iui-xs;
   }
 
   > *:first-child {
-    .iui-button {
+    .iui-header-breadcrumbs-button {
       padding-left: $iui-sm;
+
+      &:hover {
+        background-color: transparent;
```
veekeys, line 297: Is this needed? cause now hover on projects button and imodels button is different by 0.1 opacity :DD

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
+    height: $iui-icons-default + ($iui-xs * 2);
   }
 
   &.iui-active {
     @include iui-header-button-active;
   }
 
-  &:disabled {
+  &.iui-disabled,
+  &:disabled,
```
veekeys, line 228: I think hover is broken...
Can we add in demo disabled states too? Same as with split buttons, so we coudl test easier

src/header/header.scss
```diff
@@ -429,3 +386,20 @@
...
+@mixin iui-header-divider {
+  width: 1px;
+  height: $iui-baseline * 2;
+  @include themed {
+    background-color: t(iui-color-background-4);
+  }
+}
+
+@mixin iui-header-chevron {
+  @include iui-icons-small;
```
veekeys, line 401: maybe worth including this mixin for all icons in header? it adds flex display, not sure its gonna be better or worse..

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
-      padding-bottom: $iui-xs;
-    }
+  &.iui-slim {
+    height: round($iui-baseline * 2.5) + $iui-xxs; // Remove `+ $iui-xxs` when box-sizing: border-box is applied to all elements.
 
-    .iui-chevron {
-      @include iui-icons-small;
-      flex-shrink: 0;
-      margin: 0 $iui-xs;
+    .iui-user-icon {
```
veekeys, line 31: I guess ideally we could have added class on this one too... ? Not sure if that was better, though

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
+      .iui-header-breadcrumbs-button {
+        .iui-header-button-icon {
+          width: $iui-icons-small;
+          height: $iui-icons-small;
+          padding: $iui-xxs;
+          margin: 0 $iui-xs;
+        }
+
+        img.iui-header-button-icon {
+          width: $iui-icons-small + ($iui-xxs * 2);
```
veekeys, line 248: This is inconsistent.. On normal size you set width as `auto`, so maybe it should have width defined in exact number in both cases?

src/header/header.scss
veekeys: Also this one looks weird..
image

src/header/header.scss
veekeys: Can you tell me the exact place, when its broken without it? Not sure I could find it..

src/header/header.scss
FlyersPh9: When I started this PR I actually removed that hack, but it caused test images to change so I left it as is.  I think our original plan was when we add `* {box-sizing: border-box;}` we'd remove the hacky CSS because multiple test images would be changing.

backstop/tests/header.html
```diff
@@ -620,45 +620,45 @@ <h2>
...
-              <span class="iui-label">App name</span>
+              <span class="iui-header-brand-label">App name</span>
             </div>
 
-            <div class="iui-divider"></div>
+            <div class="iui-header-divider"></div>
 
             <!-- Project Button -->
             <nav aria-label="breadcrumbs">
               <span class="iui-button-split-menu iui-header-split-button">
```
mayank99, line 643: Now it's a weird situation where we have a mix of our readymade button components and these new header button variants. Will need to work on simplifying this split button probably

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
   }
 
   > * + * {
     margin-left: $iui-sm;
   }
 
   .iui-header-button-icon {
+    display: inline-flex;
     padding: $iui-xs;
+    box-sizing: content-box;
```
mayank99, line 205: What issue does this fix and how? Feels hacky and we shouldn't rely on it.

src/header/header.scss
```diff
@@ -429,3 +386,20 @@
     }
   }
 }
+// #endregion Header buttons
+
+// #region Header extras
+@mixin iui-header-divider {
+  width: 1px;
+  height: $iui-baseline * 2;
```
mayank99, line 394: I was playing with dividers in the integration sprint and realized that `width: 1px` is not reliable on high resolution screens. Try using borders instead https://ishadeed.com/article/flexbox-separator/

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
     }
 
-    &.iui-disabled {
-      .iui-button {
-        &:last-child:hover {
-          @include themed {
-            background-color: transparent;
-          }
-        }
+    > *:last-child > .iui-header-breadcrumbs-button:hover {
```
mayank99, line 321: 0-5-0 🤔

src/header/header.scss
```diff
@@ -190,181 +136,191 @@
...
-          }
+        .iui-header-breadcrumbs-button-description {
+          @include iui-display-none;
         }
       }
     }
   }
 }
 
+@mixin iui-header-button {
```
mayank99, line 261: This is getting confusing. Why is there a `header-button` and a `header-breadcrumbs-button`?

The example in your description doesn't apply here because those extra buttons are passed by the user.

src/header/header.scss
mayank99: I kinda like that no images changed in this PR
image

src/header/header.scss
FlyersPh9: You are correct, removed from here & left on `.iui-header-button-icon`.

src/header/header.scss
FlyersPh9: `.iui-header-button-icon` is the icon that appears within our breadcrumb buttons.  `.iui-button-icon` applies to all of the icons within buttons in the header (breadcrumb button icons as well as notification, help, etc.).

src/header/header.scss
FlyersPh9: We do not currently but I could add one.  I did just now test it locally and found 2 issues, so we probably should add a test for this.

FlyersPh9: Also worth pointing out that our react header is using `` & these changes will fix the display bug in react:

**React**
Screen Shot 2022-02-18 at 4 17 13 PM

**CSS**
Screen Shot 2022-02-18 at 4 17 01 PM

src/header/header.scss
FlyersPh9: I think we needed to set a width for truncation to take place and we landed on 25% viewport width. 🤷🏻‍♂️

src/header/header.scss
FlyersPh9: We can support placing 3 types of buttons in the header with unique styling:  a breadcrumb button, a header button, and a regular button (maybe 4 if you count the branding button as well).  Our previous implementation if a user passes a regular button they would get this:
Screen Shot 2022-02-18 at 4 25 37 PM

This `.iui-header-button` modifier ensures that a user passed regular button's styling doesn't get overriden.

mayank99: We cannot apply any modifiers on user's buttons, because we don't know what element they will pass.

src/header/header.scss
FlyersPh9: In iTwinUI-Layouts Vilius is applying `* { box-sizing: border-box; }` & I know we plan on doing that across all `iui-` classes eventually.  In our layout, we are seeing this:
Screen Shot 2022-02-18 at 4 31 43 PM
 
It should look like this & `content-box` fixes it:
Screen Shot 2022-02-18 at 4 32 35 PM

src/header/header.scss
```diff
@@ -5,9 +5,9 @@
 @import '../icon/index';
 @import '../button/index';
 
+// #region Header structure
```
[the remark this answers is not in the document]
FlyersPh9: I've found them helpful for collapsing sections of code when working on long files.  If you really want them removed, let me leave them until the very end and then I can remove them.
Screen Shot 2022-02-18 at 4 35 02 PM

src/header/classes.scss
```diff
@@ -6,6 +6,42 @@
...
+
+.iui-page-header-center {
+  @include iui-page-header-center;
+}
+
+.iui-page-header-right {
+  @include iui-page-header-right;
+}
+
+.iui-header-brand {
```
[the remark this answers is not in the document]
FlyersPh9: Our current implementation has a blend of both as well.  I tried to keep on trend with that by deciding `iui-page-header` has to do with the header bar and the zones within it, & `iui-header` has to do with the content inside of `iui-page-header`.  I am ok with changing to make sure all align, but which should I use?  I'm pretty sure originally we used `iui-header` but I think it was changed along the way.

src/header/header.scss
FlyersPh9: If I remove this line we get a double hover effect on the left part of a split button:
Screen Recording 2022-02-18 at 4 43 40 PM

src/header/header.scss
FlyersPh9: Replaced the math with a singular icon variable in both instances of `img.iui-header-button-icon`.

src/header/header.scss
FlyersPh9: Really seems to only be when switching themes.  Removing color transition here and fill transition for the brand icon for the same reason (although it is visually inconsistent with the other buttons in the header).

src/header/header.scss
FlyersPh9: Renamed.

src/header/header.scss
```diff
@@ -25,159 +25,105 @@
...
   }
+}
 
-  > .iui-center {
-    justify-content: center;
-  }
-
-  > .iui-right {
-    justify-content: flex-end;
+@mixin iui-page-header-zones {
```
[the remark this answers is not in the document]
mayank99: These can be combined under `&, nav`
image

src/header/header.scss
FlyersPh9: Tried to cleanup the mixin & class names.  We now have:  `.iui-header-brand-icon`, `.iui-header-breadcrumbs-button-icon`, & `.iui-button-icon`.

src/header/header.scss
FlyersPh9: Just saw that now too, appears the right side of a disabled split button is broken.  Looking into it.

src/header/header.scss
veekeys: I removed important and it seemed to still work fine as the only rule with border radius comes from buttons mixin, which is above..

src/header/header.scss
veekeys: But isnt it how it works now??

src/header/header.scss
adamhock: I like the renaming because `iui-header-brand` includes both the `icon` and `label`. Having a class name `iui-header-logo-icon` sounds redundant in my opinion

backstop/tests/menu.html
```diff
@@ -493,7 +493,7 @@ <h3>Header</h3>
               ></svg-model>
               <span class="iui-button-label">
                 <div>Model Beta</div>
```
adamhock, line 495: Does this need to be changed to have a class name of `iui-header-breadcrumbs-button-name` like the Project Button example?

adamhock (changes requested):

backstop/tests/menu.html
```diff
@@ -629,7 +629,7 @@ <h3>Header</h3>
...
```
adamhock, line 631: Does this need to be changed to have a class name of `iui-header-breadcrumbs-button-name` like the Project Button example?

src/header/header.scss
FlyersPh9: Big rewrite, most if not all `@includes` have been removed.

src/header/header.scss
FlyersPh9: That is not how it is now.  In our current implementation if you hover over the left part of the split button you do NOT get a double hover effect on the left portion.  See this GIF and compare Project Alpha's hover states:
Screen Recording 2022-03-01 at 4 16 46 PM

src/header/header.scss
mayank99: Hover is still broken
image

I've added the disabled toggle to all sections in [`75c937e`](https://github.com/iTwin/iTwinUI/pull/463/commits/75c937e2f1b510876ec5a3fde6311ef1433fd7af).

backstop/tests/menu.html
FlyersPh9: Yeah, I think it does.  It should be updated in the latest push.

src/header/header.scss
FlyersPh9: Ok, this change has been reverted.  All `.iui-button` within the right section will be forced 100% height, but other sections are uneffected (which allows the search button to work in the center section).

src/header/header.scss
FlyersPh9: I was able to reduce specificity of this by applying a class directly.

src/header/header.scss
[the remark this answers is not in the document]
FlyersPh9: Regions have been removed & mixins have been separated into 2 different .scss files.

src/header/classes.scss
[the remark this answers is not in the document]
FlyersPh9: Everything within `header.scss` is using `.iui-page-header-` and then everything within `header-buttons.scss` is using `.iui-header-`.

src/header/header.scss
FlyersPh9: Hover is now fixed.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,346 @@
...
+    cursor: not-allowed;
+
+    svg {
+      @include themed {
+        fill: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
+      }
+    }
+  }
+
+  @at-root {
```
mayank99, line 209: The rules inside this `@at-root` don't belong in this mixin. If you look at the compiled output, you will find that none of these have the `iui-header-breadcrumb-button` class (which is the name of the mixin these are rules in currently)

src/header/header-buttons.scss
```diff
@@ -0,0 +1,346 @@
...
+  will-change: transform;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: transform $iui-speed-fast ease-out;
+  }
+
+  @at-root {
+    .iui-disabled,
+    :disabled,
+    [aria-disabled='true'] {
+      .iui-header-breadcrumb-button-icon {
```
mayank99, line 341: If you remove nesting from the place that this mixin is included, then you can just use `&` instead of this line.

Every time you type out the name of a class, you should feel some "tingling" 😄

src/header/header-buttons.scss
```diff
@@ -0,0 +1,346 @@
...
+      }
+
+      .iui-header-breadcrumb-button-dropdown:hover {
+        @include themed {
+          background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
+        }
+      }
+
+      // Blue text
+      .iui-button-label {
```
mayank99, line 228: are we using `iui-button-label` anywhere?

backstop/tests/header.html
```diff
@@ -104,86 +112,122 @@ <h1>Header</h1>
...
+                    class="iui-header-breadcrumb-button-icon"
+                    src="../assets/image-test.jpg"
+                    draggable="false"
+                  />
+                  <span class="iui-header-breadcrumb-button-label">
+                    <div class="iui-header-breadcrumb-button-label-name">Project Alpha</div>
+                    <div class="iui-header-breadcrumb-button-label-description">0x0123456789</div>
+                  </span>
+                  <svg-caret-down-small
+                    class="iui-button-icon"
```
mayank99, line 169: This one feels random. Could we avoid this `iui-button-icon` class if we are not using `iui-button` anywhere?

src/header/header.scss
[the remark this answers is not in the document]
mayank99: this is what I'm suggesting
image

backstop/tests/menu.html
adamhock: I don't see the update? I thought these cases would be changed to `<div class="iui-header-breadcrumb-button-label-name">Model Beta</div>`

FlyersPh9:
Will be submitting a new PR after #568 passes.

PR #464, opened 2022-02-03
fix: Added Popover helper class

bentleyvk:
It is meant to be used only in React as it helps to easily hide tooltips, menus and other these kind of components when they are out of view.

This should help to fix https://github.com/iTwin/iTwinUI-react/issues/529

### DIFF
--- a/src/classes.scss
+++ b/src/classes.scss
@@ -25,6 +25,7 @@
 @import './modal/classes';
 @import './non-ideal-state/classes';
 @import './notification-marker/classes';
+@import './popover/classes';
 @import './progress-indicator/classes';
 @import './side-navigation/classes';
 @import './slider/classes';
--- a/src/popover/classes.scss
+++ b/src/popover/classes.scss
@@ -1,0 +1,8 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+// tippy.js helper class to hide Popover when it is scrolled out of view
+.iui-popover[data-popper-reference-hidden] {
+  visibility: hidden;
+  pointer-events: none;
+}
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -3,7 +3,7 @@
 
 // tippy.js helper class to hide Popover when it is scrolled out of view
 @mixin iui-popover {
-  &[data-popper-reference-hidden] {
+  &[data-reference-hidden] {
     visibility: hidden;
     pointer-events: none;
   }
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -1,0 +1,10 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+// tippy.js helper class to hide Popover when it is scrolled out of view
+@mixin iui-popover {
+  &[data-popper-reference-hidden] {
+    visibility: hidden;
+    pointer-events: none;
+  }
+}
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -12,11 +12,11 @@
   // Resets tippy.js default stylings that interfere with our own
   /* stylelint-disable-next-line selector-class-pattern */
   &.tippy-box {
-    all: initial;
+    all: revert;
   }
 
   /* stylelint-disable-next-line selector-class-pattern */
   .tippy-content {
-    all: initial;
+    all: revert;
   }
 }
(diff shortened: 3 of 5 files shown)


src/popover/classes.scss
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+// tippy.js helper class to hide Popover when it is scrolled out of view
+.iui-popover[data-popper-reference-hidden] {
+  visibility: hidden;
```
bentleyvk, line 6: https://popper.js.org/docs/v2/modifiers/hide/
image

src/classes.scss
```diff
@@ -25,6 +25,7 @@
 @import './modal/classes';
 @import './non-ideal-state/classes';
 @import './notification-marker/classes';
+@import './popover/classes';
```
bentleyvk, line 28: Not sure whether it should be `popover` folder or we might need to have some utils or react folder for these kind of stuff.

src/popover/popover.scss
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+// tippy.js helper class to hide Popover when it is scrolled out of view
+@mixin iui-popover {
+  &[data-popper-reference-hidden] {
```
mayank99, line 6: I thought it was just `data-reference-hidden`

bentleyvk: Good catch! I copied CSS from Popper documentation, no idea why it has different attribute names. Fixed.

src/popover/popover.scss
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+// tippy.js helper class to hide Popover when it is scrolled out of view
+@mixin iui-popover {
```
mayank99, line 5: So where is this going to be applied? On the same level as `.tippy-content`?

We were talking about also resetting background-color and other stuff, which goes on the parent (`.tippy-box`).
image

veekeys: But we also noticed it was not the case in our storybook...? Did we figure out why its different?

src/classes.scss
veekeys: Sure we wanna add all of that in our main bundle? Maybe lets just keep this as separate css file?

src/popover/popover.scss
```diff
@@ -0,0 +1,10 @@
...
+  &[data-reference-hidden] {
+    visibility: hidden;
```
veekeys, line 7: visibility is the expected solution? not `display: none` ?

mayank99: https://github.com/iTwin/iTwinUI/pull/464#discussion_r798674492

veekeys: I need to start reading previous comments :DD

src/popover/popover.scss
bentleyvk: If users already have `tippy.js` dependency and also imported it's CSS, then it will take default styling:
image

veekeys: hmm.. If it has background only when users directly import tippy css, I guess we should not override it, imo

src/classes.scss
mayank99: That might require changing our compile script

src/popover/popover.scss
bentleyvk: It won't override their tippy.js components as we are only targeting our tippy that will have `iui-popover` class.

veekeys: OK, then add it (check to be correct DOM structure)
I just wanted to save time.. :DD

src/classes.scss
bentleyvk: Idk, if imports in react won't work (e.g. Next.js) then they may decide to import `all.css` and then it will not have it these new rules.

src/popover/popover.scss
bentleyvk: Added and tested.

src/classes.scss
veekeys: Just not include that in `classes.scss` and explicitly compile popover sass in script

bentleyvk: Isn't that too much for  a few lines?
image

veekeys: ahh...
well w/e then..
as long as you are sure you are not reverting someting which might influence the result...

veekeys (approved):

PR #465, opened 2022-02-03
fix(Inputs): Toggle switch refactor

### LINKED ISSUE #315: ToggleSwitch: no need for `padding` as it requires overrides
`ToggleSwitch` has top and bottom padding of `5.5px` and users have to override it if they want to combine `ToggleSwitch` with other components.

### LINKED ISSUE #406: Circles in radio buttons and handles of toggles are offset

FlyersPh9:
- Toggle switch body & border is using the `<input>` rather than `input ~ .iui-toggle`. Significantly cleans up the CSS and fewer HTML elements.
- Toggle switch handle is placed using `::after`, removing `.iui-handle`.
- CSS is more explicit when styling checked VS unchecked.
- Class name changes:
	- `.iui-toggle-switch` is now `.iui-toggle-switch-wrapper` aligning with naming convention used in #475.
	- `.iui-toggle` is now `.iui-toggle-switch` aligning with naming convention used in #475.
	- `.iui-icon` is now `.iui-toggle-switch-icon` & specificity was reduced by using `&-icon`.
	- `.iui-label` is now `.iui-toggle-switch-label`.
	- Removed `.iui-handle`, using `::after` instead.
- Removed padding surrounding toggle switch.  Spacing is handled by `.iui-input-group`.  If toggle is by itself, spacing must be added by the user. Closes #315.
- Using CSS variables for parts that do not display as a toggle switch in IE.
- Height changed to 24px from 22px, confirmed to display correctly when OS scaling is set to 125%. Closes #406.
- Displays a checkbox in IE.
- Better tests & test images.


src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
 
-  > input {
-    width: 0;
-    height: 0;
+  > input[type='checkbox'] {
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: inline-block;
+    flex-shrink: 0;
```
mayank99, line 29: This should not be needed

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
+    position: relative;
     appearance: none;
```
mayank99, line 31: This wont work in IE.
image

Maybe that's fine though, because it is functional at least.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
-    opacity: 0;
-    position: absolute;
+    width: $toggle-switch-height * 2 - $iui-xs;
```
mayank99, line 32: Why is width using this `-height` variable?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
+      border-color: inherit;
+      box-sizing: border-box;
+    }
+
+    &::after {
+      content: '';
+      position: absolute;
+      height: $toggle-switch-height - ($iui-xxs * 3);
+      width: $toggle-switch-height - ($iui-xxs * 3);
+      aspect-ratio: 1 / 1;
```
mayank99, line 58: aspect-ratio is not supported in all browsers. What happens in older browsers?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
+      top: 50%;
+      transform: translateY(-50%);
```
mayank99, line 60: I think this center could be improved by just using flex.

```scss
input {
  display: flex;
  ...
  &::after {
    align-self: center;
  }
}
```

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
+      border-radius: 50%;
       @media (prefers-reduced-motion: no-preference) {
```
mayank99, line 62: background and opacity transitions dont need to be inside this media query

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
-        transition: background-color $iui-speed-fast ease, border-color $iui-speed-fast ease;
+        transition: left $iui-speed-fast ease, background-color $iui-speed-fast ease, opacity $iui-speed-fast ease;
```
mayank99, line 63: `left` should not be transitioned as it has poor performance. Can you try using `transform`?

https://web.dev/animations-guide/

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
           background-color: t(iui-color-foreground-body);
           opacity: t(iui-opacity-2);
         }
       }
-    }
 
-    &:hover:not(:disabled),
-    &:focus {
-      ~ .iui-toggle {
+      &:hover:enabled,
```
mayank99, line 82: Can we avoid `:enabled` by moving `:disabled` state styling to the end?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,206 @@
...
-          border: 1px solid rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
+          border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
         }
+      }
 
-        > .iui-handle {
-          @include themed {
-            opacity: t(iui-opacity-1);
-          }
+      &:focus:not(:focus-visible) {
```
mayank99, line 89: Can you use our `focus-visible` mixin here?

src/toggle-switch/toggle-switch.scss
mayank99: I see it's not even centered currently (on my 1440p screen)
image

src/toggle-switch/toggle-switch.scss
bentleyvk: You know my position about IE support 😀

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: I made some changes so IE should display a normal checkbox but all other modern browsers should get the full toggle.  Can someone with IE confirm my latest push?

src/toggle-switch/toggle-switch.scss
FlyersPh9: I did some rethinking here.  Below is an image explaining what I'm trying to achieve spacing wise.
Screen Shot 2022-02-04 at 1 19 26 PM

Before:  I was basing all the sizing off of the height of the toggle body.
Now:  I am basing all the sizing off of the toggle handle.

Everything scales properly based off the size set for the handle.  Here is what it looks like if the toggle handle is set to 24px instead of 16px.
Screen Shot 2022-02-04 at 1 22 10 PM

src/toggle-switch/toggle-switch.scss
FlyersPh9: It _should_ just be skipped.  The set width & height should equal 1:1 but when zooming in/out the toggle handle would sometimes not appear 1:1 & this seemed to fix it on my side.

src/toggle-switch/toggle-switch.scss
FlyersPh9: I made some changes, can you check again?

src/toggle-switch/toggle-switch.scss
FlyersPh9: `@media (prefers-reduced-motion: no-preference)` is only targeting motion now.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Using `transform: translateX` now.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed `:enabled` but needed to add `&, &:hover` within `&:disabled` (line 143).

src/toggle-switch/toggle-switch.scss
mayank99: Checkbox is good, grid needs to be fixed
image

I was hoping autoprefixer would handle that automatically

src/toggle-switch/toggle-switch.scss
mayank99: Still looks the same
image

FlyersPh9: 🤔 I'm at a loss here.

src/toggle-switch/toggle-switch.scss
FlyersPh9: I am using the `focus-visible` mixin on lines 126 & 162.

mayank99: You don't need to use the themed mixin with focus-visible since it is a modern feature. You can use CSS variables directly

src/toggle-switch/toggle-switch.scss
bentleyvk: Wtf guys? How it is ok? It is regression. We haven't deprecated IE yet...

src/toggle-switch/toggle-switch.scss
FlyersPh9: Now using CSS variables within the `focus-visible` mixin.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Can you confirm if this is happening to all of the toggles on the demo page?  Or only the toggles within `.iui-input-group`?

mayank99: Looks like it's on both
image
image

My guess is that the wrapping `<label>` is causing the problem because it has a 22px height

src/toggle-switch/toggle-switch.scss
FlyersPh9: I think I have an idea of how to fix this.  I'm leaving this comment here for myself on Monday so I can try it then:
I'm going to use `::before` for the entire body of the toggle (the background color & border) instead of just the border.  I will use z-index to place them above the unstyled checkbox that appears in IE.

mayank99: It's ok to not isolate the z-index in IE as long as it works. Isolation is for preventing side effects when the component interacts with other elements beside it. If you only use positive z-indexes it should be fine

adamhock: I put the following code after the `input` selector, and it seemed to work for me

```scss
&::-ms-check {
    display:none;
 }
```

image

src/toggle-switch/toggle-switch.scss
FlyersPh9: Pretty sure we discovered the cause of the off centering: `$iui-baseline` equaling 11px in combination with OS scaling.  Our toggle switch is `$iui-baseline * 2` in height, meaning 22px.  If a user's OS is set to scale at 125% then 22px * 1.25 scale = 27.5px tall.  It's that half pixel that is causing issues.  I am investigating changing our baseline from 11px to 12px but this will effect nearly all of our UI components.

src/skip-to-content/skip-to-content.scss
```diff
@@ -19,7 +19,8 @@
   transform: translateX(-50%) translateY(-170%);
   transition: background-color $iui-speed-fast ease-in-out;
   @media (prefers-reduced-motion: no-preference) {
-    transition: background-color $iui-speed-fast ease-in-out, transform $iui-speed-slow ease-in-out, box-shadow $iui-speed-fast ease-in-out;
+    transition: background-color $iui-speed-fast ease-in-out, transform $iui-speed-slow ease-in-out,
```
bentleyvk, line 22: Why it was changed?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 
+$toggle-switch-handle: $iui-icons-default;
```
bentleyvk, line 5: Can you make it more specific? Because it is completely unclear to me by reading its name for what this variable is used.

backstop/tests/header.html
```diff
@@ -609,10 +609,9 @@ <h2>
           class="iui-toggle-switch"
           style="display: inline-flex; vertical-align: middle; margin-left: 1em;"
         >
```
bentleyvk, line 611: Why header test images have changed?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+
 @mixin iui-toggle-switch {
   @include iui-reset;
   display: flex;
   align-items: center;
+  gap: $iui-s;
+  align-items: center;
   font-size: $iui-font-size;
   width: fit-content;
-  padding: ($iui-baseline/2) 0;
```
bentleyvk, line 11: > Removed padding surrounding toggle switch (https://github.com/iTwin/iTwinUI/issues/315). Spacing is handled by .iui-input-group. If toggle is by itself, spacing must be added by the user.

I am not sure it is ok to do this as I bet more users are using toggle switches alone and that's why I decided to leave it for major release:
image

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+    margin-left: $iui-s;
+    margin-right: 0;
+    @supports (gap: $iui-s) {
+      margin-left: 0;
+    }
+  }
+}
+
+@mixin iui-toggle {
+  display: grid;
```
bentleyvk, line 51: What is the point of using grid here if it is only one element?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+  grid-template: 'toggle';
+  align-items: center;
+
+  > input[type='checkbox'] {
```
bentleyvk, line 55: Do we really need to have attribute in selector because it is making specificity higher?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: flex;
+    align-items: center;
+    position: relative;
+    cursor: pointer;
```
bentleyvk, line 62: Is it needed as you have this rule on the top element?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+  display: grid;
+  grid-template: 'toggle';
+  align-items: center;
+
+  > input[type='checkbox'] {
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: flex;
+    align-items: center;
```
bentleyvk, line 60: Seems to do nothing.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
   display: flex;
   align-items: center;
+  gap: $iui-s;
+  align-items: center;
   font-size: $iui-font-size;
   width: fit-content;
-  padding: ($iui-baseline/2) 0;
   user-select: none;
   cursor: pointer;
   position: relative;
```
bentleyvk, line 17: Seems you can remove it.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+  > input[type='checkbox'] {
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: flex;
+    align-items: center;
+    position: relative;
+    cursor: pointer;
+    width: ($toggle-switch-handle + $iui-xs) * 2;
+    height: auto;
```
bentleyvk, line 64: Doesn't do anything if removed.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+    padding: 1px;
+    border-radius: 9999px;
     appearance: none;
-    opacity: 0;
-    position: absolute;
+    z-index: 1;
+
+    &::before {
```
bentleyvk, line 70: What is the benefit having toggle switch body inside `:before`? You wrote it claened up CSS, but not it has even more lines of code 😀

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: flex;
+    align-items: center;
+    position: relative;
+    cursor: pointer;
+    width: ($toggle-switch-handle + $iui-xs) * 2;
+    height: auto;
+    padding: 1px;
+    border-radius: 9999px;
```
bentleyvk, line 66: I think we might have a variable for this when Leah added skip to content.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
     appearance: none;
-    opacity: 0;
-    position: absolute;
+    z-index: 1;
```
bentleyvk, line 68: Can you check whether all z-indexes are needed?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
-        outline: 1px solid t(iui-color-foreground-primary);
-      }
-      outline-offset: 1px;
+  > .iui-toggle-switch-icon {
+    grid-area: toggle;
+    opacity: 0;
+    width: $toggle-switch-handle * 0.75;
+    height: $toggle-switch-handle * 0.75;
+    display: flex;
+    // padding: 0 ($iui-xs + 1px);
```
bentleyvk, line 202: Remove commented code.

src/skip-to-content/skip-to-content.scss
FlyersPh9: It looks like a new line was added after the transform transition.  I didn't do this manually, maybe it was the linter?

backstop/tests/header.html
FlyersPh9: The header test images changed because the toggle switch is used to enable / disable slim mode.  This PR removes the padding on top / bottom of the toggle to close #315, which caused the bottom header demo to shift upwards a little bit.

src/toggle-switch/toggle-switch.scss
FlyersPh9: The only way I can think to resolve this is to add a modifier class similar to our typography components' `.iui-text-spacing`.

src/toggle-switch/toggle-switch.scss
FlyersPh9: This was my attempt at getting toggles to display above the checkbox in Internet Explorer ([this discussion](https://github.com/iTwin/iTwinUI/pull/465#discussion_r798811687)).  My attempt didn't work, so I'm thinking if this PR does indeed fix the centering issue, we may not have a viable option for IE aside from the standard checkbox.

src/toggle-switch/toggle-switch.scss
FlyersPh9: This one does appear to be needed, if I remove it I get `cursor: default;`.

src/skip-to-content/skip-to-content.scss
FlyersPh9: If I revert this, I get:

Screen Shot 2022-02-22 at 1 45 40 PM

> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -
husky > pre-commit (node v12.22.7)
⚠ Detected incorrect braces with only single value: `*.{scss}`. Reformatted as: `*.scss`
[STARTED] Preparing...
[SUCCESS] Preparing...
[STARTED] Running tasks...
[STARTED] Running tasks for *.scss
[STARTED] Running tasks for *.{scss,html}
[STARTED] Running tasks for *.{scss,html,js}
[STARTED] prettier --write
[STARTED] stylelint --fix
[STARTED] yarn lint:copyright --fix
[SUCCESS] prettier --write
[SUCCESS] Running tasks for *.scss
[SUCCESS] yarn lint:copyright --fix
[SUCCESS] Running tasks for *.{scss,html,js}
[SUCCESS] stylelint --fix
[SUCCESS] Running tasks for *.{scss,html}
[SUCCESS] Running tasks...
[STARTED] Applying modifications...
[FAILED] Prevented an empty git commit!
[STARTED] Reverting to original state because of errors...
[SUCCESS] Reverting to original state because of errors...
[STARTED] Cleaning up...
[SUCCESS] Cleaning up...
  ⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration
husky > pre-commit hook failed (add --no-verify to bypass)

src/toggle-switch/toggle-switch.scss
FlyersPh9: Changed to `$toggle-switch-handle-size`.

src/toggle-switch/toggle-switch.scss
FlyersPh9: I have removed it because it increases specificity.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: You are correct, it is now being used.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Size has been changed to 24px from 22px height, @mayank99 can you confirm if this is displaying centered when OS scaling is set to 125%?

src/toggle-switch/toggle-switch.scss
FlyersPh9: Using `grid` we can avoid using `position: absolute;` on `.iui-toggle-switch-icon`.

src/toggle-switch/toggle-switch.scss
FlyersPh9: We did this because we thought the 1px border was the cause of the off-centering issue in checkbox / radio buttons.  It seems like that may not be the case, so my latest push puts all of the body styling (background color & border) on the `<input>` instead of in `::before`.

src/toggle-switch/toggle-switch.scss
FlyersPh9: [Reverted back to what is shown in this post](https://github.com/iTwin/iTwinUI/pull/465#discussion_r799715004).  I tried getting this to display correctly in IE a couple of ways and it doesn't seem feasible without going back to our old method of `input ~ .iui-toggle`.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Worth pointing out that this is how our checkboxes / radios work without `.iui-labeled-input` wrapping them.  Our demos don't have them touching because we have a 11px `gap` on the `<section>`s.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,212 @@
...
-    // Focus on non-disabled toggle switch
-    &:enabled:focus ~ .iui-toggle {
-      @include themed {
-        outline: 1px solid t(iui-color-foreground-primary);
-      }
-      outline-offset: 1px;
+  > .iui-toggle-switch-icon {
+    grid-area: toggle;
+    opacity: 0;
+    width: $toggle-switch-handle * 0.75;
```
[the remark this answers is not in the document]
FlyersPh9: I've swapped out the sizing SASS variables with CSS variables, so now you can play around with the sizing in dev tools.  I kind of like that the entire toggle scales proportionally to a change, but if the user puts in a bad value I could see it causing issues.

src/skip-to-content/skip-to-content.scss
bentleyvk: Don't revert. I was just curious why linter/prettier changed it, I thought we fixed this.

src/toggle-switch/toggle-switch.scss
bentleyvk: For me it feels like a bigger hack than `position: absolute` 😀 Because you are saying in template that there will be one item - toggle -  but you are also using the same area name for icon.

src/toggle-switch/toggle-switch.scss
bentleyvk: Ok, then remove it from `iui-toggle-switch`.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,209 @@
...
+    margin-left: $iui-s;
+    margin-right: 0;
+    @supports (gap: $iui-s) {
+      margin-left: 0;
+    }
+  }
+}
+
+@mixin iui-toggle-switch {
+  --_iui-toggle-switch-handle-size: #{$iui-icons-default};
```
bentleyvk, line 48: What's the point having these as CSS variables instead of SASS variables? We are not overrding them with any of modifier. Also, now you have to have fallbacks for IE.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,209 @@
...
-    opacity: 0;
-    position: absolute;
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
+    grid-area: toggle;
+    display: flex;
+    position: relative;
+    cursor: pointer;
+    -webkit-tap-highlight-color: transparent;
+    touch-action: pan-y;
```
bentleyvk, line 64: What it does?

src/toggle-switch/toggle-switch.scss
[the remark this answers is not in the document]
bentleyvk: I left comment about CSS variables.

bentleyvk:
Tests are failing

bentleyvk:
So instead of all these changes, will it be enough just to change old toggle switch height to `24px`?

src/toggle-switch/toggle-switch.scss
mayank99: This tells the browser that it should only handle the y-axis for drag events. It's useful when you have custom drag logic for touchscreen devices (e.g. our Slider component) but it's not helping in this component

src/toggle-switch/toggle-switch.scss
mayank99: There is no need for named areas. In `input-with-icon` we're using the same idea with `grid-area: 1 / -1` for overlapping areas.

The benefit is that you get to leverage the centering from grid.

src/toggle-switch/toggle-switch.scss
FlyersPh9: That one is needed too it seems.  Without it we get `cursor: default` on the label.

src/toggle-switch/toggle-switch.scss
FlyersPh9: The idea was taken from this article when I was searching for a centering solution. https://web.dev/building-a-switch-component/#touch-action

FlyersPh9:
> So instead of all these changes, will it be enough just to change old toggle switch height to `24px`?

The specificity will go back up again with all the `input ~ .iui-toggle` rules.

FlyersPh9:
> Tests are failing

My mistake, yesterday I had a catastrophic merge failure and had to roll things back.  Looks like my changes in fieldset.html were lost.  I've updated it again.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removing it because I just want to get this PR approved, can add it later if wanted.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Lines 66-67 are still using SASS variables, so I believe it should still display ok in IE.  The CSS variables allowed me to test the scaling of the components, but I can revert them back to SASS.  I thought we were starting to switch over to CSS variables tho?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,208 @@
...
 
+  // Disabled
+  &.iui-disabled {
+    cursor: not-allowed;
+    @include themed {
+      color: t(iui-text-color-muted);
+    }
+  }
+
+  // #region IE solution
```
bentleyvk, line 28: If it is not working on IE, maybe we can just remove these workaround?

src/toggle-switch/toggle-switch.scss
bentleyvk: I don't really see a point to have them as CSS variables as we are not changing them dynamically. And I think we were talking about switching to CSS variables when we will end IE support and we will have separate lib for varaibles 😀

FlyersPh9: Ok, I have removed the CSS variables.

src/toggle-switch/toggle-switch.scss
FlyersPh9: It's not displaying as a toggle switch in IE but it does still work (as a checkbox).  This CSS here makes it look ok in IE as a checkbox.

src/toggle-switch/toggle-switch.scss
bentleyvk: I started to think maybe it is ok. This is kind of fixing a bug 😀

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,207 @@
...
-      display: block;
-      position: relative;
-      height: $iui-baseline * 2 - $iui-xxs;
-      width: ($iui-m * 2) + ($iui-xxs * 3);
-      border-radius: $iui-baseline;
-      @media (prefers-reduced-motion: no-preference) {
-        transition: background-color $iui-speed-fast ease, border-color $iui-speed-fast ease;
-      }
+    @include iui-reset;
+    @include iui-focus($thickness: 1px, $offset: 1px);
```
bentleyvk, line 59: Maybe reuse thickness variable here.

src/toggle-switch/toggle-switch.scss
bentleyvk: > There is no need for named areas. In `input-with-icon` we're using the same idea with `grid-area: 1 / -1` for overlapping areas.

@FlyersPh9 Any actions for this?

src/inputs/labeled-inputs.scss
```diff
@@ -55,6 +55,10 @@
     // stylelint-enable
   }
 
+  .iui-input-group .iui-toggle-switch-wrapper {
+    padding: round($iui-baseline * 0.5) 0;
```
mayank99, line 59: What is this for?

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,205 @@
...
+  height: $iui-toggle-switch-handle-size;
+  @supports (appearance: none) {
+    appearance: none;
+    width: ($iui-toggle-switch-handle-size + ($iui-toggle-switch-handle-offset * 2)) * 2;
+    height: auto;
+    border-radius: $iui-border-radius-round;
+    box-sizing: border-box;
+    transition: background-color $iui-speed-fast ease-out, border-color $iui-speed-fast ease-out;
+    @include themed {
+      background-color: t(iui-color-background-1);
```
mayank99, line 79: Because there is no IE support, I think pretty much everything in this component could reference the CSS `var`s directly without the `themed` mixin or `t`.

backstop/tests/toggle-switch.html
```diff
@@ -150,142 +155,146 @@ <h2>Default</h2>
...
+          type="checkbox"
+          role="switch"
+          disabled
+        />
+        <svg-placeholder
+          class="iui-toggle-switch-icon"
+          aria-hidden="true"
+        ></svg-placeholder>
       </div>
+      </label>
```
mayank99, line 297: This might be leftover

src/toggle-switch/toggle-switch.scss
FlyersPh9: Yes, I collaborated with Mayank yesterday and we got this part working.

src/inputs/labeled-inputs.scss
FlyersPh9: #315 wants the padding removed from the toggle switch, so I placed this within `labeled-inputs.scss` to avoid causing issues.

src/toggle-switch/toggle-switch.scss
FlyersPh9: Removed `@include themed {` from all places _except_ `color` applied to `<label>` because that part does still work in IE.

backstop/tests/toggle-switch.html
FlyersPh9: You are correct, removed from the next push.

backstop/tests/toggle-switch.html
```diff
@@ -45,42 +52,40 @@ <h2>Default</h2>
...
-            <span class="iui-toggle">
-              <span class="iui-handle"></span>
-            </span>
+          <label class="iui-toggle-switch-wrapper iui-label-on-right">
+            <input
+              class="iui-toggle-switch"
+              type="checkbox"
+              role="switch"
+            />
             <span class="iui-label">Option 2</span>
```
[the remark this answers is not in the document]
FlyersPh9: Changed to `.iui-toggle-switch-label`.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,173 @@
...
+  // #region IE solution
+  width: $iui-toggle-switch-handle-size;
+  height: $iui-toggle-switch-handle-size;
+  @supports (appearance: none) {
+    appearance: none;
+    width: ($iui-toggle-switch-handle-size + ($iui-toggle-switch-handle-offset * 2)) * 2;
+    height: auto;
+    border-radius: $iui-border-radius-round;
+    box-sizing: border-box;
+    transition: background-color $iui-speed-fast ease-out, border-color $iui-speed-fast ease-out;
```
elephantcatdog, line 67: Some transitions in this file aren't surrounded by a `@media( prefers-reduced-motion...)`. I tested them out and they're very subtle. I noticed that in one section there's a subtle version and a more overt version. So I'm assuming this was done on purpose and it's okay? Generally there's no transitions except those in the `@media` block, so just making sure.

elephantcatdog (commented):
Mostly looks good to me, just want to confirm the transitions are okay.

src/toggle-switch/toggle-switch.scss
```diff
@@ -2,194 +2,173 @@
...
+  // Disabled
+  &.iui-disabled {
+    cursor: not-allowed;
+    @include themed {
+      color: t(iui-text-color-muted);
     }
+  }
+
+  // For labels on the left
+  &.iui-label-on-left {
```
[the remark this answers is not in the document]
FlyersPh9: We weren't able to find a way to have the icon placed within the toggle switch without a wrapper element, so this was our work around.  @mayank99, did I relay this correctly?

backstop/tests/header.html
```diff
@@ -910,12 +911,13 @@ <h2>
     <h2>Animation (Toggle default ⇔ slim)</h2>
```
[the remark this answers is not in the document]
FlyersPh9: Can you share a video of this?  I'm not seeing any content shift on my side.

src/toggle-switch/toggle-switch.scss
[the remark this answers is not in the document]
mayank99: Yes, this is the tradeoff for having only one wrapper element. It uses the same grid for everything now, so the classes are necessary for changing the grid-template

src/toggle-switch/toggle-switch.scss
mayank99: The `reduced-motion` guard only applies to transitions that consist of some motion. When there is only a color change, it is ok to keep those transitions outside the media query.

bentleyvk (approved):

PR #467, opened 2022-02-04
feat(ExpandableBlock): Add borderless styling to expandable-block

### LINKED ISSUE #454: Create borderless expandable block styling.

adamhock:
Multiple expandable blocks can appear visually heavy when placed next to each other. An iui-borderless modifier class has been added to iui-expandable-block which removes the border and background-color of the header and content and removes the padding on the content. The hover effect on the header of the expandable block still remains the same

Closes #454


src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
       margin-left: $iui-s;
     }
   }
+
+  &.iui-borderless {
+    @include themed {
```
adamhock, line 172: I think I made a mistake here. If the background color and border are removed in iui-header and iui-expandable-content, this may be redundant

src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
...
+
+      &:hover {
+        @include themed {
+          background-color: t(iui-color-background-4);
+        }
+      }
+    }
+
+    .iui-expandable-content {
+      @include themed {
```
adamhock, line 191: I could write ```.iui-header, .iui-expandable-content``` to combine the two themes

src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
...
+    @include themed {
+      background-color: transparent;
+      border: initial;
+    }
+
+    .iui-header {
+      @include themed {
+        background-color: transparent;
+        border: initial;
+      }
```
mayank99, line 181: you don't need to use the `themed` mixin if you're not using any of our variables.

```suggestion
      background-color: transparent;
      border: initial;
```

Same thing in other places that you are including this mixin

src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
...
+
+      &:hover {
+        @include themed {
+          background-color: t(iui-color-background-4);
```
mayank99, line 185: I think this hover color needs to be a bit lighter. You can inspect our borderless buttons as an example

src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
...
+    }
+
+    .iui-expandable-content {
+      @include themed {
+        background-color: transparent;
+        border: initial;
+      }
+
+      > div {
+        padding: 0;
```
mayank99, line 197: Are you sure about removing this padding?

src/expandable-block/block.scss
```diff
@@ -167,4 +167,35 @@
...
+      @include themed {
+        background-color: transparent;
+        border: initial;
+      }
+
+      &:hover {
+        @include themed {
+          background-color: t(iui-color-background-4);
+        }
+      }
```
mayank99, line 187: I think you will need to add a border-radius here to soften those sharp edges
```suggestion
        border-radius: $iui-border-radius;
      }
```

src/expandable-block/block.scss
adamhock: Isssue #454 asks that the padding be removed in the content area. I assumed it was because the padding was used to give space between the content and its border. Now that the border is removed, I don't see a need for the padding

src/expandable-block/block.scss
adamhock: The screenshot posted in issue #454 showed this background color, but I'll take a look

mayank99: I think the screenshot in the issue matches my suggestion. It even explicitly says _"We would likely want to copy the hover effect that is present in our borderless buttons"_

They look similar in light theme but quite different in dark theme.
image

```suggestion
          background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-6));
```

FlyersPh9: In light theme the hovered borderless expandable block closely matches the hovered borderless button, but in dark theme it's more obvious that they're not matching:
Screen Shot 2022-02-04 at 3 47 13 PM Screen Shot 2022-02-04 at 3 47 31 PM

[Try using this background-color](https://github.com/iTwin/iTwinUI/blob/0f789cae04237e76672d28b6af665fdb961b362d/src/button/borderless.scss#L20).

adamhock: I misread matching the hover effect with the borderless buttons instead of matching with the non-borderless expandable block headers. This makes sense now

src/expandable-block/block.scss
adamhock: Property added

src/expandable-block/block.scss
adamhock: removed

src/expandable-block/block.scss
adamhock: Background color for hover has changed

src/expandable-block/block.scss
```diff
@@ -167,4 +167,23 @@
       margin-left: $iui-s;
     }
   }
+
+  &.iui-borderless {
+    .iui-header,
+    .iui-expandable-content {
+      background-color: transparent;
```
adamhock, line 174: This results in the background color being `iui-color-background-3` from `iui-expandable-block` instead of `iui-color-background-2` from `iui-body`.

We can either change the background color here or make `iui-expandable-block-borderless` its own class and set its background color to `transparent` similar to how `iui-button-borderless` is written

mayank99: Changing it here is quite simple, so you could just do that. All you'd need to add is `&`.
image

If you were to create a `iui-expandable-block-borderless`, would it be _in addition to_ `iui-expandable-block` or _instead_ of it? The former wouldn't change much as you would still have to do some overrides. And the latter would require some extra work to make sure all styles are included.

In the case of buttons, the reason `iui-button` doesn't have anything is just because we have the `iui-default` modifier for applying default colors. So `iui-borderless` and `iui-default` are never used together.

adamhock: Your suggested change makes the most sense

Change has been implemented

mayank99: Dont forget to update test images

adamhock: The test images have been updated

mayank99: Yes, it will pass because the diff checker has issues with recognizing color changes. But if you approved those images before changing this background color, then the images are "wrong" even if the tests pass.

To force an update, you can manually delete the images, then run the tests (causing failures), and approve the new images.

src/expandable-block/block.scss
```diff
@@ -167,4 +167,24 @@
...
+      border: initial;
+    }
+
+    .iui-header:hover {
+      @include themed {
+        background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-6));
+      }
+      border-radius: $iui-border-radius;
```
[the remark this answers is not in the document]
adamhock: This has been changed

FlyersPh9 (approved):

PR #471, opened 2022-02-07
feat(Skip-to-content): Add skip-to-content button

### LINKED ISSUE #450: Skip to Content link
#### Feature
'Skip to content' link for users using tab navigation. Allows to bypass links in header/menus and land directly into a page's content. This is a basic accessibility feature we should highly consider.

#### Examples
[WebAIM recommendation](https://webaim.org/techniques/skipnav/)
[More info from Deque University](https://dequeuniversity.com/tips/add-skip-navigation-link)

CC @FlyersPh9

elephantcatdog:
Added component "Skip to content".

Component:
 - is hidden above top of screen until it is tabbed to
 - has a border and shadow on focus
 - darkens on hover
 - can have multiple lines of text

An aspect of using "Skip to content" links that are shown on tab is that it should be clearly visible to sighted keyboard users and should not be able to be overlooked if they are quickly tabbing. I used a fast transition speed when transforming the link downward so that it should be mostly visible even if tabbing is fast. An alternative, and what it looks like WebAIM is doing, is to not have a transition from off screen to on screen, and then have a slower transition back to off the screen so that the component will be immediately visible when tabbed to and have more time on-screen. I think this alternative might be better, if it's possible. Thoughts?

Current Implementation (Fast In, Fast Out) (regular tabbing speed then quick tabbing speed):

https://user-images.githubusercontent.com/37936169/152854548-67c4c199-6bf5-4226-a20d-29ff2f6e8b5f.mp4

Fast In, Slow Out Implementation (regular tabbing speed then quick tabbing speed):

https://user-images.githubusercontent.com/37936169/152870290-9cc1344e-a924-4689-8e53-8b6f6ee12702.mp4

Instant In, Slow Out Implementation (regular tabbing speed then quick tabbing speed):

https://user-images.githubusercontent.com/37936169/152870705-b9b73b59-b69e-4635-b9cd-381a9c6afa15.mp4

closes #450


src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+
+@mixin iui-skip-to-content {
+  @include iui-blur;
+  @include iui-focus($offset: 2px, $thickness: 2px);
+
+  border-radius: 99999px;
```
mayank99, line 10: I think we might want to save this border-radius as a global variable somewhere, we need it quite often

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
...
+  color: t(iui-color-foreground-accessory);
+  padding: ($iui-baseline * 0.75) $iui-m;
+  text-decoration: none;
+  position: absolute;
+  text-align: center;
+  left: 50%;
+  transform: translateX(-50%) translateY(-170%);
```
mayank99, line 17: Can you explain this magic number (170)? Or did you just pick an arbitrary off-screen value?

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
...
+  transition: background-color $iui-speed-fast ease-in-out,
+              transform $iui-speed-fast ease-in-out,
+              box-shadow $iui-speed-fast ease-in-out;
+  
+  &:hover {
+    @include iui-blur($opacity: 3);
+  };
```
mayank99, line 24: nit: You dont need a semicolon here

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
...
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+
+@mixin iui-skip-to-content {
+  @include iui-blur;
+  @include iui-focus($offset: 2px, $thickness: 2px);
+
+  border-radius: 99999px;
+  color: t(iui-color-foreground-accessory);
```
mayank99, line 11: I don't see any background-color, I think we might want to add one because the contrast ratio is inaccessible in light theme

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
...
+  &:hover {
+    @include iui-blur($opacity: 3);
+  };
+  
+  &:focus {
+    transform: translateX(-50%) translateY(0);
+    box-shadow: $iui-elevation-16;
+  }
+
+  // visible for x amount of time after tabbed away from
```
mayank99, line 31: leftover?

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+
```
mayank99, line 5: nit: leftover newline

backstop/tests/skip-to-content.html
```diff
@@ -0,0 +1,53 @@
...
+    />
+    <link
+      rel="stylesheet"
+      href="../assets/demo.css"
+    />
+    <script src="../assets/theme.js"></script>
+  </head>
+  <body
+    class="iui-body"
+    style="position: relative"
```
mayank99, line 29: is this position: relative needed? I think body automatically gets it

backstop/tests/skip-to-content.html
```diff
@@ -0,0 +1,53 @@
...
+      class="iui-skip-to-content"
+      id="test-1"
+    >Skip to main content</a>
+    <theme-button></theme-button>
+    <h1>Skip to content</h1>
+    <hr />
+
+    <section id="other-content">
+      <div style="height: 1000px; border: solid 1px;">
+        <p>This is not the main content.</p>
```
mayank99, line 42: Maybe add some interactable elements (buttons, links, etc) so that it's obvious the skip link even works

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,32 @@
...
+  transform: translateX(-50%) translateY(-170%);
+  transition: background-color $iui-speed-fast ease-in-out,
+              transform $iui-speed-fast ease-in-out,
+              box-shadow $iui-speed-fast ease-in-out;
+  
+  &:hover {
+    @include iui-blur($opacity: 3);
+  };
+  
+  &:focus {
```
mayank99, line 26: If you want to show a slower exit animation, you can change the transition-duration inside this focus rule

src/skip-to-content/skip-to-content.scss
elephantcatdog: This is the nearest round number that just barely hides the component. The closer the component is to the edge of the screen, the faster it will appear when tabbed to.

mayank99: > The closer the component is to the edge of the screen, the faster it will appear when tabbed to.

I dont think that is necessarily true. The component will always appear in 200ms because that's the duration of the transition.

I'm just wondering if there's a better way to hide this link. Maybe using visually hidden styles 🤔

Have you done the research on common ways skip links are usually implemented?

src/skip-to-content/skip-to-content.scss
FlyersPh9: `@mixin iui-blur` has a `background-color` of black w/ some level of opacity based on if the blur is supported or not.

mayank99: ok I see. My comment is still valid though, as the contrast ratio needs to be higher than 4.5
image

src/skip-to-content/skip-to-content.scss
elephantcatdog: I just tested this and that is what happens. I see what you're saying, but the component will be partially visible to the user faster if it's closer to the edge, not that the component will be finished moving sooner.

I did do the research. [Here is the WebAIM article](https://webaim.org/techniques/skipnav/) Virginie posted on the issue. [Further information about invisible content is located here.](https://webaim.org/techniques/css/invisiblecontent/) WebAIM is using a similar setup for their own site.

I am not sure what you mean by "visually hidden styles" exactly. If you have a specific suggestion please share it.

mayank99: I don't see any CSS in the first link. It is just talking about skip links at a high-level. When I say "research" I was talking about practical CSS that is recommended. This can include how-to articles as well as actual code that is used in real-world websites or component libraries. Here is an example: https://github.com/reach/reach-ui/tree/main/packages/skip-nav

The second link is about invisible content in general, and it even includes the "visually hidden" technique (tldr: hide content using `clip-path` and place it in a 1x1 box). [https://webaim.org/techniques/css/invisiblecontent/#techniques](https://webaim.org/techniques/css/invisiblecontent/#:~:text=receive%20keyboard%20focus.-,css%20clip,-%7Bclip%3A%20rect)

src/skip-to-content/skip-to-content.scss
elephantcatdog: I can add this as a global variable. Where else are similar values used? I couldn't find any (in the CSS project) when I looked.

src/skip-to-content/skip-to-content.scss
elephantcatdog: I prefer the consistency of having semicolons at the end of every line.

src/skip-to-content/skip-to-content.scss
FlyersPh9: This PR has a better chance of being approved before #465.  @elephantcatdog if you want to add a variable in `/src/style/variable.scss` for `99999px` I'll use that in the toggle switch PR after this PR gets approved.

src/skip-to-content/skip-to-content.scss
elephantcatdog: I talked to Jon when determining the design for this component - he likes that the link moves from off-screen to on-screen vertically.

WebAIM specifically recommends this style:
> Probably the most accessible method for visually hiding a skip link is to hide it off screen with CSS, then cause it to be positioned on screen when it receives keyboard focus.
> One possible drawback to this approach is that the sudden appearance of a link that was previously invisible will be unexpected and could potentially confuse the sighted keyboard user. The WebAIM.org site address this by implementing CSS transitions to animate the focused "skip" link from above the top of the page (as opposed to off-screen left) to the top left corner of the page, then back above the page when focus is lost. This ensures that the link is highly visible on screen for a longer period of time.

I looked into using clip or clip-path, and it seems that the animation style would be more pop-onto-the-page rather than slide-onto-the-page. This is what skip-nav does, and tabbing quickly can mean missing the link. I suppose it could have a fade-in and/or fade-out effect, but in either case it seems that it would be less visually noticeable than the component moving.

src/skip-to-content/skip-to-content.scss
elephantcatdog: I added $iui-border-radius-huge but it is 9999px currently instead of 99999px.

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed

backstop/tests/skip-to-content.html
elephantcatdog: Fixed

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-skip-to-content {
+  @include iui-blur($opacity: 3);
+  @include iui-focus($offset: 2px, $thickness: 2px);
+
+  border-radius: $iui-border-radius-huge;
+  color: t(iui-color-foreground-accessory);
```
elephantcatdog, line 10: Color does not work correctly on IE11.

mayank99: Wrap it in `@include themed {}`

elephantcatdog: Thank you

src/skip-to-content/skip-to-content.scss
mayank99: Maybe it should be called `$iui-border-radius-round` or `$iui-border-radius-pill`. I see tailwind calls it `rounded-full` https://tailwindcss.com/docs/border-radius

backstop/tests/skip-to-content.html
mayank99: Maybe make it more realistic? 😄 but upto you

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  border-radius: $iui-border-radius-huge;
+  text-decoration: none;
+  padding: ($iui-baseline * 0.75) $iui-m;
+  position: absolute;
+  text-align: center;
+  left: 50%;
+  transform: translateX(-50%) translateY(-170%);
+  transition: background-color $iui-speed-fast ease-in-out, transform $iui-speed-slow ease-in-out,
```
mayank99, line 19: You will need to use a media query to only transition the transform if the user doesn't `prefers-reduced-motion. Same thing in the :focus rule

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,31 @@
...
+  @include iui-blur($opacity: 3);
+  @include iui-focus($offset: 2px, $thickness: 2px);
+  @include themed {
+    color: t(iui-color-foreground-accessory);
+  }
+
+  border-radius: $iui-border-radius-huge;
+  text-decoration: none;
+  padding: ($iui-baseline * 0.75) $iui-m;
+  position: absolute;
```
mayank99, line 15: This might be better as `position: fixed` so that it's based on viewport rather than scroll position

src/skip-to-content/skip-to-content.scss
FlyersPh9: `$iui-border-radius-round` makes the most sense to me.

src/skip-to-content/skip-to-content.scss
FlyersPh9:
```suggestion
transition: background-color $iui-speed-fast ease-in-out, box-shadow $iui-speed-fast ease-in-out;
@media (prefers-reduced-motion: no-preference) {
	transition: background-color $iui-speed-fast ease-in-out, box-shadow $iui-speed-fast ease-in-out, transform $iui-speed-slow ease-in-out;
}
```
This makes it so if the user does not have a "reduced motion" preference set in their operating system, they get all 3 transition effects.  But if the user does have reduced motion turned on, they only see background-color and box-shadow transition, no motion (from transform).

src/skip-to-content/skip-to-content.scss
FlyersPh9: Oh, that is a good idea, I hadn't thought of that.

src/skip-to-content/skip-to-content.scss
elephantcatdog: Most other components have the media query for all types of transitions, including background color and box shadow.

mayank99: yes that is a mistake, the media query doesn't make sense for non-motion transitions

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed

backstop/tests/skip-to-content.html
elephantcatdog: Done

backstop/tests/skip-to-content.html
```diff
@@ -0,0 +1,196 @@
...
+  lang="en-US"
+  id="theme"
+>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1"
+    >
+    <title>skip-to-content | iTwinUI</title>
```
FlyersPh9, line 16:
```suggestion
    <title>Skip to content | iTwinUI</title>
```

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-skip-to-content {
```
FlyersPh9, line 5: Let's add `user-select: none;` to avoid accidental highlighting of the text when clicking on the anchor.
Screen Shot 2022-02-10 at 6 43 06 PM

backstop/tests/skip-to-content.html
```diff
@@ -0,0 +1,196 @@
...
+    <a
+      href="#demo-skip-to-here"
+      class="iui-skip-to-content"
+      id="test-1"
+    >Skip to main content</a>
+    <theme-button></theme-button>
+    <h1>Skip to content</h1>
+    <hr />
+
+    <section id="demo-other-content">
```
veekeys, line 67: I havent read too much about `skip-to-content` but isnt the purpose of it to scroll through not important content to be able to see important/main one in viewport? If so, could we make "not important" content much bigger? I find current html a bit weird, cause the "important" content is literally there.

src/skip-to-content/classes.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './index';
+
+.iui-skip-to-content {
```
veekeys, line 5: Should we call it `iui-skip-to-content-button` maybe? Cause thats how it visually looks like. 
Not sure it is ok to call the component by the name of what it does versus what it is.

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,38 @@
...
+  @include iui-blur($opacity: 3);
```
veekeys, line 6: Minor and not connected to this PR, but we should remove `backdrop-filter` from themed mixin in `iui-blur`
image

src/skip-to-content/skip-to-content.scss
```diff
@@ -0,0 +1,38 @@
...
+  @include themed {
+    color: t(iui-color-foreground-accessory);
+  }
+
+  border-radius: $iui-border-radius-round;
+  text-decoration: none;
+  padding: ($iui-baseline * 0.75) $iui-m;
+  position: fixed;
+  text-align: center;
+  left: 50%;
```
veekeys, line 17: Currently it looks ok, because our body has top padding applied through demo css (specific only for our html pages). In real scenario, we could expect this being put without any spacing from the top. As a result, it would be just sticked to the top.. Maybe we could think of adding `top` or some `margin`... Not sure.

image

veekeys (commented):
Looks good! Left few minor comments.

src/skip-to-content/classes.scss
FlyersPh9: Technically it's not a button, but it does visually look like a button.  I don't know, I don't have a strong feeling for or against the current name.

elephantcatdog: I changed it to skip-to-content-link in some places. I think it's a link more so than a button, even though it has more styling than links normally do.

src/skip-to-content/skip-to-content.scss
elephantcatdog: Good catch, thanks. Fixed.

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed

backstop/tests/skip-to-content.html
elephantcatdog: Fixed

src/skip-to-content/skip-to-content.scss
veekeys: We usually try to use our spacing variables for that. Could you try applying `$iui-baseline` here? Maybe `2x$iui-baseline` would be alright.

elephantcatdog: Right, thanks.

src/skip-to-content/skip-to-content.scss
mayank99: @elephantcatdog I think Vykintas meant just move it out of the mixin, not completely remove it 😄

Like this:
```scss
@include themed() {
  background-color: rgba(0, 0, 0, t(iui-opacity-#{$opacity}));
}
backdrop-filter: $iui-blur-filter;
```

Although, now that I think about, if `backdrop-filter` is supported, then you don't need the IE fallback and can just use CSS variables directly:
```scss
background-color: rgba(0 0 0 / var(--iui-opacity-#{$opacity}));
backdrop-filter: $iui-blur-filter;
```

src/skip-to-content/skip-to-content.scss
elephantcatdog: Fixed.

src/skip-to-content/skip-to-content.scss
elephantcatdog: Oh, whoops, thanks for clarifying.

backstop/scenarios/skip-to-content.js
```diff
@@ -0,0 +1,18 @@
...
+  scenario('Type focus', {
+    selectors: ['viewport'],
+    viewports: [{ width: 800, height: 600 }],
+    actions: [focus('#test-1')],
+  }),
+  scenario('Type focus and hover', {
+    selectors: ['viewport'],
+    viewports: [{ width: 800, height: 600 }],
+    actions: [focus('#test-1'), hover('#test-1')],
+  }),
```
elephantcatdog, line 17: Test cases are broken becuase I no longer have docker.

src/skip-to-content/skip-to-content.scss
elephantcatdog: "backdrop-filter" does not seem to be supported on IE11, [according to caniuse](https://caniuse.com/css-backdrop-filter).

mayank99: yes, that's why we can use CSS variables. The code is wrapped inside `@supports (backdrop-filter: $iui-blur-filter)` so it will not be used in IE.

elephantcatdog: Fixed

veekeys (approved):
After tests are updated, looks good to me.

src/skip-to-content/skip-to-content.scss
mayank99: this is now wrong syntax, `t` should only be used inside `themed` mixin.

my original comment above had the right syntax (use `var` directly)

mayank99 (commented):
Looks good

mayank99 (approved):

PR #475, opened 2022-02-09
refactor: Use inline svgs for checkbox and radio

mayank99:
Greatly simplified the code for checkbox and radio so that it doesn't rely on 0x0 checkbox and can be used without a label. Absolutely no visual changes in modern browsers. IE gets to show the default checkbox.

`appearance: none` is used for hiding the default checkbox, and `::after` pseudo-element is used for displaying the checkmark/radio-dot svgs in a mask image.

Class name changes:
- `iui-checkbox` -> `iui-checkbox-wrapper`
- `iui-radio` -> `iui-radio-wrapper`
- `iui-label` -> `iui-checkbox-label` and `iui-radio-label`
- New: `input.iui-checkbox` and `input.iui-radio`
- Removed: `iui-checkbox-checkmark` and `iui-radio-dot`

Note: radio tiles and toggle switches are not influenced by this PR.


backstop/tests/checkbox.html
```diff
@@ -167,99 +109,39 @@ <h2>Default</h2>
       <div class="iui-input-container iui-with-message iui-disabled">
         <div class="iui-label">Checkbox group label</div>
         <div class="iui-input-group">
-          <label class="iui-checkbox">
+          <label class="iui-checkbox-wrapper iui-disabled">
```
mayank99, line 112: This one is an important difference. The wrapper explicitly needs `iui-disabled` now because the checkbox-label does not rely on `input:disabled` anymore. Same thing in radio.

adamhock (approved):
Your change make the checkboxes and radio buttons look a bit different on IE, but I personally think it's for the better!

Before

image

After

image

src/inputs/checkbox-radio.scss
```diff
@@ -15,160 +15,41 @@
...
-    @include themed {
-      color: t(iui-text-color-muted);
-    }
+    color: var(--iui-text-color-muted);
 
     svg {
-      @include themed {
-        fill: t(iui-icons-color-actionable-disabled);
-      }
+      fill: var(--iui-icons-color-actionable-disabled);
```
FlyersPh9, line 52: Is this still needed?

backstop/tests/checkbox.html
FlyersPh9: That is fine, #465 is [doing the same](https://github.com/iTwin/iTwinUI/blob/977bcee9873dccb2c9b4732325918d864f386dc2/src/toggle-switch/toggle-switch.scss#L23).  I will copy your class name structure for toggle switch as well (`.iui-toggle-switch-wrapper` & `.iui-toggle-switch`).

FlyersPh9 (approved):

PR #476, opened 2022-02-10
feat(Badge): Rename CSS variables for consistency

mayank99:
Renamed the following variables to make our names consistent as we increase our variable usage:
- `--badge-color` is now `--iui-badge-background-color`
- `--badge-text-color` is now `--_iui-badge-text-color`


src/badge/badge.scss
```diff
@@ -11,14 +11,17 @@ $iui-badge-background-primary: $iui-color-foreground-primary--dark;
 $iui-badge-text-color: rgba($iui-color-foreground-body--light, $iui-opacity-3--light);
 
 @mixin iui-badge {
+  --_iui-badge-background-color: #{$iui-badge-background-default};
+  --_iui-badge-text-color: #{$iui-badge-text-color};
```
mayank99, line 15: I have to keep the `$iui-badge-text-color` on line 11 because Sass is able to use hex values as the first parameter of `rgba` but CSS variables cant do that directly in the browser.

elephantcatdog, line 15: I think these variables should be considered public and not have an underscore, as iTwin UI CSS could be (or is? Blazor?) consumed by more than just our React project.

elephantcatdog, line 15: Have you updated "--badge-color" in Badge.tsx and Badge.test.tsx in the React project?

FlyersPh9 (approved):

src/badge/badge.scss
mayank99: Removed `_` from the background-color variable. The text color variable is not used externally

mayank99: That happens after the CSS PR is merged and released

elephantcatdog (approved):

PR #479, opened 2022-02-11
feat(Tile): Add responsive tile array wrapper

FlyersPh9:
Wrap `.iui-tile-array` around a group of `.iui-tile`s and the tiles will stretch / shrink to go full page width.  Includes IE fallback.

**Before**
Screen Shot 2022-02-11 at 3 04 40 PM

**After**
Screen Recording 2021-12-28 at 10 13 59 AM


backstop/tests/tile.html
```diff
@@ -1060,5 +1060,231 @@ <h2>Folder</h2>
...
+            </div>
+          </div>
+        </div>
+
+        <div
+          class="iui-tile"
+          tabindex="0"
+        >
+          <div class="iui-tile-thumbnail">
+            <svg-imodel-hollow
```
bentleyvk, line 1099: Can you try to use image instead of SVG? Because I have concerns about this width:
image

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
     @include visually-hidden;
   }
 }
+
+@mixin iui-tile-array {
+  width: 100%;
+
+  // #region IE support
+  .iui-tile {
+    margin: ($iui-line-height * 0.5) ($iui-l * 0.5);
```
bentleyvk, line 324: You can just use `$iui-sm` instead of `$iui-l * 0.5`

backstop/tests/tile.html
FlyersPh9: I've added photos to every other tile, everything appears good.

backstop/tests/tile.html
```diff
@@ -23,15 +23,22 @@
...
-      /* Set the width of all tiles */
-      .iui-tile {
-        width: 288px;
-        margin-bottom: 11px;
+      #demo-everything-turned-on,
+      #demo-default,
+      #demo-user,
+      #demo-location,
+      #demo-video,
+      #demo-folder {
```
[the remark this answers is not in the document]
FlyersPh9: I like to think that almost every time a tile component is used they would also use the `iui-tile-array` wrapper.  But I do think we should still test at least 1 section of tiles without the wrapper.  Maybe we apply the wrapper on all but one section?

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
...
+  // #region IE support
+  .iui-tile {
+    margin: $iui-baseline $iui-sm;
+  }
+
+  @supports (grid-gap: $iui-line-height $iui-l) {
+    .iui-tile {
+      margin: 0;
+    }
+  }
```
mayank99, line 331: Don't need to target `iui-tile`.

```suggestion
  > * {
    margin: $iui-baseline $iui-sm;
  }

  @supports (gap: 0 0) {
    > * {
      margin: 0;
    }
  }
```

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
...
+  // #endregion IE support
+
+  @supports (grid-gap: $iui-line-height $iui-l) {
+    display: grid;
+    grid-template-columns: repeat(auto-fill, minmax($tile-min-width, 1fr));
+    grid-gap: $iui-line-height $iui-l;
```
mayank99, line 337: This grid stuff doesn't need to be inside `@supports`

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
...
+    align-items: stretch;
```
mayank99, line 338: I think `stretch` is the default value

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
...
+
+    .iui-tile {
+      width: 100%;
```
mayank99, line 341: This shouldn't be needed ideally, as the grid is in charge of the tile width

src/tile/tile.scss
```diff
@@ -313,3 +315,30 @@
     @include visually-hidden;
   }
 }
+
+@mixin iui-tile-array {
+  width: 100%;
+
+  // #region IE support
```
mayank99, line 322: I'm wondering two things:
- does this need to go in this repo?
- does this need to support IE?

src/tile/tile.scss
FlyersPh9: If I remove this the tiles keep their fixed width.
Screen Shot 2022-02-17 at 10 17 41 AM

src/tile/tile.scss
FlyersPh9: IE support I can take it or leave it, I added it because it was pretty light weight & it didn't take much time.

Regarding where this belongs, I think it belongs in CSS rather than layouts.  If a team wanted to use our responsive tiles and nothing else this way they'd only need to import 1 pkg, not 2.

src/tile/tile.scss
FlyersPh9: You are correct, removed.

src/tile/tile.scss
FlyersPh9: Moved out of `@supports`.

src/tile/tile.scss
mayank99: Yeah, I think you'd need to change the grid formula to account for tile's fixed width. Try this:
```css
grid-template-columns: repeat(auto-fit, minmax(min(100%, 288px), 1fr));
```

src/tile/tile.scss
mayank99: welp, it's all broken now because autoprefixer cannot translate this formula.

image

You can disable autoprefixer IE grid by adding `/* autoprefixer grid: no-autoplace */` above this rule. That way we get a less broken layout.

backstop/tests/tile.html
mayank99: Can you also add the `` tag to some of the tiles? It is a supported way of showing tile thumbnails.

src/tile/tile.scss
FlyersPh9: This suggested code is giving me an error on build:
```
[scss] Error converting files to CSS: Error: Incompatible units: 'px' and '%'.
error Command failed with exit code 1.
```

src/tile/tile.scss
FlyersPh9: IE support is removed.

src/tile/tile.scss
mayank99: Seems to work in the browser so maybe this is a Sass problem again. Try hacking around the Sass min function like this:
```scss
grid-template-columns: repeat(auto-fit, minmax(unquote("min(100%, #{$tile-min-width})"), 1fr));
```

FlyersPh9: Wow, replacing `min` with `Min` does let the code compile without error, but the tiles are still appearing like they did [above](https://github.com/iTwin/iTwinUI/pull/479#discussion_r809162689) with `.iui-tile { width: 100%; }` removed.

backstop/tests/tile.html
FlyersPh9: Tile 4 & 8 within the responsive array section are using ``.  Had to add this to the CSS:
```scss
@at-root {
  img#{&} {
    object-fit: cover;
  }
}
```

src/tile/tile.scss
FlyersPh9: Is now using `> *` and using a CSS variable.  Also renamed the mixin and class to `.iui-responsive-grid`.

mayank99: Maybe we should move it to a different file then? 🤔

src/tile/tile.scss
```diff
@@ -313,3 +320,16 @@
     @include visually-hidden;
   }
 }
+
+@mixin iui-responsive-grid {
```
gretanausedaite, line 324: This would be easy to reuse in Layouts grid:
- full row item could have `grid-column: 1 / -1`
- full column item is more dificult to get in implicit grid but I'm not sure how much users would want that.

Maybe move it to separate file if we are going to use this mixing outide tiles?

src/tile/tile.scss
FlyersPh9: @gretanausedaite can you look into adding this to iTwinUI-Layouts?  Don't spend too much time on it as this decision isn't final but I'm certainly leaning more towards it after this morning's call.

src/tile/tile.scss
```diff
@@ -313,3 +320,16 @@
...
+  --_iui-grid-item-min-width: #{$tile-min-width};
```
elephantcatdog, line 325: Why did you make this a private variable instead of just using `$tile-min-width`?

gretanausedaite (commented):
We can remove this PR as I created draft PR [#14](https://github.com/iTwin/iTwinUI-layouts/pull/14) in Layouts.

PR #481, opened 2022-02-11
feat(ColorPicker): Rename CSS variables for consistency

mayank99:
| Before | After |
| --- | --- |
| `--swatch-color` | `--iui-color-swatch-background` |
| `--selected-color` | `--iui-color-picker-selected-color` |
| `--hue` | `--iui-color-field-hue` |
| `--top` and `--left` | `--iui-color-dot-inset` |


src/color-picker/color-picker.scss
```diff
@@ -12,6 +12,9 @@ $iui-active-box-shadow: 0 0 0 $iui-xxs rgba(t(iui-color-foreground-primary-rgb),
 $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
 
 @mixin iui-color-picker {
+  --iui-color-field-hue: initial;
```
veekeys, line 15: I assume should still have `iui-color-picker` prefix, cause is a part of `ColorPicker` component.

src/color-picker/color-picker.scss
```diff
@@ -12,6 +12,9 @@ $iui-active-box-shadow: 0 0 0 $iui-xxs rgba(t(iui-color-foreground-primary-rgb),
...
+  --iui-color-picker-selected-color: initial;
```
veekeys, line 16: And here is interesting situation... `selected-color` is not even used in this mixin, its used in others, so where to define it?
`hue` same, but its used in only 1 mixin, so it should go there

src/color-picker/color-picker.scss
```diff
@@ -235,9 +240,10 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb
 }
 
 @mixin iui-color-dot {
+  --iui-color-dot-inset: 0 0;
```
veekeys, line 243: sure its `0 0`? in html you defined with auto right and bottom values

src/color-picker/color-picker.scss
```diff
@@ -235,9 +240,10 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb
...
+
   position: absolute;
-  top: var(--top, 0%);
-  left: var(--left, 0%);
+  inset: var(--iui-color-dot-inset);
```
veekeys, line 246: inset not supported in IE... wont be too big change?

src/color-picker/color-picker.scss
mayank99: It is used in multiple places so I moved it up to parent

src/color-picker/color-picker.scss
mayank99: well, CSS variables are also not supported in IE, so it doesn't matter if we use inset

src/color-picker/color-picker.scss
mayank99: ok changed

src/color-picker/color-picker.scss
mayank99: it works (dot appears at top left) but still I changed to `initial`

src/color-picker/color-picker.scss
veekeys: good point :D

src/color-picker/color-picker.scss
veekeys: Hue is not, can be moved under mixin where its used

mayank99: Ok I see. If you look at the HTML (or JSX), the variable is passed in the color picker root element, so setting it to `initial` inside a child would override that value.

src/color-picker/color-picker.scss
mayank99: Well, I went with the name of the mixin that is initializing these variables. For example, it doesn't make sense to use `iui-color-picker-background` instead of `iui-color-swatch-background`.

veekeys: i was talking more like `iui-color-picker-swatch-background`, but ok. Minor

src/color-picker/color-picker.scss
mayank99: with `initial`, it just uses the default value, which is `auto` so it will be fine. I tested and it's placing the dot at the top left

src/color-picker/color-picker.scss
veekeys: so move `--iui-color-picker-hue` under `iui-color-swatch` and lets complete this pr

PR #483, opened 2022-02-11
feat(ColorPicker): Swatch supports transparency

### LINKED ISSUE #482: ColorPicker: transparent swatch
_Originally posted by @claudiaareneee in https://github.com/iTwin/itwinjs-core/issues/3197_

### Feature

Currently, if a transparent color is passed to our ColorSwatch, then it just shows the color of the background:
image
image

Our UX standards have a diagonal line to indicate transparency, so we should handle styling for it.
image

### Example

I imagine we could add a modifier class (e.g. `iui-transparent`) on `iui-color-swatch`, and then in iTwinUI-react we can use our `ColorValue` util to check if a color is transparent and automatically apply this class.

FlyersPh9:
Add support for full & semi transparent color swatches. Closes #482.

- Checkered transparent background gradient is now a mixin: `@mixin iui-transparent-background($size: $iui-m)`
- Checkered transparent background gradient is applied on the swatch, color is applied on pseudo element overlaying the background.
- Variables are using `var(--` instead of `t()`.

Screen Shot 2022-02-15 at 5 23 47 PM


FlyersPh9:
Tests are failing on 1 unrelated component for unknown reasons.  I tried re-running tests and got the same result.  Will try again on Monday.
Screen Shot 2022-02-11 at 5 45 10 PM

src/color-picker/color-picker.scss
```diff
@@ -45,17 +61,29 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb
...
 
-  @include themed {
-    box-shadow: $iui-inset-box-shadow;
+  &::after {
+    content: '';
+    position: absolute;
+    width: inherit;
+    height: inherit;
+    border-radius: inherit;
+    background-color: var(--iui-color-swatch-background);
```
mayank99, line 83: Could we "flip" the approach so that background is set normally in the swatch, and `::after` is used for the checkered pattern?

Currently this component supports IE by setting `background-color` as an inline style (in HTML/JSX), which would not be achievable if you move the background in a pseudo-element.

FlyersPh9: At first I tried applying the background color on the swatch and the checkered background on the pseudo element & using `isolation: isolate` and `z-index` but it wasn't working.

So now what I did was apply the background color on the swatch, then override it with the checkered background.  The pseudo element then inherits the background color.  Does that work for IE?

mayank99: ok, seems to work
image

the checkered pattern is absent but there's no regression

FlyersPh9: Interesting… ok, I've removed the variable.

mayank99: You could have it in a mixin and then `@include iui-transparency`.

Also while you're at it, might as well change those hover/focus variables to use `var(--` instead of `t()`

FlyersPh9: Created `@mixin iui-transparent-background` & replaced `t()` with `var(--` on `$iui-inset-box-shadow`, `$iui-active-box-shadow`, & `$iui-hover-box-shadow`.

src/color-picker/color-picker.scss
```diff
@@ -6,10 +6,10 @@
...
 $iui-color-picker-content-width: ($iui-l * 9) + ($iui-s * 8); // (9 swatches per row) + (gaps)
 $iui-color-picker-max-width: ($iui-l * 10) + ($iui-s * 9); // (10 swatches per row) + (gaps)
-$iui-color-swatch-border-radius: 5px;
-$iui-inset-box-shadow: inset 0 0 0 1px rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
-$iui-active-box-shadow: 0 0 0 $iui-xxs rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-3));
-$iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
+$iui-color-swatch-border-radius: $iui-border-radius * 2;
+$iui-inset-box-shadow: inset 0 0 0 1px rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
+$iui-active-box-shadow: 0 0 0 $iui-xxs rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-3));
+$iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
```
elephantcatdog, line 12: If you're using `$iui-xxs + 1` instead of `3px`, why not use `$iui-xxs - 1` instead of 1px?

src/color-picker/color-picker.scss
```diff
@@ -45,36 +45,37 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb
...
+  background-color: var(--iui-color-swatch-background);
+  @include iui-transparent-background;
   @supports (gap: $iui-s) {
     margin-bottom: 0;
     margin-right: 0;
   }
 
-  @include themed {
+  &::after {
+    content: '';
```
elephantcatdog, line 59: What does this do?

src/color-picker/color-picker.scss
mayank99: These haven't changed compared to before

src/color-picker/color-picker.scss
FlyersPh9: Checkout [this CodePen demo I made for you](https://codepen.io/JonGraft/pen/podadWp?editors=1100).  `content` can be used to place text before or after an element.  But if you want to place a shape, you need to pass an empty content, otherwise it won't display (see the class `.broken`).

src/color-picker/color-picker.scss
```diff
@@ -250,17 +229,36 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb
   border-radius: 50%;
   transform: translate(-$iui-s, -$iui-s);
   cursor: crosshair;
-  @include themed {
```
mayank99, line 253: Maybe put `themed` mixin back, or change `t(` to `var(--`

src/color-picker/color-picker.scss
```diff
@@ -6,10 +6,10 @@
 
 $iui-color-picker-content-width: ($iui-l * 9) + ($iui-s * 8); // (9 swatches per row) + (gaps)
 $iui-color-picker-max-width: ($iui-l * 10) + ($iui-s * 9); // (10 swatches per row) + (gaps)
-$iui-color-swatch-border-radius: 5px;
-$iui-inset-box-shadow: inset 0 0 0 1px rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
-$iui-active-box-shadow: 0 0 0 $iui-xxs rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-3));
-$iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
+$iui-color-swatch-border-radius: $iui-border-radius * 2;
```
mayank99, line 9: That's a sneaky change. Do we really want it?

src/color-picker/color-picker.scss
FlyersPh9: Right now our spacing variables are all hard coded.  I;d like for these variables to be replaced with REM which will all scale relative to `$iui-m`.  So for example, if we switched `$iui-m` from 16px to 14px, then `$iui-l` ($iui-m * 1.5) would change from 24px to 21px.

Borders on the other hand, I think we'd want them to stay 1px no matter what.

src/color-picker/color-picker.scss
FlyersPh9: It makes sense to me to have the color swatch border radius based off of `$iui-border-radius`.  It's a pretty small visual change.

mayank99: Feels too round to me personally. But if you really want to change it, I'd recommend doing it in a separate PR.

src/color-picker/color-picker.scss
FlyersPh9: Ahh, I did miss a couple.  I've swapped the `t(` that are no longer within `themed` with `var(--`.

src/color-picker/color-picker.scss
FlyersPh9: Reverted back to 5px.

elephantcatdog (approved):

PR #484, opened 2022-02-14
fix(Tree): Made `li` element focusable

bentleyvk:
Moved focus from `.iui-tree-node` to `li.iui-tree-item` because otherwise screen readers can't properly work.
In HTML there is a lot of changes but all of them are just moving `tabindex` one level up.

I also tested screen readers with sub-trees structure and it is working properly - only node data is being read. I think screen readers are smart enough to skip `ul` content.


src/tree/tree.scss
```diff
@@ -17,6 +17,28 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
   list-style: none;
 }
 
+@mixin iui-tree-item {
+  outline: none;
```
bentleyvk, line 21: This one removes outline from whole sub-tree.

src/tree/tree.scss
```diff
@@ -17,6 +17,28 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
...
+
+    &.iui-active {
+      @include themed {
+        outline: $iui-xxs solid t(iui-color-foreground-primary);
+      }
+      outline-offset: -2px;
+    }
+  }
+
+  &:focus:not(:focus-visible) > :where(.iui-tree-node:not(.iui-active)) {
```
bentleyvk, line 37: I used `:where` to reduce specificity. This should be removed with IE support drop.

src/tree/tree.scss
```diff
@@ -17,6 +17,28 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
...
+    outline: none;
+  }
```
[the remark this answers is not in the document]
bentleyvk: Because I am not so smart 😀 and it is not working on IE but who really cares... so changed.

gretanausedaite (approved):
Looks good to me 🌵

src/tree/tree.scss
mayank99: Can you move it inside focus-visible?

bentleyvk: Why?

mayank99: To avoid removing default outline in older browsers where our focus-visible mixin doesn't work

bentleyvk: So it will be even worse, because it will add focus on node and its sub-tree :/

mayank99: How is it worse than no focus ring?

bentleyvk: Idk, it is sounds bad in both ways 😀 What is the purpose of this mixin if it isn't even working on older browsers?

mayank99: It works in most browsers and helps avoid code duplication and focus bugs when comparing with the other approach (`focus:not(:focus-visible)`). I don't understand why you want to show no outline rather than the default.

bentleyvk: "most browsers" is latest ones? So it is worthless to me as it should make `:focus` work but it doesn't... But whatever, moved.

mayank99 (approved):

PR #486, opened 2022-02-15
chore: Fixed carousel tests flakiness

bentleyvk:
Turned off carousel auto-play because it causes tests to fail randomly. You have to click play for it to start.


backstop/tests/carousel.html
```diff
@@ -483,9 +483,8 @@ <h2>Default</h2>
 
         // initialize
         updateSlide(0);
```
bentleyvk, line 485: This seems to cause issues as it takes time to add required classes. I am thinking to just add those classes in HTML initially.

mayank99: Honestly, I have only seen flakiness in `Carousel` tests once or twice, so I'm not even sure about removing autoplay.

What's a lot more annoying is this `Tile` one which keeps failing in every PR:
image

mayank99: That screenshot is from this very PR. The last commit failed.

I also noticed the same failure in your previous PR (treenode).

mayank99 (approved):

FlyersPh9 (approved):

PR #487, opened 2022-02-15
chore(Select): Clean up demo page & test images.

FlyersPh9:
- `<select>` takes auto width.
- Scenarios crop the image width at 800px.
- Added padding to `<section>` so component isn't right up against the test image edge.


backstop/scenarios/select.js
```diff
@@ -4,25 +4,30 @@ module.exports = [
   // Select types
   scenario('Type Default', {
     selectors: ['#demo-default'],
+    viewports: [{ width: 800, height: 600 }],
```
veekeys, line 7: So you have changed viewport and as a result there are new images, which now is hard to compare with previous.
But also previous one should be removed then, no?

veekeys: We have script to remove old images, might help..
`yarn clean:images`

backstop/tests/select.html
```diff
@@ -24,10 +24,11 @@
     />
     <style>
       section[id^='demo'] {
-        width: 45vw;
+        width: auto;
```
veekeys, line 27: hmm...
Was checking why its needed at all, seems we have different grid implementations.
All select examples have 1fr (coming from `iui-input-container`.
While the example which uses only `.iui-input-with-icon` does not have any rules on width... 
This is expected?
@gretanausedaite @mayank99

backstop/tests/select.html
```diff
@@ -72,7 +73,7 @@ <h1>Select</h1>
     <h2>Default</h2>
     <section id="demo-default">
 
-      <section id="demo-default-singular">
+      <div id="demo-default-singular">
```
veekeys, line 76: so if not needed, remove this and pass div on label maybe?

backstop/scenarios/select.js
FlyersPh9: Removed old test images.  I didn't think that would happen since I didn't change the name of the scenario.

backstop/tests/select.html
FlyersPh9: Derp, that makes much more sense.  ID is moved to the label.

backstop/scenarios/select.js
veekeys: Now I see.. It got appended `desktop` at the end previously in images names, now it was removed after you changed viewport...
Silly situation, ideally would be good to keep same images name so it could be easier to compare the visuals :/

Also, tests are failing, cause for some reason you pushed new user icon images.

veekeys: I see we have this default viewport change in multiple components already.. and we are not using any labels like the default one:
```js
{ label: 'desktop',  width: 1920, height: 1080 },

If we ever gonna start testing responsiveness, this might be important and we would need to refactor all of this. Set a list of accepted viewports and label them, for example.
But I guess it is not this PRs scope, sometime in the future...

File: backstop/tests/select.html (hunk 2 above)
Comment (mayank99) [reply backstop/tests/select.html:27]: The `1fr` doesn't automatically give it width. It makes sense that the `input-with-icon` don't have `grid-template-columns` because it is only being used for placing the icon.

The difference only becomes apparent when you specify `fit-content` on parent. Without that, they behave identically.

Comment (FlyersPh9) [reply backstop/tests/select.html:27]: The parent in this PR has `width: auto` defined in the `<head>`.  @veekeys if you pull this locally and build, be sure to refresh the page because for whatever reason it first renders as if `width: fit-content` is set (at least on my side).

Comment (FlyersPh9) [reply backstop/tests/select.html:27]: Update:  I swapped out the Lorem ipsum in the last example and it is shrinking where I thought it would stay full width.
Screen Shot 2022-02-22 at 3 45 05 PM

Comment (veekeys) [reply backstop/tests/select.html:27]: And where is that `fit-content` set?

Comment (mayank99) [reply backstop/tests/select.html:27]: 1. `width: fit-content` is set in demo.css
2. `width: auto` is applied on all sections on this page
3. This "in context" demo is using the `<section>` as the flex container, that's why it gets messed up.

I suggest adding a wrapper inside this section to avoid fighting with previous rules.
image

**Comment by FlyersPh9** [comment]
For some reason the table in dark theme is failing tests.
Screen Shot 2022-02-23 at 2 07 43 PM
Screen Shot 2022-02-23 at 2 07 34 PM

**Comment by mayank99** [comment]
Jon. The tests are not failing anymore. You need to look for the ✔ next to "_Build and run tests_". Flaky tests are a thing we have to deal with, and re-running them usually fixes it (as it did in this case).

The ❌ next to the "_Test results_" should be ignored. These test results are added only when a test fails and exist only as a way to view the images. As far as I can tell, there is no way to remove them after re-running.

image

If it'd be less confusing, I could make it always show ✔ so that you see "less red". But then the argument could be applied from the other direction: "why is _Build and run tests_ showing red if the _Test results_ is showing green?". Unfortunately, there is no "neutral" check in github actions.

**Comment by FlyersPh9** [comment]
@mayank99 just double checking, so we could have an ❌ in the image below but tests could actually be passing?  That's what I've been typically checking to make sure my PRs are passing. 
Screen Shot 2022-02-23 at 2 38 34 PM

**Comment by mayank99** [comment]
> we could have an ❌ in the image below but tests could actually be passing?

Yes, click on the ❌ to see more info.

File: backstop/tests/select.html (hunk 2 above)
Comment (veekeys) [reply backstop/tests/select.html:27]: My point was to figure, why selects are acting different, not to set custom rules on container to fix that...
but I guess doesnt matter :/

File: backstop/tests/select.html [hunk 4]
```diff
@@ -38,8 +39,10 @@
         height: 210px;
       }
 
-      #demo-inline-combo {
-        flex-direction: row;
+      #demo-in-context {
```
veekeys, line 42: Remove this... too much custom logic on simple demo and lets merge this..

veekeys, line 42: Remove this... too much custom logic on simple demo and lets merge this..

veekeys, line 42: Remove this... too much custom logic on simple demo and lets merge this..

FlyersPh9: Reverted back to flex.

veekeys (approved):

Issue #425, opened 2021-12-22
Modal appears larger than phone display


FlyersPh9:
If a lot of text is present within a modal & the screen is small enough, it may be impossible to access the close icon or the buttons at the bottom.
![Screen Shot 2021-12-20 at 3 36 15 PM](https://user-images.githubusercontent.com/849817/147160145-d1378352-2b7e-4613-806f-2128a339b93e.png)

mayank99:
@FlyersPh9 Can you check if this is still an issue after all the changes we made to dialog in the last few months? If not, we can close it.

FlyersPh9:
It looks like the bug is still present on default & draggable dialog, full screen dialog is ok because the body area is scrollable.

**Default dialog:**
![Screenshot 2022-12-13 at 4 40 46 PM](https://user-images.githubusercontent.com/849817/207449704-cbbed8ec-90de-4680-a748-4aa4ccd0dd52.png)

**Full screen dialog:**
![Screenshot 2022-12-13 at 4 42 08 PM](https://user-images.githubusercontent.com/849817/207449993-6584b3f8-6838-472d-9800-8a70be932552.png)

**Draggable dialog:**
![Screenshot 2022-12-13 at 4 41 51 PM](https://user-images.githubusercontent.com/849817/207450053-aa22390a-1e68-4970-9bfe-3b5bb47b3fa3.png)

FlyersPh9:
@mayank99 looks like I'm having the same issue, `svh` may be the fix.
![Screenshot 2023-02-10 at 12 33 11 PM](https://user-images.githubusercontent.com/849817/218157972-25123c78-58e1-44e5-8823-988206db80c5.png)

PR #489, opened 2022-02-18
feat(Modal): Add full-page modal

### LINKED ISSUE #474: Add "full page" option to modal
#### Feature
Currently our modals appear as a box centered within the screen.  It's been requested to have the option for the modal to take full page width and height.  This full page modal will be used in one of the layouts we are considering.  

#### Examples
Current:
Screen Shot 2022-02-08 at 11 24 32 AM

Requesting:
Screen Shot 2022-02-08 at 11 25 47 AM

elephantcatdog:
Added full-page modal option.

- Full page modal with buttons along bottom of modal
- Modal translates from right side by default, but variables --iui-full-page-modal-transform-x and -y can be set to do from left, top, and bottom
- Considering adding support for custom transitions, but that will require adding a lot more CSS as variables I think
- Changed modal box-sizing from content-box to border-box

closes #474


elephantcatdog (commented):
Width doesn't appear to go all the way to the right.

backstop/tests/modal.html
```diff
@@ -40,15 +40,30 @@ <h1>Modal</h1>
...
+
+    <a
+      id="open-full-page-modal"
+      onclick="toggleModal(true)"
+      class="iui-anchor"
+    >Open full page modal</a>
+
     <script>
-      function openModal() {
+      function toggleModal(isFullPage) {
```
elephantcatdog, line 56: Javascript function does not work well and has bugs. Needs to be fixed.

src/modal/modal.scss
```diff
@@ -74,3 +75,40 @@
...
   }
 }
+
+@mixin iui-modal-full-page {
+  --iui-full-page-modal-transform-x: -100%;
+  --iui-full-page-modal-transform-y: 0%;
+
+  visibility: visible;
+  opacity: 1;
+  transform: translateX(var(--iui-full-page-modal-transform-x)) translateY(var(--iui-full-page-modal-transform-y));
```
elephantcatdog, line 85: Use translate() instead

src/modal/modal.scss
```diff
@@ -74,3 +75,40 @@
...
+
+    .iui-button-bar {
+      position: absolute;
+      bottom: $iui-baseline;
+      right: $iui-m;
+    }
+  }
+
+  &.iui-modal-visible {
+    transform: translateX(0);
```
elephantcatdog, line 109: Needs to include Y component of translate.

backstop/tests/modal.html
elephantcatdog: The problem that is that when switching between the two types of modals, the beginning transitions are cut out if the opposite modal was just viewed (or if the full-page modal is the first viewed). So the full-page modal doesn't translate onto the page and the regular modal doesn't fade onto the page (I think, the fade is harder to tell). I think this happens because the "visibility" class name is toggled on before the modal has a chance to 'reset'? I tried to put a delay on the "toggle" code but that didn't seem to do anything.

bentleyvk: Maybe just have separate modals for these tests.

src/modal/modal.scss
```diff
@@ -31,6 +31,7 @@
     border-radius: $iui-border-radius;
     box-shadow: $iui-elevation-24;
     padding: $iui-baseline $iui-m;
+    box-sizing: border-box;
```
bentleyvk, line 34: Why it is needed now?

bentleyvk: Ok, found it 😀

src/modal/modal.scss
```diff
@@ -74,3 +75,40 @@
...
+    width: 100vw;
+    left: 0;
+    top: 0;
+    transform: none;
+    max-width: initial;
+    min-width: initial;
+    border-radius: 0;
+    box-shadow: none;
+
+    .iui-button-bar {
```
bentleyvk, line 101: If there is more content, button bar will go on top of it:
image

elephantcatdog: Fixed

backstop/tests/modal.html
elephantcatdog: I changed how the full page modal is implemented a little bit and ended up separating the modals.

src/modal/modal.scss
bentleyvk: Ugh, I am not sure whether we can apply flex to the dialog because if someone has several inline elements next to each other, then we would break their layout:
image
I don't know why we didn't applied flex in the first place though... Maybe we could just apply flex only to full screen dialog.

backstop/tests/modal.html
```diff
@@ -40,13 +40,21 @@ <h1>Modal</h1>
...
 
+    <br>
+
+    <a
+      id="open-full-page-modal"
+      onclick="toggleModal('fullPageModal')"
+      class="iui-anchor"
+    >Open full page modal</a>
+
     <script>
```
bentleyvk, line 55: Can you move this whole script to the bottom of body as we have in all others HTMLs?

backstop/tests/modal.html
```diff
@@ -40,13 +40,21 @@ <h1>Modal</h1>
 
     <a
       id="open-modal"
-      onclick="openModal()"
+      onclick="toggleModal('myModal')"
```
bentleyvk, line 43: Really strange modal id 😀 Can you rename it to `default-modal` maybe?

backstop/tests/modal.html
```diff
@@ -81,13 +90,60 @@ <h1>Modal</h1>
...
+          >
+            <span class="iui-button-label">Cancel</span>
+          </button>
+        </div>
+      </div>
+    </div>
+
+    <div
+      class="iui-modal iui-modal-full-page"
+      id="fullPageModal"
```
bentleyvk, line 109: Rename to `full-page-modal`

src/modal/modal.scss
FlyersPh9: I like the idea of including `.iui-modal-content` in both default and full page modal.  All user content would need to go within this class.  I don't think the content inside there would have the issue you described, right?

src/modal/modal.scss
```diff
@@ -66,6 +69,11 @@
         margin-right: $iui-s;
       }
     }
+
+    .iui-modal-content {
+      flex-grow: 2;
+      overflow: scroll;
```
FlyersPh9, line 75:
```suggestion
      overflow-y: auto;
      overflow-y: overlay;
```
I would suggest targeting only overflow on the Y axis.  Also I'd like to use overlay with a fallback for browsers that don't support it.

src/modal/modal.scss
```diff
@@ -74,3 +75,40 @@
...
+
+  visibility: visible;
+  opacity: 1;
+  transform: translate(var(--iui-full-page-modal-transform-x), var(--iui-full-page-modal-transform-y));
+  @media (prefers-reduced-motion: no-preference) {
+    transition: transform $iui-speed ease-in;
+  }
+
+  > .iui-modal-dialog {
+    height: 100vh;
```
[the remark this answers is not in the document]
FlyersPh9: That's a good question.  The current implementation would go above the header and side navigation.  I actually don't know what would be best here.

src/modal/modal.scss
```diff
@@ -74,3 +82,34 @@
...
   }
 }
+
+@mixin iui-modal-full-page {
+  --iui-full-page-modal-transform-x: 100%;
+  --iui-full-page-modal-transform-y: 0%;
+
+  visibility: visible;
+  opacity: 1;
+  transform: translate(var(--iui-full-page-modal-transform-x), var(--iui-full-page-modal-transform-y));
```
FlyersPh9, line 92: Can we try using `scale(0)` to `scale(1)` instead of animating in from the side?  When it animates in from the right it makes me think we are making progress (moving forward in a step like a wizard) which may not always be the case.  If you use scale you'll need to transition the transform, opacity, and visibility.  I think using scale will also allow you to remove lines 90-91.

src/modal/modal.scss
```diff
@@ -74,3 +82,34 @@
...
+    transform: none;
+    max-width: initial;
+    min-width: initial;
+    border-radius: 0;
+    box-shadow: none;
+  }
+
+  &.iui-modal-visible {
+    transform: translate(0, 0);
+    @media (prefers-reduced-motion: no-preference) {
```
FlyersPh9, line 111: If both transitions are using the same ease value, this should not be necessary.

src/modal/modal.scss
```diff
@@ -66,6 +69,11 @@
         margin-right: $iui-s;
       }
     }
+
+    .iui-modal-content {
```
FlyersPh9, line 73: I really like this idea.  If this gets implemented you should be able to apply `max-height: 100vh;` to the base `.iui-modal-dialog` and would solve #425 .

src/modal/modal.scss
bentleyvk: But users will need to do this change, otherwise it might break their current layouts.

src/modal/modal.scss
gretanausedaite: Agree. Also maybe it's good idea to use dark background overlay (same as in not full page modals) to show users what is happening. Currently it's hard to see modal border when it's appearing from the side.

mayank99: I would have thought it should animate from the bottom. That's what I'm used to seeing everywhere.

But also this animation will not work in React. Currently it relies on the dialog being present (and hidden) in the DOM, whereas in React, there isn't a distinction between "hidden" or "visible": it's either present in the DOM (when open). or it isn't (when closed).

src/modal/modal.scss
elephantcatdog: They are not using the same ease value. I am following [the suggestions made here](https://css-tricks.com/ease-out-in-ease-in-out/#:~:text=There%20are%20two,and%20without%20delay.) for moving on and off the screen.

src/modal/modal.scss
elephantcatdog: Done. This only solves #425 if the content is wrapped in `iui-modal-content`

src/modal/modal.scss
elephantcatdog: Done

backstop/tests/modal.html
elephantcatdog: Done

backstop/tests/modal.html
elephantcatdog: Done

backstop/tests/modal.html
elephantcatdog: Done

src/modal/modal.scss
elephantcatdog: I added separate mixins for slide animations vs scale animations. Slide animations now default from the bottom. I removed
`visibility: visible;
opacity: 1;`
from the slide class and am now dealing with it in the transition.

elephantcatdog: I tried to add the dark background to the full page modal during the transitions, but it's more complicated than I anticipated and need to think about it more. The dark background is currently being covered by the full page modal and moving with it (slide or scale), so I'm not sure how to have that part remain constant without adding another HTML element.

src/modal/modal.scss
bentleyvk: Yeah, I think it is the best approach right now.

src/modal/modal.scss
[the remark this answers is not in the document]
bentleyvk: Regular modal's overlay goes over everything so maybe it is ok that full page modal covers everything too.

src/modal/modal.scss
elephantcatdog: This will be fixed in CSS 1.0, when flex is added to both modals.

src/modal/modal.scss
elephantcatdog: I've moved the flex and .iui-modal-content to the full-page-modal scss only.

src/modal/modal.scss
elephantcatdog: I added the dark background. There are new videos in the PR description.

src/modal/modal.scss
```diff
@@ -71,6 +73,72 @@
...
+    width: 100vw;
+    left: 0;
+    top: 0;
+    transform: none;
+    max-width: initial;
+    min-width: initial;
+    border-radius: 0;
+  }
+
+  .iui-modal-content {
```
bentleyvk, line 94: I think you need to move this out of full-modal because regular modal users could want to use it and also in React I think we will create separate component for this one.

gretanausedaite (commented):
- Maybe in React component we could give our users a choice from where slide animation starts (top/bottom/right/left). Do we need animations here or we can do those in React (like toast notification, side nav, expandable block)?
- Moving `iui-modal-content` outside full page modal is great idea.
- Nice improvement adding `iui-speed-instant` variable 👍

elephantcatdog (commented):
The transitions I have don't affect the modal in React. I tried using `animation` in css without success.

src/modal/modal.scss
```diff
@@ -67,11 +69,77 @@
         margin-right: $iui-s;
       }
     }
+
+    .iui-modal-content {
+      flex-grow: 2;
```
FlyersPh9, line 74:
```suggestion
      flex-grow: 2;
      margin: 0 -#{$iui-m};
      padding: 0 $iui-m;
```
Can we add this margin & padding so that the scrollbars aren't indented?

**Currently:**
Screen Shot 2022-03-09 at 2 23 44 PM

**With this change:**
Screen Shot 2022-03-09 at 2 22 59 PM

backstop/tests/modal.html
```diff
@@ -62,37 +86,141 @@ <h1>Modal</h1>
...
+            onclick="toggleModal('full-page-modal-slide')"
+          >
+            <span class="iui-button-label">Cancel</span>
+          </button>
+        </div>
+      </div>
+    </div>
+
+    <div
+      class="iui-modal iui-modal-full-page iui-modal-full-page-scale-animation"
```
FlyersPh9, line 173: Let's remove the scaling transition & only use the slide up from the bottom transition.  That way `iui-modal-full-page` can include the transition within it & no need for an extra modifier class like `.iui-modal-full-page-scale-animation`.

elephantcatdog: I did it, but are you sure? You really liked the scale. I can include it in react also.

src/modal/modal.scss
elephantcatdog: Done

src/modal/modal.scss
```diff
@@ -67,11 +78,52 @@
...
+    height: 100vh;
+    width: 100vw;
+    left: 0;
+    top: 0;
+    transform: none;
+    max-width: initial;
+    min-width: initial;
+    border-radius: 0;
+
+    &.iui-modal-animation-enter {
```
bentleyvk, line 107: So it should work in React now? @elephantcatdog @mayank99

mayank99: Yeah, but we only spent ~30 mins on it. There might still be some CSS changes needed, so let's wait for the React PR (at least draft).

bentleyvk: You mean we shouldn't merge this PR until React one is created?

mayank99: Correct.

backstop/tests/modal.html
FlyersPh9: Yeah, just for simplicity let's only support 1 option for initial release.  We can think about adding other options in the future.

src/modal/modal.scss
```diff
@@ -67,11 +78,52 @@
...
+    }
   }
+}
 
-  &.iui-modal-visible {
-    visibility: visible;
-    opacity: 1;
-    transition-delay: 0s;
+@mixin iui-modal-full-page {
+  > .iui-modal-dialog {
```
mayank99, line 95: We can reduce specificity/nesting by applying the full page modifier on the dialog itself, instead of the parent

src/modal/modal.scss
```diff
@@ -67,11 +78,52 @@
...
     }
+
+    .iui-modal-content {
+      flex-grow: 2;
+      margin: 0 -#{$iui-m};
+      padding: 0 $iui-m;
+      overflow-y: auto;
+      @supports (overflow-y: overlay) {
+        overflow-y: overlay;
+      }
```
mayank99, line 89: There's no need to wrap this in `@supports`. It will be ignored in browsers that don't support it

```suggestion
      overflow-y: overlay;
```

backstop/scenarios/modal.js
```diff
@@ -7,4 +7,10 @@ module.exports = [
     actions: [click('#open-modal'), hover('body')],
     selectors: ['document'],
   }),
+  scenario('full-page-modal', {
+    // Hovering `body` to avoid random test failures
+    // when "Open modal" sometimes is and sometimes is not hovered.
+    actions: [click('#open-full-page-modal'), hover('body')],
+    selectors: ['document'],
```
FlyersPh9, line 14:
```suggestion
    selectors: ['document'],
    hideSelectors: ['h1', 'hr', 'theme-button', '#open-modal', '#open-full-page-modal'],
```
Can you add this `hideSelectors` part to both scenarios?  It will hide the non-relevant stuff from the test images.

src/modal/modal.scss
elephantcatdog: Done. So `@supports` doesn't need to be used if the condition is exactly the same as the body?

backstop/scenarios/modal.js
elephantcatdog: Done

src/modal/modal.scss
elephantcatdog: Good point, but I changed this and it doesn't apply anymore.

elephantcatdog:
My most recent changes added a lot of code for the transitions, but! I think it's worth it.

So, the way I'm doing animations now allows for the same animation to be present in both CSS and React. I made regions for each: small modal animations for CSS, small modal animations for React, full page modal animations for CSS, and full page modal animations for React. There are some overlapping selectors that are separated from each other, and some of the code should be the same for CSS and React transitions, but I think having separate sections makes the code more easily readable.

Additionally, moving the React animations from the modal dialog to the modal parent allows all of the modal to be removed from the DOM. Previously the modal parent/background overlay was remaining in the DOM in React. Also there was an issue with recreating the slide animation for the full page modal.

FlyersPh9 (approved):
To me, this looks good.  I'd [wait to see if this comment is resolved](https://github.com/iTwin/iTwinUI/pull/489#discussion_r826280278) before merging.

src/modal/modal.scss
elephantcatdog: React PR: https://github.com/iTwin/iTwinUI-react/pull/588

src/modal/modal.scss
```diff
@@ -11,15 +11,53 @@
...
   width: 100%;
   height: 100%;
+  @include themed {
+    background-color: rgba(0, 0, 0, t(iui-opacity-4));
+  }
+
+  // #region Small Modal Animations for CSS
   visibility: hidden;
   opacity: 0;
   @media (prefers-reduced-motion: no-preference) {
```
bentleyvk, line 21: It should be enough just to have transition in one place, you don't need to duplicate in every class.

bentleyvk:
Is it only me, or when window is resized to be smaller, modal gets blurry?
image
I checked deployed/original one and it is showing normally there.

src/modal/modal.scss
```diff
@@ -11,15 +11,53 @@
...
+
+  &.iui-modal-animation-enter-active {
+    visibility: visible;
+    opacity: 1;
+    @media (prefers-reduced-motion: no-preference) {
+      transition: visibility $iui-speed-instant linear, opacity $iui-speed-fast ease-out;
+    }
+  }
+
+  &.iui-modal-animation-exit {
```
bentleyvk, line 47: Is this class really needed?

src/modal/modal.scss
```diff
@@ -28,9 +66,12 @@
     z-index: 1000;
     max-width: 50%;
     min-width: 400px;
+    max-height: 100vh;
     border-radius: $iui-border-radius;
     box-shadow: $iui-elevation-24;
     padding: $iui-baseline $iui-m;
+    box-sizing: border-box;
+    will-change: transform;
```
bentleyvk, line 74: It only applies to full page modal as default one has only opacity animation.

bentleyvk: Also, I see it is causing this strange bug https://github.com/iTwin/iTwinUI/pull/489#issuecomment-1068816612. Because when removed, modal shows everything clearly.

bentleyvk: And it only causes the issue in regular modal, so definitely move it to full modal CSS.

src/modal/modal.scss
```diff
@@ -67,11 +108,99 @@
...
+    > .iui-modal-dialog {
+      transform: translateY(0);
+      opacity: 1;
+      @media (prefers-reduced-motion: no-preference) {
+        transition: transform $iui-speed ease-out;
+      }
+    }
+  }
+
+  &.iui-modal-animation-exit {
```
bentleyvk, line 180: Is it needed?

backstop/scenarios/modal.js
```diff
@@ -6,5 +6,13 @@ module.exports = [
     // when "Open modal" sometimes is and sometimes is not hovered.
     actions: [click('#open-modal'), hover('body')],
     selectors: ['document'],
+    hideSelectors: ['h1', 'hr', 'theme-button', '#open-modal', '#open-full-page-modal'],
```
bentleyvk, line 9: Maybe worth changing to more universal selector: `body > *:not(.iui-modal)`

src/modal/modal.scss
```diff
@@ -11,15 +11,53 @@
...
   }
 
+  &.iui-modal-animation-exit-active {
+    visibility: hidden;
+    opacity: 0;
+    @media (prefers-reduced-motion: no-preference) {
+      transition: visibility $iui-speed-instant linear $iui-speed-fast, opacity $iui-speed-fast ease-out;
+    }
+  }
+  // #endregion Full Page Modal Animations for React
```
gretanausedaite, line 59: Is this region name correct?

src/modal/modal.scss
```diff
@@ -28,9 +66,12 @@
     z-index: 1000;
     max-width: 50%;
     min-width: 400px;
```
gretanausedaite, line 68: Changing this min-width to 380px makes transition from `default` screen to `max-width: 400px` smoother.

gretanausedaite: https://user-images.githubusercontent.com/83585998/158596719-6f3ba969-1db3-4305-9f46-f028ab4a41c2.mov

vs

https://user-images.githubusercontent.com/83585998/158596736-e2941d90-1038-4899-a144-f49e05688a7b.mov

gretanausedaite (commented):
I agree with Vilius comment in Teams that it's probably worth it to change centering to flex.

bentleyvk:
> I agree with Vilius comment in Teams that it's probably worth it to change centering to flex.

But maybe not with this PR 😀 It is enough to remove `will-change: transform` or move it to full-modal CSS.

src/modal/modal.scss
elephantcatdog: Done

src/modal/modal.scss
elephantcatdog: Yes, the name is correct. This is the end of the section that deals with animations for the full page modal using CSSTransitions in React. It was placed a bit off though.

backstop/scenarios/modal.js
elephantcatdog: Done

src/modal/modal.scss
elephantcatdog: Done

bentleyvk (approved):
Looks good! Don't forget to add this to merge message `Release-As: 0.52.0`

PR #492, opened 2022-02-18
fix(Radio): Set blank svg for indeterminate state

mayank99:
Apparently, [it is possible](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate#:~:text=%3Cinput%20type%3D%22radio%22%3E%20elements%2C%20when%20all%20radio%20buttons%20with%20the%20same%20name%20value%20in%20the%20form%20are%20unchecked) for radio inputs to be in the indeterminate state, so in the current implementation they get the unchecked svg.
image

This change will explicitly set the mask image to use a blank svg.
image


FlyersPh9 (approved):
That's kind of a hilarious bug!  Solution makes sense.

PR #496, opened 2022-02-23
ci: Create successful 'Test results' check if not failed

mayank99:
This will fix the confusing red ❌ that stays when flaky tests are re-run, by overriding with a successful "Test results" check.


bentleyvk:
Try to push some visual changes and then undo them.

mayank99:
> Try to push some visual changes and then undo them.

That's not the same. We want to test whether re-running on the same commit will override old check.

bentleyvk:
> > Try to push some visual changes and then undo them.
> 
> That's not the same. We want to test whether re-running on the same commit will override old check.

But aren't we also have this problem that after fixing visual tests, run results are not updated and that one step to publish visual tests results still shows that it failed?

mayank99:
No, I have not noticed any issues with pushing a new commit.

mayank99:
After about 20 attempts, I have been able to test the expected behavior. This last commit failed and succeeded on a subsequent re-run
image

All green checks
image

veekeys (approved):

PR #497, opened 2022-02-25
ci: Exclude main branch when creating check

mayank99:
This should fix the failing build in main branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
           git-config-email: github-actions[bot]@users.noreply.github.com
 
       - name: Post gh pages link under Checks
-        if: always()
+        if: github.ref != 'refs/heads/main'
         uses: actions/github-script@v5
         with:
           script: |


.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           git-config-email: github-actions[bot]@users.noreply.github.com
 
       - name: Post gh pages link under Checks
-        if: always()
+        if: github.ref != 'refs/heads/main'
```
veekeys, line 52: but we need deploy results, if visual tests fail in main branch

bentleyvk, line 52: But it won't post if build fails on main? Or it really needs PR number or something similar for that to work?

bentleyvk (approved):

.github/workflows/build.yml
mayank99: we do
image

mayank99: it will
image

PR #498, opened 2022-02-25
feat(Table): Add column reordering

FlyersPh9:
Added to support [feat(Table): support for draggable columns #561](https://github.com/iTwin/iTwinUI-react/pull/561).

- Add `.iui-reorder-column-left` & `.iui-reorder-column-right` classes to give a border to show where the column is being placed.
Screen Shot 2022-02-25 at 11 45 23 AM
- Add `.iui-grabbing` modifier to give `cursor: grabbing;` when the column is being dragged (similar to slider).


src/table/table.scss
```diff
@@ -99,6 +103,12 @@
     }
   }
 
+  .iui-reorder-column {
+    @include themed {
+      border-right: $iui-xxs solid t(iui-color-foreground-primary);
```
veekeys, line 108: Not sure if its always right though

https://github.com/iTwin/iTwinUI-react/pull/561/files#r814875116

src/table/table.scss
```diff
@@ -40,6 +40,10 @@
       cursor: pointer;
     }
 
+    &.iui-grabbing {
+      cursor: grabbing;
```
veekeys, line 44: You might also want to indicate somehow, that column is available for reorder.
`cursor: grab;`, again similar to slider

src/table/table.scss
FlyersPh9: Updated to work on left or right.

src/table/table.scss
FlyersPh9: I thought about that, but the entire column header is clickable and is using `cursor: pointer;`.

mayank99: Maybe a light border on hover?

veekeys: but wont it be weird, when cursor from pointer will change to grabbing ?

veekeys (approved):
If the UI is fine, for initial version we can have just this, so PR in react would not be stopped by this.

src/table/table.scss
FlyersPh9: At least on hover the cursor changes to the pointer hand, so its not a weird transition from `default` to hand.

PR #502, opened 2022-02-28
fix(Modal): Fix clipping of text

### LINKED ISSUE #485: Modal title is slightly cut on Safari if Open Sans font is used

Modal title is being slightly cut at the bottom.

image

Bug is reproducible only on Safari with Open Sans font used (tried to reproduce same issue in storybook, but due to font differences everything looks good).

#### Expected behavior
Text should fit to height on all browsers.

FlyersPh9:
Fixes clipping on the descenders of typography (look at the bottom of the g & y characters).

**Before:**
Screen Shot 2022-02-28 at 1 26 45 PM

**After:**
Screen Shot 2022-02-28 at 1 27 10 PM

Closes #485.


src/modal/modal.scss
```diff
@@ -53,6 +53,7 @@
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+        line-height: normal;
```
mayank99, line 56: This keyword is reliant on the font family but usually means 1.2. Are you sure you want to use this?

What happened to the `align-items` fix you had?

FlyersPh9: `align-items: stretch` resulted in the text not center aligning with the close button.

Screen Shot 2022-02-28 at 1 43 03 PM

mayank99: What if add another layer of flex for centering `.iui-title`?

```suggestion
        align-self: stretch;
        display: flex;
        align-items: center;
```

FlyersPh9: That does visually align but it breaks truncation.
Screen Shot 2022-02-28 at 2 46 18 PM

[Looking into this right now](https://css-tricks.com/flexbox-truncated-text/) but not working on my initial try.

mayank99: Then maybe just use `line-height: 1.5`.

This is with `line-height: normal`:
image

This is with `line-height: 1.5` (notice the text has shifted down a bit, and is better centered)
image

FlyersPh9: Changed to `line-height: 1.5;`.

mayank99 (approved):

PR #505, opened 2022-03-02
feat(theme): Add high contrast themes

FlyersPh9:
- Add high contrast light theme with AAA score.
- Add high contrast dark theme with AAA score.
- Updated `theme-button`.
	- Updated scenarios of demo pages that included the `theme-button` in the test results since these tests were already visually changing.
- `::selection` opacity changed from opacity 4 to 5.
- Add `color.html` demo page for testing contrast easily.
- Demo.css changes:
	- Use CSS variables for `<hr>`.
	- Add `overflow: overlay` to test scrollbar above different background colors in `color.html`.
- Introducing `$iui-color-background-border` which displays background-5 in default themes but shows foreground-body w/ opacity in high contrast themes.
- Tooltip has a white border to help stand out above any background color.
- Selection overlay opacity is lowered.
- Add inner white circle to active carousel dot.
- Muted text over any overlay background color has had opacity removed.  This is to score a AAA in active header button & selected menu item.
- Header secondary text is using smaller font to align with menu items.
- Active menu item uses blue text to align with header, side nav.


backstop/assets/theme.js
```diff
@@ -1,4 +1,4 @@
-const THEMEAVAILABLE = ['iui-theme-light', 'iui-theme-dark'];
+const THEMEAVAILABLE = ['iui-theme-light', 'iui-theme-dark', 'iui-theme-light-hc', 'iui-theme-dark-hc'];
```
veekeys, line 1: Would it make sense to have another button to reach those themes faster...?

veekeys (commented):
I really liked dark theme and pink color the most!
Light theme was weird, though... Looked even more authentic than current one

backstop/assets/theme.js
bentleyvk: And all our test images changes 😕 Maybe right-click could expand dropdown with all themes.

veekeys: Most of our test images should take sections with components, not full page, I think...

FlyersPh9: For test images that do have the theme button within, we should be using something like `hideSelectors: ['theme-button'],` in the scenario JS.

backstop/assets/theme.js
```diff
@@ -37,10 +37,12 @@ class ThemeButton extends HTMLElement {
         display: inline-block;
         margin-left: 0.25rem;
       }
-      :host-context(.iui-theme-dark) svg {
+      :host-context(.iui-theme-dark) svg,
```
mayank99, line 40: This will not work in Firefox and Safari 🤔 Could we just use `currentColor` instead?

src/style/theme.scss
```diff
@@ -60,6 +61,62 @@ $iui-icons-color-positive--light: $iui-color-foreground-positive--light;
...
+$iui-color-background-4--light-hc: #D9D9D9;
+$iui-color-background-5--light-hc: #CCC;
+$iui-color-background-primary--light-hc: #0068A8;
+$iui-color-background-positive--light-hc: #37740A;
+$iui-color-background-negative--light-hc: #B90909;
+$iui-color-background-warning--light-hc: #965608;
+$iui-color-background-disabled--light-hc: $iui-color-background-3--light-hc;
+
+// Light high contrast theme > Foreground
+$iui-color-foreground-body--light-hc: #000;
```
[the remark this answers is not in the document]
veekeys: You can read `hsl` colors? :O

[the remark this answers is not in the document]
mayank99: Honestly, it is much more intuitive and easy to understand compared to hex or rgb. 

Can you read hex colors? What color is `#965608` and how would you make it 20% lighter?

backstop/assets/theme.js
mayank99: see #511

backstop/assets/theme.js
mayank99: fixed in #511

mayank99:
After #508 and #511 get merged, we also need to check two more things:
1. Contrast ratio should be at least 7:1 to meet WCAG AAA requirements (otherwise it's not really "high" contrast). 4.5:1 should be met in our normal themes.
   - https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html
   - https://webaim.org/articles/contrast/#sc146
   - https://blog.prototypr.io/high-contrast-mode-ftw-bef74e097d1e
2. Should work with forced high contrast mode in Windows.
   - https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/
   - https://blog.khanacademy.org/making-websites-work-with-windows-high-contrast-mode/

veekeys:
> After #508 and #511 get merged, we also need to check two more things:
> 
> 1. Contrast ratio should be at least 7:1 to meet WCAG AAA requirements (otherwise it's not really "high" contrast). 4.5:1 should be met in our normal themes.
>    
>    * [w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html)
>    * [webaim.org/articles/contrast/#sc146](https://webaim.org/articles/contrast/#sc146)
>    * [blog.prototypr.io/high-contrast-mode-ftw-bef74e097d1e](https://blog.prototypr.io/high-contrast-mode-ftw-bef74e097d1e)
> 2. Should work with forced high contrast mode in Windows.
>    
>    * [blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/)
>    * [blog.khanacademy.org/making-websites-work-with-windows-high-contrast-mode](https://blog.khanacademy.org/making-websites-work-with-windows-high-contrast-mode/)

You can check page situation with Css Overview in Chrome dev tools.

image

Not sure how trustful it is, though...

FlyersPh9:
> 1. Contrast ratio should be at least 7:1 to meet WCAG AAA requirements (otherwise it's not really "high" contrast). 4.5:1 should be met in our normal themes.

Our current light / dark themes do not meet WCAG AA.  The new high contrast light / dark theme meet WCAG AA but not AAA.  We could make these new themes the defaults since they do meet AA and then create new high contrast themes that meet AAA.

backstop/tests/color.html
```diff
@@ -0,0 +1,255 @@
...
+          <li class="demo-text-warning-muted">Warning muted</li>
+        </ul>
+      </div>
+      <div class="demo-status-overlay-negative">
+        <h2 class="iui-text-title iui-text-spacing">Overlay negative</h2>
+        <ul>
+          <li class="iui-text-block">Text</li>
+          <li class="iui-text-block iui-text-muted">Muted</li>
+          <li class="demo-text-negative">Negative</li>
+          <li class="demo-text-negative-muted">Negative muted</li>
```
FlyersPh9, line 249: Positive / Warning / Negative muted text on overlay background are not present anywhere in our UI.  However primary muted text on overlay is present in the header buttons.

backstop/tests/color.html
```diff
@@ -0,0 +1,255 @@
...
+        <ul>
+          <li class="iui-text-block">Text</li>
+        </ul>
+      </div>
+
+      <div class="demo-status-overlay-primary">
+        <h2 class="iui-text-title iui-text-spacing">Overlay primary</h2>
+        <ul>
+          <li class="iui-text-block">Text</li>
+          <li class="iui-text-block iui-text-muted">Muted</li>
```
FlyersPh9, line 220: I expanded on the tests on this page and discovered yesterday that muted text on overlay primary is failing in light HC.  This is the only fail I am finding across both light & dark HC themes.  Solutions:
1. Increase `opacity-3` to passing.
2. Change muted text to use `opacity-2`.
3. Remove instances of muted text over primary overlay.

Muted text over primary overlay are present in header buttons, hovered select menu items for examples.

FlyersPh9:
> 1. Contrast ratio should be at least 7:1 to meet WCAG AAA requirements (otherwise it's not really "high" contrast). 4.5:1 should be met in our normal themes.

As of right now all text, with [1 exception](https://github.com/iTwin/iTwinUI/pull/505#discussion_r827270536), is passing with AAA.  Same goes for important interface elements (border of checkboxes) and icons.

> 2. Should work with forced high contrast mode in Windows.

`@media (forced-colors: active)` is new to me.  I will look into if I can replicate this on macOS.  From what I'm seeing [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors#css) this might be on a component by component basis.

mayank99:
@FlyersPh9 This looks a lot better! I tried to go through all components, and most places, it's passing AAA, but some of the hover/focus/active states are not.
image
image

Also, disabled/muted/placeholders are obviously failing. I know they are not required to pass AAA, but we might still want to increase the contrast a little bit because some of them are really hard to see.
image

The muted text is also used in footer, which has low visibility and is hard to even tell that it's made of links.
---

One more issue I did notice is that because the dark theme background is so... dark, all our shadows are lost.
image

This is also an issue with translucent backgrounds such as our tooltip where the boundary is lost.
image

So you might want to adjust some of the background colors a little bit, or add borders. https://twitter.com/steveschoger/status/1499780803550658567

backstop/tests/color.html
FlyersPh9: This is addressed in #529.

mayank99:
> [Lines 396-408](https://github.com/iTwin/iTwinUI/blob/2f9b08228690d379a86a64bcb1a1fad432832551/src/style/theme.scss#L396) seem to be overriding changes set above.
> 
> The placeholder text opacity is set to 5, and I'd like to try changing it to 4 for high contrast themes, but I'm not seeing the change.

Looks like it is _meant_ to override the other stuff and it's not possible to add theme-specific values there. You could try moving the `iui-text-color-placeholder` from `$cssVariables` into each of the themes above.

elephantcatdog (commented):
I'm not sure how much we control of this, but the colors of many of the user icons don't pass AAA (or even AA). The color on the background or the (white) initials on the color.

Not sure how important this is, but there are four images that don't visually appear to have changed - the dark and light versions of the basic modal and the "session timed out" screen.

FlyersPh9:
> I'm not sure how much we control of this, but the colors of many of the user icons don't pass AAA (or even AA). The color on the background or the (white) initials on the color.

Hmm, I'm not sure how we can best handle this.  Maybe we add a `text-shadow` only when high contrast is on?  Let me think on this for a little bit.

elephantcatdog:
I don't know how well this would work, but you could add an overlay with box-shadow so that user icons are shaded with white for dark theme and black for light theme, and the text color should be black for dark theme (light icons) and white for light theme (dark icons). This does mess with the box-shadows already present though. I don't like the way this looks exactly - something is off with the text color, and I didn't get it working for the stacked lists, but something like this:
image
image
1st one has `box-shadow: inset 0 0 0 200px rgb(255 255 255 / 50%); color: black;` and 2nd one has `box-shadow: inset 0 0 0 200px rgb(0 0 0 / 60%);` in `iui-initials`. But with this method you have to make assumptions about how light/dark the color is initially.

src/tooltip/tooltip.scss
```diff
@@ -20,6 +20,7 @@
   @include themed {
     background-color: rgba(0, 0, 0, t(iui-opacity-3));
     color: t(iui-color-foreground-accessory);
+    border: 1px solid rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-4));
```
mayank99, line 23: Can we avoid this change in default themes? Looks really bad in default dark

src/skip-to-content/skip-to-content.scss
```diff
@@ -3,11 +3,8 @@
 @import '../style/index';
 
 @mixin iui-skip-to-content-link {
-  @include iui-blur($opacity: 3);
+  @include iui-blur($opacity: 3, $invert: true);
```
mayank99, line 6: I will say it again, can we avoid this change in default dark theme?

src/carousel/carousel.scss
```diff
@@ -120,6 +124,8 @@
     &.iui-active::after {
       @include themed {
         background-color: t(iui-color-foreground-primary);
+        border-color: t(iui-color-foreground-primary);
+        box-shadow: inset 0 0 0 1px t(iui-color-background-1);
```
mayank99, line 128: box-shadow will not work with forced colors mode.

src/carousel/carousel.scss
```diff
@@ -96,12 +96,16 @@
...
       height: $iui-icons-small;
       border-radius: 50%;
-      transition: background-color $iui-speed-fast ease;
+      box-sizing: border-box;
+      border: 1px solid transparent;
+      transition: background-color $iui-speed-fast ease, box-shadow $iui-speed-fast ease;
       @media (prefers-reduced-motion: no-preference) {
-        transition: background-color $iui-speed-fast $iui-speed-fast ease, transform $iui-speed ease;
+        transition: background-color $iui-speed-fast $iui-speed-fast ease, transform $iui-speed ease,
+          box-shadow $iui-speed-fast ease, border-color $iui-speed-fast ease;
```
mayank99, line 104: These changes are going to conflict with #544 so we should decide now how we want to handle that

src/style/global.scss
```diff
@@ -51,4 +70,32 @@ html {
       @include theme-variables(dark);
     }
   }
+
+  @media (prefers-color-scheme: light) and (prefers-contrast: more) {
+    &:not([class*='iui-theme-']) {
+      @include theme-variables(light-hc);
+
+      .iui-anchor {
```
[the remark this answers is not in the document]
FlyersPh9: How I about I remove this from this PR and introduce an anchor component _after_ this one merges?

src/carousel/carousel.scss
FlyersPh9: I've added `@media (forced-colors: active) { border-width: $iui-icons-small * 0.5; }`.
Screen Shot 2022-03-23 at 2 11 50 PM

src/tooltip/tooltip.scss
FlyersPh9: This one I feel pretty strongly that we need to leave in.  The tooltip could appear above an AppUI frame and we have no idea what color might be underneath.
Screen Shot 2022-03-23 at 2 14 21 PM
 
 I wouldn't say it looks bad either, both GitHub and VSC have a light border on their tooltips.

src/style/global.scss
```diff
@@ -51,4 +59,16 @@ html {
...
+    &:not([class*='iui-theme-']) {
+      @include theme-variables(light-hc);
+    }
+  }
+
+  @media (prefers-color-scheme: dark) and (prefers-contrast: more) {
+    &:not([class*='iui-theme-']) {
+      @include theme-variables(dark-hc);
+    }
+  }
```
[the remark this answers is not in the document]
FlyersPh9: I pushed this change.  I ran tests & tried testing myself to make sure this works and it appears to be fine.

mayank99 (approved):

mayank99 (approved):

src/style/global.scss
```diff
@@ -2,6 +2,38 @@
...
+
+  @media (prefers-color-scheme: light) and (prefers-contrast: more) {
+    @include theme-variables(light-hc);
+  }
+
+  @media (prefers-color-scheme: dark) and (prefers-contrast: more) {
+    @include theme-variables(dark-hc);
+  }
+
+  &.iui-theme-light {
```
elephantcatdog, line 20: Do you not need the extra & that was previously included with &.iui-theme-light? Looks like it was part of the fallback?

elephantcatdog (changes requested):

src/style/global.scss
FlyersPh9: I believe line 6 makes it so we don't need the extra `&`.

This:
```scss
html {
	@include theme-variables(light);

	&.iui-theme-light {
		@include theme-variables(light);
	}
}
```

Should have the same output as this:
```scss
html {
	&,
	&.iui-theme-light {
		@include theme-variables(light);
	}
}
```

The second option has less text, but I think the first option needs to be placed in this order so it runs through the media queries first & then if the class is applied, it overrides whatever was set by the media query.

FlyersPh9:
> I don't know how well this would work, but you could add an overlay with box-shadow so that user icons are shaded with white for dark theme and black for light theme, and the text color should be black for dark theme (light icons) and white for light theme (dark icons). This does mess with the box-shadows already present though. I don't like the way this looks exactly - something is off with the text color, and I didn't get it working for the stacked lists, but something like this:
1st one has `box-shadow: inset 0 0 0 200px rgb(255 255 255 / 50%); color: black;` and 2nd one has `box-shadow: inset 0 0 0 200px rgb(0 0 0 / 60%);` in `iui-initials`. But with this method you have to make assumptions about how light/dark the color is initially.

That's an interesting approach.  I vote that we let this PR pass as is and tackle both user icons and badge AAA scoring issues in separate PRs while working towards CSS 1.0.

elephantcatdog (approved):
Looks good to me

src/carousel/carousel.scss
FlyersPh9: If you are planning to come into the office tomorrow, let's figure it out in person.

PR #507, opened 2022-03-03
feat(Carousel): Use scroll-snap instead of transform

mayank99:
Instead of trying to dynamically calculate the `translateX` value in JS, I'm now using scroll snapping with a hidden scrollbar. The transition is handled using `smooth` scrolling.

Scroll snapping provides tons of benefits:
- free support for dragging on touch devices
- automatically scroll to the focused slide (no need to disable focusing for hidden slides)
- allow showing multiple slides (not possible previously because `translateX` was relying on % width)

From JS part, scroll can be triggered by simply calling `scrollIntoView`, but for synchronizing the auto-scroll cases mentioned above, I think we would need IntersectionObserver or something. 🤔

### DIFF
--- a/src/carousel/carousel.scss
+++ b/src/carousel/carousel.scss
@@ -12,17 +12,29 @@
 
 @mixin iui-carousel-slider {
   display: flex;
+  gap: $iui-s;
   list-style: none;
   margin: 0;
   padding: 0;
+  overflow-x: auto;
+  overflow-x: overlay;
+  scroll-snap-type: x mandatory;
   @media (prefers-reduced-motion: no-preference) {
-    transition: transform $iui-speed-slow ease-in-out;
+    scroll-behavior: smooth;
   }
+
+  // hide scrollbar
+  &::-webkit-scrollbar {
+    background-color: transparent;
+    height: 0;
+  }
+  scrollbar-width: none;
 
   &-item {
     width: 100%;
     flex-shrink: 0;
     box-sizing: border-box;
+    scroll-snap-align: center;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/carousel/carousel.scss
```diff
@@ -12,17 +12,29 @@
 
 @mixin iui-carousel-slider {
   display: flex;
+  gap: $iui-s;
```
FlyersPh9, line 15: What's the reason for the gap between slides?

FlyersPh9:
Cool to see scroll returning, I know that we tried that originally but ran into issues or concerns on how we could make it work.  Especially like swiping on touch.

src/carousel/carousel.scss
veekeys: Yeah, question ,if we want to show empty space between every slide.. It is visible when sliding...

src/carousel/carousel.scss
```diff
@@ -12,17 +12,29 @@
...
   list-style: none;
   margin: 0;
   padding: 0;
+  overflow-x: auto;
+  overflow-x: overlay;
+  scroll-snap-type: x mandatory;
```
veekeys, line 21: So does it do anything? I barely could notice a difference with and without it,,

veekeys: a ok... mobile devices..
Need to read description twice..

veekeys (approved):

src/carousel/carousel.scss
mayank99: it's for this third case mentioned in description
> allow showing multiple slides

image

FlyersPh9: Ahhh, ok, I understand now.  That will be a nice addition.

FlyersPh9 (approved):

PR #508, opened 2022-03-04
chore(Theme): Convert HEX to HSL

FlyersPh9:
- Convert all HEX to HSL.
- Removed unused variables that were for our old `:focus` styling.

### DIFF
--- a/src/style/theme.scss
+++ b/src/style/theme.scss
@@ -45,12 +45,6 @@ $iui-text-color--light: rgba($iui-color-foreground-body--light, $iui-opacity-2--
 $iui-text-color-muted--light: rgba($iui-color-foreground-body--light, $iui-opacity-4--light);
 $iui-text-color-placeholder--light: rgba($iui-color-foreground-body--light, $iui-opacity-5--light);
 
-// Light theme > Focus color
-$iui-focus-box-shadow--light: rgba($iui-color-foreground-primary--light, $iui-opacity-5--light) 0 0 0 2px;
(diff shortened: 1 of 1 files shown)


src/style/theme.scss
```diff
@@ -3,25 +3,25 @@
...
-$iui-color-background-1--light: #FFF;
-$iui-color-background-2--light: #F8F9FB;
-$iui-color-background-3--light: #EEF0F3;
-$iui-color-background-4--light: #DCE0E3;
-$iui-color-background-5--light: #C7CCD1;
-$iui-color-background-primary--light: #008BE1;
-$iui-color-background-positive--light: #53A21A;
-$iui-color-background-negative--light: #D30A0A;
-$iui-color-background-warning--light: #F18B12;
+$iui-color-background-1--light: hsl(0, 0%, 100%);
```
mayank99, line 6: I was going to suggest using new space-separated `hsl` syntax, e.g. `hsl(0 0% 100%)` but node-sass doesn't seem to understand that.

@bentleyvk What's the plan with the dart sass work you had started?

bentleyvk: When I have free time, I hope I will finish it 😀 But is it a big difference whether to use spaces or commas?

mayank99: The newer space-separated syntax is preferable because it allows you to use the same `hsl` function (instead of `hsla`) for adding transparency. i.e. `hsl(0 0% 100% / 0.5)` vs `hsla(0, 0%, 100%, 0.5)`.

I don't like that node-sass stops us from using new features all the time.

gretanausedaite (approved):
Approve because dart sass implementation should not block this PR. We'll change to space-separated syntax when we can.

PR #513, opened 2022-03-09
feat(ColumnManager): CSS added to make Column Manager a Dropdown Button

adamhock:
CSS has been added so that Column Manager can exist as a type of Dropdown Button

- Changed the `width `of `iui-menu-item` to `auto` so that the rows in the Dropdown menu are clickable
- Updated html files to show Column Manager as a Dropdown Button
- Because some of the CSS for the menu for Dropdown Buttons is generated by Tippy.js, I decided not to put a test example of the Column Manager when it is open. Instead, the following screenshot shows what the menu would look like

Column Manager Menu


veekeys (commented):
could you still add this to html? simple on click handler to toggle display of menu. hardcode position straight in html..
now its hard to review  :DD

adamhock:
> could you still add this to html? simple on click handler to toggle display of menu. hardcode position straight in html.. now its hard to review :DD

Does it matter if the hardcore position is directly under the button, or is this toggle only being used for reviewing the CSS in this PR?

mayank99:
> Does it matter if the hardcore position is directly under the button, or is this toggle only being used for reviewing the CSS in this PR?

It doesn't really matter how you do it, we just need something to be able to demo it. If it's being used in multiple table instances in table.html, then you might need to do some calculations. However, I'm thinking we might also need visual tests for it.

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
             </div>
             <div class="iui-cell iui-slot">
               <button
-                class="iui-button iui-borderless"
+                class="iui-button iui-borderless iui-dropdown"
```
mayank99, line 276: You don't need this `iui-dropdown` class, it's only used for adjusting the spacing when the caret icon is used.

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
...
+                onclick="openColumnManger()"
               >
                 <svg-column-manager
                   class="iui-button-icon"
                   aria-hidden="true"
                 ></svg-column-manager>
               </button>
+              <div
+                data-tippy-root=""
+                id="tippy-1"
```
mayank99, line 288: It doesn't make sense to emulate the tippy dom structure if you are not using tippy.js here. You can cleanup this whole thing to just use the `ul`.

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
...
+                style="z-index: 9999; display: none; position: fixed; inset: 0px auto auto 0px; margin: 0px; transform: translate(625px, 62px);"
```
mayank99, line 289: Positioning is messed up. You _can_ hardcode the numbers but you should base it relative to the trigger button rather than the screen. These magic numbers will never work consistently across different screens.

Also, are you planning to add the column manager to other table instances on this page? Because then you might want to calculate the position using JS.

backstop/tests/table.html
```diff
@@ -3798,5 +3849,16 @@ <h2>Empty state</h2>
         });
       });
     </script>
+
+    <script>
+      var elements = document.getElementsByClassName('iui-dropdown');
```
mayank99, line 3854: Please don't use `var` anywhere ever for any reason at all.
https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
...
+                  >
+                    <div>
+                      <ul
+                        class="iui-menu"
+                        role="menu"
+                        style="min-width: 58px;"
+                      ><label
+                          class="iui-checkbox-wrapper iui-menu-item"
+                          style="cursor: pointer;"
+                        ><input
```
mayank99, line 310: Why are you not using our `iui-checkbox` component?

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
...
+                            checked=""
+                          ><span class="iui-checkbox-label">Description</span></label><label
+                          class="iui-checkbox-wrapper iui-menu-item"
+                          style="cursor: pointer;"
+                        ><input
+                            type="checkbox"
+                            title="Toggle Column Visible"
+                            role="menu"
+                            checked=""
+                          ><span class="iui-checkbox-label">Click</span></label></ul>
```
mayank99, line 331: Formatting is all messed up, please fix

src/menu/menu.scss
```diff
@@ -47,7 +47,7 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
   cursor: pointer;
   box-sizing: border-box;
   line-height: normal;
-
+  width: auto;
```
mayank99, line 50: I don't think you really need this width, it should come from the flexbox if you use `iui-content`.

Curious how you ended up with this solution. If you try using our `MenuItem` react component, it will not let you use it without wrapping the children with `iui-content`.

mayank99 (commented):
Not sure what this PR is trying to achieve.

@adamhock Do you have a branch or example code of how you plan to use this in your React implementation?

backstop/tests/table.html
veekeys: I dont see you use this variable anywhere too

backstop/tests/table.html
```diff
@@ -3798,5 +3849,16 @@ <h2>Empty state</h2>
...
       });
     </script>
+
+    <script>
+      var elements = document.getElementsByClassName('iui-dropdown');
+
+      function openColumnManger() {
+        var element = document.getElementById('tippy-1');
+        element.style.display = element.style.display == "none" ? "block" : "none";
+        document.getelement;
```
veekeys, line 3859: whats that?

backstop/tests/table.html
veekeys: Yeah, you can clean whole html up by removing hardcoded inline styles.. I thinnk you copied this whole thing from wrong place. Check our checkbox html

backstop/tests/table.html
veekeys, line 276: Instead, you can add custom class and in this html set `position: relative` (custom class will help you to select those elements too).. This should help you using `position: absolute` put the menu in the correct place.

backstop/tests/table.html
```diff
@@ -273,15 +273,66 @@ <h2>Default</h2>
...
+                    class="tippy-content"
+                    data-state="hidden"
+                    style="transition-duration: 0ms;"
+                  >
+                    <div>
+                      <ul
+                        class="iui-menu"
+                        role="menu"
+                        style="min-width: 58px;"
+                      ><label
```
veekeys, line 307: if there is `ul`, its children semantically should be `li`. Check up our menu html file... You might need `iui-menu-content` class for this as Mayank has mentioned in other comment..

backstop/tests/table.html
adamhock: tippy structure has been removed

backstop/tests/table.html
adamhock: use of `var` has been removed

backstop/tests/table.html
adamhock: formatting has been fixed

src/menu/menu.scss
adamhock: this has been removed

backstop/tests/table.html
adamhock: this menu has been updated

adamhock:
Closed because no CSS changes are needed for Column Manager

PR #517, opened 2022-03-09
chore(Styles): Add text selection mixin

FlyersPh9:
These components use `::selection`:
- Alert
- Labeled inputs
- Progress indicator
- Table
- Toast notification

This PR adds `@mixin iui-text-selection;` so the opacity is always in sync across components.

### DIFF
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -63,6 +63,15 @@
   }
 }
 
+@mixin iui-text-selection($status: primary) {
+  &::selection,
+  *::selection {
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-#{$status}-rgb), t(iui-opacity-5));
+    }
+  }
+}
+
 @mixin iui-scrollbar {
   @include themed {
     scrollbar-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4)) transparent;
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -67,7 +67,7 @@
   &::selection,
   *::selection {
     @include themed {
-      background-color: rgba(t(iui-color-foreground-#{$status}-rgb), t(iui-opacity-5));
+      background-color: rgba(t(iui-color-foreground-#{$status}-rgb), t(iui-opacity-4));
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/style/mixins.scss
```diff
@@ -63,6 +63,15 @@
   }
 }
 
+@mixin iui-text-selection($status: primary) {
+  &::selection,
```
veekeys, line 67: Whats the difference between these 2 selectors?

veekeys (approved):
Apart the facts that PR came from nowhere and the problem stated in description does not exist (all values were the same, so it shouldnt have been inconsistent); opacity was slightly changed and we dont have visual tests to test all of that (I guess too minor); PR looks ok. 

PR title could not consist `feat`, because I dont think this change requires minor version release...

src/style/mixins.scss
mayank99: `&` will apply to the same level where the mixin is used, `*` will apply to all descendants in the tree

src/style/mixins.scss
```diff
@@ -63,6 +63,15 @@
...
+  *::selection {
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-#{$status}-rgb), t(iui-opacity-5));
```
mayank99, line 70: Shouldn't this be `opacity-4` like in every other place?

FlyersPh9:
> Apart the facts that PR came from nowhere

This came from the work taking place in #505.

> and the problem stated in description does not exist (all values were the same, so it shouldnt have been inconsistent)

I changed the description after I submitted because I realized what had happened.  I changed the opacity in `global.scss` but did not change it in all the other places.  It took me a while before I figured out what had happened.

> opacity was slightly changed

I've reverted it back to 4 for this PR but I will change it to 5 in #505.

> PR looks ok.

Thank you. ☺️ 

> PR title could not consist `feat`, because I dont think this change requires minor version release...

Changed to `chore` after reverting opacity back to 4.

src/style/mixins.scss
mayank99: Maybe make opacity a parameter in the mixin then

mayank99: or well, lets see what we need in #505 and change accordingly

mayank99 (approved):

PR #518, opened 2022-03-10
fix(Table): Fixed column reorder styling on resizable column

bentleyvk:
Before resize handle was visible:
image

After:
image

`:where` prevents from increasing specificity which is still bad `(0,5,0)` 😀


src/table/table.scss
```diff
@@ -85,7 +85,7 @@
       }
     }
 
-    &:hover > .iui-resizer {
+    &:hover:where(:not(.iui-reorder-column-left, .iui-reorder-column-right)) > .iui-resizer {
       opacity: 1;
```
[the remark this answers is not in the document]
bentleyvk: Changed to having a separate bar element like resizer is done. This way I don't encounter this resizer and reorder indicator collapse issue also fixes content shifting when hovering over column.
P.S. You can now drag first table's columns because I couldn't work without actual working example 😀 Of course it won't change the order but at least we can test drag interactions.

backstop/tests/table.html
```diff
@@ -3798,5 +3822,32 @@ <h2>Empty state</h2>
...
+      const onDragOver = (event) => {
+        event.preventDefault();
+        const column = event.currentTarget;
+        const columnIndex = Array.from(column.parentElement.childNodes).indexOf(column);
+        if (columnIndex === reorderStartIndex) {
+          return;
+        }
+        column.classList.add(columnIndex > reorderStartIndex ? 'iui-reorder-column-right' : 'iui-reorder-column-left');
+      };
+      const onDragLeave = (event) => {
```
adamhock, line 3841: Can't this class and `onDrop` be combined into one function?

bentleyvk: You mean to have one function for `onDragLeave` and `onDrop`?

mayank99 (approved):

backstop/tests/table.html
bentleyvk: Reusing.

PR #525, opened 2022-03-11
feat(Anchor): Add anchor component

FlyersPh9:
- Add **Anchor** component.
- Includes `.iui-anchor-external` modifier.
- Replaces custom anchor styling in toast & alert components.
- mixin supports theming.
- Support for `forced-colors`.

Screen Shot 2022-03-28 at 3 06 08 PM


src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-anchor($status: primary, $underline: 'false') {
+  --_iui-anchor-popout-svg: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z' /></svg>");
+
+  @include iui-anchor-status($status);
+  border-radius: $iui-border-radius;
+  text-underline-offset: 1px;
```
FlyersPh9, line 10: Most image changes are a result of this line of CSS.  This makes the underline align with the bottom of the [descender](https://lh3.googleusercontent.com/XN6ATtQ7Q_NxH5Eb_ZMiAZI6Wh61nClqWKHyx10vhIly8lzEOKmswVEAxjQOzldyCJ_q5W6dTznM0fn5wS8ArPzhpDq4vfhUxnOz=w1064-v0).

Previously:
Screen Shot 2022-03-28 at 4 34 56 PM

Now:
Screen Shot 2022-03-28 at 4 34 39 PM

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+      text-decoration: none;
+    }
+  }
+
+  &-blank::after {
+    content: '';
+    display: inline-block;
+    width: 1.5ch;
+    height: 1.5ch;
+    margin-left: 0.5ch;
```
FlyersPh9, line 31: Using `CH` so that the icon size & spacing adjusts with the font size set.
Screen Shot 2022-03-28 at 4 42 01 PM

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-anchor($status: primary, $underline: 'false') {
+  --_iui-anchor-popout-svg: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z' /></svg>");
+
+  @include iui-anchor-status($status);
+  border-radius: $iui-border-radius;
+  text-underline-offset: 1px;
+  forced-color-adjust: none;
```
FlyersPh9, line 11: Forces colors to not be overriden since we need to reserve the color in some cases for status.
Screen Shot 2022-03-28 at 4 44 03 PM

Without it we get this and the icon disappears:
Screen Shot 2022-03-28 at 4 45 22 PM

src/style/mixins.scss
```diff
@@ -32,10 +32,12 @@
   }
 }
 
-@mixin iui-focus($color: var(--iui-color-foreground-primary), $offset: -1px, $thickness: 1px) {
+@mixin iui-focus($status: primary, $offset: -1px, $thickness: 1px) {
```
mayank99, line 35: This change is unrelated, please undo from this PR

src/classes.scss
```diff
@@ -43,3 +43,4 @@
 @import './wizard/classes';
 @import './skip-to-content/classes';
 @import './surface/classes';
+@import './anchor/classes';
```
mayank99, line 46: nit: I think this file needs to be sorted alphabetically

src/anchor/anchor.scss
mayank99, line 11: This change is anti-accessibility as the colors are no longer "forced high contrast".

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+    &:hover {
+      text-decoration: underline;
+    }
+  } @else {
+    &:hover {
+      text-decoration: none;
+    }
+  }
+
+  &-blank::after {
```
mayank99, line 26: Not sure about this name. I know "blank" is supposed to mimic `target=_blank` but something like `-external` might be more obvious.

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+  &:hover {
+    @include themed {
+      color: t(iui-color-foreground-#{$status}-overlay);
+    }
+  }
+
+  &.iui-anchor-blank {
+    &::after {
+      @include themed {
+        background-color: currentColor;
```
mayank99, line 54: This doesn't need to be wrapped in themed

src/anchor/classes.scss
```diff
@@ -0,0 +1,19 @@
...
+    @include iui-anchor-status($status: positive);
+  }
+
+  &-warning {
+    @include iui-anchor-status($status: warning);
+  }
+
+  &-negative {
+    @include iui-anchor-status($status: negative);
+  }
```
mayank99, line 18: Not sure if we want to expose these to the users. It's one thing to make mixins for internal re-use, but if we provide these options then we might start seeing misuse

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+@mixin iui-anchor($status: primary, $underline: 'false') {
+  --_iui-anchor-popout-svg: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path d='m16 0v5.4l-1.9-2-8.4 8.4-1.5-1.5 8.3-8.4-1.9-1.9m5.4 16v-9h-1v8h-14v-14h8v-1h-9v16z' /></svg>");
+
+  @include iui-anchor-status($status);
+  border-radius: $iui-border-radius;
+  text-underline-offset: 1px;
+  forced-color-adjust: none;
+  cursor: pointer;
+
+  @if $underline == 'false' {
```
mayank99, line 14: So how do you plan to tackle the underline for high contrast (which was one of the original main purposes for this PR)? This condition might get complicated.

backstop/tests/html-tags.html
```diff
@@ -1,110 +0,0 @@
-<!--
```
mayank99, line 1: Why'd this get deleted?

src/anchor/anchor.scss
```diff
@@ -0,0 +1,58 @@
...
+    &:hover {
+      text-decoration: none;
+    }
+  }
+
+  &-blank::after {
+    content: '';
+    display: inline-block;
+    width: 1.5ch;
+    height: 1.5ch;
```
mayank99, line 30: This icon does not look vertically centered to me.

backstop/scenarios/anchor.js
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
+  scenario('default', {
+    selectors: ['#demo-default'],
+  }),
```
mayank99, line 10: Don't you want to test hover/focus states?

src/anchor/anchor.scss
FlyersPh9: What if I remove just this one line & we review it.  Then once everything is good, I add that line back in.  That way we know any additional image changes are only the underline change.

src/anchor/anchor.scss
FlyersPh9: I'm searching right now to see if there are provided color variables for forced high contrast mode.

The yellow hyperlink color is `hsl(60deg 100% 50%)`.  What if I find a green, orange, red color that match the S & L of the pre-defined yellow?

mayank99: or just remove this line?

the "yellow" comes from Windows settings and can be changed there. That's the whole point of "forced colors" and we shouldn't mess with it except in edge cases like ColorPicker.

These status links are meant to be used in Toast/Anchor which should already have some other indication of the status (i.e. icon), so it's completely fine if the color is lost imo.

backstop/tests/html-tags.html
FlyersPh9: HTML tags was covering `<a>`, `<abbr>`, and ``.  I moved all of the `<a>` out of that demo into the new anchor demo page.  `<abbr>` wasn't receiving any custom styling and the `@2x`, `@3x` part of the `` was removed a while back, so HTML tags wasn't serving much purpose anymore.

src/anchor/anchor.scss
FlyersPh9: I still need to figure out how to handle the appended icon.  Here's the options I can think of:

1. `@media (forced-colors: active) { display: none; }`
Screen Shot 2022-03-29 at 10 38 43 AM

2. `@media (forced-colors: active) { forced-color-adjust: none; }`
Screen Shot 2022-03-29 at 10 39 26 AM

3. `@media (forced-colors: active) { forced-color-adjust: none; background-color: hsl(60deg, 100, 50); }` but if the user changes their hyperlink color, this will not look great.
Screen Shot 2022-03-29 at 10 40 31 AM

Removing `forced-color-adjust: none;` completely results in this which I think we should try to avoid:

mayank99: try `@media (forced-colors: active) { background-color: LinkText }` in the pseudo-element

mayank99: in the CSS spec https://www.w3.org/TR/css-color-4/#css-system-colors

src/anchor/anchor.scss
FlyersPh9: The height of the `<a>` appears to be 19px & the icon is 12px.  I played around with adding a `margin-top` & `padding-top` to `::after` but wasn't happy with the results yet.  It is using `display: flex` & `align-items: center` so it is trying to center it.

Screen Shot 2022-03-29 at 12 30 09 PM

src/anchor/anchor.scss
FlyersPh9: This has been fixed.

src/anchor/anchor.scss
FlyersPh9: `-external` sounds good to me.

src/classes.scss
FlyersPh9: This has been fixed.

src/style/mixins.scss
FlyersPh9: Reverted.

src/anchor/anchor.scss
mayank99: maybe use baseline?

src/anchor/anchor.scss
FlyersPh9: I've removed it for now but would like to add it back in at the end.

src/anchor/classes.scss
FlyersPh9: Removed the status classes, we can test for those statuses within Alert and Toast.

src/anchor/anchor.scss
FlyersPh9: Thank you for the help on the call earlier, we now have underline working correctly.

src/anchor/anchor.scss
elephantcatdog: You can use values like `1px` with vertical-align. Perhaps `vertical-align: -0.1ch;` is what you're looking for?

src/anchor/anchor.scss
```diff
@@ -0,0 +1,77 @@
...
+
+  @at-root {
+    .iui-theme-light-hc &,
+    .iui-theme-dark-hc & {
+      @include iui-anchor-underline('on-default');
+    }
+  }
+}
+
+@mixin iui-anchor-status($status: primary) {
```
elephantcatdog, line 46: If possible, should we include the different uses of `iui-anchor-status` and `iui-anchor-underline` within the anchor html file?

elephantcatdog (commented):
image
image
It appears that nothing has changed for these test images.

src/anchor/anchor.scss
FlyersPh9: Originally I did have `.iui-positive`, `.iui-warning`, & `.iui-negative` modifier classes so our users could apply any of these status colors.  We removed those classes so they are not exposed to users to try to avoid misuse / abuse of those.  To achieve those status links, you would need to use `@mixin iui-anchor-status`.

That's a long way of explaining that I removed the status links from the anchor demo page because they're already being tested within toast notifications & alerts demo pages.

src/anchor/anchor.scss
FlyersPh9: `vertical-align: -0.1ch;` does seem to do the trick!  Tested on different font sizes and visually looks good.

src/anchor/anchor.scss
elephantcatdog: Gotcha, that makes sense.

bentleyvk:
>  It appears that nothing has changed for these test images.

pixelmatch doesn't show color changes, I need to investigate that more.

FlyersPh9:
> > It appears that nothing has changed for these test images.
> 
> pixelmatch doesn't show color changes in some cases, I need to investigate that more.

@bentleyvk @elephantcatdog I downloaded the original image & uploaded it, no changes detected and the changed file count dropped a little.

backstop/tests/anchor.html
```diff
@@ -0,0 +1,59 @@
...
+      href="../assets/demo.css"
+    />
+    <script src="../assets/theme.js"></script>
+  </head>
+  <body class="iui-body">
+    <theme-button></theme-button>
+    <h1>Anchor</h1>
+    <hr />
+
+    <section id="demo-default">
```
bentleyvk, line 32: Add padding so hover state would be fully visible in the imgs.

src/style/global.scss
```diff
@@ -51,10 +51,6 @@ html {
   }
 }
 
-.iui-anchor {
```
bentleyvk, line 54: So if users doesn't import `anchor.css` and they previously used `iui-anchor` class from global css, their anchors will break. But maybe that's ok with pre major release of css...

src/anchor/anchor.scss
```diff
@@ -0,0 +1,79 @@
...
+
+  &:hover {
+    @include themed {
+      color: t(iui-color-foreground-#{$status}-overlay);
+    }
+  }
+}
+
+/// $behavior should be one of: 'on-hover' or 'on-default'
+@mixin iui-anchor-underline($behavior: 'on-hover') {
```
bentleyvk, line 65: For me this mixin doesn't feel right. I would just override stylings in toast as it is the only place where `on-default` (strange name) is needed.

bentleyvk: Oh, it is also used in high-contrast themes. Maybe than need to have a better naming for `on-default`

mayank99 (approved):

backstop/tests/anchor.html
FlyersPh9: Added padding so test images include the `:focus` outline.

src/style/global.scss
FlyersPh9: There would be no change for those who use `all.css` though, I think that's a plus.

src/anchor/anchor.scss
FlyersPh9: How does `on-initial` sound?

mayank99: `on-initial` sounds better to me. 

Another option is: `show-on-hover` and `hide-on-hover`

src/anchor/anchor.scss
elephantcatdog: Great!

bentleyvk (approved):

PR #527, opened 2022-03-15
fix(Carousel): Increase flex-grow for dots

### LINKED ISSUE #516: Carousel: dots get clipped on narrow widths
When the carousel is ~350px wide, the dots start getting clipped if there are 5 or more.
image

This should not be happening because there is definitely space available around the dots.
image

mayank99:
Increasing flex grow allows the dots container to fill up more space on narrow widths. Also added max-width to prevent taking up too much space on larger widths.

Fixes #516 

Before:
image

After:
image


bentleyvk:
I would say to make as much visible dots as it fits because now it looks strange when there is enough space and we are still showing the same amount of dots. If you afraid that tests images will change, I don't see it as an issue.

veekeys:
> I would say to make as much visible dots as it fits because now it looks strange when there is enough space and we are still showing the same amount of dots. If you afraid that tests images will change, I don't see it as an issue.

So you mean putting like 10 dots in html to test the overflow too? Cause we still need to apply correct classes at the first and last ones and that could happen only with additional logic in react, from what I understood..

bentleyvk:
> > I would say to make as much visible dots as it fits because now it looks strange when there is enough space and we are still showing the same amount of dots. If you afraid that tests images will change, I don't see it as an issue.
> 
> So you mean putting like 10 dots in html to test the overflow too? Cause we still need to apply correct classes at the first and last ones and that could happen only with additional logic in react, from what I understood..

No no, not to test overflow. But you can see that there is enough space to fit like 6 additional dots. I just want us to use that additional space.
image

veekeys:
> > > I would say to make as much visible dots as it fits because now it looks strange when there is enough space and we are still showing the same amount of dots. If you afraid that tests images will change, I don't see it as an issue.
> > 
> > 
> > So you mean putting like 10 dots in html to test the overflow too? Cause we still need to apply correct classes at the first and last ones and that could happen only with additional logic in react, from what I understood..
> 
> No no, not to test overflow. But you can see that there is enough space to fit like 6 additional dots. I just want us to use that additional space. image

Well yeah, I guess we can.. But now I wonder, that we should also put `max-width` cause on pretty wide carousel it might be just too much...

src/carousel/carousel.scss
```diff
@@ -64,6 +63,9 @@
   &-dots {
     @include iui-reset;
     @include iui-focus;
+    display: flex;
+    align-items: center;
+    flex: 2;
```
veekeys, line 68: Would it make sense to add some max-width for too wide carousels..?

mayank99: Added `max-width: 256px`

mayank99:
> > I would say to make as much visible dots as it fits because now it looks strange when there is enough space and we are still showing the same amount of dots. 
...
> Well yeah, I guess we can.. But now I wonder, that we should also put `max-width` cause on pretty wide carousel it might be just too much

I've increased `MAX_DOTS` in the html code to 7, and I've set `max-width: 256px` on the dots container.

src/carousel/carousel.scss
veekeys: good, but why 256? 1 dot is 28px and dots have no gap in between... Doesnt seem to divide nicely..?

mayank99: i just went with our size variables. Think I should use `28 * 9` or something?

bentleyvk:
> I've increased `MAX_DOTS` in the html code to 7, and I've set `max-width: 256px` on the dots container.
 
No need to update imgs? 😀

src/carousel/carousel.scss
veekeys: yeah idk.. our buttons sizes are really weird :/

mayank99:
> No need to update imgs? 😀

Pushed

veekeys (approved):

elephantcatdog (commented):
Is it a concern that if you add more dots to the bottom, they still clip? I added three more to the html file and it looks like this:
image
But I don't know a lot about how carousel works.

mayank99:
> Is it a concern that if you add more dots to the bottom, they still clip? I added three more to the html file and it looks like this: image

If you read description and comment history, it is done that way intentionally because we don't want to show that many dots.

If the dots are used outside carousel, then it is easy to override max-width.

elephantcatdog (approved):

PR #528, opened 2022-03-16
fix(Table): Fix for resize calculations in react

bentleyvk:
This is the fix for https://github.com/iTwin/iTwinUI-react/issues/581. Because body rows have borders and header row doesn't, that causes issues in resize calculations and makes to show horizontal scrollbar after resizing.
Also, this makes header and body columns alignment perfect.


mayank99 (approved):

elephantcatdog (approved):
Seems fine to me

PR #529, opened 2022-03-16
feat(theme): Change colors across multiple components to pass contrast better

FlyersPh9:
- Introducing `$iui-color-background-border` which displays background-5 in default themes but shows foreground-body w/ opacity in high contrast themes.
- Skip-to-content button use an inverted color palette to always guarantee better contrast above the background.
- Tooltip has a white border to help stand out above any background color.
- Selection overlay opacity is lowered.
- Add inner white circle to active carousel dot.
- Muted text over any overlay background color has had opacity removed.  This is to score a AAA in active header button & selected menu item.
- Header secondary text is using smaller font to align with menu items.
- Active menu item uses blue text to align with header, side nav.


FlyersPh9:
I haven't figured out a good way of handling tooltip / skip-to-content other than the invert effect.

src/header/header.scss
```diff
@@ -83,7 +83,7 @@
     width: 1px;
     height: $iui-baseline * 2;
     @include themed {
-      background-color: t(iui-color-background-4);
+      background-color: t(iui-color-background-border);
```
veekeys, line 86: background color as border color.. does not make sense

src/tooltip/tooltip.scss
```diff
@@ -18,8 +19,9 @@
   box-shadow: $iui-elevation-8;
   pointer-events: none;
   @include themed {
-    background-color: rgba(0, 0, 0, t(iui-opacity-3));
-    color: t(iui-color-foreground-accessory);
+    // background-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-3));
```
veekeys, line 22: remove commented out rules

veekeys (commented):
Sure we cannot change things only for high contrast themes???
This influence old themes all over the place and results in 150+ files changes.
I have no idea, how to review that all, where mostly images are changed.
Do you have a recommendation on the tool to use for it?

backstop/assets/demo.css
```diff
@@ -34,7 +34,7 @@ h3 {
 }
 
 hr {
-  border-top: 1px solid var(--iui-color-background-5);
+  border-top: 1px solid var(--iui-color-background-border);
```
bentleyvk, line 37: Undo this, because it can cause a lot tests images to change which makes even harder to review.

mayank99: I don't think it's because of that. I was swiping through the images and didn't see any difference between left and right.

Maybe try undoing all image changes (i.e. get them to the state before this PR) and re-running all tests and then approving again. It's possible you'll only get a few changes then.

src/header/header.scss
FlyersPh9: In this case it does kind of make sense.  This element is `.iui-divider` which is basically replicating a border.  @mayank99 and I both found an article that explained why it's better to use a 1px wide element VS a `border` when inside a flexbox.  I can't find that article atm.

veekeys: oh, yes.. you are right.
Other same cases are track and rail, which I assume also makes sense..

backstop/assets/demo.css
FlyersPh9: Count is now lower than before, but it's still a little high.  Most changes are:
- Background 4 being replaced with `iui-color-background-border` which is equal to 5 in base themes.
- Selection state opacity is less than before.
- Swapping for foreground-body in some places.
- Tooltip / skip to content color is inverting from the theme.

src/carousel/carousel.scss
```diff
@@ -114,9 +118,13 @@
       transform: scale(0);
     }
 
-    &.iui-active::after {
-      @include themed {
-        background-color: t(iui-color-foreground-primary);
+    &.iui-active {
+      &::after {
```
veekeys, line 122: why syntax changed here?

backstop/assets/demo.css
veekeys: Again, cant we make changes so they apply only to high-contrast themes? I thought we were fixing issues there now...
Current images do not make sense to me, at least some of them:
- Header image shows header button descripton color change, but image captures that only in one of headers (notice iModel Beta button description color) 
image
- Select arrow somehow changes size (seems like) 
before: image
And now its color is even inconsistent with menu items above, for example
image
- Not sure we can do this drastic tooltip change 
image
- This tooltip got bigger (according to new image)
image
- Is it expected that light theme images have more changes than dark (Tile border seems to change in light, but not in dark, for example)

And there might be many more.. So again, not a very nice PR to review and figure things out here :/

src/tooltip/tooltip.scss
FlyersPh9: This has been cleaned up.

backstop/assets/demo.css
FlyersPh9: The tooltip inverting has been reverted, but it does still bring _some_ change that will effect all themes.  Only change is adding a `foreground-accessory` border to help stand out against any color background.
Screen Shot 2022-03-21 at 2 42 31 PM

FlyersPh9: Another change I brought was removing opacity from the selected secondary line of text.  This was needed to score a AAA and will help with the contrast on the base themes as well.
Screen Shot 2022-03-21 at 2 44 19 PM

This change was also brought to menu items to match:
Screen Shot 2022-03-21 at 2 45 42 PM

src/carousel/carousel.scss
FlyersPh9: I tried a couple of different ways of updating the active indicator.  I think this change is leftover from a failed attempt.  Reverting.

backstop/assets/demo.css
FlyersPh9: > * Select arrow somehow changes size (seems like)
>   before: image
>   And now its color is even inconsistent with menu items above, for example
>   image

I'm not seeing this change.  That first example looks like the dropdown arrow Titas implemented.  Where did you take those screenshots from?

veekeys: image

It is from your PR, where you think I am taking these from...

PR #530, opened 2022-03-16
feat(DatePicker): Added range date support

veekeys:
Now can mark a range of days in date picker!
- now using small buttons in date-picker header to save some space and look nicer
- fixed horizontal movement of numbers on hover
- removed some unused css rules
- changed current day border radius, so its rectangular now as day size changed and its no longer a square
- naming and colors are up for debate
- date picker height increased because I added little margin between rows
- date picker width increased and its same no matter if year buttons are there
- date picker does not have fixed size set as it is now coming from content. As a result, there is grid magic to adjust time picker height

image


FlyersPh9:
In the date span picker, the start date & end date have different cursor types on hover.  I think they should be matching.

Also I tried seeing what it would look like if i placed the Current Date within the span & aside from the font weight change it's hardly noticeable.  Date 20 is the current day in this screenshot:
Screen Shot 2022-03-17 at 5 59 46 PM

veekeys:
> In the date span picker, the start date & end date have different cursor types on hover. I think they should be matching.
> 
> Also I tried seeing what it would look like if i placed the Current Date within the span & aside from the font weight change it's hardly noticeable. Date 20 is the current day in this screenshot: Screen Shot 2022-03-17 at 5 59 46 PM

I have fixed to cursor (changed to pointer), but I am not sure how to fix the today marker..
Should we change the approach? i.e. instead of background-color add some custom marker like
image

Or if you have any suggestion on the color to change to, I'd appreciate :)

adamhock:
> I have fixed to cursor (changed to pointer), but I am not sure how to fix the today marker.. Should we change the approach? i.e. instead of background-color add some custom marker like 
> image
> 
> Or if you have any suggestion on the color to change to, I'd appreciate :)

I personally like the marker. Is the legend necessary?

veekeys:
> > I have fixed to cursor (changed to pointer), but I am not sure how to fix the today marker.. Should we change the approach? i.e. instead of background-color add some custom marker like
> > image
> > Or if you have any suggestion on the color to change to, I'd appreciate :)
> 
> I personally like the marker. Is the legend necessary?

It is just a screenshot from other example, it is not our date picker :D

src/date-picker/date-picker.scss
```diff
@@ -111,6 +112,35 @@
     @include iui-focus($color: var(--iui-color-foreground-accessory), $offset: -3px);
   }
 
+  &.iui-range-start {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    cursor: pointer;
```
adamhock, line 118: `cursor: pointer` needs to be added here

src/date-picker/date-picker.scss
```diff
@@ -111,6 +112,35 @@
...
+  &.iui-range {
+    border-radius: 0;
+    cursor: pointer;
+    @include themed {
+      background-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-5));
+      color: t(iui-text-color);
+    }
+  }
+
+  &.iui-range:hover {
```
adamhock, line 136: Should the same `font-weight` change and a new `background-color`change be added to `iui-range-start` and `iui-range-end`?

veekeys: hmm.. Maybe we need something, as this is clickable now.
@FlyersPh9 any suggestions on the colors on active days on hover?

Also, I see this focus issue between range and range-end days..
image

FlyersPh9:
How about this?  Threw this together pretty quick for proof of concept, code will need some cleanup I'm sure:
Screen Shot 2022-03-21 at 12 43 18 PM
```scss
&.iui-today {
  position: relative;
  font-weight: $iui-font-weight-semibold;

  &::before {
    content: '';
    position: absolute;
    display: block;
    width: 36px;
    height: 36px;
    border-radius: $iui-border-radius-round;
    box-sizing: border-box;
    top: 50%;
    left: 50%;
    @include themed {
      border: 2px solid rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
    }
  }
}
```

src/date-picker/date-picker.scss
FlyersPh9: The focus outline next to start / end of range I think might be ok.  The outline is still visible on 3 of the 4 sides.

veekeys: @FlyersPh9 and any suggestions about hover on range-start and range-end days?

veekeys:
> How about this? Threw this together pretty quick for proof of concept, code will need some cleanup I'm sure: Screen Shot 2022-03-21 at 12 43 18 PM
> 
> ```css-scss
> &.iui-today {
>   position: relative;
>   font-weight: $iui-font-weight-semibold;
> 
>   &::before {
>     content: '';
>     position: absolute;
>     display: block;
>     width: 36px;
>     height: 36px;
>     border-radius: $iui-border-radius-round;
>     box-sizing: border-box;
>     top: 50%;
>     left: 50%;
>     transform: translate(-50%, -50%);
>     @include themed {
>       border: 2px solid rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-5));
>     }
>   }
> }
> ```

I have pushed this change, thanks

src/date-picker/date-picker.scss
FlyersPh9: I'm thinking leave it as is.  [Google Flights](https://www.google.com/travel/flights) date range picker doesn't have a hover effect on the start / end dates.

veekeys:
Ok, moved all days to separate classes and ended up with 0.2.1 specificity as highest..
Hopefully, nothing broken.
Also, added different today styling, so should look ok also on selected and range days.
Please review again

src/date-picker/date-picker.scss
```diff
@@ -62,58 +64,251 @@
...
+      &::before {
+        @include themed {
+          background-color: initial;
+        }
+      }
+    }
+
+    &::before {
+      @include themed {
+        border-color: t(iui-color-foreground-accessory);
```
FlyersPh9, line 181:
```suggestion
        border-color: rgba(t(iui-color-foreground-accessory-rgb), t(iui-opacity-4));
```
Can you try this?  I think it looks a little less attention grabbing.
Screen Shot 2022-03-22 at 10 25 36 AM

src/date-picker/date-picker.scss
```diff
@@ -4,16 +4,23 @@
 @import '../icon/index';
 @import '../menu/index';
 
+$iui-date-picker-cell-width: 40px;
+$iui-date-picker-cell-height: 36px;
```
FlyersPh9, line 8: Nice, so basically as if `$iui-baseline * 3` where baseline is equal to 12px.  I see what you did there. 😉

src/date-picker/date-picker.scss
```diff
@@ -4,16 +4,23 @@
 @import '../icon/index';
 @import '../menu/index';
 
+$iui-date-picker-cell-width: 40px;
```
FlyersPh9, line 7: Could this be 36px as well or does this introduce issues?

src/date-picker/date-picker.scss
```diff
@@ -4,16 +4,23 @@
...
+$iui-date-picker-cell-height: 36px;
+
 @mixin iui-date-picker {
   @include iui-dropdown;
   box-shadow: $iui-elevation-2;
```
FlyersPh9, line 12: `@include iui-dropdown;` includes `box-shadow: $iui-elevation-2;` so this line can be removed.

src/date-picker/date-picker.scss
veekeys: Well yeah, Thats the minimum width I can give to the day, so everything can fit in nicely :)
But try it out, maybe you will find out better solution

src/date-picker/date-picker.scss
```diff
@@ -62,58 +64,220 @@
...
+      @include themed {
+        border-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-4));
+      }
+    }
+  }
+}
+
+@mixin iui-calendar-day-outside-month {
+  &,
+  &-today {
```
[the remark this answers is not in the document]
veekeys: oh.. not ony that.. range days also missed it. Added focus mixin to base mixin, hopefully its fine :/

src/date-picker/date-picker.scss
veekeys: removed

src/date-picker/date-picker.scss
veekeys: Just set the stage..
Looks pretty good, doesnt it?

src/date-picker/date-picker.scss
FlyersPh9: Changing it to 36px doesn't seem to cause any issues.  I think we can safely change this to 36px.

backstop/tests/date-picker.html
```diff
@@ -120,7 +120,7 @@ <h2>Date</h2>
           >
             <div class="iui-calendar-week">
               <div
-                class="iui-calendar-day iui-outside-month"
+                class="iui-calendar-day-outside-month"
```
FlyersPh9, line 123: What is the benefit of 1 class VS modifier classes?  Seems tricky when you introduce combos like today + within range.

src/date-picker/date-picker.scss
```diff
@@ -62,58 +63,220 @@
...
+@mixin iui-calendar-day-today-base {
+  position: relative;
+  font-weight: $iui-font-weight-semibold;
+
+  &::before {
+    content: '';
+    position: absolute;
+    display: block;
+    width: $iui-date-picker-cell-height;
+    height: $iui-date-picker-cell-height;
```
bentleyvk, line 109: For me it doesn't look nice when circle is from the very top to the bottom. What if we reduce its size a bit? Here it is 32x32
image

src/date-picker/date-picker.scss
```diff
@@ -62,58 +63,220 @@
...
   @include iui-focus;
   cursor: pointer;
-  display: inline-block;
-  width: $iui-baseline * 3;
-  line-height: $iui-baseline * 3;
+  text-align: center;
+  width: $iui-date-picker-cell-width;
+  height: $iui-date-picker-cell-height;
   border-radius: $iui-border-radius;
+  font-variant-numeric: tabular-nums;
```
bentleyvk, line 87: What is the benefit of having this? Because I don't see any changes when toggling it.

veekeys: Try hovering over the day and see numbers moving horizontally while changing font-weight

backstop/tests/date-picker.html
veekeys: Well, I would need to handle `iui-today` modifier for every different day (simple day, selected day, range day, etc) by adding a selector to it. So it increases complexity and most important specificity.

backstop/scenarios/date-picker.js
```diff
@@ -14,6 +14,9 @@ module.exports = [
   scenario('Type Date & Time 24 hour format', {
     selectors: ['#demo-date-and-time-24h'],
   }),
+  scenario('Type Date Range', {
+    selectors: ['#demo-date-range'],
+  }),
 
   // Hover states
```
bentleyvk, line 21: Maybe add test for hovered and focused range day.

src/date-picker/date-picker.scss
veekeys: Changed

backstop/scenarios/date-picker.js
veekeys: Added tests, but hover does not work :/

src/date-picker/date-picker.scss
FlyersPh9: This is how I found out about this CSS property: https://twitter.com/javan/status/1486059026064584711

FlyersPh9 (approved):

backstop/scenarios/date-picker.js
adamhock: Is this still broken? Looks good to me

veekeys: Yeah, if you check image, there is no hover style in it :/

veekeys: I removed hover test.. Couldnt figure why it was not working..

bentleyvk (approved):

PR #532, opened 2022-03-17
feat(UserIcon): Add icon to busy status

FlyersPh9:
- Moves the 4 status icons out of HTML and into the CSS (same approach as checkbox / radio).
- Adds an icon to the busy indicator so that we get a better accessibility score in #505.
- Adds new `.iui-offline` modifier class.

Screen Shot 2022-03-17 at 12 47 51 PM


veekeys (approved):
Icon does not really make sense to me.. In Teams there is such icon when person is in do not disturb mode. Icon name also says not really being used on that purpose :D
In teams busy has same as we had before
image

But if it helps something, is ok, I guess..

FlyersPh9:
@elephantcatdog @veekeys I changed how the icons are shown, maybe take another look and make sure everything is good.

veekeys:
Now its bad.. Can review only tomorrow..
Build is red, icons are added in html, even though not used anymore..
And might be some more potential issues..

elephantcatdog:
The build is red because the test images weren't updated.

I don't think we have to match Teams. Slack has completely different user status symbols.
image
If you search ["user busy icon"](https://www.google.com/search?q=user+busy+icon&rlz=1C1GCEA_enUS891US891&hl=en&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjOuZKE7s32AhXEjokEHRjbB7MQ_AUoAXoECAEQAw&biw=1536&bih=722&dpr=1.25) my results have the "-" symbol appear frequently, so it's not a foreign idea.

I like the consistency of having all symbols or all no symbols, but I prefer the symbols because they give a little bit more context.
Ultimately, I don't think it matters that much. Every iteration of this was fine and understandable. However, if the point of adding this icon was to improve the accessibility score, wouldn't removing the icons decrease the score?

Some other solutions:
- Change the status from "Busy" to "Do not disturb", because the idea behind them is not that different, and "Do not disturb" is perhaps more actionable on the part of other users. If these statuses are automatic I would not do this though.
- Add another status and icon for "Do not disturb". This would match Teams exactly but the "Busy" icon would not have a symbol

src/user-icon/user-icon.scss
```diff
@@ -91,23 +91,24 @@
...
     }
 
-    > .iui-status-symbol {
-      display: inline-flex;
-      @include themed {
-        fill: t(iui-icons-color);
-      }
+    &::after {
+      content: '';
+      mask: url("data:image/svg+xml,<svg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'><path d='m12.5 2-4.5 4.5-4.5-4.5-1.5 1.5 4.5 4.5-4.5 4.5 1.5 1.5 4.5-4.5 4.5 4.5 1.5-1.5-4.5-4.5 4.5-4.5z' /></svg>");
```
veekeys, line 96: Previously, we could have passed what ever icon there, now there is always gonna be `offline` one by default...
Should we add class for offline icon too?

PR #533, opened 2022-03-17
feat(Surface): Add surface component

FlyersPh9:
- Add surface component needed for iTwinUI-Layouts.
- `@mixin iui-surface;` replaces `@mixin iui-dropdown;`.

Screen Shot 2022-03-17 at 12 50 44 PM

Various elevations:
Screen Shot 2022-03-17 at 2 02 34 PM


src/surface/surface.scss
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-surface {
```
mayank99, line 5: Are you sure about this name? Surface is too generic, and use case is only type of surface

FlyersPh9: We threw around some name ideas in the layouts call, I liked "surface" most.  Greta pointed out MUI uses "paper".  I'm open to whatever we want to call it.

mayank99: ok but you're missing the important part: **_this use case is only <u>one</u> type of surface_**

What if the user wants a background-2 or background-3 surface? What if they don't want the shadow or want different elevation? All of those things should be configurable in a `<Surface>` component imo.

FlyersPh9: I've added support for different elevations.  I cant think of an instance where changing background color would be needed, we can add that later if we need it.

backstop/tests/surface.html
```diff
@@ -0,0 +1,81 @@
...
+        display: flex;
+        align-items: center;
+        height: 44px;
+        border-bottom: 1px solid var(--iui-color-background-5);
+        padding: 0 12px;
+      }
+      .demo-surface-content {
+        padding: 11px 12px;
+      }
+      #demo-elevations {
```
bentleyvk, line 36: Can you add padding to the demo sections so box-shadow would be visible?

src/surface/surface.scss
bentleyvk: For me "surface" also sounds too generic. Also, we are kind of copying this from Material UI which has category named Surface that contains these kind of components. As I told in layouts meeting, I would go for the "Card".

src/surface/classes.scss
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './index';
+
+.iui-surface {
+  &,
+  &-2 {
```
bentleyvk, line 7: I would say that these should be modifier classes like `iui-surface iui-surface-elevation-2`

src/surface/surface.scss
```diff
@@ -0,0 +1,16 @@
...
+@import '../style/index';
+
+@mixin iui-surface($elevation: $iui-elevation-2) {
+  box-shadow: $elevation;
+  box-sizing: border-box;
+  @include themed {
+    background-color: t(iui-color-background-1);
+    color: t(iui-text-color);
+  }
+
```
bentleyvk, line 12: Do we need rounded borders?

src/surface/surface.scss
mayank99: Card can be confused with Tile

src/surface/classes.scss
mayank99: What's the benefit of modifiers?

src/surface/surface.scss
gretanausedaite: I don't think `Card` is the best option as most people think that card should have more elements than just background and shadow. 
image

Maybe `iui-elevated-background`? or `iui-elevated-surface`?

gretanausedaite: We could add semi-transparent or colored surfaces later

backstop/tests/surface.html
FlyersPh9: Test images are better now with negative margin & positive padding.

src/surface/surface.scss
FlyersPh9: Surface makes the most sense to me.  It gives you an elevated "surface" above bg2 to place content, whether it be a tile looking thing or a dropdown menu.
Screen Shot 2022-03-18 at 9 29 34 AM

src/surface/classes.scss
FlyersPh9: Having a surface without a shadow didn't seem useful, so I made the 2 base classes equal the smallest shadow.  I could make the class names go by 1, it's matching the box-shadow variables now.

src/surface/surface.scss
FlyersPh9: All the use cases for it currently don't require rounded corners.  Maybe we can add it later as an option if we need it?

src/surface/classes.scss
mayank99: Should we use CSS variables to allow easily customizing elevation, background and border-radius from the React component?

```tsx
const Surface = ({ elevation = 1, background = 1, borderRadius = 1, ...props }) => {
  return (
    <div
     {...props}
     style={{
       'iui-surface-elevation': `var(--iui-elevation-${elevation})`,
       'iui-surface-background': `var(--iui-color-background-${background})`,
       'iui-surface-border-radius': `var(--iui-border-radius-${borderRadius})`,
     }}
    />
  );
};
```

Usage:
```tsx
<Surface elevation={2} background={2} borderRadius={3}>...</Surface>
```

FlyersPh9: A couple of changes:
- You can apply the surface using classes `.iui-surface .iui-elevation-1`.
- You can apply the surface using mixin `@include iui-surface;`.
- Elevations are now `.iui-elevation-1` - `.iui-elevation-5` & no number defaults to `.iui-elevation-1`.

FlyersPh9: I think it makes sense for our menu component to use the mixin because the menu should always have that surrounding box with drop shadow.
However, for date picker it might make more sense to use the classes because you may want to embed the calendar somewhere in a page w/o the drop shadow.

backstop/scenarios/surface.js
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
+  scenario('Type default', {
+    selectors: ['#demo-default'],
+    viewports: [{ width: 800, height: 600 }],
+    hideSelectors: ['h1', 'hr', '#demo-elevations'],
```
bentleyvk, line 11: Do we really need these? Since we are targeting specific sections, these elements are not even being captured.

src/surface/classes.scss
bentleyvk: > Should we use CSS variables to allow easily customizing elevation, background and border-radius from the React component?

Then we would also convert SASS variables for elevation and others to CSS variables?

src/surface/surface.scss
veekeys: I like `Card` :D

backstop/scenarios/surface.js
FlyersPh9: The negative margin & positive padding on the `<section>` to help capture the drop shadow are capturing some of the surrounding elements.  Screenshot of the test results after removing the `hideSelectors`:
Screen Shot 2022-03-21 at 12 55 01 PM

src/surface/classes.scss
FlyersPh9: Latest push switches to CSS variables, let me know if this is not what you were expecting.

src/surface/surface.scss
```diff
@@ -0,0 +1,27 @@
...
+@import '../style/index';
+
+@mixin iui-surface($background-color: var(--iui-color-background-1), $border-radius: 0, $elevation: $iui-elevation-2) {
+  @include surface($background-color, $border-radius);
+  @include elevation($elevation);
+}
+
+@mixin surface($background-color: var(--iui-color-background-1), $border-radius: 0) {
+  --iui-surface-background-color: #{$background-color};
+  --iui-surface-border-radius: #{$border-radius};
```
bentleyvk, line 12: Maybe remove for now as we don't have CSS variable for that and we only use one value (no?).

mayank99: Still doesn't hurt to keep, as you can set any value from outside, e.g. `--iui-surface-border-radius: 12px` via react prop `<Surface borderRadius='12px' />`

src/surface/classes.scss
mayank99: yes, we don't need those extra classes either because we have CSS variables

FlyersPh9: How should apply the various elevations on the surface demo page without the modifier classes?

mayank99: Use inline CSS variables with hardcoded values.

We should in the future move those Sass variables to CSS variables so that we can reference them in the inline styles.

mayank99 (approved):

PR #541, opened 2022-03-22
fix(UserIcon): Use pixel values for status icon

mayank99:
Before:
image

After:
image


src/user-icon/user-icon.scss
```diff
@@ -92,8 +92,8 @@
 
     &::after {
       content: '';
-      width: calc(100% / 1.5);
-      height: calc(100% / 1.5);
```
FlyersPh9, line 96: This definitely helps with the centering and looks better in the smaller status icons, but appears a little bit smaller in the larger status icons.  

I tried replacing with `width: $iui-icons-small;` but it looked bugged in the smaller sizes:
Screen Shot 2022-03-22 at 5 04 21 PM

I tried some combination of `calc` and `round` but wasn't having any success there either.  Without setting the width and height explicitly for each size, this may be the only approach I can think of atm.

FlyersPh9, line 96: This definitely helps with the centering and looks better in the smaller status icons, but appears a little bit smaller in the larger status icons.  

I tried replacing with `width: $iui-icons-small;` but it looked bugged in the smaller sizes:
Screen Shot 2022-03-22 at 5 04 21 PM

I tried some combination of `calc` and `round` but wasn't having any success there either.  Without setting the width and height explicitly for each size, this may be the only approach I can think of atm.

src/user-icon/user-icon.scss
```diff
@@ -92,8 +92,8 @@
 
     &::after {
```
FlyersPh9, line 93: Maybe we should include `border-radius` to ensure the icon doesn't appear outside of the status circle shape?

This is what it looks like with & without the `border-radius` and with the mask disabled:
Screen Shot 2022-03-22 at 5 08 04 PM Screen Shot 2022-03-22 at 5 07 44 PM

FlyersPh9, line 93: Maybe we should include `border-radius` to ensure the icon doesn't appear outside of the status circle shape?

This is what it looks like with & without the `border-radius` and with the mask disabled:
Screen Shot 2022-03-22 at 5 08 04 PM Screen Shot 2022-03-22 at 5 07 44 PM

src/user-icon/user-icon.scss
veekeys: I also do not really like making status icon smaller.. It looks too small on bigger user icon.

veekeys:
Tests are failing.
Also, why image without status icon changed?

src/user-icon/user-icon.scss
```diff
@@ -92,8 +92,8 @@
...
       content: '';
-      width: calc(100% / 1.5);
-      height: calc(100% / 1.5);
+      width: $iui-s;
+      height: $iui-s;
```
gretanausedaite, line 96: I don't like that all icons are the same size.
image

Though, I'm not sure how to solve this, because if we changed sizes based on user icon size, it would like 1px change and we don't have variables for that. So maybe it's ok.

veekeys: We have 3 sizes of user icons. Cant we just specify for each?

src/user-icon/sizes.scss
```diff
@@ -1,13 +1,19 @@
...
 @import '../style/index';
 @import '../text/mixins';
 
 @mixin iui-user-icon-size($size: $iui-icons-xl, $font-size: $iui-font-size-leading) {
   width: $size;
   height: $size;
-  --_iui-user-icon-status-size: #{math.round(math.div($size, 6))};
+
+  @if ($size == $iui-icons-xl) {
+    --_iui-user-icon-status-size: #{$iui-xs};
```
veekeys, line 11: hmm.. Not too small?
image

Size is 12px (for some reason),
so I guess icon should be 6px (not in our sizes :D)

mayank99: imo we shouldn't even we be showing a status icon at that size. 6px looks too big

size is 12px but it has a 2px border on all sides so it becomes 8px.

veekeys: well idk.. Now it looks super bad.. all of them except maybe green one

mayank99: | 6px | 4px |
| --- | --- |
| image | 
image |

veekeys: Yeah, checkmark seems too big, but then others are at least visible :DDD
idk... idea not to show them at all at that small size is also worth thought...
@FlyersPh9 check.
I am approving from my side 6px, but removing them completely also is fine to me.

veekeys (approved):

gretanausedaite (approved):

PR #548, opened 2022-03-25
ci: revert build script

veekeys:
Just revert the script to previous code.
If my guess is right, we would have only issue when failed tests are rerun and we have deploy step marked as red.
But if this fixes our current deployed pages issues, I think it is fine for now.


bentleyvk (approved):
Let's try this out!

PR #550, opened 2022-03-25
chore: Add Open Sans to demo pages

gretanausedaite:
- Add `Open Sans` to demo pages
- Update test images

Changed components (size because of font change):
- Badge
- Buttons
- Carousel (inner content on first slide)
- Fieldset
- Side nav
- Skip to content


veekeys (approved):
There are some width changes in certain elements (or even height?) Might be interesting to see if some of our components got broken.. But reviewing 400 images is too much, so hopefully will notice as we go.

mayank99:
This is not a feature that we are releasing in the package, so `feat` shouldn't be used

gretanausedaite:
> This is not a feature that we are releasing in the package, so `feat` shouldn't be used

chore?

FlyersPh9:
> There are some width changes in certain elements (or even height?) Might be interesting to see if some of our components got broken.. But reviewing 400 images is too much, so hopefully will notice as we go.

@veekeys that might be a result of the Open Sans font being slightly larger/smaller than whatever font was being loaded previously.

PR #553, opened 2022-03-28
chore: deploy js files

veekeys:
We exclude all folders with not needed files.
Maybe this fixes deployment with changed js files (web components)


bentleyvk:
Don't you want to remove this condition to test whether it is working? 
image

veekeys:
> Don't you want to remove this condition to test whether it is working? image

I tested and it is working!

image

bentleyvk (approved):
Maybe also worth removing these but minor, so approving.
image

yea, need to remove now manually and script should not add anymore

PR #557, opened 2022-03-28
fix(Modal): Update modal animation

elephantcatdog:
Updates modal animations to account for movement of components in React.
image
`FocusTrap` was moved from being the parent of `CSSTransition` to being its child. `CSSTransition` now applies the transition class names to `FocusTrap` instead of "iui-modal".

The following changes were made to account for the above:
- Moved all react animations into a separate mixin "iui-modal-animation" that matches the class name used in React. The class names are no longer tied to "iui-modal".
- The animations for the full page modal have increased specificity (added ".iui-modal" on lines 132 and 137) so that these are not overwritten by the rule `transform: translateY(0)` on line 122, which has a specificity of 3.
  - The modal should probably be refactored to make the specificity better, but I want to finish up the full page modal without committing to the time needed to achieve that.


veekeys (approved):

src/modal/modal.scss
```diff
@@ -137,16 +125,28 @@
...
-  &.iui-modal-animation-enter > .iui-modal-dialog {
+@mixin iui-modal-animation {
+  // Full Page Modal Animations for React
+  &-enter .iui-modal .iui-modal-dialog {
     transform: translateY(100%);
     opacity: 0;
   }
 
-  &.iui-modal-animation-enter-active > .iui-modal-dialog {
+  &-enter-active .iui-modal .iui-modal-dialog {
```
FlyersPh9, line 137: Lines 132 and 137, is the `.iui-modal` part needed?  Would this not work?
```scss
&-enter-active .iui-modal-dialog {
```

veekeys: It wouldnt, because this has higher specificity and would override
image

elephantcatdog: What Vykintas says is correct.

FlyersPh9 (approved):

Issue #512, opened 2022-03-07
Skip to content button is not accessible in Safari


FlyersPh9:
Safari doesn't support `:focus-visible` and I think that is making the element un-focusable.

mayank99:
Can you try replacing this line
https://github.com/iTwin/iTwinUI/blob/e40f387d94781fb0dcb6ebce0bbfea74c4404931/src/skip-to-content/skip-to-content.scss#L7
with this?
```css
outline: 2px solid var(--iui-color-foreground-primary);
outline-offset: 2px;
```

Even if that works, I think we might want to investigate why this fallback is not working.
https://github.com/iTwin/iTwinUI/blob/e40f387d94781fb0dcb6ebce0bbfea74c4404931/src/style/mixins.scss#L28-L32

FlyersPh9:
I followed the steps on [this page](https://stackoverflow.com/q/52783375) to enable tabbing in Safari, no action needed.

PR #558, opened 2022-03-28
fix(SkipToContent): Hide skip-to-content link when not focused

FlyersPh9:
When investigating #512 I discovered that the **skip-to-content link** was visible when touch scrolling up above the top of the page.
Skip-to-content link visible when touch scrolling

- Change opacity & apply transition effect to link.
- Skip to content demo page presentation changes.

Screen Shot 2022-03-28 at 3 48 00 PM


src/skip-to-content/skip-to-content.scss
```diff
@@ -15,23 +15,26 @@
...
     @include iui-blur($opacity: 2);
   }
 
   &:focus {
+    opacity: 1;
     transform: translateX(-50%) translateY(0);
     box-shadow: $iui-elevation-16;
     @media (prefers-reduced-motion: no-preference) {
-      transition: transform $iui-speed-fast ease-in-out;
+      transition: opacity $iui-speed-fast ease-in-out, background-color $iui-speed-fast ease-in-out,
```
mayank99, line 36: I think these should have a delay of `0` so that they're instantly visible

src/skip-to-content/skip-to-content.scss
```diff
@@ -15,23 +15,26 @@
...
   text-align: center;
   left: 50%;
+  opacity: 0;
   top: $iui-baseline * 2;
   transform: translateX(-50%) translateY(-170%);
   transition: background-color $iui-speed-fast ease-in-out;
   @media (prefers-reduced-motion: no-preference) {
-    transition: background-color $iui-speed-fast ease-in-out, transform $iui-speed-slow ease-in-out,
-      box-shadow $iui-speed-fast ease-in-out;
+    transition: opacity $iui-speed-slow ease-in-out, background-color $iui-speed-fast ease-in-out,
```
mayank99, line 23: why this weird mix of `slow` and `fast`?

backstop/tests/skip-to-content.html
```diff
@@ -177,19 +177,25 @@ <h1>Skip to content link</h1>
...
       </div>
     </section>
-    <br>
+
+    <br />
+
     <section id="demo-skip-to-here">
-      <div style="height: 1000px; border: solid 1px;">
-        <p>This is the main content. <a
+      <div style="height: 1000px; border: solid 1px var(--iui-color-background-border); padding: 11px 12px;">
```
mayank99, line 194: Can you move these into the `style` tag?

backstop/tests/skip-to-content.html
```diff
@@ -177,19 +177,25 @@ <h1>Skip to content link</h1>
...
         </div>
       </header>
-      <div style="border: solid 1px">
-        <p>This is not the main content. Press 'tab' to see skip-to-content-link component. You might need to click on
-          the top of the page or the URL first. Press 'enter' after focusing on skip-to-content-link to skip to the main
+      <div style="padding: 11px 0">
+        <p>This is not the main content. Press&nbsp;<kbd class="iui-keyboard">tab</kbd>&nbsp;to see skip-to-content-link
+          component. You might need to click on
+          the top of the page or the URL first. Press&nbsp;<kbd
+            class="iui-keyboard"
```
mayank99, line 184: Can you format this properly?

backstop/tests/skip-to-content.html
FlyersPh9: Moved from inline styling into `<style>`.  Was able to remove the styling of the top section completely and replaced with a `<br />`.

src/skip-to-content/skip-to-content.scss
FlyersPh9: Everything within the media query is now matching slow.  This is so when the button loses `:focus` it slowly fades away off the screen.
The one `transition` outside of the media query can be set to fast for the button hover effect.  Tested with reduced motion on to make sure everything is working as expected.

src/skip-to-content/skip-to-content.scss
FlyersPh9: Here's a screen recording with reduced motion on & then reduced motion off.  It is instant when reduced motion is enabled.
Screen Recording 2022-03-28 at 4 14 35 PM

mayank99: I'm talking about the opacity and background-color even when reduced-motion is off. I think @elephantcatdog intended for the link to be fully visible when it starts sliding into view

backstop/tests/skip-to-content.html
```diff
@@ -177,20 +183,23 @@ <h1>Skip to content link</h1>
           </button>
         </div>
       </header>
-      <div style="border: solid 1px">
-        <p>This is not the main content. Press 'tab' to see skip-to-content-link component. You might need to click on
-          the top of the page or the URL first. Press 'enter' after focusing on skip-to-content-link to skip to the main
-          content below.</p>
+      <br />
```
bentleyvk, line 186: Why changed?

FlyersPh9: I moved the inline styling into `<head>` and the first section didn't really need the box surrounding it.  Also using the `<kbd>` element where we call out the keys within the description.

elephantcatdog (approved):
I think the animation is sufficiently noticeable currently. The code looks fine to me.
Also I really like the changes you made to the html page.

PR #559, opened 2022-03-28
fix(ButtonGroup): Add `width: 100%` for overflow handling

mayank99:
This will address https://github.com/iTwin/iTwinUI-react/issues/602


FlyersPh9 (approved):

src/button/button-group.scss
```diff
@@ -106,3 +106,8 @@
     }
   }
 }
+
+@mixin iui-button-group-overflow-x {
```
veekeys, line 110: so when would you apply that class in react?

src/button/button-group.scss
```diff
@@ -106,3 +106,8 @@
...
+  max-width: revert;
```
veekeys, line 111: Remind again, why max-width is always needed?

src/button/button-group.scss
mayank99: `{ 'iui-button-group-overflow-x': !!overflowButton && orientation === 'horizontal' }`

src/button/button-group.scss
mayank99: Hmm I think it could be removed now actually

src/button/button-group.scss
veekeys: and if you always keep it applied, whats breaking?

src/button/button-group.scss
```diff
@@ -10,9 +10,7 @@
   align-items: center;
   isolation: isolate;
 
-  @if $orientation == 'horizontal' {
-    max-width: 100%;
-  } @else {
+  @if $orientation != 'horizontal' {
```
veekeys, line 13: so just `== 'vertical'`

mayank99: This is the personal convention I chose for this component. See https://github.com/iTwin/iTwinUI/pull/503#discussion_r816922553

veekeys: huh? But dont you want to apply that rule only when vertical is chosen (and not some random stuff) ?

src/button/button-group.scss
veekeys: But if I remember correctly, wehad 100% before your previous changes...no?

src/button/button-group.scss
mayank99: by that logic you would never use `@else`

veekeys: I really dont understand , where are you going with this or not having else :DD
But w/e, its minor not worth spending hours..

PR #561, opened 2022-03-29
feat(Tag): Classes change and specificity improvement

veekeys:
- Changed `iui-label` and `iui-button` to have specific classes
- Improved specificity a little.. still have 0.3.0 bcs of button mixins and 0.3.1 in container, which i havent really touched
- Hope this might help a little in using tags in multi select
- In container replaced selector to any children, so mixin could be reused in select too for margins.
- Images change in container, because of margin. Need to talk about this


elephantcatdog (commented):
Code seems fine to me. I don't have a lot of experience with specificity yet though.

But these changes seem like they'd break tags already in use? I had to hold off on some changes before for that reason. Is this okay for refactors or is this going to be put into CSS 1.0?

veekeys:
> Code seems fine to me. I don't have a lot of experience with specificity yet though.
> 
> But these changes seem like they'd break tags already in use? I had to hold off on some changes before for that reason. Is this okay for refactors or is this going to be put into CSS 1.0?

They break css, but react should stay the same. Css is 0.x, so we produce breaking changes with our refactors.
What was your change which you needed to hold off?

elephantcatdog:
The change I held off was adding flex to the regular modal. [Link here.](https://github.com/iTwin/iTwinUI/pull/489#:~:text=Ugh%2C%20I%20am%20not%20sure%20whether%20we%20can%20apply%20flex%20to%20the%20dialog%20because%20if%20someone%20has%20several%20inline%20elements%20next%20to%20each%20other%2C%20then%20we%20would%20break%20their%20layout%3A)

src/tag/tag.scss
```diff
@@ -1,83 +1,88 @@
...
-      @include iui-text;
-      white-space: nowrap;
-      text-overflow: ellipsis;
-      overflow: hidden;
-      margin: 0 $iui-xs 0 $iui-s;
+@mixin iui-tag {
+  @include iui-tag-base;
+  height: $iui-baseline * 3;
+  margin: $iui-component-offset 0;
+  border-radius: ($iui-baseline * 3) * 0.5;
```
FlyersPh9, line 30:
```suggestion
  border-radius: $iui-border-radius-round;
```
We have a new variable for this.

src/tag/tag.scss
```diff
@@ -1,83 +1,88 @@
...
+  box-sizing: border-box;
+  padding: 0 $iui-xxs;
+  align-items: center;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: border $iui-speed-fast ease-out;
```
FlyersPh9, line 35: The `transition: border` doesn't need to be within the `prefers-reduced-motion` media query because it doesn't have to do with motion.

FlyersPh9:
What do you think if we get rid of `.iui-tag .iui-basic` and instead use `.iui-anchor`?  I can add CSS to #525 that handles anchor links that have `href` and ones that don't:
Screen Shot 2022-04-01 at 5 06 54 PM

src/tag/tag.scss
```diff
@@ -1,83 +1,88 @@
...
-      }
+  &:hover {
+    @include themed {
+      border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-1));
     }
+  }
+}
 
-    > .iui-button {
-      border-radius: 50%;
```
bentleyvk, line 52: Hm, is it really worth it to have all button CSS when only border radius changed?

src/tag/tag.scss
```diff
@@ -1,83 +1,88 @@
...
-    > .iui-label {
-      @include iui-text;
-      white-space: nowrap;
-      text-overflow: ellipsis;
-      overflow: hidden;
-      margin: 0 $iui-xs 0 $iui-s;
+@mixin iui-tag {
+  @include iui-tag-base;
+  height: $iui-baseline * 3;
+  margin: $iui-component-offset 0;
```
bentleyvk, line 29: Is it ok to have top and bottom margins?

src/tag/classes.scss
```diff
@@ -6,7 +6,19 @@
...
+
+.iui-tag-label {
+  @include iui-tag-label;
+}
+
+.iui-tag-button {
+  @include iui-tag-button;
+}
+
+a.iui-tag-basic {
```
bentleyvk, line 21: Focus state looks weird:
image

src/tag/tag.scss
```diff
@@ -1,83 +1,88 @@
...
     color: t(iui-text-color-muted);
   }
 
-  > a.iui-basic {
+  > a.iui-tag-basic {
     margin: $iui-xxs;
   }
 
-  > .iui-tag:not(:last-child) {
+  > *:not(last-child) {
```
bentleyvk, line 81: What if you add gap with `@support` check so in the future we could easily find it and remove with IE11 deprecation?

backstop/tests/tag.html
```diff
@@ -109,13 +109,13 @@ <h2>Default tag containers</h2>
...
             title="Tag"
           >
             Tag
           </span>
           <button
-            class="iui-button iui-borderless iui-small"
+            class="iui-tag-button"
             aria-label="Delete tag"
             type="button"
             tabindex="-1"
```
bentleyvk, line 121: Why it is set to `-1`? It is not tabbable because of that.

backstop/tests/tag.html
```diff
@@ -390,13 +390,13 @@ <h2>Default tag containers</h2>
           tabindex="0"
```
bentleyvk, line 390: We need to change test background because `iui-visible` background is not so visible 😀

veekeys:
> The change I held off was adding flex to the regular modal. [Link here.](https://github.com/iTwin/iTwinUI/pull/489#:~:text=Ugh%2C%20I%20am%20not%20sure%20whether%20we%20can%20apply%20flex%20to%20the%20dialog%20because%20if%20someone%20has%20several%20inline%20elements%20next%20to%20each%20other%2C%20then%20we%20would%20break%20their%20layout%3A)

Link did not really put me to the correct comment about flex in that PR. But basically, in css we are making breaking changes cause we are still not on stable version (0.x). But it is important to think from react user perspective. If our css change might influence current react users with their solutions, we usually try to avoid it. So maybe that was the case with modal and flex?

src/tag/classes.scss
veekeys: Yeah, but it was same before :D
Might have been broken with focus refactors @FlyersPh9  ?

backstop/tests/tag.html
veekeys: No idea.. 
I also see we have `tabindex` on whole tag set in html.. in react we do not, so in react only close button is focusable.
@FlyersPh9 do you know the intention here?

src/tag/tag.scss
veekeys: I dont know.. Left as it was. Removing those would require to add some gaps between lists of tags.. Also, might break existing layouts for users?

src/tag/tag.scss
veekeys: I have tried that, but then lists with `iui-truncate` applied breaks...
image
(notice ellipsis not having margin...)

Should I then add margins back if `iui-truncate` applied?
I couldnt find any info about `-webkit-box` display we use.. Mozila page has its link broken, so not sure.

src/tag/tag.scss
veekeys: Changed

src/tag/tag.scss
veekeys: Using new var

mayank99:
> > The change I held off was adding flex to the regular modal. [Link here.](https://github.com/iTwin/iTwinUI/pull/489#:~:text=Ugh%2C%20I%20am%20not%20sure%20whether%20we%20can%20apply%20flex%20to%20the%20dialog%20because%20if%20someone%20has%20several%20inline%20elements%20next%20to%20each%20other%2C%20then%20we%20would%20break%20their%20layout%3A)
> 
> Link did not really put me to the correct comment about flex in that PR. But basically, in css we are making breaking changes cause we are still not on stable version (0.x). But it is important to think from react user perspective. If our css change might influence current react users with their solutions, we usually try to avoid it. So maybe that was the case with modal and flex?

To clarify, itwinui-css is pre-1.0 in which breaking changes are allowed, but itwinui-react is 1.0+ so we cannot break things there (from a user perspective). What this means practically is, any breaking changes we make in itwinui-css should be "consumed" fully in itwinui-react so that they are basically "invisible" to the user.

So in the case of `Modal`, you can (for example) change the classname from `iui-modal-dialog` to `iui-dialog` and we could easily handle that in react. But if you apply `display: flex` then it changes the layout mode of the children, i.e. in `<Modal><A/><B/><C/></Modal>`, it would affect/break `A`, `B`, and `C`. If we were to release itwinui-react 2.0, we could just say wrap your content in `<ModalContent>` but we can't right now.

backstop/tests/tag.html
FlyersPh9: I think the reasoning was clicking anywhere on the tag would delete it… but thinking about that now, I'm not sure I like that interaction.  It is not ideal that you have to hit the tab key twice to get to the 2nd tag (1st tab goes to the tag, 2nd tab goes to delete button).

src/tag/classes.scss
FlyersPh9: I think the top/bottom clipping was present with the old focus styling as well, it just wasn't as noticeable.  I tried using `overflow-x` but that didn't work as expected.  Maybe decrease the height of the anchor and add a padding to the container so the focus doesn't go out of bounds?

backstop/tests/tag.html
veekeys: I do agree if we let focus whole tab, something should be interactable, though :/
So idk, Im open to doing what ever you guys feel would be the best.

src/tag/classes.scss
veekeys: I changed offset for anchor focus.
@FlyersPh9 wanna make this change in your anchor pr maybe?

src/tag/tag.scss
FlyersPh9: I'll be honest, I'm going back and forth on this decision as well.  For this case, because the user can not pass any custom element in place of the close button, I vote we do this and not use the mixin:
```html
<span class="iui-tag">
	<span class="iui-tag-label">Hello</span>
	<button class="iui-button iui-borderless iui-small iui-tag-button">X</button>
</span>
```
```scss
.iui-tag-button {
	border-radius: $iui-border-radius-round;
}
```

src/tag/classes.scss
FlyersPh9: Looks fine to me, I'll make the anchor PR match in style.

src/tag/tag.scss
```diff
@@ -1,83 +1,86 @@
...
   @include iui-reset;
   @include iui-focus($offset: -2px, $thickness: 2px);
   user-select: all;
   text-transform: lowercase;
   display: inline-flex;
+}
+
+@mixin iui-tag-basic {
+  @include iui-tag-base;
+  outline-offset: 1px;
```
FlyersPh9, line 17: You have `outline-offset` set here in the CSS and you are setting it when including the `iui-focus` mixin within `classes.scss`.  Are both required?

src/tag/tag.scss
```diff
@@ -1,83 +1,86 @@
...
+  height: $iui-baseline * 3;
+  margin: $iui-component-offset 0;
+  border-radius: $iui-border-radius-round;
+  box-sizing: border-box;
+  padding: 0 $iui-xxs;
+  align-items: center;
+  transition: border-color $iui-speed-fast ease-out;
+  font-size: $iui-font-size;
+  text-transform: none;
+  cursor: default;
```
FlyersPh9, line 37:
```suggestion
  cursor: default;
  -webkit-tap-highlight-color: transparent;
```
Can you add this tap highlight color change in as well?  It's an iOS thing, I just tested it within my Simulator app.

src/tag/tag.scss
veekeys: I guess not.. removed

FlyersPh9 (approved):

elephantcatdog (approved):
Still seems fine to me.

backstop/tests/tag.html
bentleyvk: But how close icon should be clicked then? I see in react we allow to focus close icon.

src/tag/tag.scss
bentleyvk: Doesn't make sense to apply margins with truncate, then it is ok.

src/tag/tag.scss
bentleyvk: We removed margins from toggle switch 😀 It will also affect the Tile but idk if it is bad. Or were margins added to make it nicely aligned with buttons/texts.

backstop/tests/tag.html
bentleyvk: Was this change done?

veekeys: Yes

backstop/tests/tag.html
veekeys: Should we get rid of focus on tag and just have close button focusable?

src/tag/tag.scss
bentleyvk: Then we can leave as it is for now.

backstop/tests/tag.html
FlyersPh9: I'm fine with that.

veekeys: Ok, so I have removed focus from tag and its styling..
Later we can revisit, cause I see mui has whole tag focusable and interactive, not button inside it (but then we should have the button without separate styling, IMO)

bentleyvk (approved):

PR #564, opened 2022-03-30
fix(Modal): Update animation speed

elephantcatdog:
- Fixed mistake with class name selectors for full page modal react animations - previously it was affecting both regular and full page modals (lines 135 and 140)
- Fixed problem with full page modal background flickering on exit - the end of the animation was being clipped off (line 113)
- Improved (hopefully?) full page modal speed with custom variables. The changes included making the entry and exit speeds a little bit faster, with the exit speed a tiny bit faster than the entry. This makes it look more balanced, even though it actually isn't. I did not change ease-in/out as ease-in is supposed to be used for exits and ease-out is supposed to be used for entrances. I followed guidelines such as [this webpage](https://css-tricks.com/ease-out-in-ease-in-out/#:~:text=Google%20is%20saying%20don%E2%80%99t%20ease%2Dout%20on%20the%20way%20out) for both ease-in/out and timing (exit being faster than entrance). (lines 94-95, 120, and 127)
  - Another option for speed is changing it to 0.35s for entry and 0.3s for exit if current speeds (0.3s and 0.25s) are too fast

https://user-images.githubusercontent.com/37936169/160772000-52f6c61f-ad16-4464-9948-fb2142f086de.mp4


src/modal/modal.scss
```diff
@@ -91,6 +91,9 @@
 }
 
 @mixin iui-modal-full-page {
+  --_iui-full-page-modal-slide-in-speed: 0.3s;
```
veekeys, line 94: Why you decided to use css vars here? Do you plan to change these values in react? Usually, we use css vars, if we plan to change the values somewhere else (css vars are used in browser scope).
In this case, maybe just simple sass variables are enough?

gretanausedaite (commented):
I agree with using sass variables unless we plan allowing users to change animation speed in react.
Other than that - looks good to me.

PR #566, opened 2022-03-31
chore: Pass auth token on auto release

veekeys:
So our auto release is failing and the guess is that it needs auth token to be passed explicitly to the reused workflow.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -3,6 +3,9 @@ name: Publish to npm
 on:
   workflow_dispatch:
   workflow_call:
+    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-npm.yml
```diff
@@ -3,6 +3,9 @@ name: Publish to npm
 on:
   workflow_dispatch:
   workflow_call:
+    secrets:
+      NPMJS_PUBLISH_ITWIN:
+        required: true
```
mayank99, line 8: Will this affect workflow_dispatch in any way? Don't want manual publish to break

veekeys: It shouldnt, I added it only for `workflow_call`

PR #568, opened 2022-03-31
refactor(Breadcrumbs)

FlyersPh9:
**WIP**

- Chevron icons are placed using `::after`.
- Built in a way to replace the breadcrumb CSS within `header.scss`.
- Support for `forced-colors`


src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
   user-select: none;
 }
 
-@mixin iui-breadcrumbs-item {
+@mixin iui-breadcrumbs-item($breadcrumbs-height: $iui-component-height) {
```
veekeys, line 21: Plan to have different height?

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
+      background-color: CanvasText;
+    }
   }
+}
 
-  .iui-button {
-    @include iui-button-borderless;
-    margin: 0 (-$iui-s - 1);
+@mixin iui-breadcrumbs-text {
+  margin: 0 ($iui-s + 1);
```
veekeys, line 53: 9px? Something is wrong

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
+    @include iui-anchor;
   }
+}
+
+@mixin iui-breadcrumbs-button {
+  @include iui-button;
+  @include iui-button-borderless;
 
-  > *,
   .iui-button-label {
```
veekeys, line 68: so this part is weird.. We kinda create a new `component` for breadrcums button, but inside still use same elements from original button..
And in react it makes no fun for these custom classes to be applied (without creating new components).
So I am not so sure anymore, what should be the best approach.. Including mixins and having new components, or just use existing components, especially, when there is no rules to be changed/overriden

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
-    margin: 0 (-$iui-s - 1);
+@mixin iui-breadcrumbs-text {
+  margin: 0 ($iui-s + 1);
+  max-width: 24ch;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+
+  @at-root a#{&} {
+    @include iui-anchor;
```
veekeys, line 60: Are we ok with different focus stylings?

image

veekeys:
Technically, title is wrong. It should be feat, I guess, because there are a bunch of breaking changes, so we at least should increase minor version of package..

gretanausedaite (commented):
I have no additional comments for now. 
Focus does not look consistent but they are different components (iui-breadcrumbs-text and iui-breadcrumbs-button) so maybe ok?

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
+  --_iui-breadcrumbs-item-separator-svg: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path d='m4.7 0-1.4 1.4 6.6 6.6-6.6 6.6 1.4 1.4 8-8z' /></svg>");
+
   display: flex;
-  line-height: $iui-component-height;
-  height: $iui-component-height;
-  max-width: $iui-3xl * 2;
-  margin: 0 $iui-sm;
+  align-items: center;
+  line-height: $breadcrumbs-height;
+  height: $breadcrumbs-height;
```
mayank99, line 27: I'm guessing this is the part that helps with header?

But I don't see much else connection tbh. Header and breadcrumbs not only look different (apart from the separator) but also have completely different implementations, especially in React. And there is not too much code, so I think we could just repeat it in header.

Plus, these two lines don't even do anything in the base breadcrumbs.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
+  text-overflow: ellipsis;
+
+  @at-root a#{&} {
+    @include iui-anchor;
   }
+}
+
+@mixin iui-breadcrumbs-button {
+  @include iui-button;
+  @include iui-button-borderless;
```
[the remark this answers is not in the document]
mayank99: Also, how would a user add icons to the breadcrumbs now?

mayank99:
While I'm struggling to understand how this PR is blocking header, I do believe we should try to find a solution for CSS 1.0 that does not target those inner classes from `button.css` but also does not break React.

In the react component, we explicitly say the user can pass anything, so I think our solution needs to support 3 types of components that the user can pass inside the list:
- `<Button>` (which just means `<button class='iui-button iui-default'>`)
- `<a>` (including custom links, e.g. `<Link>` from react-router). Also might want to support our `<Anchor>` (`iui-anchor`).
- any custom component, which simply means that we handle things like separator, and the user provides the actual styling. Incidentally, this also sounds a lot like the `header` case.

If we make a `BreadcrumbButton` component, that's in _addition_ to what we currently support. We can't stop supporting the above until react 2.0.

I'm not sure of the best path forward, but I think we could use CSS variables similar to `--_iui-button-active-stripe-inset`

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
-  a {
-    @include iui-anchor;
+  &:not(:last-child)::after {
+    flex-shrink: 0;
+    width: $iui-icons-small;
+    height: $iui-icons-small;
+    margin: 0 $iui-xxs;
+
+    // #region IE support & future support
+    content: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m4.7 0-1.4 1.4 6.6 6.6-6.6 6.6 1.4 1.4 8-8z' fill='%23000' fill-opacity='0.4' /%3E%3C/svg%3E");
```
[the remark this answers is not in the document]
veekeys: I checked IE, was looking OK.

image

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
+    line-height: 14px;
+
+    @supports (mask: var(--_iui-breadcrumbs-item-separator-svg)) {
```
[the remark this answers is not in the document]
veekeys: I think it will work

image

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
   user-select: none;
 }
 
-@mixin iui-breadcrumbs-item {
+@mixin iui-breadcrumbs-item($breadcrumbs-height: $iui-component-height) {
+  --_iui-breadcrumbs-item-separator-svg: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path d='m4.7 0-1.4 1.4 6.6 6.6-6.6 6.6 1.4 1.4 8-8z' /></svg>");
```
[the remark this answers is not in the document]
veekeys: oh damn.. I was looking at breadcrumbs, but no idea, how I didnt notice this prop...

src/breadcrumbs/breadcrumbs.scss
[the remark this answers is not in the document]
mayank99: oh wow. i love autoprefixer

backstop/scenarios/breadcrumbs.js
```diff
@@ -5,11 +5,6 @@
 const { scenario, hover, focus } = require('../scenarioHelper');
 
 module.exports = [
-  // Breadcrumb types
-  scenario('Type Breadcrumb', {
```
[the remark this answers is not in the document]
FlyersPh9: iTwinUI_breadcrumbs_State_hover_and_focus_buttons_0_demo-buttons_0_desktop
One image tests default, hover, focus, and non-clickable / current location.  I can revert this change though if you'd like.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Our buttons have `$iui-s` padding & 1px border.
Screen Shot 2022-04-04 at 2 44 14 PM

I added this here so the hyperlink & button text align.
Screen Shot 2022-04-04 at 2 46 55 PM

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: When we refactor the buttons the mixin should then reference `&-label` so we could use `.iui-breadcrumbs-button-label` instead of `.iui-button-label`.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: They are different currently, I don't think it should be an issue.  [I do want to match the width](https://github.com/iTwin/iTwinUI/pull/568#discussion_r842041621) incase we had a scenario where the breadcrumb was switching from hyperlink to button mode, the content won't shift around.

src/breadcrumbs/breadcrumbs.scss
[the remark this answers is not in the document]
FlyersPh9: I don't like this part either at the moment.  What I think I'd like to do is update the button mixins to target `&-label` so we could use `.iui-breadcrumbs-button-label` [[other similar discussion](https://github.com/iTwin/iTwinUI/pull/568#discussion_r842045197)].

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
 
-  > *,
   .iui-button-label {
     overflow: hidden;
-    white-space: nowrap;
     text-overflow: ellipsis;
-  }
-
-  &:not(.iui-current) > .iui-button {
+    max-width: 24ch;
```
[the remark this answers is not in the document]
FlyersPh9: We were using `max-width: $iui-3xl * 2;` to start truncation.  `24ch` results in the same amount of characters so the test images don't change.  Maybe we could make this a CSS variable in the future so the truncation limit could easily be adjusted?

backstop/tests/breadcrumbs.html
```diff
@@ -23,12 +23,10 @@
       href="../assets/demo.css"
     />
     <style>
-      .iui-input { width: 450px; }
+      .iui-input { width: 517.5px; }
```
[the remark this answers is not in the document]
FlyersPh9: I set it to match the breadcrumb container.

Screen Shot 2022-04-04 at 3 02 25 PM

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,67 @@
...
-    @include iui-anchor;
+  &:not(:last-child)::after {
+    flex-shrink: 0;
+    width: $iui-icons-small;
+    height: $iui-icons-small;
+    margin: 0 $iui-xxs;
+
+    // #region IE support & future support
+    content: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='m4.7 0-1.4 1.4 6.6 6.6-6.6 6.6 1.4 1.4 8-8z' fill='%23000' fill-opacity='0.4' /%3E%3C/svg%3E");
+    line-height: 14px;
```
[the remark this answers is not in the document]
FlyersPh9: Line 36 should be undone by line 40.  Line 36 I placed the actual SVG code but on line 40 I used a CSS variable.  IE will display line 36, while modern browsers will use line 40.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: This was in prep to be used within header, but I'm going to remove it from this PR and add it in if I need it for the header because I'm uncertain atm.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I've removed `$breadcrumbs-height` but may need to add it in when the header work resumes.

src/breadcrumbs/breadcrumbs.scss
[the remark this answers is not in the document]
FlyersPh9: `breadcrumbs.scss` now handles the new breadcrumb specific components & the user passed `.iui-button` and `<a>`.  Although keeping the styling for the user passed items is increasing specificity past our 0,3,0 goal.  I believe those specificity issues are coming from `@include iui-button-borderless;`.

src/button/button.scss
```diff
@@ -38,12 +41,21 @@ $iui-button-padding-large: $iui-xs * 6;
   @include themed {
     color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
   }
+  color: var(--_iui-button-label-color);
+
+  &-label {
```
mayank99, line 46: I think this belongs in the `iui-button-label` mixin

src/button/button.scss
```diff
@@ -11,6 +11,9 @@ $iui-button-padding-large: $iui-xs * 6;
 
 @mixin iui-button {
   --_iui-button-active-stripe-inset: initial;
+  --_iui-button-label-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
```
mayank99, line 14: I don't think this is just `label-color`, it is more like `foreground-color` because it also applies to the icon fill

src/button/cta.scss
```diff
@@ -6,12 +6,16 @@
 @import './disabled';
 
 @mixin iui-button-cta {
+  --_iui-button-label-color: var(--iui-color-foreground-accessory);
+  --_iui-button-label-color-hover: --_iui-button-label-color;
+
+  @include iui-focus($color: var(--iui-color-foreground-accessory), $offset: -3px, $thickness: 1px);
```
mayank99, line 12: Can we not move the `iui-focus` around for no reason? From an organizational and cascade perspective, also doesn't make sense to have modifiers (i.e. `:focus`/`:focus-visible`) before the base styling

src/button/button.scss
FlyersPh9: Changed to `--_iui-button-foreground-color`.

src/button/cta.scss
FlyersPh9: Moved the `iui-focus` mixin back to where it was originally.

src/button/button.scss
FlyersPh9: It would appear we don't have a `@mixin iui-button-label`.  I can create one if you'd like, it was my plan to do so when we do the CSS 1.0 work on buttons.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -18,47 +18,77 @@
...
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
   }
 
-  .iui-button {
+  > .iui-button {
     @include iui-button-borderless;
-    margin: 0 (-$iui-s - 1);
+    --_iui-button-label-character-count: 24ch;
```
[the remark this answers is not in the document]
FlyersPh9: That would work for the anchor but not for the button because the max-width be set on the button's shape rather than the label text.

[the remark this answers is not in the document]
FlyersPh9: Removed CSS variable and setting max width on the parent (button body rather than label).

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Focus styling should be identical now.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Went the opposite approach and removed the border, so now only `$iui-s` is used.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -28,57 +28,64 @@
...
     @include iui-anchor;
-    max-width: 24ch;
-    margin: 0 $iui-s;
+    padding: 0 $iui-s;
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
+
+    &:focus,
+    &:focus-visible {
```
mayank99, line 37: `:focus-visible` is unnecessary here and will break the whole selector where it's not supported. https://css-tricks.com/one-invalid-pseudo-selector-equals-an-entire-ignored-selector/

```suggestion
    &:focus {
```

Same thing in the other instance below at line 71

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -28,57 +28,64 @@
...
+  > * {
+    max-width: 26ch;
+  }
+
+  &:not(.iui-current) > * {
+    &,
+    &:hover,
+    &:active {
+      --_iui-button-foreground-color: var(--iui-color-foreground-primary);
+    }
```
mayank99, line 56: This is not nice. If I remove `:hover` and `:active`, then it looks like they have the same specificity but `iui-button` rules are declared later in the stylesheet so they take preference
image

I think in the future we might want to use `:where` for the hover/active states in base button. For now this is fine.

src/button/default.scss
```diff
@@ -7,17 +7,18 @@
...
-  @include iui-focus($offset: -2px, $thickness: 2px);
   @include themed {
     background-color: t(iui-color-background-1);
     border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-4));
   }
+  @include iui-focus($offset: -2px, $thickness: 2px);
 
   &:hover,
   &:active {
+    --_iui-button-foreground-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-1));
```
mayank99, line 20: This is already set in `button.scss` so it doesn't need to be repeated. Same thing at the top of the mixin on line 10.

```suggestion
```

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Yikes, good catch, fixed.

src/button/default.scss
FlyersPh9: Removed.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I have an idea on how to solve this.  Instead of setting the button color & button hover color with two separate RGBA, what if we set the just the button color and change the button label opacity instead?  The added benefit is it would be easier for others to override the button color, needing to only set 1 color instead of 2.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -28,55 +28,74 @@
...
 
 @mixin iui-breadcrumbs-text {
-  margin: 0 ($iui-s + 1);
-  max-width: 24ch;
+  padding: 0 $iui-s;
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
 
   @at-root a#{&} {
```
mayank99, line 72: Something I didn't realize before: we are styling _all_ anchor elements that have this class because our expected usage is that the current item consists of a `<span>`. This might not be ideal because there are two cases where the current item can be an anchor instead of `<span>`:
- if the current item is an `<a>` without a `href`. See [html spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#:~:text=If%20the%20a%20element%20has%20no%20href%20attribute%2C%20then%20the%20element%20represents%20a%20placeholder).
- (less common) if clicking on the current item is meant to refresh the page. See [w3c example](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html).

In both of these cases, I think we might want to style it differently from the rest of the items.

src/breadcrumbs/breadcrumbs.scss
mayank99: Yeah you could go that route or leave it as is. I'm fine either way, as it's a button issue, not a breadcrumb issue.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -28,55 +28,74 @@
...
-      color: t(iui-color-foreground-primary);
-    }
   }
 
-  @at-root .iui-breadcrumbs-item-current {
-    .iui-button-label {
-      @include themed {
-        color: t(iui-text-color);
-      }
+  @at-root .iui-breadcrumbs-item-current & {
```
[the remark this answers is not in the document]
FlyersPh9: Love that idea, added.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: We could update the anchor to style `<a>` without a `href`, something like this:
Anchor links

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -21,45 +21,94 @@
...
   display: flex;
+  align-items: center;
   line-height: $iui-component-height;
   height: $iui-component-height;
-  max-width: $iui-3xl * 2;
-  margin: 0 $iui-sm;
 
-  a {
+  // #region Component overrides
+  > a {
```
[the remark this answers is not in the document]
veekeys: a wait.. its that ting of duplicate implementation..
Additional comment would be really appreciated, cause I spent good 10mins of trying to figure out why so many rules are duplicated :/

backstop/tests/breadcrumbs.html
```diff
@@ -55,155 +54,324 @@
...
 
-      <section id="demo-anchors">
+      <nav
+        class="iui-breadcrumbs"
+        aria-label="Breadcrumbs"
+      >
+        <button class="iui-button iui-borderless iui-dropdown">
+          <svg-folder
+            class="iui-button-icon"
+            aria-hidden
```
veekeys, line 67: wait, is it valid to have it without value? I remember there was something, that you can use that now, but I am not sure.

backstop/tests/breadcrumbs.html
```diff
@@ -55,155 +54,324 @@
...
+            <a href="#">Root</a>
+          </li>
+          <li
+            class="iui-breadcrumbs-separator"
+            aria-hidden
+          >
+            <svg-chevron-right aria-hidden></svg-chevron-right>
+          </li>
+          <li class="iui-breadcrumbs-item">
+            <span class="iui-breadcrumbs-text">&mldr;</span>
```
veekeys, line 85: so now its enough just to apply class to child element of `iui-breadcrumbs-item` no matter of its type, right? Asking for react.

backstop/tests/breadcrumbs.html
```diff
@@ -55,155 +54,324 @@
...
+          </li>
+          <li
+            class="iui-breadcrumbs-separator"
+            aria-hidden
+          >
+            <svg-chevron-right aria-hidden></svg-chevron-right>
+          </li>
+          <li class="iui-breadcrumbs-item">
+            <button
```
veekeys, line 170: oh, but for button no need that text class..
That complicates implementation in react :/
Users can pass anything there, check would not be nice at all..

backstop/tests/breadcrumbs.html
```diff
@@ -55,155 +54,324 @@
...
+              class="iui-button"
```
veekeys, line 171: for the real scenario, you forgot to add `iui-default` class here. And then it breaks styling :D

backstop/tests/breadcrumbs.html
```diff
@@ -55,155 +54,324 @@
...
+          </li>
+          <li
+            class="iui-breadcrumbs-separator"
+            aria-hidden
+          >
+            <svg-chevron-right aria-hidden></svg-chevron-right>
+          </li>
+          <li
+            class="iui-breadcrumbs-item"
+            aria-current
```
veekeys, line 199: so li element shouldnt get `aria-current`. Child element would, right? remove this

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -21,45 +21,94 @@
...
     @include iui-anchor;
```
veekeys, line 30: Interesting...

image

image

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -21,45 +21,94 @@
...
     }
   }
 }
 
 @mixin iui-breadcrumbs-separator {
   display: flex;
+  margin: 0 $iui-xxs;
 
   svg {
     @include iui-icons-small;
```
veekeys, line 108: also this :DD

image

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -21,45 +21,94 @@
...
+
+    &-label {
+      overflow: hidden;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+    }
+  }
+
+  > * {
+    max-width: 26ch;
```
veekeys, line 53: in date picker its 15, in badge its 20, here 26... what was the reason for this value?

src/breadcrumbs/breadcrumbs.scss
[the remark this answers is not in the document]
mayank99: You are misusing `aria-current`. It needs to go on the button/anchor, not the `<li>`. Also, it needs some value, like `"location"` or `"page"`

src/breadcrumbs/breadcrumbs.scss
mayank99: Um, no this is not really about the anchor component. It's a specific use-case only for breadcrumbs

src/button/button.scss
```diff
@@ -38,12 +39,12 @@ $iui-button-padding-large: $iui-xs * 6;
   @include themed {
     color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
   }
+  color: var(--_iui-button-foreground-color);
```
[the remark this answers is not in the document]
mayank99: this was my suggestion. now `color` is only set in one place and we can just change the value of the css variable. it's fine to break hover on IE imo

backstop/tests/breadcrumbs.html
mayank99: `aria-hidden` without a value doesn't do anything. It needs to be `aria-hidden="true"`.

src/breadcrumbs/breadcrumbs.scss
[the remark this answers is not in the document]
FlyersPh9: It is now targeting `aria-current="page"` on the button / anchor.

backstop/tests/breadcrumbs.html
FlyersPh9: Based off [this article](https://css-tricks.com/write-html-the-html-way-not-the-xhtml-way/#aa-examples=) I thought only `aria-hidden` would be needed, I can add `="true"`.

backstop/tests/breadcrumbs.html
FlyersPh9: But our live implementation only has `.iui-button`, no `.iui-default`?

backstop/tests/breadcrumbs.html
FlyersPh9: Moved `aria-current` to the button, no longer on the `<li>`

backstop/tests/breadcrumbs.html
veekeys: What?
Inspect it yourself
https://itwin.github.io/iTwinUI-react/?path=/story/core-breadcrumbs--basic

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Yeah, these are coming from the anchor mixin.  I suppose we could split the anchor mixin up even more so we don't include the external modifier in this case.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I had `max-width: 24ch` set on the anchor & the button label but Mayank thought it made more sense to set it on the anchor and button body instead.  We landed on 26ch only because it did not change the test image results.

src/button/button.scss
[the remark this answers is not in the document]
FlyersPh9: What I would like to do in the CSS 1.0 button refactor is set the color on the `<button>` and then change the label's opacity for default & hover state.  That way if a user wants to override the color, they only need to define 1 color for both states.

backstop/tests/breadcrumbs.html
mayank99: wow that article is overflowing with bad advice

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I have this part ready to go:
```scss
> a {
  ...

  &:not([href]) {
    &,
    &:hover {
      text-decoration: none;
      cursor: default;
      @include themed {
        color: t(iui-text-color);
      }
    }
  }
}
```
But I'm struggling to remove the underline on the high contrast themes.
Screen Shot 2022-04-15 at 8 28 37 AM

src/breadcrumbs/breadcrumbs.scss
mayank99: > We landed on 26ch 

I never suggested 26ch (or any value) fwiw. All II said is that it makes sense to `max-width` on the button/anchor instead of the text.

I had this same question in my other comment: https://github.com/iTwin/iTwinUI/pull/568#discussion_r841635928

backstop/tests/breadcrumbs.html
FlyersPh9: Oh… interesting.  We don't have it present in [the iTwinUI-CSS breadcrumbs demo](https://itwin.github.io/iTwinUI/backstop/minified/breadcrumbs.html).

veekeys: well, its a bug. Having a button with only `iui-button` is not valid. 
There is no way in react to have a button only with base class (cause its not a nicely looking button  and you cannot really use it like that)

src/breadcrumbs/breadcrumbs.scss
mayank99: Hmm, we might need to target the high-contrast classes again if we want to remove the underline.

Also, maybe the [`:any-link`](https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link) selector would help with the specificity here.

src/button/high-visibility.scss
```diff
@@ -6,19 +6,21 @@
...
 
 @mixin iui-button-high-visibility {
+  --_iui-button-foreground-color: var(--iui-color-foreground-accessory);
+
   @include themed {
     background-color: t(iui-color-background-primary);
     border-color: t(iui-color-background-primary);
     color: t(iui-color-foreground-accessory);
   }
+  color: var(--_iui-button-foreground-color);
```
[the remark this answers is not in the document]
gretanausedaite: Line 14 color could be removed.

src/breadcrumbs/breadcrumbs.scss
gretanausedaite: I don't think users will add classes on SVG components.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Updated `/anchor/` so `-external` is now in it's own mixin.

src/breadcrumbs/breadcrumbs.scss
mayank99: So now it only works with `<a>`? Doesn't that break the existing approach of using `<span>`?

Also what about the non-override section? Will need to make it consistent there too.

FlyersPh9: `<a>` with no `href` works.
`<span class="iui-breadcrumbs-text">` works.
`<span>` with no class does _not_ work currently.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -28,55 +28,74 @@
...
+      --_iui-button-foreground-color: var(--iui-color-foreground-primary);
+    }
   }
   // #endregion Component overrides
 }
 
 @mixin iui-breadcrumbs-text {
-  margin: 0 ($iui-s + 1);
-  max-width: 24ch;
+  padding: 0 $iui-s;
```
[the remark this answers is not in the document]
FlyersPh9: I left the padding on `.iui-breadcrumbs-text` and instead set the padding on `.iui-breadcrumbs-button` to 0 for the same end result, eliminating `.iui-breadcrumbs-button-label`.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Do we want `<span>` with no class to work?

src/button/high-visibility.scss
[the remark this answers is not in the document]
FlyersPh9: I think we left line 14 as a fallback for browsers that don't support CSS variables.

backstop/tests/breadcrumbs.html
FlyersPh9: `.iui-breadcrumbs-item` will work for `<a>`, `<a href="">`, & `<span>`.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: If we really want to remove these info, positive, warning, negative styles, I could remove the `@include iui-icons-small;` and just place the sizing and color CSS by itself.

src/button/high-visibility.scss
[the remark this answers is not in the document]
veekeys: Can we at least rename the variable? IT is literally used for setting text color only.

backstop/tests/breadcrumbs.html
veekeys: You still havent fixed it

src/breadcrumbs/breadcrumbs.scss
veekeys: It should work with anything passed under breadcrumb item, IMO..

FlyersPh9: It now works with `<span>`.

src/button/high-visibility.scss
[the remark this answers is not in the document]
FlyersPh9: Variable renamed to `var(--_iui-button-text-color);`.

backstop/tests/breadcrumbs.html
FlyersPh9: Works now with `.iui-button` and `.iui-button .iui-default`.

src/breadcrumbs/breadcrumbs.scss
mayank99: > I could remove the @include iui-icons-small; and just place the sizing and color CSS by itself.

yes, please do that

src/button/button.scss
```diff
@@ -53,7 +53,9 @@ $iui-button-padding-large: $iui-xs * 6;
   &:disabled {
     @include iui-button-disabled;
   }
+}
 
+@mixin iui-button-active {
   &.iui-active {
```
mayank99, line 59: What's the reason for this mixin? Looks really weird that it expects an `iui-active` class to do _anything_.

src/button/button.scss
```diff
@@ -11,7 +11,7 @@ $iui-button-padding-large: $iui-xs * 6;
 
 @mixin iui-button {
   --_iui-button-active-stripe-inset: initial;
-  --_iui-button-foreground-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
+  --_iui-button-text-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
```
mayank99, line 14: What's the reason for this change? @veekeys

This color is used for the icon as well.
image

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -50,9 +50,25 @@
...
+    &.iui-default {
+      @include iui-button-borderless;
+      background-color: transparent;
+      border: none;
+
+      &:focus-visible {
+        outline-offset: -1px;
+        outline-width: 1px;
+      }
+    }
```
mayank99, line 71: Are you sure about all these overrides? I think it's just a specificity issue, which could be worked around by duplicating the class.

```suggestion
    &#{&} {
      @include iui-button-borderless;
      border: none;
    }
```

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -50,9 +50,25 @@
     }
   }
 
+  > span:not([class]) {
```
mayank99, line 53: and what if it's `div` instead of `span`?

FlyersPh9: I've replaced this line with `> *:not([class]) {`.

mayank99: And what if it has a class for some other purpose? You can just use `> *` and even remove these styles from the anchor overrides.

image

src/button/button.scss
FlyersPh9: The breadcrumbs button doesn't need the active state (which was increasing the specificity) so I separated it from the `@mixin iui-button`.  I pushed an update just now that removes the `&.iui-active` from the mixin and instead am applying `&.iui-active { @include iui-button-active; }`. Does that help?

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I have removed `@include iui-icons-small;` and only added what was needed.

src/button/button.scss
mayank99: Yeah, that's better but now it's inconsistent as we also have `iui-active` inside `borderless.scss` and `default.scss`.

I guess lets fix that when we tackle #605

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Damn this PR is kicking my butt. 😣 Thank you for the example, I've implemented it almost exactly as you have it.  I did have to add `.iui-button { &.iui-button { @include iui-focus; } }` because the outline was appearing as 2px instead of 1px.

I did notice that `> * {` within `@mixin iui-breadcrumbs-item-overrides` is overriding `@mixin iui-breadcrumbs-text`.  Should be ok for pre CSS 1.0 I guess.

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: I believe this has been solved with [your suggestion here](https://github.com/iTwin/iTwinUI/pull/568#discussion_r863943914).

src/button/button.scss
veekeys: amm, but not directly. I just went through the variable use, and it was used only for the `color`, so for me it is much clearer when it is named that way.. We can think of other name, just foreground color does not sound obvious what it is, for me at least

veekeys:
Everything looks fine to me, except..
We ended up having 0.5.0 specificity with this one. I think it might be even worse than what we had before :((

gretanausedaite:
> Everything looks fine to me, except.. We ended up having 0.5.0 specificity with this one. I think it might be even worse than what we had before :((

I only found 0,4,0. 

Look good to me. I have no comments on code.

src/breadcrumbs/breadcrumbs.scss
```diff
@@ -21,45 +21,101 @@
...
+  > :any-link {
+    @include iui-anchor;
+
+    &:focus {
+      outline-offset: -1px;
+    }
+  }
+
+  .iui-button {
+    &.iui-button {
```
FlyersPh9, line 53: This is what is causing the 0,5,0.  We need to increase the specificity so that the borderless mixin overrides a button with `.iui-default`.  If you apply `.iui-default` to the buttons on [the live breadcrumb demo](https://itwin.github.io/iTwinUI/backstop/minified/breadcrumbs.html), we get this:
Screen Shot 2022-05-04 at 4 44 04 PM

I'm not sure how we're able to use `.iui-default` in React.

veekeys (approved):

src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Vykintas and I discussed this, seems like the CSS is being imported in a different order in React than in this repo.  In our live demo we were not including the `.iui-default` modifier on our buttons, so they appeared ok.  When adding the modifier class using dev tools they appear incorrectly.  In this PR I am adding the modifier class AND increasing the specificity to 0,5,0 so the import order does not matter.

The only good news is that this older method can be removed in React 2.0.

gretanausedaite (approved):

PR #569, opened 2022-04-01
fix(ColorPicker): Add `forced-colors` support.

FlyersPh9:
- Add `forced-color-adjust: none;` on color swatches, color field, and slider rails.

Before:
Screen Shot 2022-04-01 at 5 16 15 PM

After:
Screen Shot 2022-04-01 at 5 16 30 PM

### DIFF
--- a/src/color-picker/color-picker.scss
+++ b/src/color-picker/color-picker.scss
@@ -49,6 +49,7 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
   margin-right: $iui-xs;
   position: relative;
   background-color: var(--iui-color-swatch-background);
+  forced-color-adjust: none;
   @include iui-transparent-background;
   @supports (gap: $iui-s) {
     margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


src/color-picker/color-picker.scss
```diff
@@ -49,6 +49,7 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
   margin-right: $iui-xs;
   position: relative;
   background-color: var(--iui-color-swatch-background);
+  forced-color-adjust: none;
```
bentleyvk, line 52: I may have missed this earlier but are there any recommendations to do this? Aren't we breaking anything?

veekeys: And not missing this?
image

mayank99: It only affects high contrast mode, so it doesn't break anything else and media query is not really necessary

https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/

bentleyvk (approved):

PR #570, opened 2022-04-04
fix(Tree): Fix spacing for tree node icon

### LINKED ISSUE #567: Tree: Spacing is not applied for tree node icon
We use padding for some reason and if we pass into that place something like checkbox, it does not work.

https://github.com/iTwin/iTwinUI-react/issues/611

I think switching to margin should do the job..

veekeys:
Closes #567 

I needed to increase margin left to child icon, after switching from padding to margin for the icon itself.

### DIFF
--- a/src/tree/tree.scss
+++ b/src/tree/tree.scss
@@ -54,11 +54,11 @@ $iui-expander-button-width: ($iui-icons-default) + ($iui-expander-inline-padding
 
     &-icon {
       @include iui-icons-default;
-      padding: 0 $iui-expander-inline-padding;
+      margin: 0 $iui-expander-inline-padding;
       flex-shrink: 0;
 
       &:first-child {
-        margin-left: $iui-expander-button-width;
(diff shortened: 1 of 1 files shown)


mayank99:
Can you add a screenshot showing before/after? Because I don't really understand what this is solving.

The linked issue says "We use padding for some reason and if we pass into that place something like checkbox, it does not work" but it's not clear what doesn't work.

veekeys:
> Can you add a screenshot showing before/after? Because I don't really understand what this is solving.
> 
> The linked issue says "We use padding for some reason and if we pass into that place something like checkbox, it does not work" but it's not clear what doesn't work.

Sure, sorry.

Before:
image
There are issues with inputs and applying padding on them

image

If you know some other fix, would be really good...

After:
image

mayank99 (approved):

FlyersPh9 (approved):

PR #572, opened 2022-04-04
fix(SkipToContent): Fix Overlapping Visibility Bug

elephantcatdog:
The problem looks like this: 
image
The result of the fix:
image

I added a z-index of 99 - higher than most things but lower than the modals, overlays, and toasts (and tooltips). The alternative was 998 but I thought this followed the pattern better.

**There might be other changes I want to make to this component.

### DIFF
--- a/src/skip-to-content/skip-to-content.scss
+++ b/src/skip-to-content/skip-to-content.scss
@@ -15,8 +15,9 @@
   position: fixed;
   text-align: center;
   left: 50%;
-  opacity: 0;
   top: $iui-baseline * 2;
+  opacity: 0;
+  z-index: 99;
   transform: translateX(-50%) translateY(-170%);
   transition: background-color $iui-speed-fast ease-in-out;
   @media (prefers-reduced-motion: no-preference) {


src/skip-to-content/skip-to-content.scss
```diff
@@ -15,8 +15,9 @@
   position: fixed;
   text-align: center;
   left: 50%;
-  opacity: 0;
   top: $iui-baseline * 2;
+  opacity: 0;
+  z-index: 99;
```
FlyersPh9, line 20: I think this should be fine as is, but I want to share with you something that I do every time I use `z-index`:  I do a repo wide search for `z-index` and then think through "ok, what do I want this to be above & below?
Screen Shot 2022-04-06 at 9 38 54 AM
So for example, we'd want this be beneath the modal but above most of the rest of the stuff.  So 99 should be fine.
"

elephantcatdog: Good tip, that's what I did! 😄

mayank99 (approved):
I would have gone with a higher z-index number as skip links should always be on top, but this works for now.

gretanausedaite (approved):

PR #576, opened 2022-04-05
test(Menu): Remove padding from header demo

mayank99:
This padding was causing flaky tests probably because of some miscalculation in menu position.


mayank99:
The miscalculation is still happening and the test is still flaky. I don't understand.

PR #577, opened 2022-04-07
feat(Menu): Skeleton menu items

bentleyvk:
menu-item-skeleton

Prepares CSS for https://github.com/iTwin/iTwinUI-react/issues/345


src/menu/menu.scss
```diff
@@ -149,6 +150,32 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
       }
     }
   }
+
+  &-skeleton {
```
bentleyvk, line 154: I was thinking whether to create completely separate class (not overriding existing css) but it doesn't have a lot of css and overrides so just extended the existing menu item css.

src/menu/menu.scss
```diff
@@ -149,6 +150,32 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
...
+    --iui-menu-item-skeleton-max-width: 30%;
+    cursor: auto;
+
+    .iui-content {
+      max-width: var(--iui-menu-item-skeleton-max-width);
+    }
+
+    .iui-menu-label {
+      width: 100%;
+      height: $iui-line-height * 0.7;
```
bentleyvk, line 164: Full line height doesn't looks nice as it takes full height so I just to the nicest looking multiplier.

src/menu/menu.scss
```diff
@@ -149,6 +150,32 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
       }
     }
   }
+
+  &-skeleton {
+    --iui-menu-item-skeleton-max-width: 30%;
+    cursor: auto;
+
+    .iui-content {
+      max-width: var(--iui-menu-item-skeleton-max-width);
```
bentleyvk, line 159: I wasn't sure about this variable but it allows easier to override if users decide to make some changes trough css classes.

backstop/scenarios/menu.js
```diff
@@ -11,11 +11,21 @@ module.exports = [
...
+    viewports: [{ width: 800, height: 600 }],
+  }),
 
   // Hover & focus states
   scenario('State hover & focus', {
     actions: [hover('#test-menu-1'), focus('#test-menu-2')],
     selectors: ['#demo-default'],
     viewports: [{ width: 800, height: 600 }],
   }),
+  // Skeleton should not be have hover background
```
bentleyvk, line 25: Just to prevent bugs in the future if something changes with the menu item background.

src/menu/menu.scss
FlyersPh9: I tested it using our font sizes, I like this approach.
Screen Shot 2022-04-07 at 11 36 46 AM

src/menu/menu.scss
elephantcatdog, line 154: I suppose this is a bit out of scope as it deals with the colors defined in skeleton.scss, but the colors for high contrast light don't appear to change from light. High contrast dark does, although I'm guessing that is just because the background is darker? None of the (base) colors pass AA or AAA on the backgrounds used.
Light:
image
High Contrast Light:
image
Dark:
image
High Contrast Dark:
image

FlyersPh9: Yikes, we will definitely need to look into this.

src/menu/menu.scss
```diff
@@ -149,6 +150,32 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
...
+
+    .iui-menu-description {
+      width: 70%;
+      height: $iui-line-height * 0.5;
+      vertical-align: middle;
+      margin-top: $iui-xs;
+    }
+
+    &:hover {
+      background-color: unset;
```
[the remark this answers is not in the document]
bentleyvk: But it also adds `cursor: not-allowed` so it is kind of the same.

src/menu/menu.scss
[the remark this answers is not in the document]
bentleyvk: > this is kinda misleading as it doesnt include the icon in this max-width

Is it really that big of a problem that icon size is not added? Because it is used just to mimic content.

>I assume skeleton menu items wouldn't be selectable, but if they are then the max-width seems to cause the checkmark icon to be placed incorrectly.

Is there a case where selected checkmark should be shown in skeleton? Because then I could work to fix it.

src/menu/menu.scss
bentleyvk: Does it need to be AA or AAA on skeleton?

src/menu/menu.scss
bentleyvk: Thanks for the suggestion! Added.

src/menu/menu.scss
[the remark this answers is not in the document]
mayank99: I just meant the name is misleading, maybe add `content` in it. Of course ideally it would be nice if the icon could be somehow included in the max-width, but that might be difficult

src/menu/menu.scss
mayank99: I think it needs to be AA in default themes and AAA in high contrast themes. Maybe create a separate issue for now?

src/menu/menu.scss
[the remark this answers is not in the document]
bentleyvk: Renamed.

backstop/tests/menu.html
```diff
@@ -429,6 +429,144 @@ <h2>Default</h2>
...
+    <section id="demo-skeleton">
+
+      <ul
+        class="iui-menu"
+        role="menu"
+      >
+        <li
+          class="iui-menu-item iui-menu-item-skeleton"
+          role="presentation"
+          style="--iui-menu-item-content-skeleton-max-width: 30%;"
```
FlyersPh9, line 442: In #587 it uses the width of the text.  Rather than set `max-width` you could use "Lorem ipsum" or something.  Not sure if that would be considered better or worse, just thought I'd share another idea.

bentleyvk: But to make different width content I would need to generate some random text with selected width

bentleyvk:
@FlyersPh9 @mayank99 @elephantcatdog Is anything missing or could this PR get an approve?

bentleyvk:
> Probably not worth holding this PR up, but can you add `aria-hidden` and `aria-busy` to the skeleton classes? I added those in #587 but those decisions haven't been approved yet. Idea taken from here: https://adrianroselli.com/2020/11/more-accessible-skeletons.html

Can `aria-hidden` and `aria-busy` be used together? Won't `aria-hidden` hide element so screen readers won't even know that it is busy?

PR #579, opened 2022-04-07
feat(Tile): Add hover effect

FlyersPh9:
- Hover effect (box-shadow & thumbnail scale) & focus only apply to tiles with `.iui-actionable` modifier.
- Use `@mixin iui-surface` for better `forced-colors` support.

This is to support https://github.com/iTwin/iTwinUI-layouts/pull/20.


backstop/scenarios/tile.js
```diff
@@ -18,104 +18,50 @@ module.exports = [
...
-  // Hover states
-  //// Tile
-  scenario('State hover', {
-    actions: [hover('#test-tile-1')],
-    selectors: ['#test-tile-1'],
-  }),
-  scenario('State hover with thumbnail', {
-    actions: [hover('#test-tile-2')],
-    selectors: ['#test-tile-2'],
+  // Entire tile hover & focus states
```
bentleyvk, line 21: Why changing all of these? Maybe at least with separate PR...

FlyersPh9: I changed the test images because the hover effects were not showing, they were getting clipped.

bentleyvk: But you also renamed and changed a lot of them. That ends up showing as fully changed tests. Therefore we can't see if this hover effect didn't break old UI.

src/tile/tile.scss
```diff
@@ -56,6 +53,20 @@
     }
   }
 
+  &[tabindex] {
```
veekeys, line 56: hm... Not sure any `tabindex` is enough.. There is -1 value, which should disable keyboard.

backstop/scenarios/tile.js
veekeys: yeah.. I agree. Ideally would be good to firstly see the changes and then rename, if needed with separate pr

FlyersPh9: Reverted back to original tests and test images.

src/tile/tile.scss
bentleyvk: I would prefer concrete class. 
Also, `tabindex=-1` is used to allow to focus element through JS but it is not focusable with keyboard. We are using this to make arrow navigation to work.

src/tile/tile.scss
```diff
@@ -56,6 +53,20 @@
...
+    cursor: pointer;
+    transition: box-shadow $iui-speed-fast ease-in-out;
+
+    &,
+    * {
+      user-select: none;
```
bentleyvk, line 62: Pls remove it, now it is impossible to select any text from the Tile.

gretanausedaite (commented):
I agree with current comments. Other than that looks good!

src/tile/tile.scss
FlyersPh9: I added this only to the clickable tiles to avoid accidentally selecting text when clicking on a tile.  My thought was if the entirety of the tile can not be clicked then the text should be selectable.  Do you still think this should be removed?

src/tile/tile.scss
```diff
@@ -56,6 +53,20 @@
...
+    }
+
+    &:hover {
+      box-shadow: $iui-elevation-8;
```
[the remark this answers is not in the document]
FlyersPh9: Updated tests to include these.

src/tile/tile.scss
FlyersPh9: Removed.

src/tile/tile.scss
FlyersPh9: Using `.iui-actionable`.

backstop/scenarios/tile.js
```diff
@@ -21,12 +21,12 @@ module.exports = [
   // Hover states
   //// Tile
   scenario('State hover', {
-    actions: [hover('#test-tile-1')],
-    selectors: ['#test-tile-1'],
+    actions: [hover('#test-tile-4')],
+    selectors: ['#demo-default'],
   }),
   scenario('State hover with thumbnail', {
```
gretanausedaite, line 27: This might be redundant if we remove thumbnail images from tests with [M1 docker PR](https://github.com/iTwin/iTwinUI/pull/571#issuecomment-1098450285)

gretanausedaite: Although, we should have tests for thumbnail scale change. Somehow.

gretanausedaite (approved):
Code looks good to me. Love this new hover drop shadow.

Future ideas: do we want to add hover effects to surface component?

FlyersPh9:
> Future ideas: do we want to add hover effects to surface component?

If the entire surface is actionable, then I suppose we may want to.  Here's a screenshot from Design Review where I imagine the whole surface would be clickable.
Screen Shot 2022-04-14 at 11 42 45 AM

src/tile/tile.scss
```diff
@@ -1,21 +1,18 @@
...
 @import '../style/index';
-@import '../icon/index';
 @import '../button/index';
+@import '../icon/index';
+@import '../surface/index';
 
 @mixin iui-tile {
   @include iui-reset;
+  @include iui-surface;
+  @include iui-focus($offset: $iui-xxs, $thickness: $iui-xxs);
```
veekeys, line 11: So non actionable tile also can be focused? I'd say only interactable ones should get it, no?

backstop/tests/tile.html
```diff
@@ -310,8 +310,9 @@ <h2>Default</h2>
     <section id="demo-default">
       <!-- No thumbnail -->
       <div
-        class="iui-tile"
+        class="iui-tile iui-actionable"
```
veekeys, line 313: Tile width override in html can be also removed as tile has width set of same value

src/tile/tile.scss
```diff
@@ -56,6 +53,15 @@
     }
   }
 
+  &.iui-actionable {
```
veekeys, line 56: So then thumbnail should still have scale change on hover if tile is non actionable?

src/tile/tile.scss
FlyersPh9: I've moved the focus mixin within `.iui-actionable`.

backstop/tests/tile.html
FlyersPh9: Removed tile width from HTML head.

src/tile/tile.scss
FlyersPh9: Moved the thumbnail scaling into actionable hover, focus within, and `.iui-selected`.

veekeys (approved):
This introduces kinda breaking change on Tile which will be visible for react users. Thumbnail will not react on hover anymore, which might be an important change.
Apart from that, hopefully, nothing else will break.

PR #580, opened 2022-04-08
fix(Anchor): Bringing Anchor back to global.scss

bentleyvk:
This change removed iui-anchor from `global.scss` https://github.com/iTwin/iTwinUI/pull/525/files#diff-43b2198fc818aad454f5201f751c89af47c4c7ec583e863a695be7b126e2cdfaL54 and it might cause a lot of breaking UIs so I am bringing it back until v1.


src/style/global.scss
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import './index';
+@import './anchor';
```
mayank99, line 4: Are you not able to keep it in `../anchor/anchor` and import it from there instead of from `style` folder?

bentleyvk: `anchor.scss` was importing from `style` folder. `global.scss` is also importing from `style`. So it would be circular dependency.

mayank99: Could we avoid importing `style` in `anchor.scss` somehow? Maybe by importing from individual files in that folder.

Because it doesn't make sense to move a component out of the main structure.

bentleyvk: Is it really that bad?

bentleyvk: And do you have an idea how not use `themes.scss` then in `anchor`?

mayank99 (approved):

PR #587, opened 2022-04-12
fix(Typography): Accessible skeleton text

### LINKED ISSUE #583: Skeleton: contrast issues
The colors for high contrast light don't appear to change from light. None of the (base) colors pass AA or AAA on the backgrounds used.
Light:
image
High Contrast Light:
image
Dark:
image
High Contrast Dark:
image

_Originally posted by @elephantcatdog in https://github.com/iTwin/iTwinUI/pull/577#discussion_r845387127_

FlyersPh9:
- Use text color variables instead of background colors for better contrast.
- `forced-colors` support.
	- Gradient slanted to look a little better due to lack of animation.
- Cursor change to `progress`.
- Demo page uses toggle switch to enable / disable skeleton text on all typography classes:
	- Demonstrates how the height is inherited by the typography class applied.
	- Applies `aria-hidden`.

Closes #583, [some accessibility ideas taken from this article](https://adrianroselli.com/2020/11/more-accessible-skeletons.html).

**Light:**
Screen Shot 2022-04-12 at 11 33 03 AM

**Dark:**
Screen Shot 2022-04-12 at 11 33 13 AM

**Light high contrast:**
Screen Shot 2022-04-12 at 11 33 22 AM

**Dark high contrast:**
Screen Shot 2022-04-12 at 11 33 30 AM

**Forced colors:**
Screen Shot 2022-04-12 at 11 33 45 AM


src/text/skeleton.scss
```diff
@@ -3,23 +3,29 @@
 @import '../style/index';
 
 @mixin iui-text-skeleton {
-  display: inline-block;
+  width: fit-content;
   user-select: none;
   color: transparent;
   border-radius: $iui-border-radius;
+  cursor: progress;
```
adamhock, line 10: Any reason we shouldn't use `wait` here?

FlyersPh9: `wait` probably makes more sense, I will change it tomorrow when I sign on.

src/text/skeleton.scss
```diff
@@ -3,23 +3,29 @@
 @import '../style/index';
 
 @mixin iui-text-skeleton {
-  display: inline-block;
+  width: fit-content;
```
bentleyvk, line 6: Why changed?

src/text/skeleton.scss
```diff
@@ -3,23 +3,29 @@
...
   @include themed {
     background: linear-gradient(
-      270deg,
-      t(iui-color-background-3),
-      t(iui-color-background-3),
-      t(iui-color-background-2),
-      t(iui-color-background-3),
-      t(iui-color-background-3)
+      292deg,
+      t(iui-text-color-muted),
```
bentleyvk, line 15: Idk, for me previous one looked better 😀 Maybe we could use some color combination?

FlyersPh9: I liked the appearance of the older one better as well, but it did not pass accessibility.

src/text/skeleton.scss
FlyersPh9: `inline-block` took full width, `fit-content` only goes the width of the text it is covering up.

src/text/skeleton.scss
FlyersPh9: Using `wait` cursor now.

src/text/skeleton.scss
FlyersPh9: 1 idea I had was to use `currentColor` so the skeleton text inherits the color.  You could use skeleton text in a CTA button for example.  I think to do that I would need to use `::after` to display the skeleton though.

src/text/skeleton.scss
bentleyvk: I don't think we can use some text for skeletons because when we are showing skeleton, we are waiting for some content to be downloaded so for me it doesn't make sense to add some random text there to show skeleton, also it should break my skeleton menu item because:
image

src/text/skeleton.scss
FlyersPh9: Screen Shot 2022-04-14 at 11 04 23 AM

src/text/skeleton.scss
FlyersPh9: Removed `width: fit-content;`.  Did not replace `display: inline-block;` because I was thinking it makes sense to keep the `display` set to whatever it should be when skeleton is not applied.

FlyersPh9: I also added an additional demo, a `<p>` with no content inside to replicate what you have going on in menu.

src/text/skeleton.scss
```diff
@@ -3,21 +3,26 @@
...
+      t(iui-text-color-placeholder),
+      t(iui-text-color-muted),
+      t(iui-text-color-muted)
     );
-    background-size: 200% 100%;
+  }
+  background-size: 200% 100%;
+
+  @media (forced-colors: active) {
+    forced-color-adjust: none;
```
veekeys, line 24: So when forced colors are turned on, should this marker in middle be moving a little? Looks weird..
typography

src/text/skeleton.scss
```diff
@@ -3,21 +3,26 @@
...
+      292deg,
+      t(iui-text-color-muted),
+      t(iui-text-color-muted),
+      t(iui-text-color-placeholder),
+      t(iui-text-color-muted),
+      t(iui-text-color-muted)
     );
-    background-size: 200% 100%;
+  }
+  background-size: 200% 100%;
```
veekeys, line 21: It is expected that skeleton is same width on all elements, no matter the content?
image

src/text/skeleton.scss
```diff
@@ -3,21 +3,26 @@
 @import '../style/index';
 
 @mixin iui-text-skeleton {
-  display: inline-block;
```
veekeys, line 6: hm...
Sure about this removal?
Thats the reason to that issue in my other comment about same widths.

backstop/tests/typography.html
```diff
@@ -29,20 +29,51 @@
...
-    <p class="iui-text-small">Small</p>
+    </section>
 
-    <p class="iui-text-block iui-text-spacing iui-skeleton">Lorem ipsum dolor sit amet.</p>
+    <script>
+      document.querySelector('#skeleton-toggle input[type=checkbox]').addEventListener('change', (e) => {
+        document.querySelectorAll('.iui-text-spacing').forEach((skeletonText) => {
+          skeletonText.classList.toggle('iui-skeleton');
+          skeletonText.toggleAttribute("aria-hidden");
+          skeletonText.toggleAttribute("aria-busy");
```
bentleyvk, line 74: Busy shouldn't be used as it is ignored in most screen readers.

src/text/skeleton.scss
bentleyvk: Check here 😀
https://adrianroselli.com/2020/11/more-accessible-skeletons.html#Update04

src/text/skeleton.scss
bentleyvk: Aren't they too dark in the light theme?

src/text/skeleton.scss
```diff
@@ -3,21 +3,26 @@
...
   user-select: none;
   color: transparent;
   border-radius: $iui-border-radius;
+  cursor: wait;
```
bentleyvk, line 9: Article is using `progress`

src/text/skeleton.scss
FlyersPh9: That is discussed [here](https://github.com/iTwin/iTwinUI/pull/587#discussion_r849120409).  You can set `width: fit-content` to match the width of the text (if text is present) or you can set `width: 30%` like Vilius did [here](https://github.com/iTwin/iTwinUI/blob/f6657791d42d6818760728615ae4cc45f4a0edb9/src/menu/menu.scss#L155).

bentleyvk: I thought you will just revert your change 😀

veekeys: What?
But now its worse. If I return `inline-block` skeletons take the width of the text, and thats what we want.
Then if users want to have skeleton without actual content (which is often the case), they can set the width (or flex container, which should grow children, i think?)

src/text/skeleton.scss
FlyersPh9: By using the text colors we know that we're passing contrast scores.  I know they are more prominent visually, but I think accessibility trumps visuals here.

src/text/skeleton.scss
FlyersPh9: If I set `display: inline-block;` I'm getting this:
Screen Shot 2022-04-15 at 11 21 29 AM
 
It doesn't seem right to override the `display` when skeleton is set in my opinion.

FlyersPh9: Put `display: inline-block` back & set `section[id^='demo'] { display: flex; flex-direction: column; }`.

src/text/skeleton.scss
FlyersPh9: It's put back now.

src/text/skeleton.scss
FlyersPh9: Switched back to progress.

backstop/tests/typography.html
FlyersPh9: Removed `aria-busy`

adamhock (approved):
Looks good to me

src/text/skeleton.scss
FlyersPh9: Removed `width: fit-content;` again.

PR #588, opened 2022-04-12
fix(ToggleSwitch): Add `forced-colors` support.

FlyersPh9:
Issue discovered when testing `forced-colors` in #587.

- Add `@media (forced-colors: active)` support to make toggle switch usable.  [Source for color names](https://www.w3.org/TR/css-color-4/#css-system-colors).
- Using same active fill color that Microsoft is using [here](https://myaccount.microsoft.com/settingsandprivacy/privacy).

**Before:**
Screen Shot 2022-04-12 at 12 02 05 PM

**After:**
Screen Shot 2022-04-12 at 4 00 47 PM


src/toggle-switch/toggle-switch.scss
```diff
@@ -111,11 +117,18 @@
   &:checked {
     background-color: var(--iui-color-background-primary);
     border-color: var(--iui-color-background-primary);
+    @media (forced-colors: active) {
+      background-color: ActiveText;
+      border-color: ActiveText;
```
mayank99, line 122: I think you are misusing this color as it's meant for active link text (foreground). It's too heavy for use in the background. I would just keep the default background, and use `CanvasText` for the handle.
image

src/toggle-switch/toggle-switch.scss
```diff
@@ -141,6 +154,10 @@
     cursor: not-allowed;
     background-color: var(--iui-color-background-disabled);
     border-color: var(--iui-color-background-disabled);
+    @media (forced-colors: active) {
+      background-color: GrayText;
```
mayank99, line 158: I don't get why this "gray" text shows random colors in the emulated mode.

Can someone on Windows check what this looks like with the actual (not emulated) high contrast setting?

src/toggle-switch/toggle-switch.scss
FlyersPh9: I'm seeing some UIs do that. The one downside to that is not knowing which one is checked / unchecked.
Screen Shot 2022-04-12 at 3 22 48 PM

Microsoft appears to be using `background-color: highlight` on [this page](https://myaccount.microsoft.com/settingsandprivacy/privacy):
Screen Shot 2022-04-12 at 3 27 08 PM
Screen Shot 2022-04-12 at 3 27 13 PM

src/toggle-switch/toggle-switch.scss
FlyersPh9: These are taken from my Windows 10 VM running the latest version of Edge.

Screen Shot 2022-04-12 at 3 39 00 PM
Screen Shot 2022-04-12 at 3 39 18 PM

Screen Shot 2022-04-12 at 3 39 38 PM
Screen Shot 2022-04-12 at 3 39 48 PM

src/toggle-switch/toggle-switch.scss
FlyersPh9: Just confirmed that toggle styling is identical within Windows 10 High Contrast settings menu as well.

src/toggle-switch/toggle-switch.scss
bentleyvk: Windows toggle switch:
image
Ours:
image

bentleyvk: Active looks like disabled in our case. And handle could have brighter color.

elephantcatdog: Vilius' image goes with this theme: 
image
The other themes have a more noticeable difference between disabled text and selected text. The difference between text and disabled text is more different; we could use the text color instead of the selected text color?

src/toggle-switch/toggle-switch.scss
```diff
@@ -111,11 +118,19 @@
   &:checked {
     background-color: var(--iui-color-background-primary);
     border-color: var(--iui-color-background-primary);
+    @media (forced-colors: active) {
+      background-color: Highlight;
+      border-color: Highlight;
```
elephantcatdog, line 123: image
The border color doesn't return to the default color if it's been checked previously.

FlyersPh9: I've added lines 98-100, I think they fixed this.

src/toggle-switch/toggle-switch.scss
FlyersPh9: I changed the toggle switch handle background color & the icon fill to `HighlightText`.  There doesn't appear to be a visual change in the last 2 themes, but the first 2 themes did see a change for the better.
Screen Shot 2022-04-15 at 3 33 52 PM

src/toggle-switch/toggle-switch.scss
elephantcatdog: They did.

src/toggle-switch/toggle-switch.scss
elephantcatdog: I think this looks nice. Well, nice is a strong word. I don't like high contrast, lol, makes my head hurt. But I think the change in handle color is good.

src/toggle-switch/toggle-switch.scss
```diff
@@ -141,10 +159,19 @@
...
+      background-color: GrayText;
+      border-color: GrayText;
+    }
 
     &::after {
       background-color: var(--iui-color-foreground-body);
       opacity: var(--iui-opacity-5);
+
+      @media (forced-colors: active) {
+        background-color: Canvas;
```
mayank99, line 172: Should we try to use different styling for checked/unchecked when in disabled state?

image


vs

image

src/toggle-switch/toggle-switch.scss
```diff
@@ -92,10 +95,17 @@
...
   border-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-4));
+  @media (forced-colors: active) {
+    border-color: CanvasText;
+  }
 
   &::after {
     background-color: var(--iui-color-foreground-body);
     opacity: var(--iui-opacity-2);
+    @media (forced-colors: active) {
+      background-color: CanvasText;
```
mayank99, line 106: Because we are changing the color of the handle between checked/unchecked states, the transition is messed up (in forced-colors mode). Not sure it's worth fixing but just something to be aware of

src/toggle-switch/toggle-switch.scss
FlyersPh9: I support this, change made:
Screen Shot 2022-04-25 at 10 41 58 AM
Screen Shot 2022-04-25 at 10 42 30 AM

src/toggle-switch/toggle-switch.scss
FlyersPh9: It does seem to have a delay & no smooth transition.  I tried duplicating the `transition` within the `@media (forced-colors: active)` but that didn't help, so I'm not really sure how to fix this.  I find it weird that the transition breaks in toggle switches but not on side navigation.  Would be nice if we could fix it, but I would say let's proceed without it for now.

src/toggle-switch/toggle-switch.scss
mayank99: Looks great. This is from Windows 11 with more realistic colors:
image

mayank99 (approved):

elephantcatdog (approved):
Seems good to me, but need to rerun tests

PR #593, opened 2022-04-14
fix(Menu): Improved skeleton menu item accessibility

bentleyvk:
Improved skeleton menu item accessibility based on these 2 articles: 
 - https://adrianroselli.com/2020/11/more-accessible-skeletons.html
 - https://css-tricks.com/a-bare-bones-approach-to-versatile-and-reusable-skeleton-loaders/

Also, fixed JS in HTML demo to not select skeletons.


backstop/tests/menu.html
```diff
@@ -442,7 +442,11 @@ <h2>Skeleton</h2>
           style="--iui-menu-item-content-skeleton-max-width: 30%;"
         >
           <span class="iui-content">
-            <div class="iui-menu-label iui-skeleton"></div>
+            <div
+              class="iui-menu-label iui-skeleton"
+              aria-hidden="true"
+            ></div>
+            <div class="iui-visually-hidden">Loading...</div>
```
FlyersPh9, line 449:
```suggestion
            <div class="iui-visually-hidden">Loading&mldr;</div>
```
@VirginieGaudreault can you check how `Loading...` and `Loading…` are read via screen readers, is there any difference / benefit?

FlyersPh9:
Can you change [line 156 in `menu.scss`](https://github.com/iTwin/iTwinUI/blob/8cd710b7a26745dbc0673f12b328a3552131db54/src/menu/menu.scss#L156) from `cursor: auto;` to `cursor: wait;` to match #587?

bentleyvk:
> Can you change [line 156 in `menu.scss`](https://github.com/iTwin/iTwinUI/blob/8cd710b7a26745dbc0673f12b328a3552131db54/src/menu/menu.scss#L156) from `cursor: auto;` to `cursor: wait;` to match #587?

I am using the same `iui-skeleton` class that you modified in your PR, so no need as this cursor will be applied everywhere skeleton class is used:
image

backstop/tests/menu.html
bentleyvk: I used NVDA and it read to me as `Loading` no dots or ellipses mentioned.

src/style/global.scss
```diff
@@ -56,6 +56,10 @@ html {
   @include iui-anchor;
 }
 
+.iui-visually-hidden {
```
bentleyvk, line 59: I am not sure whether it is ok to have this class globally.

bentleyvk:
> Can you change [line 156 in `menu.scss`](https://github.com/iTwin/iTwinUI/blob/8cd710b7a26745dbc0673f12b328a3552131db54/src/menu/menu.scss#L156) from `cursor: auto;` to `cursor: wait;` to match #587?

What about that article that you dropped?
image

src/style/global.scss
veekeys: hm..
Initially, Id thought to have some file like `utils` which you would include if you need such classes.
But as this is a single class, maybe it can stay :/

backstop/tests/menu.html
veekeys, line 449: So in React you would have a component, which gonna deliver all of these elements and have prop for `loading` text?

bentleyvk: Yup

bentleyvk: `<MenuItemSkeleton hasIcon hasDescription />` something like that.

veekeys (approved):

FlyersPh9:
> > Can you change [line 156 in `menu.scss`](https://github.com/iTwin/iTwinUI/blob/8cd710b7a26745dbc0673f12b328a3552131db54/src/menu/menu.scss#L156) from `cursor: auto;` to `cursor: wait;` to match #587?
> 
> What about that article that you dropped? image

`wait` and `progress` are the same cursor on macOS, are they different on Windows?

adamhock:
> > > Can you change [line 156 in `menu.scss`](https://github.com/iTwin/iTwinUI/blob/8cd710b7a26745dbc0673f12b328a3552131db54/src/menu/menu.scss#L156) from `cursor: auto;` to `cursor: wait;` to match #587?
> > 
> > 
> > What about that article that you dropped? image
> 
> `wait` and `progress` are the same cursor on macOS, are they different on Windows?

Progress:
image

Wait: 
image

adamhock (approved):
Changes look good to me. Good job!

FlyersPh9 (approved):

PR #594, opened 2022-04-19
fix(Checkbox): Add border to disabled checkbox

gretanausedaite:
Added border to disabled checkbox for better visibility.


FlyersPh9 (approved):

src/inputs/checkbox.scss
```diff
@@ -86,7 +86,7 @@
 
   &:disabled {
     --_iui-checkbox-svg-color: var(--iui-icons-color-actionable-disabled);
-    --_iui-checkbox-border-color: var(--iui-color-background-disabled);
+    --_iui-checkbox-border-color: var(--iui-color-background-border);
```
mayank99, line 89: Do we want to exclude it from visibility checkbox? This looks a bit cluttered 
image

FlyersPh9:
Wait for someone else's approval, I only gave a quick glance.  Might be able to apply the border in `checkbox-radio.scss` instead of only in `checkbox.scss` but I don't have time to review more thoroughly.

mayank99 (approved):

src/inputs/checkbox.scss
mayank99: Oh, you could avoid the duplicate variable declarations by using `:enabled` selector inside `:where` (as it does not increase specificity)

image

gretanausedaite: To me it would be a bit more clearer what colours are used in visibility checkbox when variables are duplicated. But it's not good to duplicate code. 
Weren't we removing `:enabled` selector from buttons when refactoring them?

mayank99: We were removing `:enabled` from button for two main reasons:
- specificity (which is solved by `:where`)
- polymorphic buttons (e.g. rendered as `<a>` which does not support `:disabled` or `:enabled`)

I'm fine with duplicate code here, it's only two lines.

bentleyvk (approved):

PR #597, opened 2022-04-22
fix(Toast): Increase z-index

veekeys:
Increased z-index to the same value as we have for popover as toasts were not visible in modal frontstage in itwin.js.
Ideally, I think we should reuse Popover for toasts, modals, so we are sure our components are always visible.

### DIFF
--- a/src/toast-notification/classes.scss
+++ b/src/toast-notification/classes.scss
@@ -17,7 +17,7 @@
   align-items: center;
   flex-direction: column;
   justify-content: center;
-  z-index: 1070;
+  z-index: 50000;
 
   @media screen and (max-width: 400px) {
     width: calc(100vw - 36px);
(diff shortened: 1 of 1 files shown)


src/toast-notification/classes.scss
```diff
@@ -17,7 +17,7 @@
   align-items: center;
   flex-direction: column;
   justify-content: center;
-  z-index: 1070;
+  z-index: 99999;
```
bentleyvk, line 20: Should we have a variable for it?

veekeys: You mean SASS one? Where would you reuse it?

bentleyvk: I wanted to say in popover but then understood that it is set in React. So no need then.

bentleyvk (approved):

mayank99 (approved):

PR #598, opened 2022-04-22
feat(RadioTile): Class name and specificity refactor

adamhock:
- `iui-icon` replaced with `iui-radio-tile-icon`
- `iui-label` replaced with `iui-radio-tile-label`
- `iui-description` replaced with `iui-radio-tile-sublabel`
- `iui-radio` removed
- `iui-radio-tile-input` created

~~- Instead of the `<label>` element being the parent of `<input>`, `<label>` is now the first sibling of `<input>` with class name `iui-radio-tile` to reduce specificity.~~

- hover styling for a radio tile that is enabled and not checked is now moved to `iui-radio-tile-input` and changed when the tile is disabled or checked.

Collab with @FlyersPh9 

EDIT:

- `<label>` element now has class name `iui-radio-tile`
- `<div>` element class name has been changed from `iui-radio-tile` to `iui-radio-tile-content`


adamhock:
It may help to review each commits' changes after the "initial push" commit to see what has changed

src/index.scss
```diff
@@ -42,3 +42,4 @@
 @import './wizard/index';
 @import './skip-to-content/index';
 @import './surface/index';
+@import './radio-tile/index';
```
veekeys, line 45: Would be nice to keep the alphabetical order here.. I see, 2 last additions did not do that :(

src/classes.scss
```diff
@@ -44,3 +44,4 @@
 @import './wizard/classes';
 @import './skip-to-content/classes';
 @import './surface/classes';
+@import './radio-tile/classes';
```
veekeys, line 47: Same here with order.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+@mixin iui-radio-tile {
+  cursor: pointer;
+  width: $iui-xl * 5;
```
veekeys, line 8: Not for this PR, but if we keep hardcoded width, we might need to have `small` size for these.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+  height: 100%;
+  box-sizing: border-box;
+  padding: $iui-s;
+  position: relative;
+  z-index: 1;
+  border: 1px solid rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-4));
```
veekeys, line 14: Not sure whats the process now, but I thought we agreed to have IE support drop for version 1 only.
So these could be better wrapped under `themed` mixin, instead of using css vars directly.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+  background-color: var(--iui-color-background-1);
+  @media (prefers-reduced-motion: no-preference) {
+    transition: border-color $iui-speed-fast ease-out;
+  }
+
+  // Rounded corners on first tile
```
veekeys, line 20: No need these comments. Code is self-explanatory.
Best practices are to have readable code, so you dont need to add comments.
So the only case for comments when the code is not obvious what is solving or you are having something more complex in there where comments might help.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+
+  // All tiles shift 1px left so we don't have 2px thick dividers
+  &:not(:first-of-type) {
+    margin-left: -1px;
+  }
+
+  &-container {
+    display: inline-flex;
+    flex-wrap: wrap;
+    width: fit-content;
```
veekeys, line 38: Is this needed? `inline-flex` should handle width according to the content.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+  &:last-of-type {
+    border-radius: 0 $iui-border-radius $iui-border-radius 0;
+  }
+
+  // All tiles shift 1px left so we don't have 2px thick dividers
+  &:not(:first-of-type) {
+    margin-left: -1px;
+  }
+
+  &-container {
```
veekeys, line 35: Not sure I am a fan of this...
In other compnents we just had separate mixins for separate parts and then separate classes.
`.iui-radio-tile-container { @include iui-radio-tile-container }`
THis allows also for an easier search for that specific class.
Any reason you chose this approach?

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+    font-size: $iui-font-size;
+    font-weight: $iui-font-weight-semibold;
+    line-height: $iui-line-height;
+    text-align: center;
+    color: var(--iui-text-color);
+  }
+
+  &-description {
+    font-size: $iui-font-size-small;
+    line-height: round($iui-baseline * 1.375);
```
veekeys, line 139: wow, `1.375` :D

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+    margin: 0 auto;
+    margin-top: round($iui-baseline * 0.5);
+    margin-bottom: $iui-baseline;
+    fill: var(--iui-icons-color);
+    @media (prefers-reduced-motion: no-preference) {
+      transition: fill $iui-speed-fast ease-out;
+    }
+  }
+
+  &-label {
```
veekeys, line 129: What is the expected overflow behavior? We just let to put what ever length label/description and wrap it?

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+    font-size: $iui-font-size;
+    font-weight: $iui-font-weight-semibold;
+    line-height: $iui-line-height;
+    text-align: center;
+    color: var(--iui-text-color);
+  }
+
+  &-description {
```
veekeys, line 137: Hmm.. A little bit inconsistent here. If we have label, then we should have sublabel.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+      border: 2px solid rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-4));
+    }
+  }
+
+  &-icon {
+    width: $iui-icons-xl;
+    height: $iui-icons-xl;
+    display: block;
+    margin: 0 auto;
+    margin-top: round($iui-baseline * 0.5);
```
veekeys, line 121: hmm.. Wondering, is this margin needed? We already have padding for whole tile.

src/inputs/classes.scss
```diff
@@ -66,6 +66,6 @@
   @include iui-radio;
 }
 
-.iui-radio-tile-container {
-  @include iui-radio-tile;
-}
+// .iui-radio-tile-container {
```
veekeys, line 69: ?

backstop/tests/radio-tile.html
```diff
@@ -0,0 +1,610 @@
...
+    <h1>Radio tile</h1>
+    <hr />
+
+    <section id="demo-default">
+
+      <div class="iui-input-container">
+        <div class="iui-label iui-required">Radio tile label</div>
+
+        <div class="iui-radio-tile-container">
+          <input
```
veekeys, line 62: Not sure about this change.. Previously, we had `input` under `label` and thats how its done in react. Now if you move it out, `id` is required, which means we will need to generate that in react, if user does not pass it...
Any reason you kept it out?

backstop/tests/radio-tile.html
```diff
@@ -0,0 +1,610 @@
...
+              <path
+                d="m12 0a7.98189 7.98189 0 0 0 -6.9688 11.906c.1079.192.221.381.3438.563l6.625 11.531 6.625-11.531c.102-.151.19-.311.281-.469l.063-.094a7.98217 7.98217 0 0 0 -6.969-11.906zm0 4a4 4 0 1 1 -4 4 4.00011 4.00011 0 0 1 4-4z"
+                fill="#e74c3c"
+              />
+              <path
+                d="m12 3a5 5 0 1 0 5 5 5 5 0 0 0 -5-5zm0 2a3 3 0 1 1 -3 3 2.99988 2.99988 0 0 1 3-3z"
+                fill="#c0392b"
+              />
+            </svg>
+
```
veekeys, line 275: These empty lines look weird. Did auto formatter put those like that?

backstop/scenarios/radio-tile.js
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const { scenario } = require('../scenarioHelper');
+
+module.exports = [
+  scenario('default', {
+    selectors: ['#demo-default'],
```
veekeys, line 9: Would be good to check hover, focus too

veekeys (commented):
Highest specificity is 0.3.0 which is accepted and I checked stylelint it passes with rules set to `3` which is good.

src/classes.scss
adamhock: Changed order for radio tile, surface, and skip to content

src/inputs/classes.scss
adamhock: Removed

src/radio-tile/radio-tile.scss
adamhock: Removed these comments

backstop/scenarios/radio.js
```diff
@@ -11,7 +11,4 @@ module.exports = [
     selectors: ['#demo-label'],
     hideSelectors: ['h3'],
   }),
-  scenario('tiles', {
```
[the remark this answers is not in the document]
adamhock: This commit shows the difference in images. The only difference detected is that the checkmark has moved down one pixel

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+      border-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
+
+      > * {
+        fill: var(--iui-icons-color-actionable);
+      }
+    }
+    // Styles checked radio tile
+    &:checked + * {
+      padding: $iui-s - 1;
+      z-index: 3;
```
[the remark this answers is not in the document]
adamhock: The blue border of a `checked` tile with `z-index: 3` goes over the black border of an unchecked hovered tile with `z-index: 2`

image

src/radio-tile/radio-tile.scss
adamhock: I've seen instances in `header`, `menu`, `tabs`, and `tile` that use `-description` in their classes. I don't see any that use `sublabel`

backstop/tests/radio-tile.html
adamhock: Moving `input` outside of `label` essentially reduced the specificity of radio tiles by `0, 1, 0` since everything in the radio tile was being referenced using the `> label` selector.

src/radio-tile/radio-tile.scss
adamhock: Removed

src/radio-tile/radio-tile.scss
veekeys: ahh.. we need to review that. Cause if we go `name-description` pattern, we cannot have same in `react` because `name` is a reserved native property.
`label-sublabel` kinda makes more sense than `label-description` to me.

backstop/tests/radio-tile.html
veekeys: amm, couldnt you skip `label` selector as you have separate classes?
I mean.. Im not sure we can go with this implementation in react for the reason I mentioned above.

src/radio-tile/radio-tile.scss
adamhock: A label/description with multiple words that causes overflow is wrapped according to the tile's width and the tile's height is increased (see the description of the sample radio tiles).

A single word that causes overflow in label/description causes issues.

adamhock: `word-break: break-word` has been added to `iui-radio-tile-label` and `iui-radio-tile-description`

backstop/scenarios/radio-tile.js
adamhock: Added tests for hover and focus. Not sure what this scenario should be called. default? Type default? All?

src/radio-tile/radio-tile.scss
FlyersPh9: @veekeys do you mean you'd like it to align with [this sort of structure](https://github.com/iTwin/iTwinUI/blob/77480a0f6c0019b80724bef10821bd50eb9b2448/src/breadcrumbs/classes.scss)?  It wouldn't make it easier to search for a specific class, but it does keep the mixins smaller and easier to reuse.

veekeys: no, this is like that right now, almost.
Im tlaking more like this:
https://github.com/iTwin/iTwinUI/blob/main/src/date-picker/classes.scss
It does allow easy search for the specific class :)

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+        outline: 2px solid var(--iui-color-foreground-primary);
+        outline-offset: -4px;
+      }
+    }
+
+    + *:hover {
+      z-index: 2;
+      border-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
+
+      > * {
```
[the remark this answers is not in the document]
FlyersPh9: I think it makes sense to replace `> *` with `> svg` because it only sets `fill` which is exclusive to SVG.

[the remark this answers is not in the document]
FlyersPh9: Same for line 68.

backstop/tests/radio-tile.html
adamhock: This has been updated. `<label>` is a parent of `<input>` ; therefore, `<input>` no longer needs an `id`

Check the updated PR description

src/radio-tile/radio-tile.scss
FlyersPh9: After our CSS 1.0 meeting this morning, I think we decided that this PR will be pre-1.0 so the CSS variables will need to go back to the `themed` `t()` variables.

src/radio-tile/radio-tile.scss
FlyersPh9: Ok, I'll update breadcrumbs to align, @adamhock i guess do the same to match.  Should be fairly easy.

src/radio-tile/radio-tile.scss
FlyersPh9: This is unchanged from the original, but yeah, i don't recall how we landed on that value.

src/radio-tile/radio-tile.scss
FlyersPh9: I was considering the structure like this when I worked on the failed header PR because the ID # wasn't really a "description":
```scss
.name {
	.label { … }
	.sublabel { … }
}
```
I do agree that whatever we decide on we should be consistent across all of our components.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../icon/index';
+
+@mixin iui-radio-tile {
+  cursor: pointer;
```
FlyersPh9, line 7:
```suggestion
  cursor: pointer;
  @include themed {
    -webkit-tap-highlight-color: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
  }
```
Can you include this, it's an iOS only thing.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+  &-input {
+    width: 0;
+    height: 0;
+    appearance: none;
+    opacity: 0;
+    position: absolute;
+    @include focus-visible {
+      + * {
+        outline: 2px solid var(--iui-color-foreground-primary);
+        outline-offset: -4px;
```
FlyersPh9, line 53:
```suggestion
        outline-offset: -4px;
        @media (forced-colors: active) {
          outline-color: Highlight;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+    + *:hover {
+      z-index: 2;
+      border-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
+
+      > * {
+        fill: var(--iui-icons-color-actionable);
```
FlyersPh9, line 62:
```suggestion
        fill: var(--iui-icons-color-actionable);
        @media (forced-colors: active) {
          fill: CanvasText;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+    &:checked + * {
+      padding: $iui-s - 1;
+      z-index: 3;
+      border: 2px solid var(--iui-color-foreground-primary);
```
FlyersPh9, line 69:
```suggestion
      border: 2px solid var(--iui-color-foreground-primary);
      @media (forced-colors: active) {
        border-color: Highlight;
      }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+
+      > * {
+        fill: var(--iui-icons-color-primary);
```
FlyersPh9, line 72:
```suggestion
        fill: var(--iui-icons-color-primary);
        @media (forced-colors: active) {
          fill: Highlight;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+
+      &::after {
+        content: '';
+        width: $iui-icons-default;
+        height: $iui-icons-default;
+        position: absolute;
+        top: $iui-s;
+        right: $iui-s;
+        background-color: var(--iui-color-foreground-primary);
+        mask: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path d="M6,14L0,8l2-2l4,4l8-8l2,2L6,14z" /></svg>');
```
FlyersPh9, line 83:
```suggestion
        mask: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path d="M6,14L0,8l2-2l4,4l8-8l2,2L6,14z" /></svg>');
        @media (forced-colors: active) {
          background-color: Highlight;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    &:disabled + * {
+      cursor: not-allowed;
+      z-index: 0;
+      border-color: var(--iui-color-background-disabled);
+      background-color: var(--iui-color-background-disabled);
+
+      &::after {
+        background-color: var(--iui-icons-color-actionable-disabled);
```
FlyersPh9, line 94:
```suggestion
        background-color: var(--iui-icons-color-actionable-disabled);
        @media (forced-colors: active) {
          background-color: GrayText;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+      }
+
+      .iui-radio-tile-icon {
+        fill: var(--iui-icons-color-actionable-disabled);
+        filter: $iui-icons-color-multicolor-disabled;
```
FlyersPh9, line 99:
```suggestion
        filter: $iui-icons-color-multicolor-disabled;
        @media (forced-colors: active) {
          fill: GrayText;
        }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+    width: $iui-icons-xl;
+    height: $iui-icons-xl;
+    display: block;
+    margin: 0 auto;
+    margin-top: round($iui-baseline * 0.5);
+    margin-bottom: $iui-baseline;
+    fill: var(--iui-icons-color);
+    @media (prefers-reduced-motion: no-preference) {
+      transition: fill $iui-speed-fast ease-out;
+    }
```
FlyersPh9, line 124:
```suggestion
    }
    @media (forced-colors: active) {
      fill: CanvasText;
    }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+
+      .iui-radio-tile-label {
+        color: var(--iui-text-color-muted);
+      }
+    }
+
+    // Styles disabled checked radio tile
+    &:disabled:checked + * {
+      z-index: 3;
+      border: 2px solid rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-4));
```
FlyersPh9, line 110:
```suggestion
      border: 2px solid rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-4));
      @media (forced-colors: active) {
        border-color: GrayText;
      }
```
Adds support for `forced-colors`.

src/radio-tile/radio-tile.scss
```diff
@@ -0,0 +1,143 @@
...
+        background-color: var(--iui-icons-color-actionable-disabled);
+      }
+
+      .iui-radio-tile-icon {
+        fill: var(--iui-icons-color-actionable-disabled);
+        filter: $iui-icons-color-multicolor-disabled;
+      }
+
+      .iui-radio-tile-label {
+        color: var(--iui-text-color-muted);
```
FlyersPh9, line 103:
```suggestion
        color: var(--iui-text-color-muted);
        @media (forced-colors: active) {
          color: GrayText;
        }
      }

      .iui-radio-tile-description {
        @media (forced-colors: active) {
          color: GrayText;
        }
```
Adds support for `forced-colors`.

FlyersPh9:
I added a handful of comments adding support for `forced-colors`.  There may be some additional places we want to add support as well, but I think this is a good start.

src/radio-tile/radio-tile.scss
veekeys: I would say change. It is just a rename of class if we agree from now on to go label/sublabel pattern.
Should be do this rename everywhere before 1.0 then too? @FlyersPh9

adamhock: > I like `sublabel` more than `description`
> 
> Should I change `description` to `sublabel` in this PR or is out of scope?

I meant to ask if I should rename this everywhere in this PR such as in `header` , `menu` , etc.

src/radio-tile/radio-tile.scss
adamhock: Added

adamhock:
@FlyersPh9 I have added all of your `forced color` changes

src/radio-tile/radio-tile.scss
adamhock: Mixins added

src/radio-tile/radio-tile.scss
adamhock: `iui-radio-tile-description` changed to `iui-radio-tile-sublabel`

src/radio-tile/radio-tile.scss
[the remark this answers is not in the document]
adamhock: Fixed

src/radio-tile/radio-tile.scss
veekeys: It is not really what I meant :D
I was talking more about idea to remove it, cause not it looks more like a hack to visually align checkmark with the icon.
But I guess doesnt matter for now.

backstop/scenarios/radio-tile.js
veekeys: COuld you take screenshot of whole group instead of single tile? So leave the selector in `actions` just change selector of whole group in `selectors`. Would be easier to see if focus and especially hover visuals are fine.

veekeys (approved):
Looks good.
Approving.
Just test selectors could be updated.

backstop/scenarios/radio-tile.js
adamhock: Fixed

FlyersPh9 (approved):

PR #600, opened 2022-04-26
chore: Improve build script and error handling

### LINKED ISSUE #599: build fails when an empty folder is present in `src`
When an empty folder is present in `src`, the `yarn build` command fails with this error.

```
Error converting files to CSS: TypeError [ERR_INVALID_URL]: Invalid URL
```

This is not ideal when switching between branches that _add_ a new component folder, because `git checkout` will leave behind an empty folder instead of removing it. And there is no indication of where this error occurred, which makes it frustrating.

mayank99:
Added `fs.existsSync` to check if `classes.scss` is present before trying to compile it. Fixes #599

If the compilation does fail for some reason, then the path will be shown.

image

### DIFF
--- a/scripts/generateCss.js
+++ b/scripts/generateCss.js
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 const fs = require('fs');
+const path = require('path');
 const sass = require('sass-embedded');
 const postcss = require('postcss');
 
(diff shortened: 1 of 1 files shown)


adamhock (approved):

FlyersPh9:
I don't know if I should be reviewing this PR, it's outside of my knowledge.  But I like the reasoning behind this change.

bentleyvk (approved):

PR #618, opened 2022-04-28
chore(Table): Fix sorting within demos

FlyersPh9:
https://github.com/iTwin/iTwinUI-react/issues/641 pointed out that the sort arrow is inverted from our standards documentation, and what is shown in macOS and Windows.  The default should be:

**Alphabetically:**
- Arrow points up = A to Z
- Arrow points down = Z to A

**Dates:**
- Arrow points up = oldest to newest
- Arrow points down = newest to oldest

I fixed the demos by replacing the arrow icon & changing some of the text to be sorted correctly.  Note that on the table demos we are sorting the Folders above Files.


mayank99 (approved):
Changes look fine but PR title should not say `fix` as there is no change in the component css. Maybe `chore` or `test`.

FlyersPh9:
> What if we create new icons that are more clear? Because for me those arrows are always not clear whether it is ascending or descending. I really like Excel icons:

I'd rather not use icons like that, we may need to introduce an icon that show 1-9 & 9-1 for a number specific column.  Also from a visual design perspective I like it as minimal as possible.  I'd say if an arrow is good enough for macOS and Windows, they should be good for us as well.

PR #619, opened 2022-04-29
feat(Select): Multi select support

veekeys:
Added multi-select example.
Have a class for container of tags.
Have a class for multi-select tag, which is reusing some of the tag styles.
Separate classes for label and close button too. 

Screen Shot 2022-05-03 at 10 01 51 AM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3715,9 +3715,9 @@ shebang-regex@^3.0.0:
   integrity sha512-7++dFhtcx3353uBaq8DDR4NuxBetBzC7ZQOhmTQInHEd6bSrXdiEyzCvG07Z44UYdLShWUyXt5M/yhz8ekcb1A==
 
 shell-quote@^1.6.1:
-  version "1.7.2"
-  resolved "https://registry.yarnpkg.com/shell-quote/-/shell-quote-1.7.2.tgz#67a7d02c76c9da24f99d20808fcaded0e0e04be2"
-  integrity sha512-mRz/m/JVscCrkMyPqHc/bczi3OQHkLTqXHEFu0zDhK/qfv3UcOA4SVmRCLmos4bhjr9ekVQubj/R7waKapmiQg==
+  version "1.7.3"
+  resolved "https://registry.yarnpkg.com/shell-quote/-/shell-quote-1.7.3.tgz#aa40edac170445b9a431e17bb62c0b881b9c4123"
+  integrity sha512-Vpfqwm4EnqGdlsBFNmHhxhElJYrdfcxPThu+ryKS5J8L/fhAwLazFZtq+S+TWZ9ANj2piSQLGj6NQg+lKPmxrw==
 
 shelljs@^0.8.4:
   version "0.8.5"


src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
+  height: 80%;
+  max-height: $iui-baseline * 3;
+
+  @at-root .iui-small#{&} {
+    .iui-multi-select-tag-label,
+    .iui-multi-select-tag-button {
+      font-size: $iui-font-size-small;
+    }
+
+    .iui-button-icon {
```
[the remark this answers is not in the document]
mayank99: Or even just duplicate those roles in a new `iui-multi-select-tag-button-icon` mixin.

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
+  margin: 0;
+  display: inline-flex;
+  align-items: center;
+  height: 80%;
+  max-height: $iui-baseline * 3;
+
+  @at-root .iui-small#{&} {
+    .iui-multi-select-tag-label,
+    .iui-multi-select-tag-button {
+      font-size: $iui-font-size-small;
```
mayank99, line 90: I feel like we could avoid nesting these classes because `font-size` gets inherited. But I see both of them reset the font-size, so maybe not.

We could still move it into the mixins
image

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
   }
 }
+
+@mixin iui-multi-select-tag-container {
+  position: absolute;
+  inset: 0 $iui-icons-default + $iui-l 0 $iui-sm;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  gap: $iui-xs;
```
mayank99, line 76: might need fallback for flex gap

backstop/tests/select.html
```diff
@@ -348,5 +352,263 @@ <h2>In context</h2>
...
+    <hr />
+
+    <h2>Multi select</h2>
+
+    <section id="demo-multi">
+      <div class="iui-input-with-icon">
+        <div
+          tabindex="0"
+          class="iui-select-button iui-small"
+        > <span class="iui-content">
```
mayank99, line 365: why is formatting messed up?

FlyersPh9:
Not introduced by your PR, but I noticed that the hover state for select the border changes to match the status color (red below):
Screen Shot 2022-04-29 at 4 45 16 PM

But the rest of the inputs don't do this:
Screen Shot 2022-04-29 at 4 46 34 PM
Screen Shot 2022-04-29 at 4 46 44 PM

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
+
+@mixin iui-multi-select-tag {
+  @include iui-tag;
+  margin: 0;
+  display: inline-flex;
+  align-items: center;
+  height: 80%;
+  max-height: $iui-baseline * 3;
+
+  @at-root .iui-small#{&} {
```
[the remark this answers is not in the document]
veekeys: I think it is needed, because this weird selector is generated otherwise
image

Versus
image

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
+  align-items: center;
+  gap: $iui-xs;
+}
+
+@mixin iui-multi-select-tag {
+  @include iui-tag;
+  margin: 0;
+  display: inline-flex;
+  align-items: center;
+  height: 80%;
```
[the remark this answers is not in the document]
veekeys: What would you suggest? padding around is already pretty small.. If I increase height, there would be even worse :/

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
...
+  align-items: center;
+}
+
+@mixin iui-multi-select-tag-button {
+  @include iui-button;
+  @include iui-button-borderless;
+  @include iui-tag-button;
+  padding: 0 $iui-xs;
+  height: 100%;
+  aspect-ratio: 1 / 1;
```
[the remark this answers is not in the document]
veekeys: 84% global support is not enough?
I would need to go through all 3 sizes to set the width, unless you have some smart suggestions?

src/inputs/select.scss
veekeys: amm, you mean add margins in children or what?

src/inputs/select.scss
veekeys: But I guess this selector, which might be messed up, if there are other components with `iui-small`
image

Unless I could write something different?

src/inputs/select.scss
```diff
@@ -64,3 +66,53 @@
     }
   }
 }
+
+@mixin iui-multi-select-tag-container {
+  position: absolute;
+  inset: 0 $iui-icons-default + $iui-l 0 $iui-sm;
```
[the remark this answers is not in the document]
veekeys: added comment

backstop/tests/select.html
veekeys: no idea..
Fixed

src/inputs/select.scss
mayank99: What about `@at-root .iui-select.iui-small`? In the future we can change the modifier to be `iui-select-small` maybe.

src/inputs/select.scss
mayank99: Yup. Because `gap` support in flex is relatively new

src/inputs/select.scss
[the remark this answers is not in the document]
mayank99: Maybe we should use `basic` tags in small select? or would that be too confusing?

/cc @FlyersPh9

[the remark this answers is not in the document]
veekeys: I know we tried invisible tags, looked weird too :D
Basic would not have on hover, I guess.. But I had idea of just displaying text and separate with coma when got tired of tags...

src/inputs/select.scss
veekeys: amm.. So I should add some fancy check, if gap is not suppored add margins? Cause margins go on children, whereas gap goes on parent.

src/inputs/select.scss
veekeys: Yeah, that helps!
pushed

src/inputs/select.scss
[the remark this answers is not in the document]
veekeys: This was changed

src/inputs/select.scss
mayank99:
```suggestion
  gap: $iui-xs;
  > * + * {
    margin-left: $iui-xs;
    @supports (gap: $iui-xs) {
      margin-left: 0;
    }
  }
```

src/inputs/select.scss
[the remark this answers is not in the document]
mayank99: looks like it's not too bad without `aspect-ratio`
image

So we can keep it as is.

src/inputs/select.scss
veekeys: Added.
Needed some cleanup on tags, but working now.

src/inputs/select.scss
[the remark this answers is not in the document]
FlyersPh9: If we used basic tags I'd like it if the user hit the delete key that it would delete the entire tag instead of just 1 character from the tag.  1 benefit to using the basic tag over the other tag would be that we could display more tags because there is no wasted space for the X icon.

src/inputs/select.scss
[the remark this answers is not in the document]
veekeys: Ok. moved to separate class all the logic of `button-icon`

veekeys:
Last question I have is about naming. Is it ok to call it `multi-select-tag` ?
I see MUI called `Chip`.
Antd call it `select-selection-item`.
Maybe we should consider calling it somehow else?

src/inputs/select.scss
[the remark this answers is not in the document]
FlyersPh9: > Have in mind, it is select, not comobobox

That seems weird to me.  If you can select multiple items, maybe you should need to use a combobox?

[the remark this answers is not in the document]
veekeys: hm.... Difference between select and combobox is that combobox has input where you can type, while select just shows list.
So idk.. I guess having multiselect in Select makes sense too :/

FlyersPh9:
How can we handle overflow?  Currently it does this:
Screen Shot 2022-05-05 at 5 41 05 PM

Our tag containers have truncation and wrapping.
Screen Shot 2022-05-05 at 5 41 37 PM

Can we truncate or wrap?

veekeys:
@FlyersPh9 I added scroll if overflow.. So it is still possible to find the tag you want to remove and do it.
Was trying to investigate wrapping option, but because of position absolute of tag container it was complicated to adjust selects height.
I did not add scroll with overlay, cause it went on tags, but with `auto` i do face issue because of aspect ratio of buttons. When overflow happens, scrollbar moves all elements up (aspect ratio moves them also to the left), so I get this situation

image

src/inputs/select.scss
```diff
@@ -64,3 +66,67 @@
...
+
+@mixin iui-multi-select-tag-container {
+  position: absolute;
+  // align wth Select's padding
+  inset: 0 $iui-icons-default + $iui-l 0 $iui-sm;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  gap: $iui-xs;
+  overflow-x: auto;
```
FlyersPh9, line 78:
```suggestion
  overflow-x: auto;
  overflow-y: hidden;
```
Fixes this issue seen on the small select.
Screen Shot 2022-05-16 at 6 43 07 PM

FlyersPh9:
> @FlyersPh9 I added scroll if overflow.. So it is still possible to find the tag you want to remove and do it.
Was trying to investigate wrapping option, but because of position absolute of tag container it was complicated to adjust selects height.
I did not add scroll with overlay, cause it went on tags, but with `auto` i do face issue because of aspect ratio of buttons. When overflow happens, scrollbar moves all elements up (aspect ratio moves them also to the left), so I get this situation
>
> image

I'll be honest, I'm not sure what the best option here is.

`overflow: auto` shrinks the height of the tag to a nearly unusable size when the input is set to small:
Screen Shot 2022-05-16 at 6 51 49 PM

And `overflow: overlay` covers up a large portion of the tag:
Screen Shot 2022-05-16 at 6 52 13 PM

I was looking to see if we could place the scrollbar beneath the body of the `<select>`. I couldn't find anything, and I'm sure that would have looked really ugly anyways.

Another option might be to copy what we do in the user icon list overflow where the last user icon / tag could contain the number of how many tags can't be shown.  The user would need to select / deselect those overflowed options from the menu (extra clicks) but it certainly looks better and might actually be better for the user.
Screen Shot 2022-05-16 at 6 50 28 PM

src/inputs/select.scss
veekeys: added `overflow: hidden`

veekeys:
> > @FlyersPh9 I added scroll if overflow.. So it is still possible to find the tag you want to remove and do it.
> > Was trying to investigate wrapping option, but because of position absolute of tag container it was complicated to adjust selects height.
> > I did not add scroll with overlay, cause it went on tags, but with `auto` i do face issue because of aspect ratio of buttons. When overflow happens, scrollbar moves all elements up (aspect ratio moves them also to the left), so I get this situation
> > image
> 
> I'll be honest, I'm not sure what the best option here is.
> 
> `overflow: auto` shrinks the height of the tag to a nearly unusable size when the input is set to small: Screen Shot 2022-05-16 at 6 51 49 PM
> 
> And `overflow: overlay` covers up a large portion of the tag: Screen Shot 2022-05-16 at 6 52 13 PM
> 
> I was looking to see if we could place the scrollbar beneath the body of the `<select>`. I couldn't find anything, and I'm sure that would have looked really ugly anyways.
> 
> Another option might be to copy what we do in the user icon list overflow where the last user icon / tag could contain the number of how many tags can't be shown. The user would need to select / deselect those overflowed options from the menu (extra clicks) but it certainly looks better and might actually be better for the user. Screen Shot 2022-05-16 at 6 50 28 PM

Ok, removed scroll.
Added example with tag showing overflow.
But functionality should be achieved in react, we have some utils for it.

backstop/assets/multiSelectTag.js
```diff
@@ -0,0 +1,38 @@
...
+      tabindex="0"
+    >
+      <span
+        class="iui-multi-select-tag-label"
+        title="${value}"
+      >
+        ${value}
+      </span>
+      ${
+        dismissible
```
elephantcatdog, line 26: Would it make sense to switch from nondismissable being the default to dismissable being the default? Not sure what the expected default would be from a user's viewpoint.

veekeys: Maybe it would, but I didnt like the prop name in that case, so left as this (or couldnt figure one). I personally prefer to add features through props than to disabling them. In react we add `close` button only if user provides `onRemove` callback, to answer your question about the default.
Maybe it was not clear, but these web components in css repo are just to make our lives easier. Readability is better and we do not need to flood the same html. So thats the only purpose of this, it is not exposed to the users in anyway and it is not something we will need to support actively.

elephantcatdog (approved):

PR #620, opened 2022-04-29
chore(Inputs): Break `/inputs/` into multiple folders.

### LINKED ISSUE #607: [V1] Input files separation
Deliver separate css per component

FlyersPh9:
Break `/inputs/` into multiple folders:
- `/checkbox/`
- `/input/`
- `/label/`
- `/radio/`
- `/select/`

Demo pages, tests, & test images name changes to align with the HTML element:
- text-field >> input
- text-area >> textarea

Closes #607.


src/classes.scss
```diff
@@ -19,7 +20,8 @@
 @import './header/classes';
 @import './icon/classes';
 @import './information-panel/classes';
-@import './inputs/classes';
+@import './label/classes'; // Needs to be above Input
```
FlyersPh9, line 23: If this line is below `/input/` we get this:
Screen Shot 2022-04-29 at 4 38 30 PM

Not a fan that this is no longer alphabetical.  Suggestions?

FlyersPh9:
I'm wondering if it makes sense to introduce a `label.html` & `scenarios/label.js` to test any input with a wrapping label rather than testing those within the individual component pages?

src/classes.scss
veekeys: If order starts to matter, it is wrong. We might have issues of specificty and import order..

Not sure I understand that `label` part at all.. It should be named differently, I think, cause there is a bunch of different mixins starting from container, ending up with certain placement..

Maybe it would make sense to separate those even further?

veekeys: If order starts to matter, it is wrong. We might have issues of specificty and import order..

Not sure I understand that `label` part at all.. It should be named differently, I think, cause there is a bunch of different mixins starting from container, ending up with certain placement..

Maybe it would make sense to separate those even further?

veekeys:
So then this PR should be merged into new branch, not `main`, right?

src/classes.scss
FlyersPh9: I could rename `/label` to `/input-container/` if that's what you're suggesting.

FlyersPh9:
> So then this PR should be merged into new branch, not `main`, right?

I don't think it has to be CSS 1.0.  The new names (Input, Textarea) align with the names of our current React components.

src/classes.scss
bentleyvk: No, we need to investigate what causes issues with having different import order. Because in React we have no idea in what order CSS files will be imported, nor we can control that.

veekeys:
> > So then this PR should be merged into new branch, not `main`, right?
> 
> I don't think it has to be CSS 1.0. The new names (Input, Textarea) align with the names of our current React components.

I dont understand anymore :D You all voted to go a new branch, but now going my suggestion? :D

src/classes.scss
gretanausedaite: `input-with-label` imports `'../label/index';`
classes.scss in Label folder imports ``'../input/index'

Maybe this loop is causing failures?

src/input/classes.scss
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './index';
+
+.iui-input {
+  @include iui-input;
+}
+
+.iui-input-with-icon.iui-input-with-icon {
```
FlyersPh9, line 9: This was the fix to the ordering issue in `classes.scss` discussed [here](https://github.com/iTwin/iTwinUI/pull/620#discussion_r862136220).

src/classes.scss
FlyersPh9: Loop is removed & ordering has been fixed by doing [this](https://github.com/iTwin/iTwinUI/pull/620/files#r865318081).

src/input/classes.scss
veekeys: :(

src/classes.scss
veekeys: bad fix, but...
Can we still rename label to something else? I barely find connection between name and content in there...

FlyersPh9: How about `/input-container/`?

veekeys: I guess its better, but still.. there are 2 mixins about input container.. there are many more for labels, icons, statuses, etc..
Seems like a file with a bunch of stuff which is generic to all the inputs?

src/input/input-with-icon.scss
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
+@import '../label/index';
```
bentleyvk, line 4: Why you need this import?

FlyersPh9: It's needed for `.iui-end-icon`.  I could move that into `/input-container/` if you'd rather.
Screen Shot 2022-05-06 at 1 33 41 PM

src/classes.scss
FlyersPh9: Renamed to `/input-container/`.

bentleyvk: SASS file name should be the same as folder name.

src/input/input-with-icon.scss
bentleyvk: Ugh, I am not sure. `iui-end-icon` is for `Selelect`, `Combobox` and `input-container` has stuff related with container. Maybe it is ok how it is right now.

veekeys (approved):

gretanausedaite (approved):
Looks good to me.

PR #621, opened 2022-05-02
feat(Table): Sticky columns

bentleyvk:
CSS for #572

sticky-columns
- Moved all background and borders logic from rows to cells.
- Left borders only for expanded content since wits content can be anything.
- Added sticky columns to `Additional features` tables (not to all of them).
- Added JS to handle shadow for sticky columns.
- Fixed JS for columns selection, now they actually work.

I recommend to review HTML with this flag turned on in GitHub:
image


src/table/table.scss
```diff
@@ -177,22 +185,35 @@
...
     min-width: 100%;
     display: flex;
-    border: solid 1px transparent;
-    @media (prefers-reduced-motion: no-preference) {
-      transition: border $iui-speed-fast ease-out;
-    }
-    @include themed {
-      border-bottom-color: t(iui-color-background-border);
+
+    &:not(.iui-expanded-content) .iui-cell {
```
bentleyvk, line 189: Not sure what to do with cell inside expanded content.

backstop/tests/table.html
```diff
@@ -165,147 +165,149 @@ <h2>Default</h2>
...
-                  class="iui-button-icon"
-                  aria-hidden="true"
-                ></svg-filter-hollow>
-              </button>
-              <div class="iui-cell-end-icon">
-                <svg-sort-up
-                  class="iui-button-icon iui-sort"
-                  aria-hidden="true"
-                ></svg-sort-up>
+        <div class="iui-table-header-wrapper">
```
bentleyvk, line 168: This wrapper is needed because sticky won't work with `overflow: hidden` that was on `iui-table-header`.

backstop/tests/table.html
```diff
@@ -1775,104 +1781,134 @@ <h2>Additional options</h2>
...
-            <div class="iui-cell iui-actionable iui-sorted">
-              Product
-              <button
-                class="iui-button iui-borderless iui-filter-button"
-                aria-label="Apply filter"
-                type="button"
+        <div class="iui-table-header-wrapper">
+          <div class="iui-table-header">
+            <div class="iui-row">
+              <div
```
bentleyvk, line 1787: I didn't want for first tables to have too much functionality so I added some stuff to these ones.

src/table/table.scss
```diff
@@ -350,12 +386,20 @@
   display: flex;
   flex-grow: 1;
   min-width: $iui-xxl;
-  min-height: $iui-baseline * 5;
+  min-height: $iui-baseline * 5 + 2px; // +2px for border
```
bentleyvk, line 389: This makes inner cell height to remain `55px` because previously it was `55px` and `2px` from borders on the row. It also makes header a bit bigger than it was before but now header row and body rows are the same height.

backstop/tests/table.html
```diff
@@ -3793,32 +4184,112 @@ <h2>Empty state</h2>
...
-        element.indeterminate = true;
-        element.onclick = function () {
-          var checkboxes = document.querySelectorAll(`#${tableId} .iui-row input[type="checkbox"]:not(:disabled)`);
-          for (var checkbox of checkboxes) {
-            checkbox.checked = this.checked;
-          }
-          element.indeterminate = false;
-          element.checked = this.checked;
-        };
+      const setTableHeaderCheckbox = (table) => {
```
bentleyvk, line 4187: This makes row to get `iui-selected` when checkbox is clicked. Also all header checkboxes now selects table body rows.

adamhock (commented):
image

I'm trying my best to reproduce this visual bug consistently, but I notice that some of the shadows get misaligned

src/table/condensed.scss
```diff
@@ -5,8 +5,11 @@
 
 @mixin iui-table-condensed {
   .iui-table-header .iui-cell,
-  .iui-row .iui-cell,
   .iui-paginator {
     min-height: $iui-baseline * 4;
```
mayank99, line 9: so header and paginator don't have border?

src/table/table.scss
```diff
@@ -367,6 +411,33 @@
...
+    bottom: -1px;
+    width: 30px;
+    transition: box-shadow 0.3s;
+    pointer-events: none;
+  }
+
+  .iui-shadow-right {
+    right: 0;
+    transform: translate(100%);
+    box-shadow: inset 10px 0 8px -8px hsla(200, 12%, 0%, 0.25);
```
mayank99, line 432: Shouldn't we try to use our elevation variab;es?

src/table/table.scss
```diff
@@ -367,6 +411,33 @@
...
+    position: sticky;
+    z-index: 1;
+  }
+
+  .iui-shadow-right,
+  .iui-shadow-left {
+    position: absolute;
+    top: -1px;
+    bottom: -1px;
+    width: 30px;
```
mayank99, line 424: Random width, maybe use variable?

src/table/table.scss
```diff
@@ -350,12 +386,20 @@
...
   flex-basis: $iui-xxl;
   position: relative;
   word-break: break-word;
 
+  &:first-of-type {
+    border-left: solid 1px transparent;
+  }
+
+  &:last-of-type {
+    border-right: solid 1px transparent;
```
mayank99, line 400: what does this do?

src/table/table.scss
```diff
@@ -16,23 +16,31 @@
...
+  overflow: hidden;
+  display: flex;
+  flex-shrink: 0;
+}
+
 @mixin iui-table-header {
+  display: flex;
   user-select: none;
-  overflow: hidden;
+  min-width: 100%;
```
mayank99, line 28: Can you explain this width stuff?

src/table/table.scss
```diff
@@ -367,6 +411,33 @@
     flex-basis: $iui-l * 2;
   }
 
+  &.iui-sticky {
+    position: sticky;
+    z-index: 1;
```
mayank99, line 416: Maybe add `isolation: isolate` somewhere so this doesn't leak out.

src/table/condensed.scss
bentleyvk: Yup

src/table/table.scss
```diff
@@ -233,36 +257,47 @@
...
-        border-color: t(iui-color-foreground-primary);
-        background: rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6));
+      .iui-cell {
+        @include themed {
+          border-color: t(iui-color-foreground-primary);
+          background: linear-gradient(
+            rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6)),
+            rgba(t(iui-color-foreground-primary-rgb), t(iui-opacity-6))
+          ),
+            linear-gradient(t(iui-color-background-1), t(iui-color-background-1));
```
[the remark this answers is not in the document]
bentleyvk: This solution looked cleaner. Where we need to animate this?

src/table/table.scss
bentleyvk: Without it:
image

src/table/table.scss
bentleyvk: Where would you suggest on cell or whole table?

mayank99: I think table would make sense because we also have some other other z-indices, like in resizer

src/table/table.scss
mayank99: ah ok I see `iui-row` also has it. But then how was header working before?

src/table/table.scss
FlyersPh9: Our elevation values have the shadow go downwards where here we need them to go left / right.

src/table/table.scss
```diff
@@ -367,6 +411,43 @@
...
+    z-index: 1;
+  }
+
+  .iui-shadow-right,
+  .iui-shadow-left {
+    position: absolute;
+    top: -1px;
+    bottom: -1px;
+    width: 30px;
+    transition: box-shadow 0.3s;
```
FlyersPh9, line 425: This doesn't seem to do anything because the `<div class="iui-shadow-right">` is removed from the DOM.  Transitions would be nice, but if it's too tricky to get working it's fine as is.

src/table/table.scss
bentleyvk: I think it is now needed because of the header wrapper.

src/table/table.scss
bentleyvk: Changed

src/table/table.scss
bentleyvk: Added

src/table/table.scss
bentleyvk: Good catch! Removed

src/table/table.scss
mayank99: we have some random values in [`information-panel`](https://github.com/iTwin/iTwinUI/blob/main/src/information-panel/information-panel.scss#L184) too. Maybe we should align the two?

FlyersPh9: Yeah, that would make sense.  Maybe we could make our drop shadow into a mixin that allows you to set the elevation & direction?

backstop/tests/table.html
```diff
@@ -1775,104 +1781,134 @@ <h2>Additional options</h2>
...
-            </div>
-            <div class="iui-cell iui-actionable">
-              Price
-              <button
-                class="iui-button iui-borderless iui-filter-button"
-                aria-label="Apply filter"
-                type="button"
+              <div
+                class="iui-cell iui-actionable iui-sorted iui-sticky"
+                style="left: 48px;"
```
mayank99, line 1798: How are these inline styles supposed to work? We need to apply them in react too?

What if the user wants the first 3 columns to be sticky? We need to find their widths using JS?

And if we really need to pass those from JS then maybe we should use CSS variables.

src/table/table.scss
```diff
@@ -367,6 +412,42 @@
     flex-basis: $iui-l * 2;
   }
 
+  &.iui-sticky {
+    position: sticky;
+    z-index: 1;
+  }
+
+  .iui-shadow-right,
```
mayank99, line 420: Maybe `iui-shadow-left` and `iui-shadow-right` should be in a separate mixin. There is no reason to nest this css inside cell.

src/table/table.scss
```diff
@@ -337,6 +373,7 @@
     justify-content: center;
     align-items: center;
     flex-grow: 1;
+    width: 100%;
```
mayank99, line 376: what about this one? I don't see any difference when I toggle it off

src/table/table.scss
```diff
@@ -367,6 +412,42 @@
     flex-basis: $iui-l * 2;
   }
 
+  &.iui-sticky {
```
mayank99, line 415: I think we should be more specific here than just `iui-sticky` as we're working on avoiding class name collisions between components.

```suggestion
  &.iui-cell-sticky {
```

src/table/table.scss
```diff
@@ -177,22 +186,35 @@
...
     display: flex;
-    border: solid 1px transparent;
-    @media (prefers-reduced-motion: no-preference) {
-      transition: border $iui-speed-fast ease-out;
-    }
-    @include themed {
-      border-bottom-color: t(iui-color-background-border);
+
+    &:not(.iui-expanded-content) .iui-cell {
+      border-top: solid 1px transparent;
```
mayank99, line 191: Shouldn't all these go in the `iui-table-cell` mixin?

mayank99: Also I feel like we should be able to apply borders to cell for `iui-expanded-content` too, because it's just one big cell.

mayank99 (commented):
Some more comments to help clean up.

Also there are some merge conflicts that need to be resolved.

backstop/tests/table.html
bentleyvk: Yup, those needed to be set through JS. I also thought about using CSS variables but I am not sure what the benefit of using CSS variables here would be.

src/table/table.scss
bentleyvk: `iui-table-cell` mixin is reused for header and body and I only need these rules for body cells.
Only in HTML example we are demoing expanded content that has `iui-cell` inside but in React we are not using it as it can be anything so I don't know how can I add borders to expanded content only when there is no `iui-cell` except for using `:has` which is not fully supported yet. Because otherwise I will get double borders.

src/table/table.scss
bentleyvk: Maybe I should revert header to 55px 😕
image

src/table/table.scss
bentleyvk: But is it needed for a modifier?

backstop/tests/table.html
```diff
@@ -3793,32 +4184,112 @@ <h2>Empty state</h2>
...
+            });
+            scrollableBody.querySelectorAll('.iui-sticky').forEach((sticky) => {
+              const shadow = sticky.querySelector('.iui-shadow-right');
+              if (shadow) {
+                sticky.removeChild(shadow);
+              }
+            });
+          }
+
+          if (this.scrollWidth - this.scrollLeft - this.offsetWidth > 0) {
```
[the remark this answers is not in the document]
bentleyvk: Added

src/table/table.scss
bentleyvk: Moved

backstop/tests/table.html
bentleyvk: Added

src/table/table.scss
bentleyvk: Idk, our shadows feels very different and it would be too much work to make mixin because mine uses some hacks to position inset shadow outside, info-panel has parent with `overflow: hidden`, surface just has a regular outside shadow. Maybe at least we could have some variables for shadow blur.
Sticky shadow:
image
Info-panel:
image
Surface:
image

src/table/table.scss
bentleyvk: Must be a leftover. Removed.

backstop/tests/table.html
mayank99: We try to avoid inline styles in react even if there is no obvious benefit.

Same reason we use `--left` instead of `left` in slider and color-picker

src/table/table.scss
mayank99: nothing is "needed" but it's better to avoid collisions just in case

src/table/table.scss
mayank99: > `iui-table-cell` mixin is reused for header and body and I only need these rules for body cells.

`@at-root iui-table-body iui-row:not(.iui-expanded-content) &` should work, no?

> Only in HTML example we are demoing expanded content that has `iui-cell` inside but in React we are not using it

Can we remove iui-cell from the html example then? Looks confusing.

src/table/table.scss
bentleyvk: Reverted

src/table/table.scss
bentleyvk: > `@at-root iui-table-body iui-row:not(.iui-expanded-content) &` should work, no?

How it is better? Also we have a lot of `.iui-cell` rules that are wrapped with other classes so then we should be consistent and move those to mixin too. And I don't want to do that, too much changes then.

> Can we remove iui-cell from the html example then? Looks confusing.

I think we can. It doesn't feel right to use `iui-cell` in random place. Maybe I should also remove those `:not(.iui-expanded-content)`?

src/table/table.scss
bentleyvk: Changed

backstop/tests/table.html
bentleyvk: Using CSS variables right now

src/table/table.scss
bentleyvk: Ok, sticky cell shadow is a bit different, I tried setting the same blur value as in info-panel but because it is inset shadow I need to adjust spread and then shadows doesn't look the same. I then tried to copy whole shadow from info-panel but then it doesn't look like one shadow because there are whiter spaces:
image
So maybe I will try to make sticky cell shadow to visually look the same as info-panel.

FlyersPh9: I took a look at how you're implementing your shadows… pretty interesting way to make that visually work.  This screenshot shows the sticky table shadow on top & `$iui-elevation-4` but casting to the left rather than down on the bottom.  They match up pretty close.
Screen Shot 2022-05-16 at 6 21 43 PM

I do agree with changing `information-panel.scss` to better match our elevation, something like this for line 184 `box-shadow: -1px 0 10px hsla(200, 12%, 0%, 0.25);` which matches `$iui-elevation-4` but the X & Y swapped.

I'd like to see the sticky columns using something less elevated like `$iui-elevation-2` but that might not be achievable with the way you have the box shadow implemented.

bentleyvk: Tried to make sticky shadow the same as elevation-2 (top sticky, bottom one is with elevation-2):
image
How table looks like:
image

bentleyvk: And for info panel maybe I will just create a separate PR after this one if we want to make shadows similar.

bentleyvk:
@mayank99 @FlyersPh9 is there anything else you want me to fix or can I get an approval?

mayank99 (approved):
Waiting for this but otherwise looks good:
> > Can we remove iui-cell from the html example then? Looks confusing.
>
> I think we can. It doesn't feel right to use iui-cell in random place. Maybe I should also remove those :not(.iui-expanded-content)?

bentleyvk:
> Waiting for this but otherwise looks good:
> 
> > > Can we remove iui-cell from the html example then? Looks confusing.
> > 
> > 
> > I think we can. It doesn't feel right to use iui-cell in random place. Maybe I should also remove those :not(.iui-expanded-content)?

Oh shoot. I missed this one, will do right now.

src/table/table.scss
bentleyvk: Removed `iui-cell` from expanded contents

adamhock (approved):

PR #622, opened 2022-05-02
fix(Table): Table header truncation

### LINKED ISSUE #473: Last row in table when expanded then need to be scrolled
Now :
image

Should be like this when last row expanded

image

### LINKED ISSUE #500: Table: fix column headers in smaller screens

FlyersPh9:
- Table header truncates rather than wrapping to a new line. Must be placed within `<div class="iui-cell-label">`. Closes #500.

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -241,7 +241,6 @@
 
     &.iui-expanded-content {
       overflow: hidden;
-      flex-shrink: 0;
       @include iui-transition-group;
     }
 
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -35,54 +35,62 @@
     border-right: solid 1px transparent;
   }
 
-  .iui-cell:not(.iui-slot) {
-    @include iui-focus;
-
-    &.iui-actionable {
-      cursor: pointer;
+  .iui-cell {
+    &-label {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
     }
 
-    &.iui-grabbing {
-      cursor: grabbing;
-    }
+    &:not(.iui-slot) {
+      @include iui-focus;
 
-    > .iui-filter-button {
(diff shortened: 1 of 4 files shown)


mayank99:
Can you split this into two PRs? I don't think the scroll snapping will be compatible with react-table and our animations, so it would be better to handle and review it separately.

https://github.com/iTwin/iTwinUI/blob/c30031b9d52f0be93a64712dfdfb45be5aa64424/CONTRIBUTING.md?plain=1#L129

src/table/table.scss
```diff
@@ -35,54 +35,62 @@
...
-  .iui-cell:not(.iui-slot) {
-    @include iui-focus;
-
-    &.iui-actionable {
-      cursor: pointer;
+  .iui-cell {
+    &-label {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
```
mayank99, line 42: I don't think this is a proper fix. 
image

What was wrong with `flex-wrap`? I know it wouldn't automagically handle the sort icon, but that was part of the task.

FlyersPh9: I first tried applying `white-space: nowrap; flex-wrap: wrap;` but wasn't very happy with how it was looking.
Screen Shot 2022-05-02 at 12 51 59 PM

FlyersPh9:
> Can you split this into two PRs? I don't think the scroll snapping will be compatible with react-table and our animations, so it would be better to handle and review it separately.
> 
> https://github.com/iTwin/iTwinUI/blob/c30031b9d52f0be93a64712dfdfb45be5aa64424/CONTRIBUTING.md?plain=1#L129

Removed the scroll snapping portion.

src/table/table.scss
bentleyvk: I agree with Mayank, truncated headers doesn't give any info to the user.

PR #623, opened 2022-05-02
feat(Alert): Class name & specificity refactor

### LINKED ISSUE #604: [V1] Alert refactor
- [ ] Specificity max 0.3.0

- [ ]  Dont target `iui-button`, have alert-button component instead

- [ ] sticky alert should use `position: fixed` or  `position: sticky`

adamhock:
- Reduced specificity to `(0, 3, 0)`

- `iui-button` is now `iui-alert-button`
- `iui-button-icon` is now `iui-alert-button-icon`
- the `position` property for `iui-sticky` has been changed from `absolute` to `sticky`

Closes #604

EDIT: 
- `iui-alert iui-status` is now `iui-alert-status` where `status` is `positive`, `negative`, `warning`, and `informational`


src/alert/classes.scss
```diff
@@ -4,10 +4,30 @@
...
+
+.iui-alert-button-icon {
+  @include iui-alert-button-icon;
+}
 
-  @each $status in positive, negative, warning, informational {
-    &.iui-#{$status} {
-      @include iui-alert-category($statusColor: if($status == informational, 'primary', $status));
-    }
+@each $status in positive, negative, warning, informational {
```
adamhock, line 29: I moved this outside of the `.iui-alert` selector to reduce specificity. The selectors that `iui-status` targets (`iui-alert-icon`, `iui-alert-message`, `iui-alert-link`, and `iui-alert-button`) are understood to be part of `iui-alert` anyway.

src/alert/alert.scss
```diff
@@ -16,53 +17,64 @@
...
-    cursor: pointer;
-    margin-left: $iui-s;
-    user-select: none;
-    white-space: nowrap;
-  }
-
-  > .iui-button {
-    margin-left: auto;
-    margin-right: $iui-s;
+  @each $status in positive, negative, warning, informational {
```
adamhock, line 20: This is to generate the `.iui-alert.iui-status` css for each of the alert statuses.

adamhock (commented):
Any mention of `iui-status` in my previous comments refers to `iui-positive`, `iui-negative`, `iui-warning`, and `iui-primary`

src/alert/alert.scss
```diff
@@ -16,53 +17,64 @@
...
+          linear-gradient(t(iui-color-background-1), t(iui-color-background-1));
+        border-color: t(iui-color-foreground-#{$statusColor});
+      }
+    }
   }
 
   &.iui-sticky {
     border-radius: 0;
-    position: absolute;
+    position: fixed;
```
veekeys, line 36: hm... Wondering.. This will always make it fixed relatively to the viewport.
When we say `sticky`, I'd expect possibility to stick it relative to certain element.

src/alert/alert.scss
```diff
@@ -16,53 +17,64 @@
...
+  @include iui-button;
+  @include iui-button-borderless;
+  @include iui-button-size(small, borderless);
+  margin-left: auto;
+  margin-right: $iui-s;
+}
+
+@mixin iui-alert-button-icon {
+  @include iui-button-icon;
+  margin-left: 3px;
```
veekeys, line 70: weird place.. maybe use `$iui-component-offset` variable and add comment, why it was needed.

src/alert/classes.scss
veekeys: Not sure I like it.. It is still risky.
Checking the generated selectors shows that if there is any parent element of alert with any of status classes, alert would get the styling according to that.. 
Maybe instead of having `iui-alert iui-positive`, we could think about having `iui-alert-positive` ?
It is not that we really need to support `iui-alert` just by itself as currently its looking pretty empty (color-wise).. Even though it gives ability to customize it.

src/alert/alert.scss
veekeys: @FlyersPh9 does our design system limit sticky alert to be only in page? or it can be in modal or just any container?

veekeys (commented):
Specificity is not 0.3.0
image

But not sure what to do about our buttons.. maybe see if certain mixins could be dropped out?

src/alert/classes.scss
adamhock: I'll give that a shot

adamhock: This has been changed

src/alert/alert.scss
adamhock: This has been updated

adamhock:
> Specificity is not 0.3.0 image
> 
> But not sure what to do about our buttons.. maybe see if certain mixins could be dropped out?

I believe @FlyersPh9 has started on Button refactoring, so when that specificity is reduced to 0,3,0, the changes should apply here

src/alert/alert.scss
FlyersPh9: The only thing I can find that might be relevant is this 2 year old mockup that shows an alert fixed to the top of the list in the right pane.  If the user were to scroll, the alert would not scroll with it and would stay in place.
Screen Shot 2022-05-03 at 4 21 19 PM

src/alert/alert.scss
```diff
@@ -17,43 +18,53 @@
...
-    user-select: none;
-    white-space: nowrap;
-  }
-
-  > .iui-button {
-    margin-left: auto;
-    margin-right: $iui-s;
-  }
-
   &.iui-sticky {
```
FlyersPh9, line 21: I think it makes sense to make this into it's own mixin and then include it in `classes.scss` like this:
```scss
.iui-alert {
  @include iui-alert;

  &.iui-sticky {
    @include iui-alert-sticky;
  }
}
```
I ran into a similar issue in #568 with the anchor mixin.  That mixin included the `.iui-anchor-external` modifier which I did not need in breadcrumbs & was causing the specificity to go high.

FlyersPh9, line 21: I think it makes sense to make this into it's own mixin and then include it in `classes.scss` like this:
```scss
.iui-alert {
  @include iui-alert;

  &.iui-sticky {
    @include iui-alert-sticky;
  }
}
```
I ran into a similar issue in #568 with the anchor mixin.  That mixin included the `.iui-anchor-external` modifier which I did not need in breadcrumbs & was causing the specificity to go high.

src/alert/alert.scss
```diff
@@ -17,43 +18,53 @@
...
     border-radius: 0;
-    position: absolute;
+    position: fixed;
     top: 0;
     left: 0;
     width: 100vw;
```
FlyersPh9, line 26: Expanding upon the comment above this one, if we do allow an alert nested within another component like a modal, I think this `width: 100vw` may cause issues.  `100%` might make more sense.

veekeys:
Everything looks fine to me, except..
We ended up having 0.5.0 specificity with this one. I think it might be even worse than what we had before :((

src/alert/alert.scss
veekeys: Yeah, so I guess lets use `%` for width OR can test out having `display: flex` with idea that it will stretch.
Then we need to think about `position` here..

veekeys: `sticky` might be an option as it is relative to the first scrolling parent.

src/alert/alert.scss
gretanausedaite: developer.bentley.com has this alert:
image

src/alert/classes.scss
```diff
@@ -4,10 +4,30 @@
...
 .iui-alert {
   @include iui-alert;
+}
 
-  @each $status in positive, negative, warning, informational {
-    &.iui-#{$status} {
-      @include iui-alert-category($statusColor: if($status == informational, 'primary', $status));
-    }
+@each $status in positive, negative, warning, informational {
+  .iui-alert-#{$status} {
```
gretanausedaite, line 10: I like this change!

src/alert/alert.scss
gretanausedaite: Maybe make it as status alert - `iui-alert-sticky`?

src/alert/alert.scss
adamhock: `display: flex` and `position: sticky` have been added here

src/alert/alert.scss
```diff
@@ -17,43 +18,54 @@
...
-
   &.iui-sticky {
     border-radius: 0;
-    position: absolute;
+    position: sticky;
     top: 0;
     left: 0;
-    width: 100vw;
+    display: flex;
+    z-index: 1;
```
adamhock, line 27: image

Came across this issue when the sticky alert was over another alert. It seems `z-index: 1` solves it.

src/alert/alert.scss
adamhock: I've set `position: sticky` so it can stick relative to any container. The visual test has been updated to reflect this

src/alert/alert.scss
veekeys: Not sure we should control that.. Would assume its on user in what container sticky alert is put.. Or I misunderstood the issue

backstop/tests/alert.html
```diff
@@ -23,9 +23,15 @@
       href="../assets/demo.css"
     />
     <style>
-      body,
-      theme-button {
-        margin-top: 44px; /* Height of sticky alert*/
+      #stickyContainer {
```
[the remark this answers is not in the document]
veekeys: amm.. I dont see that space on my end
image

src/alert/alert.scss
veekeys: `display: flex` is not needed, cause its coming from `iui-alert`

backstop/tests/alert.html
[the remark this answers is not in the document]
adamhock: It's pretty inconsistent, When I had both the old and the new sticky alerts on the HTML page, I would get this issue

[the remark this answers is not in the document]
veekeys: amm.. can you put together then sticky on the whole page and leave this exmple too? If you see something, you might be up to something in there. We need to be sure, there is nothing wrong with our css :/

src/alert/alert.scss
adamhock: Removed

src/alert/alert.scss
adamhock: I agree. Removed

backstop/tests/alert.html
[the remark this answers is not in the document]
adamhock: Open the gif in a new tab to clearly see the changes

AlertResizeBorder

So the double border appears and disappears on resizing? Changing the border thickness of the container to `0.5px` fixes this issue.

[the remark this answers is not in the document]
adamhock: I changed the border thickness of the container. I'm open to feedback on if there's a better way to fix this

src/alert/alert.scss
```diff
@@ -17,43 +18,52 @@
     background-color: t(iui-color-background-1);
   }
```
FlyersPh9, line 19:
```suggestion
  }
  @media (forced-colors: active) {
    background-color: Canvas;
  }
```
Adds `forced-color` support.

FlyersPh9 (approved):
Once [this](https://github.com/iTwin/iTwinUI/pull/623#discussion_r867017832) is merged in, looks good to go.

PR #626, opened 2022-05-05
feat(Table): Add container over column actions for wrapping

veekeys:
Would something like this work out?
table


src/table/table.scss
```diff
@@ -155,6 +156,12 @@
   }
 }
 
+@mixin iui-table-header-actions-container {
```
adamhock, line 159: Does this mean the filter icon will be shown all the time instead of just on hover?

backstop/tests/table.html
```diff
@@ -184,21 +184,23 @@ <h2>Default</h2>
...
-                  class="iui-button-icon"
-                  aria-hidden="true"
-                ></svg-filter-hollow>
-              </button>
-              <div class="iui-cell-end-icon">
-                <svg-sort-up
-                  class="iui-button-icon iui-sort"
-                  aria-hidden="true"
-                ></svg-sort-up>
+              <div class="iui-table-header-actions-container">
```
adamhock, line 187: Visual Tests 😄

src/table/table.scss
adamhock, line 159: Also make sure this is applied in `information-panel.html` and `file-upload.html`

veekeys: Yeah, good catch.. firstly want to verify, if we go this way.

bentleyvk:
Maybe we could switch filter to small button because this misalignment bings pain to my eyes:
image

src/table/table.scss
veekeys: no, it was a bug :D 
fixed

backstop/tests/table.html
veekeys: fixed with the selector fix.

veekeys:
> Maybe we could switch filter to small button because this misalignment bings pain to my eyes: image

Did something, looks like that now
image

FlyersPh9 (approved):
Once you apply the changes to `file-upload` and `information-panel` [like Adam pointed out](https://github.com/iTwin/iTwinUI/pull/626#discussion_r866230097) it should be good to go.

adamhock (approved):
Approved once file upload and information panel are updated. I like this approach

src/table/table.scss
veekeys: done

src/table/table.scss
```diff
@@ -155,6 +156,12 @@
...
+  display: flex;
+  flex-grow: 1;
```
bentleyvk, line 161: Why this one is needed?

veekeys: Without it does not take full width
image

PR #632, opened 2022-05-10
chore: Clean up z-index values

adamhock:
These are the following components I looked at when modifying z-index values

**button-group.scss**
- z-index values are used for prioritizing overlapping borders for disabled, hovered, and focused buttons
- **Solution:** I left this alone.

**file-upload.scss**
- z-index value was used so that `cursor: pointer` is applied to `iui-browse-input`, but the size of the element is really small and I'm not sure why it isn't `0 x 0`.
- **Solution:** Removed z-index.

**information-panel.scss**
- z-index was used to make sure the information panel was in front of the column resizer. However, since the information panel comes after the table in the HTML I don't see a need for it.
- **Solution:** Removed z-index for both information panel and table resizer.

**radio-tile.scss**
- Similar to button group, z index was used to prioritize overlapping borders for disabled, hovered, focused, and checked radio tiles.
- **Solution:** Decremented all z-index values by 1 and added `isolation: isolate` to `iui-radio-tile-container`.

**slider.scss**
- z-index is used to keep `iui-slider-thumb` above `iui-slider-ticks`. I tried fixing this by placing the `iui-slider-ticks` before the `iui-slider-thumb` in the HTML, but that caused issues with the `iui-tooltip` positioning.
- **Solution:** Left z-index alone, but added `isolation: isolate` to the slider container.

**table.scss**
- z-index was used for the table resizer.
- **Solution:** Removed z-index.

**tile.scss**
- z-index was used on thumbnail icons for tiles which already had a thumbnail.
- **Solution:** Removed z-index.

**toggle-switch.scss**
- z-index is used for hiding an icon when unchecking a switch. `iui-toggle-switch-icon` is initially set to `z-index: 1`, but `iui-toggle-switch` gets switched to `z-index: 2` while it's being unchecked.
- **Solution:** Decremented each z-index value by 1.

### DIFF
--- a/src/file-upload/file-upload.scss
+++ b/src/file-upload/file-upload.scss
@@ -36,7 +36,6 @@
       opacity: 0;
       overflow: hidden;
       position: absolute;
-      z-index: -1;
     }
 
     > svg,
--- a/src/information-panel/information-panel.scss
+++ b/src/information-panel/information-panel.scss
@@ -17,7 +17,6 @@
   box-sizing: border-box;
   max-width: 100%;
   max-height: 100%;
-  z-index: 2; // Needs to be higher than table's column resizers.
   @media (prefers-reduced-motion: no-preference) {
     transition: visibility $iui-speed-instant $iui-speed-fast ease-in, transform $iui-speed-fast ease-out,
       opacity $iui-speed-fast ease;
--- a/src/radio-tile/radio-tile.scss
+++ b/src/radio-tile/radio-tile.scss
@@ -27,6 +27,7 @@
   box-sizing: border-box;
   padding: $iui-s;
   position: relative;
+  z-index: 1;
   @media (prefers-reduced-motion: no-preference) {
     transition: border-color $iui-speed-fast ease-out;
   }
@@ -36,7 +37,7 @@
   }
 
   &:hover {
-    z-index: 1;
+    z-index: 2;
     @include themed {
       border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
     }
@@ -81,7 +82,7 @@
 
   &:checked + * {
     padding: $iui-s - 1;
-    z-index: 2;
+    z-index: 3;
     @media (forced-colors: active) {
       border-color: Highlight;
     }
@@ -117,7 +118,7 @@
 
   &:disabled + * {
     cursor: not-allowed;
-    z-index: -1;
+    z-index: 0;
     @include themed {
       border-color: t(iui-color-background-disabled);
       background-color: t(iui-color-background-disabled);
@@ -157,7 +158,7 @@
   }
 
   &:disabled:checked + * {
-    z-index: 2;
+    z-index: 3;
     @media (forced-colors: active) {
       border-color: GrayText;
     }
--- a/src/radio-tile/radio-tile.scss
+++ b/src/radio-tile/radio-tile.scss
@@ -27,7 +27,6 @@
   box-sizing: border-box;
   padding: $iui-s;
   position: relative;
-  z-index: 1;
   @media (prefers-reduced-motion: no-preference) {
     transition: border-color $iui-speed-fast ease-out;
   }
@@ -37,7 +36,7 @@
   }
 
   &:hover {
-    z-index: 2;
+    z-index: 1;
     @include themed {
       border-color: rgba(t(iui-color-foreground-body-rgb), t(iui-opacity-2));
     }
@@ -59,7 +58,7 @@
   user-select: none;
   // Create stacking context
   position: relative;
-  z-index: 0;
+  isolation: isolate;
 }
 
 @mixin iui-radio-tile-input {
@@ -82,7 +81,7 @@
 
   &:checked + * {
     padding: $iui-s - 1;
-    z-index: 3;
+    z-index: 2;
     @media (forced-colors: active) {
       border-color: Highlight;
     }
@@ -118,7 +117,7 @@
 
   &:disabled + * {
     cursor: not-allowed;
-    z-index: 0;
+    z-index: -1;
     @include themed {
       border-color: t(iui-color-background-disabled);
       background-color: t(iui-color-background-disabled);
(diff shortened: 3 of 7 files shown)


src/table/table.scss
```diff
@@ -64,7 +64,6 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
-      z-index: 1;
```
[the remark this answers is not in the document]
FlyersPh9: If you add `isolation: isolate;` to one of the parent elements (I tested with `.iui-table-header`) it should still be safe to remove the `z-index` from information panel.

src/radio-tile/radio-tile.scss
```diff
@@ -59,7 +58,7 @@
   user-select: none;
   // Create stacking context
   position: relative;
-  z-index: 0;
+  isolation: isolate;
```
mayank99, line 61: If there's isolation: isolate, then none of the other z-indexes in this file need to be changed

src/table/table.scss
mayank99, line 67: This needs to be reverted because the resizer needs to show up over the next column. Without this, the draggable area would become smaller.

src/toggle-switch/toggle-switch.scss
```diff
@@ -85,7 +85,7 @@
     aspect-ratio: 1 / 1;
     border-radius: 50%;
     transition: background-color $iui-speed-fast ease-out, opacity $iui-speed-fast ease-out;
-    z-index: 2;
+    z-index: 1;
```
mayank99, line 88: There is already `isolation: isolate` in this file, so these can be reverted.

src/tile/tile.scss
```diff
@@ -197,7 +197,6 @@
   }
 
   ~ .iui-thumbnail-icon {
-    z-index: 1;
```
mayank99, line 200: I think this was added for an edge case. @FlyersPh9 Do you remember?

src/file-upload/file-upload.scss
```diff
@@ -36,7 +36,6 @@
       opacity: 0;
       overflow: hidden;
       position: absolute;
-      z-index: -1;
```
FlyersPh9, line 39: In the PR description you wrote:
> the element is really small and I'm not sure why it isn't 0 x 0.

If I remember correctly this was an accessibility or SEO based decision.  I tried to find the original source that helped us make our decision but [this is the best I am finding](https://kittygiraudel.com/2021/02/17/hiding-content-responsibly/#the-width-0-and-height-0-declarations) so far.

We may be able to replace lines 34-38 with `@include visually-hidden;` instead.  Pretty sure that mixin did not exist when we created the file upload component.

src/information-panel/information-panel.scss
```diff
@@ -17,7 +17,6 @@
   box-sizing: border-box;
   max-width: 100%;
   max-height: 100%;
-  z-index: 2; // Needs to be higher than table's column resizers.
```
FlyersPh9, line 20: Good to remove as long as we [isolate within the header](https://github.com/iTwin/iTwinUI/pull/632/files#r869425199).

src/tile/tile.scss
FlyersPh9: I thought that this had to do with a tile that has a placeholder icon & a thumbnail present.  I quickly scanned `tile.html` and didn't see any demos that are testing that scenario.  I'm thinking we may be able to remove the `z-index` as long as the HTML is in this order (`.iui-thumbnail-icon` placed before `.iui-tile-thumbnail-picture`):
```html
<div class="iui-tile-thumbnail">
  <svg-imodel-hollow
    class="iui-thumbnail-icon"
    aria-hidden="true"
  ></svg-imodel-hollow>
  <div class="iui-tile-thumbnail-picture demo-photo"></div>
  <button
    aria-label="Info"
    type="button"
    class="iui-button iui-borderless iui-small iui-tile-thumbnail-quick-action"
  >
    <svg-info
      class="iui-button-icon"
      aria-hidden="true"
    ></svg-info>
  </button>
  <div class="iui-tile-thumbnail-badge-container">
    <div
      class="iui-badge"
      style="background-color: hsl(2, 90%, 83%);"
    >
      Draft
    </div>
  </div>
</div>
```

src/table/table.scss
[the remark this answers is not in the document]
FlyersPh9: Looks like Vilius is adding `isolation: isolate;` in his his sticky column PR: https://github.com/iTwin/iTwinUI/pull/621#discussion_r867648112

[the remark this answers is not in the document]
adamhock: Added isolation to the table and added z-index back to resizer

adamhock: reverted

src/toggle-switch/toggle-switch.scss
adamhock: reverted

mayank99: left behind a typo
image

adamhock: fixed

mayank99 (approved):

src/radio-tile/radio-tile.scss
```diff
@@ -59,7 +59,7 @@
   user-select: none;
   // Create stacking context
   position: relative;
-  z-index: 0;
```
elephantcatdog, line 62: Was the removal of this z-index supposed to be reverted as well?

elephantcatdog (commented):
The PR description was nice and helpful!

src/radio-tile/radio-tile.scss
adamhock: I'm not exactly sure. @FlyersPh9 ?

FlyersPh9: I tried reverting back to `z-index: 0` and tried removing `z-index: 0` without `isolation: isolate;` and I'm not seeing any changes.  Maybe this `z-index` was never needed.

FlyersPh9 (approved):

PR #636, opened 2022-05-11
feat(SideNavigation): Change stripe location

FlyersPh9:
Moving the active stripe from the left to the right for better visual consistency with the header & for better display on mobile devices with a notch.


mayank99:
I'm afraid this change is a no-go from me.

"consistency" is not a strong enough argument because the header is horizontal whereas the side-nav is vertical; you can look at vertical button-group and default tabs as counter-arguments.

And the iPhone notch is an extreme edge case: only applies to some phones, and only on landscape orientation (which is not the default orientation people use their phones in), and side-nav should probably not even be used on mobile so it's an anti-pattern.

veekeys:
> I'm afraid this change is a no-go from me.
> 
> "consistency" is not a strong enough argument because the header is horizontal whereas the side-nav is vertical; you can look at vertical button-group and default tabs as counter-arguments.
> 
> And the iPhone notch is an extreme edge case: only applies to some phones, and only on landscape orientation (which is not the default orientation people use their phones in), and side-nav should probably not even be used on mobile so it's an anti-pattern.

Consistency is the argument here. It is not about vertical vs horizontal. It is about having an `active` marker 'inside'rather than 'outside'.
Header and SideNav are the components we usually expect to be displayed on the edges of the viewport. When this happens, active stripe is literally on the border of your screen, so visibility of it is decreased. We `already` have active stripe on the bottom (or in other words `inside`) of the header button components, whereas SideNav currently more is hidden than actually visible.
I think this change makes a lot of sense and was especially noticeable in layouts demos on mobile devices.

src/side-navigation/side-navigation.scss
```diff
@@ -85,7 +85,7 @@
     justify-content: flex-start;
 
     // Adds stripe to the left of active button
```
adamhock, line 87: Change this comment and it looks good to me

mayank99:
> consistency

image
image

veekeys:
> > consistency
> 
> image image

You didnt read.
Consistent:
image

Inconsistent:
image

Examples of button group, expandable block are again different, cause they will go inside the content, they will never be edge of screen.

mayank99:
> Inconsistent: image

tbh I don't see this as inconsistent because they are different directions. But the "outside" vs "inside" comparison makes sense I guess.

It still looks weird for me, especially when submenu is involved.
image

I think the original reason for the left stripe was so we could omit the right border and have the sidenav-button bleed into the submenu or even the page. Since we are not going for that design anymore, I think this change might be ok.

The inconsistency with tabs and button-group still bothers me. I don't think we can change tabs but maybe we should change the button-group? Although, it could also be placed near the left or right edge of the iTwin viewer and might require different stripe locations depending on which.

Mostly just thinking out loud here, there is no "right answer". I'm interested in @FlyersPh9's thoughts (outside the iPhone notch scenario, which is still a non-issue imo).

veekeys:
> > Inconsistent: image
> 
> tbh I don't see this as inconsistent because they are different directions. But the "outside" vs "inside" comparison makes sense I guess.
> 
> It still looks weird for me, especially when submenu is involved. image
> 
> I think the original reason for the left stripe was so we could omit the right border and have the sidenav-button bleed into its submenu. Since we are not going for that design anymore (and instead rely on changing the icon fill), I think this change might be ok.
> 
> The inconsistency with tabs and button-group still bothers me. I don't think we can change tabs but maybe we should change the button-group? Although, it could also be placed near the left or right edge of the iTwin viewer and might require different stripe locations depending on which.
> 
> Mostly just thinking out loud here, there is no "right answer". I'm interested in @FlyersPh9's thoughts (outside the iPhone notch scenario, which is still a non-issue imo).

I agree sub menu looks weird... Maybe it is worth investigating possibility to mark it as active in other way.. Not sure.
Component is pretty unique, so it is up to us to decide its design. But we need to put into consideration all the use cases and possible issues.

FlyersPh9:
Placing the stripe on the right will help with cameras that cut into the view like the notch or hole punches:
Screen Shot 2022-05-17 at 11 50 45 AM

It will also help screens with rounded corners:
Screen Shot 2022-05-17 at 11 51 34 AM

> It still looks weird for me, especially when submenu is involved.
> I think the original reason for the left stripe was so we could omit the right border and have the sidenav-button bleed into its submenu. Since we are not going for that design anymore (and instead rely on changing the icon fill), I think this change might be ok.

I completely agree, I haven't been in love with the visuals since we decided against the sidenav-button bleed, but I know that we needed to drop the bleed for better visual consistency.  I am open to rethinking the visuals in a future update.

mayank99 (approved):

bentleyvk (approved):

Issue #333, opened 2021-10-11
Reduce specificity and nesting in all components


mayank99:
We need to reduce specificity and nesting across the board (follow up to #55).
 
This issue should be used to track the status of the various components/tasks that need to be worked on:
- [x] #225
- [x] #398
- [x] #462
- [x] #428 
- [x] #598
- [x] #430 
- [x] #432 
- [x] Reduce stylelint max nesting depth to 4

PR #640, opened 2022-05-13
feat(Header): Class name & specificity refactor

FlyersPh9:
Work in progress, don't bother reviewing yet.


backstop/tests/menu.html
```diff
@@ -617,72 +617,87 @@ <h3>Header</h3>
...
-                class="iui-button-icon iui-header-button-icon"
-                aria-hidden="true"
-              ></svg-model>
-              <span class="iui-button-label">
-                <div>Model Beta</div>
-                <div class="iui-description">0x0123456789</div>
-              </span>
-              <svg-caret-down-small
-                class="iui-button-icon"
+          <nav
```
FlyersPh9, line 637: There appears to be an overflow bug somewhere in the breadcrumbs.  I'm trying to figure out where the overflow needs to be applied & then figure out if the change belongs in `header.scss`, `breadcrumbs.scss`, or `@mixin iui-label-with-sublabel`.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+  @media (max-width: 1024px) {
+    display: none;
+  }
+  @at-root .iui-slim & {
+    font-size: $iui-font-size;
+  }
+}
+
+@mixin iui-header-breadcrumb-button-container($active-color: primary) {
+  max-width: 25vw;
```
veekeys, line 71: After you added this container, having breadcrumbs and so many wrappers seem like not needed anymore :/

backstop/tests/header.html
```diff
@@ -235,7 +279,7 @@ <h2>Default</h2>
 
             <!-- Profile dropdown button -->
             <button
-              class="iui-button iui-dropdown iui-borderless"
+              class="iui-button iui-borderless"
```
veekeys, line 282: why removed?

backstop/tests/header.html
```diff
@@ -394,78 +454,110 @@ <h2>Default</h2>
 
       <hr />
 
-      <h2>Slim</h2>
+      <h2>
+        Slim
+        <!-- <label
```
veekeys, line 459: remove commented out code

backstop/tests/header.html
```diff
@@ -804,92 +917,100 @@ <h2>
...
+              <ol class="iui-breadcrumbs-list">
+                <li class="iui-breadcrumbs-item">
+                  <span class="iui-header-breadcrumb-button-container">
+                    <button
+                      class="iui-header-breadcrumb-button"
+                      type="button"
+                      aria-label="Project Alpha"
+                      id="test-button-5"
+                    >
+                      <img
```
veekeys, line 949: so now we support only `img` and `svg` ? div with bacgkround image was not working well?

backstop/tests/header.html
```diff
@@ -104,108 +106,150 @@ <h1>Header</h1>
...
+                </li>
+                <li class="iui-breadcrumbs-item">
+                  <span class="iui-header-breadcrumb-button-container">
+                    <button
+                      class="iui-header-breadcrumb-button"
+                      type="button"
+                      aria-label="Model picker"
+                      aria-expanded="false"
+                    >
+                      <svg-model
```
veekeys, line 187: In test images I see this images background is different. expected?

backstop/tests/menu.html
```diff
@@ -617,72 +617,87 @@ <h3>Header</h3>
...
+                    aria-expanded="false"
+                    id="test-button-1"
+                  >
+                    <img
+                      class="iui-header-breadcrumb-button-icon"
+                      src="../assets/image-test.jpg"
+                      draggable="false"
+                    />
+                    <span class="iui-header-breadcrumb-button-label">
+                      <div class="iui-header-breadcrumb-button-label-label">Project Alpha</div>
```
veekeys, line 657: looking at difference of images... is project alpha misaligned now (text with caret) or it was misaligned before?

src/header/header.scss
```diff
@@ -3,441 +3,61 @@
...
-  // iPhone notch support
-  padding: env(safe-area-inset-top) env(safe-area-inset-right) 0 env(safe-area-inset-left);
+  background-color: var(--iui-color-background-1);
+  border-bottom: $iui-xxs solid var(--iui-color-background-border);
+  color: var(--iui-text-color);
+  padding: env(safe-area-inset-top) env(safe-area-inset-right) 0 env(safe-area-inset-left); // iPhone notch support
+}
+
+@mixin iui-page-header-slim {
   @media (prefers-reduced-motion: no-preference) {
```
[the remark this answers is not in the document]
FlyersPh9: My thinking here was that the `transition` isn't needed unless you have the option to make it slim.  And the transition needs to be applied on `.iui-page-header`, can't be applied on `.iui-slim` or it would only animate when going from slim to thick.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+      }
+    }
+  }
+}
+
+@mixin iui-header-brand-icon {
+  width: $iui-icons-large;
+  height: $iui-icons-large;
+  margin: 0 $iui-side-navigation-icon-margins;
+  fill: var(--iui-icons-color-actionable);
```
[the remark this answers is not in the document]
FlyersPh9: This might be an exception to the rule because it is a brand icon and changing it to the non-actionable fill looks wrong to me.
Screen Shot 2022-05-19 at 11 50 35 AM

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+  @at-root .iui-slim & {
+    font-size: $iui-font-size;
+  }
+}
+
+@mixin iui-header-breadcrumb-button-container($active-color: primary) {
+  max-width: 25vw;
+  transition: background-color $iui-speed-fast ease-out;
+  height: 100%;
+  display: flex;
```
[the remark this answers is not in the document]
FlyersPh9: If I remove it I end up with failing test images.
Screen Shot 2022-05-19 at 12 16 42 PM

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+  }
+  @media (max-width: 1024px) {
+    display: none;
+  }
+  @at-root .iui-slim & {
+    font-size: $iui-font-size;
+  }
+}
+
+@mixin iui-header-breadcrumb-button-container($active-color: primary) {
```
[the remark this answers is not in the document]
FlyersPh9: I don't really have a plan to use it so I'll remove it.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+  &:hover {
+    background-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-6));
+  }
+
+  // Active state
+  &[aria-current] {
+    background-color: rgba(var(--iui-color-foreground-#{$active-color}-rgb), var(--iui-opacity-6));
+    box-shadow: 0 $iui-xxs 0 0 currentColor;
+    color: var(--iui-color-foreground-#{$active-color});
+    pointer-events: none; // Removes hover effect, reduces specificity
```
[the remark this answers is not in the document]
FlyersPh9: Yeah, it's a bit of trickery to keep specificity low.  If the wrapper span is `aria-current` it sets `pointer-events: none;` on the wrapper but then reverts it on the nested buttons.  I'm trying to recall the alternative, but I think it would have been something like `[aria-current]:enabled` to avoid hover effect on disabled state.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+    @media (forced-colors: active) {
+      color: LinkText;
+      border-bottom: 3px solid LinkText;
+    }
+  }
+
+  // Disabled state
+  &.iui-disabled,
```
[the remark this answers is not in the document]
FlyersPh9: For the wrapping container I have `&.iui-disabled, &[aria-disabled='true']` but on the buttons I have those 2 plus `&:disabled`.  I may be wrong but I don't think you can set `:disabled` on a `<span>`.

I did update the `@at-root .iui-disabled {` parts to this:
```scss
@at-root {
  .iui-disabled &,
  :disabled &,
  [aria-disabled='true'] & {
	… 
  }
}
```

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+@mixin iui-header-breadcrumb-button {
+  @include iui-reset;
+  @include iui-focus;
+  display: inline-flex;
+  align-items: center;
+  vertical-align: middle;
+  justify-content: center;
+  height: 100%;
+  border-radius: 0;
+  max-width: 25vw;
```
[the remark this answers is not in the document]
FlyersPh9: That is [what we had set previously](https://github.com/iTwin/iTwinUI/blob/bcb65221a7f6ceadd0b9b4506686f73aa686f1d8/src/header/header.scss#L242).  Maybe it would make sense to switch to `ch` to align with [breadcrumbs](https://github.com/iTwin/iTwinUI/blob/bcb65221a7f6ceadd0b9b4506686f73aa686f1d8/src/breadcrumbs/breadcrumbs.scss#L29)?

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+        border-bottom: 3px solid GrayText;
+      }
+    }
+  }
+}
+
+@mixin iui-header-breadcrumb-button {
+  @include iui-reset;
+  @include iui-focus;
+  display: inline-flex;
```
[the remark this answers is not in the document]
FlyersPh9: Correct, but I did not use `@include iui-button` because it brought styling that was not needed along with it.  We could use the button mixin in the future after we make the mixin more scoped.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,239 @@
...
+
+  // Disabled state
+  &.iui-disabled,
+  &:disabled,
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
+}
+
+@mixin iui-header-breadcrumb-button-dropdown {
```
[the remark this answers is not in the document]
FlyersPh9: Correct, without this class we end up with really wide spacing:
Screen Shot 2022-05-19 at 12 53 55 PM

I could achieve it without this class, but then buttons without a dropdown arrow would have uneven left/right padding.
Screen Shot 2022-05-19 at 12 55 40 PM

src/header/classes.scss
```diff
@@ -1,11 +1,63 @@
...
+
+.iui-header-breadcrumb-button-icon {
+  @include iui-header-breadcrumb-button-icon;
+}
+
+.iui-header-breadcrumb-button-dropdown-icon {
+  @include iui-header-breadcrumb-button-dropdown-icon;
+}
+
+.iui-header-breadcrumb-button-label {
```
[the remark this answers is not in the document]
FlyersPh9: I also don't really like `label-label` & am open to suggestions.

src/style/mixins.scss
```diff
@@ -186,3 +186,83 @@
...
+  $sublabel-font-size: $iui-font-size-small,
+  $label-gap: 0
+) {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  font-size: $label-font-size;
+  text-align: left;
+  overflow: hidden;
+  row-gap: $label-gap;
```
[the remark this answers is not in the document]
FlyersPh9: I was trying to plan ahead for using this mixin on something like the [expandable block](https://itwin.github.io/iTwinUI/backstop/minified/expandable-block.html) where the main label is a larger font size.  We _might_ want to increase the gap in that scenario.

src/style/mixins.scss
```diff
@@ -186,3 +186,83 @@
...
+  &-sublabel {
+    font-size: $sublabel-font-size;
+    opacity: var(--iui-opacity-3);
+
+    @at-root .iui-active & {
+      opacity: var(--iui-opacity-1);
+    }
+
+    @at-root [aria-current] & {
+      opacity: var(--iui-opacity-1);
```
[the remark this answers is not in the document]
FlyersPh9: You can't, I tried that but got an error.
Screen Shot 2022-05-19 at 1 00 18 PM

However I did find another way to pull off the same effect:
```scss
@at-root {
  .iui-active &,
  [aria-current] & {
    …
  }
}
```

src/style/mixins.scss
```diff
@@ -186,3 +186,83 @@
...
+  width: $size;
+  height: $size;
+  background-color: $background-color;
+  border-radius: $border-radius;
+  display: inline-flex;
+  flex-shrink: 0;
+  box-sizing: border-box;
+  padding: $iui-xs;
+  fill: var(--iui-icons-color);
+  color: inherit;
```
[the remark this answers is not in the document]
FlyersPh9: I was setting `color: inherit` because the `<svg>` color was being set using `fill: currentColor;`, but it appears I can remove `color: inherit`.

src/side-navigation/side-navigation.scss
```diff
@@ -4,6 +4,8 @@
 @import '../icon/index';
 @import '../text/index';
 
+$iui-side-navigation-icon-margins: $iui-icons-large;
```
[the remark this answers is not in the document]
FlyersPh9: Maybe I could use CSS variables instead but I'm struggling to figure out the best way to do so.  Setting `--iui-side-navigation-icon-margins: #{$iui-icons-large};` within `side-navigation.scss` means it's not accessible within `header-buttons.scss`.

src/header/header-buttons.scss
FlyersPh9: Yup, didn't even need to change the CSS!  Pushed.

backstop/tests/header.html
FlyersPh9: It doesn't appear to be applying any styling.

backstop/tests/header.html
FlyersPh9: Oops, cleaned up.

backstop/tests/header.html
FlyersPh9: As I understand it the header in React is not using the `<div>` with background image.  That's part of the reason why https://github.com/iTwin/iTwinUI-react/issues/447 wasn't caught, because we weren't testing `` on our HTML demo page.

backstop/tests/header.html
FlyersPh9: Yes, that was intentional.  This passes our contrast scores where previously it did not.

backstop/tests/menu.html
FlyersPh9: **Before:**
Screen Shot 2022-05-19 at 3 51 40 PM

**After:**
Screen Shot 2022-05-19 at 3 52 19 PM

src/header/classes.scss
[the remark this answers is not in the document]
FlyersPh9: How about:
```scss
.iui-header-breadcrumb-button-label {
    &-primary { … }
    &-secondary { … }
}
```

elephantcatdog (commented):
The test images for narrow screens don't look right; it looks like the header is getting cut off and not adjusting at the correct widths.
image
In the browser I get problems too. The width here was about 1100px. There are various widths when this problem occurs.
image

.stylelintrc.json
```diff
@@ -12,7 +12,7 @@
     },
     "no-eol-whitespace": true,
     "function-parentheses-space-inside": "never-single-line",
-    "max-nesting-depth": 5,
+    "max-nesting-depth": 4,
```
elephantcatdog, line 15: What do these do/mean?

src/header/header-buttons.scss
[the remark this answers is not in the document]
veekeys: I think you misunderstood me...
What I had in mind is.. maybe we can tackle just aria attributes and get rid of classes and states selectors?
I mean.. get rid of class for sure, we can replace it with aria attributes.. about state I am not that sure yet..

src/header/header-buttons.scss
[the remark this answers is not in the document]
veekeys: sure, just somehow need not to forget that :D
Maybe add this to buttons item or something..

src/header/classes.scss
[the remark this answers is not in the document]
veekeys: that looks nice, but it is a new thing and we used label and sublabel everywhere else.. so would be nice to be consistent :/
ahh..
Maybe just `-text` as container name for now? And we can reconsider our label and sublabel naming in all components later?
Not sure on this one

src/style/mixins.scss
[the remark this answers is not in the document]
veekeys: just I dont think it works on header examples at all.. so maybe remove now?

src/style/mixins.scss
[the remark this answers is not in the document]
veekeys: Color is applicable only for text color, svg has its own property called `fill`, so safe to remove

.stylelintrc.json
veekeys: https://stylelint.io/user-guide/rules/list/max-nesting-depth/
https://stylelint.io/user-guide/rules/list/selector-max-compound-selectors/

And we have this item to work on these, cause they are connected with our general goal of reducing specificity level too
https://github.com/iTwin/iTwinUI/issues/333

I think we want to end up these rules having value of 3. They will help also for new components/features, so we develop those straight away in a good level.

backstop/tests/header.html
veekeys: oh ,itneresting.. I think we expected to have an arrow there, to show its a dropdown menu, when yo uclick on it..
but ok, it wasnt there before, so no need.

backstop/tests/header.html
veekeys: oh.. it is just an example..
We let user provide any of the following:
- `svg`
- `img`
- `div` with background image

So if it easy to support all these cases, lets have a test care with all of them.

backstop/tests/header.html
```diff
@@ -23,38 +23,19 @@
...
-
-      .demo-side-navigation-icon {
-        width: 24px;
-        height: 24px;
-        fill: var(--iui-icons-color-actionable);
-      }
-
       .demo-searchbar {
-        border-radius: 22px;
+        border-radius: 9999px;
```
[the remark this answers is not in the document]
veekeys: He will not.
`$iui-border-radius-round` is a SASS variable and it cannot be used in html files. And we do not have it as css variable yet.

src/header/header.scss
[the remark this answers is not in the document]
veekeys: and how we choose if we have option to make it slim or not?
Not this mixin is by default included under `iui-page-header`.
So this transition rule is always there.

src/header/header-buttons.scss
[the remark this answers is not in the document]
veekeys: I dont know... Maybe we want to limit whole width of button depending on the width of the header or something..
Dont have a single opinion.
I would have expected more stuff to be reused from breadcrumbs, but I guess not really needed

src/header/header-buttons.scss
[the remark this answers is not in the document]
veekeys: I meant, if class is `dropdown`, it should be used on both, split and doprodwn buttons..
So maybe rename, that this is applicable only on split buttons?

src/header/header-buttons.scss
veekeys: And what was the need of having `ol` with `li`?
Previously we jsut had `nav` and buttons as children..

src/side-navigation/side-navigation.scss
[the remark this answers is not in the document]
veekeys: I dont know..
Later maybe when we have some css variables to be shared between cmopnents we could do that..
Now I would just put the value and maybe add comment that it needs to be aligned with side nav..

Or you can put that variable in some common sass file and use it from there :/

backstop/tests/header.html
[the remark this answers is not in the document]
elephantcatdog: Ah, sorry.

src/header/header-buttons.scss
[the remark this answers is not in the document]
FlyersPh9: I have renamed it from `.iui-header-breadcrumb-button-dropdown` to `.iui-header-breadcrumb-button-split`.

src/header/classes.scss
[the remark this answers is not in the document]
FlyersPh9: I like that, renamed to `-text`.

src/style/mixins.scss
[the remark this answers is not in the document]
FlyersPh9: Ok, I have removed `$label-gap`.

src/header/header-buttons.scss
FlyersPh9: [W3 recommends using `<ol>` and `<li>` for breadcrumbs](https://web.archive.org/web/20220406233822/https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html), so we tried to align.

backstop/tests/header.html
FlyersPh9: Added support for `<div>` with `background-image`.  Will add a test in my next push.

backstop/tests/header.html
```diff
@@ -1083,7 +1140,7 @@ <h2>Animation (Toggle default ⇔ slim)</h2>
         </div>
       </header>
 
-      <div class="demo-side-navigation">
+      <!-- <div class="demo-side-navigation">
```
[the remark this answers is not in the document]
FlyersPh9: Oops, the commented out code has been removed now.

FlyersPh9:
> The test images for narrow screens don't look right; it looks like the header is getting cut off and not adjusting at the correct widths. image In the browser I get problems too. The width here was about 1100px. There are various widths when this problem occurs. image

The overflow issues have been fixed by adding `@mixin iui-page-header-left;`.

backstop/tests/header.html
FlyersPh9: This demo has been updated to show Version Gamma.  This tests against:
- ``
- `<svg>`
- `<svg>` in active state.
- `<div>` with `background-image.

And all can be tested in disabled state as well.

Screen Shot 2022-05-23 at 4 12 08 PM

src/header/header.scss
```diff
@@ -32,6 +32,13 @@
   height: 100%;
 }
 
+@mixin iui-page-header-left {
```
veekeys, line 35: Just applying this to whole header is not enough?

src/header/header-buttons.scss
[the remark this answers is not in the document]
FlyersPh9: Removed `.iui-disabled` and relying on `aria-disabled=true` to style `.iui-header-breadcrumb-button-container`.

src/header/header.scss
```diff
@@ -3,441 +3,61 @@
...
-  }
-}
-
-@mixin iui-header-logo {
-  display: inline-flex;
+@mixin iui-page-header-zone {
+  display: flex;
   align-items: center;
-  white-space: nowrap;
   height: 100%;
```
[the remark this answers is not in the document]
FlyersPh9: Seems like it is not needed, removed.

src/header/classes.scss
```diff
@@ -38,6 +38,16 @@
...
 }
 
+.iui-header-breadcrumbs-list {
+  @include iui-breadcrumbs-list;
+}
+
+.iui-page-header-left,
+.iui-header-breadcrumbs,
+.iui-header-breadcrumbs-list {
+  @include iui-page-header-left;
```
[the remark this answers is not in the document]
FlyersPh9: We were doing [something similar previously](https://github.com/iTwin/iTwinUI/blob/6c94ffbb7e93049dc6d1c81049ddf55e57418c68/src/header/header.scss#L46).  The only real difference is now we have a `<ol>` within the `<nav>` which needs to be included in here as well.

src/header/header.scss
FlyersPh9: `box-sizing`, `margin-bottom`, and `padding-bottom` could be applied to all zones (left, center, right) but if we apply `overflow` on the right zone we get a result that is different from what is currently live.

Correct on the top, incorrect on the bottom.
Screen Shot 2022-05-24 at 2 08 47 PM

src/style/mixins.scss
```diff
@@ -261,8 +264,22 @@
...
+      fill: var(--iui-icons-color-actionable-disabled);
+
+      @media (forced-colors: active) {
+        fill: GrayText;
+      }
+    }
+
     .iui-disabled img#{&},
-    .iui-disabled div#{&} {
+    .iui-disabled div#{&},
```
veekeys, line 278: so these classes also not needed?

FlyersPh9: I'm anticipating this mixin being used in other places (menu for example) which are still using `.iui-disabled`.  But since header is the only one using this mixin at the moment, I'll remove it.

src/header/header.scss
FlyersPh9: Ok, so this is possible:

- Use `:before` to place the 2px bottom border.
- `overflow` still needs to be applied on `.iui-page-header-left`, `.iui-header-breadcrumbs`, & `.iui-header-breadcrumbs-list`.  We could apply `overflow` on `@mixin iui-breadcrumbs` & `@mixin iui-breadcrumbs-list` instead of the header specific ones if we'd like.
- Inset the `box-shadow` of the active breadcrumb button.

However this does re-introduce this issue of text & icons appearing off center.
Screen Shot 2022-05-19 at 3 51 40 PM

That could be fixed by giving ALL buttons within the header `padding-bottom: 2px`.  I'm not sure this is any better than what we have currently.

veekeys (approved):

elephantcatdog (approved):

src/header/classes.scss
```diff
@@ -1,11 +1,73 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import './index';
+@import '../breadcrumbs/index';
 
 .iui-page-header {
   @include iui-page-header;
+  @include iui-page-header-slim;
```
mayank99, line 8: I would suggest moving the modifier out of the mixin and putting it here like this:

```scss
.iui-page-header {
  @include iui-page-header;

  &.iui-slim {
    @include iui-page-header-slim;
  }
}
```

If you make the name more specific, you can even get rid of the `&`
```scss
.iui-page-header {
  @include iui-page-header;
}

.iui-page-header-slim {
  @include iui-page-header-slim;
}
```

src/header/classes.scss
```diff
@@ -1,11 +1,73 @@
...
+.iui-header-brand-icon {
+  @include iui-header-brand-icon;
+}
+
+.iui-header-brand-label {
+  @include iui-header-brand-label;
+}
+
+.iui-header-breadcrumbs {
+  @include iui-breadcrumbs($breadcrumbs-height: 100%);
```
mayank99, line 38: I don't think `classes.scss` should have any logic so I suggest creating a new `@mixin iui-header-breadcrumbs` which handles this.

src/style/mixins.scss
```diff
@@ -186,3 +186,99 @@
     }
   }
 }
+
+@mixin iui-label-with-sublabel($label-font-size: $iui-font-size, $sublabel-font-size: $iui-font-size-small) {
```
mayank99, line 190: I do not think we should be putting these mixins here (especially the `iui-thumbnail-or-icon`) because currently it still feels very specific to header.

And when we want to create a generic solution, we might want to make it a utility component similar to `iui-surface`.

src/header/header-buttons.scss
```diff
@@ -0,0 +1,230 @@
...
+  width: $iui-icons-large;
+  height: $iui-icons-large;
+  margin: 0 $iui-side-navigation-icon-margins;
+  fill: var(--iui-icons-color-actionable);
+  will-change: transform;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: transform $iui-speed-fast ease-out;
+  }
+  @at-root .iui-slim & {
+    transform: scale(0.66666);
```
mayank99, line 55: I think this can be better handled through CSS variables so that you don't need to reference `iui-slim` inside the mixin. The css variable can be called something like `--_iui-header-brand-icon-scale` and the value can be changed inside `@mixin iui-page-header-slim`

src/header/header-buttons.scss
```diff
@@ -0,0 +1,230 @@
...
+  margin-right: $iui-sm;
+  font-size: $iui-font-size-leading;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: font-size $iui-speed-fast ease-out;
+  }
+  @media (max-width: 991px) {
+    display: none;
+  }
+  @at-root .iui-slim & {
+    font-size: $iui-font-size;
```
mayank99, line 69: Same comment here about CSS var

src/header/header-buttons.scss
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+
+@mixin iui-header-brand-label {
+  margin-right: $iui-sm;
+  font-size: $iui-font-size-leading;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: font-size $iui-speed-fast ease-out;
+  }
+  @media (max-width: 991px) {
```
mayank99, line 65: Why this random magic number?

src/header/header-buttons.scss
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+}
+
+@mixin iui-header-breadcrumb-button-text {
+  @include iui-label-with-sublabel;
+
+  &-sublabel {
+    @at-root .iui-slim & {
+      display: none;
```
mayank99, line 214: I think we were smoothly transitioning the height before, but now it abruptly disappears.

src/header/header-buttons.scss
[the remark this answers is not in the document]
mayank99: This is dangerous as it will break cursor (pointer/not-allowed, etc) and also break any javascript events that get applied on the `li`.

Plus I'm not seeing any hover effect even when I remove these `pointer-events` declarations, so I'm not sure it's even needed.

src/header/classes.scss
```diff
@@ -1,11 +1,73 @@
...
+
+.iui-header-breadcrumb-button {
+  @include iui-header-breadcrumb-button;
+}
+
+.iui-header-breadcrumb-button-split {
+  @include iui-header-breadcrumb-button-split;
+}
+
+.iui-header-breadcrumb-button-container {
```
mayank99, line 59: nit: I think this makes more sense as a `-wrapper` than a `-container`.

I like to think of container as something that has multiple instances of something else. e.g. tile container holds multiple tiles. But in this case it is just applied to every `li`.

src/header/header-buttons.scss
FlyersPh9: This was changed to align with [the breakpoints in iTwinUI-layouts](https://github.com/iTwin/iTwinUI-layouts/blob/9661022753480fc4901ffc59ca1cf08dd44dd160/packages/layouts-css/src/breakpoints.scss#L16).

src/header/classes.scss
FlyersPh9: I like that differentiator between wrapper and container, I've made the change.

src/header/classes.scss
FlyersPh9: Can do, change made.

src/header/classes.scss
FlyersPh9: @veekeys had a similar comment.  [This was my reasoning behind it](https://github.com/iTwin/iTwinUI/pull/640#discussion_r877207632), but now that we've had 2 comments on the same thing I'll make the change.

src/header/header-buttons.scss
FlyersPh9: Great suggestion!  I need to reset my thinking, I'm used to not being able to use CSS variables like this.

src/style/mixins.scss
FlyersPh9: I removed these 2 universal mixins and placed them in `header-buttons.scss`.  We can revisit making these parts more easily used in other components in a separate PR.

src/header/header-buttons.scss
FlyersPh9: Transition effect is back in place.

veekeys:
There is a bunch of stuff changed, would be nice if you could summarize all of that and leave a comment in V1 branch after merge (classes, html structure changes, etc)

packages/itwinui-css/src/header/header-buttons.scss
```diff
@@ -0,0 +1,310 @@
...
+  // Disabled state
+  &:disabled,
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
+}
+
+@mixin iui-header-breadcrumb-button-split {
+  padding: 0 $iui-s;
+  margin-left: -$iui-s;
```
FlyersPh9, line 162: @mayank99 lines 161-162 allow us to get the same split button width as previously without needing to introduce any additional classes or increasing specificity.  Are you ok with this?

packages/itwinui-css/backstop/tests/menu.html
```diff
@@ -617,72 +617,87 @@ <h3>Header</h3>
...
+                class="iui-breadcrumbs-separator"
                 aria-hidden="true"
-              ></svg-caret-down-small>
-            </button>
+              >
+                <svg-chevron-right aria-hidden="true"></svg-chevron-right>
+              </li>
+              <li class="iui-breadcrumbs-item">
+                <span
+                  class="iui-header-breadcrumb-button-wrapper"
```
mayank99, line 675: I am seeing some inconsistency in the various html files. Can you verify and document the correct html structure?

mayank99 (commented):
Looks good

packages/itwinui-css/backstop/tests/menu.html
FlyersPh9: Good catch, I've updated `skip-to-content.html` & `menu.html` to match the latest HTML structure in `header.html`.

adamhock:
> Looks good but I'm seeing some inconsistencies in dom structure and not sure if it will work in react, particularly the split-button part. We should verify it before merging.

I see the inconsistency that Mayank is talking about in react.

This PR handles split buttons for breadcrumbs by wrapping the two `<button>` elements in an `<li>` due to its parent being an `<ol>`. The current implementation for all split buttons is to wrap each button part individually in a `<div>` and then wrapping those two `<div>`s in a `<span>`.

I believe this refactor would call for a new React component `HeaderBreadcrumbSplitButton` so that these split buttons could follow the `<ol> <li> <button>` structure.

This could be a bit excessive, but if the following structure is used for header breadcrumb split buttons, we can still use `<ol>` while keeping the DOM structure consistent with split buttons without having to create a new component. Thoughts?

`<ol>`
`____<li>`
`________<span>`
`____________<div>`
`________________<button>`
`____________</div>`
`____________<div>`
`________________<button>`
`____________</div>`
`________</span>`
`____</li>`
`</ol>`

packages/itwinui-css/src/button/split-button.scss
```diff
@@ -10,12 +10,6 @@
   }
 
   > * {
-    &:first-child {
```
FlyersPh9, line 13: This is no longer needed for the header split button.

mayank99:
I just realized we are using divs inside spans inside buttons. This markup is considered invalid. https://validator.w3.org/nu/#textarea


We should change the divs to be spans and add `display: block` to them.

FlyersPh9:
> I just realized we are using divs inside spans inside buttons. This markup is considered invalid. [validator.w3.org/nu/#textarea](https://validator.w3.org/nu/#textarea) 
> 
> We should change the divs to be spans and add `display: block` to them.

Good catch, I wasn't even aware that was invalid.  I've swapped them over to `<span>`, no CSS changes seem to be needed.

mayank99:
@adamhock Can you please verify and leave a comment about whether react works fine with the latest changes?

adamhock:
React works fine with the latest changes. My only suggestion would be to change the name of `iui-header-breadcrumbs-item` to `iui-header-breadcrumb-item` since it's only one item, unlike `iui-header-breadcrumbs-list`

PR #641, opened 2022-05-16
chore: Rename Components

elephantcatdog:
Renamed components:
- ToastNotification to Toast
- UserIcon to Avatar
  - custom html component `user-icon` renamed to `x-avatar`
- Wizard to Stepper

Replaced tests with files with new names.


backstop/tests/stepper.html
```diff
@@ -0,0 +1,199 @@
+<!--
```
bentleyvk, line 1: Have you tried using `git mv` command so that it wouldn't show this as a new file?

elephantcatdog: Made a new PR: https://github.com/iTwin/iTwinUI/pull/642

PR #646, opened 2022-05-23
fix(Table): Wrap header actions conditionally

bentleyvk:
If you don't like this approach then I will just close the PR.

This approach uses JS to handle header column actions placement - when there is enough space so header label and actions are shown in one line, then filter icon is aligned to the left, but if there is not enough space and column content wraps, then action icons are placed to the right and placed in column order.
table-actions-fit-js


src/table/table.scss
```diff
@@ -47,8 +47,8 @@
 
   .iui-cell:not(.iui-slot) {
     @include iui-focus;
-    flex-wrap: wrap;
     column-gap: $iui-xs;
+    padding-block: $iui-baseline * 0.5;
```
veekeys, line 51: <90% support is enough?

backstop/tests/table.html
```diff
@@ -4396,5 +4396,29 @@ <h2>Empty state</h2>
...
       };
     </script>
+
+    <script>
+      const setMargin = (actionsContainer) => {
+        const filterButton = actionsContainer.querySelector('.iui-filter-button');
+        if (!filterButton) {
+          return;
+        }
+        if (actionsContainer.clientHeight > 28) {
```
veekeys, line 4406: Would need to check about that height (for different table sizes?)
Maybe we can think of some other identifier...

veekeys (commented):
Visually I think its nice.
Might be no other way to make it good without using JS...

mayank99:
Why do we need all these JS hacks? I removed all the JS and added `margin-left: auto` sort icon and it seems to work just fine.

In this video I'm showing three cases:
1. current JS version, no changes
2. no JS, no CSS changes
3. no JS, with `margin-left: auto` on the sort icon.

1 and 3 seem to behave identically.

https://user-images.githubusercontent.com/9084735/170090169-4d5bc536-ce30-4fba-9e48-9de5c821d4e9.mov

backstop/tests/table.html
bentleyvk: Removed scripts

src/table/table.scss
bentleyvk: Idk, most of the browsers have this since last year, some of them even since 2020.

veekeys (approved):

PR #647, opened 2022-05-23
fix(Table): Allows to resize table header as before

bentleyvk:
My sticky columns PR apparently broke Table resizing that after some point header wouldn't reduce in size while body still can:
table-header-resize-bug

### DIFF
--- a/src/table/table.scss
+++ b/src/table/table.scss
@@ -36,6 +36,7 @@
   .iui-row {
     display: flex;
     flex-grow: 1;
+    min-width: 100%;
   }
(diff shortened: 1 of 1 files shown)


src/table/table.scss
```diff
@@ -36,6 +36,7 @@
   .iui-row {
     display: flex;
     flex-grow: 1;
+    min-width: 100%;
```
bentleyvk, line 39: Table body also has this.

mayank99 (approved):

PR #650, opened 2022-05-23
feat(GDPRBanner): add GDPR Banner

### LINKED ISSUE #637: Create GDPR banner
We should offer a GDPR banner component.  Something like this that attaches to the bottom of the page:
Screen Shot 2022-05-11 at 5 08 44 PM

marashad001:
- Created GDPR banner. Closes #637 
- Reused dialog mixins where applicable


backstop/tests/gdpr-banner.html
```diff
@@ -0,0 +1,78 @@
...
+          >
+            <svg-close
+              class="iui-button-icon"
+              aria-hidden="true"
+            ></svg-close>
+          </button>
+        </div>
+        <div class="iui-gdpr-banner-content">
+          <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
FlyersPh9, line 56: Please replace this with the text in #637.  For the anchor link you will need to use `<a class="iui-anchor" href="#">`.

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,35 @@
...
+
+@mixin iui-gdpr-banner-content {
+  @include iui-dialog-content;
+}
+
+@mixin iui-gdpr-banner-button-bar {
+  @include iui-dialog-button-bar;
+}
+
+.iui-gdpr-banner-cookie-policy {
```
FlyersPh9, line 33: Within the `{{ ComponentName }}.scss` files, all CSS must be within a `@mixin` and then that mixin is included in `classes.scss`.  Here's an example:

**gdpr-banner.scss:**
```scss
@mixin iui-gdpr-banner-cookie-policy {
	color: var(--iui-color-foreground-primary);
}
```

**classes.scss:**
```scss
.iui-gdpr-banner-cookie-policy {
	@include iui-gdpr-banner-cookie-policy;
}
```

However, if you set `<a class="iui-anchor">` then none of this is necessary.

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../dialog/index';
+
+@mixin iui-gdpr-banner {
+  @include iui-dialog;
+  border-radius: 0;
+  box-shadow: none;
+  background-color: var(--iui-color-background-5);
```
FlyersPh9, line 10: Let's make a slight visual change to increase the contrast score of the text against the background. 

```suggestion
  background-color: var(--iui-color-background-3);
  border-top: 1px solid var(--iui-color-background-border);
```

Because the background color is getting lighter with this change, I think we need to make the division from the body content and the GDPR banner more prominent.  Adding `border-top: 1px solid var(--iui-color-background-border);` should help with that.

**Results after this change:**
Screen Shot 2022-05-25 at 1 49 06 PM
Screen Shot 2022-05-25 at 1 49 36 PM

backstop/tests/gdpr-banner.html
```diff
@@ -0,0 +1,79 @@
...
+          </button>
+        </div>
+        <div class="iui-gdpr-banner-content">
+          <p>
+            We use cookies for our best experience on our website, including to personalize content and ads, to provide
+            social media features and to analyze traffic.
+            These cookies include targeted media cookies and analytics cookies. For more information see our &#32;<a
+              class="iui-gdpr-banner-cookie-policy"
+              href="https://www.bentley.com/en/cookie-policy"
+            > cookie policy</a>.
```
FlyersPh9, line 61:
```suggestion
            >cookie policy</a>.
```
You can remove the space between `>` and `cookie`.

backstop/tests/gdpr-banner.html
```diff
@@ -0,0 +1,79 @@
...
+              class="iui-button-icon"
+              aria-hidden="true"
+            ></svg-close>
+          </button>
+        </div>
+        <div class="iui-gdpr-banner-content">
+          <p>
+            We use cookies for our best experience on our website, including to personalize content and ads, to provide
+            social media features and to analyze traffic.
+            These cookies include targeted media cookies and analytics cookies. For more information see our &#32;<a
```
FlyersPh9, line 58:
```suggestion
            These cookies include targeted media cookies and analytics cookies. For more information see our&nbsp;<a
```
Our demo pages within `/minified/` remove unnecessary spaces which sometimes results in spaces that we actually want being removed.  To work around this, use `&nbsp;` to keep the space.

src/gdpr-banner/gdpr-banner.scss
marashad001: So I add iui-anchor and just remove lines 25-27?

backstop/tests/gdpr-banner.html
```diff
@@ -0,0 +1,79 @@
...
+              class="iui-gdpr-banner-cookie-policy"
```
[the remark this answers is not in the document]
marashad001: Fixed

src/gdpr-banner/classes.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.iui-gdpr-banner-content {
+  @include iui-gdpr-banner-content;
+}
+
+.iui-gdpr-banner-button-bar {
+  @include iui-gdpr-banner-button-bar;
+}
+
+.iui-gdpr-banner-cookie-policy {
```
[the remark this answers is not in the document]
marashad001: Fixed

src/gdpr-banner/gdpr-banner.scss
marashad001: ^ yep and fixed

src/gdpr-banner/gdpr-banner.scss
marashad001: Made these adjustments

backstop/tests/gdpr-banner.html
marashad001: Fixed

backstop/tests/gdpr-banner.html
marashad001: Thank you, fixed 👍

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+@import '../dialog/index';
+
+@mixin iui-gdpr-banner {
+  @include iui-dialog;
+  border-radius: 0;
+  box-shadow: none;
+  background-color: var(--iui-color-background-3);
```
mayank99, line 10: Are we sure about this color? Looks inconsistent and goes against our documented use case which is _"Disabled components and headers of surfaces"_

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,40 @@
...
+  border-top: 1px solid var(--iui-color-background-border);
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
mayank99, line 15: I don't think we should be specifying `right` and `left` because that makes it span the entire width. It's better to have the lines start wrapping after a certain width (e.g. `75ch`).

On an ultrawide display, this will become unnecessarily wide.

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,40 @@
...
+  @include iui-dialog-button-bar;
+}
+
+//@mixin iui-anchor {
+//  @include iui-anchor;
+//}
+
+//.iui-gdpr-banner-cookie-policy {
+// color: var(--iui-color-foreground-primary);
+//}
```
mayank99, line 40: Is this leftover? unfinished?

src/gdpr-banner/classes.scss
```diff
@@ -0,0 +1,27 @@
...
+  @include iui-gdpr-banner-content;
+}
+
+.iui-gdpr-banner-button-bar {
+  @include iui-gdpr-banner-button-bar;
+}
+
+//.iui-anchor {
+//  @include iui-anchor;
+//}
```
mayank99, line 27: Is this leftover? Should be removed as we already have `iui-anchor` available.

mayank99 (commented):
There is not much code in this component but I left some comments.

I'm thinking it would make more sense if we make this a generic variant inside `dialog`. Something like: `<Dialog placement='bottom'>`. The CSS doesn't really have anything specific to GDPR.

src/gdpr-banner/gdpr-banner.scss
marashad001: Oh yeah I forgot to take out those comments haha

src/gdpr-banner/gdpr-banner.scss
marashad001: Fixed

mayank99: I see you are using `background-1` now. The `iui-dialog` mixin already includes that, so you don't need to repeat it.

src/gdpr-banner/gdpr-banner.scss
mayank99: This makes it worse because it looks fine for "medium" size screens but breaks on narrow or ultrawide screens.



My suggestion is to avoiding hardcoding `left` and `right` to a specific value, and set a `max-width` instead. I think you will also need to center it after that.

marashad001: I got the max-width in, still trying to figure out how to center it

src/index.scss
```diff
@@ -43,3 +43,4 @@
 @import './tree/index';
 @import './user-icon/index';
 @import './wizard/index';
+@import './gdpr-banner/index';
```
elephantcatdog, line 46: This file is also sorted alphabetically.

src/gdpr-banner/gdpr-banner.scss
```diff
@@ -0,0 +1,33 @@
...
+  @include iui-dialog-title-bar;
+}
+
+@mixin iui-gdpr-banner-content {
+  @include iui-dialog-content;
+}
+
+@mixin iui-gdpr-banner-button-bar {
+  @include iui-dialog-button-bar;
+}
```
elephantcatdog, line 33: For all of the mixins that only contain `@include ...` I think you should be able to remove those mixins and instead just place, for example `iui-dialog-button-bar` where you now have `iui-gdpr-banner-button-bar`. Then you can also remove the declarations in the gdpr-banner/classes.scss file. Unless you were asked to do it this way?

elephantcatdog (commented):
I tried to help with finding a solution for centering it, but couldn't figure it out in time. :( I do know that the `justify-content: center` that you have is currently not doing anything. This is meant to be on the parent of a flexbox setup and centers its children. I don't know if you've seen this yet, but I find https://css-tricks.com/snippets/css/a-guide-to-flexbox/ to be a good resource on flexbox. I don't know that flexbox is a good solution for this setup because as far as I know that would require another container div.

src/gdpr-banner/gdpr-banner.scss
marashad001: Yes, the styling for the banner is closely related to the dialog just to keep everything standard and consistent with the rest of the components. I understand what you are saying about replacing the mixins, could you give me an example?

FlyersPh9: That is a good point Leah!  I may have told you to do it this way, but Leah's suggestion makes more sense.

Line 6, `@mixin iui-gdpr-banner`, uses `@include iui-dialog;` and has additional CSS (lines 7-16), so that one makes sense to leave as is.  The rest of the mixins only use a mixin & have no extra CSS, so we don't need to have a mixin reference an include with no additional CSS.

To achieve this:
- You can remove all the mixins that have no additional CSS.
- Within `classes.scss` update the classes to include the dialog mixins directly except for the one mixin that has custom CSS.  Screenshot shows how `classes.scss` should look like:
Screen Shot 2022-06-01 at 1 28 00 PM

marashad001: Got it fixed, thanks guys!

FlyersPh9:
Closing, work will continue in #679.

Issue #603, opened 2022-04-27
[V1] Move variables to separate css-variables package


veekeys:
Important to gather a list of all the variables we want to separate.
Think about vision of this package. i.e. would it consist some sass variables/mixins too?
Separate repo or monorepo?

PR #651, opened 2022-05-23
chore: Convert to monorepo

mayank99:
- Moved all source files into `packages/itwinui-css`
- Added yarn workspaces + [turborepo](https://turborepo.org/)
- Changed the build slightly to output to `css` and `scss` folders and included those in `package.json#files`.
  - This makes the package easier to link locally as there is only one package.json. Copying files is also no longer needed.
  - This resulted in a change in the stylesheet import in all html files.
- Created `shared` package to run copyrightLinter from root
- Updated all paths in workflows (confirmed CI passing)

This is in preparation for the variables package (#603)


README.md
```diff
@@ -1,80 +0,0 @@
-<p align="center">
```
bentleyvk, line 1: Same as changelog.

LICENSE.md
```diff
@@ -1,9 +0,0 @@
-# MIT License
```
bentleyvk, line 1: Maybe leave it a it was. I think github takes it to show license type

packages/itwinui-css/scripts/createComponent.js
```diff
@@ -3,11 +3,7 @@
...
  *--------------------------------------------------------------------------------------------*/
 const fs = require('fs');
-const {
-  copyrightBannerScss,
-  copyrightBannerHtml,
-  copyrightBannerJs,
-} = require('./copyrightLinter');
+const { copyrightBannerScss, copyrightBannerHtml, copyrightBannerJs } = require('./copyrightLinter');
 
 const componentName = process.argv.slice(2).join('-');
```
bentleyvk, line 9: Does it still work?

README.md
mayank99: It's symlinked, it will show normally. Check it out in my branch: https://github.com/iTwin/iTwinUI/tree/mayank/monorepo

bentleyvk: Glad that github understands unix symlinks because windows doesn't 😀
image

packages/itwinui-css/scripts/createComponent.js
mayank99: I didn't change anything, so it should

.gitignore
```diff
@@ -4,7 +4,8 @@ backstop/results/*
 !backstop/results/bitmaps_reference
 
 # Built Files
-lib
+scss
```
veekeys, line 7: Review this file, maybe.. I think paths needs to be udpated, cause after running build, all minified files are shown as new untracked files

veekeys (commented):
Please update contributing guide straight away.
And I am not sure about the MD files and windows :/

packages/itwinui-css/scripts/createComponent.js
```diff
@@ -105,7 +101,7 @@ const demoHtmlFactory = (directory) => {
     <title>${componentName} | iTwinUI</title>
     <link
       rel="stylesheet"
-      href="../../lib/css/all.css"
+      href="../../css/all.css"
```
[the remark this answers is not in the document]
mayank99: What script? Not sure I understand.

.gitignore
mayank99: Updated

README.md
mayank99: It's a symlink that github recognizes. What's the problem?

packages/itwinui-css/scripts/createComponent.js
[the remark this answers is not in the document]
veekeys: `createComponent`, cause it fails for me
image

.gitignore
veekeys: I think that one not enough.. have you checked what happens with test images? They all showed as untracted files after tests run..

README.md
veekeys: That it does not work on my machine (and most of users).

.gitignore
mayank99: Should I move these things to a second gitignore inside `packages/itwinui-css`? Would be more reliable there

packages/itwinui-css/scripts/createComponent.js
mayank99: whoops, I thought you were talking about the copyrightLinter, not createComponent.

this script is fixed now

packages/itwinui-css/scripts/createComponent.js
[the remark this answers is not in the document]
mayank99: this script is fixed now

.gitignore
veekeys: Oh, i see we have multiple gitignores in monorepo in react, so its fine..
I guess just needed to add `*` at the start, so it would pick those anywhere.

README.md
veekeys: image

It shows those linked folders images and if you press on files, you get nothing. This is expected too?
image

mayank99: Hmm, so it won't matter in the case of `README` and `LICENSE` because github knows how to display those, but this `CHANGELOG` is not helpful. Should I just move it from root into the package? I'd imagine itwinui-variables will have its own `CHANGELOG`

veekeys: I dont know.. I see mui has a single changelog file in root with all packages listed in there. Not sure I am fan of that..
If we move changelog per package, how our `Releases` tab would look like? I like having it there.
I would say `LICENSE` and `README` keep in root. License copy from there to package on build and for separate pcakge we need to have separate `README`, while the root one could have general info ?

mayank99: Ok I'm duplicating the files now

package.json
```diff
@@ -38,26 +22,13 @@
...
-    "audit:ci": "audit-ci --moderate --report-type=full",
-    "build-docker": "node scripts/buildDocker.js"
+    "build": "turbo run build",
+    "dev": "turbo run dev",
+    "test": "turbo run test",
+    "lint": "turbo run lint",
+    "lint:copyright": "node packages/shared/copyrightLinter.js"
   },
+  "dependencies": {},
   "devDependencies": {
```
bentleyvk, line 32: Shouldn't a big part of these go to iTwinUI-CSS package.json?

mayank99: Moved the CSS-specific ones into itwinui-css

PR #655, opened 2022-05-24
chore: Remove IE support

### LINKED ISSUE #654: Get rid of custom IE implementation
Get rid of custom IE implementation (gather a list of these, i.e. Inputs)

adamhock:
- Removed `iui-progress-indicator-radial-ie`
- Removed use of `-ms-grid` in `input-container`
- Removed fallback support for `gap`, `-webkit-line-clamp`, and `appearance`
- Removed `focus-visible` mixin

Closes #654


veekeys:
Hope after resolving conflicts, it will not show 140 files as changed..

src/style/mixins.scss
```diff
@@ -87,15 +75,11 @@
...
-  white-space: nowrap;
   text-overflow: ellipsis;
-
-  @supports (-webkit-line-clamp: 1) {
-    white-space: unset;
-    display: -webkit-box; /* stylelint-disable-line */
-    -webkit-line-clamp: $lines;
-    -webkit-box-orient: vertical; /* stylelint-disable-line */
-  }
+  white-space: unset;
```
veekeys, line 79: should not need to unset anymore as you removed `nowrap`

veekeys (commented):
Also update `browserslistrc`

src/checkbox/checkbox.scss
```diff
@@ -19,14 +19,9 @@
 
     gap: $iui-s;
 
-    // #region flex gap fallback
     > * + * {
-      margin-left: $iui-s;
-      @supports (gap: $iui-s) {
-        margin-left: 0;
-      }
+      margin-left: 0;
```
bentleyvk, line 23: So is it needed?

src/color-picker/color-picker.scss
```diff
@@ -26,15 +26,10 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
...
-    row-gap: $iui-baseline;
-  }
+  row-gap: $iui-baseline;
 
   > div:not(:last-child) {
-    margin-bottom: $iui-baseline;
-    @supports (row-gap: $iui-baseline) {
-      margin-bottom: 0;
-    }
+    margin-bottom: 0;
```
bentleyvk, line 32: Again, if we use gap, is this needed?

src/color-picker/color-picker.scss
```diff
@@ -45,16 +40,12 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
   width: $iui-l;
   cursor: pointer;
   border-radius: $iui-color-swatch-border-radius;
-  margin-bottom: $iui-xs;
-  margin-right: $iui-xs;
+  margin-bottom: 0;
```
bentleyvk, line 43: Remove?

src/color-picker/color-picker.scss
```diff
@@ -131,10 +120,7 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
     }
 
     > *:not(:last-child) {
-      margin-right: $iui-xxs;
-      @supports (gap: $iui-xxs) {
-        margin-right: 0;
-      }
+      margin-right: 0;
```
bentleyvk, line 123: Not needed?

src/dialog/dialog.scss
```diff
@@ -192,22 +192,12 @@
...
   align-items: center;
   justify-content: flex-end;
-
+  column-gap: $iui-s;
   @at-root .iui-dialog-draggable & {
     padding: 0 $iui-m $iui-baseline $iui-m;
   }
 
-  // #region IE support
   > .iui-button:not(:last-child) {
```
bentleyvk, line 200: Shouldn't be needed when gap is used

src/input-container/input-container.scss
```diff
@@ -131,11 +79,7 @@
       > .iui-checkbox-wrapper,
       > .iui-radio-wrapper,
       > .iui-toggle-switch-wrapper {
-        margin-right: $iui-m;
-
-        @supports (gap: $iui-m) {
-          margin-right: 0;
-        }
+        margin-right: 0;
```
bentleyvk, line 82: Not needed

src/input-container/input-container.scss
```diff
@@ -131,11 +79,7 @@
...
       }
 
       &:not(:last-child) {
```
bentleyvk, line 85: I think this should also be removed

src/toggle-switch/toggle-switch.scss
```diff
@@ -18,10 +18,7 @@
   }
 
   > * + * {
-    margin-left: $iui-s;
-    @supports (gap: $iui-s) {
-      margin-left: 0;
-    }
+    margin-left: 0;
```
bentleyvk, line 21: Remove?

bentleyvk:
> I think we will need to keep the `focus-visible` mixin because it is a _very_ new addition to Safari.
> 
> We might even want to keep some of the flex `gap` fallbacks because it was only added last year, although `@supports (gap: 2px)` is not entirely reliable because it will return true when grid gap is supported and flex gap is not supported. One workaround here could be to replace flex with grid in these places.

Is one year not enough with `gap` support is not enough?

mayank99:
> Is one year with `gap` support is not enough?

It's not bad but still feels kinda new. If we used gap with `grid` instead of `flex`, we could get it up to 93%.

but ultimately it's just whitespace so I'm fine either way.

elephantcatdog (commented):
The other comments have covered everything I noticed.

src/input-container/input-container.scss
adamhock: I tried removing this one and it lead to some visual changes. I'll look into it a little further

src/input-container/input-container.scss
adamhock: fixed

src/dialog/dialog.scss
adamhock: fixed

src/color-picker/color-picker.scss
adamhock: fixed

src/color-picker/color-picker.scss
adamhock: fixed

src/color-picker/color-picker.scss
adamhock: fixed

src/checkbox/checkbox.scss
adamhock: fixed

src/style/mixins.scss
adamhock: fixed

adamhock:
> > Is one year with `gap` support is not enough?
> 
> It's not bad but still feels kinda new. If we used gap with `grid` instead of `flex`, we could get it up to 93%.
> 
> but ultimately it's just whitespace so I'm fine either way.

I have a local stash with changes that change the majority of the flex gap changes to grid gap. I'm still trying to fix one or two more cases in color picker.

.browserslistrc
```diff
@@ -1,12 +1,14 @@
 # This file is used by autoprefixer and stylelint to adjust CSS output to support the specified browsers below.
+
```
bentleyvk, line 2: Just remove those new lines for less changes.

src/radio-tile/radio-tile.scss
```diff
@@ -68,6 +68,7 @@
   appearance: none;
   opacity: 0;
   position: absolute;
+
```
bentleyvk, line 71: Remove this new line so it won't be shown as a change

src/style/mixins.scss
```diff
@@ -19,7 +19,6 @@
     text-decoration: none;
   }
 }
-
```
bentleyvk, line 22: Bring back the ne lines

src/toggle-switch/toggle-switch.scss
bentleyvk: Is it?

src/style/mixins.scss
adamhock: fixed

src/toggle-switch/toggle-switch.scss
adamhock: Now it is

scripts/generateCss.js
```diff
@@ -18,10 +18,10 @@ const compileScss = async (path, outFile) => {
   return new Promise(async (resolve, reject) => {
     try {
       const cssResult = await sass.compileAsync(path);
-      const processedCssResult = await postcss([
-        require('autoprefixer')({ grid: 'autoplace' }),
-        require('postcss-discard-comments'),
-      ]).process(cssResult.css, { from: undefined }); // `from` is required
+      const processedCssResult = await postcss([require('autoprefixer'), require('postcss-discard-comments')]).process(
```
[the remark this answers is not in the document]
adamhock: Fixed

veekeys (approved):

bentleyvk:
So are we going to replace all `:not(:last-child)` that I mentioned in previous comment and that were done because of IE or not?

adamhock:
> * color-picker
>   image

I kept the use of `margin-bottom` for `iui-color-field` because not every child of `iui-color-selection-wrapper` is going to be `iui-color-field`, so we only want that margin if the color field is the last child

> * information-panel
>   image

I'm not sure if there's a way to use `gap` for this because this margin is applied to every other child so that the labels and inputs are grouped together

> * column-filter
>   image
> * pill tabs
>   image
> * tags
>   image
> * tile
>   image

These instances have been changed to incorporate `gap`

src/input-container/input-container.scss
adamhock: This use of `margin-right` is to separate `iui-input-group` from `iui-end-icon`, so it must be kept.

packages/itwinui-css/src/tag/tag.scss
```diff
@@ -59,6 +59,7 @@
 
 @mixin iui-tag-container {
   color: var(--iui-text-color-muted);
+  column-gap: $iui-xs;
```
bentleyvk: Don't you need to remove?
```sass
> *:not(last-child) {
  margin-right: $iui-xs
}
```

adamhock: Actually, because `iui-tag-container` does not have `display: flex`, I cannot use `gap` here

bentleyvk (approved):

PR #661, opened 2022-05-27
chore: Add `box-sizing: border-box` as default

### LINKED ISSUE #614: [V1] Add box-sizing: border-box as default

elephantcatdog:
Generally:
- added rules to default box-sizing property to `border-box` in global.scss
- removed all individual css properties `box-sizing: border-box`
- fixed problems that came up with visual tests, except for some that I determined were ok to change which were often heights/widths being a little bit different, though I did try to get them as close as I could

Notable changes described below:

Alert
- test: sticky/scrolling alert is a bit taller

Avatar
- adjusted width and height of status bubble from `25%` to a variable that's calculated alongside the status icon size. Percentages were no longer reliable as something like 31.5% was close to the current size but not exact, and the current sizes are nice numbers like '12x12'
- added check for extra-small icon size (tested it visually but didn't add it to the tests. I still have the changes saved and can change the tests)

Badge
- test: long label width is a bit longer

Codeblock
- adjusted width of numbers in codeblock
- test: codeblock is a bit shorter

Color Picker
- test: slider handle is moved left/right a bit because of use of percentages to place the handle

Header
- didn't need `+ $iui-xxs` removed, unless change of header height is desired
- redefined height to account for border width

Non-Ideal State
- test: body height changed a bit

Notification Marker
- redefined width/height to account for border width

Radio Tile
- redefined height to account for padding

Side Navigation
- redefined width to account for border width in header.html
- redefined widths to account for border width in side-navigation.scss

Slider
- redefined height to account for border width
- slider handle is moved left/right a bit because of use of percentages to place the handle

Stepper
- redefined width/height to account for border width

Surface
- redefined height to account for border width in surface.html

Tag
- redefined .iui-tag-container max-width to account for padding
- specified that `#demo-basic` should still have a max-width of 420px as it has no padding

Tile
- redefined height to account for border witdth

Toast
- redefined body height in toast.html as the 'real height' changed with box-sizing change
- test: the test body height changed a bit, so every test image is a little off

Toggle Switch
- redefined width/height as it changed with box-sizing change

closes #614


src/header/header.scss
```diff
@@ -227,7 +226,7 @@
 
   gap: 0;
   max-width: 25vw;
-  height: 100%;
+  height: calc(100% - 2px); // `- 2px` accounts for border width
```
elephantcatdog, line 229: I think I will create a variable like `border-width: 1px` to make these changes clear without the comments, and for use in all cases where border is 1px.

src/slider/slider.scss
```diff
@@ -3,7 +3,7 @@
 @import '../style/index';
 @import '../icon/index';
 
-$thumb-height: $iui-m; // 16px
+$thumb-height: $iui-m + $iui-xxs; // 18px to account for border
```
veekeys, line 6: hm.. Maybe we should keep 16px? Not sure..

src/stepper/stepper.scss
```diff
@@ -29,8 +29,8 @@
     align-items: center;
     border-radius: 50%;
     display: flex;
-    width: $iui-l;
-    height: $iui-l;
+    width: $iui-l + $iui-xxs; // `+ $iui-xxs` accounts for border width
```
veekeys, line 32: again.. same.. maybe keep previous size? All these new ones kinda go against the ones we have set as standard.

src/style/global.scss
```diff
@@ -33,6 +33,14 @@ html {
...
     @include theme-variables(dark-hc);
   }
+
+  box-sizing: border-box;
+}
+
+*,
+*::before,
+*::after {
+  box-sizing: inherit;
```
veekeys, line 43: Why did it this way? Just adding for all to have `bordex-box` did not work out?

src/tile/tile.scss
```diff
@@ -153,7 +153,7 @@
 
 @mixin iui-tile-thumbnail {
   width: 100%;
-  height: $iui-baseline * 14;
+  height: $iui-baseline * 14 + 1px; // `+ 1 px` to account for border width
```
veekeys, line 156: Same.. return `content-box` or keep the size as previous? But here ight be sensitive, cause border is only on the bottom..

veekeys (commented):
In general, looks good except those places where we need to change sizes by adding padding/border. I would like to read/find more material on best practices here, becuase not sure if it is good to have now these 'weird' values.
Till now what I have found is only using width/height on parent element and apply needed paddings/borders on child. That way we keep standard sizes and fully use `bordex-box` advantages.
If you have found/read some articles, would appreciate if you could share so I get a better understanding.

src/slider/slider.scss
veekeys: Oh wait.. I am lost.. even though you were adjusting sizes, images changed. So maybe didnt need to do that?
i.e. slider image now looks like thumb is not centered (or it was not centered) ?

src/style/global.scss
elephantcatdog: I was following the[ recommendation I found at css-tricks](https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/).

src/slider/slider.scss
elephantcatdog: 16px wasn't centered vertically. The images changed because the slider is not in the same place horizontally. If it doesn't look vertically centered, it was that way before.

src/header/header.scss
```diff
@@ -11,9 +11,8 @@
   display: flex;
   justify-content: space-between;
   width: 100%;
-  height: $iui-baseline * 5 + $iui-xxs; // Remove `+ $iui-xxs` when box-sizing: border-box is applied to all elements.
+  height: $iui-baseline * 5 + $iui-xxs;
```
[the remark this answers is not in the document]
elephantcatdog: Generally for changing from `content-box` to `border-box`, the height and width needs to increase to account for any padding or borders or anything that wasn't included in the height/width previously. So removing `+ $iui-xxs` made the header shorter, but I didn't need to add anything to keep it the same size. I am assuming that the height should stay the same rather than the being shorter? I will ask Jon about it when he gets back; I think he's the one that made the comment.

src/dialog/dialog.scss
```diff
@@ -35,10 +35,8 @@
 @mixin iui-dialog {
   border-radius: $iui-border-radius;
   box-shadow: $iui-elevation-24;
-  box-sizing: border-box;
   position: absolute;
   padding: $iui-baseline $iui-m;
```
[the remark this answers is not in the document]
elephantcatdog: I removed this duplicate. I didn't see any more duplicates in this file, but I'll scan other files too.

src/avatar/sizes.scss
```diff
@@ -8,11 +8,17 @@
   height: $size;
 
   @if ($size == $iui-icons-xl) {
-    --_iui-avatar-status-size: #{$iui-xs * 1.5};
+    --_iui-avatar-status-icon-size: #{$iui-xs * 1.5};
+    --_iui-avatar-status-bubble-size: #{$iui-sm};
+  } @else if ($size == $iui-icons-large) {
```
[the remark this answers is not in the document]
elephantcatdog: There was no catch for icons with sizes other than xl, xxl. The default (`else: ...`) is for icon size 3xl. There are also no catches for `$iui-icons-default` and `$iui-icons-small`. I can add those, but at that point it's going to be difficult to even see the status, and I don't know if we support avatars that are that small.

src/badge/badge.scss
```diff
@@ -24,7 +24,7 @@ $iui-badge-text-color: rgba($iui-color-foreground-body--light, $iui-opacity-3--l
   color: var(--_iui-badge-text-color);
   text-transform: uppercase;
   user-select: none;
-  max-width: 20ch;
+  max-width: 22ch;
```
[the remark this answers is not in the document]
elephantcatdog: `22` is closer to the previous size than `20`. Perhaps I could do `calc(20ch + 2px)` but that's... another weird value. [It's how these people are doing it though.](http://revoltpuppy.com/articles/89/using-ch-an-underappreciated-css-length#:~:text=But%20wait%2C%20there%20are%20a%20couple%20extra%20considerations.%20We%E2%80%99re%20reasonable%20people%2C%20and%20we%20set%20our%20box%2Dsizing%20to%20border%2Dbox.)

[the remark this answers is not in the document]
bentleyvk: Maybe it is worth making this one explicitly `content-box`?

src/notification-marker/notification-marker.scss
```diff
@@ -8,8 +8,8 @@
   &::before {
     content: '';
     position: absolute;
-    width: $iui-s;
-    height: $iui-s;
+    width: $iui-sm;
```
bentleyvk, line 11: Why changed?

bentleyvk, line 11: Maybe better just add +4 for the border it will be more clear or content-box

src/slider/slider.scss
bentleyvk: But it is not centered horizontally because it can be seen clearly that 50% mark is not in the middle of the thumb. You might need to tweak this one 
image

backstop/tests/toast.html
```diff
@@ -70,7 +70,7 @@
   </head>
   <body
     class="iui-body"
-    style='height: 90vh;'
+    style='height: 95vh;'
```
bentleyvk, line 73: Why changed?

backstop/tests/tag.html
```diff
@@ -25,6 +25,9 @@
     <style>
       /* Forces grouped tags to display on 1 line, can be set by the dev team. */
       section .iui-tag-container {
+        max-width: 444px;
+      }
+      section#demo-basic .iui-tag-container {
```
bentleyvk, line 30: Why needed?

src/badge/badge.scss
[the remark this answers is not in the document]
veekeys: I would go either then for `content-box` or `calc`. Just `ch` value seems confusing, cause it does not really fit 22 chars in there :/

packages/itwinui-css/src/style/global.scss
```diff
@@ -65,3 +65,10 @@ html {
   // Scrollbar
   @include iui-scrollbar;
 }
+
+[class*='iui-'],
+[class*='iui-'] *,
```
veekeys, line 70: hm... when thinking about it, I dont feel like we can just apply border-box to all children of iui elements.. That ends up in all children of `iui-body` having border-box and thats guaranteed to mess up users projects..
Maybe lets tackle ONLY iui elements? See which elements are getting broken when changing this..
Also, I think we might want to exclude body from iui selector, because body is also a tricky part to change box-sizing :/
This change also should fix some of the test images.

veekeys (commented):
Lots of test images shouldnt have changed. This is what I noticed:
- `Alert`: should be fixed after changing selector for global `box-sizing`
- `Badge`: Try reverting the image and running the test. I dont see the difference.
- `ColorPicker`: Same as Badge
- `NonIdealState`: Can you check what has broken the bottom border of last element?? Selector is not applied anymore:
image
- `Slider`: Same as Badge
- `Tag`: Apply maxWidth to section, so we dont get the image to change
- `Tile`: Thumbnail now is 153px, which is not good... Use `content-box` on it ? All images should be kept old then
- `Toasts`: These also shouldnt have an influence and its connected with the comment on global style and applying box-sizing to body. If we exclude body, old images should be fine

packages/itwinui-css/src/style/global.scss
veekeys: Opinions @FlyersPh9 @mayank99 @bentleyvk  ?

packages/itwinui-css/src/header/header.scss
```diff
@@ -227,7 +226,7 @@
 
   gap: 0;
   max-width: 25vw;
-  height: 100%;
+  height: calc(100% - 2px); // `- 2px` accounts for border width
```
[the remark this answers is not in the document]
veekeys: Button does not take 100% full height and bottom border is not on place as a result

veekeys:
If this is not needed, please close it
https://github.com/iTwin/iTwinUI/pull/674

packages/itwinui-css/src/style/global.scss
bentleyvk: Yeah, we can't target all children either target only `iui-*` or we should omit body by adding `:not(.iui-body)` but it will increase specificity.

elephantcatdog:
image
Using `:last-child` doesn't work when the live-server injects a script as a sibling to the non-ideal-state scrolling divs. I changed it to `:last-of-type`, which doesn't seem like a perfect solution, but it works in this case. As long as the setup of this demo page doesn't change it'll be fine. Alternatively, I can wrap all of the `"iui-non-ideal-state-scrolling"` divs in another div.

packages/itwinui-css/src/badge/badge.scss
```diff
@@ -24,7 +24,7 @@ $iui-badge-text-color: rgba($iui-color-foreground-body--light, $iui-opacity-3--l
   color: var(--_iui-badge-text-color);
   text-transform: uppercase;
   user-select: none;
-  max-width: 20ch;
+  max-width: calc(20ch + 12px); // `+ 12px` accounts for padding;
```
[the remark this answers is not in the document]
elephantcatdog: Updated

packages/itwinui-css/src/header/header.scss
```diff
@@ -84,7 +83,7 @@
   }
 
   &.iui-slim {
-    height: round($iui-baseline * 2.5) + $iui-xxs; // Remove `+ $iui-xxs` when box-sizing: border-box is applied to all elements.
+    height: $iui-baseline * 3;
```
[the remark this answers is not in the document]
elephantcatdog: Now the default header is 55px tall and the slim header is 33px, which are multiples of `$iui-baseline`.

packages/itwinui-css/src/header/header.scss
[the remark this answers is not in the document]
elephantcatdog: Fixed

packages/itwinui-css/src/notification-marker/notification-marker.scss
```diff
@@ -8,8 +8,8 @@
   &::before {
     content: '';
     position: absolute;
-    width: $iui-s;
-    height: $iui-s;
+    width: $iui-s + 4px; // `+ 4px accounts for borders
```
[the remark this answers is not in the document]
elephantcatdog: Updated

packages/itwinui-css/src/style/global.scss
elephantcatdog: I updated this and removed `*` for now.

veekeys:
I have resolved most of my comments, but you still need to review/update the images and we can see, whats the situation there after you changed the selector for `border-box`

packages/itwinui-css/backstop/tests/non-ideal-state.html
```diff
@@ -35,7 +35,7 @@
         border-bottom: solid 2px #000;
       }
 
-      .iui-non-ideal-state-scrolling:last-child {
+      .iui-non-ideal-state-scrolling:last-of-type {
```
elephantcatdog, line 38: Why is this necessary for this PR specifically?

packages/itwinui-css/src/notification-marker/notification-marker.scss
```diff
@@ -2,19 +2,21 @@
...
 @mixin iui-notification-marker($status: primary) {
   position: relative;
 
   &::before {
     content: '';
     position: absolute;
-    width: $iui-s;
-    height: $iui-s;
+    width: $iui-s + $iui-notification-marker-border-width * 2;
+    height: $iui-s + 4px; // `+ 4px accounts for borders
```
veekeys, line 14: ups, forgot to reuse variable here.

packages/itwinui-css/backstop/tests/tag.html
```diff
@@ -24,8 +24,11 @@
     />
     <style>
       /* Forces grouped tags to display on 1 line, can be set by the dev team. */
-      section .iui-tag-container {
-        max-width: 420px;
+      section#demo-default-container {
```
veekeys, line 27: Lets do simpler:
`#demo-default-container, #demo-basic {
  width: 440px;
}
`

packages/itwinui-css/src/style/global.scss
```diff
@@ -65,3 +65,9 @@ html {
   // Scrollbar
   @include iui-scrollbar;
 }
+
+[class*='iui-'],
```
veekeys, line 69: ok, lets update this selector to exclude body and toast images should be reverted then
```
[class*=iui-]:where(:not(.iui-body)),
[class*=iui-]:where(:not(.iui-body))::before,
[class*=iui-]:where(:not(.iui-body))::after{
  box-sizing:border-box;
}
```

This should not increase specificity and also we will not influence users app in any way

packages/itwinui-css/backstop/tests/non-ideal-state.html
veekeys: This it not needed specifically.. It was needed only because we run live server which appends that script..But leave it, I think it is ok for future. It does not have any influece on tests, though

packages/itwinui-css/src/style/global.scss
elephantcatdog: Did it this way and reverted toast tests

packages/itwinui-css/src/notification-marker/notification-marker.scss
elephantcatdog: Whoops, fixed

packages/itwinui-css/backstop/tests/tag.html
veekeys: amm.. weird.. I have tried putting 440px locally, and there were no visual changes...

elephantcatdog: Yeah, it looks the same on the html page. I don't know why the tests don't show the tag placement properly. I set the width to `445px` to fix that part. They're still changed now, but it's because the image is being taken of `#demo-default-container` and `#demo-basic`, which now have a set width of `445px` vs `fit-to-content`. The widths used to be 420px and 444px.

packages/itwinui-css/backstop/tests/non-ideal-state.html
elephantcatdog: But why didn't it cause problems for people running the live server previously?

elephantcatdog: But why didn't it cause problems when running the live server previously?

packages/itwinui-css/backstop/tests/non-ideal-state.html
```diff
@@ -33,9 +33,10 @@
         scroll-snap-align: start;
         position: relative;
         border-bottom: solid 2px #000;
+        box-sizing: content-box;
```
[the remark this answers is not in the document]
elephantcatdog: Yeah that makes sense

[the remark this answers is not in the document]
elephantcatdog: Fixed

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
```diff
@@ -44,6 +44,7 @@
 
   > :any-link {
     @include iui-anchor;
+    box-sizing: border-box;
```
adamhock, line 47: I don't understand why this was added instead of redefining the width of the link to fix the visual issues.

adamhock (commented):
Looks good to me for the most part. Once we can come up with some solutions to revert some of the visual changes, we should be good to go

packages/itwinui-css/src/radio-tile/radio-tile.scss
```diff
@@ -147,8 +146,8 @@
   height: $iui-icons-xl;
   display: block;
   margin: 0 auto;
-  padding-top: round($iui-baseline * 0.5);
-  padding-bottom: $iui-baseline;
+  margin-top: round($iui-baseline * 0.5);
+  margin-bottom: $iui-baseline;
```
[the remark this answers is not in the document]
veekeys: Because `border-box` does include `padding` into `width` or `height.
Read more here
https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing#values

I say its much better if you leave 2 explicit rules rather than merging those, better readability. Also, `auto` is not the same, as actually the default `margin` value is `0`.
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#formal_definition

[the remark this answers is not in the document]
adamhock: The line before these defines the margin already though.

`margin: 0 auto;`

I just thought it was weird to define the margin and then redefine `margin-top` and `margin-bottom`.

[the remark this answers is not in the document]
veekeys: oh, right.. Havent seen that one. Then makes sense, I guess

packages/itwinui-css/backstop/tests/non-ideal-state.html
veekeys: It did, just no one needed to practically check last image as we didnt have failures like you did. So at first we (you and me) thought that this is the issue (that selector), but actually it was not, cause no script is embedded while running tests.

veekeys (approved):

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
elephantcatdog: In this case, only the "generic anchor" breadcrumbs example was failing because it did not have a class name that begins with `"iui-"`. The method I'm using to set `box-sizing: border-box` only defines it on elements with class names starting with `"iui-"`. `"iui-anchor"` gets `border-box`, but `:any-link` doesn't, so I'm manually including it. Also, Vykintas said that the "generic anchor" example was going to be deprecated soonish and that this solution was fine for that little time in between.
image

packages/itwinui-css/src/radio-tile/radio-tile.scss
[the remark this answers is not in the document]
elephantcatdog: Updated

packages/itwinui-css/src/style/global.scss
FlyersPh9: Would this work?
```scss
[class*='iui-'],
[class*='iui-']::before,
[class*='iui-']::after { … }
```

adamhock (approved):
Great job! 🎉

bentleyvk (approved):

PR #639, opened 2022-05-13
chore: v1 branch

mayank99:
This draft PR will help run CI on each commit and also provide the "Update branch" button.

Live preview at: https://itwin.github.io/iTwinUI/639

PR #668, opened 2022-06-01
testing github pages stuff

gretanausedaite:
lets see what happens 🤷

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -151,3 +151,75 @@ jobs:
             ${{ steps.open_prs.outputs.result }}
           git-config-name: github-actions[bot]
           git-config-email: github-actions[bot]@users.noreply.github.com
+          
+  deploy:
+    name: Deploy demo html pages
+    if: github.ref == 'refs/heads/v1.0'
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


mayank99:
I think `v1.0` might need to be added in the `branches` array to trigger builds automatically

.github/workflows/build.yml
```diff
@@ -3,9 +3,13 @@ name: CI
...
   push:
-    branches: [main]
+    branches:
+      - 'main'
+      - 'v1.0'
   pull_request:
-    branches: [main]
+    branches:
+      - 'main'
+      - 'v1.0'
```
mayank99, line 12: I think this can be reverted now that we have tested it's working

mayank99: So there's two pieces: `push` and `pull_request`.

1. `push` is redundant, because #639 is created to merge with `main`, so this workflow should run every time we merge a PR into the `v1.0` branch

2. `pull_request` we could leave to run build/tests on every individual v1 PR, but it won't publish test results to gh-pages. That might be ok.

Just thinking out loud, maybe this is fine.

mayank99 (approved):

PR #670, opened 2022-06-01
feat: Separate out `Backdrop` from `Dialog`

mayank99:
Four main changes:
1. Moved dialog backdrop into a generic `iui-backdrop` component in `backdrop.scss`
2. Moved all modifiers from `iui-dialog-backdrop` into `iui-dialog`.
3. Changed dialog to use `position: fixed` instead of `absolute`.
4. Added `iui-dialog-wrapper` which should be used for confining the draggable 

Combined, this means the dialog does not rely on the backdrop and does not even need a wrapper except for the "relative to container" case:
```html
<div class='iui-backrop' />
<div class='iui-dialog'>...</div>
```

Also removed a lot of nesting/overrides and added a new `iui-dialog-title-bar-draggable` modifier class.

_Pending: properties/classes for react-transition-group_


FlyersPh9:
This all looks good, I'll give the approval once the pending changes are in.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,144 +3,105 @@
...
+  position: fixed;
   padding: $iui-baseline $iui-m;
   box-sizing: border-box;
   padding: $iui-baseline $iui-m;
   overflow: hidden;
-  @include themed {
-    background-color: t(iui-color-background-1);
-  }
+  pointer-events: auto;
+  background-color: var(--iui-color-background-1);
```
bentleyvk, line 31: I don't think we can use plain CSS variables until v1 as it is not a new component.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -181,33 +140,12 @@
...
-
-  @supports (column-gap: $iui-s) {
-    column-gap: $iui-s;
-
-    > .iui-button:not(:last-child) {
-      margin-right: 0;
-    }
-  }
-  // #endregion IE support
+  gap: $iui-s;
```
bentleyvk, line 150: Same as with CSS varaibles, I think we still need to support IE and this will be removed with Adam's PR.

packages/itwinui-css/backstop/tests/dialog.html
```diff
@@ -75,11 +75,9 @@ <h1>Dialog</h1>
       Toggle draggable dialog
     </button>
 
-    <div
-      class="iui-dialog-backdrop iui-dialog-default"
-      id="default-dialog"
-    >
-      <div class="iui-dialog">
+    <div id="default-dialog">
```
bentleyvk, line 78: Shouldn't we have some wrapper? We could even use it for later to center dialog using flex.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,144 +3,105 @@
...
-    @media screen and (max-width: 400px) {
-      max-width: 95%;
-      width: 95%;
-      min-width: 95%;
-    }
-  }
+  left: 50%;
+  top: 33%;
+  transform: translate(-50%, -33%);
+  max-width: Max(50%, $iui-dialog-width);
```
bentleyvk, line 53: `Max` not supported by IE and what is the point of using it here?

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,144 +3,105 @@
...
+  min-width: $iui-dialog-width;
+  max-height: 100vh;
```
bentleyvk, line 55: No `dvh` here?

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,144 +3,105 @@
...
-    min-height: 144px;
-    display: flex;
-    flex-direction: column;
-    padding: 0;
-    @include themed {
-      border: 1px solid t(iui-color-background-border);
-    }
-  }
+  max-width: 100%;
+  max-height: 100vh;
```
bentleyvk, line 85: `dvh`?

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,144 +3,105 @@
...
+  min-height: 144px;
+  display: flex;
+  flex-direction: column;
+  border: 1px solid var(--iui-color-background-border);
 }
 
 @mixin iui-dialog-animation {
-  // Full Page Dialog Animations for React
-  &-enter .iui-dialog-full-page .iui-dialog {
+  // TODO: fix react-transition-group for default dialog
```
bentleyvk, line 94: After fixing those, it would be great to also test in React whether animations are still working by linking CSS.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -158,20 +119,18 @@
...
-  @at-root .iui-dialog-draggable & {
-    user-select: none;
-    font-size: $iui-font-size-leading;
-    padding: round($iui-baseline * 0.5) $iui-m;
-    cursor: grab;
-    @include themed {
-      background-color: t(iui-color-background-3);
-      border-bottom: 1px solid t(iui-color-background-border);
-    }
+@mixin iui-dialog-title-bar-draggable {
```
bentleyvk, line 124: Why changed?

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -181,33 +140,12 @@
   padding: 0 $iui-m;
   overflow-y: auto;
   overflow-y: overlay;
-
-  @at-root .iui-dialog-draggable & {
```
bentleyvk, line 185: Is it not needed?

packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-backdrop {
```
bentleyvk, line 5: Would be nice to also have progress backdrop here too but maybe in the future.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: 3 reasons I used css var:
- The PR for removing `themed` mixin was already completed in v1 branch, so I don't want to create additional work when we merge branches.
- IE will be officially considered "dead" in 12 days and we might not even release this until then, so there really is absolutely no reason to go through the extra time.
- draggable is new so we can ignore that. In the existing dialog, `background-1` is just white which is the default background for html anyway.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: It dynamically gets the maximum of 50% or 380px. Useful on mobile devices where 50% could be too narrow.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: This says `max-height` which is kind of an edge case. I only added it for the `height` of full page dialog because otherwise it would _always_ overflow on iOS.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: See https://github.com/iTwin/iTwinUI/pull/670/files#r888873008

packages/itwinui-css/src/dialog/dialog.scss
mayank99: To reduce nesting and decouple it from the modifier class. The styling is not directly related to draggability (other than the cursor) so it could even be useful for custom dialogs (in itwinjs for example). Just not sure about the name.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: Nope, because I'm setting negative margins in the title bar

packages/itwinui-css/src/backdrop/backdrop.scss
mayank99: I think "backdrop" is kinda different from "overlay" which is what the progress one is. Backdrop goes _under_ something else (e.g. dialog or hamburger menu).

packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -0,0 +1,26 @@
...
+  z-index: 998;
+  isolation: isolate;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, var(--iui-opacity-4));
```
gretanausedaite, line 13: We don't have variable for this one?

mayank99: what do you mean? A combined variable for `rgba(0, 0, 0, var(--iui-opacity-4))`? I don't see any

gretanausedaite (commented):
Looks good. I think some time ago we talked about changing top and left possition in `@mixin iui-dialog-default` to flex and center. But I might be wrong.

mayank99:
> Looks good. I think some time ago we talked about changing top and left position in `@mixin iui-dialog-default` to flex and center. But I might be wrong.

Yeah, we did bring it up some time ago. And I brought it up again in another PR because we want to add additional placements to dialog (top, bottom, bottom-left, etc similar to toasts).

However, using flex or grid would make adding a wrapper element required. Currently it works without a wrapper.

gretanausedaite:
> 
> Yeah, we did bring it up some time ago. And I brought it up again in another PR because we want to add additional placements to dialog (top, bottom, bottom-left, etc similar to toasts).
> 
> However, using flex or grid would make adding a wrapper element required. Currently it works without a wrapper.

It won't be breaking change if we make wrapper required later?

mayank99:
@bentleyvk @FlyersPh9 @gretanausedaite i linked the css changes into react, made the necessary changes in react (same as in our html), and the animation Just Works™️ without any extra css changes needed.

https://user-images.githubusercontent.com/9084735/171881241-1fd011b0-f3bb-4cc1-ad2e-d237c3139ef8.mp4

packages/itwinui-css/src/dialog/dialog.scss
bentleyvk: But we announced to everyone that we will drop IE with v1 and now we are starting to break IE before that.

packages/itwinui-css/src/dialog/dialog.scss
bentleyvk: Maybe `iui-filled`, `iui-bordered`? Idk, @FlyersPh9 always has ideas 😀

packages/itwinui-css/src/dialog/dialog.scss
bentleyvk: If it would default to white I would be ok but: 
image

mayank99: ok, thanks for checking. I've hardcoded fallback background colors now. Is that fine?

packages/itwinui-css/backstop/tests/dialog.html
gretanausedaite: Related (conversation)[https://github.com/iTwin/iTwinUI/pull/670#issuecomment-1145957843].

We have `wrapper` for contained dialog. It might be hard to think other name.

gretanausedaite (approved):
Looks good to me even without wrapper around `Dialog`. 
If we want to add it, we could name it `iui-dialog-container` or something.

packages/itwinui-css/src/dialog/dialog.scss
FlyersPh9: At the moment I think this is only intended for a draggable dialog because it gives the user a clear area where they can click & drag.  I also think only a draggable dialog can be resized.  I'm ok with the current name & `@mixin iui-dialog-title-bar-filled`, but if we want to think of alternatives maybe `@mixin iui-dialog-title-bar-controllable` or something along those lines?

mayank99: Changed to `iui-dialog-title-bar-filled`

Issue #625, opened 2022-05-04
[V1] Change font family


gretanausedaite:
Font family to consider: https://fonts.google.com/noto/fonts
It has a lot of languages and monospace font.

Things to consider:
- Does it support variable font sizes?
- How big is font file?

PR #673, opened 2022-06-03
feat(style): Add `Noto` font family

gretanausedaite:
Issue #625 

Add `Noto Sans` and `Noto Sans Mono` fonts


packages/itwinui-css/backstop/assets/demo.css
```diff
@@ -1,32 +1,40 @@
...
+  src: url('fonts/NotoSans-Light.ttf') format('truetype');
 }
+
 @font-face {
-  font-family: 'Open Sans';
+  font-family: 'Noto Sans';
   font-style: normal;
   font-weight: 400;
-  src: url('fonts/open-sans-v28-latin-regular.woff2') format('woff2');
+  src: url('fonts/NotoSans-Regular.ttf') format('truetype');
```
mayank99, line 16: Why `ttf`? We should be using `woff2`.

gretanausedaite: Google fonts doesn't have woff2? https://fonts.google.com/noto/specimen/Noto+Sans

veekeys (approved):
I have scrolled through images, mostly look fine. 
Non ideal state ones were weird, that font change also influence the width of the image.
Also input button combo image looked weird, that it influenced width and height I think.

I would agree to check `woff2` type too. t showed some results, but not sure if reliable (i.e. https://github.com/GreatWizard/notosans-fontface)

packages/itwinui-css/backstop/assets/demo.css
gretanausedaite: I added Noto Sans woff2 files. Noto sans mono is variable font - maybe we can leave it as is?

mayank99: Noto Sans Mono size:
- variable ttf: 1.3MB
- variable woff2 latin full: 51KB
- variable woff2 latin weights: 28KB
- weight 400 woff2 latin: 9KB

I would be ok with any of the latter three. 1.3MB is not an acceptable size for any webfont.

gretanausedaite: Changed to variable woff2 latin full: 51KB

mayank99 (approved):

adamhock (approved):

gretanausedaite (commented):
Changed Mono font file to more extended one because previous didn't have arrow symbols (therefore tests failing). Unfortunately, it's `woff` and 156KB

gretanausedaite:
Could not make tests fail locally and I don't know what's wrong, adding back `woff` file.

PR #675, opened 2022-06-03
feat: Move variables to new `@itwin/itwinui-variables` package

mayank99:
Created new workspace `@itwin/itwinui-variables` located inside `packages/`.

Right now it just hardcodes the compiled variables and spits them to `index.css`. Not being used in itwinui-css yet.

---

#### Pending

- [ ] Remove `theme.scss` and use this new package in `@itwin/itwinui-css` 
- [ ] Autoprefixer
- [ ] Narrow down list / add new vars?
- [ ] Remove opacity-based variables?
- [ ] JS API?


mayank99:
@iTwin/itwinui-css-1 Would be good to have initial thoughts on this while I work on the pending tasks.

@FlyersPh9 We should go through the list of variables together and come up with a final list that needs to be exposed to our users.

packages/itwinui-variables/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "itwinui",
+    "itwin",
+    "design system",
+    "design token"
+  ],
+  "dependencies": {},
+  "devDependencies": {
+    "autoprefixer": "^10.4.7",
+    "postcss": "^8.4.14",
+    "sass": "^1.52.2"
```
bentleyvk, line 26: Not sass-embeded?

yarn.lock
```diff
@@ -3253,6 +3316,15 @@ postcss@^8.3.11, postcss@^8.4.5:
     picocolors "^1.0.0"
     source-map-js "^1.0.1"
 
+postcss@^8.4.14:
```
bentleyvk, line 3319: Why these are added? Don't we already have this dep?

packages/itwinui-variables/mixins.scss
```diff
@@ -0,0 +1,275 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin light-variables {
+  --iui-color-background-1-rgb: 255, 255, 255;
```
bentleyvk, line 4: Not hsl?

bentleyvk: Also, are those `*-rgb` still needed?

packages/itwinui-variables/.gitignore
```diff
@@ -0,0 +1,2 @@
+index.css
+index.css.*
```
bentleyvk, line 2: What it ignores?

packages/itwinui-variables/package.json
mayank99: sass-embedded doesn't seem to have a cli, and it's only file so performance is not an issue

packages/itwinui-variables/mixins.scss
mayank99: i copied these from our current generated css. I can convert to hsl I suppose

the `-rgb`/`-hsl` would allow easily changing the opacity if needed

packages/itwinui-variables/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+# MIT License
+
+Copyright © Bentley Systems, Incorporated. All rights reserved.
```
veekeys, line 3: Year is missing 😄

veekeys, line 3: Year is missing 😄

packages/itwinui-variables/index.scss
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './mixins';
+
+:root {
+  @include light-variables;
```
veekeys, line 6: would it make sense to add `theme` ?
`light-theme-variables`

veekeys, line 6: would it make sense to add `theme` ?
`light-theme-variables`

packages/itwinui-variables/mixins.scss
veekeys: As I mentoined, having separate `rgb` variable looks weird.. WOuld you mind dropping some examples from other products, which you mentioned have the same?

packages/itwinui-variables/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@itwin/itwinui-variables",
+  "version": "0.1.0",
+  "author": "Bentley Systems",
+  "license": "MIT",
+  "main": "index.css",
+  "style": "index.css",
+  "files": [
+    "index.css",
+    "LICENSE.md"
```
veekeys, line 10: maybe includ CHANGELOG and README straight away? Or would it have errors if files do not exist?

veekeys, line 10: maybe includ CHANGELOG and README straight away?

yarn.lock
veekeys: Agree, that lock file changes look weird..

veekeys: Agree, that lock file changes look weird..

packages/itwinui-variables/mixins.scss
```diff
@@ -0,0 +1,275 @@
...
+  --iui-icons-color-actionable: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-2));
+  --iui-icons-color-actionable-hover: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-1));
+  --iui-icons-color-actionable-disabled: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
+  --iui-icons-color-primary: var(--iui-color-foreground-primary);
+  --iui-icons-color-positive: var(--iui-color-foreground-positive);
+  --iui-icons-color-warning: var(--iui-color-foreground-warning);
+  --iui-icons-color-negative: var(--iui-color-foreground-negative);
+}
+
+@mixin dark-variables {
```
veekeys, line 73: So all variables are stored in mixins now.. We were talking also about possibility to use certain theme variables and with current implementation it would be possible to do ONLY with sass. Just a thought for future, so we are sure we would not need to change much.

veekeys, line 73: So all variables are stored in mixins now.. We were talking also about possibility to use certain theme variables and with current implementation it would be possible to do ONLY with sass. Just a thought for future, so we are sure we would not need to change much.

packages/itwinui-variables/index.scss
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './mixins';
+
+:root {
```
veekeys, line 5: Maybe you know the pros of using `:root` here?

veekeys, line 5: Maybe you know the pros of using `:root` here?

packages/itwinui-variables/index.scss
```diff
@@ -0,0 +1,35 @@
...
+  @media (prefers-color-scheme: light) and (prefers-contrast: more) {
+    @include light-hc-variables;
+  }
+
+  @media (prefers-color-scheme: dark) and (prefers-contrast: more) {
+    @include dark-hc-variables;
+  }
+}
+
+[data-iui-theme='light'] {
```
veekeys, line 21: So we would end up with a single set of variables included right? 

First goes default OS and then override with attribute value (if it exists)

veekeys, line 21: So we would end up with a single set of variables included right? 

First goes default OS and then override with attribute value (if it exists)

veekeys:
I would also suggest consuming this package in css only after it is published. (separate PR)

packages/itwinui-variables/index.scss
mayank99: `:root` is interchangeable with `html` usually. Most places I've seen use `:root` to declare CSS variables.

Two differences I'm aware of:
- `:root` has higher specificity than `html`. I think that's fine in this case.
- `:root` refers to the root of the shadow DOM if used there, but it doesn't apply in this case.

packages/itwinui-variables/index.scss
mayank99: yeah, that's if the data attribute is applied on the `<html>` element.

if it's applied on some bespoke element deeper in the DOM, it will override the one from `:root`, as intended.

packages/itwinui-variables/index.scss
mayank99: it's just a name of an internal mixin and it's one of the two files in the repo, so it doesn't really matter. but i can change it if you want

packages/itwinui-variables/package.json
mayank99: README gets included automatically

cant we just leave this until we have a CHANGELOG?

packages/itwinui-variables/mixins.scss
mayank99: you mean like `--light` and `--dark` versions of every CSS variable?

packages/itwinui-variables/LICENSE.md
mayank99: it's the same as our CSS package

packages/itwinui-variables/mixins.scss
mayank99: i think we would have `--light-hc` and `--dark-hc` as well

packages/itwinui-variables/.gitignore
bentleyvk: Do we even need to generate map?

mayank99: Removed sourcemap from prod build. I think it can be useful for development

yarn.lock
mayank99: should be deduped now

packages/itwinui-variables/package.json
veekeys: Well, just now I remembered, when we have CHANGELOG, might forget :D 
but yeah, as long as we ensure its there at the end result, Im fine

packages/itwinui-variables/index.scss
veekeys: Yeah, I was aware of the 2nd point..
First one didnt know.. interesting.

packages/itwinui-variables/mixins.scss
veekeys: > > I would also suggest consuming this package in css only after it is published. (separate PR)
> 
> I do not think that is going to be very nice. Many of the variables will be changed / removed. Or do you mean keep the outdated variables and code in `itwinui-css`?

Yeah, the latter.

veekeys:
> > I would also suggest consuming this package in css only after it is published. (separate PR)
> 
> I do not think that is going to be very nice. Many of the variables will be changed / removed. Or do you mean keep the outdated variables and code in `itwinui-css`?

Yeah, the latter.

packages/itwinui-variables/src/space.scss
```diff
@@ -0,0 +1,18 @@
...
+  --iui-size-xs: calc(8 / 16 * 1rem);
+  --iui-size-s: calc(12 / 16 * 1rem);
+  --iui-size-m: calc(16 / 16 * 1rem);
+  --iui-size-l: calc(24 / 16 * 1rem);
+  --iui-size-xl: calc(32 / 16 * 1rem);
+  --iui-size-2xl: calc(64 / 16 * 1rem);
+  --iui-size-3xl: calc(96 / 16 * 1rem);
+}
+
+@mixin border-radii {
```
bentleyvk, line 15: Mistype?

packages/itwinui-variables/src/typography.scss
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin typography {
+  --iui-font-sans: 'Noto Sans', 'Open Sans', system-ui, sans-serif;
+  --iui-font-mono: 'Noto Sans Mono', ui-monospace, 'Segoe UI Mono', Consolas, 'Roboto Mono', monospace;
+
+  --iui-font-size-0: calc(12 / 16 * 1rem);
```
bentleyvk, line 7: Didn't we decide to switch to rem later?

bentleyvk: So what is the point of having them if they are not being used? I expect you will use them together with all the variables.

mayank99: None of the variables are used. See https://github.com/iTwin/iTwinUI/pull/675#issuecomment-1150828601

I don't want to release with `px` first and then change to `rem` later because it is tough switching between units. I expect that when we use the variables, we will make all the required changes together without having to come back to it later

mayank99:
**Update**: I've been playing around with the build process and output. Here's my findings:
- postcss is not needed because sass automatically discards comments and there isn't much else in this package. also, adding postcss will complicate the process and slow down build times.
- the output file is currently 30+KB uncompressed and 2.2KB gzipped.
- if I remove the `-hsl` variables, the size goes down to 18KB uncompressed and 1.8KB gzipped
- if I also remove the duplication that we have in `:root`, and only keep the ones in `[data-iui-theme]`, then the size goes down to 9KB uncompressed and 1.3KB gzipped.

Currently, I'm thinking of (1) moving the media query vars into a different file because those are not used in react, and (2) moving the hsl vars into a different file because they will not be used by most apps. This way, we still get to keep those files for those who need it, and reduce bundle size for those who don't.

I'm also currently looking into converting the hsl colors into hex if they don't have opacity to further reduce bundle size.

Thoughts? @iTwin/itwinui-css-1

packages/itwinui-variables/mixins.scss
veekeys: Example links use it even weirder.. They have separate `rgb` var, but they do not use it directly for the main color, they use `hex` 😄 So honestly, I am getting lost... And I am trying to imagine users doing something like in the article examples, that they would take our `base` color and play around with opacity...
I just see 2 use cases:
- User just uses our 'final' values in his solution
- User fully changes the theme in w/e way he wants (so he changes our variables values). Now how he does that, its fully up to him, he can have his own 'base' variables and from them construct his custom colors.
But if you just want to keep those `rbg` or `hsl` values, moving to separate file makes sense to me.

veekeys (commented):
When you say `postcss` is needed, that doesnt mean we will drop `autoprefixer`, right? I think that one is still useful.
Thinking about theming.. I would say having a default behavior by OS theme might be more useful than data attributes?
But we still need data attributes, for inner components theming, right? Or would we be able to handle that by just letting users import a set of variables under some selector? That way we could remove data attributes selectors.

bentleyvk:
> * the output file is currently 30+KB uncompressed and 2.2KB gzipped.
> * if I remove the `-hsl` variables, the size goes down to 18KB uncompressed and 1.8KB gzipped

Are these big numbers for the size of CSS lib? At least 18KB seems not very much especially when gzipped. 

> Currently, I'm thinking of (1) moving the media query vars into a different file because those are not used in react, and 

But if user sets theme to `os` in react, then it won't work, so they are kind of needed. And somehow it still feels that as a CSS lib it is nice to support OS theming out of the box.

> (2) moving the hsl vars into a different file because they will not be used by most apps. This way, we still get to keep those files for those who need it, and reduce bundle size for those who don't.

Can you do that? Because you are using those `-hsl` variables inside normal color variables. I am just not sure whether we need them at all.

packages/itwinui-variables/src/space.scss
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin sizes {
+  --iui-size-3xs: calc(2 / 16 * 1rem);
```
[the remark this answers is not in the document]
mayank99: replaced `calc` with `math.div`, so the output is a bit cleaner now

yarn.lock
```diff
@@ -825,6 +825,21 @@ chokidar-cli@^2.1.0:
     lodash.throttle "^4.1.1"
     yargs "^13.3.0"
 
+"chokidar@>=3.0.0 <4.0.0":
```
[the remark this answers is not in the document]
mayank99: these come from `sass`, which is a new dependency (previously we only had `sass-embedded`) needed for the cli commands

mayank99:
@veekeys @bentleyvk Please have a look at the latest changes and let me know what you think. Pull the branch and run `yarn build` to see the output in `index.css` and `os.css`. Responding to individual points below:

> When you say `postcss` is needed, that doesnt mean we will drop `autoprefixer`, right? I think that one is still useful.

autoprefixer is not useful for variables. sass already converts the new color formats to more widely supported ones, and there is nothing else to autoprefix.

---

> Thinking about theming.. I would say having a default behavior by OS theme might be more useful than data attributes? But we still need data attributes, for inner components theming, right? Or would we be able to handle that by just letting users import a set of variables under some selector? That way we could remove data attributes selectors.

> But if user sets theme to `os` in react, then it won't work, so they are kind of needed.

Guys, we don't make use of the default css media queries in react. Our `useTheme` hook has additional media queries and always adds a `iui-theme-` class. After these changes, the hook would just need to set `data-iui-theme=` instead of the class.

> And somehow it still feels that as a CSS lib it is nice to support OS theming out of the box.

It's just an additional import (`os.css` in addition to `index.css`). I think it's nice that we don't have to ship all the extra stuff in the bundle when not needed.

---

> > * the output file is currently 30+KB uncompressed and 2.2KB gzipped.
> > * if I remove the `-hsl` variables, the size goes down to 18KB uncompressed and 1.8KB gzipped
> 
> Are these big numbers for the size of CSS lib? At least 18KB seems not very much especially when gzipped.

These were just for the `index.css` file in itwinui-variables. After the latest changes, it's 1.3KB.

---

> > (2) moving the hsl vars into a different file because they will not be used by most apps. This way, we still get to keep those files for those who need it, and reduce bundle size for those who don't.
> 
> Can you do that? Because you are using those `-hsl` variables inside normal color variables. I am just not sure whether we need them at all.

I am now using sass maps, so I can do more static stuff without the need to use css variables in intermediate calculations.

veekeys:
I have checked both files, look ok. Just sad cannot combine css selector with media-queries. That way we could have just 1 file without duplication...
- Should we add `light` theme checker in `os.css` ?
- Have we thought about delivering sass files too? Or we gonna have just separate css files which you can import under any selector to get all the css vars (does it work that way)? I think it is connected with the question I had [here ](https://github.com/iTwin/iTwinUI/pull/675#discussion_r898828013). In other libs I saw just css files with the variables defined.. Would be nice if we could achieve all native css solution for the users too, so we do not depend on sass.
- Do you need that mapping in every file instead of having a simple list of vars? Seems readability just got worse... I think it might be much better if @FlyersPh9 could see the exact variables names instead of figuring the values through the map or going to the built css ant finding it there.. But I'd like to have some sass magic to ensure, that we have all the variables defined in every theme, for example...

mayank99:
> * Should we add `light` theme checker in `os.css` ?

`index.scss` already defaults to light theme.

> * Have we thought about delivering sass files too? Or we gonna have just separate css files which you can import under any selector to get all the css vars (does it work that way)? I think it is connected with the question I had [here ](https://github.com/iTwin/iTwinUI/pull/675#discussion_r898828013). In other libs I saw just css files with the variables defined.. Would be nice if we could achieve all native css solution for the users too, so we do not depend on sass.

I do not plan on exposing Sass in the deliverable. But in native CSS you cannot define variables outside a selector, so you cannot import it under any selector. What's the use case for doing that?

> * Do you need that mapping in every file instead of having a simple list of vars? Seems readability just got worse... I think it might be much better if @FlyersPh9 could see the exact variables names instead of figuring the values through the map or going to the built css ant finding it there.. But I'd like to have some sass magic to ensure, that we have all the variables defined in every theme, for example...

What part is hard to figure out? I need the intermediate Sass map for producing derived values at the end (e.g. `--iui-color-background-disabled` and `--iui-text-color`). Plus it has the additional benefit of converting the new hsl syntax into hex.

packages/itwinui-variables/src/index.scss
```diff
@@ -0,0 +1,39 @@
...
+  @include shadows;
+  @include soft-backgrounds;
+}
+
+:root {
+  // global vars shared between all themes
+  @include common;
+
+  // default to light theme if data attribute not set anywhere
+  @include themes.light-theme;
```
veekeys, line 22: I know I commented before on naming, but looking at this include, maybe enough having just `light`? 
We already know its theme from the file and import

packages/itwinui-variables/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@itwin/itwinui-variables",
+  "version": "0.1.0",
+  "author": "Bentley Systems",
+  "license": "MIT",
+  "main": "index.css",
+  "style": "index.css",
+  "files": [
+    "index.css",
```
veekeys, line 9: I think I asked that before, but we are sure we do not want to export our mixins, so users could include only parts of variables (if that makes sense) ?
I also assume, after this pckage is rleased, we will ask users to stop using itwinui-css and switch to this one? 
But then how variables inclusion would work (if user uses itwinui-react which takes itiwnui-css with variables and also uses itwinui-variables directly. Would it end up with duplicated declarations of variables? We need to test these combos in an app)

veekeys (commented):
Left some minor comments.
@FlyersPh9 Can you go one more time through all the variables (if you havent yet) and verify the values/list of those?
Now I think we need to test this out in multiple scenarios
- use it in itwinui-css package
- use it in an app
	- where there is no itwinui-react
	- with itwinui-react
	- maybe even with itwinui-css ?
	
we can also test out pipelines with some dev release. Hopefully our bot with auto release PR will also work?

packages/itwinui-variables/src/colors.scss
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Bentley Systems Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@use 'sass:color';
```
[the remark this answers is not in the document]
mayank99: Not anymore. Removed

packages/itwinui-variables/src/space.scss
```diff
@@ -0,0 +1,20 @@
...
+  --iui-size-s: #{math.div(12, 16)}rem;
+  --iui-size-m: #{math.div(16, 16)}rem;
+  --iui-size-l: #{math.div(24, 16)}rem;
+  --iui-size-xl: #{math.div(32, 16)}rem;
+  --iui-size-2xl: #{math.div(64, 16)}rem;
+  --iui-size-3xl: #{math.div(96, 16)}rem;
+}
+
+@mixin border-radii {
+  --iui-border-radius-1: 3px;
```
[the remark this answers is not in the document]
mayank99: Changed

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -0,0 +1,55 @@
...
+  2: 0.94,
+  3: 0.88,
+  4: 0.76,
+  5: 0.34,
+  6: 0.07,
+);
+
+@mixin theme {
+  @each $color, $value in $base-colors {
+    --iui-#{$color}: #{hex($value)};
```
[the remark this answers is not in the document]
mayank99: Smaller output + more compatible with older browsers

packages/itwinui-variables/src/index.scss
mayank99: In `themes.light-theme`:
- "themes" comes from the import (we say `as themes`)
- "light-" comes from the `@forward` in `themes/index.scss`.
- "theme" comes from the name of the mixin in `themes/light.scss`.

I don't think I can get rid of the `-theme` suffix.

packages/itwinui-variables/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "css",
+    "theme",
+    "itwinui",
+    "itwin",
+    "design system",
+    "design token"
+  ],
+  "dependencies": {},
+  "devDependencies": {
+    "sass": "^1.52.2"
```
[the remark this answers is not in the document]
mayank99: Updated

packages/itwinui-variables/src/index.scss
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@use './themes/index' as themes;
+@use './space' as *;
```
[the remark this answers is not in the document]
veekeys: True, but still.. size for border radius sounds more logical than space for border radius?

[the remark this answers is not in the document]
mayank99: how about `spacing.scss`?

packages/itwinui-variables/mixins.scss
mayank99: I just found a place where we use these `-rgb`/`-hsl` values internally
https://github.com/iTwin/iTwinUI/blob/888eb9ec1407af22196d914cbc9d47a0529b1ee5/packages/itwinui-css/src/style/mixins.scss#L49-L53

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -0,0 +1,60 @@
...
+  // Background colors - default
+  background-1: hsl(200 12% 20%),
+  background-2: hsl(200 12% 15%),
+  background-3: hsl(200 12% 10%),
+  background-4: hsl(200 12% 25%),
+  background-5: hsl(200 12% 30%),
+  background-1-hover: hsl(0 0% 95%),
+  //
+  // Background colors - statuses subtle
+  background-subtle-primary: hsl(204 14% 44%),
```
mayank99, line 16: @FlyersPh9 We still need to fix these subtle background values in high contrast themes. Leaving this comment here as a reminder.

packages/itwinui-variables/index.scss
bentleyvk: In recent tweets I saw someone changing from `:where(html)` to `:host` in order to support web components

veekeys (approved):

packages/itwinui-variables/src/typography.scss
FlyersPh9: If switching to REM results in the same visuals / no test image changes, I don't see any issue with switching.

packages/itwinui-variables/src/themes/dark-hc.scss
FlyersPh9: Here are the values for dark high contrast subtle colors:
```scss
  background-subtle-primary: hsl(199 16% 24%),
  background-subtle-positive: hsl(167 17% 23%),
  background-subtle-warning: hsl(144 4% 24%),
  background-subtle-negative: hsl(210 5% 24%),
```

packages/itwinui-variables/src/themes/light-hc.scss
```diff
@@ -0,0 +1,60 @@
...
+  background-3: hsl(200 12% 94%),
+  background-4: hsl(200 12% 88%),
+  background-5: hsl(200 12% 80%),
+  background-1-hover: hsl(0 0% 95%),
+  //
+  // Background colors - statuses subtle
+  background-subtle-primary: hsl(202 73% 94%),
+  background-subtle-positive: hsl(94 41% 93%),
+  background-subtle-warning: hsl(34 85% 95%),
+  background-subtle-negative: hsl(0 67% 94%),
```
FlyersPh9, line 19: Here are the values for light high contrast subtle colors:
```scss
  background-subtle-primary: hsl(200 25% 95%),
  background-subtle-positive: hsl(135 15% 95%),
  background-subtle-warning: hsl(48 20% 95%),
  background-subtle-negative: hsl(0 36% 96%),
```

mayank99: Added

packages/itwinui-variables/src/index.scss
```diff
@@ -0,0 +1,44 @@
...
+
+:root {
+  // global vars shared between all themes
+  @include common;
+
+  // default to light theme if data attribute not set anywhere
+  @include themes.light-theme;
+}
+
+[data-iui-theme='light'] {
```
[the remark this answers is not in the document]
bentleyvk: I was thinking if light high contrast theme is set then only its variables are present cause now there will be regular light theme variables and then they are overridden with high contrast. Or I just don't fully understand and there is no problem 😀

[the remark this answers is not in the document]
mayank99: But that's how CSS works: the last values in the cascade are the ones that are used 😅  Even if I revert to the "old" way, there will always be light theme variables in the `:root` selector.

One benefit of this approach is that we are able to share some "unthemed" variables when in high contrast mode (see bottom of this file). Another benefit is the logic to use high contrast theme in iTwinUI-react becomes much simpler and less messy.

Another potential benefit I can see in the future is we can remove some of the overrides that use the same value between default and high-contrast themes (e.g. background-1 for both `light` and `light-hc` is white so it doesn't need to be re-specified)

elephantcatdog (approved):
Looks good to me. I really like the move to separate the themes into different files; I think that was a good idea.

PR #678, opened 2022-06-06
feat(Stepper): Class name & specificity refactor

FlyersPh9:
- Wizard default & wizard long renamed to Stepper.
- Wizard workflow renamed to Workflow diagram.
- Reduced stepper & workflow diagram specificity from 0,5,1 to 0,2,1.

**I submitted this PR early so we can take a look at it and see if this is something we want to target.**. If not, we can close it before I spend any more time on it.


packages/itwinui-css/src/stepper/stepper.scss
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // First step
+  &:first-of-type::before {
+    --_iui-stepper-track-first-half: transparent;
+  }
+
+  // Last step
+  &:last-of-type::before {
+    --_iui-stepper-track-second-half: transparent !important;
```
FlyersPh9, line 112: Back when I was working on this, this was one area I couldn't figure out.

packages/itwinui-css/src/stepper/stepper.scss
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // Clickable steps
+  &.iui-clickable {
```
FlyersPh9, line 116: I can reduce specificity by making this no longer nested.  However, when trying this, I was getting incorrect results.  So to do this I will need to work on it a bit more.  I'd like to determine if I should continue work on this PR or close it before spending any more time.

bentleyvk:
It is a lot of changes to review, it might take a while. We either close this PR or leave it till the end of v1 to see how many stuff is still there and how much time we have.

bentleyvk:
It is a lot of changes to review, it might take a while. We either close this PR or leave it till the end of v1 to see how many stuff is still there and how much time we have.

FlyersPh9:
Closing for now, I would like to revisit this PR after CSS 1.0 because of the reduced specificity and fewer HTML elements.

PR #679, opened 2022-06-06
feat(Dialog): add positioning to dialog

### LINKED ISSUE #637: Create GDPR banner

marashad001:
- Change positioning of dialog on page using using `data-position`.
- Closes #637.


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -51,11 +51,48 @@
...
+  transform: translate(-50%, -33%);
+
+  &[data-position] {
+    position: fixed;
+    top: auto;
+    left: auto;
+    transform: none;
+  }
+
+  &[data-position='centered'] {
```
FlyersPh9, line 66: I think ideally `data-position` would only be used for the corner positioning & would not be used for centering.

**Centered:**
```html
<div class="iui-dialog">
```

**Corners:**
```html
<div class="iui-dialog" data-position="top-left">
```

packages/itwinui-css/backstop/tests/dialog.html
```diff
@@ -275,13 +331,21 @@ <h1>Dialog</h1>
...
       function toggleDialog(dialogId) {
         let element = document.getElementById(dialogId);
         element.classList.toggle('iui-dialog-visible');
       }
 
+      const dialog = document.querySelector('#dialog-click');
+      // Clicking on menu-item should toggle selection
+      document.querySelectorAll('input').forEach((item) => {
+        item.addEventListener('click', () => {
+          // dialog.classList.value = `iui-toast-wrapper iui-placement-${item.value}`;
```
FlyersPh9, line 344: You can remove this commented out line of code.

packages/itwinui-css/src/dialog/dialog.scss
marashad001: It is difficult to try to center the dialog while not affecting the way it aligns to the corners. I decided it would be easier to just include the center as a position for the component so we can see how it looks.

marashad001: ^ I'll keep trying though

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -51,11 +51,48 @@
...
 
-@mixin iui-dialog-default {
-  .iui-dialog {
-    left: 50%;
+@mixin iui-dialog-position {
+  top: 33%;
+  left: 50%;
+  transform: translate(-50%, -33%);
+
+  &[data-position] {
```
bentleyvk, line 59: Why you decided to use `data-position` instead of regular class as it is done in Toast?

FlyersPh9: @mayank99 suggested it to me and I like the idea.  I would like to see us use `data-position` for toasts eventually too.  Is there any negative reason why we shouldn't do this?

mayank99: i suggested `data-placement` btw. `position` doesn't make sense to me. we might also want to add our prefix, e.g. `data-iui-placement`.

but yeah, the benefit of data attribute is that you can kinda enforce that only one of the options is active at a time. also has a nice JS API

bentleyvk: I can't think of any negative reasons besides consistency.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -74,6 +111,7 @@
   }
 
   .iui-dialog {
+    position: relative;
```
mayank99, line 114: is this needed?

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -51,11 +51,48 @@
   }
 }
 
-@mixin iui-dialog-default {
-  .iui-dialog {
-    left: 50%;
+@mixin iui-dialog-position {
+  top: 33%;
+  left: 50%;
+  transform: translate(-50%, -33%);
```
mayank99, line 57: i think we should leave this default case untouched where it previously was.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -51,11 +51,48 @@
...
+
+  &[data-position] {
+    position: fixed;
+    top: auto;
+    left: auto;
+    transform: none;
```
mayank99, line 63: This looks a bit random. Any reason you added `position: fixed` here instead of to the base dialog? And `top: auto`/`left: auto` could also be moved into the individual placement cases where you know the exact value.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -51,11 +51,48 @@
...
+  &[data-position='centered'] {
+    position: fixed;
     top: 33%;
+    left: 50%;
     transform: translate(-50%, -33%);
+  }
+
+  &[data-position='top-left'] {
+    top: $iui-baseline;
+    left: $iui-m;
```
mayank99, line 75: I think it would make more sense to use `top: 0`, `left: 0`, etc and then apply margin

packages/itwinui-css/backstop/tests/dialog.html
marashad001: fixed

packages/itwinui-css/src/dialog/dialog.scss
marashad001: Yes it sets the default and pushes the dialog to the different corners based on that particular point on the screen.

packages/itwinui-css/src/dialog/dialog.scss
marashad001: They're needed in every iteration of the placement, including the centering. Honestly I placed them in there because that was the only solution I was able to get working for a default center.

mayank99 (approved):

mayank99:
Looks good!

PR #681, opened 2022-06-08
fix(Tabs): Improve visuals with wrapped labels

### LINKED ISSUE #677: Tab header is not responsive
APIM team is facing issues with Tabs not being responsive enough:
- Borderless tab underline does not extend full width on small screens.
- Tab does not adjust hight to match label.

Eg.:
image
image
image

Contact person:
@irenajur

veekeys:
Closes #677 

- added `min-height` instead of `height` to all tabs
- Added some padding so when text is big there is some space around (and space between the marker line)
- Changed test content to handle longer text behavior
- Added `flex-shrink` to icon as it was getting smaller on overflow
- Needed to to change `align-items` value (flex default - `stretch`) and add `height: 100%` to button to make all tabs take full available height
- Removed not used flex rule from vertical tabs (they dont even use display flex)


packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -68,7 +68,8 @@ <h2>Default</h2>
                 aria-hidden="true"
               ></svg-placeholder>
               <span class="iui-tab-label">
-                <div>Active</div>
+                <div>This is a very long 'Active' tab content, which hopefully will wrap if all tab headers content is
```
mayank99, line 71: this is really messing up the demos. should we add new ones instead of changing the existing ones?

also the text probably doesn't need to be as long

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -16,6 +15,8 @@
     display: flex;
     align-items: center;
     font-size: $iui-font-size;
+    min-height: $iui-component-height;
+    height: 100%;
```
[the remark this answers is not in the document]
veekeys: `flex-grow` I assume would not work, cause I need vertical size..
Have any better suggestions?

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -16,6 +15,8 @@
...
     cursor: pointer;
     @media (prefers-reduced-motion: no-preference) {
```
FlyersPh9, line 21: We can remove the `@media (prefers-reduced-motion: no-preference)` here and in other places that aren't effecting motion.

FlyersPh9, line 21: We can remove the `@media (prefers-reduced-motion: no-preference)` here and in other places that aren't effecting motion.

packages/itwinui-css/src/tabs/default.scss
```diff
@@ -78,7 +77,7 @@
 
   &.iui-large {
     .iui-tab {
-      height: $iui-component-height-large;
+      min-height: $iui-component-height-large;
```
FlyersPh9, line 80: How would you feel about setting a `max-height` or a `-webkit-line-clamp` to prevent extreme cases like this?
Screen Shot 2022-06-09 at 12 46 21 PM

FlyersPh9, line 80: How would you feel about setting a `max-height` or a `-webkit-line-clamp` to prevent extreme cases like this?
Screen Shot 2022-06-09 at 12 46 21 PM

packages/itwinui-css/backstop/tests/tabs.html
FlyersPh9: I would also prefer if the first demo looks normal, maybe we can use the third demo of each category for long labels?

FlyersPh9: I would also prefer if the first demo looks normal, maybe we can use the third demo of each category for long labels?

packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -85,7 +86,8 @@ <h2>Default</h2>
                 aria-hidden="true"
               ></svg-placeholder>
               <span class="iui-tab-label">
-                <div>Left</div>
+                <div>This is a very long 'Left' tab content, which hopefully will wrap if all tab headers content is
+                  pretty long and cannot fit in viewport width</div>
                 <div class="iui-tab-description">Sublabel</div>
```
FlyersPh9, line 91: Maybe test a long sublabel instead of main label for the 2nd tab?

FlyersPh9, line 91: Maybe test a long sublabel instead of main label for the 2nd tab?

gretanausedaite (commented):
`+` button should be vertically centered.  
image

packages/itwinui-css/src/tabs/tabs.scss
veekeys: You will need to help me here..
You want me to remove all the media queries from this file?

packages/itwinui-css/backstop/tests/tabs.html
veekeys: Added separate demos with long text

packages/itwinui-css/src/tabs/default.scss
veekeys: I dont know... What would you feel you like to see here? Limit height as a standard?

packages/itwinui-css/backstop/tests/tabs.html
veekeys: added

gretanausedaite:
> > `+` button should be vertically centered. image
> 
> Yeah.. nice catch.. We do not really have implementation for such case and would need to handle some additional handler.. Possibly, add separate class for this custom button in tabs. Maybe with separate PR?

In my mind "improve visuals with wrapped labels" should include button centering. But sure, separate PR is ok.

packages/itwinui-css/src/tabs/default.scss
FlyersPh9: I'd be fine with setting it to 3.  If a user needs more lines it wouldn't be hard for them to override the default.

packages/itwinui-css/src/tabs/tabs.scss
FlyersPh9: Hypothetical example not from `tabs.scss`:
```scss
transform: rotate(90deg);
transition: background-color $iui-speed-fast ease;
  transition: background-color $iui-speed-fast ease, transform $iui-speed-fast ease;
}
```
The above example would allow `background-color` to transition for all users & the media query would allow `background-color` & `transform` (motion) to animate for users who don't enable reduced motion.  For users who don't enable reduced motion, that media query would override the previous line.

Looking through `tabs.scss`, most of the transitions have to do with `background-color` & `fill` which are not motion.  The transitions that target `width` & `height` are motion so those can stay within the media query.

We have been misusing the media query and wrapping pretty much all transitions, but now we know it should only target motion specifically.

FlyersPh9:
>> Yeah.. nice catch.. We do not really have implementation for such case and would need to handle some additional handler.. Possibly, add separate class for this custom button in tabs. Maybe with separate PR?
>
> In my mind "improve visuals with wrapped labels" should include button centering. But sure, separate PR is ok.

Our vertical tabs stretch the button to full width.
Screen Shot 2022-06-14 at 3 27 27 PM

But our horizontal tabs do not effect the height.
Screen Shot 2022-06-14 at 3 28 42 PM

Stretching might be what we want for buttons, but maybe not for other elements like a `<select>`.  Maybe that last `<li>` should have a new class to handle non-tab elements?
Screen Shot 2022-06-14 at 3 36 42 PM

packages/itwinui-css/src/tabs/default.scss
```diff
@@ -82,6 +82,9 @@
 
     .iui-tab-description {
       display: block;
+      @supports (-webkit-line-clamp: 1) {
+        display: -webkit-box; /* stylelint-disable-line */
```
mayank99, line 86: Is this needed? The line clamp mixin should already set it

veekeys: It is needed.. We have `display: none` which needs to be overriden and its happening here :/

veekeys: I thought so too after seeing >90% support.. But kinda left same as in the mixin :/

veekeys: changed

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
mayank99: ah ok, this is because we have a wrapper `<li>`. Maybe you could apply grid or flex on that. Or maybe it's fine this way because 100% height is controlled by the tallest element inside the `<ul>`

packages/itwinui-css/src/tabs/default.scss
veekeys: It is 0.2.0
image

Sure too bad?
Excluding full refactor, I would need to add new class on label `iui-tab-name` and then move this whole selector outside of tabs mixin.

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -75,6 +73,10 @@
 
   .iui-tab-label {
     text-align: left;
+
+    > * {
+      @include iui-line-clamp(3);
```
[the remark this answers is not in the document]
mayank99: The general guidance is to always have text be limited to 45-75 characters. We control the label styling, but we don't dictate anything about the content (and the content may not be text), so it makes sense to me that max-width should go on the label

packages/itwinui-css/src/tabs/default.scss
mayank99: If you look at how the user would normally override this, they need to apply a custom classname on the Tab and then target our `iui-tab-label` like so:

```scss
.my-tab > .iui-tab-label {
  // ...
}
```

Because our specificity is 0-2-0, this will not always work (depending on order of imports) and then they might reach for `!important`. So yeah, it's better to reduce it to 0-1-0.

Why do you need a new class? You could use the same `iui-tab-label`.

veekeys: what? I need to apply overflow rules on both, name and description divs. I cannot apply it just on parent, it would not work.
So I need to introduce a new class for the first div, because currently it doesnt have any.

mayank99: The `> *` selector can help with that, no need for a new class.

veekeys: ok, moved into mixin, now it has 0.1.0

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
veekeys: ok, added max width of `60ch` on label

gretanausedaite (approved):
Looks good

mayank99 (approved):

PR #687, opened 2022-06-15
chore: Review folder structure

### LINKED ISSUE #615: [V1] Review folder structure in repo
- [ ] Checkup components(inputs, etc.)

- [ ] Add utils folder for helper mixins like visually hidden and others. 

- [ ] Review styles folder

elephantcatdog:
Co-authored-by: adamhock <adamhock@users.noreply.github.com>

Our general strategy was move what we determined were helper mixins and helper components into the new `utils` folder. We thought about how specific mixins/components were and whether they are or could be used by other components. For components, we also considered the lack of a .html file as a point for it being a helper.

Major Changes:
- Added `utils` folder
  - Added `classes.scss` and `index.scss` files for this folder
- Moved files and folders into `utils` folder
  - ~`avatar/sizes.scss` - helper mixin~
  - `style/mixins.scss` - helper mixins
  - `style/ripple.scss` - helper mixin
  - `input-container` folder - used by other components and doesn't have its own .html file
  - `notification-marker` folder - used by other components and doesn't have its own .html file
  - `popover` folder - used by other components and doesn't have its own .html file
  - `text` folder - used by other components and doesn't have its own .html file
- Moved files into `style` folder
  - `icon/variables.scss` -> `style/icons-sizes.scss` - moved into the collection of variable files contained in `style` folder
  - ~`table/variables.scss` -> `style/table-densities.scss` - moved into the collection of variable files contained in `style` folder~
- Removed mixin `no-link` from `utils/mixins.scss` - was not being used
- Consolidated `anchor/anchor.scss` and `style/anchor.scss` into `anchor/anchor.scss`

Minor Changes:
- Fixed all `@import` references to moved/renamed files and folders
- Renamed `mixins.scss` in `icon` to `icon.scss` 
- Cleaned up unecessary `@import` staements and downgraded some from using `style/index` to a more specific file
- Added `@import` statements for files that depended on another file for that import

closes #615


elephantcatdog:
There are two mixin files besides `mixins.scss` included in the `utils` folder: `sizes.scss` and `ripple.scss`. These could be moved into the `mixins.scss` file to consolidate all of the util mixins for ease of use. Alternatively, the mixins within `mixins.scss` could be separated into different files so that individual mixins would be referenced when needed rather than every mixin being included in a component file. Thoughts?

packages/itwinui-css/src/utils/sizes.scss
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
-@import '../text/mixins';
 
 @mixin iui-avatar-size($size: $iui-icons-xl, $font-size: $iui-font-size-leading) {
```
gretanausedaite, line 5: Maybe rename file to `avatar-sizes`. Similar to `icon-sizes` in styles folder.

gretanausedaite (commented):
Could you explain your thought process and difference between `utils` and `style` folder. Why icon-sizes are in style and avatar-sizes are in utils?
Why text is in utils?

elephantcatdog:
We tried to arrange it so that the `style` folder has variables defined in it, and the `utils` folder has mixins in it. I see that [what's now called] `avatar-sizes.scss` is in between, because it's a mixin but it's being used to set variables. There are also some mixins in `style/theme.scss` and `style/typography.scss`.

So it might make sense to move `avatar-sizes.scss` into `style` or to move every mixin out of `style`. There's another [PR to move variables into a separate package, #603](https://github.com/iTwin/iTwinUI/issues/603), so I lean towards bringing all the mixins out of `style`.

packages/itwinui-css/src/utils/sizes.scss
elephantcatdog: Renamed

packages/itwinui-css/src/style/variables.scss
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
```
gretanausedaite, line 2: Everything in `styles` are variables, maybe it makes sense to rename this file to something more specific?

packages/itwinui-css/src/style/table-densities.scss
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
+@import './baseline';
+@import './space';
 
 $iui-table-density-vertical: $iui-baseline * 1.5;
```
veekeys, line 6: hm... I was thinking, maybe it makes more sense to move this together with table? And all other files, if they are specific to certain component.. No point to move those to generic styles

veekeys (commented):
I will leave comment in here as github does not allow to comment on moved files..
- You are not right about text. We do have demo html page for it
https://itwin.github.io/iTwinUI/backstop/minified/typography.html
So it should not be under utils
- As I left comment, I'd consider moving these to the component folder they being used in
avatar-sizes to avatar
notification-marker to button (not sure about this one)
table-densities to table

packages/itwinui-css/src/style/table-densities.scss
elephantcatdog: That makes sense for table-densities and avatar-sizes, but notification-marker is used in multiple places, including button, side-navigation, and header.

packages/itwinui-css/src/style/variables.scss
elephantcatdog: Do you think doing something like 'misc-variables' is better, or to separate the variables out into more distinct groups like 'component-padding' and 'component-height'? I couldn't think of anything concise that represented all of the variables in this file.

gretanausedaite: - rename to `component.scss`
- move everything to `global.scss`

elephantcatdog: Thanks, done

elephantcatdog:
There are several mixin files besides mixins.scss included in the utils folder: font-family.scss, theme.scss, and ripple.scss. These could be moved into the mixins.scss file to consolidate all of the util mixins for ease of use. Alternatively, the mixins within mixins.scss could be separated into different files so that individual mixins would be referenced when needed rather than every mixin being included in a component file. Thoughts? @gretanausedaite @veekeys @bentleyvk

packages/itwinui-css/src/style/component.scss
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import './space';
 @import './baseline';
-@import '../icon/variables';
+@import './icon-sizes';
```
veekeys, line 5: I dont think its needed here

packages/itwinui-css/src/style/global.scss
```diff
@@ -49,10 +49,6 @@ html {
   }
 }
 
-.iui-anchor {
```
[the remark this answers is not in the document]
bentleyvk: Previously it was in global css then we decided to move it to its own folder but then we would break users layouts as they will need to import anchor css. So with major release we can do that now.

packages/itwinui-css/src/style/variables.scss
```diff
@@ -2,9 +2,9 @@
...
 @import './space';
 @import './baseline';
-@import '../icon/variables';
+@import '../utils/icon/variables';
 
-$iui-border-radius: 4px;
+$iui-border-radius: 3px;
```
bentleyvk, line 7: I don't think this change was intended.

packages/itwinui-css/src/style/component.scss
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
bentleyvk, line 1: Why this file was added when `variables.scss` is left?

bentleyvk:
I agree this Vykintas. Just for me `notification-marker` seems more like a separate component than an util. Also, why we can't leave `variables.scss`, why rename it to `component.scss`?

elephantcatdog:
@bentleyvk 
Greta left a comment about it above:
image

packages/itwinui-css/src/style/variables.scss
elephantcatdog: Fixed

packages/itwinui-css/src/style/component.scss
elephantcatdog: I don't know; I'm guessing an auto-merge mistake
Fixed

elephantcatdog:
The original 'rules' that Adam and I went by for putting components in `utils` were that 1) they were used by other components and 2) didn't have an .html file of their own.

The icon refactor PR has a note to move the `icon` folder to `utils`, which I've done in this PR. But `icon` has its own .html file. So, if `icon` should be considered a util then maybe the guidelines should change, and components like `text` should be reincluded. Or should the `rules` stand and not add icons?

packages/itwinui-css/src/style/component.scss
elephantcatdog: Fixed

packages/itwinui-css/src/style/global.scss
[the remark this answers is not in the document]
elephantcatdog: this seems to be okay then?

elephantcatdog:
Recent Changes:

1. Put `ripple.scss` into `utils/mixins.scss`
2. Moved `utils/theme` back into `style/theme` and `utils/font-family` back into `style/typography`
3. Renamed `component.scss` (formerly `variables.scss`) to `global-variables.scss` based on poll
4. Moved `icon` folder out of `utils` folder
5. Made imports from `utils` folder more generic - `@import '../utils/index';`

bentleyvk (approved):

gretanausedaite (approved):

packages/itwinui-css/src/style/global.scss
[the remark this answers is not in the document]
veekeys: @bentleyvk But wont it happen here, that its a breaking change for react users??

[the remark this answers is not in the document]
veekeys: I suggest leave it here like that and in react we import anchor css in `useTheme` together with global, so users dont get breaking changes. In v2 release, can get rid of that.
@mayank99 might want to be aware of this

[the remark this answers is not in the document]
mayank99: I don't think it makes sense to keep it global. If people are using this class, it's expected that something will break with v1.

mayank99:
Can we get rid of `style` folder after variables package is used in v1 branch?

packages/itwinui-css/src/style/global.scss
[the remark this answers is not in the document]
veekeys: css - yes, but react.. we are not doing there a major release, so we cant say users will have breaking change, but its not react direct fault :?

veekeys:
> Can we get rid of `style` folder after variables package is used in v1 branch?

If all the vars are there, I would assume yes

packages/itwinui-css/src/style/global.scss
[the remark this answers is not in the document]
mayank99: using a class = css, not react. They should be using our `<Anchor>` component if they want to be shielded from breaking css changes.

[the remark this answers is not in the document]
veekeys: But it will break with the react minor version release. class is implementation detail (as is our global css inclusion in react), but technically, stuff would break.

Issue #551, opened 2022-03-25
Artefact on Button border after mouse interaction.


mathieu-fournier:
#### Describe the bug (actual behavior)
After interacting (clicking/hovering) a button, with high-visibility style, in a ButtonGroup, there is a weird visual artifact that appear in  the button border.
![image](https://user-images.githubusercontent.com/82905333/160188815-81301045-c810-410b-9abf-072f497eb124.png)

#### Expected behavior
The border should stay sharp. 

#### Reproduction
Normally hovering the button triggered the artifact, but it's not consistent in codesandbox. 
I found a way to do it consistently in codesandbox by moving/hovering the mouse back and forward from the editor to the button. (Following the red line in the image below. )
![image](https://user-images.githubusercontent.com/82905333/160189918-b92f9971-ca65-4421-a41a-ac2e59ba8e73.png)

[Link to the codesandbox. ](https://codesandbox.io/s/itwinui-react-minimal-example-forked-vewoe3?file=/src/App.tsx)

#### Additional information
I noticed that the artifact is stack on itself each time I do the repro steps. (It become more opaque).
The artifact is not visible in firefox. 
It's visible in Chrome Version 99.0.4844.51 and Edge Version 99.0.1150.46

I am on windows and my display setting are : 
![image](https://user-images.githubusercontent.com/82905333/160190296-26b68daf-2697-4f78-a97d-5fa31b391a52.png)

And my google chrome browser Display setting is : 
![image](https://user-images.githubusercontent.com/82905333/160190518-11482a65-c7e4-41d4-9fd5-95452423e924.png)

Thanks (:

PR #688, opened 2022-06-15
feat(style): Border radius change

gretanausedaite:
#551 

Updated border-radius to be 4px.


adamhock:
I'm a little confused on how this change fixes the artifact issue? Is it because 4px is more divisible than 3px?

veekeys (approved):
Looks good.
The only question, do we want to merge this into v1 or main?

adamhock (approved):
I believe this should be merged into main since the issue wasn't planned for v1 and this fix can potentially be released sooner

gretanausedaite:
> I'm a little confused on how this change fixes the artifact issue? Is it because 4px is more divisible than 3px?

I'm not sure WHY this is working, but a lot of artefact or other rendering problems disappear when changed to even number of pixels (or more specifically base 2 number system). A lot of suggestions do not have explanations (eg. [this](https://stackoverflow.com/questions/12505872/3px-border-radius-has-different-look-in-ie-and-other-browsers))

My guess: It's because computer processors gets and process data in binary. This is just a guess and I did not find any info to confirm.

gretanausedaite:
> Looks good. The only question, do we want to merge this into v1 or main?

I was thinking V1 because it's a visual change. 

I can rebase my branch and merge to main if that's better and we want to deliver this faster.

PR #689, opened 2022-06-16
feat(Tile): Add new status states

### LINKED ISSUE #461: Tile & table row should have matching options / states
#### Feature
[Tiles](https://itwin.github.io/iTwinUI/backstop/minified/tile.html) & [table](https://itwin.github.io/iTwinUI/backstop/minified/table.html) rows should have matching options:
- **Uploading / disabled states.**  Table has it, tile does not.  I would also like italic text to be exclusive to "upload in progress", not tied to disabled state.
- **Error states.**. Table has it, tile does not.

We already have matching states for:
- Newly uploaded state (although the symbol displayed is different between the 2 components).
- Selected state.

We will need this when we go to deliver the table / tile iTwinUI-Layout.

marashad001:
- Added 4 additional tile states with styling from `table` (uploading, positive, negative, warning)
- Closes #461


packages/itwinui-css/src/tile/tile.scss
```diff
@@ -121,6 +138,10 @@
     text-overflow: ellipsis;
   }
 
+  .iui-tile-uploading {
```
veekeys, line 141: Naming is weird.. why not just `iui-informational`

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -121,6 +138,10 @@
...
+    font-style: italic;
```
veekeys, line 142: amm.. we sure about this?

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -97,6 +97,23 @@
   flex-grow: 2;
   min-width: 0;
 
+  .iui-disabled {
+    filter: grayscale(100%);
+    cursor: not-allowed;
+  }
+
+  .iui-negative {
```
veekeys, line 105: The expectation is, that you would need to put tile status class only on root and it would get styling.. With your current code, you need to put on all the sections, which is too much.
Try moving these up. I have a feeling, that setting color of text on root might be enough for it to be inherried in all places.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -175,9 +196,31 @@
...
+    filter: grayscale(100%);
+    cursor: not-allowed;
+    border-bottom: 4px solid t(iui-color-background-border);
+  }
+
   .iui-thumbnail-icon {
     @include iui-icons-3xl;
   }
+
+  &.iui-negative {
```
veekeys, line 209: These also might be not needed...
Instead, you could have a private variable (i.e. `_iui-tile-thumbnail-border-bottom`) and set its different values on root

```
...
&.iui-negative {
  _iui-tile-thumbnail-border-bottom: 4px solid var(--iui-icons-color-negative);
}
```

So you would move up the default value from line 196 and update it per different status class

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -175,9 +196,31 @@
     border-bottom: 1px solid t(iui-color-background-border);
   }
 
+  &.iui-disabled {
+    filter: grayscale(100%);
+    cursor: not-allowed;
+    border-bottom: 4px solid t(iui-color-background-border);
```
veekeys, line 202: this should not even work as you are using `t` without `themed` mixin...
Are you sure about border of 4px?
Could use variable for it too (`$iui-xs`)

veekeys (commented):
Adding new statuses should not change all tile images.. Need to review what has changed or what influenced the changes and possibly revert..
Also, maybe put all status tiles in new section so we get only new images and test those cases instead of updating current one

gretanausedaite (commented):
I agree with all Vykintas comments. We should set status (informational, warning, negative, positive) on root `iui-tile`.

packages/itwinui-css/src/tile/tile.scss
marashad001: Fixed this

packages/itwinui-css/src/tile/tile.scss
marashad001: Yes, italics were requested in the original issue.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -97,6 +97,23 @@
   flex-grow: 2;
   min-width: 0;
 
+  .iui-disabled {
```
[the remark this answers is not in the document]
marashad001: I don't know, I was really focused on trying to make the tiles and keep it standard with the classes. Probably would've made revising the code a lot easier haha

packages/itwinui-css/src/tile/tile.scss
marashad001: Fixed

packages/itwinui-css/src/tile/tile.scss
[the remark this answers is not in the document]
veekeys: well, I am not a fan of attributes, just I noticed some of the new changes are being done using those, so inconsistencies start to appear everywhere :(
Keep it with classes, I guess

packages/itwinui-css/src/tile/tile.scss
veekeys: I dont think you did...
Specificity is higher with current solution. Variables could help solve that.

packages/itwinui-css/src/tile/tile.scss
veekeys: @FlyersPh9 We do not use italics in any other `informational` components. Whats the difference here?

packages/itwinui-css/src/tile/tile.scss
marashad001: Okay, now I understand what you are saying, let me try to implement it

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -87,6 +87,35 @@
       -webkit-line-clamp: 2;
     }
   }
+
+  //#region Statuses
+  &.iui-uploading {
```
bentleyvk, line 92: Maybe use more generic name like `iui-progress`

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -299,6 +352,7 @@ <h2>Everything turned on</h2>
           <button class="iui-button iui-default">Only button</button>
         </div>
       </div>
+
```
bentleyvk, line 355: Remove so it won't show as unnecessary change

packages/itwinui-css/src/tile/tile.scss
marashad001: Fixed

bentleyvk: Forgot to push? 😀

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -505,10 +559,320 @@ <h2>Default</h2>
           </div>
         </div>
       </div>
+
```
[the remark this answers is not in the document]
marashad001: Fixed

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -505,10 +559,320 @@ <h2>Default</h2>
...
+          </div>
+
+        </div>
+
+      </div>
+
+      <div
+        class="iui-tile iui-actionable"
+        tabindex="0"
```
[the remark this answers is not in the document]
marashad001: fixed

gretanausedaite (approved):
Code part looks good. Test images need to be fixed.

veekeys (approved):
@FlyersPh9 verify the visuals here.
I agree with @bentleyvk, you should recheck images.
Else looks ok.

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -503,8 +556,316 @@ <h2>Default</h2>
...
+                  cy="20"
+                  r="15.91549"
+                  style="stroke-dashoffset: 20"
+                ></circle>
+              </svg>
+            </div>
+            <span class="iui-tile-name-label">Upload in progress&mldr;</span>
+
+          </div>
+
```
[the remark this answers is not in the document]
marashad001: fixed

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: We've removed the tile with the blue stripe, we only include positive, warning, negative statuses.  The italics is tied to "upload in progress" which [I would like to separate from disabled state in table as well](https://github.com/iTwin/iTwinUI/pull/697#issuecomment-1184764342).

Screen Shot 2022-07-14 at 2 26 54 PM

PR #693, opened 2022-06-21
chore: Review radio tile responsive design

### LINKED ISSUE #322: RadioTiles: review responsive design, border issues and group selector
#321 mentions:

> We have bug now because of our borders (see picture below), so will create new item to change borders approach (outline or need to figure out something else...).
Also, border radius now looks weird...
> 

Also, if want to add tooltip, need to wrap in a div every RadioTile. Doing this, brakes selector of group. Need to update those.

elephantcatdog:
Problems:
1. border-radius looks odd when radio tiles are stacked
2. borders on radio tiles do not overlap when stacked
3. tooltips don't work when using group selectors (`last-of-type`, etc.)

Solutions:
1. removed border-radius entirely. difficult (impossible?) to determine which tile is where when they can become stacked
  - exception is perhaps using `grid`? You'd still need to know how many columns there are

2. current solution is using `outline`. this isn't perfect because there's small overlap when the radio tile is hovered on or selected ...however this is in the current version of radio tile. <b>*Update: Fixed the small overlap by using `gap: 1px`. I've noticed two small problems with the current solution: the interior lines look brighter because of the overlap (because we use colors with opacity). Also when a checked and disabled tile is adjacent to an undisabled tile the thin borders overlay the checked border and it looks bad. I don't anticipate this scenario happening, but who knows.*</b>
  - also tried:
    - border: doesn't work with border-box for double vertical borders for some reason (trying margin: -1px, etc.)
    - border with content-box: similar issue to using outline, but with the negative of switching to content-box
    - box-shadow: this looks great! except that we'd have to use colors without opacity for the borders and I couldn't find a solution in line with current practices. also changes tile dimensions a bit
  - thought about but haven't tried yet:
    - grid: overpowered for the problem-space? might work well, but could just cause the same problem with outline and would continue the issues with group selectors
    - gap: can't change color of border for a single radio tile and would need to change tactics to highlight the tile itself rather than its borders
    
4. stopped using group selectors
  - added tooltips on new example

Also:
- added new examples and tests for 'alternative layouts'

Also found new problem:
image
Is this expected behavior or should all tiles in a container be the same height?

closes #322


packages/itwinui-css/backstop/tests/radio-tile.html
```diff
@@ -27,6 +27,15 @@
...
         gap: 11px;
+        margin-bottom: 11px;
+      }
+      .right {
+        top: 0;
+        left: 100%;
+        margin-left: 5px;
+      }
+      .iui-input-container {
+        margin: 2px;
```
elephantcatdog, line 38: This margin allows the test images to capture the outlines on the edges of the images. I think they aren't captured correctly without this margin due to the difference between placement of margin vs border.

packages/itwinui-css/src/radio-tile/radio-tile.scss
```diff
@@ -134,9 +117,20 @@
...
+  }
+
+  @include focus-visible {
+    + * { // outline for checkmark
+      z-index: 2;
+      outline-offset: -2px;
+      outline: 3px solid var(--iui-color-foreground-primary);
+      @media (forced-colors: active) {
+        outline-color: Highlight;
+      }
```
elephantcatdog, line 133: `focus-visible` was moved so that its rules have precedence over the other rules.

packages/itwinui-css/backstop/tests/radio-tile.html
veekeys: Could you add custom class instead? So we know its for demo purposes

packages/itwinui-css/src/radio-tile/radio-tile.scss
```diff
@@ -59,22 +49,15 @@
...
-      }
-    }
-  }
 
   &:checked + * {
-    padding: $iui-s - 1;
+    padding: $iui-s;
     z-index: 3;
-    border: 2px solid var(--iui-color-foreground-primary);
+    outline: 2px solid var(--iui-color-foreground-primary);
```
veekeys, line 56: minor, but could you variable `iui-xxs` for 2px

veekeys (commented):
All looks good. Left some minor findings.

packages/itwinui-css/backstop/tests/radio-tile.html
```diff
@@ -343,7 +380,7 @@ <h1>Radio tile</h1>
               ></svg-network>
               <div class="iui-radio-tile-label">Web</div>
               <div class="iui-radio-tile-sublabel">
-                Dimensions in px and % CSS and HTML exports, PNG and SVG assets
+                Dimensions in px and %, CSS and HTML exports, PNG and SVG assets
```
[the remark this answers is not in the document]
elephantcatdog: The image is going to change anyway because of the margin I had to add for `iui-import-container`.

packages/itwinui-css/backstop/tests/radio-tile.html
```diff
@@ -27,6 +27,15 @@
         display: flex;
         flex-direction: column;
         gap: 11px;
+        margin-bottom: 11px;
+      }
+      .right {
```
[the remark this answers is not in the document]
elephantcatdog: Ah, thanks. I took this from tooltip.html. Fixed.

packages/itwinui-css/src/radio-tile/radio-tile.scss
elephantcatdog: I changed it in lines 56, 120, and 129-130. I went back and forth on leaving lines 129-130 as is or changing them. If you think it's better the other way let me know. (And also your reasoning.)

packages/itwinui-css/backstop/tests/radio-tile.html
elephantcatdog: Done

packages/itwinui-css/src/radio-tile/radio-tile.scss
veekeys: yeah, leave 3px instead, agree

FlyersPh9 (approved):
Took a call w/ @elephantcatdog to make a change to `forced-colors` on the primary radio tile label.  I couldn't leave a comment on this because it was not a change made within this PR.  Reviewed everything else and looks good to me.

PR #697, opened 2022-06-28
feat(Table): Class refactor

### LINKED ISSUE #610: [V1] Table refactor
Review current classes and structure, see if we can do some minimal changes to improve it, without taking it too long.
Do not break react!

adamhock:
- Reduced `condensed.scss` and `extra-condensed.scss` to `condensed-sizes.scss`
- Moved `variables.scss` to `table.scss`
- Removed unnecessary use of `@media-transition` in `table.scss`
- Separated `iui-paginator-page-size-label` into its own mixin

TODO:

- [ ] See if `iui-button-bar` can be reduced in `iui-column-filter` without breaking React
- [ ] Use `@at-root` for some mixins if possible

Closes #610


packages/itwinui-css/src/table/classes.scss
```diff
@@ -22,6 +20,10 @@
   @include iui-table-header;
 }
 
+.iui-row {
```
bentleyvk, line 23: I thought we were planning to prefix everything with `iui-table-*`

packages/itwinui-css/src/table/condensed-size.scss
```diff
@@ -0,0 +1,28 @@
...
+    --_iui-table-condensed-header-size: #{$iui-baseline * 4};
+    --_iui-table-condensed-row-size: #{$iui-baseline * 4 + 2px};
+  } @else {
+    --_iui-table-condensed-header-size: #{$iui-baseline * 3};
+    --_iui-table-condensed-row-size: #{$iui-baseline * 3 + 2px};
+  }
+
+  .iui-table-header .iui-cell,
+  .iui-paginator {
+    min-height: var(--_iui-table-condensed-header-size);
```
bentleyvk, line 16: These are duplicated in the compiled CSS so it lost the point of using CSS variables.
Maybe just hardcode sizes in this file:
```scss
@mixin iui-table-condensed-size() {
  .iui-table-condensed {
    --_iui-table-condensed-header-size: #{$iui-baseline * 4};
    --_iui-table-condensed-row-size: #{$iui-baseline * 4 + 2px};
  } 

  .iui-table-extra-condensed {
    --_iui-table-condensed-header-size: #{$iui-baseline * 3};
    --_iui-table-condensed-row-size: #{$iui-baseline * 3 + 2px};
  }

  .iui-table-header .iui-cell,
  .iui-paginator {
    min-height: var(--_iui-table-condensed-header-size);
  ...
```

packages/itwinui-css/src/table/table.scss
```diff
@@ -2,7 +2,13 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../style/index';
 @import '../icon/index';
-@import './variables';
+
+$iui-table-density-vertical: $iui-baseline * 1.5;
```
bentleyvk, line 6: What is the idea of moving variables here? Better visibility?

packages/itwinui-css/src/table/condensed-size.scss
adamhock: I tried using your suggested code but ran into some issues when dealing with tables that weren't condensed. I tweaked the file a bit to remove duplication in the CSS.

packages/itwinui-css/src/table/table.scss
adamhock: This change is just for consistency. Other components such as Badge, Color Picker, Date Picker, Menu, Slider, Toggle Switch, and Tree list their variables at the top of the file instead of having a separate file for them.

packages/itwinui-css/src/table/table.scss
```diff
@@ -215,22 +183,22 @@
...
 
-    .iui-row-expander > .iui-button-icon {
+    .iui-table-row-expander > .iui-button-icon {
       @media (prefers-reduced-motion: no-preference) {
         transition: transform $iui-speed-fast ease-out;
       }
     }
 
-    &.iui-row-expanded {
+    &.iui-table-row-expanded {
```
bentleyvk, line 192: Now the question is to what extent we want to prefix with `iui-table-*` 😀 I guess here find replace all did the job. Maybe we should prefix all elements but not modifiers.

packages/itwinui-css/src/table/condensed-size.scss
bentleyvk: Is it possible to set default sizes too, so that specificity would be even lower because it is kind of the same as before.

adamhock: Default values have been updated. The test images for condensed tables have changed as a result, but I believe it is for the better.

In our current implementation of condensed tables, the `.iui-table-header .iui-cell` selector was being overriden by the `.iui-row .iui-cell` selector even on cells that were part of the header. Look at the following example of a header cell of an extra condensed table

Extra condensed screenshot



Now, our calculated value for the `min-height` for table headers in condensed tables is `2px` less than the `min-height` for the body cells. This is consistent with our non-condensed table

There are other slight visual changes as a result of this push, but I'm not entirely sure what's going on there.

bentleyvk:
Tests are failing because you need to get this change https://github.com/iTwin/iTwinUI/pull/702 into your branch.

packages/itwinui-css/src/information-panel/information-panel.scss
```diff
@@ -22,16 +22,16 @@
       opacity $iui-speed-fast ease;
   }
 
-  > .iui-resizer {
+  > .iui-table-resizer {
```
bentleyvk, line 25: What it has to do with table?

packages/itwinui-css/src/table/classes.scss
```diff
@@ -30,34 +27,38 @@
...
 }
 
-.iui-cell-shadow-right {
+.iui-table-cell-shadow-right {
   @include iui-table-cell-shadow-right;
 }
 
-.iui-paginator {
-  @include iui-paginator;
+.iui-table-paginator {
```
bentleyvk, line 42: I don't remember whether we intentianally named it without prefix. Maybe @FlyersPh9 you remember?

packages/itwinui-css/src/table/paginator.scss
```diff
@@ -4,14 +4,12 @@
...
 @import '../text/index';
-@import './variables';
 
-@mixin iui-paginator {
+@mixin iui-table-paginator {
   @include iui-reset;
   display: flex;
   justify-content: space-between;
   align-items: center;
-  min-height: $iui-baseline * 5;
```
bentleyvk, line 14: Why removed?

FlyersPh9:
#689 is separating the "upload in progress" and "disabled" states.  Can we include that in this PR as well?

- **Upload in progress:** italic text, radial progress indicator, progress cursor.
- **Disabled:** muted text color, grayscale icon, not-allowed cursor.

Screen Shot 2022-07-14 at 2 23 24 PM

packages/itwinui-css/src/table/classes.scss
FlyersPh9: I suppose we could have a paginator on a page with tiles, but I think it might look out of place with the darker background color that it currently has.

packages/itwinui-css/src/table/condensed-size.scss
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../style/index';
+
+@mixin iui-table-condensed-size {
```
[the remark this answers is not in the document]
FlyersPh9: I've got a proposal:
```scss
/// @arg size - must be one of: default, condensed, extra-condensed
@mixin iui-table-size($size: default) {
  @if $size == default {
    --_iui-table-header-size: #{$iui-baseline * 5};
    --_iui-table-row-size: #{$iui-baseline * 5 + 2px};
  }
  @if $size == condensed {
    --_iui-table-header-size: #{$iui-baseline * 4};
    --_iui-table-row-size: #{$iui-baseline * 4 + 2px};
  }
  @if $size == extra-condensed {
    --_iui-table-header-size: #{$iui-baseline * 3};
    --_iui-table-row-size: #{$iui-baseline * 3 + 2px};
  }
}
```

@adamhock I'll get with you tomorrow to collab on this.

packages/itwinui-css/src/table/table.scss
```diff
@@ -136,107 +133,77 @@
...
     }
 
     &:hover:not(.iui-disabled) {
-      .iui-cell {
+      .iui-table-cell {
         background: linear-gradient(
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6)),
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6))
-        ),
+            rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6)),
```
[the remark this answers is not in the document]
adamhock: Fixed. Not sure what happened there

packages/itwinui-css/src/information-panel/information-panel.scss
adamhock: Incidental issue when using find/replace. Fixed

packages/itwinui-css/src/table/classes.scss
adamhock: From your previous comment

> Maybe we should prefix all elements but not modifiers.

I figured `iui-paginator` was an element, so I prefixed it with `iui-table`. I assume when you mean "modifiers" you are talking about things like `iui-slot` and `iui-sorted` which I remained untouched

bentleyvk:
> #689 is separating the "upload in progress" and "disabled" states. Can we include that in this PR as well?
> 
> * **Upload in progress:** italic text, radial progress indicator, progress cursor.
> * **Disabled:** muted text color, grayscale icon, not-allowed cursor.
> 
> Screen Shot 2022-07-14 at 2 23 24 PM

Please don't, it is completely unrelated so leave it in another PR. Also, this PR is almost finished and with suggested changes it will probably would take another 2-3 weeks.

packages/itwinui-css/src/table/condensed-size.scss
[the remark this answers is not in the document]
bentleyvk: But in reality it does the same 😀 Also, since this mixing would be used only in Table, it kind of looses the need for reusability.

gretanausedaite (approved):

bentleyvk:
Are you planning to work on these conflicts?

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -751,22 +751,22 @@ <h2>Default</h2>
       <div class="iui-table iui-condensed">
```
FlyersPh9, line 751: Since this is a v1.0 PR can we replace the modifiers with data attributes?
```suggestion
      <div class="iui-table" data-iui-size="condensed">
```

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -939,8 +939,8 @@ <h2>Default</h2>
             </div>
           </div>
 
-          <div class="iui-row iui-selected">
-            <div class="iui-cell iui-slot">
+          <div class="iui-table-row iui-selected">
```
FlyersPh9, line 942: Because this is a v1.0 PR we could use `aria-selected` instead of a class modifier.
```suggestion
<div class="iui-table-row" aria-selected="true">
```
You'll need to set `aria-multiselectable="true"` on one of the parents, you can read about it [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#description).

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -1253,46 +1253,46 @@ <h2>Default</h2>
             </div>
           </div>
 
-          <div class="iui-row iui-negative">
-            <div class="iui-cell iui-slot">
+          <div class="iui-table-row iui-negative">
```
FlyersPh9, line 1256: We could use `data-iui-status` on both rows and cells.
```suggestion
          <div class="iui-table-row" data-iui-status="negative">
```

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -1253,46 +1253,46 @@ <h2>Default</h2>
...
               </div>
             </div>
             <div
-              class="iui-cell iui-negative"
+              class="iui-table-cell iui-negative"
               title="Jul 24, 2020 2:10pm"
             >A few minutes ago</div>
-            <div class="iui-cell iui-disabled">
-              <div class="iui-cell-start-icon">
+            <div class="iui-table-cell iui-disabled">
```
FlyersPh9, line 1283: Could use:
```suggestion
            <div class="iui-table-cell" aria-disabled="true">
```

adamhock: `iui-disabled` has been replaced with `aria-disabled='true'`

packages/itwinui-css/backstop/tests/table.html
adamhock: `iui-positive/negative/warning` has been replaced with `data-iui-status='positive/negative/warning'`

packages/itwinui-css/backstop/tests/table.html
adamhock: `iui-selected` has been replaced with `aria-selected='true'`.

For tables with selectable rows, the `iui-table-body` element now has `aria-multiselectable='true'`

packages/itwinui-css/backstop/tests/table.html
adamhock: `iui-condensed/extra-condensed` has been replaced with `data-iui-size='condensed/extra-condensed'`

bentleyvk (approved):

PR #701, opened 2022-07-08
feat(Icon): Separate color from icon sizes & status color from default color.

elephantcatdog:
Changes I made:

<b>Separated icon size mixins.</b> There are now mixins for sizes (iui-icons-default/small/large/xl/xxl/3xl) separate from two new categories of mixins: color and display. I made one mixin called `iui-icons-display` that contains `display: inline-flex; flex-shrink: 0;`. I made two mixins for color. One is called `iui-icons-color` and is just for the default gray color. The other is called `iui-icons-color-status` and contains the colors for the different statuses, and `@include`s the mixin `iui-icons-color` for the default color.

<b>Added `flex-shrink: 0`.</b> This is part of the mixin `iui-icons-display`.

The two above changes allowed me to go through the code and consolidate some css for icons that were just changing the size, or just the color, etc. The vast majority of the file changes are due to this change.

<b>Within the `@itwinui-icons` repo:</b> I added `width: 16px`, `height: 16px`, and `fill: currentColor` attributes. The size is 16px because that's the default size. This can be found in the branch `leah/icons-refactor` on the `@itwinui-icons` repo.

<b>Move to utils folder:</b> Will be much easier after the [PR for folder structure](https://github.com/iTwin/iTwinUI/pull/687) is finished.

Changes I need to make:

<b>Icon css will need to be changed to target the svg instead of the svg container.</b>

<b>Targeting icons in `@itwinui-icons-elements` repo with common selectors</b>

<b>Targeting icons in `@itwinui-css` with data attributes</b>

Reference:
image


elephantcatdog:
I added everyone as a reviewer so you can all see this PR.

packages/itwinui-css/src/icon/mixins.scss
```diff
@@ -3,50 +3,45 @@
...
 }
 
 @mixin iui-icons-3xl {
-  @include iui-icons-color();
-  display: inline-flex;
   width: $iui-icons-3xl;
   height: $iui-icons-3xl;
 }
 
+@mixin iui-icons-display {
```
bentleyvk, line 35: Is it bad if we include this to every icon? Because it feels that it should be used for every icon.

packages/itwinui-css/src/icon/mixins.scss
```diff
@@ -3,50 +3,45 @@
 @import '../style/index';
 
 @mixin iui-icons-default {
```
bentleyvk, line 5: Maybe we should consider renaming it? Because now when there is color and display mixins, default doesn't tell anything, I had to look at the implementation what it does 😀 Maybe rename it to `-medium` or smth, but that is just my opinion.

bentleyvk:
> Within the @itwinui-icons repo: I added width: 16px, height: 16px, and fill: currentColor attributes. The size is 16px because that's the default size.

Won't this be kind of a breaking change for the icons package? Because of taking all available space, now icons by default be only 16x16.

packages/itwinui-css/src/icon/mixins.scss
veekeys: Or we can add `size` suffix to make it clear.
But `medium` could work too.

packages/itwinui-css/src/information-panel/information-panel.scss
```diff
@@ -63,12 +63,11 @@
...
 
       svg {
-        display: flex;
-        width: $iui-icons-large;
-        height: $iui-icons-large;
+        @include iui-icons-large;
+        @include iui-icons-color;
         margin-right: $iui-s;
+        display: flex;
         flex-shrink: 0;
```
veekeys, line 70: should be able to reuse `iui-icons-display` here and get rod of `display` and `flex-shrink` rules

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -157,12 +157,11 @@
 
 /// Icon to be shown near a message.
 @mixin iui-input-icon {
+  @include iui-icons-default;
+  @include iui-icons-color;
   display: flex;
```
veekeys, line 162: `iui-icons-display` also might be good here

packages/itwinui-css/src/icon/mixins.scss
veekeys: yeah, Was thinking the same..
But then if parent of icon is not `display: flex`, `flex-shrink` rule would do nothing :/

veekeys (commented):
If we decide that all our icons should have flex display  and shrink rule, we might want to have a single mixin/class for combo of those to simplify usage. Color is also up for discussion (just its applicable only to svg).
And I guess would need to see how these changes look if we merge the ones in icons repo, cause then it will be different here.

packages/itwinui-css/src/icon/mixins.scss
bentleyvk: But it is not a big problem if it does nothing 😀

packages/itwinui-css/src/icon/mixins.scss
bentleyvk: I also was thinking we could have `size` word in the name.

FlyersPh9: Actually, here's an idea I like better:
```scss
/// @arg $size Can be one of: s, m, l, xl, 2xl, 3xl
@mixin iui-icon-size($size: m) {
	@if $size == 's' {
	  width: $iui-icons-small;
	  height: $iui-icons-small;
	}
	@if $size == 'm' {
	  width: $iui-icons-default;
	  height: $iui-icons-default;
	}
	@if $size == 'l' {
	  width: $iui-icons-large;
	  height: $iui-icons-large;
	}
	@if $size == 'xl' {
	  width: $iui-icons-xl;
	  height: $iui-icons-xl;
	}
	@if $size == '2xl' {
	  width: $iui-icons-xxl;
	  height: $iui-icons-xxl;
	}
	@if $size == '3xl' {
	  width: $iui-icons-3xl;
	  height: $iui-icons-3xl;
	}
}
```

packages/itwinui-css/src/icon/mixins.scss
elephantcatdog: This can't be used for every icon. It causes a lot of bad changes. Not every icon/element is currently using `display: inline-flex`, and adding this mixin to every icon breaks a lot of test images.

packages/itwinui-css/src/icon/mixins.scss
elephantcatdog: Implemented. Mostly

packages/itwinui-css/backstop/tests/icon.html
```diff
@@ -105,22 +105,22 @@ <h2>Actionable icons</h2>
     <h2>Icons with status</h2>
     <section id="demo-status">
       <svg-info-circular
-        class="iui-icons-default iui-informational"
+        class="iui-icons-default iui-icons-color-status iui-informational"
```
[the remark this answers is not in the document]
elephantcatdog: It didn't need it. I can add it, but it doesn't change the way it looks.

packages/itwinui-css/src/information-panel/information-panel.scss
elephantcatdog: `iui-icons-display` uses `display: inline-flex`, not `flex`. This change breaks the test images.

packages/itwinui-css/src/input-container/input-container.scss
elephantcatdog: Same as above: `iui-icons-display` uses `display: inline-flex`, not `flex`. This change breaks the test images.

packages/itwinui-css/src/icon/mixins.scss
```diff
@@ -3,50 +3,45 @@
...
 @mixin iui-icons-3xl {
-  @include iui-icons-color();
-  display: inline-flex;
   width: $iui-icons-3xl;
   height: $iui-icons-3xl;
 }
 
+@mixin iui-icons-display {
+  display: inline-flex;
+  flex-shrink: 0;
```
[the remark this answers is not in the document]
elephantcatdog: This mixin can't be used for every icon. It causes a lot of bad changes. Not every icon/element is currently using `display: inline-flex`, and adding this mixin to every icon breaks a lot of test images.

I separated `display: inline-flex` and `flex-shrink: 0` so that the icon size mixins can be used in more places. Putting them together again means that I can't do that.

bentleyvk:
> @bentleyvk
> 
> > Within the @itwinui-icons repo: I added width: 16px, height: 16px, and fill: currentColor attributes. The size is 16px because that's the default size.
> 
> > Won't this be kind of a breaking change for the icons package? Because of taking all available space, now icons by default be only 16x16.
> 
> Can you discuss this with @mayank99? This is his suggestion to decrease the amount of css needed, and I'm not very familiar with the icons package yet. Is it bad if it is a breaking change?

I think we discussed this already last week and it should be ok.

packages/itwinui-css/src/information-panel/information-panel.scss
veekeys: I would like to see those changes. Lets meet up about this one.

elephantcatdog:
With help from Vykintas, I incorporated `iui-icon-display` into `iui-icon-size`. In doing so, I renamed `iui-icon-size` to `iui-icon-style` to represent the contents of the mixin a bit better, removed `iui-icon-display`, and fixed some issues that came up with the new changes.

There are now quite a few new test images. I did not fix the button images (or any that contain button, including header) because the icons now look more aligned than before. I also did not fix the icons for the radios for the same reason. Location marker and slider both got fixes that changed the height of the examples a little.

FlyersPh9:
> There are now quite a few new test images. I did not fix the button images (or any that contain button, including header) because the icons now look more aligned than before.

Those changes are likely matching the changes in #712.  Should be good to push those new test images.

packages/itwinui-css/backstop/tests/icon.html
```diff
@@ -60,32 +60,32 @@ <h1>Icon</h1>
     <h2>Non-actionable icons</h2>
     <section id="demo-icons">
       <svg-imodel-hollow
-        class="iui-icons-3xl"
+        class="iui-icon-style-3xl iui-icon-display iui-icon-color"
```
bentleyvk, line 63: Forgot to remove `iui-icon-display`?

packages/itwinui-css/backstop/tests/icon.html
```diff
@@ -105,22 +105,22 @@ <h2>Actionable icons</h2>
     <h2>Icons with status</h2>
     <section id="demo-status">
       <svg-info-circular
-        class="iui-icons-default iui-informational"
+        class="iui-icon-style-m iui-icon-color-status iui-informational"
```
bentleyvk, line 108: For me `iui-icon-color-status` class looks a bit odd as it doesn't do anything. What if we just have status classes only like `iui-icon-status-informational`?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -18,14 +17,19 @@ $tick-height: $iui-baseline; // 11px
...
+      @include iui-icon-color;
       margin-top: $iui-xs;
     }
   }
 
   .iui-slider-min {
     margin-right: $iui-sm;
     text-align: right;
+
+    svg {
```
bentleyvk, line 30: Why this change is needed?

packages/itwinui-css/src/carousel/carousel.scss
```diff
@@ -68,10 +68,8 @@
     }
 
     &::after {
+      @include iui-icon-style('s');
       content: '';
-      display: block;
```
adamhock, line 72: Why was `display: block` removed if `iui-icon-style` has `display: flex`? I saw the same thing in `radio-tile.scss` as well

packages/itwinui-css/src/checkbox/checkbox.scss
```diff
@@ -65,10 +64,9 @@
   --_iui-checkbox-background-color: var(--iui-color-background-1);
   --_iui-checkbox-mask-image: initial;
 
+  @include iui-icon-style('m');
```
adamhock, line 67: Because the `iui-icon-style` mixin includes `flex-shrink: 0`, we can remove it from this class

packages/itwinui-css/src/header/header-buttons.scss
```diff
@@ -248,15 +248,15 @@
 }
 
 @mixin iui-header-breadcrumb-button-icon {
```
adamhock, line 250: Can the `iui-icon-style` mixin be used for the icons in `header-buttons.scss`?

FlyersPh9:
I reviewed each image 1 by 1, here are my 2 comments:

1. Header's center zone search icon seems to have disappeared.
2. The location marker test image is so wide it makes it hard to compare, can we set `viewports: [{ width: 800, height: 600 }],` since the test images are changing anyways?

Aside from that, all other changes that I am seeing appear to be icons being more correctly vertically centered. 👍

packages/itwinui-css/src/utils/icon/icon.scss
```diff
@@ -0,0 +1,55 @@
...
+    width: $iui-icon-2xl;
+    height: $iui-icon-2xl;
+  }
+  @if $size == '3xl' {
+    width: $iui-icon-3xl;
+    height: $iui-icon-3xl;
+  }
+}
+
+@mixin iui-icon-color {
```
FlyersPh9, line 35: Suggestion to replace `@mixin iui-icon-color` and `@mixin iui-icon-color-status` with this:
```scss
/// @arg $color Can be one of: informational, positive, warning, negative, or left blank for default color
@mixin iui-icon-color($color) {
  @if $color == 'informational' {
    fill: var(--iui-icon-color-primary);
  }
  @if $color == 'positive' {
    fill: var(--iui-icon-color-positive);
  }
  @if $color == 'warning' {
    fill: var(--iui-icon-color-warning);
  }
  @if $color == 'negative' {
    fill: var(--iui-icon-color-negative);
  }
  @else {
    fill: var(--iui-icon-color);
  }
}
```

Which can be used in `classes.scss` like this:
```scss
.iui-icon-color {
  @include iui-icon-color;
}
.iui-icon-color-informational {
  @include iui-icon-color(informational);
}
.iui-icon-color-positive {
  @include iui-icon-color(positive);
}
.iui-icon-color-warning {
  @include iui-icon-color(warning);
}
.iui-icon-color-negative {
  @include iui-icon-color(negative);
}
```

That would keep our classes a little bit cleaner:
```diff
- class="iui-icon-style-m iui-icon-color-status iui-informational"
+ class="iui-icon-style-m iui-icon-color-informational"
```

packages/itwinui-css/backstop/tests/icon.html
FlyersPh9: I agree, my suggestion shared [here](https://github.com/iTwin/iTwinUI/pull/701#discussion_r932362274).

packages/itwinui-css/src/utils/icon/icon.scss
elephantcatdog: How would this work for the places like this?
image

packages/itwinui-css/src/carousel/carousel.scss
elephantcatdog: I was testing to see if `@include iui-icon-style('s')` should go before or after `display: block` to see if the display type mattered, and it worked with `display: flex` so instead of putting the `include` after `display: block` I just removed it.

packages/itwinui-css/src/checkbox/checkbox.scss
elephantcatdog: Thanks, fixed

packages/itwinui-css/src/header/header-buttons.scss
elephantcatdog: Yes, thank you. I also changed the width/height in select.scss.

packages/itwinui-css/backstop/tests/icon.html
elephantcatdog: Yes, thank you. Fixed

packages/itwinui-css/backstop/tests/icon.html
elephantcatdog: In cases where `@include iui-icon-color-status` is used, wouldn't I then have to `include` every type?

packages/itwinui-css/src/slider/slider.scss
elephantcatdog: This targets only the svg on the left side of the slider and makes it flush against the slider. Without this change, the slider looks like this: 
image
vs the correct: 
image

This is needed because of the change to include `display: flex` in the icon size.

packages/itwinui-css/src/utils/icon/icon.scss
FlyersPh9: Leah and I just had a call about this, I can think of 2 or 3 different ways we could handle this & would like to make a team decision in our call tomorrow.

One option is to place this in place of all instances of `@include iui-icon-color-status`:
```scss
.iui-tile-status-icon {
  @include iui-icon-style('m');
  @include iui-icon-color;
  margin-right: $iui-s;

  &.iui-icon-color-informational {
    @include iui-icon-color(informational);
  }
  &.iui-icon-color-positive {
    @include iui-icon-color(positive);
  }
  &.iui-icon-color-warning {
    @include iui-icon-color(warning);
  }
  &.iui-icon-color-negative {
    @include iui-icon-color(negative);
  }
}
```

Another option is we have a mixin to handle all class modifiers (example: `.iui-tile-status-icon .iui-informational`) that is _not_ used in `classes.scss`.  I think this is my least favorite option but it also introduces the least amount of changes for pre-existing components like tile & expandable block.

The last option I can think of is to use data attributes for icon color.  Because the data attribute is specific (`data-iui-icon-color` VS `data-iui-color`) it doesn't need to be nested under a parent class.  This would remove the need for `@include iui-icon-color-status` inside expandable block & tile.
```scss
[data-iui-icon-color] {
  @include iui-icon-color;
}

[data-iui-icon-color='informational'] {
  @include iui-icon-color(informational);
}

[data-iui-icon-color='positive'] {
  @include iui-icon-color(positive);
}

[data-iui-icon-color='warning'] {
  @include iui-icon-color(warning);
}

[data-iui-icon-color='negative'] {
  @include iui-icon-color(negative);
}
```

elephantcatdog: I think if we use data attributes for color, we should use them for icon size also.

packages/itwinui-css/backstop/tests/icon.html
bentleyvk: Every status will be included together with `@include iui-icon-color-status`

packages/itwinui-css/src/utils/icon/icon.scss
bentleyvk: I would disagree with having such mixins. More code - same result because you need to explicitly enumerate all status colors where with current mixin you just include it and have all the colors available.

FlyersPh9 (approved):

bentleyvk (approved):

elephantcatdog:
Before:
image
After:
image

PR #702, opened 2022-07-11
tests: Bump chromium version to 103

mayank99:
Fixes pipeline which started failing recently: https://github.com/iTwin/iTwinUI/runs/7282461001?check_suite_focus=true


packages/itwinui-css/backstop/docker/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:16.14.2-bullseye
 ENV BACKSTOPJS_VERSION=6.0.4
 
 # from https://packages.debian.org/bullseye/chromium
-ENV CHROMIUM_VERSION=99.0.4844.74-1~deb11u1
```
veekeys, line 6: And we cannot use latest, for possible random images failures, right..?

mayank99: you mean automatically? yeah tests will randomly start failing whenever there is a new version

veekeys (approved):

PR #703, opened 2022-07-11
fix(Tabs): add styling for button within tab list

### LINKED ISSUE #696: Tabs: handle button as tab
Currently, we have hardcoded size for button to take full width in vertical tabs demo html:
image

And we do not support button in horizontal tabs
image

Maybe we want to add possibility to have button as tab label in our css

marashad001:
- Removed hardcoded styling for `.iui-vertical li .iui-button` in `tabs.html`
- Set `button` to 100% height in horizontal tabs / 100% width in vertical tabs.  Is targeting `.iui-button`.

Closes #696


FlyersPh9:
> These modifiers are going to be hard to consume in iTwinUI-react. What was the reason we wanted to handle this in iTwinUI-css rather than leave it up to the user? It's just one line of code so I feel like it won't be worth it.

When @veekeys did #681 we realized the "Add tab" button was not stretching to full height in horizontal tabs but it was stretching to full width in vertical tabs due to CSS that was included in the `<head>` of the demo page.
Screen Shot 2022-07-12 at 12 21 33 PM
Screen Shot 2022-07-12 at 12 21 54 PM

This change applies the styling for both horizontal and vertical scenarios & places it within the SCSS rather than in the `<head>`.  We used `.iui-tabs-button` because I thought this change was going into the `v1.0` branch and we would try to replace `.iui-button .iui-borderless` with something like `.iui-tabs-button`, but now that I know this is going into `main` maybe it makes sense to target `.iui-borderless` instead.

adamhock (commented):
To @FlyersPh9 's point, we could also target `iui-button` as well. That's how it was targeted in tabs.html before this change

Other than that, I see no issues

bentleyvk:
Are you planning to do changes from comments?

FlyersPh9:
I've updated this PR to target `.iui-button` instead of `.iui-tabs-button`.  This should work in both `main` and `v1.0` as it doesn't add any modifier classes and the class name is unchanged between the 2 branches.

adamhock (approved):

bentleyvk (approved):

packages/itwinui-css/src/tabs/default.scss
```diff
@@ -43,35 +43,39 @@
...
 
-      ~ .iui-tabs-content {
-        margin-left: -1px;
-      }
+    ~ .iui-tabs-content {
+      margin-left: -1px;
+    }
+
+    .iui-button {
+      width: 100%;
```
mayank99, line 78: Can we undo all the extra changes except for this part? Will make it easier to avoid merge conflicts.

Also maybe we should use `:where` selector to reduce specificity of this override in case the user wants a different width/height on the button.

FlyersPh9: Ok, change made.

mayank99 (approved):

PR #705, opened 2022-07-14
WIP: Rohan/vertical slider

rohan-kadkol:
Goals:

* Have a vertical slider as mentioned in #699 
  * Extend the current horizontal slider functionality to also support having vertical sliders

Tasks:

* [ ] Default sliders
  * [x] Without ticks
  * [x] With ticks
  * [ ] Disabled with ticks
* [ ] Range sliders
  * [x] Without ticks
  * [x] With ticks
  * [ ] Disabled with ticks

Screenshots (click to enlarge):

vertical_slider_with_ticks

vertical_range_slider_with_ticks

Bugs to solve:

* Unexpected thumb offset when the page is scrolled: I think has something to do with `event.pageY` inside the `mousemove` event callback. _I need to investigate it more but would appreciate your input if you know why this is happening._
  * 
* The below thumb of the range slider has an unexpected vertical offset (shown in the second screenshot above)

Issue #699, opened 2022-07-06
Vertical Sliders


Josh-Schifter:
Some examples from the web:

![image](https://user-images.githubusercontent.com/38294200/177568358-40ee12ab-8faf-433c-af6b-6118e38e5f6b.png)
![image](https://user-images.githubusercontent.com/38294200/177568464-7c6b5925-5214-4569-a263-d2437acdd3c4.png)
![image](https://user-images.githubusercontent.com/38294200/177568693-e3ae6760-e6ba-45f2-b6ab-0e6e6bc1bb61.png)

PR #707, opened 2022-07-15
feat(Slider): Add vertical slider

rohan-kadkol:
**NOTE**: This is the newer version of PR #705 

---

### Goals:

* Have a vertical slider as mentioned in #699 
  * Extend the current horizontal slider functionality to also support having vertical sliders

### Tasks:

* [x] Default vertical sliders
  * [x] Without ticks
  * [x] With ticks
  * [x] Disabled with ticks
* [x] Range vertical sliders
  * [x] Without ticks
  * [x] With ticks
  * [x] Disabled with ticks
* [x] Clean and optimize code
* [x] Create and run JS tests
* [x] Remove all TODOs from code (The last TODO is in the code of one of the [conversations](https://github.com/iTwin/iTwinUI/pull/707#discussion_r923511547). Once it is resolved, then I will remove the TODO)
* [x] Rename the scss variable of `track-height` to `track-thickness`, and `tick-height` to `tick-length` to be meaningful for horizontal and vertical sliders

### Screenshots:

https://user-images.githubusercontent.com/45748283/179300419-ec53e433-d127-44c4-9a84-32c4ee23441d.mp4

<details>
<summary>Expand to see old screenshots</summary>

#### 07/14/22 
https://user-images.githubusercontent.com/45748283/179027102-23f2ab0a-89d2-44e0-9207-16c44cd0bdca.mp4

</details>

### Bugs to solve:

* [x] Unexpected thumb offset when the page is scrolled: I think has something to do with `event.pageY` inside the `mousemove` event callback. _I need to investigate it more but would appreciate your input if you know why this is happening._
<details>
  <summary>Expand to see screenshot</summary>
  
</details>

* [x] The below thumb of the range slider has an unexpected vertical offset (shown in the second screenshot above)

Testing issues:

* [x] The test is not passing. Need to investigate it further. But as of now, I'm confused on why the test image's disabled slider is not showing ticks even when the real HTML file's disabled slider is showing ticks. One reason the test is failing is because the reference image's disabled slider has ticks whereas the test image's disabled slider does not have ticks.

<details>
<summary>Expand to see screenshot</summary>
https://user-images.githubusercontent.com/45748283/179313897-6daeb5a1-c4a4-45cb-95c6-c67464cb9d67.mp4
</details>


rohan-kadkol:
@veekeys, To solve the testing problem that I mentioned above, I tried to check if the ticks were outside the section with id = `​demo-default`​​. They seem to be inside the section.

I am currently comparing the old and the new code to see if I made any breaking changes. But do you have any other idea that comes to mind on why that is happening?

rohan-kadkol:
I was able to fix the above testing error for the horizontal sliders. For some reason, converting the tick numbers from being in a `<b>`​ to a `​<span>​` and adding a `​margin-bottom: $tick-height * 2;` to the tick​ worked. But I am not able to pass the tests for the vertical slider because it says "Reference file not found".

Should I remove the tests for the `<section>`s with the vertical sliders? What should I do to fix the failing backstop tests?

packages/itwinui-css/backstop/scenarios/slider.js
```diff
@@ -10,6 +10,14 @@ module.exports = [
...
     viewports: [{ width: 800, height: 600 }],
   }),
+  scenario('Type Default Vertical', {
+    selectors: ['#demo-default-vertical'],
+    viewports: [{ width: 800, height: 600 }],
+  }),
+  scenario('Type Range Vertical', {
+    selectors: ['#demo-range-vertical'],
+    viewports: [{ width: 800, height: 600 }],
+  }),
```
rohan-kadkol, line 20: I was able to fix the above testing error for the horizontal sliders. For some reason, converting the tick numbers from being in a `<b>`​ to a `​<span>​` and adding a `​margin-bottom: $tick-height * 2;` to the tick​ worked. But I am not able to pass the tests for the vertical slider because it says "Reference file not found".

Should I remove the tests for the `<section>`s with the vertical sliders? What should I do to fix the failing backstop tests? @FlyersPh9 @veekeys

veekeys: > I was able to fix the testing error for the horizontal sliders I mentioned in the above comment. For some reason, converting the tick numbers from being in a `<b>`​ to a `​<span>​` and adding a `​margin-bottom: $tick-height * 2;` to the tick​ div worked. But I am not able to pass the tests for the vertical slider because it says "Reference file not found".
> 
> Should I remove the tests for the `<section>`s with the vertical sliders? What should I do to fix the failing backstop tests? @FlyersPh9 @veekeys
> 
> 

Have you read the contributing guide?
The part about approving images especially.

https://github.com/iTwin/iTwinUI/blob/main/CONTRIBUTING.md#testing

rohan-kadkol: Oops, sorry, I read it a while ago. Forgot to read it again to see about testing. Checking it now

packages/itwinui-css/src/slider/classes.scss
```diff
@@ -6,6 +6,13 @@
   @include iui-slider-component-container;
 }
 
+// TODO: See if this is the correct place for non-iui related selectors
+#demo-default-vertical,
+#demo-range-vertical {
+  display: flex;
+  gap: 2rem;
+}
+
```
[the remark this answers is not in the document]
rohan-kadkol: After running `yarn test`, I realized that there is a `<section>` in tabs.html with the same id as the sliders. `#demo-default-vertical`. Hence, `tabs.html` is also affected by the selected lines above and so the backstop test for `tabs.html` also fails.

What do you recommend? Here are some solutions I can think of:
* Change the id of the `sliders.html`'s `<section>` to something else, like `#demo-default-vertical-slider`?
* Remove the selected lines of code above. Allow the vertical slides to be arranged vertically instead of horizontally.

rohan-kadkol:
After running `yarn test`, I noticed that one of the failing tests is for the `Type Split State hover & focus (split)` even though the reference and the test look nearly identical. The only diffs it found were the two images near Project Alpha (ie. the slim and default project alpha images). Should I run `yarn approve --filter=header`?


https://user-images.githubusercontent.com/45748283/179555965-75de2369-4a11-4ad4-a7b0-054894dd816d.mp4

image
image

packages/itwinui-css/src/slider/classes.scss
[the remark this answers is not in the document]
rohan-kadkol: Thanks, moved that code from `classes.scss` to `slider.html`. The failing backstop test now passes

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -88,6 +122,13 @@ $tick-height: $iui-baseline; // 11px
   @include themed {
     background-color: t(iui-color-foreground-primary);
   }
+
+  @at-root .iui-vertical & {
+    width: $track-height;
+    bottom: 0;
```
bentleyvk, line 128: Maybe in order to avoid constantly overriding horizontal rule we could have `iui-horizontal` and `iui-vertical` like we have in Tabs.

bentleyvk:
Your tests are failing because you forgot to approve image from new slider tests.

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: Thanks, I went through `tabs.scss`, trying to do something similar in `sliders.scss` now

bentleyvk: Hm, I was thinking to kind of keeping the same as it was where common styling is left but there would be `iui-horizontal` and `iui-vertical` modifiers that have orientation specific rules.

rohan-kadkol: @bentleyvk Ohh, yes, maybe I misunderstood. Could you please give a small code sample to help me understand what you meant? Also, just curious, do you feel the newer version is more cleaner and better? Or do you prefer the old version?

rohan-kadkol: I tried implementing the changes that you suggested. Kindly take a look and let me know if you suggest any other changes.

packages/itwinui-css/backstop/scenarios/slider.js
rohan-kadkol: **[DONE]** I ran `yarn approve` to approve the new vertical slider images. The backstop tests now pass

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -144,3 +131,118 @@ $tick-height: $iui-baseline; // 11px
     }
   }
 }
+
+@mixin iui-slider-horizontal {
+  // TODO: Vilius said this selector might not be needed and we can directly write the attributes. It works, but confirm if that is okay
```
bentleyvk, line 136: It is ok 😀

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -144,3 +131,118 @@ $tick-height: $iui-baseline; // 11px
...
+  .iui-slider-max {
+    margin-left: $iui-sm;
+  }
+}
+
+@mixin iui-slider-vertical {
+  // TODO: Vilius said this selector might not be needed and we can directly write the attributes. It works, but confirm if that is okay
+  // &.iui-slider-component-container {
+  flex-direction: column-reverse;
+  min-height: auto;
```
bentleyvk, line 193: This one shouldn't be needed anymore.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -382,17 +463,444 @@ <h2>Ranges</h2>
...
-              <span class="iui-slider-tick">20</span>
-              <span class="iui-slider-tick">30</span>
-              <span class="iui-slider-tick">40</span>
-              <span class="iui-slider-tick">50</span>
-              <span class="iui-slider-tick">60</span>
-              <span class="iui-slider-tick">70</span>
-              <span class="iui-slider-tick">80</span>
-              <span class="iui-slider-tick">90</span>
-              <span class="iui-slider-tick">100</span>
+              <div class="iui-slider-tick">
```
bentleyvk, line 466: Why tick bacame wrapper and now has two elements inside?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -144,3 +131,118 @@ $tick-height: $iui-baseline; // 11px
...
+  min-width: $iui-baseline * 2;
+  height: 100%;
```
bentleyvk, line 195: This doesn't seem to work. Try removing `height: 500px` from the demo and slider won't take full height. It's either problem with input container that it doesn't work properly with vertical slider or need to check other fixes. Because I removed input container wrapper from one of the vertical sliders and removed `height: 500px` then slider shrinked, then removed `height: 100%` and it expanded trough all height.

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -183,45 +183,49 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
...
-        to right,
-        $iui-color-system-red 0%,
-        $iui-color-system-yellow 17%,
-        rgb(0, 255, 0) 33%,
-        rgb(0, 255, 255) 50%,
-        $iui-color-system-blue 67%,
-        $iui-color-system-magenta 83%,
-        $iui-color-system-red 100%
-      );
+  .iui-slider-container {
```
bentleyvk, line 186: Can you check whether these changes are still needed?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -82,8 +64,6 @@ $tick-height: $iui-baseline; // 11px
 @mixin iui-slider-track {
   pointer-events: none;
   position: absolute;
-  height: $track-height;
-  left: 0;
   top: $iui-sm;
```
bentleyvk, line 67: If think this one is also horizontal specific.

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -144,3 +131,118 @@ $tick-height: $iui-baseline; // 11px
...
+
+  .iui-slider-max {
+    margin-left: $iui-sm;
+  }
+}
+
+@mixin iui-slider-vertical {
+  // TODO: Vilius said this selector might not be needed and we can directly write the attributes. It works, but confirm if that is okay
+  // &.iui-slider-component-container {
+  flex-direction: column-reverse;
```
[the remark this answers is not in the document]
rohan-kadkol: To pass the `minLabel` and the `maxLabel` in the same order as the horizontal but display them opposite to the horizontal. For e.g. the horizontal slider's `row` flex has the `minLabel`, the `slider-container`, and the `maxLabel`. The vertical slider's `column-reverse` flex also has the same order of tags but displays the `maxLabel` first, then the `slider-container`, and then the `minLabel`.

An alternative could be to change it to `flex-direction: column` and reverse the order of the flex tags too. So `maxLabel` - `slider-container` - `minLabel` for the vertical slider

packages/itwinui-css/backstop/tests/slider.html
rohan-kadkol: This was done since in the old code for the horizontal slider, the number would go below the tick because the number could not fit in the iui-slider-tick. However, in the vertical slider, the ticks needed to be horizontal. So when the number would overflow out of the iui-slider-tick, the number would still go below the tick (as below).

Actual


Required


I was not sure how to use the same HTML code to achieve it. So I thought to use flex on the iui-slider-tick and separate it into tick-mark and tick-number that are arranged in a row flex

Old code for a tick

```html
<div class="iui-slider-ticks">
  <span class="iui-slider-tick">0</span>
</div>
```

New code for a tick

```html
<div class="iui-slider-ticks">
  <div class="iui-slider-tick">
    <span class="iui-slider-tick-mark"></span><span class="iui-slider-tick-number">10</span>
  </div>
</div>
```

packages/itwinui-css/src/color-picker/color-picker.scss
rohan-kadkol: I just rechecked that after removing this change, the backstop tests failed. The reason I had to do this change was because the color-picker's css selector for the rail gets lesser priority than the slider's css rail selector. As a result, the color picker can't use the override height of 8px. Instead, it falls back to the slider's specified height of 4px.

https://user-images.githubusercontent.com/45748283/180248425-a18277a9-8fe3-4277-8924-cd7125d61168.mp4

I solved this by making the color-picker's css selector more specific.

Old

image


New

image

Can you recommend any better way to solve this problem?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -62,10 +47,7 @@ $tick-height: $iui-baseline; // 11px
   height: $thumb-height;
   width: $thumb-height;
   border-radius: 100%;
-  top: $track-height + 1; // position center of thumb in center of track
-  transform: translateX(
-    (-$thumb-height - 2) * 0.5
-  ); // Adjust the left (set in code) by radius of thumb so thumb center is at value position
+  transform: translateX(-50%); // Adjust the left (set in code) by radius of thumb so thumb center is at value position
```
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Oops, sorry I missed that. Moved line 50 under the horizontal mixin

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: **[DONE]** I think this code is from an old commit. But thanks, yes, I reconfirmed that it is now under `iui-slider-horizontal`

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -9,30 +9,18 @@ $tick-height: $iui-baseline; // 11px
...
-    margin-left: $iui-sm;
-  }
+@mixin iui-slider-endpoints {
+  user-select: all;
 
-  &.iui-disabled {
-    @include iui-slider-disabled;
+  svg {
+    @include iui-icons-default;
+    margin-top: $iui-xs;
```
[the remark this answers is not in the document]
rohan-kadkol: Yes, it is needed for both. The svg icons need to look the same for the vertical and the horizontal sliders. Putting lines 21-23 under just `iui-slider-horizontal` makes the vertical's and horizontal's min/max icons look different, which fails the tests.


https://user-images.githubusercontent.com/45748283/180340037-d4bc41d5-bfdb-4647-abe5-7a70c8f04035.mp4

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: I tried replicating what you said. After removing `height: 500px` from the `iui-slider-container`, removing the slider from the input container, and removing `height: 100%`, the vertical slider does expand completely. But that was because it was in the row flex (the div that arranges the vertical sliders horizontally).

I'm not entirely sure I understood the problem you are highlighting. But if I understood correctly, then if the slider is not in a flex (like a row flex) and not in a `display: grid` (input container) but in a container with a fixed height, then the `height: 100%` makes the vertical slider extend the entire height of the container. That way users can size the wrapping container to their requirements and the vertical/horizontal slider fits the container.

Please let me know if I misunderstood you.

packages/itwinui-css/src/color-picker/color-picker.scss
bentleyvk: I think it is fine. Just wanted to double check.

packages/itwinui-css/src/slider/slider.scss
bentleyvk: Yup

packages/itwinui-css/backstop/tests/slider.html
bentleyvk: It seems previously it was done in a hacky way, so I guess it is ok.

bentleyvk (approved):

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -505,39 +1041,29 @@ <h2>Ranges</h2>
         return tooltips[0];
       }
 
+      // TODO: Confirm if calculating this every mousemove has a negligible cost. Or do it just once and save it as a data property in the slider container (instead of its parent)
+      function getIsVerticalSlider(sliderContainer) {
+        return sliderContainer.parentNode.classList.contains("iui-vertical");
+      }
+
```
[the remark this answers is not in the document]
rohan-kadkol: **[OUTDATED]** Resolving as outdated. Will re-add this comment to the new code.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -505,39 +1056,29 @@ <h2>Ranges</h2>
         return tooltips[0];
       }
 
+      // TODO: Confirm if calculating this every mousemove has a negligible cost. Or do it just once and save it as a data property in the slider container (instead of its parent)
+      function getIsVerticalSlider(sliderContainer) {
+        return sliderContainer.parentNode.classList.contains("iui-slider-vertical");
+      }
```
rohan-kadkol, line 1062: @mayank99 Can you please let me know if calling this method on every thumb movement is okay? Or should I save a data attribute of the orientation in the thumb also? (Currently, the orientation class of `iui-slider-vertical` is only in `iui-slider-component-container`) I am leaning towards the first option because with the second, I feel there is a source of inconsistency between the orientation data attribute in the `iui-slider-component-container` and the `iui-slider-thumb`.

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: **[DONE]** Marking it as done as Vilius approved this PR. @bentleyvk, can you please mark this as resolved? I don't want to do it myself in case you wanted to recommend some more changes. Thanks.

packages/itwinui-css/backstop/tests/slider.html
bentleyvk: It is demo/test html so don't worry, also it shouldn't be that costly to call it constantly.

rohan-kadkol: **[DONE]** Thanks, then will leave it as it is

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -142,17 +157,39 @@ <h2>Default</h2>
...
-              <span class="iui-slider-tick">30</span>
-              <span class="iui-slider-tick">40</span>
-              <span class="iui-slider-tick">50</span>
-              <span class="iui-slider-tick">60</span>
-              <span class="iui-slider-tick">70</span>
-              <span class="iui-slider-tick">80</span>
-              <span class="iui-slider-tick">90</span>
-              <span class="iui-slider-tick">100</span>
+              <div class="iui-slider-tick">
+                <span class="iui-slider-tick-mark"></span><span class="iui-slider-tick-number">0</span>
```
FlyersPh9, line 161: You might be able to remove `<span class="iui-slider-tick-mark"></span>` & use a pseudo element (`::before`) instead.

rohan-kadkol: Thanks for the idea, I removed the `iui-slider-tick-mark` span and reverted back to the old tick mark HTML structure. I replaced the `iui-slider-tick-mark` span with a `::before` on the `iui-slider-tick`.

#### Old (reverted back to this) `iui-slider-tick` structure
```html
<span class="iui-slider-tick">0</span>
```

#### New (deleted this and reverted back to the above version) `iui-slider-tick` structure
```html
<div class="iui-slider-tick">
  <span class="iui-slider-tick-mark"></span><span class="iui-slider-tick-number">0</span>
</div>
```

FlyersPh9: Awesome, looks great & good use of a pseudo element!

rohan-kadkol: @FlyersPh9 Great, thanks! If there aren't any other changes that you recommend, can you please give your approval so that I can merge the branch? Thanks

FlyersPh9 (approved):

Issue #645, opened 2022-05-20
CONTRIBUTING.md needs to be updated


mayank99:
The `build-docker` command needs to be mentioned.

PR #708, opened 2022-07-15
chore: Use vite for build and dev steps

mayank99:
- [x] Dev server: `yarn dev`
- [x] Build step: `yarn build`
- [x] Index file 
- [x] Backstop tests
- [x] Github pages

Notable changes:
- moved `assets/` folder under `tests/` so that it is scoped
- replaced `generateHtml.js` with vite plugins
- added deploy preview for every branch!
- replaced skypack version of icons with locally installed one, currently using `0.2.0` to prevent image changes


packages/itwinui-css/backstop/tests/header.html
```diff
@@ -14,17 +14,18 @@
...
-      href="../../css/all.css"
-    />
-    <link
-      rel="stylesheet"
-      href="../assets/demo.css"
-    />
+
+    <script type="module">
+      import "./assets/theme.js";
+      import "./assets/userIcon.js";
```
mayank99, line 20: I am running into an issue with backstop tests. Because this is an ESM import, it requires a dev server to be able to resolve. Simply opening the file (e.g. visiting the `file:///` path in browser) does not work.

@bentleyvk Do you know if there is an easy way to make backstop tests run on a server rather than on hardcoded html paths?

packages/itwinui-css/backstop/tests/alert.html
```diff
@@ -14,15 +14,12 @@
...
-    <link
-      rel="stylesheet"
-      href="../assets/demo.css"
-    />
+    <script type="module">
+      import "./assets/theme.js";
+    </script>
     <style>
+      @import url("./assets/demo.css") layer(demo);
+      @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
```
mayank99, line 22: I'm using CSS layers here to work around this Vite bug which messes up order of CSS (thus causing visuals to change): https://github.com/vitejs/vite/issues/3924

VSCode will currently show red squigglies even though this is valid syntax. Should be fixed soon: https://github.com/microsoft/vscode-css-languageservice/pull/276

packages/itwinui-css/backstop/tests/header.html
bentleyvk: Try chaning `scenario.url` in `packages\itwinui-css\backstop\scenarios\index.js`
image

mayank99: So I would need to start a server inside the docker container? Could use a custom entrypoint to do that, but then I won't have access to this `{backstopCommand}`

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -15,14 +15,27 @@
...
-      href="../../css/all.css"
-    />
-    <link
-      rel="stylesheet"
-      href="../assets/demo.css"
-    />
+
+    <script type="module">
+      import "./assets/theme.js";
+      import "./assets/userIcon.js";
```
bentleyvk, line 21: Ran `yarn dev` and opened Table html. I see that user icon is not shown properly as it doesn't have its css.
image

packages/itwinui-css/package.json
```diff
@@ -29,30 +29,34 @@
...
+    "@itwin/itwinui-icons-elements": "0.2.0",
     "autoprefixer": "^10.4.1",
     "backstopjs": "~6.0.4",
     "html-minifier": "^4.0.0",
+    "npm-run-all": "^4.1.5",
     "postcss": "^8.4.5",
     "postcss-discard-comments": "^5.0.1",
     "postcss-scss": "^4.0.3",
     "sass-embedded": "^1.49.9",
+    "serve": "^14.0.1",
```
bentleyvk, line 41: Why it is needed? Can't you use preview from vite?

packages/itwinui-css/backstop/scenarios/header.js
```diff
@@ -4,44 +4,45 @@ module.exports = [
   // Header types
   scenario('Type Default', {
     selectors: ['#demo-default'],
+    misMatchThreshold: 0.1,
```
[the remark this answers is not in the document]
mayank99: was getting random failures repeatedly

packages/itwinui-css/backstop/tests/alert.html
bentleyvk: Don't you need to set layer order?

packages/itwinui-css/package.json
mayank99: preview wasn't showing anything for me

packages/itwinui-css/backstop/tests/table.html
mayank99: i think it might need build first, can you try changing turbo.json like this and just run `yarn dev` after that?

dependsOn build

bentleyvk: I ran `yarn build` and it helped. So this might help.

packages/itwinui-css/package.json
bentleyvk: Have you tried looking into that?

.github/workflows/build.yml
```diff
@@ -49,28 +49,19 @@ jobs:
...
       - run: yarn lint:copyright ${{ github.workspace }}/*/**.{scss,html,js,ts}
       - run: yarn build
-      - run: yarn test
-
-      - name: Publish test results artifact
-        if: failure()
-        uses: actions/upload-artifact@v2.2.3
-        with:
-          name: 'backstop_test_results'
-          path: '${{ github.workspace }}/packages/itwinui-css/backstop/results/'
```
bentleyvk, line 52: Have you tried to fail tests on purpose?

packages/itwinui-css/backstop/tests/table.html
mayank99: Changed

packages/itwinui-css/package.json
mayank99: I don't see any logs or anywhere so I was very confused and decided to just go with `serve`.

packages/itwinui-css/package.json
```diff
@@ -29,30 +29,34 @@
...
     "build:css": "node ./scripts/generateCss.js src css",
-    "build:html": "node ./scripts/generateHtml.js backstop/tests backstop/minified",
-    "build:watch": "yarn build && concurrently -n scss,html -k \"yarn build:watch:scss\" \"yarn build:watch:html\"",
+    "build:html": "vite build backstop/tests --config vite.config.ts",
     "build:watch:scss": "chokidar \"src/**/*\" -c \"yarn build:scss && yarn build:css\"",
-    "build:watch:html": "chokidar \"backstop/tests/**/*\" -c \"yarn build:html\"",
     "createComponent": "node ./scripts/createComponent.js",
     "clean": "rimraf ./css ./scss ./backstop/results/bitmaps_test",
-    "dev": "yarn build:watch",
+    "dev": "concurrently -n scss,html -k \"yarn build:watch:scss\" \"yarn vite backstop/tests --config vite.config.ts\"",
```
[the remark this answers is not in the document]
mayank99: What documentation?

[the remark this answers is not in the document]
mayank99: Oh you mean contributing guide? There is already an item for that. #645

packages/itwinui-css/package.json
bentleyvk: Maybe ok, just not to waste time for this minor stuff.

packages/itwinui-css/backstop/tests/table.html
bentleyvk: Would be great if others could check whether it is working now for them @veekeys @FlyersPh9 @gretanausedaite

packages/itwinui-css/package.json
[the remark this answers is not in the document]
bentleyvk: But it is different because now need to document how to run demo html locally. Just one or two sentences so guess you can do with this PR.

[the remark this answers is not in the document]
mayank99: Ok, good point. Updated contributing.md

packages/itwinui-css/backstop/tests/table.html
veekeys: I ran `yarn` and `yarn dev`. The html files seem to work fine

packages/itwinui-css/backstop/scenarios/header.js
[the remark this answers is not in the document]
veekeys: Doesnt sound positive

packages/itwinui-css/backstop/tests/alert.html
veekeys: You say if we swapped demo and all css files order, it got messed up??? Should have fixed demo.css instead.

package.json
```diff
@@ -24,7 +24,7 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev",
-    "test": "turbo run test",
+    "test": "turbo run test --",
```
bentleyvk, line 27: What `--` does?

packages/itwinui-css/backstop/scenarios/header.js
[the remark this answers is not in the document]
bentleyvk: In other PRs it is also failing pretty commonly. Maybe at least we should try with a bit lower value like `0.05`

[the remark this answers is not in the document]
veekeys: There was a reason with put these specific ones before, wasnt it?

packages/itwinui-css/backstop/tests/alert.html
veekeys: Also, it shows this as error
image

packages/itwinui-css/backstop/tests/alert.html
```diff
@@ -14,41 +14,22 @@
...
-    <link
-      rel="stylesheet"
-      href="../../css/all.css"
-    />
-    <link
-      rel="stylesheet"
-      href="../assets/demo.css"
-    />
+    <script type="module">
+      import "./assets/theme.js";
```
veekeys, line 18: Why need it this way now? Something because of Vite?

packages/itwinui-css/package.json
veekeys, line 54: I couldnt figure from the scripts and configs...
If we run `yarn dev`, do we use `serve` anywhere? if yes, where this is defined? Cause i see only `preview` using it, but cant see we run `preview` while runnig `yarn dev`.

mayank99: `serve` is used in `yarn preview` which is used in `yarn test`

veekeys: amm... and it is used to only show up test results?
Cant we use `vite` same as with backstop html folder?

veekeys: aaa, ok.. now I see.. it is hosting backstop minified files, so tests could access those...?

mayank99: yeah, we need to run a server now because module imports are not allowed by simply visiting an html file.

packages/itwinui-css/backstop/tests/alert.html
mayank99: i did it this way to make sure vite knows the correct path

package.json
mayank99: it allows passing args to the subpackage commands.

in this case it will allow running `yarn test --filter=alert` from the monorepo root

packages/itwinui-css/backstop/scenarios/header.js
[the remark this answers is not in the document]
mayank99: previously, those specific ones were just the lowest values i could get the tests to pass with

`0.1` is still pretty low, it's `0.1%` so more like `0.001`

packages/itwinui-css/backstop/tests/alert.html
mayank99: > You say if we swapped demo and all css files order, it got messed up??? Should have fixed demo.css instead.

No it's about the bespoke styles present in the `.html` pages (inside `<head>`). Vite places imported styles at the end of `<head>`, which overrides those page styles. The only other way to get around it would be to increase specificity in every .html page.

> Also, it shows this as error

see above: https://github.com/iTwin/iTwinUI/pull/708#discussion_r922521350

bentleyvk (approved):

mayank99:
> Will this change allows us using fontsource instead of font files?

Yes it should. But we already have font files in source so I just kept it unchanged.

veekeys (approved):

PR #711, opened 2022-07-19
docs: Added `yarn build-docker` to CONTRIBUTING.md

rohan-kadkol:
Added `yarn build-docker` to `CONTRIBUTING.md` as mentioned in issue #645

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,14 +79,6 @@ We provide a script that can automatically create all the necessary files for yo
 
 For running tests you will need [Docker](https://www.docker.com/products/docker-desktop). It helps to avoid cross-platform rendering differences.
 
-#### How to set up Docker:
-
-- Make sure Docker is running.
-
-- Create the docker image needed for testing by running `yarn build-docker` from the root folder.
-
-#### How to run tests:
-
 - To run tests for a specific component, use this command:
 
   `yarn test --filter=[component_name]` (e.g. `yarn test --filter=side-navigation`)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,8 @@ For running tests you will need [Docker](https://www.docker.com/products/docker-
 
 #### How to set up Docker:
 
-- Download and install [Docker](https://www.docker.com/products/docker-desktop) and make sure Docker is running
+- Make sure Docker is running
+
 - Create the docker image needed for testing by running `yarn build-docker` from the `/packages/itwinui-css/` folder
 
 #### How to run tests:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,14 @@ We provide a script that can automatically create all the necessary files for yo
 
 ### Testing
 
+#### How to set up Docker:
+
+- Make sure Docker is running.
+
+- Create the docker image needed for testing by running `yarn build-docker` from the root folder.
+
+#### How to run tests:
+
 For running tests you will need [Docker](https://www.docker.com/products/docker-desktop). It helps to avoid cross-platform rendering differences.
 
 - To run tests for a specific component, use this command:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,9 +81,9 @@ For running tests you will need [Docker](https://www.docker.com/products/docker-
 
 #### How to set up Docker:
 
-- Make sure Docker is running
+- Make sure Docker is running.
 
-- Create the docker image needed for testing by running `yarn build-docker` from the `/packages/itwinui-css/` folder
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -100,23 +107,17 @@ For running tests you will need [Docker](https://www.docker.com/products/docker-
   - For actions like click, hover use according functions from `scenarioHelper.js` and pass them as scenario options `actions` property.
     ```js
     const { scenario, hover } = require('../scenarioHelper');
-    module.exports = [
-      scenario('hover', { actions: [hover('.element-selector')] }),
-    ];
+    module.exports = [scenario('hover', { actions: [hover('.element-selector')] })];
```
bentleyvk, line 110: Why changed here?

CONTRIBUTING.md
```diff
@@ -69,16 +69,23 @@ We provide a script that can automatically create all the necessary files for yo
   - Use existing Sass variables for spacing (margin, padding, sizing, etc).
   - Use [`themed`](https://github.com/iTwin/iTwinUI/blob/main/src/style/theme.scss#L430) mixin wherever color is involved.
 - Define classes for your mixin in `src/[component-name]/classes.scss` file.
-  - *Running the `createComponent` command will do this for you.*
+  - _Running the `createComponent` command will do this for you._
```
bentleyvk, line 72: What it does? Cause I can't see the difference 😀

CONTRIBUTING.md
rohan-kadkol: Ohh! I didn't make that change. Looks like it was the linter/formatter that runs after saving/committing the file.

### Before saving

image

### After saving

https://user-images.githubusercontent.com/45748283/180015109-28b8ed5f-b99c-4b8a-a8d2-c5011e71b10c.mp4

### After committing

Same as after saving

CONTRIBUTING.md
rohan-kadkol: Right, there is no visual difference. This change is also not by me! It was by the linter/formatter that runs after saving/committing the file.

CONTRIBUTING.md
```diff
@@ -69,16 +69,23 @@ We provide a script that can automatically create all the necessary files for yo
...
 - Write tests for your new component in `backstop/tests/[component-name].html` and `backstop/scenarios/[component-name].js`. See [Testing](#Testing) section below.
 - After running `yarn build` you can open minified html in browser to check up, how your component looks like from `backstop/minified`.
 
 ### Testing
 
 For running tests you will need [Docker](https://www.docker.com/products/docker-desktop). It helps to avoid cross-platform rendering differences.
 
+#### How to set up Docker:
+
+- Download and install [Docker](https://www.docker.com/products/docker-desktop) and make sure Docker is running
```
mayank99, line 84: Do we need to specify this again? The line right above says you will need docker, so maybe we can just say "Make sure Docker is running".

CONTRIBUTING.md
```diff
@@ -69,16 +69,23 @@ We provide a script that can automatically create all the necessary files for yo
...
+- Create the docker image needed for testing by running `yarn build-docker` from the `/packages/itwinui-css/` folder
```
mayank99, line 85: Maybe we should add an alias for `build-docker` at the root package.json so that it doesn't need to be run from `/packages/itwinui-css/`.

CONTRIBUTING.md
rohan-kadkol: Ohh, right, that makes sense. Edited it to just `Make sure Docker is running`

CONTRIBUTING.md
mayank99: you can use the [yarn workspaces](https://classic.yarnpkg.com/en/docs/cli/workspace) syntax, no need to `cd`

```json
"build-docker" "yarn workspace @itwin/itwinui-css build-docker"
```

rohan-kadkol: Thanks, replaced the cd command with the above `yarn workspace` command. I checked that the command works. But would be great if you can reconfirm it just to be sure. Thanks

veekeys (approved):

bentleyvk (approved):

mayank99 (approved):
Looks good but i would say lets undo the unformatting changes (and "Save without formatting") before merging.

CONTRIBUTING.md
rohan-kadkol: **[DONE]** I saved the file without formatting as mentioned [below](https://github.com/iTwin/iTwinUI/pull/711#pullrequestreview-1046485696). Thus, we can ignore the changes due to formatting for now.

rohan-kadkol:
> Looks good but i would say lets undo the unformatting changes (and "Save without formatting") before merging.

I saved `CONTRIBUTING.md` with just the `yarn build-docker` and without formatting it. After doing that, I then merged this PR

PR #712, opened 2022-07-20
feat(Button): Borderless button with only icon is 1:1

### LINKED ISSUE #605: [V1] Button refactor
- [ ] split-menu rename to split-button

- [ ] specificity max 0.3.0

- [ ] have single class versus modifiers (i.e. iui-button-high-visibility))

- [ ] Icon button component. Should be square

- [ ] Button group sub components 

- [ ] Borderless active button needs active stripe

FlyersPh9:
- Borderless button with only an icon present is 1:1 width & height.

Closes #605.


packages/itwinui-css/src/header/header.scss
```diff
@@ -325,6 +325,12 @@
     }
   }
 
+  > *:last-child {
```
FlyersPh9, line 328: This can be removed after #640.

bentleyvk (approved):

FlyersPh9:
Handled in #718.

PR #716, opened 2022-07-22
chore: Use published docker image

veekeys:
- Update `buildDocker` script to build and push the docker image to docker hub. 
- Now it builds multi-arch image to support all architectures.
- It publishes 2 versions: `latest` and `{VERSION}` specified by `VERSION` variable in script.
- `VERSION` needs to be updated before new build, otherwise it will override the existing versions.
- Renamed `buildDocker` -> `pushDocker`.
- Added `VERSION` variable, which needs to be updated.
- For now, I have decided to remove build action from workflow. If we want to automate some of this process, we can do that later, but now it seems we are updating dockerfile pretty rarely. So when that happens, we can manually publish it again by running `yarn pushDocker` command.


packages/itwinui-css/backstop/backstop.js
```diff
@@ -31,7 +31,7 @@ const config = {
   debug: false,
   debugWindow: false,
   dockerCommandTemplate:
-    'docker run --rm -i --mount type=bind,source="{cwd}",target=/src itwinui/backstopjs:latest {backstopCommand} {args}',
+    'docker run --rm -i --mount type=bind,source="{cwd}",target=/src bentleysystemsinc/itwinui:backstop {backstopCommand} {args}',
```
bentleyvk, line 34: Is it correct? Because previously it was `:latest` instead of some version and that makes sense but now it has `:backstop`. What it means?

bentleyvk:
You should have just ignored your local header test failures.

gretanausedaite (commented):
I think we need two versions of this image pushed to docker. It fails on mac
image

mayank99:
> I think we need two versions of this image pushed to docker. It fails on mac

We can publish a multi-arch image! In [`bcb2e5a` (#571)](https://github.com/iTwin/iTwinUI/pull/571/commits/bcb2e5a18a5156248326a03c4c479cdf9ffe09d8), I changed the approach to detect the platform in a script. This was done to reduce the time of build. But if we are publishing the image, we should use `--platform`.

packages/itwinui-css/backstop/backstop.js
veekeys: I have changed things a bit and updated whole description accordingly. @mayank99 @bentleyvk 

@gretanausedaite I am building now multiarch image, check up again, if it passes on MAC please.

CONTRIBUTING.md
```diff
@@ -79,8 +79,6 @@ We provide a script that can automatically create all the necessary files for yo
 
 - Make sure Docker is running.
```
mayank99, line 80: I think we can just remove this section and mention it under "How to run tests".

packages/itwinui-css/backstop/backstop.js
mayank99: Working for me on M1 mac 👍

CONTRIBUTING.md
veekeys: done

mayank99 (approved):

bentleyvk (approved):

gretanausedaite (approved):

PR #718, opened 2022-07-25
feat: Use itwinui-variables

mayank99:
Added `@itwin/itwinui-variables` import in html pages. Changed theming logic to use `data-iui-theme` attribute, both in `<theme-button>` component as well as backstop tests.

Major changes made:
- [x] **Spacing**
- [x] **Text and icon colors**
- [x] **Baseline**: `$iui-baseline` replaced with `--iui-size-s`, `$iui-line-height` with `--iui-size-l`.
- [x] **Typography**: _Had to fix `--iui-font-weight-semibold` to be 600 instead of 500._ 
- [x] **Shadows**
- [x] **Border-radii**
- [x] Removed files with old sass variables wherever applicable
- [x] **Speed variables**: Added new `duration.scss` in variables pkg.
- [x] **Component height/offset**: _Replaced with existing variables and also added new component-height vars._
- [x] Transparency (existing usage of `-rgb` + opacity): _This required some non 1:1 changes/rewrites in a few places._
- [x] Icon vars removed
- [x] **Badge**: Previously was using specific colors from other themes. The values are now hardcoded.

---

_See deploy preview at https://itwin.github.io/iTwinUI/718/_


mayank99:
I think this is now ready for review. The only thing pending is to potentially add a few more CSS variables for the colors that we are not happy with.

I have not updated the test images yet because I would like to get the color changes in first.

packages/itwinui-css/src/button/split-button.scss
```diff
@@ -31,7 +30,7 @@
   }
 
   :where(.iui-button:last-child) {
-    --_iui-button-padding-inline: #{$iui-xs};
+    --_iui-button-padding-inline: var(--iui-size-2xs);
```
FlyersPh9, line 33:
```suggestion
    --_iui-button-padding-inline: var(--iui-size-2xs);
    
    min-width: unset;
```
We need to remove the min-width on the 2nd button within the split button.

How it looks currently:
Screen Shot 2022-08-12 at 10 00 34 AM

How it should look:
Screen Shot 2022-08-12 at 10 01 36 AM

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -73,12 +72,12 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
   display: flex;
   flex-wrap: wrap;
   align-content: flex-start;
-  min-height: $iui-l;
-  min-width: $iui-l;
-  gap: $iui-s;
+  min-height: var(--iui-size-l);
+  min-width: var(--iui-size-l);
+  gap: var(--iui-size-xs);
```
FlyersPh9, line 77:
```suggestion
  gap: var(--iui-size-xs);
  align-items: center;
```
The borderless button & swatches are not aligned in this branch:
Screen Shot 2022-08-12 at 10 12 00 AM

FlyersPh9:
I'm seeing a visual bug with the dropdown icon in the small select within the information panel demo:

Screen Shot 2022-08-12 at 10 23 05 AM

FlyersPh9:
All instances of `@include iui-icon-style('3xl')` need to have the `'` removed.  Tagging @elephantcatdog just so she is aware because I think these changes came from #701. Example:

```diff
- @include iui-icon-style('3xl');
+ @include iui-icon-style(3xl);
```

FlyersPh9:
I noticed local & live both list **index** as a demo page & it just refreshes the homepage.  Not sure when this got introduced.

Screen Shot 2022-08-12 at 10 56 43 AM

packages/itwinui-css/src/table/table.scss
```diff
@@ -154,11 +154,7 @@
 
     &:hover:not([aria-disabled='true']) {
       .iui-table-cell {
-        background: linear-gradient(
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6)),
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6))
-        ),
-          linear-gradient(var(--iui-color-background-1), var(--iui-color-background-1));
+        background-color: var(--iui-color-background-subtle-primary);
```
FlyersPh9, line 157:
```suggestion
        background-color: var(--iui-color-background-transparent-hover);
```
To align with menu.

packages/itwinui-css/src/progress-indicator/overlay.scss
```diff
@@ -15,17 +15,10 @@
...
 
   // Blur fallback
-  background-color: rgba(var(--iui-color-background-1-rgb), var(--iui-opacity-2));
-
-  // With blur
-  @supports (backdrop-filter: $iui-blur-filter) {
-    background-color: rgba(var(--iui-color-background-1-rgb), var(--iui-opacity-3));
-    backdrop-filter: $iui-blur-filter;
-  }
+  @include iui-blur($opacity: 3);
```
FlyersPh9, line 18: This needs to be looked into:
Screen Shot 2022-08-12 at 11 55 45 AM

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -124,7 +128,7 @@
     }
 
     &.iui-expand {
-      height: $iui-line-height;
+      height: var(--iui-size-l);
```
FlyersPh9, line 131:
```suggestion
      height: var(--iui-size-l);
      min-height: unset;
```

We need to remove the min-height (or change it somehow else) because it looks much larger than it was previously.
Screen Shot 2022-08-12 at 11 59 42 AM

FlyersPh9:
The slider icons & drag handle are misaligned and the start / end icons are not receiving the correct color:
Screen Shot 2022-08-12 at 12 02 26 PM

@rohan-kadkol could probably offer a solution as he worked on sliders fairly recently.

FlyersPh9:
Might need to introduce some custom styling for the active state in paginator as we have some really bad contrasting colors.

Screen Shot 2022-08-12 at 12 06 12 PM

packages/itwinui-css/src/tag/tag.scss
```diff
@@ -15,29 +14,28 @@
...
 }
 
 @mixin iui-tag {
   @include iui-tag-base;
-  height: $iui-baseline * 3;
-  margin-top: $iui-component-offset;
-  margin-bottom: $iui-component-offset;
-  border-radius: $iui-border-radius-round;
-  padding: 0 $iui-xxs;
+  height: calc(var(--iui-size-s) * 3);
```
FlyersPh9, line 23:
```suggestion
  --_iui-tag-height: var(--iui-component-height);
  
  height: var(--_iui-tag-height);
```

Our tags should probably use the new component height variable.

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -1,11 +1,14 @@
...
 
 $iui-active-outline: thin solid var(--iui-color-foreground-primary);
-$iui-expander-inline-padding: $iui-component-offset + $iui-xxs + 1px; // margin + padding + border
-$iui-expander-button-width: ($iui-icon-default) + ($iui-expander-inline-padding * 2); // icon width + inline padding on both sides
+$iui-expander-inline-padding: calc(
+  var(--iui-size-2xs) + 1px
+); // padding of small borderless button + 1px transparent border
+$iui-expander-button-width: calc(
+  var(--iui-component-height-large) + 2px
+); // width of small borderless button (button is 1:1 ratio)
```
FlyersPh9, line 11: The formatting on this could be improved.  Also I think the math is wrong, our indentations are too large now:

**Before:**
Screen Shot 2022-08-12 at 12 16 03 PM

**After:**
Screen Shot 2022-08-12 at 12 16 38 PM

FlyersPh9:
I've looked through every component 1 by 1.  Only things remaining to investigate are `::selection` with status colors (I would _really_ like to keep this) and looking into opacity for scrollbars (I would prefer if we include opacity but I could live without it, just verify the end result looks good on the `color.html` demo page).

rohan-kadkol:
> The slider icons & drag handle are misaligned and the start / end icons are not receiving the correct color: Screen Shot 2022-08-12 at 12 02 26 PM
> 
> @rohan-kadkol could probably offer a solution as he worked on sliders fairly recently.

From a first look, I remember I also faced the problem of different min and max icon colors. I fixed it by checking that one of the old `svg` tag selectors was not accidentally deleted or placed in the wrong scss hierarchy/code block. For the misaligned thumb, I too faced similar problems. It just involved playing around with the values. I guess when including the variables, it might have caused some other problem.

Also, I noticed that lines 41 and 42 were removed from the below screenshot of `sliders.scss`. Not sure, but maybe re-adding that might fix the thumb alignment?

image

I will take a closer look, play with the values, and update you. Thanks.

mayank99:
@FlyersPh9 Thanks for the detailed review, will try to get these soon.

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -20,7 +19,7 @@ $tick-height: $iui-baseline; // 12px
 
   svg {
     @include iui-icon-style('m');
-    margin-top: $iui-xs;
+    margin-top: var(--iui-size-2xs);
   }
 }
```
rohan-kadkol, line 24: To solve the slider icons color problem as mentioned [here](https://github.com/iTwin/iTwinUI/pull/718#issuecomment-1213274552), I think you can include the `iui-slider-color` mixin in the `iui-icon-style` mixin or something similar.

image

<details>
<summary>Expand to see old code</summary>

#### slider.scss
image

#### mixins.scss
image

#### mixins.scss
image

</details>

mayank99: That is a specifically a change we don't want to make. The `iui-icon-style` mixin already has too many side effects (hence https://github.com/iTwin/iTwinUI-icons/pull/43) considering how widely it is used.

But I think we could `@include iui-icon-color` in addition to `iui-icon-style` right here in this rule (after line 21).

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -38,9 +37,10 @@ $tick-height: $iui-baseline; // 12px
...
   position: absolute;
-  width: 100%;
-  height: $track-height;
   background-color: var(--iui-color-background-border);
+  @media (forced-colors: active) {
+    background-color: CanvasText;
+  }
 }
 
 @mixin iui-slider-thumb {
```
rohan-kadkol, line 46: I don't know if this is important, but one small thing I noticed was that this property might be increasing the size of the thumb from `14x14` to `16x16`. I think this is part of [Leah's PR](https://github.com/iTwin/iTwinUI/pull/661) and might be a needed property, but just mentioning what I noticed. Because this property was not there when I working on the [vertical slider PR](https://github.com/iTwin/iTwinUI/pull/707)

rohan-kadkol: Oops, looks like it selected the wrong lines. I was referring to this property
image

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -102,30 +109,34 @@ $tick-height: $iui-baseline; // 12px
...
+  min-height: calc(var(--iui-size-s) * 2);
 
   .iui-slider-rail {
     width: 100%;
     height: $track-height;
-    top: $iui-sm;
+    top: var(--iui-size-s);
   }
 
   .iui-slider-thumb {
```
rohan-kadkol, line 142: For the thumb positioning problem mentioned [here](https://github.com/iTwin/iTwinUI/pull/718#issuecomment-1213274552), I think it is because `$track-height + 1` is producing an unexpected result in the actual webpage's CSS.

#### Selected lines (slider.scss):

image

#### Declaration of $track-height (slider.scss):

image

#### What appears in the browser:

image

Here there is an expected 1 that is causing a problem. It works fine when we replace it in the browser with the calculated value of 5px.

packages/itwinui-css/src/slider/slider.scss
mayank99: What property?

packages/itwinui-css/src/slider/slider.scss
mayank99: Good catch, should be fixed now

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: `box-sizing: content-box` on line 48. This was just a small thing I noticed, so I thought I'd just mention it.

mayank99: I was looking into this and didn't notice any difference between the size of the thumb in the [live version](https://itwin.github.io/iTwinUI/slider.html) and [current version](https://itwin.github.io/iTwinUI/718/slider.html). They are both 18x18 (16 + 1px border on both sides).

packages/itwinui-css/src/progress-indicator/overlay.scss
mayank99: Fixed by adding a `-background-1-hsl` variable.

packages/itwinui-css/src/color-picker/color-picker.scss
mayank99: Fixed

packages/itwinui-css/src/tag/tag.scss
mayank99: I don't think tags need to always have the same height as inputs/buttons (especially since they have different shapes), so this variable wouldn't make sense here.

packages/itwinui-css/src/tree/tree.scss
mayank99: Fixed. There was a typo (`large` -> `small`).

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: Ohh, I meant when I removed the `box-sizing` property, I noticed that the size of the thumb went from `16x16` to `14x14`. When I was working on the [vertical slider PR](https://github.com/iTwin/iTwinUI/pull/707), this property was not there. I think it came from [Leah's PR](https://github.com/iTwin/iTwinUI/pull/661) done after I worked on mine.

[Screencast from 08-17-2022 01:12:32 AM.webm](https://user-images.githubusercontent.com/45748283/185040275-20d7fdd6-a4a0-41bf-b20b-5f05e8141ff3.webm)

mayank99: Well yeah, it's not meant to be 14x14 so `content-box` was added to restore the size back to 16x16.

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
```diff
@@ -33,20 +34,20 @@
...
-      color: var(--iui-text-color);
+      color: var(--iui-color-foreground-2);
     }
   }
 
   > :any-link {
     @include iui-anchor;
     box-sizing: border-box;
-    line-height: $iui-component-height;
+    line-height: var(--_iui-breadcrumbs-height);
```
veekeys, line 50: sure we wanna have `line-height` with value of `100%` ?

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: **[RESOLVED]** Ohh, great. Was just confirming

packages/itwinui-css/src/anchor/anchor.scss
```diff
@@ -50,13 +49,12 @@
 @mixin iui-anchor-status($status: primary) {
   @include iui-focus($color: var(--iui-color-foreground-#{$status}), $offset: 1px);
   color: var(--iui-color-foreground-#{$status});
-  -webkit-tap-highlight-color: rgba(var(--iui-color-foreground-#{$status}-rgb), var(--iui-opacity-5));
   @if $status != primary {
```
veekeys, line 52: why highlight color was removed?

packages/itwinui-css/backstop/tests/information-panel.html
```diff
@@ -1310,7 +1318,8 @@ <h1>Information panel</h1>
               </label>
               <input
                 value="2021"
-                class="iui-input iui-small"
+                class="iui-input"
+                data-iui-size="small"
```
veekeys, line 1322: all of that could have been done separately...

veekeys (commented):
I havent run any of that locally and verified values, would like to firstly see the new visual images, cause it seems everything will be different.

Something I noticed and I really dont like with the "new" variables is that the naming now is so broad, that the readability is so much harder (I am not talking about values here). Some of the examples

- `icon-color` -> `foreground-n`
- `text-color` -> `foreground-n`
- `background-overlay` ->`background-hover` (this variable is also used in multiple places and not even for hover, so hard to understand if its naming issue or something else)
- `font-size-title` -> `font-size-4` (this one is pretty cruel, previously it was much better to guess what to use where we were reusing some of the common naming for different sizes)
- even the baseline change, I do feel having a baseline variable would be so much better

Some of these could be fixed by just adding new variables with duplicated values and some of them could be revisited, cause this naming with random numbers as suffix does not work well in every case as in elevation (which is the only one we have discussed and decided to go with simple numbering system).

These are just some of the thoughts, but the biggest concern is how our apps gonna migrate to the new iTwinUI version, cause I have no idea how they could verify that no layouts are broken or colors arent too different (knowing some of the apps are in the middle of migration to iTwinUI). Of course, fact is, barely any app has visual tests to ensure changes in their UI is acceptable.

Another really important thing is to have all other changes prior to this published as separate dev version, so we could see, how this single commit is influencing things in react side, also we might want to take some of the existing products and try to migrate them to our `alpha` release.

Dont forget to document all of the changes, especially variables replacement.

In general, good job. Im trying to be positive, cause all of that is leading us to something really nice and easy to use, right :)

packages/itwinui-css/src/tabs/pill.scss
```diff
@@ -1,34 +1,43 @@
...
       bottom: 0;
       left: 0;
     }
 
     &:hover {
-      background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6));
+      background-color: var(--iui-color-background-transparent-hover);
+    }
+
+    &.iui-active:hover {
```
bentleyvk, line 39: It doesn't seem to work on regular tab. But I am more of a fan of having colorful hover states as before.

packages/itwinui-css/src/textarea/textarea.scss
```diff
@@ -1,12 +1,11 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/global-variables';
 @import '../input/index';
 
 @mixin iui-textarea {
   @include iui-input;
   resize: vertical;
-  min-height: $iui-component-height;
+  overflow-y: overlay;
```
bentleyvk, line 8: ?

bentleyvk (commented):
I checked the code only, need to check visuals later.
The main concerns of mine:
 - foreground changes to background
 - previously we had variables that was clear for what it is like text-muted, font-size-heading, and now there is only number postfix that doesn't tell anything to the users
 - This one looks odd `--iui-color-background-transparent-hover`

packages/itwinui-variables/src/typography.scss
```diff
@@ -15,6 +15,6 @@
 
   --iui-font-weight-light: 300;
   --iui-font-weight-normal: 400;
-  --iui-font-weight-semibold: 500;
+  --iui-font-weight-semibold: 600;
```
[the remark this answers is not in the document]
mayank99: it's the correct value
https://github.com/iTwin/iTwinUI/blob/3b9b327ed8cf84db87dc1d61d356445cd5b051ba/packages/itwinui-css/src/style/typography.scss#L30

packages/itwinui-variables/src/sizes.scss
```diff
@@ -18,3 +18,10 @@
   --iui-border-radius-1: 4px;
```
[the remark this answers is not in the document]
mayank99: Not sure I understand what's scary or confusing. I changed 3px -> 4px

packages/itwinui-css/src/utils/icon/icon.scss
```diff
@@ -1,56 +1,51 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../../style/index';
+
+$iui-icon-color-multicolor-disabled: grayscale(100%);
+
+$iui-icon-sizes: (
+  s: var(--iui-size-s),
+  // 12px
```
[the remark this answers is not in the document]
mayank99: Looks like our pre-commit hook puts any comments that are at the of the line into the next line.

I've moved it into the previous line.

packages/itwinui-css/src/tooltip/tooltip.scss
```diff
@@ -1,26 +1,25 @@
...
-  font-size: $iui-font-size-small;
+  border-radius: var(--iui-border-radius-1);
+  font-size: var(--iui-font-size-0);
   overflow: hidden;
   max-width: 400px;
   width: max-content;
   overflow-wrap: break-word;
   box-sizing: content-box;
-  padding: round($iui-baseline * 0.33) $iui-s;
+  padding: calc(var(--iui-size-s) * 0.33) var(--iui-size-xs);
```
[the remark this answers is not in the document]
mayank99: This was a find-and-replace. I kept the same math as before

packages/itwinui-css/src/tooltip/tooltip.scss
```diff
@@ -1,26 +1,25 @@
...
   z-index: 999;
-  box-shadow: $iui-elevation-8;
+  box-shadow: var(--iui-shadow-3);
   pointer-events: none;
   background-color: rgba(0, 0, 0, var(--iui-opacity-3));
   color: var(--iui-color-foreground-accessory);
-  border: 1px solid rgba(var(--iui-color-foreground-accessory-rgb), var(--iui-opacity-4));
+  border: 1px solid rgba(255, 255, 255, var(--iui-opacity-4));
```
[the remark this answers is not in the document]
mayank99: it was not changing across themes previously too.

`-foreground-accessory` is just a fancy name for "white"
https://github.com/iTwin/iTwinUI/blob/3b9b327ed8cf84db87dc1d61d356445cd5b051ba/packages/itwinui-variables/src/colors.scss#L25

packages/itwinui-css/src/button/button.scss
```diff
@@ -1,33 +1,31 @@
...
   user-select: none;
   cursor: pointer;
   white-space: nowrap;
   border: 1px solid var(--_iui-button-border-color);
   background: var(--_iui-button-background-color);
   color: var(--_iui-button-text-color);
   gap: var(--_iui-button-gap);
-  height: var(--_iui-button-height);
+  min-height: var(--_iui-button-min-height);
+  min-width: var(--_iui-button-min-height);
```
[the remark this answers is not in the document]
mayank99: this makes the icon-button square (same as #712)

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
mayank99: but this is the same as before

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -37,11 +36,11 @@
     margin: 0;
     border-radius: 50%;
     font-style: normal;
-    font-weight: $iui-font-weight-semibold;
+    font-weight: var(--iui-font-weight-semibold);
     text-align: center;
     text-decoration: none;
     text-transform: uppercase;
-    background-color: $iui-color-dataviz-celery;
+    background-color: hsl(72, 51%, 56%);
```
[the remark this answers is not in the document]
mayank99: yeah, because those colors should not really be used anywhere. Even for `avatar` we should find a better default color

packages/itwinui-css/src/badge/badge.scss
```diff
@@ -1,27 +1,23 @@
...
 
-$iui-badge-background-default: $iui-color-background-5--light;
-$iui-badge-background-positive: $iui-color-foreground-positive--dark;
-$iui-badge-background-negative: $iui-color-foreground-negative--dark;
-$iui-badge-background-warning: $iui-color-foreground-warning--dark;
-$iui-badge-background-primary: $iui-color-foreground-primary--dark;
-$iui-badge-horizontal-padding: $iui-xs * 1.5;
-
-$iui-badge-text-color: rgba($iui-color-foreground-body--light, $iui-opacity-3--light);
+$iui-badge-background-positive: hsl(96, 45%, 78%); // foreground-positive--dark
```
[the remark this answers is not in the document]
mayank99: we don't have `--dark` or `--light` versions of variables anymore so i hardcoded it, because currently our badges are supposed to look the same across themes.

in the future, i think we want the `soft-background` colors to be themeable so we might be able to use those

packages/itwinui-css/src/blockquote/blockquote.scss
```diff
@@ -1,23 +1,22 @@
...
   display: block;
   position: relative;
-  padding: $iui-baseline $iui-m;
-  margin: $iui-baseline 0;
+  padding: var(--iui-size-s) var(--iui-size-m);
+  margin: var(--iui-size-s) 0;
   text-align: justify;
-  box-shadow: $iui-elevation-2;
-  border-radius: $iui-border-radius;
+  box-shadow: var(--iui-shadow-1);
```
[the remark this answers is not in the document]
mayank99: this was done in #675 because it just makes more sense 🤷

packages/itwinui-css/src/button/borderless.scss
```diff
@@ -21,16 +19,16 @@
...
-    --_iui-button-text-color: var(--iui-text-color-placeholder);
+    --_iui-button-text-color: var(--iui-color-foreground-5);
     --_iui-button-background-color: transparent;
     --_iui-button-border-color: transparent;
   }
 }
 
 @mixin iui-button-borderless-active {
-  --_iui-button-background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6));
+  --_iui-button-background-color: var(--iui-color-background-subtle-primary);
```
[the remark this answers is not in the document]
mayank99: well, it makes sense as it's used for the background

packages/itwinui-css/src/button/default.scss
```diff
@@ -37,11 +36,7 @@
 
 @mixin iui-button-default-active {
   @include iui-button-active;
-  --_iui-button-background-color: linear-gradient(
-    rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6)),
-    rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6))
-  ),
-    linear-gradient(var(--iui-color-background-1), var(--iui-color-background-1));
+  --_iui-button-background-color: var(--iui-color-background-subtle-primary);
```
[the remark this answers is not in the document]
mayank99: the `-background-subtle` variables are meant to replace the gradient hacks

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -73,14 +72,10 @@ $iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body
   display: flex;
   flex-wrap: wrap;
   align-content: flex-start;
-  min-height: $iui-l;
-  min-width: $iui-l;
-  gap: $iui-s;
-
-  button[data-iui-variant='borderless'] {
```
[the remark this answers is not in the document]
mayank99: not needed anymore because our borderless button is now square.

packages/itwinui-css/src/date-picker/date-picker.scss
```diff
@@ -135,7 +128,7 @@ $iui-date-picker-today-circle-size: 32px;
   }
 
   &::before {
-    border-color: rgba(var(--iui-color-foreground-accessory-rgb), var(--iui-opacity-4));
+    border-color: rgba(255, 255, 255, var(--iui-opacity-4));
```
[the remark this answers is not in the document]
mayank99: see https://github.com/iTwin/iTwinUI/pull/718#discussion_r948123325

packages/itwinui-css/src/textarea/textarea.scss
mayank99: this file doesn't exist anymore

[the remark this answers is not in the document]
mayank99: this file doesn't exist anymore

packages/itwinui-css/backstop/tests/information-panel.html
mayank99: I think these changes came from Jon's commits. This was a side effect of removing the component-height variables and aligning the logic with buttons.

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -38,9 +38,10 @@ $tick-height: $iui-baseline; // 12px
 
 @mixin iui-slider-rail {
   position: absolute;
-  width: 100%;
-  height: $track-height;
```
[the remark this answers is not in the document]
mayank99: This code is repeated down below in the `-horizontal` mixin. I think there was a bad merge in the v1 branch when vertical slider was added to the main branch.

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -162,22 +162,26 @@ $iui-active-outline: thin solid var(--iui-color-foreground-primary);
 }
 
 @mixin iui-menu-content {
-  padding: $iui-baseline ($iui-component-padding-horizontal + 1px);
+  padding-block: var(--iui-size-xs);
```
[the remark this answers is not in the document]
mayank99: For readability I guess. Jon made these changes.

packages/itwinui-css/src/input/input.scss
```diff
@@ -1,70 +1,76 @@
...
-    }
+    --_iui-input-border-color: var(--iui-color-foreground-2);
   }
 
   &[disabled] {
-    background-color: var(--iui-color-background-disabled);
-    border-color: var(--iui-color-background-disabled);
+    --_iui-input-background-color: var(--iui-color-background-disabled);
+    --_iui-input-border-color: var(--iui-color-background-disabled);
+    --_hover-color: var(--iui-color-background-disabled); // RENAME
```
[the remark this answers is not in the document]
mayank99: I don't know why it says rename, this variable has been around for 6 months

packages/itwinui-css/src/date-picker/date-picker.scss
```diff
@@ -78,47 +77,41 @@ $iui-date-picker-today-circle-size: 32px;
...
   height: $iui-date-picker-cell-height;
-  border-radius: $iui-border-radius;
+  border-radius: var(--iui-border-radius-1);
   font-variant-numeric: tabular-nums;
 }
 
 @mixin iui-calendar-day-base-hover {
   &:hover {
-    font-weight: $iui-font-weight-semibold;
-    color: var(--iui-color-foreground-primary);
```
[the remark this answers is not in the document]
mayank99: There was some large changes made in date-picker to make the hover states simpler.

packages/itwinui-css/src/date-picker/date-picker.scss
```diff
@@ -78,47 +77,41 @@ $iui-date-picker-today-circle-size: 32px;
...
-    border: 2px solid rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
+    border: 2px solid var(--iui-color-foreground-5);
   }
 
   &:hover {
-    background-color: initial;
-    color: var(--iui-color-foreground-primary);
-  }
-
-  &:hover::before {
```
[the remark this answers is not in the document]
mayank99: There was some large changes made in date-picker to make the hover states simpler.

packages/itwinui-css/src/date-picker/date-picker.scss
```diff
@@ -205,29 +198,25 @@ $iui-date-picker-today-circle-size: 32px;
...
   }
 
   &-today {
     @include iui-calendar-day-today-base;
 
     &:hover {
-      background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-5));
-      color: var(--iui-text-color);
-
-      &::before {
```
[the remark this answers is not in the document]
mayank99: There was some large changes made in date-picker to make the hover states simpler.

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -54,27 +53,23 @@
...
 
       > label {
-        font-weight: $iui-font-weight-bold;
+        font-weight: var(--iui-font-weight-bold);
       }
     }
   }
 
   &.iui-drag > .iui-content {
-    background: linear-gradient(
```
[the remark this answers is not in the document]
mayank99: the `-background-subtle` variables are meant to replace the gradient hacks

packages/itwinui-css/src/location-marker/me.scss
```diff
@@ -9,18 +8,23 @@
   position: relative;
   border-radius: 50%;
   cursor: default;
-  background: radial-gradient(
-    circle at center,
-    rgba(var(--iui-color-foreground-primary-rgb), 0),
-    rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-5))
-  );
+
+  &::before {
```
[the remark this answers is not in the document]
mayank99: This pseudo-element allows us to specify opacity

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -162,22 +162,26 @@ $iui-active-outline: thin solid var(--iui-color-foreground-primary);
...
+  padding-inline: calc(var(--iui-size-m) + 1px);
+  min-height: var(--iui-component-height);
+  display: grid;
```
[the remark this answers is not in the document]
mayank99: for centering. the height changed a little bit so it was looking off center

packages/itwinui-variables/src/themes/light.scss
```diff
@@ -57,4 +58,9 @@ $opacities: (
   --iui-color-background-transparent-hover: #{rgba(#000, map.get($opacities, 6))};
   --iui-color-background-disabled: #{hex(map.get($base-colors, background-3))};
   --iui-color-background-border: #{hex(map.get($base-colors, background-5))};
+
+  // Only HSL channels, for customizing transparency
+  --iui-color-background-1-hsl: #{get-hsl(map.get($base-colors, background-1))};
```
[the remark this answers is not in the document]
mayank99: the plan is to use these in more places

packages/itwinui-css/src/anchor/anchor.scss
mayank99: because we don't/didn't have the transparency variables (and because jon was ok with removing it). i think i might add it back though

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -240,18 +234,20 @@
...
   }
   @if $variant == quick-action {
-    right: $iui-s;
+    right: var(--iui-size-xs);
   }
 
   &[data-iui-active='true'] {
-    --_iui-button-text-color: var(--iui-color-foreground-primary);
-    --_iui-button-background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6));
+    // background-color: var(--iui-color-foreground-primary);
```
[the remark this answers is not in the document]
mayank99: removed

mayank99:
> I havent run any of that locally and verified values, would like to firstly see the new visual images, cause it seems everything will be different.

You can check the deploy preview: https://itwin.github.io/iTwinUI/718

> * `icon-color` -> `foreground-n`
> * `text-color` -> `foreground-n`

Having aliases for these sounds nice, but it just gets complicated with the different variations of icons and text, and in some cases the variables start getting misused. Whereas with a single foreground scale, you only need to look at the documentation once and it just makes sense after that

> * `background-overlay` ->`background-hover` (this variable is also used in multiple places and not even for hover, so hard to understand if its naming issue or something else)

yeah i don't think the name makes sense here

> * `font-size-title` -> `font-size-4` (this one is pretty cruel, previously it was much better to guess what to use where we were reusing some of the common naming for different sizes)

i don't agree with this in particular. previously the font-size variables seemed completely random to me, whereas now, `1` is the base font size, `0` is the smaller one, and `2`, `3`, etc are progressively getting larger.

> * even the baseline change, I do feel having a baseline variable would be so much better

how does having a baseline variable help? we never followed the baseline grid anyway, and now it's just part of our normal spacing scale.

packages/itwinui-css/backstop/tests/index.html
```diff
@@ -68,7 +68,10 @@
 
     <style>
       @import url("./assets/demo.css");
+      @import url("@itwin/itwinui-variables");
+      @import url("@itwin/itwinui-variables/os.css");
```
elephantcatdog, line 72: Why do you need to import `os.css` specifically if you've already imported `@itwin/itwinui-variables`? Is it because this is a url import?

packages/itwinui-css/src/table/table-densities.scss
```diff
@@ -1,11 +1,10 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/baseline';
 @import '../style/space';
```
elephantcatdog, line 3: `style/space` doesn't exist anymore either.

packages/itwinui-variables/src/themes/light.scss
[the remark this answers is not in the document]
veekeys: Now or after v1 release? if the latter, we dont need them for this release

packages/itwinui-css/src/button/button.scss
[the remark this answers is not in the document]
veekeys: Shouldnt it make it square with just having set icon size + padding...?
And I am not sure if `min-height` is better than `height` ? doesnt it allow the button to be larger than we expect it to be then?

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
veekeys: Well no, previously it was using `iui-component-height` variable values for `line-height`, now you are using param of the mixin value, which is passed as `100%` when breadcrumbs mixin is used in header.

packages/itwinui-css/backstop/tests/information-panel.html
veekeys: Cause then we have few more components, which also use class modifiers versus data attribute modifiers.. Sounds weird this change was required because of some variables change :/

packages/itwinui-css/src/input/input.scss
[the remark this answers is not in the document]
veekeys: > > > * `icon-color` -> `foreground-n`
> > > * `text-color` -> `foreground-n`
> 
> > Having aliases for these sounds nice, but it just gets complicated with the different variations of icons and text, and in some cases the variables start getting misused. Whereas with a single foreground scale, you only need to look at the documentation once and it just makes sense after that

Previously I knew, if I need to style icon, i use `icon-` variables, if I need to apply styling to text, i use `text-` variables. I dont think there were anymore variations of these groups. I think we even had a little different opacity for fill of icon and for text color at some point.

Can I see documentation, so it would make sense to me?

> 
> > > * `font-size-title` -> `font-size-4` (this one is pretty cruel, previously it was much better to guess what to use where we were reusing some of the common naming for different sizes)
> 
> > i don't agree with this in particular. previously the font-size variables seemed completely random to me, whereas now, `1` is the base font size, `0` is the smaller one, and `2`, `3`, etc are progressively getting larger.

So how I am supposed to know which of the number to use if I want to have a title of the section? Or which number is the `default` for general text in my app?
Shouldnt we at least start numbering from `1` so this is at least synced in some kind of a way with heading tags of html?
Does this mean our react components will become `FontElement2` too?

> 
> > > * even the baseline change, I do feel having a baseline variable would be so much better
> 
> > how does having a baseline variable help? we never followed the baseline grid anyway, and now it's just part of our normal spacing scale.

I know products have started to adopt baseline vs horizontal spacing recommendations after lots of years of pushing this, so they have their padding set to use `baseline` and `space` variables accordingly. So having an easier replace of the variables might look as an easier transition to the major release.
And in different variable it might be easier to explain, why they should use always `12px` instead of `16px` for vertical default spacing.

packages/itwinui-css/src/avatar/avatar.scss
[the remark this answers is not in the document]
veekeys: This might be big.. We have whole section about use cases for those colors.. Was this verified with the UX team...?

veekeys:
> > > * `icon-color` -> `foreground-n`
> > > * `text-color` -> `foreground-n`
> 
> > Having aliases for these sounds nice, but it just gets complicated with the different variations of icons and text, and in some cases the variables start getting misused. Whereas with a single foreground scale, you only need to look at the documentation once and it just makes sense after that

Previously I knew, if I need to style icon, i use `icon-` variables, if I need to apply styling to text, i use `text-` variables. I dont think there were anymore variations of these groups. I think we even had a little different opacity for fill of icon and for text color at some point.

Can I see documentation, so it would make sense to me?

> 
> > > * `font-size-title` -> `font-size-4` (this one is pretty cruel, previously it was much better to guess what to use where we were reusing some of the common naming for different sizes)
> 
> > i don't agree with this in particular. previously the font-size variables seemed completely random to me, whereas now, `1` is the base font size, `0` is the smaller one, and `2`, `3`, etc are progressively getting larger.

So how I am supposed to know which of the number to use if I want to have a title of the section? Or which number is the `default` for general text in my app?
Shouldnt we at least start numbering from `1` so this is at least synced in some kind of a way with heading tags of html?
Does this mean our react components will become `FontElement2` too?

> 
> > > * even the baseline change, I do feel having a baseline variable would be so much better
> 
> > how does having a baseline variable help? we never followed the baseline grid anyway, and now it's just part of our normal spacing scale.

I know products have started to adopt baseline vs horizontal spacing recommendations after lots of years of pushing this, so they have their padding set to use `baseline` and `space` variables accordingly. So having an easier replace of the variables might look as an easier transition to the major release.
And in different variable it might be easier to explain, why they should use always `12px` instead of `16px` for vertical default spacing.

packages/itwinui-css/src/input/input.scss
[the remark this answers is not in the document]
bentleyvk: Maybe rename it to meet our standards because it will never be renamed.

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -1,39 +1,38 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
 @import '../surface/index';
 @import '../button/borderless';
 @import '../utils/index';
```
bentleyvk, line 6: Why color picker border went from darker same color to just gray?

bentleyvk: this one also changed to solid color:
image

packages/itwinui-css/src/date-picker/date-picker.scss
```diff
@@ -78,47 +77,41 @@ $iui-date-picker-today-circle-size: 32px;
...
   font-variant-numeric: tabular-nums;
 }
 
 @mixin iui-calendar-day-base-hover {
   &:hover {
-    font-weight: $iui-font-weight-semibold;
-    color: var(--iui-color-foreground-primary);
-    background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6));
+    font-weight: var(--iui-font-weight-semibold);
+    background-color: var(--iui-color-background-transparent-hover);
```
bentleyvk, line 87: I don't like this change from blue to the gray as our UI are black, gray and white, we are lacking more colors and this one is removing what is left.

packages/itwinui-css/src/button/borderless.scss
bentleyvk, line 29: This color is barely visible on darker background e.g. inside info-panel 
image

packages/itwinui-css/backstop/tests/progress-indicator.html
```diff
@@ -25,9 +25,10 @@
     </script>
     <style>
       @import url("./assets/demo.css") layer(demo);
+      @import url('@itwin/itwinui-variables') layer(variables);
       @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
       section[id^='demo'] {
-        width: 33vw;
+        width: 35vw;
```
bentleyvk, line 31: Why changed?

packages/itwinui-css/src/skip-to-content/skip-to-content.scss
```diff
@@ -1,27 +1,26 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
 @import '../utils/index';
```
bentleyvk, line 4: Broken
image

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -20,7 +19,8 @@ $tick-height: $iui-baseline; // 12px
 
   svg {
     @include iui-icon-style('m');
-    margin-top: $iui-xs;
+    @include iui-icon-color;
```
bentleyvk, line 22: Is this color change ok?
image

packages/itwinui-css/src/table/table.scss
```diff
@@ -154,11 +154,7 @@
 
     &:hover:not([aria-disabled='true']) {
       .iui-table-cell {
-        background: linear-gradient(
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6)),
-          rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6))
-        ),
-          linear-gradient(var(--iui-color-background-1), var(--iui-color-background-1));
+        background-color: var(--iui-color-background-transparent-hover);
```
bentleyvk, line 157: I don't like this color change.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 // See LICENSE.md in the project root for license terms and full copyright notice.
-@import '../style/index';
 @import '../button/index';
```
bentleyvk, line 3: Idk if this is known but icons background is missing 
image

packages/itwinui-css/src/code/codeblock.scss
```diff
@@ -39,21 +38,21 @@
       white-space: nowrap;
       hyphens: auto;
       counter-increment: section;
-      font-family: $iui-monospace;
+      font-family: var(--iui-font-mono);
       background-color: var(--iui-color-background-1);
 
       &:nth-child(even) {
-        background: rgba(var(--iui-color-foreground-body-rgb), 0.02);
+        background: var(--iui-color-background-1-hover);
```
bentleyvk, line 45: Why on dark theme these got inverted? 
image

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: I'm not sure, but I think the grey one is the correct color. That is how it looks on the website. It was also grey when I was working on the [vertical slider PR](https://github.com/iTwin/iTwinUI/pull/707)

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -102,37 +110,39 @@ $tick-height: $iui-baseline; // 12px
     cursor: not-allowed;
     background-color: var(--iui-color-background-4);
     border-color: var(--iui-color-background-4);
+    @media (forced-colors: active) {
```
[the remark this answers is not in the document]
mayank99: This might have been a bad merge in the v1 branch from when the vertical slider PR was merged into main
https://github.com/iTwin/iTwinUI/blob/3b9b327ed8cf84db87dc1d61d356445cd5b051ba/packages/itwinui-css/src/slider/slider.scss#L121

packages/itwinui-css/src/input/input.scss
```diff
@@ -1,70 +1,76 @@
...
-    font-size: $iui-font-size-leading;
-  }
+  min-height: var(--_iui-input-min-height);
+  padding-block: var(--_iui-input-padding-block);
+  padding-inline: var(--iui-size-s);
+  color: var(--_iui-input-text-color);
+  background-color: var(--_iui-input-background-color);
+  border: 1px solid var(--_iui-input-border-color);
+  transition: border-color var(--iui-duration-1) ease-out;
+  box-sizing: border-box;
```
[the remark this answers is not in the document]
mayank99: I think these were again done by Jon. I've removed the `box-sizing` property because it's now set globally, but I've kept the removal of media query as it's not necessary.

packages/itwinui-css/backstop/tests/index.html
mayank99: `@itwin/itwinui-variables` refers to `index.css`

we specifically want to also import `os.css` in `index.html` because there is no javascript theme switcher in here.

packages/itwinui-css/backstop/tests/progress-indicator.html
mayank99: this was done to prevent line wrapping as the sizes are now slightly bigger

packages/itwinui-css/src/tile/tile.scss
mayank99: where is that from? i'm not seeing any missing icons

packages/itwinui-css/src/button/borderless.scss
mayank99: Should be fixed now

mayank99:
> Might need to introduce some custom styling for the active state in paginator as we have some really bad contrasting colors.
> 
> Screen Shot 2022-08-12 at 12 06 12 PM

I'm now using `foreground-primary-hsl` + `opacity-6` for button active states, so this is fixed now (behaving same as before).

packages/itwinui-css/src/tabs/pill.scss
mayank99: what doesn't work? this code is for pill tabs

packages/itwinui-css/src/code/codeblock.scss
mayank99: Should be fixed now

packages/itwinui-css/src/skip-to-content/skip-to-content.scss
mayank99: Fixed

packages/itwinui-css/src/anchor/anchor.scss
mayank99: added back

packages/itwinui-variables/src/themes/light.scss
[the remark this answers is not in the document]
mayank99: using these in a few more places now

packages/itwinui-css/src/expandable-block/block.scss
```diff
@@ -14,11 +13,11 @@
     @include iui-focus;
     display: flex;
     align-items: center;
-    padding: $iui-baseline $iui-sm;
+    padding: var(--iui-size-s) var(--iui-size-s);
```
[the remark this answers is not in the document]
mayank99: done

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -1,39 +1,38 @@
...
-$iui-color-picker-max-width: ($iui-l * 10) + ($iui-s * 9); // (10 swatches per row) + (gaps)
+$iui-color-picker-content-width: calc(var(--iui-size-l) * 9 + var(--iui-size-xs) * 8); // (9 swatches per row) + (gaps)
+$iui-color-picker-max-width: calc(var(--iui-size-l) * 10 + var(--iui-size-xs) * 9); // (10 swatches per row) + (gaps)
 $iui-color-swatch-border-radius: 5px;
-$iui-inset-box-shadow: inset 0 0 0 1px rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
-$iui-active-box-shadow: 0 0 0 $iui-xxs rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-3));
-$iui-hover-box-shadow: 0 0 1px $iui-xxs + 1 rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-5));
+$iui-inset-box-shadow: inset 0 0 0 1px var(--iui-color-foreground-5);
+$iui-active-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-foreground-primary);
+$iui-hover-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-foreground-4); // 0 0 1px calc(var(--iui-size-3xs) + 1px) var(--iui-color-background-1-hover);
```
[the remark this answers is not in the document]
mayank99: i think this is intentional. When pairing with Jon, we changed the visuals a little bit

packages/itwinui-css/src/button/button.scss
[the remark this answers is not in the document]
mayank99: setting just one expected value sounds a lot simpler than relying on icon size + padding

also, `min-height` makes sense because we want the height to increase if the content inside is larger for whatever reason. Setting fixed `height` should generally be avoided

packages/itwinui-css/src/color-picker/color-picker.scss
mayank99: I think these changes are intentional and verified

packages/itwinui-css/src/table/table.scss
mayank99: This was done to make our hover states more consistent

packages/itwinui-css/src/header/header-buttons.scss
```diff
@@ -163,16 +163,19 @@
...
-    background-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-6));
+  &::before {
+    content: '';
+    position: absolute;
+    inset: 0;
   }
 
-  @at-root [aria-current] &:hover {
-    background-color: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-6));
+  &:hover::before {
```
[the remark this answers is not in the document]
mayank99: this was needed because we didn't have transparency variables. i've now reverted this

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
mayank99: Reverted 👍

rohan-kadkol:
> > I'm seeing a visual bug with the dropdown icon in the small select within the information panel demo:
> > Screen Shot 2022-08-12 at 10 23 05 AM
> 
> @rohan-kadkol Did you get a chance to look at this^?

Sorry for the delay. I am reposting this here just in case someone else knows why this could be happening and how to fix it.

I have been trying to figure out the dropdown menu button for quite some time now. Unfortunately, I haven't found the solution so far. All I notice is that:
- When passing a width to the `span` with class of `iui-input-icon iui-actionable`, it for some reason also included the padding. E.g. when I pass a width of `19px`, the width becomes `1px` because of a `9px` padding on both the left and the right (`9px` + `9px` = `18px` and `19px` - `18px` = `1px` which becomes the width). But when I pass the same width of `19px` in the hosted demo website's same `span`, the box gets the width of `19px`.
image
Screenshot from 2022-08-24 16-34-19

- When removing the width attribute from line 27 of icon.scss, for some reason, the span with class automatically gets a width of 16px and looks almost correct.
image
Screenshot from 2022-08-24 16-35-45

mayank99:
I've fixed all remaining issues and gone through each component one to verify, before updating test images.

Next thing I'm going to document all the variables and create a migration guide to help users switch over from old sass variables. But that shouldn't really hold this PR.

packages/itwinui-css/src/utils/input-container/input-container.scss
```diff
@@ -186,10 +186,11 @@
 
   &.iui-actionable {
     align-items: center;
-    height: 90%;
+    height: calc(100% - 4px);
```
veekeys, line 189: comment for 4px might be useful

veekeys (commented):
SOme points I noticed while going through images.
- Why `Alert` close icon moved to the right
- Now seems our buttons are smaller height
- disabled checkbox checkmark color changes is connected with increase in wcag score, I assume
- `CodeBlock` title now less height
- `ColorPicker` thumb to choose color (or w/e is called) lost border - IMPORTANT
- Color swatches now have some weird border and colors look blurry compared with previous version (might be images issues though)
- Now seems focus and selected outlines are the same? thats weird
- In iTwinUI_dark_button_Grouped_Borderless_0_demo-grouped-borderless_0_desktop disabled llooks also changed
- Menu hover color looks bad :( (we talked about it)
- Why some screenshots increased width??? i.e. prgoress indicator
- Vertical slider icons look misaligned - IMPORTANT
- Table has lost selected border in between different color
- Why only some od the disabled states colors have changed i.e. radio tile, but not toggle switch (in dark mode)

In general, looks weird, because in most of the places height has become less (seems baseline went back from 12 to 11 :/

packages/itwinui-css/src/utils/input-container/input-container.scss
mayank99: I'm subtracting 2px on both sides to avoid overlapping with border/outline

Added comment

mayank99:
> * Why `Alert` close icon moved to the right
> * Now seems our buttons are smaller height

button size changes are intentional

> * disabled checkbox checkmark color changes is connected with increase in wcag score, I assume

I don't think I see a difference. If there is a small difference, it's likely because of transparency changes, which is expected in most places.

> * `CodeBlock` title now less height

i think this is because of the copy button, which is intentional

> * `ColorPicker` thumb to choose color (or w/e is called) lost border - IMPORTANT

fixed

> * Color swatches now have some weird border and colors look blurry compared with previous version (might be images issues though)

i think those color swatch changes are intentional

> * Now seems focus and selected outlines are the same? thats weird

they still look different to me

> * In iTwinUI_dark_button_Grouped_Borderless_0_demo-grouped-borderless_0_desktop disabled llooks also changed

aligning disabled styling for default vs borderless buttons is intentional

> * Why some screenshots increased width??? i.e. prgoress indicator

i increased the width because i noticed some of the sections were wrapping

> * Vertical slider icons look misaligned - IMPORTANT

I don't think this is related to my PR. It looks like the vertical slider spacing could use a bit of cleanup so I've opened a new issue: https://github.com/iTwin/iTwinUI/issues/753

> * Table has lost selected border in between different color

Should be fixed now

> * Why only some od the disabled states colors have changed i.e. radio tile, but not toggle switch (in dark mode)

Not sure I see any changes in radio tile

packages/itwinui-css/src/tabs/pill.scss
bentleyvk: tab-v1-hover

packages/itwinui-css/src/color-picker/color-picker.scss
bentleyvk: For real??? Previously swatches looked better in my opinion.

packages/itwinui-css/src/tile/tile.scss
bentleyvk: ok, why background was changed from darker to this, barely visible one? Now it is harder to see that there is an icon in the corner if image is bright and previously we didn't have such problem since icon background was darker.

mayank99: I'm still not sure what part you're talking about. 

https://itwin.github.io/iTwinUI/718/tile.html and https://itwin.github.io/iTwinUI/tile.html look the same

bentleyvk: I am seeing this in local build. As usual build must be fucked up...

gretanausedaite (commented):
Is this color picker border change intentional? 
image

mayank99:
> Is this color picker border change intentional?

I believe it is, because it was harder to see previously. But I can also revert it if we don't want random changes like this.

FlyersPh9:
> > Is this color picker border change intentional?
>
> I believe it is, because it was harder to see previously. But I can also revert it if we don't want random changes like this.

Here's what the [production color picker](https://itwin.github.io/iTwinUI-react/?path=%2Fstory%2Fcore-colorpicker--advanced) looks like when in light theme & placed in the top left corner in light theme. Pretty much invisible unless you hover it. 😕
Screen Shot 2022-09-01 at 2 25 52 PM Screen Shot 2022-09-01 at 2 27 13 PM

But it seems like we have the same issue when the border is set to black and placed in the bottom left corner.
Screen Shot 2022-09-01 at 2 29 35 PM Screen Shot 2022-09-01 at 2 30 09 PM

elephantcatdog (approved):

gretanausedaite (approved):
Variables looks good to me.
We should fix ColorPicker border issue in next PR before releasing v1.

bentleyvk:
Are you planning to fix these?
https://github.com/iTwin/iTwinUI/pull/718#discussion_r948006908
https://github.com/iTwin/iTwinUI/pull/718#discussion_r947907694

rohan-kadkol (approved):

packages/itwinui-css/src/input/input.scss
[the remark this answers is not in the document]
mayank99: Renamed

packages/itwinui-css/src/tabs/pill.scss
mayank99: Fixed

mayank99:
> Are you planning to fix these? [#718 (comment)](https://github.com/iTwin/iTwinUI/pull/718#discussion_r948006908) [#718 (comment)](https://github.com/iTwin/iTwinUI/pull/718#discussion_r947907694)

Somehow missed these because github decided to hide them. Fixed them just now

bentleyvk (approved):

PR #719, opened 2022-07-26
feat(Button): Refactor split button

### LINKED ISSUE #706: [v1] SplitButton refactor
Currently we are using button-group styles to construct a split-button. This might be unnecessary and results in a weird DOM structure where each button is wrapped in a `<div>`. Need to investigate if we can improve the DOM structure by avoiding button-group styles.

FlyersPh9:
- `split-button.scss` no longer imports `iui-button-group`.
- Specificity of split button is 0, 2, 0 at it's highest.
- `.iui-button-split` no longer wraps buttons in `<div>`.

Closes #706.


packages/itwinui-css/src/button/split-button.scss
```diff
@@ -3,43 +3,54 @@
...
+      border-right-color: rgba(var(--iui-color-foreground-accessory-rgb), var(--iui-opacity-4));
+    }
+
+    &[disabled],
+    &:disabled,
+    &[aria-disabled='true'] {
+      border-right-color: var(--iui-color-background-4);
+    }
+
+    &:where(:enabled:hover) {
```
mayank99, line 25: We should not be using `:enabled` as it won't work with the `[disabled]` or `aria-disabled` approaches.

Instead, this block should be moved up before the disabled block, and then we should set another z-index (maybe -1) for the disabled block.

packages/itwinui-css/src/button/split-button.scss
```diff
@@ -3,43 +3,54 @@
...
 
-  &:hover {
-    background-color: rgba(var(--iui-color-foreground-body-rgb), var(--iui-opacity-6));
+  :where(.iui-button:first-child) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+
+    &[data-iui-variant='high-visibility'],
+    &[data-iui-variant='cta'] {
+      border-right-color: rgba(var(--iui-color-foreground-accessory-rgb), var(--iui-opacity-4));
```
mayank99, line 16: Can we use a different value for this part? We won't be able to do this with the new variables package.

packages/itwinui-css/src/button/split-button.scss
```diff
@@ -3,43 +3,54 @@
...
+    }
+
+    &[disabled],
```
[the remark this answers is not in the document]
FlyersPh9: I just matched what we have in [`button.scss` line 62](https://github.com/iTwin/iTwinUI/blob/0425f911ba23869f341421de7d948cc667b48a3d/packages/itwinui-css/src/button/button.scss#L62).

packages/itwinui-css/src/button/split-button.scss
FlyersPh9: Oh, I hadn't thought that `:enabled` wouldn't effect `aria-disabled` or `[disabled]`.

I've made some changes that work with all 3 of the examples in the screenshot below (not saying they are valid cases, but the z-index is ordered correctly for all situations).
Screen Shot 2022-07-27 at 12 12 57 PM

mayank99: i think we don't need this extra rule
image

any time we have to write `&, &:hover`, that feels like a mistake. Does adding `z-index: -1` (without `:hover`) after line 28 not work?

packages/itwinui-css/src/button/split-button.scss
mayank99: > that variable combo is the same used in [default.scss border](https://github.com/iTwin/iTwinUI/blob/0425f911ba23869f341421de7d948cc667b48a3d/packages/itwinui-css/src/button/default.scss#L9).

no, that is using `foreground-body`, this is using `foreground-accessory`, for which we don't have a scale. (see [current variables list](https://unpkg.com/@itwin/itwinui-variables))

If we are ok with the visuals in `background-primary-hover`/`background-positive-hover`, then that should be fine.

packages/itwinui-css/src/button/split-button.scss
FlyersPh9: The middle example has a z-index order issue, but that may be a non-issue since this button combo may be invalid.

Screen Shot 2022-07-27 at 2 56 16 PM

packages/itwinui-css/src/button/split-button.scss
FlyersPh9: I'm ok with it for now, I want to update the base light & dark theme to meet WCAG AA anyways and this should be improved then.

bentleyvk: I like white/bight one more 😕

FlyersPh9: Ok, lets revert to use our original color for this PR & we will address the new color in #718.

bentleyvk (approved):

packages/itwinui-css/src/button/split-button.scss
mayank99: Why not just apply z-index: -1 on the disabled button instead of unsetting it?

packages/itwinui-css/src/button/split-button.scss
mayank99: It's also gonna be blocking #718 so we need to have this conversation one way or another.

mayank99: Actually I think we can just hardcode `0, 0, 0`, there is no need for a new color variable.

packages/itwinui-css/src/button/split-button.scss
FlyersPh9: This suggestion works and keeps specificity down at 0, 2, 0!  Change made.

packages/itwinui-css/src/button/split-button.scss
FlyersPh9: That makes sense to me, in #718 we can hard code a color since the value does not change from theme to theme.

mayank99: Can we do it in this PR? I don't want to deal with merge conflicts again.

FlyersPh9: Ok, lines 18 & 42 are now using `--_iui-button-split-divider-color` which is set to `rgba(255, 255, 255, var(--iui-opacity-4));` which results in no visual change from what is live right now.

mayank99: there is no need for a css variable here, i would either make it a sass variable or hardcode it. css variables should be for values that need to change later

FlyersPh9: Replaced with a SASS variable.

PR #720, opened 2022-07-26
chore(Tile): Update buttons to use `data-iui-variant` and `data-iui-size`

FlyersPh9:
Some buttons did not get updated to use the new data attributes.


marashad001 (approved):

mayank99 (approved):
PR title should not say "fix" as there is no change in the component code.

PR #727, opened 2022-07-29
fix(Table): Zebra stripes content no longer goes on top of stick column

### LINKED ISSUE #691: Zebra rows render on top of sticky columns
When scrolling horizontally, zebra-styled cells appear on top of stickied columns.

#### Expected behavior
Stickied columns should appear on top of everything else.

#### Reproduction
https://codesandbox.io/s/itwinui-react-example-forked-3dghps?file=/src/App.tsx

##### Steps to reproduce
Scroll horizontally.

bentleyvk:
zebra-on-top-fixed

Closes #691

### DIFF
--- a/packages/itwinui-css/src/table/table.scss
+++ b/packages/itwinui-css/src/table/table.scss
@@ -196,6 +196,11 @@
   &.iui-zebra-striping .iui-row:nth-child(even):not(.iui-selected) .iui-cell {
     @include themed {
       background-color: rgba(t(iui-color-foreground-body-rgb), 0.02);
+      background: linear-gradient(
+        rgba(t(iui-color-foreground-body-rgb), 0.02),
+        rgba(t(iui-color-foreground-body-rgb), 0.02)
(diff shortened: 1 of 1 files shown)


veekeys (approved):

packages/itwinui-css/src/table/table.scss
```diff
@@ -196,6 +196,11 @@
   &.iui-zebra-striping .iui-row:nth-child(even):not(.iui-selected) .iui-cell {
     @include themed {
       background-color: rgba(t(iui-color-foreground-body-rgb), 0.02);
```
mayank99, line 198: looks like this `background-color` gets overridden by the `background` in next line so it shouldn't be needed

```suggestion
```

bentleyvk: Oh wow, forgot to remove old styling. Good catch!

mayank99 (approved):

Issue #586, opened 2022-04-12
Modal support of icon


pranaykute06:
#### Feature
Add support for icon in the modal.

User can provide the required icon if they want to.

such modal can be used for delete confirmation etc....

#### Examples
![image](https://user-images.githubusercontent.com/82197220/162974880-4daaa9d2-68ee-4a5e-947e-10c3eb8f69ae.png)

![image](https://user-images.githubusercontent.com/82197220/162974904-6be3fa47-94ac-4327-af8e-79e45b7f5960.png)

mayank99:
Hi @pranaykute06, thanks for the issue.

This would be a(nother) great use case for a generic "text with icon" component. I don't think we would want to add any modal-specific CSS for this.

@FlyersPh9 What is the correct spacing that should be applied here?

FlyersPh9:
Areas that I think could make use of the icon + text + right aligned stuff that I can think of:
- Information panel title bar
- Modal title bar
- Expandable block title bar
- Menu items

I would probably replicate what is present in information panel, which is an 8px gap using `$iui-s`.
![Screen Shot 2022-04-12 at 5 43 22 PM](https://user-images.githubusercontent.com/849817/163059424-1fe9ba4a-a054-4b55-b918-00e2fc634834.png)

FlyersPh9:
A mixin like this could be reused in a lot of the places I mentioned in my last comment:
```scss
@mixin iui-title-bar($title-bar-label-font-size: $iui-font-size-subheading, $title-bar-sublabel-font-size: $iui-font-size-small) {
  justify-content: space-between;

  &,
  &-left,
  &-right {
    display: flex;
    align-items: center;
  }

  &-left,
  &-right {
    gap: $iui-s;
  }

  &-icon {
    display: flex;
    width: $iui-icons-default;
    height: $iui-icons-default;
    @include themed {
      fill: t(iui-icons-color);
    }
  }

  &-title {
    display: flex;
    flex-direction: column;
    flex: auto;
    min-width: 0;
    @include themed {
      color: t(iui-color-foreground-body);
    }

    &-label,
    &-sublabel {
      white-space: nowrap;
      overflow: hidden;
      text-overflow: ellipsis;
    }

    &-label {
      font-size: $title-bar-label-font-size;
      @include themed {
        opacity: t(iui-opacity-2);
      }
    }

    &-sublabel {
      font-size: $title-bar-sublabel-font-size;
      @include themed {
[...]

PR #729, opened 2022-08-02
fix(Dialog): Changed `space-between` inside title bar

bentleyvk:
`space-between` was bad for customization so changed it with a margin.
~Why I chose margin on the close icon~:
#### Margin on title:
| Pros | Cons |
| ------|------|
| User doesn't need to do anything to place close button on the right. | If user wants to have something, e.g. icon, on the right side of the title, he will need to override CSS. |

#### Margin on close button:
| Pros | Cons |
| -----|-------|
| User can have something on the right side of the title easily. |  Requires new class on the close button. |
| | User needs to know this new close button class. Solution: have sub-component. |

### DIFF
--- a/packages/itwinui-css/src/dialog/dialog.scss
+++ b/packages/itwinui-css/src/dialog/dialog.scss
@@ -137,15 +137,14 @@ $iui-dialog-width: 380px;
 }
 
 @mixin iui-dialog-title {
+  display: flex;
+  align-items: center;
   font-size: inherit;
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
   line-height: 1.5;
-}
-
-@mixin iui-dialog-title-close-button {
-  margin-left: auto;
+  margin-right: auto;
 }
 
 @mixin iui-dialog-title-bar {
--- a/packages/itwinui-css/src/dialog/dialog.scss
+++ b/packages/itwinui-css/src/dialog/dialog.scss
@@ -144,11 +144,14 @@ $iui-dialog-width: 380px;
   line-height: 1.5;
 }
 
+@mixin iui-dialog-title-close-button {
+  margin-left: auto;
+}
+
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -144,11 +144,14 @@ $iui-dialog-width: 380px;
   line-height: 1.5;
 }
 
+@mixin iui-dialog-title-close-button {
+  margin-left: auto;
```
mayank99, line 148: I would have still gone with margin on title. Is it really a con that you can't have anything on the immediate right of the title? The user could put their stuff inside the title.

Plus the same con would also apply to the close icon, as you can't easily put stuff on the immediate left of the icon.

bentleyvk: But there are less chances that you would want to put something next to the close icon. And putting inside a title doesn't work normally as you need additional classes to center things inside.

mayank99: > But there are less chances that you would want to put something next to the close icon

I can't find the screenshot but @FlyersPh9 had a request from someone who wanted an icon right next to the close icon.

> And putting inside a title doesn't work normally as you need additional classes to center things inside.

Maybe we should handle that then? It would also help towards https://github.com/iTwin/iTwinUI/issues/586

FlyersPh9: We've had 2 requests that i can recall:
1. An icon in front / to the left of the dialog title text.
2. More than 1 borderless button where the close icon is located.

bentleyvk: Ok, moved margin to the title and added some centering logic to it.

mayank99 (approved):

adamhock (approved):
LGTM 🎉

PR #739, opened 2022-08-11
fix(Dialog): Add `position: fixed` as an option for dialogs

### LINKED ISSUE #738: Dialog: backdrop is placed badly inside scrollable container
image

In order to reproduce add `margin-top: 100vh` to `iui-input-container` inside demo.

Changing from `position: absolute` to `position: fixed` seems to fix the issue.

FlyersPh9:
- Changed backdrop `position: absolute;` to `position: fixed;`.

Bug can be observed if you set `<body style="height: 2000px">`, open a dialog with a backdrop, and scroll down.  Closes #738.


packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -5,12 +5,11 @@
 @mixin iui-backdrop {
   z-index: 998;
   isolation: isolate;
-  position: absolute;
+  position: fixed;
   top: 0;
   left: 0;
   width: 100%;
   height: 100%;
-  background-color: rgba(0, 0, 0, 0.4); // IE fallback
```
bentleyvk, line 13: Restore this

packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -5,12 +5,11 @@
 @mixin iui-backdrop {
   z-index: 998;
   isolation: isolate;
-  position: absolute;
+  position: fixed;
```
mayank99, line 8: I am not sure if `position: fixed` is the best way to go about this, because then backdrop will only work for full-screen elements

elephantcatdog: I added a backdrop to the container dialog example, and it seems to work fine there.
image
What case are you thinking of?

packages/itwinui-css/src/backdrop/backdrop.scss
elephantcatdog: Restored

packages/itwinui-css/src/backdrop/backdrop.scss
mayank99: it works with the dialog wrapper because of this line:
https://github.com/iTwin/iTwinUI/blob/3b9b327ed8cf84db87dc1d61d356445cd5b051ba/packages/itwinui-css/src/dialog/dialog.scss#L16

but backdrop is meant to be a more generic component that will not necessarily be used with dialog

elephantcatdog: So I made a new example with only the backdrop inside a div. Is this what you mean? Even with `position: absolute` the backdrop already covers the full page, not the space inside the div.

image

bentleyvk: Maybe we need some modifier? If it is full-screen then it should be `position: fixed`, if not then `position: absolute`

elephantcatdog: I added:
image

I also considered making the acceptable options for `$position` either `fixed` or `absolute` and writing it like `position: #{$position}`. That way is a lot more succinct programming wise, but the current way is more clear why there's two different options. I can change it to the other way, or another way, if you guys think that'd be better.

bentleyvk: I think it will be enough to have just a simple modifier that overrides position. Because now you create two backdrop styles that only differ in position but has all the same properties like background color, transition...

packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -3,9 +3,9 @@
 @import '../style/index';
 
 @mixin iui-backdrop {
+  position: fixed;
```
bentleyvk, line 6: Could you move back where it was so it won't show that you made a change?

packages/itwinui-css/src/backdrop/backdrop.scss
```diff
@@ -24,4 +24,8 @@
   &.iui-backdrop-visible {
     transition-delay: $iui-speed-instant; // remove delay for entry animation
   }
+
+  &.iui-backdrop-contained {
```
bentleyvk, line 28: It would be great to get other native speaker opinion whether this name is ok.

mayank99: i think it should be flipped: `absolute` by default and then this modifier can be called `iui-backdrop-fixed`.

packages/itwinui-css/src/backdrop/backdrop.scss
elephantcatdog: I moved it

packages/itwinui-css/src/backdrop/backdrop.scss
elephantcatdog: I flipped it

mayank99 (approved):

adamhock (approved):
Looks good to me! 🎉

bentleyvk (approved):

PR #741, opened 2022-08-11
fix(Menu): Define color on menu item

### LINKED ISSUE #683: Menu item text color is not set
After updating iTwinUI react from 1.23.2 to 1.37.1 menu item text color is not explicitly set. As we were wraping menu item with anchor, now we need to explicitly override anchor text color. Before: 
image
After:
image

FlyersPh9:
- Define `color` on menu item.

Tested by wrapping menu item in an `<a>`.  Closes #683.


bentleyvk:
If you wrapped `li` with an `a` then it is a bad test because it is an invalid HTML structure. `ul`/`ol` can only have `li` elements as direct children. Can you try using `a` inside `li` and tell whether your fix still works.

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -40,6 +40,9 @@ $iui-active-outline: thin solid t(iui-color-foreground-primary);
   cursor: pointer;
   box-sizing: border-box;
   line-height: normal;
+  @include themed {
+    color: t(iui-text-color);
```
mayank99, line 44: This variable and the themed mixin are both removed in v1, so it doesn't make sense to release this fix to `main` branch, especially since it's not urgent

mayank99 (changes requested):

packages/itwinui-css/src/menu/menu.scss
elephantcatdog: Jon/Adam should check that the issue is fixed in v1, and if it is then we can label the issue as ready for release and not merge this.

adamhock: I agree with Mayank that this issue should be addressed after v1 is merged into `main`

veekeys: So use css variable and merge? Its new addition, not something we break.

adamhock: This has been changed

mayank99: there is no `iui-text-color` in variables package so this would still cause conflicts

mayank99 (approved):

veekeys:
Anything else left in here before we merge? @adamhock 
anchor issue we could address later on, right @mayank99  ?

PR #743, opened 2022-08-12
feat: Remove `iui-textarea`

mayank99:
`iui-textarea` only has `resize` rules in addition to `iui-input`, so it doesn't need to be a separate component. I've moved the resize declaration to `iui-input`, it will have no effect when used on a `<input>` element.


packages/itwinui-css/src/input/input.scss
```diff
@@ -57,6 +57,8 @@
     cursor: not-allowed;
     --_hover-color: var(--iui-color-background-disabled);
   }
+
+  @include textarea;
```
veekeys, line 61: Can we apply this only for `textarea` elements? Looks invalid, when we have `resize` prop on `input`.
We can use your loved `:where` to keep specificity down.

veekeys: Trash in css, which is not needed

mayank99: This looks like a lot more "trash" to me


but changed

veekeys: Something got messed up with lock file?

mayank99: i just ran `yarn` when i merged v1 branch here. maybe it was messed up in the v1 branch

mayank99: pushed updated lockfile in v1 branch so it should not show changed here anymore

veekeys (approved):

PR #745, opened 2022-08-15
feat: Remove sass from deliverable

### LINKED ISSUE #648: Missing md files in published package
I was converting this repo to a monorepo when I noticed that our published package doesn't include `LICENSE.md` and `CHANGELOG.md`.
https://unpkg.com/browse/@itwin/itwinui-css@0.58.0/

Also the `main` field in package.json points to a non-existing `src/index.scss`.

### LINKED ISSUE #737: v1: Remove sass from published package
With itwinui-css 1.0 + itwinui-variables, there should be no need for our users to access sass mixins or variables. We should remove it from our deliverable (npm package).

mayank99:
1. Removed `build:scss` command so the sass files are not copied into the output.
2. Removed "scss" from the "files" field in package.json
3. Changed dev server to use `sass --watch` so that it builds incrementally instead of waiting for the whole css build.
   - Used autoprefixer from vite postcss config rather than `generateCss.js`

Closes #737, closes #648


packages/itwinui-css/package.json
```diff
@@ -3,19 +3,16 @@
...
-  "main": "src/index.scss",
   "files": [
     "css",
-    "scss",
     "CHANGELOG.md",
     "LICENSE.md"
   ],
-  "description": "CSS (Sass/SCSS) library for building beautiful and well working web UI components within Bentley Systems & iTwin.js applications.",
+  "main": "css/all.css",
+  "description": "CSS library for building beautiful and well working web UI components within Bentley Systems & iTwin.js applications.",
```
gretanausedaite, line 12: Not related to this change: Are we still targeting only Bentley Systems & iTwin.js applications?

gretanausedaite (approved):

packages/itwinui-css/package.json
mayank99: I _think_ it's still accurate? Not sure who we are targeting tbh 😅

packages/itwinui-css/package.json
```diff
@@ -36,19 +33,19 @@
     "npm-run-all": "^4.1.5",
     "postcss": "^8.4.5",
     "postcss-discard-comments": "^5.0.1",
-    "postcss-scss": "^4.0.3",
+    "postcss-scss": "^4.0.4",
+    "sass": "^1.54.4",
```
bentleyvk, line 37: Why not `sass-embeded`?

bentleyvk: Oh, I see it it left bellow. Then wtf 😁

packages/itwinui-css/package.json
```diff
@@ -3,19 +3,16 @@
...
     "CHANGELOG.md",
     "LICENSE.md"
   ],
-  "description": "CSS (Sass/SCSS) library for building beautiful and well working web UI components within Bentley Systems & iTwin.js applications.",
+  "description": "CSS library for building beautiful and well working web UI components within Bentley Systems & iTwin.js applications.",
   "homepage": "https://github.com/iTwin/iTwinUI",
   "keywords": [
     "css",
-    "sass",
-    "scss",
```
bentleyvk, line 18: Maybe remove it from `.gitignore` too.

packages/itwinui-css/vite.config.ts
```diff
@@ -17,6 +17,9 @@ export default defineConfig({
       },
     },
   },
+  css: {
+    postcss: { plugins: [require('autoprefixer')] }
```
bentleyvk, line 21: Does it take our `.browserlistrc` since it is in the root?

packages/itwinui-css/package.json
mayank99: I kept it in the gitignore for now because otherwise everyone will have a bunch of untracked scss files when switching branches

packages/itwinui-css/package.json
mayank99: `sass` is used for the `-watch` flag in `scripts`
`sass-embedded` is used for the build (in `generateCss.js`)

packages/itwinui-css/vite.config.ts
mayank99: I'm not sure but I don't think it matters much, because this is just for the dev server.

packages/itwinui-css/package.json
bentleyvk: Can't you use `sass-embeded` for both?

mayank99: `sass-embedded` doesn't come with a CLI or watch flag

packages/itwinui-css/vite.config.ts
bentleyvk: It matters because I think nobody checks the result with "prod" build.

mayank99: Visual tests still run on the prod build, and the github pages deployment also uses the prod build.

But to verify, I logged autoprefixer info at both project root and inside vite config and got the same output.

bentleyvk: Huh, IE11, when it is removed in v1 branch.

PR #750, opened 2022-08-22
feat: Move visually-hidden from global to utils

mayank99:
Was working on global styles when i noticed that `iui-visually-hidden` is just sitting in `global.scss` for no reason, so I've moved it to utils.

Also noticed that `iui-display-none` is not used anywhere so removed it.


packages/itwinui-css/src/utils/visually-hidden/classes.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import './visually-hidden';
+
+.iui-visually-hidden {
```
veekeys, line 5: will be important for react to include this css in `Menu` where we use this class.

veekeys (approved):

PR #752, opened 2022-08-25
fix(Table): Keyboard Navigation Support

elephantcatdog:
I changed all `visibility: hidden` to `opacity: 0` and all `visibility: visible` to `opacity: revert`. This helps to solve the problem of keyboard navigation in the table in the react repo.

I tried to use `iui-visually-hidden` and `iui-visually-hidden-revert`, but it negatively affected the look of the table. I thought this was the better solution to 1) keeping the button in the DOM, 2) keeping the button dimensions the same throughout, and 3) allowing keyboard navigation.

Goes with https://github.com/iTwin/iTwinUI-react/pull/802

### DIFF
--- a/packages/itwinui-css/src/table/table.scss
+++ b/packages/itwinui-css/src/table/table.scss
@@ -88,7 +88,7 @@
 
       .iui-table-sort,
       .iui-table-filter-button {
-        opacity: revert;
+        opacity: 1;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


packages/itwinui-css/src/table/table.scss
```diff
@@ -88,7 +88,7 @@
 
       .iui-table-sort,
       .iui-table-filter-button {
-        visibility: visible;
+        opacity: revert;
```
bentleyvk, line 91: Minor, but `opacity: 1` looks better to me.

adamhock (approved):
I agree that `opacity: 1` looks a bit better than `opacity: revert`

packages/itwinui-css/src/table/table.scss
elephantcatdog: I made this change.

PR #754, opened 2022-08-29
fix(Slider): Fixed vertical & horizontal slider icons misalignment

### LINKED ISSUE #753: Vertical slider icons are misaligned
Might need to do some cleanup and fixes on how things are spaced/aligned in the vertical slider.

_Originally posted by @veekeys in https://github.com/iTwin/iTwinUI/pull/718#pullrequestreview-1086899678_

rohan-kadkol:
Closes #753 

- Removed the hardcoded `width` in `slider.html` for the min and max labels of the horizontal and vertical sliders. Replaced it with styling in the `slider.scss`. [(as mentioned here)](https://github.com/iTwin/iTwinUI/issues/753#issuecomment-1231652732)
- Fixed vertical slider's max and min labels misallginment as mentioned in #753 
- Also, improved the alignment of the min and the max labels in the horizontal slider (as they were similar to what was done to fix the vertical slider)


rohan-kadkol (commented):
I just wanted to confirm, is it okay for the `iui-slider-rail` of the vertical slider with the icons (red) (`height = 436px`) to be slightly longer than the `iui-slider-rail` of the vertical slider with the default numbers (green) (`height = 428px`)?

image

I tried solving this by adding a fixed height (`height: 24px`) to the `iui-slider-min` and `-max` for only the vertical slider, but I didn't know if that was okay since if users have a taller icon, the rail will not get shorter to accomodate the larger icons

image

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -215,8 +219,9 @@ $tick-height: $iui-baseline; // 12px
 
   .iui-slider-min,
   .iui-slider-max {
-    margin-left: $iui-xxs;
+    margin-left: calc($iui-xxs + $iui-sm);
     text-align: center;
+    transform: translate(-50%, 0);
```
rohan-kadkol, line 224: Did this to make the alignment relative to the mid point of the label

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -215,8 +219,9 @@ $tick-height: $iui-baseline; // 12px
 
   .iui-slider-min,
   .iui-slider-max {
-    margin-left: $iui-xxs;
+    margin-left: calc($iui-xxs + $iui-sm);
```
rohan-kadkol, line 222: Combined the `margin-left` of the inner svg tag into line 222 itself. This was the only way to make it work.

What I wanted to do (to make it similar to the hierarchy structure of the horizontal slider) (lines 220-225):

```scss
.iui-slider-min,
  .iui-slider-max {
    margin-left: $iui-xxs;
    text-align: center;
    transform: translate(-50%, 0);

    svg {
      margin: $iui-sm;
    }
  }
```

What I ended up having to do to make it work:

```scss
.iui-slider-min,
  .iui-slider-max {
    margin-left: calc($iui-xxs + $iui-sm);
    text-align: center;
    transform: translate(-50%, 0);
  }
```

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -20,7 +21,6 @@ $tick-height: $iui-baseline; // 12px
 
   svg {
```
gretanausedaite, line 22: Noticed that icons are black. Previously (currently in main) `iui-icon-default` had colors mixin included. I think we need to include it here to keep icons default color.
`@include iui-icon-color;`

gretanausedaite:
> I just wanted to confirm, is it okay for the `iui-slider-rail` of the vertical slider with the icons (red) (`height = 436px`) to be slightly longer than the `iui-slider-rail` of the vertical slider with the default numbers (green) (`height = 428px`)?
> 
No, it's not okay. Let's keep our components consistent.

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: Thanks, yes, I thought of leaving it out since Mayank is already handling this in his PR #718 ([Comment](https://github.com/iTwin/iTwinUI/pull/718#discussion_r944672596)). To avoid merging confusion later, maybe I should leave it as it is?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -215,8 +223,11 @@ $tick-height: $iui-baseline; // 12px
 
   .iui-slider-min,
   .iui-slider-max {
-    margin-left: $iui-xxs;
+    margin-left: calc($iui-xxs + $iui-sm);
     text-align: center;
+    transform: translate(-50%, 0);
+    height: $iui-line-height;
```
rohan-kadkol, line 229: Is it okay to use `$iui-line-height` here? Or is it better to use some other hard coded value here?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -158,6 +160,12 @@ $tick-height: $iui-baseline; // 12px
   .iui-slider-min,
   .iui-slider-max {
     margin-top: $iui-xxs;
+    width: $iui-line-height;
```
rohan-kadkol, line 163: Is it okay to use `$iui-line-height` here (since it is the width property)? Should I create a new variable just for this (line 163) and line 229? Or is it better to use some other hard coded value here?

rohan-kadkol:
> > I just wanted to confirm, is it okay for the `iui-slider-rail` of the vertical slider with the icons (red) (`height = 436px`) to be slightly longer than the `iui-slider-rail` of the vertical slider with the default numbers (green) (`height = 428px`)?
> 
> No, it's not okay. Let's keep our components consistent.

I have currently given a fixed `height` to the min and max labels (`.iui-slider-min`/`-max`) of the vertical slider with the same value as the line height of the text in the min and max labels of the horizontal slider (`24px`). ([Comment](https://github.com/iTwin/iTwinUI/pull/754#discussion_r961882180)). This makes the `.iui-slider-rail` height (vertical slider) and width (horizontal) the same for sliders with icons or text as labels

image

packages/itwinui-css/src/slider/slider.scss
mayank99: `$iui-line-height` doesn't exist anymore. I'm currently rebasing the main branch onto `v1` so I would say wait a bit and then update your branch to see what variables are available

PR #757, opened 2022-09-05
fix(style): Adjust components height

gretanausedaite:
Suggesting to make these adjustments to components height calculation to make them more similar to previous size:

| Component  | main size  | v1.0 size  | this PR size  |
|---|---|---|---|
| colorpicker - color field height | 209px | 216px (var(--iui-size-s) * 18) | 204px (var(--iui-size-s) * 17) |   
| colorpicker - max height | 495px | 540px (var(--iui-size-s) * 45) | 504px (var(--iui-size-s) * 42) |   
| radio-tile - sublabel line height | 15px | 16.5px (--iui-size-s) * 1.375)  | 15px (var(--iui-size-s) * 1.25)  |   
| tile - thumbnail height | 154px | 168px (var(--iui-size-s) * 14) | 156px (var(--iui-size-s) * 13) |


gretanausedaite:
@mayank99 Maybe you'll have some comments too.

elephantcatdog (approved):
It looks good to me! I would say that the colorpicker max height would be closer to the original value (495px) if it were multiplied by 41 (492px) rather than 42 (504px), but I totally understand if it's better to be bigger than smaller.

gretanausedaite:
Changed target branch, made new PR #760

PR #762, opened 2022-09-09
feat: All themes meet WCAG contrast score

### LINKED ISSUE #771: Variable renames
- Hover variable is sometimes not used for hover, rename?
- Remove `color` prefix?
- Look through [discussions](https://github.com/iTwin/iTwinUI/pull/718) in this PR for possibly more renames.

Variables MUST be stable after 1.0.

FlyersPh9:
**Main changes:**
- Base themes meet WCAG AA score of 4.5.
- High contrast themes meet WCAG AAA score of 7.1.
- User friendly variables, closes #771.

**Other changes:**
- Background colors get lighter the closer the element is to the viewer.  This means the darker gray color used on table header, unused space in side navigation is now lighter.
- Added tests for notification marker within buttons to test for white notification marker.
- `primary` is renamed to `accent` which is useful for custom themes.  `informational` is introduced to always display as blue despite any custom branding theme.
- Alert has new visual styling.
- Toast has new visual styling.
- `@mixin iui-surface` defaults changed, `@mixin iui-surface-component` introduced (altho without customization at the moment, to be addressed by #798).
- Disabled label uses muted text.

**Other notes:**
- Does not address contrast scores on badge or avatar.
- `:focus` issue on multiselect tags, #793.


FlyersPh9:
Comment mostly for myself as a reminder:  anchor links should underline when in high contrast mode & they are not currently.

packages/itwinui-css/backstop/tests/color.html
```diff
@@ -17,11 +17,6 @@
 
     <script type="module">
       import "./assets/theme.js";
-      import "@itwin/itwinui-icons-elements/info-circular";
-      import "@itwin/itwinui-icons-elements/status-success";
-      import "@itwin/itwinui-icons-elements/status-warning";
-      import "@itwin/itwinui-icons-elements/status-error";
-      import "@itwin/itwinui-icons-elements/close-small";
```
FlyersPh9, line 24: None of these were being used, must have been left over from a copy / paste.

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -74,7 +74,7 @@
     @include iui-anchor-status($statusColor);
   }
 
-  #{$rootSelector}-button {
+  #{$rootSelector}-button:focus-visible {
```
FlyersPh9, line 77: I don't like that `:focus-visible` is needed here but not needed in [`/toast/categories.scss` line 22](https://github.com/iTwin/iTwinUI/blob/f384be3406512a11b486c583aca14bb427779138/packages/itwinui-css/src/toast/categories.scss#L23).

packages/itwinui-css/src/checkbox/checkbox.scss
```diff
@@ -16,8 +16,8 @@
     gap: var(--iui-size-xs);
 
     &.iui-loading {
-      cursor: wait;
-      color: var(--iui-color-foreground-4);
+      cursor: progress;
+      font-style: italic;
```
FlyersPh9, line 20: Was focusing on changing colors so we pass contrast scores.  Decided to remove the muted text color here & apply italics to align with [tile's loading state](https://github.com/iTwin/iTwinUI/blob/f384be3406512a11b486c583aca14bb427779138/packages/itwinui-css/src/tile/tile.scss#L109).  Also aligned the `cursor`.  Maybe should align the class name as well.

packages/itwinui-css/src/checkbox/checkbox.scss
```diff
@@ -112,7 +112,7 @@
 
   &:disabled {
     --_iui-checkbox-svg-color: var(--iui-color-foreground-5);
-    --_iui-checkbox-border-color: var(--iui-color-background-border);
+    --_iui-checkbox-border-color: var(--iui-color-background-5);
```
FlyersPh9, line 115: Changed so that the disabled border color isn't more prominent than the default state in high contrast mode.

**Old, using background border:**
Screen Shot 2022-09-15 at 2 42 07 PM

**New, using background 5:**
Screen Shot 2022-09-15 at 2 43 31 PM

packages/itwinui-css/src/progress-indicator/linear.scss
```diff
@@ -11,7 +11,7 @@
     width: 100%;
     height: $height;
     overflow: hidden;
-    background-color: var(--iui-color-background-border);
+    background-color: var(--iui-color-background-5);
```
FlyersPh9, line 14: Changed so the filled track is more visible in high contrast mode.

**Old, using background border:**
Screen Shot 2022-09-15 at 2 45 32 PM
Screen Shot 2022-09-15 at 2 45 39 PM

**New, using background 5:**
Screen Shot 2022-09-15 at 2 45 06 PM
Screen Shot 2022-09-15 at 2 45 13 PM

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -40,7 +40,7 @@ $tick-height: var(--iui-size-s); // 12px
 
 @mixin iui-slider-rail {
   position: absolute;
-  background-color: var(--iui-color-background-border);
```
FlyersPh9, line 43: Changed so the fill stands out stronger in high contrast modes.

**Old, using background border:**
Screen Shot 2022-09-15 at 2 49 49 PM

**New, using background 5:**
Screen Shot 2022-09-15 at 2 50 19 PM

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -40,7 +40,7 @@ $tick-height: var(--iui-size-s); // 12px
...
+  background-color: var(--iui-color-background-5);
```
FlyersPh9, line 43: Changed so the fill stands out stronger in high contrast modes.

**Old, using background border:**
Screen Shot 2022-09-15 at 2 49 49 PM

**New, using background 5:**
Screen Shot 2022-09-15 at 2 50 19 PM

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -184,15 +185,19 @@
     display: flex;
     align-items: center;
     margin-top: auto;
-    color: var(--iui-color-foreground-4);
 
     > svg,
     .iui-tile-metadata-icon {
       @include iui-icon-style('m');
-      @include iui-icon-color;
+      fill: currentColor;
```
FlyersPh9, line 192: Changed so the icon can change fill between default state & disabled state.  Could also apply the color directly instead of using `currentColor`.

packages/itwinui-css/src/toast/categories.scss
```diff
@@ -18,4 +18,8 @@
   > .iui-toast-anchor {
     @include iui-anchor-status($category);
   }
+
+  > .iui-button {
```
FlyersPh9, line 22: Probably should be targeting a class like `.iui-toast-button` instead of `.iui-button`.

packages/itwinui-css/src/checkbox/checkbox.scss
FlyersPh9: Reverted this change, introduced this change in #765.

mayank99 (commented):
I like the contrast ratio improvements but I think the default dark mode visuals can be improved.

Just from an overall "look and feel", something feels wrong. I think the background colors feel a bit too unsaturated? I can't quite tell what exactly it is (or maybe it is a combination of things). A lot of this _is_ subjective, but I hope you can trust my overall judgement from years of using dark interfaces. And you can look at examples of dark themes on dribbble, they all seem to have a strong tint (often blue but usually derived from the accent color) rather than being low-saturation grays.

I also think we might need a color that sits between background-1 and background-2, because we want to be able to indicate multiple levels of elevation above the background body. This is most obvious in areas like table, where the table header seemingly sits "below" the page itself?

Also a few specific places I noticed that just look "off":
- **Alert**: The foreground color on links looks too saturated, and the background color on the whole alert looks too dark. It almost feels like a colored filter has been applied on a default background, rather than an intentionally handpicked color.
- **Input**: Disabled inputs don't have enough contrast with the background, making them hard to see.
- **Menu**: Selected state has the same problem as alert, where the background looks too dark.
- **Table**: The table headers still look messed up to me. There is too much of a color change when hovering a header cell.

packages/itwinui-css/src/alert/alert.scss
mayank99: You can get rid of `:focus-visible` if you move things around in `alert/classes.scss`. The alert-button and categories have the same specificity so categories should come after the button if you want them to take preference.

rohan-kadkol:
@FlyersPh9 In CSS PR #796, I think we accidentally did not approve the test images changes from that. Would it be possible to approve those images in this PR? I think @mayank99 suggested to approve them here since this PR is huge and already has a lot of image changes.

FlyersPh9:
> @FlyersPh9 In CSS PR #796, I think we accidentally did not approve the test images changes from that. Would it be possible to approve those images in this PR? I think @mayank99 suggested to approve them here since this PR is huge and already has a lot of image changes.

Yup, I just pulled the updates from the v1 branch, ran tests, & approved the new images.

packages/itwinui-css/src/table/table.scss
```diff
@@ -320,17 +325,19 @@
...
 
   &.iui-zebra-striping .iui-table-row:nth-child(even):not([aria-selected='true']) .iui-table-cell {
-    @include pseudo-content {
-      background-color: hsl(var(--iui-color-foreground-1-hsl) / 0.02);
-    }
+    background-color: var(--iui-color-background-zebra);
+  }
+
+  &.iui-zebra-striping .iui-table-row:nth-child(even):not([aria-selected='true']):hover .iui-table-cell {
+    background-color: var(--iui-color-background-hover);
```
mayank99, line 340: this is already set in the base row hover styles. i think you just need to reduce the specificity of the zebra striping selector above using `:where`

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -18,27 +18,31 @@
...
+    color: var(--iui-color-text);
     transition: background-color var(--iui-duration-1) ease-out;
 
     &.iui-active {
-      color: var(--iui-color-foreground-primary);
+      color: var(--iui-color-text-primary);
 
       .iui-tab-icon {
-        fill: var(--iui-color-foreground-primary);
+        fill: var(--iui-color-icon-primary);
```
mayank99, line 28: This icon is standing out quite a bit, compared to the text and stripe. Is this the intended look?

packages/itwinui-css/src/text/skeleton.scss
```diff
@@ -9,11 +9,11 @@
   cursor: progress;
   background: linear-gradient(
     292deg,
-    var(--iui-color-foreground-4),
-    var(--iui-color-foreground-4),
-    var(--iui-color-foreground-5),
-    var(--iui-color-foreground-4),
-    var(--iui-color-foreground-4)
+    var(--iui-color-text-muted),
```
mayank99, line 12: would `icon` colors look better here? these look too dark

i'm unsure if these need to meet the text contrast requirements or not

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -189,10 +191,15 @@
...
     .iui-tile-metadata-icon {
       @include iui-icon-style('m');
-      @include iui-icon-color;
+      fill: var(--iui-color-icon-muted);
       margin-right: var(--iui-size-xs);
     }
 
+    .iui-tag-container,
+    .iui-tag-basic {
+      color: inherit;
```
mayank99, line 200: I think these should be moved under disabled styling

packages/itwinui-css/src/utils/mixins.scss
```diff
@@ -12,7 +12,7 @@
   }
 }
 
-@mixin iui-focus($color: var(--iui-color-foreground-primary), $offset: -1px, $thickness: 1px) {
+@mixin iui-focus($color: var(--iui-color-text-primary), $offset: -1px, $thickness: 1px) {
```
mayank99, line 15: I think this should use `border-primary`. They refer to same color underneath but border makes more sense

packages/itwinui-variables/src/colors.scss
```diff
@@ -19,34 +19,7 @@
 }
 
 /// Colors that don't change between light and dark themes
-/// but do change between default and high contrast.
-/// @arg $contrast - default or high
 @mixin unthemed($contrast: default) {
```
mayank99, line 22: the `$contrast` arg can be removed.

packages/itwinui-variables/src/sizes.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 @mixin border-radii {
-  --iui-border-radius-1: 4px;
+  --iui-border-radius-1: var(--iui-size-2xs);
```
mayank99, line 18: I think this was intentionally left as 4px because it doesn't need to scale with base font size.

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -47,23 +82,73 @@ $opacities: (
...
+  --iui-color-border-disabled: #{hex($bg-9)};
+  --iui-color-border-primary: #{hex($fg-primary-1)};
+  --iui-color-border-positive: #{hex($fg-positive-1)};
+  --iui-color-border-warning: #{hex($fg-warning-1)};
+  --iui-color-border-negative: #{hex($fg-negative-1)};
+  // #endregion
+  // #region Icon variables
+  --iui-color-icon: #{hex($fg-3)};
+  --iui-color-icon-hover: #{hex($fg-2)};
+  --iui-color-icon-muted: #{hex($fg-7)};
```
mayank99, line 124: Any chance we can pass 3:1 for muted icon color?

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -47,23 +82,73 @@ $opacities: (
...
+  --iui-color-icon-positive: #{hex($fg-positive-2)};
+  --iui-color-icon-warning: #{hex($fg-warning-2)};
+  --iui-color-icon-negative: #{hex($fg-negative-2)};
+  // #endregion
+  // #region Text variables
+  --iui-color-text: #{hex($fg-2)};
+  --iui-color-text-hover: #{hex($fg-1)};
+  --iui-color-text-muted: #{hex($fg-5)};
+  --iui-color-text-disabled: #{hex($fg-7)};
+  --iui-color-text-primary: #{hex($fg-primary-1)};
```
mayank99, line 137: I'm wondering if we should introduce a new set of aliases for `accent`, to separate it from status colors. Or we could rename `primary` to `accent`. I think it would be nice for customizability.

mayank99 (commented):
I tried to go through all components. I'm quite liking the new changes.

0. The "pool of colors" that live only in the code (bg0 through bg10, etc) are quite nice as a low layer. The fact that we were able to easily "shift" the dark and dark-hc themes in lightness so quickly makes me feel good about the maintainability.
1. Just looking at the diff in all the files, I'm convinced that this is the right approach! The variable names make perfect sense in relation to how they are used and it's easier to explain/document for our users too 👍
2. In addition to passing contrast requirements, all four themes look really nice and more consistent now. The new background colors and status colors feel more intentionally picked and we are following the conventions of elevation (lighter color = more elevated). I also like that we are now using subtle borders in more places rather than relying only on background colors for separation. Our themes feel polished!

Just a few minor comments and then should be good to go.

packages/itwinui-css/src/toast/categories.scss
mayank99: i think you can just target `button` tbh

elephantcatdog (commented):
Things that Jon and I discussed on a call:

1. I think there should be a warning for developers about how changing the color options can impact the contrast ratings.
2. Related to above, Jon questioned whether the first variable was needed as it's not being used to define the colors anymore.
3. The color that's used for zebra striping doesn't go with the rest of the color series very well as it's an odd value. An alternative would be to just move the definition of that color further down the page to where the zebra-striping-color is defined. Another way would be to define the zebra-striping-color from a related color - Jon mentioned there might be a way to do this in the future. You could also move the odd color out of the color series and call it 'bg-special' or 'bg-1b' or something so that it's not part of the regular series.
4. Jon and I were looking at the alerts vs the toasts - Jon redesigned the toast to avoid color blocking but also wants to differentiate it from the alerts. First question - why do we have separate toasts and alerts when they're very similar? Second question - do they need to look significantly different from each other?
5. Related the the color blocking above - this was removed because the orange is now brown, which was done to have similar luminosity values. How important is luminosity really? I think it's more important for our orange warning color to be a defined warning color. I don't think 'brown' says 'warning' to most people as we've been habituated to see orange/yellow/gold as the 'warning' color.

packages/itwinui-css/src/toast/categories.scss
FlyersPh9: Ok, I will leave it as is then.

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -6,12 +6,16 @@
...
   border-radius: var(--iui-border-radius-1);
   display: flex;
   align-items: center;
-  border: 1px solid var(--iui-color-background-5);
-  color: var(--iui-color-foreground-2);
-  background-color: var(--iui-color-background-1);
+  color: var(--iui-color-text);
+  background-color: var(--iui-color-background);
+  border: 1px solid var(--_iui-alert-border-color);
+  box-shadow: inset var(--iui-size-2xs) 0 0 var(--_iui-alert-border-color);
```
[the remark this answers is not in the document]
FlyersPh9: Totally agree that we should focus on forced-colors and ensure all components are working as expected.  This can happen after our major release.

packages/itwinui-css/src/carousel/carousel.scss
```diff
@@ -39,7 +39,7 @@
 
     button[data-pressed='true'] {
       outline-offset: -1px;
-      outline: 1px solid var(--iui-color-foreground-primary);
+      outline: 1px solid var(--iui-color-icon-primary);
```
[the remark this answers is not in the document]
FlyersPh9: Corrected.

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -7,9 +7,9 @@
 $iui-color-picker-content-width: calc(var(--iui-size-l) * 9 + var(--iui-size-xs) * 8); // (9 swatches per row) + (gaps)
 $iui-color-picker-max-width: calc(var(--iui-size-l) * 10 + var(--iui-size-xs) * 9); // (10 swatches per row) + (gaps)
 $iui-color-swatch-border-radius: 5px;
-$iui-inset-box-shadow: inset 0 0 0 1px var(--iui-color-foreground-5);
-$iui-active-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-foreground-primary);
-$iui-hover-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-foreground-4); // 0 0 1px calc(var(--iui-size-3xs) + 1px) var(--iui-color-background-1-hover);
+$iui-inset-box-shadow: inset 0 0 0 1px var(--iui-color-border-foreground);
+$iui-active-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-text-primary);
```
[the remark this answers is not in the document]
FlyersPh9: Corrected the variable used & increased the offset from 1px to 2px.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -169,6 +170,7 @@ $iui-dialog-width: 380px;
   padding: 0 var(--iui-size-m);
   overflow-y: auto;
   overflow-y: overlay;
+  color: var(--iui-color-text);
```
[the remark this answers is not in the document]
FlyersPh9: I was testing to make sure the new variables were being applied correctly and thought it made sense to define the color explicitly for each component based off of #683, but maybe that is not necessary for all components.

Basically I was taking `var(--iui-color-text)`, setting it to red, and making sure icons / borders weren't turning red.

packages/itwinui-variables/src/colors.scss
```diff
@@ -19,34 +19,7 @@
...
-      background-primary-hover: hsl(203 100% 34%),
-      background-positive-hover: hsl(95 72% 27%),
-    );
-  }
-
-  @each $color, $value in $background-statuses {
-    --iui-color-#{$color}: #{hex($value)};
-  }
+  --iui-color-text-accessory: #{hex(hsl(0 0% 100%))};
+  --iui-color-icon-accessory: #{hex(hsl(0 0% 100%))};
```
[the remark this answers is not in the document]
FlyersPh9: Agreed that this "accessory" name doesn't make much sense.  I am open to suggestions.

Also supportive of only offering 1 variable that spans text, icon, border, etc.

packages/itwinui-css/src/toast/categories.scss
mayank99: just to be clear, i'm suggesting `button` instead of `.iui-button`

packages/itwinui-variables/src/colors.scss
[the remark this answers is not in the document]
mayank99: `iui-white`?

packages/itwinui-css/src/dialog/dialog.scss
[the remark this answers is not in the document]
mayank99: this makes it a lot harder to change the text color globally. can we revert it if the components work without it?

packages/itwinui-css/src/header/header-buttons.scss
```diff
@@ -81,14 +81,14 @@
 
   // Hover state
   &:hover {
-    background-color: var(--iui-color-background-transparent-hover);
+    background-color: var(--iui-color-background-hover);
```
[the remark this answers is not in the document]
FlyersPh9: Good catch, I missed this one.  Corrected on lines 53-55 within `header.scss`.

packages/itwinui-css/src/input/input.scss
```diff
@@ -26,25 +26,25 @@
   @include iui-input-size;
 
   &::placeholder {
-    --_iui-input-text-color: var(--iui-color-foreground-5);
+    --_iui-input-text-color: var(--iui-color-text-disabled);
```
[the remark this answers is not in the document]
FlyersPh9: I did originally try this with `muted` but wasn't happy with the visual outcome.  This is one of the few instances of questionably using a variable (applying disabled on something that isn't technically disabled).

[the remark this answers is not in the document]
FlyersPh9: We could introduce `var(--iui-color-text-placeholder)` or we can leave it as is.

packages/itwinui-css/src/input/input.scss
```diff
@@ -54,8 +54,8 @@
 }
 
 @mixin iui-input-autofill {
```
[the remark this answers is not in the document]
FlyersPh9: I've removed `@mixin iui-input-autofill`.

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -111,28 +111,28 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
     &,
     &:hover,
     &[disabled] {
-      border-color: var(--iui-color-background-border);
+      border-color: var(--iui-color-border);
```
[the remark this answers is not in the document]
FlyersPh9: I would be ok with increasing `border-right` from 1px to 2px to align with header's `border-bottom`.  What do you think about removing `border-top` & `border-bottom` from all buttons _except_ for the expander?
Screenshot 2022-11-03 at 11 47 46 AM Screenshot 2022-11-03 at 11 47 59 AM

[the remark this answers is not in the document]
FlyersPh9: We would need to update the dividing line thickness in the header to match.
Screenshot 2022-11-03 at 11 52 22 AM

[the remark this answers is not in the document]
rohan-kadkol: > What do you think about removing `border-top` & `border-bottom` from all buttons _except_ for the expander?

I'm not too familiar with this PR since I'm not a reviewer, but just wanted to mention that I think the borderless buttons add a more modern feel imo.

packages/itwinui-css/src/alert/alert.scss
FlyersPh9: Change made & `:focus-visible` removed, good suggestion.

packages/itwinui-css/src/toast/categories.scss
FlyersPh9: Ah, I did not pickup on that, change made to target `button`.

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: Yikes, yes absolutely.  Color assigned changed and tested to confirm passing 3.1 now.

packages/itwinui-variables/src/sizes.scss
FlyersPh9: Reverted.

packages/itwinui-variables/src/colors.scss
FlyersPh9: Removed.

packages/itwinui-css/src/utils/mixins.scss
FlyersPh9: Agreed, change made.

packages/itwinui-css/src/text/skeleton.scss
FlyersPh9: I think your suggestion makes sense.  Text requires a higher score than icons because text is so thin, so using the icon variables and meeting 3.1 should be acceptable.  Change made.

packages/itwinui-css/src/table/table.scss
```diff
@@ -113,20 +116,21 @@
...
     > .iui-table-reorder-bar {
-      background-color: var(--iui-color-foreground-primary);
+      background-color: var(--iui-color-border-primary);
     }
   }
 
   // Sort icon
   .iui-table-sort {
     opacity: 0;
+    fill: var(--iui-color-icon-muted);
```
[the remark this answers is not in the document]
FlyersPh9: Removed `.iui-button-icon` and corrected the CSS.

packages/itwinui-css/src/table/table.scss
```diff
@@ -404,27 +411,30 @@
   &[contenteditable] {
     outline-offset: -1px;
 
-    &:focus,
+    &:focus {
+      outline: 1px solid var(--iui-color-text-primary);
```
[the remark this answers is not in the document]
FlyersPh9: Fixed.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -2,19 +2,19 @@
...
 
   background-color: var(--iui-surface-background-color);
+  border: 1px solid var(--iui-surface-border-color);
   border-radius: var(--iui-surface-border-radius);
   box-shadow: var(--iui-surface-elevation);
-  color: var(--iui-color-foreground-2);
-  @media (forced-colors: active) {
-    border: 1px solid transparent;
-  }
+  color: var(--iui-color-text);
```
FlyersPh9, line 19: @mayank99, should `color` be removed from `surface.scss` for the same reason mentioned here? https://github.com/iTwin/iTwinUI/pull/762/#discussion_r1013071131

packages/itwinui-css/src/dialog/dialog.scss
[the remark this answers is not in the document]
FlyersPh9: Reverted these additions.  I think I caught all of them, please comment on any I may have missed.

packages/itwinui-css/src/tabs/tabs.scss
FlyersPh9: Note to self:  use `currentColor` when icon is directly next to active text like this.

packages/itwinui-css/src/surface/surface.scss
mayank99: i think it's fine to keep color in surface because it's intended to be a starting point for our users

gretanausedaite (commented):
I love new colours!
Do we need to update iModel Loader green to use our positive colour or we are using that green because of iModel branding?

packages/itwinui-css/src/table/table.scss
FlyersPh9: We tried to simplify this but decided to leave it as is, table needs a more thorough refactor.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Done.

packages/itwinui-css/src/side-navigation/side-navigation.scss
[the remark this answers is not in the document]
FlyersPh9: I've increased `border-right` thickness to 2px to align with the header's `border-bottom`.  Everything else we'll revisit at a later date.

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -47,23 +82,73 @@ $opacities: (
...
+  --iui-color-text-negative: #{hex($fg-negative-1)};
+  --iui-color-text-negative-hover: #{hex($fg-negative-0)};
+  // #endregion
+  // #region HSL variables for transparency
+  --iui-color-background-hsl: #{get-hsl($bg-6)};
+  --iui-color-foreground-hsl: #{get-hsl($fg-1)};
+  --iui-color-foreground-primary-hsl: #{get-hsl($fg-primary-2)};
+  --iui-color-foreground-positive-hsl: #{get-hsl($fg-positive-2)};
+  --iui-color-foreground-warning-hsl: #{get-hsl($fg-warning-2)};
+  --iui-color-foreground-negative-hsl: #{get-hsl($fg-negative-2)};
```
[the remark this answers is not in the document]
FlyersPh9: Renamed these to `--iui-color-positive-hsl`.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
```diff
@@ -14,8 +14,8 @@ $iui-notification-marker-border-width: var(--iui-size-3xs);
     top: calc(var(--iui-size-s) * -0.5);
     right: calc(var(--iui-size-s) * -0.5);
     border-radius: 100%;
-    background-color: var(--iui-color-foreground-#{$status});
-    border: $iui-notification-marker-border-width solid var(--iui-color-background-1);
+    background-color: var(--iui-color-icon-#{$status}); // TODO this won't work with white
```
FlyersPh9, line 17: @mayank99 this won't work with `--iui-color-white`.

packages/itwinui-css/backstop/tests/iModel-loader.html
```diff
@@ -49,7 +53,7 @@
           <g>
             <polygon
               data-id="outer3"
-              fill="#62bb47"
+              fill="var(--iui-color-icon-positive)"
```
FlyersPh9, line 56: @gretanausedaite I've updated this to use CSS vars.  I recall we had some color assignment issues in the past, is there any issue using CSS vars here instead of HEX?

gretanausedaite: There is no issue in this example. We have this loader in our illustrations package (and we recommend developers use it from there, if I remember correctly). The problem in illustrations package - we can't use CSS vars without variables package. So we should have some kind of hardcoded value there.

We can leave this issue for after new releases. As example code, this looks good to me.

elephantcatdog (commented):
In `badge.html` there are four colors for 'information', 'positive', 'negative', and 'warning' defined on lines 122-148. Should these be using a variable? `badge.scss` has some variables defined for these colors it looks like. Also, there are 'soft-backgrounds' defined in `colors.scss` that aren't used for the badge colors.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
mayank99: what does this have to do with `iui-color-white`?

packages/itwinui-css/src/tabs/tabs.scss
mayank99: what about other places like active tree nodes?

i feel like we could change the color value of `--iui-color-icon-accent` to be a little more consistent with `--iui-color-text-accent`, at least in dark theme where it stands out the most

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
FlyersPh9: Notification marker's `classes.scss` originally supported `accessory` which is now replaced with `white`:
```scss
@each $status in positive, warning, negative, accessory {
  &[data-iui-variant='#{$status}'] {
    @include iui-notification-marker-status($status);
  }
}
```

The color of the notification dot is set here:
```scss
--_iui-notification-marker-color: var(--iui-color-icon-#{$status});
```

But we don't have a `--iui-color-icon-white`, only `--iui-color-white`.  I think I can come up with a fix using `@if`.

packages/itwinui-css/backstop/tests/iModel-loader.html
FlyersPh9: I just confirmed that something like this works:
```html
fill="var(--iui-color-icon-positive, hsl(118 74% 35%))"
```
If iTwinUI-Variables is present, it uses the CSS variable value.  If iTwinUI-Variables i not present, it uses the hardcoded value to the right (which is the same as light theme's `--iui-color-icon-positive`).

FlyersPh9:
> In `badge.html` there are four colors for 'information', 'positive', 'negative', and 'warning' defined on lines 122-148. Should these be using a variable? `badge.scss` has some variables defined for these colors it looks like. Also, there are 'soft-backgrounds' defined in `colors.scss` that aren't used for the badge colors.

@elephantcatdog sorry, this is a mistake on my part.  The original PR description mentioned that we were ignoring badge & avatar colors but when I updated the PR description I failed to mention that again.  I will address avatar & badge colors in a separate PR since they are using unique colors that aren't used anywhere else.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -2,19 +2,19 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 
 @mixin iui-surface(
-  $background-color: var(--iui-color-background-1),
+  $background-color: var(--iui-color-background),
+  $border-color: var(--iui-color-border-subtle),
```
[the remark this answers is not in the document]
FlyersPh9: Does `@mixin iui-surface` need to match the React surface component?  I think `@mixin iui-surface` should provide the best defaults for use within iTwinUI-CSS so we don't need to specify elevation in most cases, even if that is different from the React Surface component's defaults.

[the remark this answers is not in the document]
mayank99: It doesn't need to match, but I'm just saying that these changes *do* affect Surface, meaning we need to override them in react.

Maybe there needs to be a different mixin (or a predetermined set of default values) for the Surface component.

The distinction between "internal surface mixin" and "public Surface component" is useful, since the latter is meant to be used more like "cards" where the users can put other components inside.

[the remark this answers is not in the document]
FlyersPh9: > The distinction between "internal surface mixin" and "public Surface component" is useful, since the latter is meant to be used more like "cards" where the users can put other components inside.

This makes sense to me.

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -4,66 +4,159 @@
...
+
+// #region Background colors
+// Darkest @ top, lightest @ bottom
+$bg-0: hsl($bg-hue $bg-saturation 0%);
+$bg-1: hsl($bg-hue $bg-saturation ($bg-lightness-gap));
+$bg-2: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 2));
+$bg-3: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 3));
+$bg-4: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 4));
+$bg-5: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 5));
+$bg-6: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 5.5));
```
elephantcatdog, line 22: Have you thought about moving this color to be outside of the series? Maybe call this one `$bg-5b` instead of `$bg-6`?

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -4,38 +4,73 @@
...
+// Darkest @ top, lightest @ bottom
+$bg-0: hsl($bg-hue $bg-saturation 0%);
+$bg-1: hsl($bg-hue $bg-saturation ($bg-lightness-gap));
+$bg-2: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 2));
+$bg-3: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 3));
+$bg-4: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 4));
+$bg-5: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 5));
+$bg-6: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 6));
+$bg-7: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 7));
+$bg-8: hsl($bg-hue $bg-saturation ($bg-lightness-gap * 7.5));
```
elephantcatdog, line 24: Same as other comment - have you thought about moving this color to be outside of the series? Maybe call it `$bg-7b` instead of `&bg-8`?

packages/itwinui-variables/src/themes/light-hc.scss
```diff
@@ -4,66 +4,159 @@
...
+$bg-hue: 203;
+$bg-saturation: 5%; // Less saturated than base theme.
+$bg-lightness-gap: 4%;
+$fg-lightness-gap: 6%;
+// #endregion
+
+// #region Background colors
+// Lightest @ top, darkest @ bottom
+$bg-0: hsl($bg-hue $bg-saturation 100%);
+$bg-1: hsl($bg-hue $bg-saturation (100% - $bg-lightness-gap * 0.5));
```
elephantcatdog, line 17: Same as other comment - have you thought about moving this color to be outside of the series? Maybe call it `$bg-0b` instead of `$bg-1`?

packages/itwinui-variables/src/themes/light.scss
```diff
@@ -4,38 +4,73 @@
...
+$bg-hue: 203;
+$bg-saturation: 10%; // More saturated than high contrast theme.
+$bg-lightness-gap: 3%;
+$fg-lightness-gap: 7%;
+// #endregion
+
+// #region Background colors
+// Lightest @ top, darkest @ bottom
+$bg-0: hsl($bg-hue $bg-saturation 100%);
+$bg-1: hsl($bg-hue $bg-saturation (100% - $bg-lightness-gap * 0.5));
```
elephantcatdog, line 17: Same as other comment - have you thought about moving this color to be outside of the series? Maybe call it `$bg-0b` instead of `$bg-1`?

packages/itwinui-variables/src/themes/light.scss
```diff
@@ -4,38 +4,73 @@
...
-  foreground-positive: hsl(95 72% 37%),
-  foreground-warning: hsl(33 89% 51%),
-  foreground-negative: hsl(0 91% 43%),
-  foreground-primary-hover: hsl(203 100% 34%),
-  foreground-positive-hover: hsl(95 72% 27%),
-  foreground-warning-hover: hsl(33 89% 41%),
-  foreground-negative-hover: hsl(0 91% 33%)
-);
+// #region SASS variables
+$bg-hue: 203;
```
elephantcatdog, line 8: Same as other comment - did you decide not to remove this?

packages/itwinui-variables/src/themes/light.scss
```diff
@@ -4,38 +4,73 @@
...
+$bg-saturation: 10%; // More saturated than high contrast theme.
+$bg-lightness-gap: 3%;
+$fg-lightness-gap: 7%;
+// #endregion
```
elephantcatdog, line 12: Same as other comment - maybe add a warning about how changing the above values could negatively impact contrast scores?

packages/itwinui-variables/src/themes/light-hc.scss
```diff
@@ -4,66 +4,159 @@
...
-  foreground-positive: hsl(140 100% 16%),
-  foreground-warning: hsl(45 100% 19%),
-  foreground-negative: hsl(0 100% 28%),
-  foreground-primary-hover: hsl(203 100% 34%),
-  foreground-positive-hover: hsl(95 72% 27%),
-  foreground-warning-hover: hsl(33 89% 41%),
-  foreground-negative-hover: hsl(0 91% 33%)
-);
+// #region SASS variables
+$bg-hue: 203;
```
elephantcatdog, line 8: Same as other comment - did you decide not to remove this?

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -4,38 +4,73 @@
...
-  foreground-positive: hsl(96 45% 78%),
-  foreground-warning: hsl(34 87% 82%),
-  foreground-negative: hsl(0 69% 80%),
-  foreground-primary-hover: hsl(203 80% 90%),
-  foreground-positive-hover: hsl(96 44% 88%),
-  foreground-warning-hover: hsl(34 85% 92%),
-  foreground-negative-hover: hsl(0 69% 90%)
-);
+// #region SASS variables
+$bg-hue: 203;
```
elephantcatdog, line 8: Same as other comment - did you decide not to remove this?

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -4,66 +4,159 @@
...
-  foreground-positive: hsl(140 100% 72%),
-  foreground-warning: hsl(45 100% 75%),
-  foreground-negative: hsl(0 100% 88%),
-  foreground-primary-hover: hsl(203 100% 34%),
-  foreground-positive-hover: hsl(95 72% 27%),
-  foreground-warning-hover: hsl(33 89% 41%),
-  foreground-negative-hover: hsl(0 91% 33%)
-);
+// #region SASS variables
+$bg-hue: 203;
```
elephantcatdog, line 8: Did you decide not to remove this?

packages/itwinui-variables/src/themes/light-hc.scss
```diff
@@ -4,66 +4,159 @@
...
+$bg-saturation: 5%; // Less saturated than base theme.
+$bg-lightness-gap: 4%;
+$fg-lightness-gap: 6%;
+// #endregion
```
elephantcatdog, line 12: Same as other comment - maybe add a warning about how changing the above values could negatively impact contrast scores?

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -4,38 +4,73 @@
...
+$bg-saturation: 12%; // More saturated than high contrast theme.
+$bg-lightness-gap: 3.25%;
+$fg-lightness-gap: 7%;
+// #endregion
```
elephantcatdog, line 12: Same as other comment - maybe add a warning about how changing the above values could negatively impact contrast scores?

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -4,66 +4,159 @@
...
+$bg-saturation: 6%; // Less saturated than base theme.
+$bg-lightness-gap: 4.25%;
+$fg-lightness-gap: 5%;
+// #endregion
```
elephantcatdog, line 12: Maybe add a warning about how changing the above values could negatively impact contrast scores?

elephantcatdog:
I never saw that you addressed my longer comment, so I made comments on the code. ^

packages/itwinui-css/src/tabs/tabs.scss
FlyersPh9: `dark.scss` has new `$fg-*-2` values.  How I created the new values:

Take `$fg-*-1` & current `$fg-*-2` value, plug them into this [middle color finder web app](https://www.dannybrien.com/middle/), and use that new value for `$fg-*-2`.  Here is the luminosity test:
Screenshot 2022-11-07 at 1 41 30 PM

packages/itwinui-css/src/surface/surface.scss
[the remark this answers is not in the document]
FlyersPh9: Introduced `@mixin iui-surface-component` which is used by `.iui-surface`.  We have a separate `@mixin iui-surface` which has had `border-radius` changed to simplify the general usage of surface across components.

packages/itwinui-variables/src/themes/dark-hc.scss
FlyersPh9: Is now removed.

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: Is now removed.

packages/itwinui-variables/src/themes/light-hc.scss
FlyersPh9: Is now removed.

packages/itwinui-variables/src/themes/light.scss
FlyersPh9: Is now removed.

packages/itwinui-css/src/surface/surface.scss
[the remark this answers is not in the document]
FlyersPh9: Issue created #798.

mayank99 (approved):
LGTM

packages/itwinui-variables/src/themes/dark-hc.scss
```diff
@@ -151,8 +150,8 @@ $opacities: (
   --iui-color-text-negative-hover: #{hex($fg-negative-0)};
   // #endregion
   // #region HSL variables for transparency
-  --iui-color-background-hsl: #{get-hsl($bg-6)};
-  --iui-color-foreground-hsl: #{get-hsl($fg-1)};
+  --iui-color-background-hsl: #{get-hsl($bg-500)};
```
elephantcatdog, line 153: I think bg-6 should be converted to bg-550?

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -5,25 +5,25 @@
...
+$bg-200: hsl(203 $bg-saturation ($bg-lightness-gap * 2));
+$bg-300: hsl(203 $bg-saturation ($bg-lightness-gap * 3));
+$bg-400: hsl(203 $bg-saturation ($bg-lightness-gap * 4));
+$bg-500: hsl(203 $bg-saturation ($bg-lightness-gap * 5));
+$bg-600: hsl(203 $bg-saturation ($bg-lightness-gap * 6));
+$bg-700: hsl(203 $bg-saturation ($bg-lightness-gap * 7));
+$bg-750: hsl(203 $bg-saturation ($bg-lightness-gap * 7.5));
+$bg-800: hsl(203 $bg-saturation ($bg-lightness-gap * 8));
+$bg-900: hsl(203 $bg-saturation ($bg-lightness-gap * 9));
+$bg-1000: hsl(203 $bg-saturation ($bg-lightness-gap * 10));
```
elephantcatdog, line 26: You added an extra background in this file compared to the dark-hc.scss file

packages/itwinui-variables/src/themes/dark-hc.scss
FlyersPh9: `--iui-color-background-hsl` should match whatever is being used by `--iui-color-background`, so for `dark-hc.scss` `$bg-500` appears to be correct.

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -111,34 +111,34 @@ $opacities: (
   --iui-color-background-transparent-hover: #{hsl(0 0% 100% / 0.065)};
   // #endregion
   // #region Border variables
-  --iui-color-border-subtle: #{hex($bg-9)};
-  --iui-color-border: #{hex($bg-10)};
-  --iui-color-border-foreground: #{hex($fg-8)};
-  --iui-color-border-foreground-hover: #{hex($fg-6)};
-  --iui-color-border-disabled: #{hex($bg-9)};
+  --iui-color-border-subtle: #{hex($bg-900)};
```
elephantcatdog, line 114: bg-9 converts to bg-800

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -111,34 +111,34 @@ $opacities: (
...
+  --iui-color-border: #{hex($bg-1000)};
```
elephantcatdog, line 115: bg-10 converts to bg-900

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -111,34 +111,34 @@ $opacities: (
...
+  --iui-color-border-foreground: #{hex($fg-800)};
+  --iui-color-border-foreground-hover: #{hex($fg-600)};
+  --iui-color-border-disabled: #{hex($bg-900)};
```
elephantcatdog, line 118: bg-9 converts to bg-800

packages/itwinui-variables/src/themes/dark.scss
```diff
@@ -151,8 +151,8 @@ $opacities: (
   --iui-color-text-negative-hover: #{hex($fg-negative-0)};
   // #endregion
   // #region HSL variables for transparency
-  --iui-color-background-hsl: #{get-hsl($bg-6)};
-  --iui-color-foreground-hsl: #{get-hsl($fg-1)};
+  --iui-color-background-hsl: #{get-hsl($bg-700)};
```
elephantcatdog, line 154: bg-6 converts to bg-600

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: Yes, that was intentional.  I noticed that prior to this `--iui-color-background` & `--iui-color-border-subtle` were set to the same color.  `--iui-color-border` was already to set to 900 and we couldn't go any higher, so I introduced `1000`.

That is one benefit to this autogenerated 0-1000 system is we can introduce new #s whenever needed (like this case).

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: This is related to the explaination here https://github.com/iTwin/iTwinUI/pull/762#discussion_r1015991109

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: This is related to the explaination here https://github.com/iTwin/iTwinUI/pull/762#discussion_r1015991109

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: This is related to the explaination here https://github.com/iTwin/iTwinUI/pull/762#discussion_r1015991109

packages/itwinui-variables/src/themes/dark-hc.scss
elephantcatdog: previous mistake

packages/itwinui-variables/src/themes/dark.scss
elephantcatdog: previous mistake

packages/itwinui-variables/src/themes/dark.scss
elephantcatdog: previous mistake

packages/itwinui-variables/src/themes/dark.scss
elephantcatdog: previous mistake

packages/itwinui-variables/src/themes/dark.scss
elephantcatdog: previous mistake

packages/itwinui-variables/src/themes/light-hc.scss
```diff
@@ -111,34 +110,34 @@ $opacities: (
...
   // #endregion
   // #region Border variables
-  --iui-color-border-subtle: #{hex($bg-8)};
-  --iui-color-border: #{hex($bg-10)};
-  --iui-color-border-foreground: #{hex($fg-8)};
-  --iui-color-border-foreground-hover: #{hex($fg-6)};
-  --iui-color-border-disabled: #{hex($bg-8)};
+  --iui-color-border-subtle: #{hex($bg-700)};
+  --iui-color-border: #{hex($bg-900)};
+  --iui-color-border-foreground: #{hex($fg-700)};
```
elephantcatdog, line 115: fg-8 converts to fg-800

FlyersPh9: Good catch, correcting soon.

elephantcatdog (approved):
Looks good, assuming tests are good 👍

PR #763, opened 2022-09-13
feat(Table): Row selection count for paginator

adamhock:
Updated table demo to show the number of rows selected in a given paginated table. The element is removed if there is not any rows selected.


FlyersPh9 (approved):
Working as expected, good job!  Please wait for someone else to review the JS before merging.

mayank99:
the PR title should not say `feat` because there are no changes in the actual CSS

elephantcatdog:
> the PR title should not say feat because there are no changes in the actual CSS

@mayank99 What should it say then? "docs"?
image

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -4616,22 +4616,54 @@ <h2>Empty state</h2>
...
+        if (!paginatorCheckboxCount) {
+          const newPaginatorCheckboxCount = document.createElement('span');
+          newPaginatorCheckboxCount.classList.add('iui-table-paginator-selection-count');
+          newPaginatorCheckboxCount.textContent = paginatorCheckboxCountText;
+          leftPaginator.appendChild(newPaginatorCheckboxCount);
+        }
+        else {
+          paginatorCheckboxCount.textContent = paginatorCheckboxCountText;
+        }
+      };
```
elephantcatdog, line 4640: Very nitpicky, but I'd put an empty line after this function.

elephantcatdog (commented):
Would it be better to have "0 rows selected" versus the text disappearing? The text disappearing is very nice and clean, but I think showing "0 rows selected" gives more information to the user.

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -4616,22 +4616,54 @@ <h2>Empty state</h2>
...
+        const numBoxesChecked = checkboxes.filter((checkbox) => checkbox.checked).length;
+        const leftPaginator = table.querySelector('.iui-table-paginator .iui-left');
+        const paginatorCheckboxCount = leftPaginator.querySelector('.iui-table-paginator-selection-count');
+        if (!numBoxesChecked) {
+          if (paginatorCheckboxCount) {
+            leftPaginator.removeChild(paginatorCheckboxCount);
+          }
+          return;
+        }
+        const paginatorCheckboxCountText = numBoxesChecked === 1 ? '1 row selected' : `${numBoxesChecked} rows selected`;
```
elephantcatdog, line 4630: Would it be better to have "0 rows selected" versus the text disappearing? The text disappearing is very nice and clean, but I think showing "0 rows selected" gives more information to the user.

gretanausedaite: I think we should not show "0 rows selected". It gives 0 useful information and might add to visual clutter.

gretanausedaite (approved):

packages/itwinui-css/backstop/tests/table.html
elephantcatdog: I was thinking that if a user can't see all rows at the same time, it tells them that there are no rows they can't see that are selected. If you're watching for it, it's easy to understand that no label means nothing selected. But if you're not watching it, there's nothing explicitly saying that nothing is selected.

FlyersPh9: We initially went with removing the text when there is no selection because that is what [this demo (our inspiration) did.](https://codesandbox.io/s/fervent-sutherland-37vruc)

packages/itwinui-css/backstop/tests/table.html
adamhock: new line added. I forgot there's no stylelint for the demo pages

packages/itwinui-css/backstop/tests/table.html
elephantcatdog: Gotcha. I don't think the current way is bad; just wanted to point out another way

elephantcatdog (approved):
Looks good!

PR #764, opened 2022-09-14
fix(Slider): Dynamic resize of rail based on label size

rohan-kadkol:


elephantcatdog (commented):
Is there a way for users to make long labels on multiple lines and still be centered?

Is that something that we should even consider? @mayank99 do you have thoughts on whether we should plan for that?

rohan-kadkol:
> Is there a way for users to make long labels on multiple lines and still be centered?
> 
> Is that something that we should even consider? @mayank99 do you have thoughts on whether we should plan for that?

If I understood you correctly, I guess they have to manually add new line characters in the text or manually add a `max-width` property for the labels.

E.g. `max-width: 50px` for both labels
image

To center the text that spans across multiple lines within the label, then they may have to use `text-align: center`, because currently, minLabel has `test-align: right` and maxLabel has `text-align: left`

E.g. `test-align: center` for both labels
image

elephantcatdog:
> If I understood you correctly, I guess they have to manually add new line characters in the text or manually add a max-width property for the labels.

Yes, much like this. I meant vertically centered though, like so:
image

rohan-kadkol:
> > If I understood you correctly, I guess they have to manually add new line characters in the text or manually add a max-width property for the labels.
> 
> Yes, much like this. I meant vertically centered though, like so: image

Ohh, I see. As of now, we don't center vertically. I experimented with simply applying `align-items: center` to the horizontal span, it centers the (rail + ticks) for the second slider, but it affects the first and third sliders. So we might need to look into it more to support it.

image

elephantcatdog:
I imagine most users won't have several stacked sliders like in our examples. Seeing that it is possible to vertically center multiline labels is good enough for me.

elephantcatdog (approved):

FlyersPh9 (approved):

PR #765, opened 2022-09-15
feat: Unified disabled & loading states

FlyersPh9:
- Loading state and disabled state visual styling and interactions unified across checkbox, table, & tile:
  - Disabled: muted text, grayscale icon, disabled cursor.
  - Loading: muted text, grayscale icon, loading cursor, italic text.
- Tile `.iui-progress` modifier class renamed to `.iui-loading` to match [the class modifier used on checkbox](https://github.com/iTwin/iTwinUI/blob/b67cfac985e5e7668eaee58ec7a4eb0cb937f13e/packages/itwinui-css/src/checkbox/checkbox.scss#L18).
- Table `.iui-loading` introduced.
- Disabled / loading table rows & tiles display the show more button. 

I originally had some of these changes in #762 but felt they were out of scope for that PR, so created this instead.


FlyersPh9:
Previously we had both `.iui-loading` and `.iui-progress` modifier classes.  I picked `.iui-loading` to replace `.iui-progress`, but I am wondering if instead should use `data-iui-loading="true"` or some `aria-` property.

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -124,6 +124,16 @@
...
     </script>
+    <script>
+      customElements.define(
+        'svg-filetype-video',
+        class extends HTMLElement {
+          connectedCallback() {
+            this.innerHTML = `<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="m11 3v-3h-9v16h12v-13z" fill="#e64821"/><path d="m7 7.2919 3 1.7081-3 1.756zm5 1.7081a4 4 0 1 1 -4-4 4 4 0 0 1 4 4zm-1 0a3 3 0 1 0 -3 3 3 3 0 0 0 3-3z" fill="#fff"/><path d="m11 0 3 3h-3z" fill="#a13217"/></svg>`;
+          }
+        }
+      );
```
rohan-kadkol, line 135: I'm slightly new to this. Could you please explain what these lines of code do?

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -124,6 +124,16 @@
         }
       );
     </script>
+    <script>
```
rohan-kadkol, line 127: Just wondering, can we merge all the `customElements.define` code from the above lines into just one `<script>`? Or is there a reason they are all in separate tags?

packages/itwinui-css/backstop/tests/table.html
mayank99: Have you tried to google? It's creating a custom element called `<svg-filetype-video>` so that when the page loads, the svg code is placed inside it.

https://developer.mozilla.org/en-US/docs/Web/Web_Components
https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements

packages/itwinui-css/backstop/tests/table.html
mayank99: It can be done in the same script, but this is just the convention we've been following. No reason in particular.

packages/itwinui-css/src/table/table.scss
```diff
@@ -254,14 +254,13 @@
...
     .iui-table-cell[aria-disabled='true'],
     &.iui-table-expanded-content[aria-disabled='true'] {
-      font-style: italic;
-      cursor: not-allowed;
       color: var(--iui-color-foreground-4);
 
-      &.iui-slot:hover > .iui-table-more-options {
-        opacity: 0;
+      .iui-table-more-options {
+        display: none;
```
mayank99, line 263: This was made `opacity: 0` to keep it tabbable through keyboard

packages/itwinui-css/backstop/tests/table.html
```diff
@@ -712,7 +701,7 @@ <h2>Default</h2>
               <div class="iui-table-cell-start-icon">
                 <svg-filetype-vector aria-hidden="true"></svg-filetype-vector>
               </div>
-              MyFileWithAReallyLongNameThatWillBeSplitIntoMultipleLinesAfterSoLong_FinalVersion_V2.dgn
+              MyFileWithAReallyLongNameThatWillBeSplitIntoMultipleLinesAfterSoLong_FinalVersion_V2.svg
```
rohan-kadkol, line 704: Is the change on line 704 (`.dgn` -> `.svg`) because the icon on line 702 is of the type vector/svg?

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -280,7 +290,7 @@ <h2>Everything turned on</h2>
           </div>
         </div>
 
-        <div class="iui-tile-content ">
+        <div class="iui-tile-content">
```
rohan-kadkol, line 293: Is this an avoidable change? Because when I tried to format the file after readding the deleted space, the formatter kept it as is.

Or maybe it is good to clean up spaces and cause these small changes in the PR, in this case

FlyersPh9: I've reverted the empty space cleanup for easier reviewing.

packages/itwinui-css/backstop/tests/tile.html
```diff
@@ -1345,16 +1355,16 @@ <h2>Action States</h2>
...
               aria-hidden="true"
-            ></svg-filetype-imodel>
+            ></svg-filetype-video>
           </div>
         </div>
 
         <div class="iui-tile-content">
           <div class="iui-tile-name">
-            <span class="iui-tile-name-label">Disabled</span>
+            <span class="iui-tile-name-label">unavailable.mov</span>
```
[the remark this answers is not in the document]
FlyersPh9: I changed them to match the file names in the table demo, but that's probably not necessary so I've reverted this change.

packages/itwinui-css/src/table/table.scss
rohan-kadkol: When I changed it to `opacity: 0`, it allows tabbing through the loading's row's more options. Not sure if that is intended, because we don't allow tabbing through the textbox with "alpha.dgn"

[vokoscreenNG-2022-09-28_10-47-59.webm](https://user-images.githubusercontent.com/45748283/192811544-f3fd3a19-cde6-4527-af18-0545347797a5.webm)

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -106,17 +109,10 @@
...
-
+  &.iui-loading,
   &[aria-disabled='true'] {
-    cursor: not-allowed;
-
-    .iui-tile-name-label {
-      color: var(--iui-color-foreground-4);
-    }
+    --_iui-tile-title-text-color: var(--iui-color-foreground-4);
+    --_iui-tile-body-text-color: var(--iui-color-foreground-4);
```
rohan-kadkol, line 115: I'm not too sure of css variable scopes, but is --_iui-tile-body-text-color already defined for this selector scope on [line 9](https://github.com/iTwin/iTwinUI/pull/765/files#diff-875f9cda040043eba8e927b119436e84c98074bbd862870ed84930dd5265dc25R9)? Because we are still in the code block of [line 9](https://github.com/iTwin/iTwinUI/pull/765/files#diff-875f9cda040043eba8e927b119436e84c98074bbd862870ed84930dd5265dc25R7)

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -132,6 +128,15 @@
...
   }
+
+  &.iui-loading {
+    cursor: progress;
+    font-style: italic;
+  }
+
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
```
rohan-kadkol, line 139: Is this selector the same as the selector on [L113](https://github.com/iTwin/iTwinUI/pull/765/files#diff-875f9cda040043eba8e927b119436e84c98074bbd862870ed84930dd5265dc25R113)? Or maybe I am missing something

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -5,6 +5,9 @@
 @import '../utils/index';
 
 @mixin iui-tile {
+  --_iui-tile-title-text-color: var(--iui-color-foreground-2);
+  --_iui-tile-body-text-color: var(--iui-color-foreground-4);
```
rohan-kadkol, line 9: Is the `_` prefix on the variable name a convention rule to follow when creating css variables in a specific component's scss file? And without `_` means the variable is from the variables package?

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -350,8 +356,4 @@
   @at-root &.iui-folder .iui-tile-thumbnail {
     box-shadow: var(--iui-size-2xs) 0 0 var(--iui-color-foreground-#{$status});
   }
-
-  .iui-tile-name-label {
-    color: var(--iui-color-foreground-#{$status});
-  }
```
rohan-kadkol, line 356: Are these lines removed because it gets the color from the `.iui-tile-content .iui-tile-name`?

image

packages/itwinui-css/src/table/table.scss
mayank99: > When I changed it to `opacity: 0`, it allows tabbing through the loading's row's more options.

Yes, that is my comment. We want to keep the button accessible to keyboard users.

> we don't allow tabbing through the textbox with "alpha.dgn"

That is a separate issue that we should fix separately.

FlyersPh9: Reverted the `display: none` to `opacity: 0` for table rows.  Now the more options button can be accessed in both the disabled and loading states.  Previously we were disabling the tile's more options button in disabled / loading states, the tile has been updated to align with table rows.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Lines 112-126 are the styling for **_both_** loading & disabled states.
Lines 128-131 are the styling for just the loading state.
Lines 133-135 are the styling for just the disabled state.

mayank99 (approved):
Looks good

1. CI doesn't run on PRs merging into v1, so make sure tests are passing locally before merging.
2. Make sure to document user-facing api changes (e.g. class names) after merging.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Correct, the tile name is getting it's color from `var(--_iui-tile-title-text-color)`.

Previously, we set `.iui-tile-name`'s color and then would override `.iui-tile-name-label` when the status modifier was applied.
```scss
.iui-tile-name {
  color: var(--iui-color-foreground-2);
}
.iui-tile.iui-positive .iui-tile-name-label {
  color: var(--iui-color-foreground-positive);
}
```

Now , we set `.iui-tile-name` to a component specific CSS variable mapped to generic CSS variable, allowing us to then change the component specific CSS variable instead.
```scss
.iui-tile-name {
  --_iui-tile-title-text-color: var(--iui-color-foreground-2);
  color: var(--_iui-tile-title-text-color);
}
.iui-tile.iui-positive {
  --_iui-tile-title-text-color: var(--iui-color-foreground-positive);
}
```

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: You may find some instances where we are not following this pattern.  We introduced this pattern a little bit after we started switching to CSS variables.  We should update areas that are not adhering to this pattern.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Ah, good catch, I have removed the redundant line 115.

FlyersPh9:
> 1. CI doesn't run on PRs merging into v1, so make sure tests are passing locally before merging.

Tests are passing! ✅

> 2. Make sure to document user-facing api changes (e.g. class names) after merging.

Documented all user-facing changes in the [v1 Migration Guide Wiki](https://github.com/iTwin/iTwinUI/wiki/v1-Migration-guide#feat-unified-disabled--loading-states-jon), will update the link to point to #756 once it's merged.

rohan-kadkol (approved):

PR #767, opened 2022-09-23
feat(ComboBox): Add multi-select tags to Combobox

gretanausedaite:
- Add `position: relative` to input-with-icon to compensate for `iui-select-tag-container` position: absolute.
- Targeting `iui-select-tag-container` from input with icon to add max-width and right: unset.


packages/itwinui-css/backstop/tests/select.html
```diff
@@ -99,6 +99,36 @@ <h2>Default</h2>
         </div>
         <div class="iui-message">Help message.</div>
       </label>
+      <label class="iui-input-container iui-with-message">
+        <div class="iui-label">ComboBox label</div>
+        <div class="iui-input-with-icon">
+          <input
+            class="iui-input"
+            placeholder="Placeholder…"
+            style="padding-left: 414px;"
```
gretanausedaite, line 108: This padding will be dynamically changed in React.
Calc should be something like: left inset of tag container + tag container width + gap between tags.

packages/itwinui-css/src/input/input-with-icon.scss
```diff
@@ -27,6 +28,11 @@
       border-color: var(--_iui-input-with-icon-hover-border-color, var(--iui-color-foreground-2));
     }
   }
+
+  > .iui-select-tag-container {
+    max-width: 70%;
```
gretanausedaite, line 33: max-width value is just something I chose. We can discuss/change it.

packages/itwinui-css/backstop/tests/select.html
rohan-kadkol:
```suggestion
            style="padding-left: min(414px, calc(70% + 20px))"
```
Something like this might be a temporary fix to also show the place holder when the width is small. I used 70% since the `max-width` of the iui-selected-container is 70% ([code](https://github.com/iTwin/iTwinUI/pull/767/files#diff-d21bc8a15e3189fb58f907a3c1fa5eb327ec5586ca65a4d55fc17b0f5847c411R33))

But since this will be dynamically updated in React, maybe just `414px` might be good

rohan-kadkol (approved):

bentleyvk:
Everything looks ok but I think we could wait until you have figured out the bug where tag container does not increase on resize.

PR #768, opened 2022-09-23
fix(NotificationMarker): Clip icon with `mask` instead of border

rohan-kadkol:
* Benefit: Can now have any background color
  * Old (with background colored border on circle):
    * image
  * New (cut circle from top right of content using masking):
    * image

<br/>

* Used masking instead of border to cut a partial circle from the top right corner of the content of the notification marker
  * Masking with `radial-gradient` instead of svg/png
* No major change in test images
  * Minor pixel changes due to the smoothing effect of the cuts of the top right circle and also since now we're using masking instead of border
    * [Diff in side-navigation](https://user-images.githubusercontent.com/45748283/192276853-4860e9b4-3b82-41c6-8e29-c6cf8fe3d863.webm)

### Test image changes

Most test changes resulted from the following:

* Changing `line-height` to `1` ([L14](https://github.com/iTwin/iTwinUI/pull/768/files#diff-b2e4ed571b9f44ba530e2dfa36a85a31706270f62ca9fc97791963f55ce101feL14-R14))
  * As a result, buttons became smaller (`26px` -> `24px`)
  * E.g. `buttons.html -> Default`
    * image

* Removed `margin-block` from `.iui-button-icon` ([L39](https://github.com/iTwin/iTwinUI/pull/768/files#diff-b2e4ed571b9f44ba530e2dfa36a85a31706270f62ca9fc97791963f55ce101feL39))
  * As a result, top and bottom margin become smaller for some icon buttons
  * E.g. `color-picker.html -> Advanced`
    * image

### Testing

* ~~[ ]~~ Pass Manual CI: [#3379](https://github.com/iTwin/iTwinUI/actions/runs/3152062009)
* [x] Approved test image changes


packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
```diff
@@ -2,25 +2,38 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 
 $iui-notification-marker-border-width: var(--iui-size-3xs);
+$iui-notification-before-size: calc(var(--iui-size-xs) + $iui-notification-marker-border-width * 2);
+$iui-notification-before-size-radius: calc($iui-notification-before-size * 0.5);
```
mayank99, line 6: Can you explain what these variables mean and how they help in positioning? The names are not very clear.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
```diff
@@ -2,25 +2,38 @@
...
     }
   }
 
+  > * {
+    mask: radial-gradient(
+      ellipse $iui-notification-before-size-radius $iui-notification-before-size-radius at
+        calc(100% - $iui-notification-before-size-radius - var(--iui-size-s) * -0.5)
+        calc($iui-notification-before-size-radius + var(--iui-size-s) * -0.5),
+      transparent 95%,
+      #000 100%
```
mayank99, line 33: nit: why not `black`?

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
```diff
@@ -2,25 +2,38 @@
...
+
     border-radius: 100%;
     background-color: var(--iui-color-foreground-#{$status});
-    border: $iui-notification-marker-border-width solid var(--iui-color-background-1);
     @media (prefers-reduced-motion: no-preference) {
       transition: background-color var(--iui-duration-1) ease-out, border-color var(--iui-duration-1) ease-out;
     }
   }
 
+  > * {
```
mayank99, line 27: Since we now rely on always having the svg as a child, I would suggest updating the html so that the notification-marker class is applied on a wrapper element (e.g. `span`) rather than the custom `svg-` element

mayank99 (commented):
PR title should have "fix", not "chore" because this is a user-affecting change. Also, did you mean "cutout" instead of "knock out"?

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: I was getting this error. Maybe because of the linting restrictions for named colors? [(Stackoverflow)](https://stackoverflow.com/a/50484451/11547064)
image

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: Ohh, yes, I think the names are not clear. I added comments and also changed the names as follows:
* `$iui-notification-marker-border-width` -> `$iui-notification-marker-spacing` (since now there is no border)
* `$iui-notification-before-size` -> `$iui-notification-marker-size`
* `$iui-notification-before-size-radius` -> `$iui-notification-marker-size-radius`

Do you recommend some other names? Also, do you think the comments are needed or the names are descriptive enough?

mayank99: Comments shouldn't be necessary if the names are descriptive. But even with your explanations, I don't see how these variables help with positioning/sizing.

1. The `width` and `height` of the `::before` element can just be `var(--iui-size-xs)`, replacing the need for marker-size and marker-spacing.
2. The -radius variable is not used directly (it has more math applied on it), so I'm not sure what's even the purpose of it.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: **[DONE]** I changed the dom structure for all the `.iui-button-icon`s with the notification marker (without notifications icons are unchanged).

Old
image

New
Screenshot from 2022-09-29 13-37-22

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: After calculating the actual values of the math calculations, some of them turned out to be 0. Also, some calculations were unnecessary (like the `width` and `height` you mentioned in point 1).

I simplified those calculations, tried to use more descriptive variable names, and also limited to only using variables that I felt were useful.

rohan-kadkol:
> PR title should have "fix", not "chore" because this is a user-affecting change. Also, did you mean "cutout" instead of "knock out"?

* Title changed to "fix".
* I think Jon called it the "knock out" effect, but "cutout" also works. @FlyersPh9, what do you suggest?

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
mayank99: the `iui-button-icon` class shouldn't go on the span, it should go on the svg

packages/itwinui-css/src/button/button.scss
```diff
@@ -37,6 +37,10 @@
   transition: fill var(--iui-duration-1) ease-out;
   fill: currentColor;
   margin-block: var(--iui-size-2xs);
+
+  > * {
+    display: flex;
+  }
```
rohan-kadkol, line 43: Jon suggested this idea of using `display: flex` so that the `<svg-document>` inside the wrapper `<span>` will expand to the size of the wrapper `<span>`. Without it, if the `<span>` for example was `24x24`, then the `<svg-document>` would be `0x24`.

This works, but I was just wondering, why does it work? Also for e.g. in React, we directly have `<svg>` instead of `<svg-document>`. There it works without L41-43. Why is that so?

CSS
image

React
Screenshot from 2022-09-29 13-38-32

FlyersPh9:
@rohan-kadkol I have heard and used "knock out", "cut out", and probably some other terms as well.  We should probably use the term "clipped" as that's more defined.  How about **fix(NotificationMarker): Clip icon with `mask` instead of border**?

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: I did this to match what happens in React. `IconButton` adds the `.iui-button-icon` to the child tag itself (Using `React.cloneChild`. [Code](https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/Buttons/IconButton/IconButton.tsx#L61-L67)). 

As a result, the below React code gives the below dom:

image

image

In the above screenshot, `.iui-button-icon` is on the `span`, which is why I did the same in CSS. Also, I couldn't get it to work in CSS by putting `.iui-button-icon` on the svg and not the span, but I can try again.

rohan-kadkol:
> @rohan-kadkol I have heard and used "knock out", "cut out", and probably some other terms as well. We should probably use the term "clipped" as that's more defined. How about **fix(NotificationMarker): Clip icon with `mask` instead of border**?

That sounds clearer, thanks. Changed the PR title

FlyersPh9:
`@mixin iui-notification-marker-hover` no longer seems necessary & can be removed from `/button/default.scss` as well.

packages/itwinui-css/src/button/button.scss
FlyersPh9: Typically `> *` has me pretty nervous because we're never sure what might be nested, but because this is being applied on `.iui-button-icon` I'm thinking it might be ok here.

rohan-kadkol:
> `@mixin iui-notification-marker-hover` no longer seems necessary & can be removed from `/button/default.scss` as well.

Ohh, thanks, removed. All tests are still passing.

To confirm, is this because since we don't use border anymore, we don't need to change the border color on hover to match the background's color on hover?

FlyersPh9:
I tested the notification marker on areas that we have not intended for it to be used just to see how flexible the CSS is.  I applied the notification marker to an `<h2>` element and the notification marker went far right on the page.  Screenshot:
Screen Shot 2022-10-04 at 5 17 59 PM

If I add `width: fit-content;` to the notification marker, it does visually look correct:
Screen Shot 2022-10-04 at 5 19 00 PM

However, I think `width: fit-content;` could cause unintended issues.  For example, in `<span class="iui-button-icon iui-notification-negative iui-urgent">` `.iui-button-icon` may have it's own width set & notification marker might conflict with it.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
mayank99: > I did this to match what happens in React. `IconButton` adds the `.iui-button-icon` to the child tag itself (Using `React.cloneChild`. [Code](https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/Buttons/IconButton/IconButton.tsx?rgh-link-date=2022-09-30T18%3A28%3A19Z#L61-L67)).

Then we might need to change the approach, instead of working around it in css.

> I couldn't get it to work in CSS by putting `.iui-button-icon` on the svg and not the span, but I can try again.

What problem where you having?

I would much rather _always_ have a span for `iui-button-icon` or _never_ have it, rather than trying to cover both cases like we need to do now.

rohan-kadkol: > What problem where you having?

After adding `.iui-button-icon` to the svg-document, the width of it goes to 0 (`0x24`). Even adding a `width: 24px` inline style doesn't fix it. I could look into it more, but here's what I noticed so far.

image

rohan-kadkol:
> Currently we apply a notification marker with urgent status like so: `class="iui-notification-positive iui-urgent"`
What if we use data attributes instead of class modifiers, something like: `class="iui-notification-marker" data-iui-urgent="true" data-iui-variant="positive"` .......

Thanks for the detailed explanation! I made the changes you suggested.

We now use data attributes: `data-iui-variant` and `data-iui-urgent` instead of classes: `.iui-notification-#{status}` and `.iui-urgent` respectively.

rohan-kadkol:
> I tested the notification marker on areas that we have not intended for it to be used just to see how flexible the CSS is. I applied the notification marker to an `<h2>` element and the notification marker went far right on the page. Screenshot: Screen Shot 2022-10-04 at 5 17 59 PM
> 
> If I add `width: fit-content;` to the notification marker, it does visually look correct: Screen Shot 2022-10-04 at 5 19 00 PM
> 
> However, I think `width: fit-content;` could cause unintended issues. For example, in `<span class="iui-button-icon iui-notification-negative iui-urgent">` `.iui-button-icon` may have it's own width set & notification marker might conflict with it.
> 
> UPDATE: `display: inline-flex` also visually works but I think could also cause unintended issues.

Thanks for noting this. After our discussion, we decided to not specify a `width` on the notification span (e.g. `fit-content`, etc.). Instead, we will let the user specify the width of the notification span themselves according to their needs.

This is to make the notification marker generic for all use cases and not cause unintended issues with using `width: fit-content`, like the width conflict you mentioned.

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: > the iui-button-icon class shouldn't go on the span, it should go on the svg

I changed it to apply `.iui-button-icon` on the `svg-document` and not the wrapper span.

##### Previous
Screenshot from 2022-09-29 13-37-22

##### Latest
image

1. The `0` width problem was because of `overflow: hidden` property near [L107](https://github.com/iTwin/iTwinUI/pull/768/files/8398b1e7a2580b66e959a441765316a28fa2bbec..acd42e3e527d13ef1ee2fe19ca667b53955ce87a#diff-187108cb5bf3f6ce0ad8a514769689db86ffe4102ee775050d0b4c02dda54bedR107). I changed that to only target the label span and not the motivation marker span.

2. There was also a `margin-block` (orange portion in screenshot) on the `.iui-button-icon` that was causing the circle's incorrect vertical placement.
image
image
I fixed this by removing this when the root of the `.iui-button-icon` has `.iui-notification-marker` [L41-44](https://github.com/iTwin/iTwinUI/pull/768/files/8398b1e7a2580b66e959a441765316a28fa2bbec..acd42e3e527d13ef1ee2fe19ca667b53955ce87a#diff-b2e4ed571b9f44ba530e2dfa36a85a31706270f62ca9fc97791963f55ce101feR41-R44)

All tests still pass, but just to confirm, would the changes in `button.scss` and `side-navigation.scss` in these [two commits](https://github.com/iTwin/iTwinUI/pull/768/files/8398b1e7a2580b66e959a441765316a28fa2bbec..acd42e3e527d13ef1ee2fe19ca667b53955ce87a#diff-b2e4ed571b9f44ba530e2dfa36a85a31706270f62ca9fc97791963f55ce101fe) break anything?

Also, for the React side, I guess we have to think of how to implement it there, since there, `.iui-button-icon` is applied on the wrapper `span`

mayank99: I think we might want to do away with that `margin-block`. It was added to make the height of the small icon button match the height of the small text button.

We explicitly set the `min-height` to 24px, but since the line-height of text is already 24px and there's a 1px on each side, the total comes out to be 26px. To match this 24px text, we add 4px margin above and below the 16px icon.

We could fix this one of two ways (@FlyersPh9 might want to comment on this part):
1. Set min-height to explicitly be 26px instead of 24px, OR
2. Decrease line-height so that small button actually becomes 24px tall instead of 26px.

---

As for React, I think our users are already having issues with `React.cloneElement`, so we might want to change the approach to _always_ have a wrapping span. The DOM structure could look something like this:
```html
<button>
  <span class='iui-button-icon'>
    <svg>...</svg>
  </span>
  <span>Text</span>
</button>
```

And then react could would just wrap the icon instead of cloning (similar change in IconButton):

```jsx
const Button = ({ startIcon, children }) => (
  <button>
    <span class='iui-button-icon'>{startIcon}</span>
    <span>{children}</span>
  </button>
);
```

Based on that react change, if the user were to do this:

```jsx
<IconButton>
  <NotificationMarker>
    <SvgSomething />
  </NotificationMarker>
</IconButton>
```

it would result in this markup:
```html
<button>
  <span class='iui-button-icon'>
    <span class='iui-notification-marker'>
      <svg>...</svg>
    </span>
  </span>
</button>
```

which shouldn't be a problem for us if we make sure our `iui-button-icon` handles svgs nested deeply like this:

```css
.iui-button-icon {
  svg {
    width: 16px;
    height: 16px;
    fill: currentColor;
  }
}
```

FlyersPh9: I just now tried removing `line-height` from the button and `margin-block` from the button icon.  Height wise, I like that the small button is now 24px tall.  What I don't like is that the text label appears to be slightly off center vertically.  Maybe playing around with the `line-height` will resolve this?
Screen Shot 2022-10-12 at 10 28 21 AM

mayank99: `line-height: 1.2` seems to work

rohan-kadkol: I made the following changes yesterday:
* In the CSS repo, I changed the line height to 1 ([Commit](https://github.com/iTwin/iTwinUI/pull/768/commits/e015a91490bd6fdf629261ae96092ee2b180998c)). I thought of line height of 1.2, as Mayank suggested, but saw now difference between 1 and 1.2. Using 1 also solved the alignment problem Jon was mentioning above. Both red rectangles are 20px tall.
  * button-line-height
  * Also the small icon button and small text button are 24px tall.
  * This led to 69 test images fail, because buttons are used in many components.

* In the React repo's NotificationMarker PR, I removed `React.cloneElement` ([Commit](https://github.com/iTwin/iTwinUI-react/pull/829/commits/48ae082db32515d384fa2f79cae47ab634435888)).
  * 4 unit tests failed because of incorrect selectors after this change (e.g. using `>` selector or `.firstElementChild`).

mayank99: i saw a difference between `1` and `1.2` on my machine (better aligned with `1.2`) and also if the button ever has multiple lines of text for some reason, 1.2 will look better

rohan-kadkol: Ohh, I think the multiple lines point is important, thanks. Changed `line-height` to `1.2`
`line-height: 1`
image

`line-height: 1.2`
image

adamhock (approved):
I approve on the contingency that Mayank's comments are resolved which it appears they are

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: Is it possible that browser rendering differences could cause unexpected results? I tested `line-height` with `1` and `1.2` on the `buttons.html -> Default -> Small Right Icon` manually/visually on Windows and Linux and got different results.

1. On Windows, `1` and `1.2` gave **same results**. The top and bottom gaps were 8 and 9 units respectively. Both red rectangles are 8 units tall.

##### 1 and 1.2
image

2. On Linux, `1` and `1.2` gave slightly **different results**. the top and bottom gaps were 9 & 12 units for `1`, and 10 & 11 units for `1.2`. The red rectangles are 9 units tall for `1` and 10 units tall for `1.2`

##### 1
MicrosoftTeams-image (5)

##### 1.2
MicrosoftTeams-image (4)

3. When running automated tests (`yarn test`), with `1`, **75 tests failed** and with `1.2`, **95 tests failed**.

Since the automated tests always run in the same environment, and since `1` caused the least test changes, I chose the `line-height` to be `1`. What do you think @mayank99 @FlyersPh9?

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -102,7 +102,7 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
       }
     }
 
-    > span {
+    > span:last-of-type {
```
rohan-kadkol, line 105: To select the label span and not the new notification span

packages/itwinui-css/src/utils/notification-marker/notification-marker.scss
rohan-kadkol: As per our previous discussion, changed the `line-height: 1.2` since that seems to be the most vertically centered in the screenshots and on Mayank's system.

I made these non-notification related changes in CSS PR [#788](https://github.com/iTwin/iTwinUI/pull/788) and React PR [#895](https://github.com/iTwin/iTwinUI-react/pull/895).

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -114,7 +114,7 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
       border-color: var(--iui-color-background-border);
     }
 
-    > .iui-button-icon:not(.iui-avatar) svg {
+    .iui-button-icon:not(.iui-avatar) svg {
```
rohan-kadkol, line 117: Removed the `>`, since now `span.iui-notification-marker` comes between `.iui-sidenav-button` and `.iui-button-icon`, as on [L132-145](https://github.com/iTwin/iTwinUI/pull/768/files#diff-324482a88e5ac05671e3d09fda294087190f1cef2c7af7631b34b7df82e9da86R132-R145).

rohan-kadkol:
After recently merging CSS PR #788 (the PR to which we transferred the "remove margin-block" and "change line-height" tasks since they were not related to the notification marker), a lot of the test changes impertinent to the notification marker were approved in that PR.

Test changes related to the combobox arrow from CSS PR [#796](https://github.com/iTwin/iTwinUI/pull/796), will be approved in CSS PR [#762](https://github.com/iTwin/iTwinUI/pull/762) (as mentioned in https://github.com/iTwin/iTwinUI/pull/762#issuecomment-1301194369). Thus, those changes are ignored in this PR and not approved.

Thus, after merging the `v1` branch with this branch and fixing the failing image tests, only two test image changes remain and both are related to the notification marker.

packages/itwinui-css/src/side-navigation/side-navigation.scss
mayank99: I don't think we should be relying on dom position here.

rohan-kadkol: Earlier just the `span` selector worked since there was just one span under `.iui-sidenav-button`.

But now with the notification marker `span` also between `.iui-sidenav-button` and `.iui-button-icon` (as on [L132-145](https://github.com/iTwin/iTwinUI/blob/b23406cca9655588a3e474ee76a3c35d6b32f361/packages/itwinui-css/backstop/tests/side-navigation.html#L132-L145)), there are two spans under `.iui-sidenav-button` ([L140](https://github.com/iTwin/iTwinUI/blob/b23406cca9655588a3e474ee76a3c35d6b32f361/packages/itwinui-css/backstop/tests/side-navigation.html#L140) and [L149](https://github.com/iTwin/iTwinUI/blob/b23406cca9655588a3e474ee76a3c35d6b32f361/packages/itwinui-css/backstop/tests/side-navigation.html#L149)).

What alternative do you recommend to select the label span but not the notification span?

rohan-kadkol: Thanks, replaced with `> span:where(:not(.iui-notification-marker))`, since `.iui-notification-marker` is the span we want to exclude, and not `.iui-button-icon`

mayank99: Are you sure? I think `iui-button-icon` will be a parent of `iui-notification-marker`, so that's what we want to exclude.

rohan-kadkol: Ohh, right, but I think we need to exclude both, `.iui-button-icon` and `.iui-notification-marker`.

Because the parent for the icon is `.iui-button-icon` when there's no notification (E.g. [L181](https://github.com/iTwin/iTwinUI/blob/b23406cca9655588a3e474ee76a3c35d6b32f361/packages/itwinui-css/backstop/tests/side-navigation.html#L181)), but is `.iui-notification-marker` when there's a notification. (E.g. [L140](https://github.com/iTwin/iTwinUI/blob/b23406cca9655588a3e474ee76a3c35d6b32f361/packages/itwinui-css/backstop/tests/side-navigation.html#L140))

Changed it to `> span:where(:not(.iui-button-icon, .iui-notification-marker))`

mayank99: i don't think `iui-notification-marker` should ever be a parent of `iui-button-icon`. this is an impossible scenario in react because we wrap the icon prop (where the user will pass `NotificationMarker`) with `<span className='iui-button-icon'>`

rohan-kadkol: Ohh, I see, good point. I made the change like you suggested:

Old
image

New
image

Now `.iui-notification-marker` is never the parent of `.iui-button-icon`. Also updated the svg selector to what you suggested: `> span:where(:not(.iui-button-icon))`

mayank99 (approved):

rohan-kadkol:
Just to mention, the 18 test cases failing in job [#3399](https://github.com/iTwin/iTwinUI/actions/runs/3395983553/jobs/5646527002) after this merge are the changes related to the combobox arrow from PR [#796](https://github.com/iTwin/iTwinUI/pull/796) that are/will be approved in PR [#762](https://github.com/iTwin/iTwinUI/pull/762).

PR #772, opened 2022-10-03
fix(Table): Disabled cell with monochrome icon use consistent fill color

### LINKED ISSUE #751: Disabled table cell icons are different colours
Grayscale filter makes icons different colours. 
image

FlyersPh9:
- All start / end monochrome icons are given a default fill color of `--iui-color-foreground-4`.
- If row has `data-iui-status`, then end monochrome icon has `filter` removed.
- Removed unnecessary `.iui-button-icon` class in some areas.

Closes #751.

**Before:**
Screen Shot 2022-10-04 at 12 40 45 PM Screen Shot 2022-10-04 at 12 42 02 PM

**After:**
Screen Shot 2022-10-04 at 12 40 45 PM Screen Shot 2022-10-04 at 12 40 55 PM


packages/itwinui-css/src/table/table.scss
```diff
@@ -440,6 +442,7 @@
 
     svg:not(.iui-radial) {
       @include iui-icon-style('m');
+      fill: var(--_iui-table-cell-icon-fill);
```
gretanausedaite, line 445: Test images did not change?

FlyersPh9: We don't have any tests that catch this change.  I would need to add a new demo that tests combinations of status applied on rows & disabled / loading states.

FlyersPh9:
@gretanausedaite @adamhock I changed my approach.  

Originally I had all monochrome icons fill color change so when `filter: grayscale(100%)` was applied all the icons visually appeared the same color of gray.
Now the same can be said _**except**_ for icons that have a status color applied.  See the new screenshots in the PR description.

packages/itwinui-css/src/table/table.scss
```diff
@@ -251,12 +253,15 @@
     &.iui-loading,
     .iui-table-cell[aria-disabled='true'],
     &.iui-table-expanded-content[aria-disabled='true'] {
+      --_iui-table-cell-icon-fill: var(--iui-color-foreground-4);
+      --_iui-table-cell-image-filter: grayscale(100%);
```
FlyersPh9, line 257: This applies to icons, images, and `iui-avatar` so I'm not sure what the variable should be named.

packages/itwinui-css/src/table/table.scss
gretanausedaite: We probably should add tests for disabled state to catch any unwanted css changes in the future. It can be added to the same table - just add 3 more rows with status and disabled state.

packages/itwinui-css/src/table/table.scss
gretanausedaite: Naming looks good to me.

gretanausedaite:
> Originally I had all monochrome icons fill color change so when `filter: grayscale(100%)` was applied all the icons visually appeared the same color of gray. Now the same can be said _**except**_ for icons that have a status color applied. See the new screenshots in the PR description.

What made you change the approach? Is there a reason why status icons have to be coloured all the time? To me it looks like a bug now - someone forgot to include those icons into disabled state. But I guess it's consistent with loading icon blue. 
image

FlyersPh9:
> What made you change the approach? Is there a reason why status icons have to be coloured all the time?

The way it is setup now is if there is a right aligned icon but the row has no status, the icon is set to muted gray color.  If the row _does_ have a status applied, then the right aligned icon ignores the grayscale filter.  My thinking was that these status icons as well as the left aligned stripe are placed there to help reinforce the status of the row & if those status icons are gray they lose some of that meaning.

gretanausedaite:
> The way it is setup now is if there is a right aligned icon but the row has no status, the icon is set to muted gray color. If the row _does_ have a status applied, then the right aligned icon ignores the grayscale filter. My thinking was that these status icons as well as the left aligned stripe are placed there to help reinforce the status of the row & if those status icons are gray they lose some of that meaning.

Makes sense. Just need to test what happens when row has status but right aligned icon is not status icon (we allow users to add anything they like even if row has status).

gretanausedaite:
Added test for row and cell status in both default and disabled state. Added icon to not status disabled row.
image

adamhock (approved):
Looks good to me! 🎉

gretanausedaite (approved):

PR #773, opened 2022-10-04
fix: Make dialog positioned and remove unneeded styles

mayank99:
1. Changed from `position: relative` to `position: fixed` and added a modifier `data-iui-relative` for `position: absolute`
2. Removed sizing rules from draggable dialog because they are already set in default dialog

After spending some time on it, I was unable to remove default placement rules from draggable dialog because the JS (in itwinui-react) relies on it to calculate the initial position.

### DIFF
--- a/packages/itwinui-css/src/dialog/dialog.scss
+++ b/packages/itwinui-css/src/dialog/dialog.scss
@@ -113,10 +113,6 @@ $iui-dialog-width: 380px;
 }
 
 @mixin iui-dialog-draggable {
-  max-width: 100%;
-  max-height: 100vh;
-  min-width: $iui-dialog-width;
-  min-height: 144px;
   display: flex;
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -5,14 +5,18 @@
...
+  position: fixed;
   overflow: hidden;
   top: 0;
   left: 0;
   width: 100%;
   height: 100%;
   pointer-events: none;
   transform: translateX(0); // creates containing block for position: fixed
+
+  &[data-iui-relative] {
```
FlyersPh9, line 17: How about targeting when equal to `true`?
```suggestion
  &[data-iui-relative='true'] {
```

FlyersPh9 (approved):

PR #779, opened 2022-10-11
fix(carousel): Dots are centered

### LINKED ISSUE #685: Carousel: dots are not centered
When there are only few dots they are not centered:
image

gretanausedaite:
Fixes #685

### DIFF
--- a/packages/itwinui-css/src/carousel/carousel.scss
+++ b/packages/itwinui-css/src/carousel/carousel.scss
@@ -46,7 +46,6 @@
   &-dots {
     @include iui-reset;
     @include iui-focus;
-    flex: 2;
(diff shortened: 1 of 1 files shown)


packages/itwinui-css/src/carousel/carousel.scss
```diff
@@ -46,7 +46,6 @@
   &-dots {
     @include iui-reset;
     @include iui-focus;
-    flex: 2;
```
gretanausedaite, line 49: The problem mentioned in issue only appears in dev mode. Should be good to go.

adamhock (approved):
Looks good to me! 🎉

FlyersPh9 (approved):

PR #781, opened 2022-10-14
chore: Add demo page background color switcher

FlyersPh9:
- Add ability to switch the `<body>` background color between 1 & 2 from within `theme.js`.
- Updated visual styling within `theme.js`.
- Page settings button is now fixed top right of the viewport to prevent needing to scroll to the top of the page to adjust settings.
- Hiding the page settings button in `footer.js` test images.

Screen Shot 2022-10-14 at 6 32 31 PM

I found this useful when working on #762.

### DIFF
--- a/packages/itwinui-css/backstop/scenarios/footer.js
+++ b/packages/itwinui-css/backstop/scenarios/footer.js
@@ -10,6 +10,7 @@ module.exports = [
   scenario('Type visible in viewport', {
     selectors: ['viewport'],
     viewports: [{ width: 800, height: 600 }],
+    hideSelectors: ['theme-button'],
     scrollToSelector: ['.iui-legal-footer'],
   }),
 
@@ -17,6 +18,7 @@ module.exports = [
   scenario('State hover and focus', {
     selectors: ['viewport'],
     viewports: [{ width: 800, height: 600 }],
+    hideSelectors: ['theme-button'],
     scrollToSelector: ['.iui-legal-footer'],
     actions: [hover('#test-link-1'), focus('#test-link-2')],
   }),
--- a/packages/itwinui-css/backstop/tests/assets/demo.css
+++ b/packages/itwinui-css/backstop/tests/assets/demo.css
@@ -8,28 +8,24 @@
   font-weight: 300;
   src: url('fonts/noto-sans-v27-latin-300.woff2') format('woff2');
 }
-
 @font-face {
   font-family: 'Noto Sans';
   font-style: normal;
   font-weight: 400;
   src: url('fonts/noto-sans-v27-latin-regular.woff2') format('woff2');
 }
-
 @font-face {
   font-family: 'Noto Sans';
   font-style: normal;
   font-weight: 600;
   src: url('fonts/noto-sans-v27-latin-600.woff2') format('woff2');
 }
-
 @font-face {
   font-family: 'Noto Sans';
   font-style: normal;
   font-weight: 700;
   src: url('fonts/noto-sans-v27-latin-700.woff2') format('woff2');
 }
-
 @font-face {
   font-family: 'Noto Sans';
   font-style: italic;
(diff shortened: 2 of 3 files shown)


packages/itwinui-css/backstop/tests/assets/theme.js
```diff
@@ -33,49 +39,74 @@ class ThemeButton extends HTMLElement {
...
         background-color: var(--iui-color-background-1);
-        padding: 0.5rem 0.25rem;
-        margin: 0.25rem;
+        padding: var(--iui-size-xs) var(--iui-size-s);
+        margin: var(--iui-size-2xs);
+        display: flex;
+        flex-direction: column;
+        gap: var(--iui-size-xs);
+        opacity: 0;
+        user-select: none;
```
mayank99, line 76: why disable user-select?

gretanausedaite (commented):
I have no additional comments

packages/itwinui-css/backstop/tests/assets/theme.js
```diff
@@ -3,27 +3,33 @@ class ThemeButton extends HTMLElement {
...
             <label tabindex="-1"><input type="radio" name="theme" value="light" /><span>Light</span></label>
             <label tabindex="-1"><input type="radio" name="theme" value="dark" /><span>Dark</span></label>
             <label tabindex="-1"><input type="radio" name="theme" value="light-hc" /><span>High contrast light</span></label>
             <label tabindex="-1"><input type="radio" name="theme" value="dark-hc" /><span>High contrast dark</span></label>
           </fieldset>
+
+          <fieldset>
+            <legend>Background color</legend>
+            <label tabindex="-1"><input type="radio" name="background" value="bg1" onclick="document.body.classList.add('force-background-1');" /><span>Background 1</span></label>
+            <label tabindex="-1"><input type="radio" name="background" value="bg2" onclick="document.body.classList.remove('force-background-1');" checked /><span>Background 2</span></label>
```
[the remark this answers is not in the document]
FlyersPh9: Yes, @elephantcatdog helped me with this and I think we've got the change applied.

packages/itwinui-css/backstop/tests/assets/demo.css
```diff
@@ -48,38 +44,42 @@ p {
...
 }
 
 /** For sections in backstop tests */
 section[id^='demo'] {
   width: -moz-fit-content; /* stylelint-disable */
   width: fit-content;
 }
+
+.force-background-1 {
+  background-color: var(--iui-color-background-1) !important;
```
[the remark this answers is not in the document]
FlyersPh9: This is now removed and the style is being applied inline via JS.

packages/itwinui-css/backstop/tests/assets/theme.js
FlyersPh9: I had this happen a couple of times while working on the theme PR, quickly opening, clicking, and closing the popup.
Screen Shot 2022-10-18 at 5 07 45 PM

packages/itwinui-css/backstop/tests/assets/theme.js
```diff
@@ -3,27 +3,33 @@ class ThemeButton extends HTMLElement {
...
         </article>
       </div>
     `;
     const style = /* css */ `
       :host {
-        position: absolute;
-        top: 0;
-        right: 8px;
+        position: fixed;
+        top: var(--iui-size-xs);
```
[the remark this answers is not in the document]
FlyersPh9: Prior to this PR, the page settings button didn't have any background unless hovered.  Now in this PR the button always has a background, and I felt it visually looked a little better with empty space separating it from the edge of the viewport.

packages/itwinui-css/backstop/tests/assets/theme.js
mayank99: Disabling user-select is generally very bad practice. 🫤

Can you share a screen recording of how this happened? I can't seem to reproduce no matter how fast i open->click->dismiss.

packages/itwinui-css/backstop/scenarios/footer.js
```diff
@@ -10,13 +10,15 @@ module.exports = [
   scenario('Type visible in viewport', {
     selectors: ['viewport'],
     viewports: [{ width: 800, height: 600 }],
+    hideSelectors: ['theme-button'],
```
mayank99, line 13: Did this not cause test image changes?

FlyersPh9: This prevented test images from changing.  In this test, it scrolls down the page.  Previously the page settings button was `position: absolute` top right of the page and was above the fold.  Now with the page settings button being `position: fixed` top right, it will appear in our tests unless we hide the button.

packages/itwinui-css/backstop/tests/assets/theme.js
```diff
@@ -120,16 +151,32 @@ class ThemeButton extends HTMLElement {
     `;
   };
 
+  changeBackground = ({ target: { value: _background } }) => {
+    if (_background == 'bg1') {
```
elephantcatdog, line 155: Ah, sorry, can you make this `===` instead of `==`? It's more type safe - the code won't change the type to make comparisons. It doesn't really matter in a lot of circumstances, but it's best practice.

packages/itwinui-css/backstop/tests/assets/theme.js
FlyersPh9: Here's a screen recording of 2 ways I've been accidentally doing it.
1. Double clicking on the label.  I know a single click works, sometimes I accidentally double click.
2. After clicking the button, while moving the cursor to the body of the page i click & sometimes it registers as a click & drag.  In the video I'm purposefully trying to make it happen, but it has happened to me on accident when naturally working.

https://user-images.githubusercontent.com/849817/196744584-cb0a50c1-281c-4d6a-9541-c7375a693a91.mov

packages/itwinui-css/backstop/tests/assets/theme.js
FlyersPh9: Changed to:
```js
if (_background === 'bg1') {
```

packages/itwinui-css/backstop/tests/assets/theme.js
mayank99: Ok, as long as we are not messing with `user-select` in all our components 👍

mayank99 (approved):

PR #785, opened 2022-10-18
fix(Table): Changing start icon fill from <svg> tag

rohan-kadkol:
> @rohan-kadkol This still needs a CSS fix, as we can't expect users to always provide the fill as an inline style.

_Originally posted by @mayank99 in https://github.com/iTwin/iTwinUI-react/issues/881#issuecomment-1281052058_

* Removed fill attribute from start icon but kept it on end icon
  * End icon still needs the fill color for the sort icon on header, cell status icon on the cell, progress indicator, etc.

#### TODOs

- [x] All tests pass
  - **IGNORING** failing two Dialog tests and approving images. Since they are very small pixel changes in unrelated files. Also these failures happened before this PR.
- [x] Tested that this PR actually fixes the problem in React


packages/itwinui-css/src/table/table.scss
```diff
@@ -135,8 +135,6 @@
 }
 
 @mixin iui-table-row {
-  --_iui-table-cell-icon-fill: var(--iui-color-foreground-4);
```
rohan-kadkol, line 138: @mayank99 Removing L138 was not enough, unfortunately. It is giving a variable undefined error, as expected. 

Screenshot from 2022-10-18 09-54-49

But, the fill attribute is still considered. Thus it still overrides the fill on the `<svg>` tag. In this example, I removed the fill from the path of the pdf icon and added a pink fill to the pdf `<svg>` tag. It still shows as black.

Screenshot from 2022-10-18 09-57-35

Thus, I also changed [L452-465](https://github.com/iTwin/iTwinUI/pull/785/files#diff-3ac545d94bb9609fcd00db7227644a76d684ddad2b678eeaa75ac0d0209a251cL452-R465)

mayank99: can you try one more thing? set the fill on the wrapping element instead of the svg selector (it should get inherited by the svg)

I think that will allow the `fill` attribute on the `<svg>` to take preference, while still avoiding the default black fill that happens when no fill is specified.

mayank99: i think this means you should be able to revert most of the changes, right? the final diff should just have line 452 moved up a couple lines.

rohan-kadkol: I tried it before and also now. I think the changes are needed or make the code cleaner.
* If I remove [L465-467](https://github.com/iTwin/iTwinUI/pull/785/files#diff-3ac545d94bb9609fcd00db7227644a76d684ddad2b678eeaa75ac0d0209a251cR465-R467), then the status icons become grey
image

* And then I removed [L138](https://github.com/iTwin/iTwinUI/pull/785/files#diff-3ac545d94bb9609fcd00db7227644a76d684ddad2b678eeaa75ac0d0209a251cL138), since you suggested that, also because now `fill` on [L441](https://github.com/iTwin/iTwinUI/pull/785/files#diff-3ac545d94bb9609fcd00db7227644a76d684ddad2b678eeaa75ac0d0209a251cR441), directly uses the color and not the variable. i.e. the variable is no longer used for the start icon fill

mayank99: Are you sure? I undid all your changes and just have this one line diff. It works fine

rohan-kadkol: Ohh, thanks for confirming. You're right it works. I accidentally was using `fill: var(--iui-color-foreground-4);` on L449 instead of `var(--_iui-table-cell-icon-fill)`.

Updated the code to just include your above mentioned change.

rohan-kadkol: Actually, with the new code, looks like two tests failed: iTwinUI_table_Type_Expandable_Rows_0_demo-expandable-rows_0_desktop.png (Light and Dark theme)

image

Here the status icons are black instead of their status colors. This gets fixed with the code before the latest commit.

This is because in the latest commit, we only pass a fill as a property to the parent that will be inherited by the svg (very low specificity). `.iui-button-icon`'s black fill (`currentColor`) easily overrides it.

MicrosoftTeams-image (3)

In the previous commit, we pass the fill specifically on the svg of the end icon (higher specificity). This overrides the black fill of `.iui-button-icon`.

MicrosoftTeams-image (2)

gretanausedaite (commented):
Try undoing Dialog images. It's not showing any changes.

packages/itwinui-css/src/table/table.scss
mayank99: Is the `iui-button-icon` class necessary? There is no button there.

rohan-kadkol:
> Try undoing Dialog images. It's not showing any changes.

image

I approved them since they were very small pixel changes. I had mentioned this in the description to confirm with others.

After today's discussion, I guess it'd be better to approve these two image change here to avoid these tests failing in future PRs and causing confusion of what broke these tests by 1px. But I can undo these two image changes if you still feel so.

packages/itwinui-css/src/table/table.scss
mayank99: It's fine to leave it as is

mayank99 (approved):

adamhock (approved):
Looks good to me! 🎉

gretanausedaite (approved):

PR #787, opened 2022-10-21
fix(Dialog): Apply z-index on the wrapper

mayank99:
In itwinui-react 1.48.0, the `Modal` component uses `iui-dialog-wrapper`, which creates a stacking context and prevents children's z-index from being applied correctly.

So I've moved the z-index to the root. Still need to investigate if this is the correct solution, but I think it is.

After this change:


mayank99:
Superseded by https://github.com/iTwin/iTwinUI-react/pull/892.

mayank99:
Reopening this and targeting `v1` branch.

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -13,6 +13,8 @@ $iui-dialog-width: 380px;
   height: 100%;
   pointer-events: none;
   transform: translateX(0); // creates containing block for position: fixed
+  z-index: 999;
+  isolation: isolate;
```
rohan-kadkol, line 17: `z-index: 999` is the same as the React PR [#892](https://github.com/iTwin/iTwinUI-react/pull/892). However, L17 is new.

Just curious why is `isolation: isolate;` needed here but was not needed before? Or maybe this is just an enhancement and not a requirement?

mayank99: It creates a new stacking context to prevent the z-index from leaking out. I think it would be useful when there are nested or adjacent dialogs. I excluded it from react because I would consider it an enhancement.

You can read about it here: https://www.joshwcomeau.com/css/stacking-contexts/

rohan-kadkol (approved):

PR #788, opened 2022-10-24
fix(Button): Remove `margin-block`, change `line-height`, style `.iui-button-icon`'s child

rohan-kadkol:
<details>
<summary>Show more of comment (Only pertinent portion)</summary>

> I think we might want to do away with that `margin-block`. It was added to make the height of the small icon button match the height of the small text button.
> 
> We explicitly set the `min-height` to 24px, but since the line-height of text is already 24px and there's a 1px on each side, the total comes out to be 26px. To match this 24px text, we add 4px margin above and below the 16px icon.
> 
> We could fix this one of two ways (@FlyersPh9 might want to comment on this part):
> 
> 1. Set min-height to explicitly be 26px instead of 24px, OR
> 2. Decrease line-height so that small button actually becomes 24px tall instead of 26px.
> 
> As for React, I think our users are already having issues with `React.cloneElement`, so we might want to change the approach to _always_ have a wrapping span. The DOM structure could look something like this:
> 
> ```
> <button>
>   <span class='iui-button-icon'>
>     <svg>...</svg>
>   </span>
>   <span>Text</span>
> </button>
> ```
>
> ... (shortened) 
> 
> which shouldn't be a problem for us if we make sure our `iui-button-icon` handles svgs nested deeply like this:
> 
> ```css
> .iui-button-icon {
>   svg {
>     width: 16px;
>     height: 16px;
>     fill: currentColor;
>   }
> }
> ```

</details>

_Originally posted by @mayank99 in https://github.com/iTwin/iTwinUI/pull/768#discussion_r991647670_

---

* Partial code from PR [#768](https://github.com/iTwin/iTwinUI/pull/768/) moved here since they are impertinent to that PR
<br/>

* Remove `margin-block` from `.iui-button-icon`
* Change `line-height` to `1.2` from `var(--iui-size-l)`
* Style inner child of `.iui-button-icon` instead of the class itself
  * This is needed by React PR [#895](https://github.com/iTwin/iTwinUI-react/pull/895), since it removes `React.cloneElement()` from `Button` & `IconButton`.

### Checklist

- [x] Approved test images
- [x] Test/Implement in React: React PR [#895](https://github.com/iTwin/iTwinUI-react/pull/895)


packages/itwinui-css/backstop/tests/information-panel.html
```diff
@@ -1361,13 +1361,17 @@ <h1>Information panel</h1>
...
-                  xmlns="http://www.w3.org/2000/svg"
+                  aria-hidden="true"
                 >
-                  <path d="m0 2v12h16v-12zm11 11h-10v-10h10z" />
-                </svg>
+                  <svg
+                    viewBox="0 0 16 16"
+                    xmlns="http://www.w3.org/2000/svg"
+                  >
+                    <path d="m0 2v12h16v-12zm15 7h-14v-6h14z" />
```
rohan-kadkol, line 1372: I changed the path from [L1369](https://github.com/iTwin/iTwinUI/pull/788/files#diff-697a9ad7c898ef8688dea9c11a68c149c5d000f54c7113193a6e94d1faaaf9ccL1369) since I found a bug: The icon first shows the dock right icon. Then after clicking it, it still shows the dock right icon. After clicking it again, it shows the dock bottom icon.

Now, the icon starts off with the dock bottom icon.

Bug screenshot:

[Video1.webm](https://user-images.githubusercontent.com/45748283/198397265-7c26e779-b2e1-44f3-8584-c0ceac01448c.webm)

packages/itwinui-css/src/button/button.scss
```diff
@@ -33,10 +33,11 @@
 }
 
 @mixin iui-button-icon {
-  @include iui-icon-style('m');
-  transition: fill var(--iui-duration-1) ease-out;
-  fill: currentColor;
-  margin-block: var(--iui-size-2xs);
+  > * {
```
mayank99, line 36: I think this should target all svg descendants instead of just the direct child.

```suggestion
  svg {
```

This is necessary for notification marker to work in react with the following dom structure:

```html
<button>
  <span class='iui-button-icon'>
    <span class='iui-notification-marker'>
      <svg>...</svg>
```

rohan-kadkol: I had it as `svg` first, but then thought that since `.iui-button-icon` was styling let's say tag `X`, and we are now wrapping `X` with a span with `.iui-button-icon`, the CSS style should style `> *` to keep the styling the same.

For nested svgs like in the example you gave, I think the properties would be inherited down the subtree to the SVG. I tested it out in React and it seems to work:

image

mayank99: The styles tab will show all properties from every ancestor all the way up the tree if you scroll up. It is not an accurate representation of which styles are actually taking effect (Computed tab is more reliable for that). But I'm pretty sure `fill` is the only property out of all of those that gets inherited. The rest stay on the parent span.  The fact that it "works" is just a coincidence.

I'm not sure what you mean by "tag X" or how it would "break" if we target all descendant svgs. The `iui-button-icon` has one job: style svgs inside it.

rohan-kadkol: **[DONE]** Sounds good, yes, I think you're right. Not all attributes were inherited. So we need to style the svg directly instead of relying on inheritance. I implemented your suggested change.

> I'm not sure what you mean by "tag X"

By tag `X` I meant any arbitrary tag

> ... what you mean by "tag X" or how it would "break" if we target all descendant svgs

I didn't mean it would break anything but I felt that since we were not explicitly styling `svg` before, now also there is no need to explicitly style `svg`. That's why I styled `> *` instead of `svg`.

packages/itwinui-css/src/button/button.scss
```diff
@@ -11,7 +11,7 @@
   justify-content: center;
   position: relative;
   border-radius: var(--iui-border-radius-1);
-  line-height: var(--iui-size-l);
+  line-height: 1.2;
```
mayank99, line 14: One more thing before we merge: have you gone through all the images and tried to verify if all the changes look good? Some of them could be undesirable.

rohan-kadkol: **[DONE]** Yes, I have re-gone through all the test image changes. Most undesirable changes I found before were the other components targeting `.iui-button-icon` and not targeting the inner svg. That's why I made changes like [L117](https://github.com/iTwin/iTwinUI/pull/788/files#diff-187108cb5bf3f6ce0ad8a514769689db86ffe4102ee775050d0b4c02dda54bedR117) or [L130](https://github.com/iTwin/iTwinUI/pull/788/files#diff-30ad83636b1219b4049da82968e8c6e9e8ca2a8fa13072e432ef4a67b7b0c11aR130).

The other change was adding a wrapper span on [L648](https://github.com/iTwin/iTwinUI/pull/788/files#diff-0562194b7c80cbcdb6d2d8b799a8b3839b3c3e5e67da4287cb5a495fc2f18a36R648) and [L719](https://github.com/iTwin/iTwinUI/pull/788/files#diff-0562194b7c80cbcdb6d2d8b799a8b3839b3c3e5e67da4287cb5a495fc2f18a36R719) for uniformity.

Another point just to mention is that `line-height: 1` caused `68` test image changes, while the current `line-height: 1.2` causes `90` test image changes. I guess this does not matter since like we discussed, we should focus on what looks better and not what causes the least test changes. Thus, I went with `line-height: 1`.

mayank99 (approved):

FlyersPh9 (approved):

PR #797, opened 2022-11-02
chore: Fix `@charset` warning caused by copyright banner

mayank99:
Our build was causing this warning:
```
@charset must precede all other statements.
```

So I've modified the copyright linter to insert the comment *after* the `@charset` rule.


rohan-kadkol (approved):
Code LGTM. Tested `yarn build` to confirm that the warning has gone.

PR #800, opened 2022-11-08
feat(SideNavigation): Active stripe shows above `border-right` & `.iui-submenu-open` hides `border-right`

FlyersPh9:
- Active side navigation button stripe appears above `border-right`, similar to the active header button stripe appears above `border-bottom`.
- `.iui-submenu-open` has different visual styling, hides `border-right`.
- Added styling for when button is active AND submenu is also open.
- Both header & side navigation are using `2px` for border width instead of CSS var so border does not scale with text size.
- Header button uses `var(--iui-color-border-accent)` for active stripe instead of `currentColor` to align with button / side navigation button.

Screenshot 2022-11-08 at 1 52 16 PM
Screenshot 2022-11-08 at 1 52 45 PM
Screenshot 2022-11-09 at 3 31 44 PM


packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -118,8 +121,13 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
...
 
-    &.iui-submenu-open {
-      background-color: var(--iui-color-background-hover);
+    &.iui-submenu-open::after {
+      content: '';
+      position: absolute;
+      right: -2px;
+      width: 2px;
+      height: 100%;
+      background-color: var(--iui-color-background);
```
mayank99, line 130: This creates an empty space when hovered. I think we want the background to occupy that space.


also, have you thought about how you want it to look when it's active + submenu open? i think we might want to keep the right stripe for consistency.

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -86,11 +86,14 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
...
-    overflow: hidden;
     justify-content: flex-start;
 
     // Adds stripe to the right of active button
-    --_iui-button-active-stripe-inset: 0 0 0 calc(100% - #{var(--iui-size-3xs)});
+    &::after {
+      right: -2px;
+      width: 2px;
+      height: 100%;
+    }
```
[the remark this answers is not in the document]
FlyersPh9: Nice, much cleaner, updated.

packages/itwinui-css/src/side-navigation/side-navigation.scss
FlyersPh9: When hovered, the area where the border once was now matches the button's hover color.
Screenshot 2022-11-08 at 3 45 28 PM

When a tab is active AND submenu open, now looks like this:
Screenshot 2022-11-08 at 3 40 07 PM

packages/itwinui-css/src/side-navigation/side-navigation.scss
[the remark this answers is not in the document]
mayank99: Can you move it back to where it was and bring back the comment? Smaller diff + comment is useful because this code is not self-explanatory.

packages/itwinui-css/src/side-navigation/side-navigation.scss
mayank99: Any chance we can get rid of one of those two pseudo elements (`::before` and `::after`)? There is only one stripe so it doesn't make sense. I think we might want to improve the border handling instead.

packages/itwinui-css/src/side-navigation/side-navigation.scss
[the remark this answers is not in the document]
FlyersPh9: Yes, the deed is done.

packages/itwinui-css/src/side-navigation/side-navigation.scss
FlyersPh9: I've removed `::before` and am instead using a linear-gradient to achieve nearly the same visual result.  I did need to change `background-color` to `background` for the active button stripe to make this work.

mayank99: Are you sure you need to change `background-color` to `background` in `button.scss`? I changed it back and it seems to work fine.

Also, transitioning `background` is not recommended for performance reasons, and it also leads to undesirable results in lots of cases.

mayank99: try this:

```css
    &.iui-submenu-open {
      &::before {
        content: '';
        z-index: -1;
        position: absolute;
        inset: 0 -2px 0 0;
        background-color: inherit;
      }
    }
```

mayank99: try this:

```css
    &.iui-submenu-open {
      --_iui-button-active-stripe-color: linear-gradient(
        0deg,
        var(--iui-color-background-accent-muted) 25%,
        var(--iui-color-border-accent) 25%,
        var(--iui-color-border-accent) 75%,
        var(--iui-color-background-accent-muted) 75%
      );

      &::before {
        content: '';
        position: absolute;
        inset: 0 -2px 0 0;
        z-index: -1;
        background-color: inherit;
      }
    }
```

also i think `--_iui-button-active-stripe-color` should be renamed to `--_iui-button-active-stripe-background`.

elephantcatdog:
> image

If there's a special case for this scenario, there should probably be a test image for it as well.

packages/itwinui-css/src/button/button.scss
```diff
@@ -51,7 +51,7 @@
     content: '';
     position: absolute;
     inset: var(--_iui-button-active-stripe-inset);
-    background-color: var(--_iui-button-active-stripe-color);
+    background: var(--_iui-button-active-stripe-color);
```
elephantcatdog, line 54: Why did you change this to `background`?

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -119,7 +119,29 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
...
+
+      &[data-iui-active='true'] {
+        --_iui-button-active-stripe-color: linear-gradient(
+          0deg,
+          var(--iui-color-background-accent-muted) 25%,
+          var(--iui-color-border-accent) 25%,
+          var(--iui-color-border-accent) 75%,
+          var(--iui-color-background-accent-muted) 75%
+        );
+      }
```
elephantcatdog, line 144: I really like the visual changes for moving the stripe over for active tabs and removing the stripe and matching the color for the open tabs. It looks good and is a nice improvement.

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -86,11 +86,11 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
     border-radius: 0;
     border-left: none;
     border-right: none;
-    overflow: hidden;
+    isolation: isolate;
```
elephantcatdog, line 89: Why did you remove `overflow: hidden`?

FlyersPh9: If `overflow: hidden` was still applied, the blue stripe would not show because technically the stripe is positioned -2px outside of the button on the right hand side.

packages/itwinui-css/src/side-navigation/side-navigation.scss
FlyersPh9: Using your suggested code and renamed CSS variable.  Did have to leave the change in button to use `background`, `background-color` wouldn't accept a `linear-gradient`.

FlyersPh9:
> If there's a special case for this scenario, there should probably be a test image for it as well.

@elephantcatdog I've added a new test for this scenario.

mayank99 (approved):

elephantcatdog (approved):
Looks good to me

PR #802, opened 2022-11-09
chore: Update visual testing

elephantcatdog:
Image testing now includes high contrast dark and high contrast light. The naming conventions for themes was updated to apply the corresponding theme to the end of each test's label. This means that 1) light theme visual tests now includes "light" in the test image names and 2) images of the same scenario but with different themes will now be adjacent to each other when test images are sorted alphabetically (as they normally are). The second change should make it easier to review PRs with test image changes.

In case git doesn't recognize that I only renamed and added images and didn't delete any, I ran the tests for just the renamed light and dark theme images:
image
for all themes before appoving the high contrast images:
image
and for all themes after approving the high contrast images:
image

Code changes can be found in the first and last files in the 'Files changed' tab: onReady.js and index.js


elephantcatdog:
Merged in https://github.com/iTwin/iTwinUI/pull/822

PR #803, opened 2022-11-09
feat(Alert): Use data attributes

FlyersPh9:
- **Use `class="iui-alert" data-iui-status="informational"`.** In #623 we changed from using `.iui-alert .iui-informational` to `.iui-alert-informational` to achieve lower specificity.  We want to keep the base `.iui-alert` for targeting regardless of status.
- **Use `class="iui-alert" data-iui-variant="sticky"`**.

Specificity is under 0, 3, 0 & `alert.css` went from 406 lines to 310.

### DIFF
--- a/packages/itwinui-css/src/alert/alert.scss
+++ b/packages/itwinui-css/src/alert/alert.scss
@@ -39,7 +39,6 @@
 }
 
 @mixin iui-alert-link {
-  @include iui-anchor-status;
   @include iui-anchor-underline('on-initial');
   border-radius: var(--iui-border-radius-1);
   cursor: pointer;
--- a/packages/itwinui-css/src/alert/alert.scss
+++ b/packages/itwinui-css/src/alert/alert.scss
@@ -29,6 +29,9 @@
   @include iui-icon-style('m');
   margin-left: var(--iui-size-m);
   fill: var(--_iui-alert-icon-color, var(--iui-color-icon-muted));
+  @media (forced-colors: active) {
+    fill: CanvasText;
+  }
 }
 
 @mixin iui-alert-message {
--- a/packages/itwinui-css/src/alert/alert.scss
+++ b/packages/itwinui-css/src/alert/alert.scss
@@ -9,28 +9,27 @@
 
   --_iui-alert-border-color: var(--iui-color-border);
 
-  border-radius: var(--iui-border-radius-1);
+  border-radius: var(--_iui-alert-border-radius, var(--iui-border-radius-1));
   display: flex;
   align-items: center;
   color: var(--iui-color-text);
   background-color: var(--iui-color-background);
   border: 1px solid var(--_iui-alert-border-color);
   box-shadow: inset var(--iui-size-2xs) 0 0 var(--_iui-alert-border-color);
-  @media (forced-colors: active) {
-    background-color: Canvas;
-  }
+}
 
-  &.iui-sticky {
-    border-radius: 0;
-    position: sticky;
-    top: 0;
-    left: 0;
-  }
+@mixin iui-alert-sticky {
+  --_iui-alert-border-radius: 0;
+  position: sticky;
+  top: 0;
+  left: 0;
+  right: 0;
 }
 
 @mixin iui-alert-icon {
   @include iui-icon-style('m');
   margin-left: var(--iui-size-m);
+  fill: var(--_iui-alert-icon-color, var(--iui-color-icon-muted));
 }
 
 @mixin iui-alert-message {
@@ -38,6 +37,7 @@
 }
 
 @mixin iui-alert-link {
+  @include iui-anchor-status;
   @include iui-anchor-underline('on-initial');
   border-radius: var(--iui-border-radius-1);
   cursor: pointer;
@@ -62,16 +62,10 @@
 /// @arg $statusColor Can be one of: informational, positive, warning, negative
 /// @arg $rootSelector Selector of the root alert element. Defaults to .iui-alert
 @mixin iui-alert-category($statusColor, $rootSelector: '.iui-alert') {
-  @include iui-alert;
   --_iui-alert-border-color: var(--iui-color-border-#{$statusColor});
+  --_iui-alert-icon-color: var(--iui-color-icon-#{$statusColor});
 
-  #{$rootSelector}-icon {
-    fill: var(--iui-color-icon-#{$statusColor});
-  }
-
-  #{$rootSelector}-message {
-    @include iui-text-selection($statusColor);
-  }
+  @include iui-text-selection($statusColor);
 
   #{$rootSelector}-link {
     @include iui-anchor-status($statusColor);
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/src/alert/alert.scss
```diff
@@ -9,35 +9,35 @@
 
   --_iui-alert-border-color: var(--iui-color-border);
 
-  border-radius: var(--iui-border-radius-1);
+  border-radius: var(--_iui-alert-border-radius, var(--iui-border-radius-1));
```
adamhock, line 12: Can you explain the reason for this change?

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -9,35 +9,35 @@
...
   display: flex;
   align-items: center;
   color: var(--iui-color-text);
   background-color: var(--iui-color-background);
   border: 1px solid var(--_iui-alert-border-color);
   box-shadow: inset var(--iui-size-2xs) 0 0 var(--_iui-alert-border-color);
-  @media (forced-colors: active) {
```
adamhock, line 19: Do we not need forced colors anymore?

adamhock (commented):
Code looks good to me off first glance

packages/itwinui-css/src/alert/alert.scss
FlyersPh9: Using CSS vars versus overriding.  Decided it's unnecessary for the point of this PR and reverted.

packages/itwinui-css/src/alert/alert.scss
FlyersPh9: `@media (forced-colors: active) { background-color: Canvas; }` was needed when we [set the background color using this gradient hack](https://github.com/iTwin/iTwinUI/blob/3b9b327ed8cf84db87dc1d61d356445cd5b051ba/packages/itwinui-css/src/alert/alert.scss#L58).  Now that we use `background-color` the `background-color: Canvas` isn't necessary. 
Screenshot 2022-11-09 at 2 56 59 PM

elephantcatdog (commented):
There's a lot of things that were changed that seem like they aren't related to the description - adding back .iui-alert and adding data attributes for alert type?

packages/itwinui-css/src/alert/alert.scss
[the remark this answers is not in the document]
elephantcatdog: I see your explanation above.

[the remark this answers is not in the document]
FlyersPh9: That's answered [here](#discussion_r1018359852).

packages/itwinui-css/src/alert/classes.scss
```diff
@@ -4,18 +4,22 @@
...
 .iui-alert {
   @include iui-alert;
+
+  @each $status in informational, positive, warning, negative {
+    &:where([data-iui-variant='#{$status}']) {
+      @include iui-alert-category($statusColor: $status);
+    }
+  }
+
+  &:where([data-iui-sticky='true']) {
```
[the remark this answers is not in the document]
FlyersPh9: In `main` branch we are using `.iui-alert .iui-informational .iui-sticky`.
In `v1` branch we are using `.iui-alert-informational .iui-sticky`
This PR I figured if we're changing from `.iui-informational` to `data-iui-variant="informational"`, now is the time to also switch over `.iui-sticky` to use a data attribute.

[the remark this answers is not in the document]
elephantcatdog: Hm... I guess my question is more, why should .iui-sticky be a data attribute? I thought we were mostly using data attributes for things like size and color and type, where there can only be one of a kind. But for .iui-sticky, an alert is either sticky or not. Are we using data attributes in other true/false situations?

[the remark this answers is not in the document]
elephantcatdog: If you're preparing for other variants for `Alert` would `data-iui-variant="sticky"` be a good solution? And call `data-iui-variant` -> `data-iui-status` instead?

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -16,28 +16,30 @@
...
+    fill: CanvasText;
+  }
 }
 
 @mixin iui-alert-message {
   margin: var(--iui-size-s) var(--iui-size-m);
 }
 
 @mixin iui-alert-link {
+  @include iui-anchor-status;
```
rohan-kadkol, line 42: I think this may not be needed. Because all the six examples in alert.html had a variant provided. And when a variant is provided, its styling is handled in [L72-74](https://github.com/iTwin/iTwinUI/blob/28421289ed0fa9f3dedd3549a5a4275f8098f01f/packages/itwinui-css/src/alert/alert.scss#L72-L74).

Also, removing this line still passed the tests.

FlyersPh9: Removed.

packages/itwinui-css/src/alert/classes.scss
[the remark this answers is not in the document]
FlyersPh9: I think your suggestion makes sense.  Changed it to `data-iui-status="informational" data-iui-variant="sticky"`.

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -16,21 +16,22 @@
...
+  border-radius: 0;
+  position: sticky;
+  top: 0;
+  left: 0;
 }
 
 @mixin iui-alert-icon {
   @include iui-icon-style('m');
   margin-left: var(--iui-size-m);
+  fill: var(--_iui-alert-icon-color, var(--iui-color-icon-muted));
```
elephantcatdog, line 31: What does this `var(x, y)` do?

adamhock (approved):

packages/itwinui-css/src/alert/alert.scss
FlyersPh9: `y` becomes the default value for `x`.  So this example does `--_iui-alert-icon-color: var(--iui-color-icon-muted)`.

rohan-kadkol (approved):

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -62,16 +64,10 @@
 /// @arg $statusColor Can be one of: informational, positive, warning, negative
 /// @arg $rootSelector Selector of the root alert element. Defaults to .iui-alert
 @mixin iui-alert-category($statusColor, $rootSelector: '.iui-alert') {
-  @include iui-alert;
   --_iui-alert-border-color: var(--iui-color-border-#{$statusColor});
+  --_iui-alert-icon-color: var(--iui-color-icon-#{$statusColor});
 
-  #{$rootSelector}-icon {
-    fill: var(--iui-color-icon-#{$statusColor});
-  }
```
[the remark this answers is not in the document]
FlyersPh9: Correct, it keeps the nesting 1 lower.

packages/itwinui-css/src/alert/alert.scss
elephantcatdog: Ah, neat

elephantcatdog (approved):

PR #816, opened 2022-11-15
chore(createComponent): Targets `all.scss` instead of `classes.scss`

### LINKED ISSUE #776: `yarn createComponent` is not working as expected in v1 branch
Needs to add `@import` to `/packages/itwinui-css/all.scss`. Instead it is adding somewhere else.

adamhock:
- `yarn createComponent` now appends the component's classes to `all.scss` instead of `src/classes.scss`. 
- `yarn createComponent` no longer appends the component's index to `src/index.scss`.
- Removed excess slashes

Closes #776

### DIFF
--- a/packages/itwinui-css/scripts/createComponent.js
+++ b/packages/itwinui-css/scripts/createComponent.js
@@ -50,14 +50,6 @@ const componentIndexFactory = (directory) => {
   };
 };
 
-/** Appends component index to global index.scss */
-const globalIndexFactory = (directory) => {
-  return {
-    path: `${directory}/index.scss`,
-    template: `@import './${componentName}/index';\n`,
-  };
-};
-
 /** Creates classes.scss under component directory */
 const componentClassesFactory = (directory) => {
   const template = `${copyrightBannerScss}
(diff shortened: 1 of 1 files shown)


mayank99:
PR title should not say `fix` because this is not a user-facing change. Adding `feat` or `fix` triggers the release bot and auto adds the the title to the changelog, which is not what we want for maintenance and docs-related fixes.

packages/itwinui-css/scripts/createComponent.js
```diff
@@ -50,14 +50,6 @@ const componentIndexFactory = (directory) => {
   };
 };
 
-/** Appends component index to global index.scss */
-const globalIndexFactory = (directory) => {
```
gretanausedaite, line 54: We don't have global index anymore, right?

gretanausedaite (approved):
Looks good to me!

packages/itwinui-css/scripts/createComponent.js
adamhock: Right

FlyersPh9 (approved):

PR #817, opened 2022-11-16
feat(ComboBox): Add multi-select tags

gretanausedaite:
Continuing from #767 

Needed for react implementation of combobox multiselect.


adamhock (approved):

packages/itwinui-css/backstop/tests/select.html
```diff
@@ -99,6 +99,38 @@ <h2>Default</h2>
         </div>
         <div class="iui-message">Help message.</div>
       </label>
+      <label class="iui-input-container iui-with-message">
+        <div class="iui-label">ComboBox multiselect label</div>
+        <div class="iui-input-with-icon">
+          <input
+            class="iui-input"
+            placeholder="Placeholder…"
+            style="padding-left: min(414px, calc(70% + 20px))"
```
rohan-kadkol, line 108: I'm just wondering, how would we set this padding in React? Like, how would we know the width of `.iui-select-tag-container`?

packages/itwinui-css/backstop/tests/select.html
```diff
@@ -99,6 +99,38 @@ <h2>Default</h2>
...
+          />
+          <svg-caret-down-small
+            class="
+            iui-end-icon
+            iui-actionable"
```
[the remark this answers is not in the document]
gretanausedaite: Good catch!

packages/itwinui-css/backstop/tests/select.html
gretanausedaite: In react I'm using `useContainerWidth` hook:
image

Adding ref object to multiple tags container:
image

And using width object for input padding:
image

rohan-kadkol (approved):

PR #818, opened 2022-11-17
feat(Avatar): Sizes have lower specificity, use CSS vars.

### LINKED ISSUE #631: Tile: avatar in thumbnail displays differently depending on import order
#### Actual behavior

When user-icon is used as a thumbnail for tile, the size of it changes depending on the import order. `tile.css` and `user-icon.css` both try to specify the size with a 0-2-0 specificity, which means the one that comes later will win.

screenshot of devtools showing the thumbnail css being overridden by user-icon size, resulting in 48x48 icon instead of 64x64

#### Expected behavior

The user-icon should either always display as 64x64 when used in tile, or it should always depend on the modifier class. The order of imports should not change the behavior, because we have no control over that in iTwinUI-react.

#### Additional information

If the goal is to display it as 64x64, then this inconsistency can be worked around by always using the `x-large` size of user-icon.

FlyersPh9:
- Reduces specificity of the size modifier classes.
- When avatar is nested within tile, the tile defines the size of the avatar.  Closes #631.
- Deleted `avatar-sizes.scss`, moved that mixin into `avatar.scss`.


gretanausedaite (approved):
Looks good to me.

adamhock (approved):
Things look good to me! Just make sure to update your branch before merging

PR #821, opened 2022-11-21
feat(ExpandableBlock): Add disabled state

FlyersPh9:
- Reduced CSS specificity from `0, 4, 0` to `0, 3, 0` without introducing breaking changes.  Specificity could be lowered more with breaking changes.
- Add `disabled` state.
- `forced-colors` support.
- `.iui-header` now supports `<button>` type, & `<div>` should have `role="button"` added.  [Source.](https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion)
- `:focus` on default styling now has 2px thickness, borderless stays at 1px.
- Slightly improve code structure (example: use `gap` instead of `margin` to separate text and icons)


packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+  display: flex;
+  gap: var(--_iui-expandable-block-header-gap);
+  align-items: center;
+  padding: var(--_iui-expandable-block-header-padding);
+  cursor: pointer;
+  user-select: none;
+  color: var(--iui-color-text);
+  line-height: inherit;
+  font-family: inherit;
+  text-align: left;
```
FlyersPh9, line 79: `@include iui-reset;`, `text-align`, `font-family`, & `line-height` were added for `<button>` support.

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+    --_iui-expandable-block-border: 1px solid var(--iui-color-border);
+    --_iui-expandable-block-background-color: var(--iui-color-background);
+    --_iui-expandable-block-content-padding: var(--iui-size-s);
+    --_iui-expandable-block-header-background-color-disabled: var(--iui-color-background-disabled);
+    --_iui-expandable-block-header-background-color-hover: var(--iui-color-background-hover);
+    --_iui-expandable-block-header-focus-offset: -2px;
+    --_iui-expandable-block-header-focus-thickness: 2px;
+
+    &:where(.iui-expanded) {
+      --_iui-expandable-block-header-bottom-border-radii: 0;
```
gretanausedaite, line 32: Why `radii`, not `radius`?

FlyersPh9: That variable is setting both bottom left & bottom right border radius, so I thought it would be appropriate.  I'm fine with changing it though, just let me know if you'd like it changed.

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+/// @arg $size Can be one of: default, small
+@mixin iui-expandable-block-size($size: 'default') {
+  @if $size == 'default' {
+    --_iui-expandable-block-expander-icon-size: var(--iui-size-m);
+    --_iui-expandable-block-header-gap: var(--iui-size-s);
+    --_iui-expandable-block-header-padding: var(--iui-size-s);
+  }
+  @if $size == 'small' {
+    --_iui-expandable-block-expander-icon-size: var(--iui-size-s);
+    --_iui-expandable-block-header-gap: var(--iui-size-xs);
```
rohan-kadkol, line 59: Earlier, the padding for a small block's header was `calc(var(--iui-size-s) * 0.5) var(--iui-size-xs)`. To confirm, was this a wanted change?

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+
+/// @arg $size Can be one of: default, small
+@mixin iui-expandable-block-size($size: 'default') {
+  @if $size == 'default' {
+    --_iui-expandable-block-expander-icon-size: var(--iui-size-m);
+    --_iui-expandable-block-header-gap: var(--iui-size-s);
+    --_iui-expandable-block-header-padding: var(--iui-size-s);
+  }
+  @if $size == 'small' {
+    --_iui-expandable-block-expander-icon-size: var(--iui-size-s);
```
rohan-kadkol, line 58: Previously, this was done by `@include iui-icon-style('s')` ([L123](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/src/expandable-block/block.scss#L123)). To confirm, was this a wanted change?

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+}
+
+@mixin iui-expandable-block-header {
+  @include iui-reset;
+  @include iui-focus(
+    $offset: var(--_iui-expandable-block-header-focus-offset),
+    $thickness: var(--_iui-expandable-block-header-focus-thickness)
+  );
+  display: flex;
+  gap: var(--_iui-expandable-block-header-gap);
```
rohan-kadkol, line 71: Not sure, but I can't find `gap` on  property in the old code. Is this alright?

To confirm, is this PR trying to change the styling too apart from just adding the diabled state? If so, then maybe the missing 1-1 mappings of old-new code might be alright.

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  transition: color var(--iui-duration-1) ease;
+  @media (forced-colors: active) {
+    transition: none;
+  }
+
+  @if $type == 'sublabel' {
+    font-size: var(--iui-font-size-0);
+    color: var(--_iui-expandable-block-header-caption-color, var(--iui-color-text-muted));
```
rohan-kadkol, line 168: `--_iui-expandable-block-header-caption-color` is not found. So the forced colors don't get applied. In the screenshot, the variable is greyed out and not blue.

image

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+@mixin iui-expandable-block-header-label {
+  display: flex;
+  flex-direction: column;
+  flex: auto;
+  min-width: 0;
+  font-size: var(--iui-font-size-2);
+  color: var(--_iui-expandable-block-header-label-color, inherit);
```
rohan-kadkol, line 153: `--_iui-expandable-block-header-label-color` not found. So forced colors don't get applied

image

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+    --_iui-expandable-block-expander-icon-fill: var(--iui-color-icon-hover);
+    --_iui-expandable-block-header-caption-color: var(--iui-color-text-hover);
+    --_iui-expandable-block-header-label-color: var(--iui-color-text-hover);
+
+    background-color: var(--_iui-expandable-block-header-background-color-hover);
+
+    @media (forced-colors: active) {
+      --_iui-expandable-block-header-caption-color: Highlight;
+      --_iui-expandable-block-header-label-color: Highlight;
+      --_iui-expandable-block-expander-icon-fill: Highlight;
```
rohan-kadkol, line 100: nit: The order of these 3 variables can be the same as on L91-93, makes it easier to read

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,187 @@
...
+      border-color: Highlight;
```
rohan-kadkol, line 101: I might be wrong, but should we update `--iui-color-border` since border is set on [L81](https://github.com/iTwin/iTwinUI/pull/821/files#diff-7139d3558365a97d23934152a27c2e4c8faba3aaeb01912529f15c90f764b7a2R81) that includes the color?

packages/itwinui-css/src/expandable-block/expandable-block.scss
gretanausedaite: It's just a bit strange. CSS has 'radius' for all (eg. border-radius: 10% 30% 50% 70%). 
If it's not already in variables package, I think "radius" would fit better.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Done.

packages/itwinui-css/backstop/tests/expandable-block.html
```diff
@@ -183,9 +220,10 @@ <h2>Small</h2>
 
       <div class="iui-expandable-block iui-expanded iui-small">
         <div
-          aria-expanded="true"
           class="iui-header"
           tabindex="0"
+          role="button"
+          aria-expanded="true"
```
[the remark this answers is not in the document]
FlyersPh9: Done.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Changed to `radius`.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Generally we don't override variables from iTwinUI-Variables, instead we override component specific variables (theoretical example: `--_iui-expandable-block-header-border-color`).  I could introduce a component variable instead if you'd prefer.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Changed line 61, that was a mistake on my part.

```diff
- --_iui-expandable-block-header-padding: 6px var(--iui-size-xs);
+ --_iui-expandable-block-header-padding: calc(var(--iui-size-s) * 0.5) var(--iui-size-xs);
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
rohan-kadkol: Ohh, I didn't realize that was a variables package var and not a local var. In that case, whatever you prefer would be good. I just wanted to confirm that since this PR is trying to add local vars for everything, maybe this could also be a variable.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: This was intentional.  We set the size of expandable block on the parent element.  If we wanted to use `@include iui-icon-style('s')` we would have to target it something like this:
```scss
.iui-expandable-block.iui-small .iui-header .iui-icon { @include iui-icon-style('s') }
```

If I could introduce breaking changes, I could reduce that nesting by giving the icon a more explicit class name like so:
```scss
.iui-expandable-block.iui-small .iui-expandable-block-icon { @include iui-icon-style('s') }
```

This new way, I am able to keep the nesting lower by setting the width / height to a CSS variable `var(--_iui-expandable-block-expander-icon-size)` and then changing that value on the parent.  Something like this:
```scss
.iui-icon {
	width: var(--_iui-expandable-block-expander-icon-size);
	height: var(--_iui-expandable-block-expander-icon-size);
}

.iui-expandable-block.iui-small {
	--_iui-expandable-block-expander-icon-size: 12px;
}
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: You are correct, we were using `margin` on the nested classes instead of `gap`.  This was done because we were still supporting Internet Explorer and [IE does not support `gap`](https://caniuse.com/?search=gap).  Now we have dropped IE support so I'd like to use `gap` instead of the `margin` hack.

If you are ok with me keeping it I would prefer to, but if you'd rather we address that in a separate PR I am ok with that too.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: I _think_ this is working as expected.  I am both introducing and setting the default value for that new variable in the same line.  This:
```scss
color: var(--my-new-variable, blue);
```
Is the same as this:
```scss
--my-new-variable: blue;
color: var(--my-new-variable);
```

When I turn on `forced-colors` mode, I do see the hover & disabled effects taking place (in order of default state, hovered state, disabled state):
Screenshot 2022-11-23 at 12 28 11 PM

I'll give you a call just to confirm I'm not misunderstanding.

rohan-kadkol: >  I am both introducing and setting the default value for that new variable in the same line

Ohh, wow, this is a cool trick, didn't know this before. In that case, I think the missing variable indication in the inspector makes sense. Thanks for explaining this.

packages/itwinui-css/src/expandable-block/expandable-block.scss
rohan-kadkol: Sounds good, was just confirming, thanks. Since this PR is also improving the code structure, I think adding this change in this PR would be good.

Maybe we can also add that this PR is "improving the code structure" to the PR description.

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -0,0 +1,188 @@
...
+    transform: var(--_iui-expandable-block-expander-icon-transform, none);
+    transition: fill var(--iui-duration-1) ease-out;
+    @media (prefers-reduced-motion: no-preference) {
+      transition: fill var(--iui-duration-1) ease-out, transform var(--iui-duration-1) ease-out;
+    }
+  }
+
+  @if $type == 'status' {
+    @media (forced-colors: active) {
+      fill: CanvasText;
```
rohan-kadkol, line 143: In table, the status icons have color in the forced color mode too.

image

But here, I think they are not colored. 

image

Not sure which one is correct, but maybe both examples should follow the same status colors in forced colors mode.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Description updated.

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9: Removed so the status color shows in `forced-colors`, matching what we have in table.

packages/itwinui-css/backstop/tests/expandable-block.html
```diff
@@ -46,7 +46,7 @@ <h1>Expandable block</h1>
     <h2>Default</h2>
     <section id="demo-default">
       <div class="iui-expandable-block">
-        <div
+        <button
```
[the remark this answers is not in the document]
FlyersPh9: `.iui-header` can remain being a `<div>` with the addition of `role="button"` or it can be a `<button>`, following [this accessibility guide](https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion).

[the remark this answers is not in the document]
rohan-kadkol: Ohh, sounds good, just confirming, since the other examples in the file used `div` with `role="button"`.

rohan-kadkol (approved):

gretanausedaite (approved):

PR #822, opened 2022-11-21
chore: Update visual testing

elephantcatdog:
Image testing now includes high contrast dark and high contrast light. The naming conventions for themes was updated to apply the corresponding theme to the end of each test's label. This means that 1) light theme visual tests now includes "light" in the test image names and 2) images of the same scenario but with different themes will now be adjacent to each other when test images are sorted alphabetically (as they normally are). The second change should make it easier to review PRs with test image changes.

In case git doesn't recognize that I only renamed and added images and didn't delete any, I ran the tests for just the renamed light and dark theme images:
image
for all themes before appoving the high contrast images:
image
and for all themes after approving the high contrast images:
image

Code changes can be found in the first and last files in the 'Files changed' tab: onReady.js and index.js


FlyersPh9 (approved):

adamhock (approved):
Nice! :tada:

packages/itwinui-css/backstop/scenarios/index.js
```diff
@@ -15,10 +15,28 @@ fs.readdirSync('backstop/scenarios/').forEach((file) => {
   scenarios = scenarios.concat(newScenarios);
 });
 
+const lightScenarios = scenarios.map((s) => {
+  const lightScenario = Object.assign({}, s);
```
rohan-kadkol, line 19: Just curious, the spread operator does not work here to create a clone?

packages/itwinui-css/backstop/engine_scripts/puppet/onReady.js
```diff
@@ -2,9 +2,14 @@ module.exports = async (page, scenario, vp) => {
   console.log('SCENARIO > ' + scenario.label);
   await require('./overrideCss')(page, scenario);
 
-  if (scenario.label.startsWith('dark_')) {
+  if (scenario.label.endsWith('_light')) {
+    await page.evaluate(() => (document.documentElement.dataset.iuiTheme = 'light'));
+  } else if (scenario.label.endsWith('_dark')) {
     await page.evaluate(() => (document.documentElement.dataset.iuiTheme = 'dark'));
   }
+  if (scenario.label.includes('_hc_')) {
```
rohan-kadkol, line 10: To make it safer, you could check of `_hc_light` or `_hc_dark` to avoid catching `_hc_` that's part of the real label
```suggestion
  if (scenario.label.endsWith('_hc_light') || scenario.label.endsWith('_hc_dark')) {
```

packages/itwinui-css/backstop/scenarios/index.js
elephantcatdog: I'm not sure. I copied this from darkScenarios, which was already here.

packages/itwinui-css/backstop/engine_scripts/puppet/onReady.js
elephantcatdog: _hc_ should not be part of the label because everything with underscores is something we add to the name.

rohan-kadkol: I'm slightly confused by what you meant. But to confirm, should it be `scenario.label.endsWith` or `scenario.label.includes`? [L10](https://github.com/iTwin/iTwinUI/pull/822/commits/e3298df92fdc7b99b28b3137792da6e0b7f2c42c#)

Since I think the labels always have the theme information at the end, maybe `endsWith` is a better check.

elephantcatdog: Oh, yeah, I can update that.

elephantcatdog: Is it good now?

rohan-kadkol: Yes, thanks

rohan-kadkol (approved):

PR #823, opened 2022-11-22
chore: Fixing release automation

### LINKED ISSUE #815: (maintenance) Release automation broken
After merging PR from release-please, the `CD` action fails with this error:

```
> npm publish --access public --tag 
npm ERR! Tag name must not be a valid SemVer range: 
```

Looks like the "tag" is not being correctly set to `latest` after https://github.com/iTwin/iTwinUI/pull/766

adamhock:
Link to successful job in which I echo the `inputs.tag` https://github.com/iTwin/iTwinUI/actions/runs/3532304714/jobs/5926494084

Closes #815


gretanausedaite (approved):
Should be good to go.

FlyersPh9 (approved):
I'm probably not the best person to review this, but from what I can see seems good.

elephantcatdog (approved):
Seems ok. Maybe you should tag Mayank to review though, as he made the issue?

.github/workflows/publish-npm.yml
```diff
@@ -26,6 +26,10 @@ on:
         required: true
         type: string
         default: itwinui-css
+      tag:
+        required: true
```
mayank99, line 30: If this was `required: false`, then I think you wouldn't need to pass the `tag` in release-please.yml because it already has a default value.

But since this is already tested working, I'm fine with it.

mayank99 (approved):

Issue #824, opened 2022-11-25
Slider: change align-items to flex-start


gretanausedaite:
User got warning after upgrading to v1.

![image](https://user-images.githubusercontent.com/83585998/203945270-88b0b20a-1bae-4630-9bb5-b528defd0055.png)

PR #825, opened 2022-11-25
fix(Slider): Change align `start` to `flex-start`

gretanausedaite:
Solves #824


FlyersPh9 (approved):

packages/itwinui-css/backstop/tests/non-ideal-state.html
```diff
@@ -30,7 +30,7 @@
 
       .demo-non-ideal-state-scrolling {
         height: 100vh;
-        scroll-snap-align: start;
+        scroll-snap-align: flex-start;
```
rohan-kadkol, line 33: flex-start didn't work here when I tested it. Maybe can replace with `center` ([Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-align#syntax)).
```suggestion
        scroll-snap-align: center;
```
Also, to confirm, the warning in [#824](https://github.com/iTwin/iTwinUI/issues/824) is for slider.css, so maybe this change is unneeded.

gretanausedaite: I can undo this change.

mayank99:
This PR will only get rid of the warning in one component. Have you gone through all instances of align-items: start​ and justify-content: start​ and even end​ in all components?
________________________________

rohan-kadkol:
> I could update align/justify `end`.

Not sure, but I think `start` and `end` are recommended to be replaced with `flex-start` and `flex-end` respectively, when using `display: flex`.

But for `scroll-snap-align`, I think `start`/`end` only work, and so don't need any replacements. So I don't think the mentioned warning was for `scroll-snap-align`.

mayank99:
> > I could update align/justify `end`.
> 
> Not sure, but I think `start` and `end` are recommended to be replaced with `flex-start` and `flex-end` respectively, when using `display: flex`.
> 
> But for `scroll-snap-align`, I think `start`/`end` only work, and so don't need any replacements. So I don't think the mentioned warning was for `scroll-snap-align`.

Yes, this is correct. Safari does not support `start` and `end` values for `justify-` and `align-` properties when using `display: flex` only (grid works fine). I went through all instances and didn't find any more that need to be changed (either they were already using `flex-start`/`flex-end` or they were using `end` with grid).

mayank99 (approved):

rohan-kadkol (approved):

PR #827, opened 2022-11-28
fix: Progress overlay should intercept clicks

### LINKED ISSUE #826: Overlay allows to click behind
You can select the text behind overlay. Also if there was button, its possible to click it, which should not be allowed.

Screenshot 2022-11-25 at 12 42 54

mayank99:
Main code change: removed `pointer-events: none`. Not sure why we had it before.

Also:
- removed tabindex and focus rule from overlay
- added `inert` attribute to content to fix focus, with polyfill for firefox
- moved `iui-overlay-exiting` class to `classes.scss` (unrelated change, because it was confusing me when reading the mixin)

Fixes #826

### DIFF
--- a/packages/itwinui-css/backstop/tests/progress-indicator.html
+++ b/packages/itwinui-css/backstop/tests/progress-indicator.html
@@ -1050,8 +1050,7 @@
           magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
           consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
           pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id
-          est
-          laborum.
+          est laborum.
         </div>
         <div class="iui-progress-indicator-overlay">
           <!-- Linear indeterminate with label -->
@@ -1075,8 +1074,7 @@
           magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
           consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
           pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id
-          est
-          laborum.
+          est laborum.
         </div>
 
         <div class="iui-progress-indicator-overlay">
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/backstop/tests/progress-indicator.html
```diff
@@ -1039,17 +1041,18 @@ <h3>Indeterminate</h3>
...
-        consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
-        pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est
-        laborum.
-
+      <div class="demo-overlay-wrapper">
         <div
-          class="iui-progress-indicator-overlay"
-          tabindex="0"
+          class="demo-overlay-content"
+          inert
```
gretanausedaite, line 1047: We will be adding `inert` to content under overlay in react implementation?

gretanausedaite (approved):
Looks good.

packages/itwinui-css/backstop/tests/progress-indicator.html
mayank99: I'm not sure. I think we could ask the user to do it or offer an `<Inert>` utility component that imports the polyfill? Doesn't need to be decided right now.

packages/itwinui-css/src/progress-indicator/overlay.scss
```diff
@@ -12,33 +12,27 @@
...
   }
 
-  .iui-overlay-exiting {
-    animation: closeAnimation var(--iui-duration-1) linear;
+  to {
+    opacity: 0;
   }
 }
+
+@mixin iui-progress-overlay-exiting {
```
rohan-kadkol, line 36: Just curious, why did you make this class into a plugin? Just better code or something else?

mayank99: Did you read the PR description?

packages/itwinui-css/backstop/tests/progress-indicator.html
```diff
@@ -1099,5 +1104,13 @@ <h2>Overlay</h2>
         </div>
       </div>
     </section>
+
+    <script type="module">
+      (async () => {
+        if (!HTMLElement.prototype.hasOwnProperty('inert')) {
```
rohan-kadkol, line 1110: What does this condition do? Why is it required?

Also, what is `HTMLElement.prototype` supposed to return? Because VSCode was not autocompleting `hasOwnProperty`

image

rohan-kadkol (approved):

PR #828, opened 2022-11-29
chore(dev-deps): Bump `braces`

mayank99:
Fixes npm audit failing from https://github.com/advisories/GHSA-w573-4hg7-7wgq.

Turns out we were force resolving the wrong minor version of `braces`. Removing it from resolutions makes it resolve to minor version `3` which does not have a vulnerable dependency on `decode-uri-component`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,14 +44,12 @@
     "turbo": "~1.3.2"
   },
   "resolutions": {
-    "braces": "^2.3.1",
     "hosted-git-info": ">=3.0.8",
     "jimp": ">=0.16.1",
     "trim-newlines": "^3.0.1",
     "minimatch": "^3.0.5",
     "glob-parent": "^6.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,14 +44,12 @@
...
   },
   "resolutions": {
-    "braces": "^2.3.1",
     "hosted-git-info": ">=3.0.8",
     "jimp": ">=0.16.1",
     "trim-newlines": "^3.0.1",
     "minimatch": "^3.0.5",
     "glob-parent": "^6.0.1",
     "jpeg-js": "^0.4.3",
-    "set-value": "^4.1.0",
```
gretanausedaite, line 54: This is also not needed?

mayank99: yup. it was coming from one of the transitive deps which no longer requires it, so the resolution is not needed

gretanausedaite (approved):

PR #833, opened 2022-12-06
feat: Add `.iui-root-background` to opt into setting background

mayank99:
First step towards https://github.com/iTwin/iTwinUI-react/issues/967.

### DIFF
--- a/packages/itwinui-css/src/global.scss
+++ b/packages/itwinui-css/src/global.scss
@@ -27,8 +27,8 @@
 }
 
 // auto apply background only if <body> is used as root.
-// for other elements, user will have to opt in by adding .iui-root-bg
-:where(body.iui-root, .iui-root.iui-root-bg) {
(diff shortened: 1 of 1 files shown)


rohan-kadkol (approved):

packages/itwinui-css/src/global.scss
```diff
@@ -27,6 +21,17 @@
...
 }
 
+// reset margin only for <body> element.
+:where(body.iui-root) {
+  margin: 0;
+}
+
+// auto apply background only if <body> is used as root.
+// for other elements, user will have to opt in by adding .iui-root-bg
+:where(body.iui-root, .iui-root.iui-root-bg) {
```
rohan-kadkol, line 31: nit: One small consideration could be whether to name it `.iui-root-bg` or `.iui-root-background`. Because I think all public class names/css variables have "background" and not "bg".

FlyersPh9 (approved):

PR #835, opened 2022-12-06
feat(Tag): Tag body receives focus instead of close button within it

### LINKED ISSUE #804: Tag should receive `:focus`
Tag component should receive `:focus`. Tried to address this issue with #793 but it should receive a better review.

FlyersPh9:
- Tag uses `focus-within` and hides `outline` on the close `<button>`.  Closes #804.
- `.iui-tag-container` has `aria-label="My tags"` for better screenreader experience.
- `<button aria-label>` includes the name of the tag for better screenreader experience.


packages/itwinui-css/backstop/tests/tag.html
```diff
@@ -46,49 +47,59 @@ <h2>Default</h2>
         >
           Tag
         </span>
-        <button
+        <span
```
FlyersPh9, line 50: Changed `<button>` to `<span role="button">` which does not receive `:focus`.  `tabindex="0"` is set on the parent.

packages/itwinui-css/src/tag/tag.scss
```diff
@@ -37,6 +36,11 @@
   &:hover {
     border-color: var(--iui-color-border-foreground-hover);
   }
+
+  &:focus {
```
FlyersPh9, line 40: Using `:focus` instead of `@mixin iui-focus` because I want to show the selection when clicked on.

FlyersPh9:
I don't understand why the scrollbar width changed. 🤔
image

gretanausedaite (approved):
Looks good to me 🌵

packages/itwinui-css/backstop/tests/tag.html
```diff
@@ -632,18 +731,21 @@ <h2>Default tag containers</h2>
...
+          class="iui-tag"
+          role="option"
+        >
           <span
             class="iui-tag-label"
             title="Long tag"
           >
-            Long tag
+            Very long tag
```
adamhock, line 742: I think either this should be "Long tag" or the `title` and `aria-label` should be "Very long tag"

adamhock (commented):
Make sure to mention the `aria-multiselectable` and `aria-orientation` changes in the PR description so that the React PR can update this change as well

packages/itwinui-css/backstop/tests/tag.html
FlyersPh9: That would explain [the unexpected test image change](https://github.com/iTwin/iTwinUI/pull/835#issuecomment-1343052369)!  Fixed.

FlyersPh9:
> Make sure to mention the `aria-multiselectable` and `aria-orientation` changes in the PR description so that the React PR can update this change as well

Updated description to cover both of these.

mayank99:
Why was `role=listbox`/`role=option` added?  Have you tested these in a screenreader to see what it actually achieves? Our tags are not selectable, so I'm pretty this role should not be applied here _at all_.

Selectable tags would be a new feature we would need styling and behavior for.

FlyersPh9:
> Have you tested these in a screenreader to see what it actually achieves?

This is with `role="listbox"`:
Screenshot 2022-12-13 at 1 24 22 PM

And without `role="listbox"`:
Screenshot 2022-12-13 at 1 23 40 PM

> Our tags are not selectable, so I'm pretty this role should not be applied here at all.

I'm reading up on [`role="listbox"` on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role#description).  My original thinking was "the ability to remove an option from the listbox makes it valid" but from your comment and re-reading the documentation has me second guessing.  I'll remove.

adamhock (approved):

PR #836, opened 2022-12-08
fix(Code): Remove `user-select: all` and extra declarations

mayank99:
Removed `user-select: all` because it is bad from a usability perspective. It breaks lots of expected behaviors like being able to select part of the text or triple-click to select the whole line.

Also removed `iui-reset` because most of those delcarations are unnecessary (either they are the default or they are being overridden or both).


FlyersPh9 (approved):

rohan-kadkol (approved):
This would also help in the migration tool. Since it was not possible to select portions of the code in the code line.

PR #846, opened 2022-12-08
fix(Select): Use correct placeholder text color

FlyersPh9:
Select placeholder text color now matches [`input.scss` placeholder text color](https://github.com/iTwin/iTwinUI/blob/9d07a550660638b36f3029af00c2ef4d36ce8c38/packages/itwinui-css/src/input/input.scss#L29).


adamhock (approved):

gretanausedaite (approved):

mayank99:
@FlyersPh9 why is this PR prefixed with `style` instead of `fix`? this is user-facing and we want it in the changelog.

PR #847, opened 2022-12-09
fix(Inputs): border color changes to status color on hover

### LINKED ISSUE #624: Inputs: Status border color on hover is not aligned
Textarea
image
Input
image
Select
image

On hover, select gets borders colored according to status, while others do not.

@gretanausedaite suggested that all the inputs have a change of color as this would add some colors to our design system.

adamhock:
- Changes made to `Input` and `Textarea` components so that their border gets colored according to status when they're hovered.
- This is to match the behavior seen in the `Select` component.

Input Before
image
Input After
image

Textarea Before
image
Textarea After
image

Closes #624


FlyersPh9:
Can you add tests to `input.js` & `textarea.js` to check hover & focus on inputs with & without status?

elephantcatdog (commented):
No additional comments

adamhock:
> Can you add tests to `input.js` & `textarea.js` to check hover & focus on inputs with & without status?

Tests have been added

packages/itwinui-css/backstop/scenarios/input.js
```diff
@@ -13,4 +13,20 @@ module.exports = [
   scenario('Type Without Label', {
     selectors: ['#demo-no-label'],
   }),
+  scenario('State hover in default', {
+    actions: [hover('#test-input')],
+    selectors: ['#demo-input'],
```
FlyersPh9, line 18: Can you add `viewports: [{ width: 800, height: 600 }],` to all of these new tests?

packages/itwinui-css/backstop/scenarios/input.js
```diff
@@ -13,4 +13,20 @@ module.exports = [
...
+  }),
+  scenario('State hover with status', {
+    actions: [hover('#test-input-with-status')],
+    selectors: ['#demo-input-with-status'],
+  }),
+  scenario('State focus in default', {
```
FlyersPh9, line 24: You can reduce the amount of test images by covering hover & focus with a single test.  Here's an example from `carousel.js`:
```js
// Hover & focus states
scenario('State hover & focus', {
  actions: [hover('#test-1'), focus('#test-2')],
  selectors: ['#demo-default'],
}),
```

adamhock: Since multiple `iui-input-container` elements aren't wrapped in another element that isn't the entire demo page, I think it's best to just have separate test images for focus and hover

packages/itwinui-css/backstop/scenarios/input.js
adamhock: Added

FlyersPh9: Can you add these to textarea as well?

adamhock: Good catch! Fixed

FlyersPh9 (approved):

elephantcatdog (commented):
It looks like tests for hc dark, hc light, and light are missing for 
image

elephantcatdog (approved):
> It looks like tests for hc dark, hc light, and light are missing

I found them; they just don't render on github for me for some reason.

PR #854, opened 2022-12-14
feat: Add new `Icon` component

mayank99:
**CSS**: Added new `iui-svg-icon` class that takes an svg as a child. Allows changing size and fill using `data-iui-icon-color` and `data-iui-icon-size` attributes.

**React**: Added an `Icon` component, used like this:
```tsx
<Icon><SvgCloseSmall /></Icon>
<Icon fill='informational'><SvgInfoCircular /></Icon>
```

Other notes:
- I used the name `iui-svg-icon` to prevent conflicts with `iui-icon` that is already used in a bunch of components.
- I kept backwards compatibility with the `data-iui-icon-color` attributes because those are already part of our released css.
- I tried not to change the `icon.scss` file because it contains mixins that are used all over the project. I wanted to reuse the icon-sizes map but I was having issues because sass sees `xl` as string but `2xl` as number. and changing the whole map to be explicitly strings would result in dozens of changes in other components, so I just duplicated the icon sizes map.
  - It's not ideal that our `utils` folder is a catch-all that is used to dump all our helper mixins and components. It is confusing to keep track of what's internal vs user facing, makes it hard to change things, and results in a bloated `util.css` file. We should improve this folder in the future.

#### Pending
- [x] narrow down list of sizes (see [comment](https://github.com/iTwin/iTwinUI/pull/854#discussion_r1048959371))
- [x] test with new svgs (after https://github.com/iTwin/iTwinUI-icons/pull/56)
- [x] test with our components (example: passing `<Icon>` to `<Button>`)


packages/itwinui-css/src/utils/icon/icon.scss
```diff
@@ -5,17 +5,17 @@ $iui-icon-color-multicolor-disabled: grayscale(100%);
...
+  'l': calc(1.5 * var(--iui-size-m)),
   // 32px
-  xl: var(--iui-size-xl),
+  'xl': var(--iui-size-xl),
   // 48px
-  2xl: calc(3 * var(--iui-size-m)),
+  '2xl': calc(3 * var(--iui-size-m)),
   // 64px
-  3xl: var(--iui-size-2xl)
+  '3xl': var(--iui-size-2xl)
```
mayank99, line 18: I added data attributes for this entire list of sizes that we already have defined, but i think we should narrow this down. Some of these are not used even internally by any components, and it doesn't feel right that some of these are using `calc` with magic numbers.

This was also the reason icon classes were excluded from the initial v1 release.

mayank99: @FlyersPh9

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@use './icon' as *;
+
+@mixin iui-svg-icon {
+  --_iui-svg-size: #{map-get($iui-icon-sizes, m)};
+  --_iui-svg-fill: var(--iui-color-icon-muted);
+
+  display: inline-flex;
+  flex-shrink: 0;
```
rohan-kadkol, line 10: What are these two lines for?

packages/itwinui-css/src/utils/icon/classes.scss
```diff
@@ -1,23 +1,26 @@
...
+@use './icon-component.scss' as *;
 
-[data-iui-icon-color='default'] {
-  @include iui-icon-color;
-}
-
-[data-iui-icon-color='informational'] {
-  @include iui-icon-color(informational);
+.iui-svg-icon,
+[data-iui-icon-color] {
```
rohan-kadkol, line 6: Is there a reason why having `data-iui-icon-color` without `.iui-svg-icon` is still treated like an icon? I believe we should not change this logic to prevent breaking changes, but don't know why it was like this from before.

rohan-kadkol, line 6: If having `data-iui-icon-color` without `.iui-svg-icon` is treated like an icon, should having `data-iui-icon-size` without `.iui-svg-icon` also be treated like an icon?

```suggestion
.iui-svg-icon,
[data-iui-icon-color], {
[data-iui-icon-size], {
```

packages/itwinui-css/src/utils/icon/icon.scss
FlyersPh9: I'd like for users to be able to pair an icon size with [each font size we offer](https://itwin.github.io/iTwinUI/typography.html).  How we do that, I am open to suggestions.

packages/itwinui-css/src/utils/icon/icon-component.scss
mayank99: `flex-shrink: 0` is to ensure that the icon maintains its size when used in a flex container.

`inline-flex` is to make sure the element wrapping the svg only takes up the height of the svg. without it, you will find that the height increases by a few pixels and the vertical alignment gets messed up.

packages/itwinui-css/src/utils/icon/classes.scss
mayank99: yes it's to prevent breaking changes, and i believe it was done because it's already used in expandable-block (including in react)

mayank99: no need, because that was done for backwards compatibility whereas `data-iui-icon-size` is a new addition

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@use 'sass:map';
+
+$icon-sizes: (
```
rohan-kadkol, line 5: Is there a benefit to adding this here and not using `$iui-icon-sizes` in `icon.scss` like were doing before?

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -18,7 +33,7 @@
 
 /// @arg $size - must be one of: 's', 'm', 'l', '2xl', '3xl'
 @mixin svg-icon-size($size) {
-  --_iui-svg-size: #{map-get($iui-icon-sizes, $size)};
+  --_iui-svg-size: #{map-get($icon-sizes, $size)};
```
rohan-kadkol, line 36: If you are using `map.get` on [L21](https://github.com/iTwin/iTwinUI/pull/854/commits/1a3fb7eeec869b4e906dd1e4375be175a6fef823#diff-b0734b3e90e13cc37780569bf40d58ef79c786c43f23240f1ae7ad180edf9b0bR21), this could be `map.get` too

mayank99: updated

packages/itwinui-css/src/utils/icon/icon-component.scss
mayank99: yes, updated description. it was resulting in too many changes

packages/itwinui-css/src/utils/icon/icon.scss
mayank99: Currently we offer these font sizes (in rem): 12px, 14px, 16px, 18px, 24px, 32px

And our current icon map has the following (in rem): 12px, 16px, 24px, 32px, 48px, 64px

In avatar, we are using: 24px, 32px, 48px, 64px
In progress-indicator, we are using: 16px, 24px, 32px, 48px
In radio-tile, we are using: 32px
In tile and file-upload, we are using: 64px
(and in a bunch of other components, we are using: 12px, 16px, 24px)

so it will not be enough to only have sizes for each font size. in fact some of it will probably not even be useful because of how close together they are (14px, 18px), and the icons also don't correspond 1:1 to the font size they need.

with that in mind, here's what i suggest:
- offer a limited set of commonly-used sizes. lets say: 12px, 16px, 24px, and maybe 32px
- consider avatar and progress-indicator as outliers. they don't need to follow the icon sizing scale.
- for any bespoke icons that don't fit the scale, suggest calculating manually.

thoughts?

FlyersPh9: > the icons also don't correspond 1:1 to the font size they need.

I know that we like the 16px icon size + 14px font size pairing.  Any other sizes could scale accordingly to match.  I do think people will want recommended sizes (wether we provide them or not) to pair with the font sizes.

> consider avatar and progress-indicator as outliers. they don't need to follow the icon sizing scale.

Agreed.

mayank99: removed `xl`, `2xl` and `3xl`. there are now only 4 sizes: `s` (12px), `m` (16px), `l` (24px) and `auto` (1em, scales with text).

let me know what you think. looking at it 100% zoom level, i think some of these are too small

mayank99: clamping it to a minimum of 1rem using this value `max(1rem, 1em)` looks like this:

mayank99: `1.125rem`


`calc(16/14 * 1rem)`

gretanausedaite (commented):
No additional comments.

packages/itwinui-css/src/utils/icon/icon.scss
FlyersPh9: I don't like `max(1rem, 1em)` because the small text & body text share the same icon size.

I can't see a visual difference between `1.125rem` & `calc(16/14 * 1rem)` but they both look good to me.

mayank99: now using 16/14 which equals `1.14286em`

packages/itwinui-css/backstop/tests/icon.html
```diff
@@ -64,24 +82,87 @@ <h2>Actionable icons</h2>
...
 
       <svg-status-error
+        class="iui-svg-icon"
         data-iui-icon-color="negative"
         aria-hidden="true"
       ></svg-status-error>
     </section>
+
+    <h2>Auto scale with text</h2>
+    <section id="demo-typography">
```
FlyersPh9, line 110: Can you update `icon.js` to screenshot the `<section>`s instead of the entire page?

mayank99: done

FlyersPh9: The 4 test images from the prior test can be deleted.

mayank99: done

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -135,11 +136,7 @@ export const ExpandableBlock = (props: ExpandableBlockProps) => {
           <div className='iui-title'>{title}</div>
           {caption && <div className='iui-caption'>{caption}</div>}
         </span>
-        {icon &&
-          React.cloneElement(icon, {
-            className: cx('iui-status-icon', icon.props.className),
-            'data-iui-icon-color': status,
-          })}
+        {icon && <Icon fill={status}>{icon}</Icon>}
```
gretanausedaite, line 139: We have several places in react package where we are wrapping icon with span (eg. Button start and end icons). Does it makes sense to also update those with new component or is it breaking change?

packages/itwinui-react/src/core/utils/components/Icon.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * <Icon size='default'>
+ *   <SvgPlaceholder />
+ * </Icon>
+ *
+ * @example
+ * <Icon fill='negative'>
+ *   <SvgStatusError />
+ * </Icon>
+ */
+export const Icon = (props: IconProps) => {
```
gretanausedaite, line 51: I believe our users would benefit from having storybook examples of new Icon component. If we don't want to add it separately, then we should add usage examples to other components.

gretanausedaite (commented):
Code looks good. I'm missing usage examples in react (storybook).

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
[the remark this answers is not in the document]
mayank99: I want to leave it unchanged to avoid any potential breaking changes for anyone who might be relying on those classes.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.test.tsx
```diff
@@ -129,11 +129,14 @@ it.each(['positive', 'negative', 'warning', 'informational'] as const)(
     expect(container.querySelector('.iui-expandable-block')).toBeTruthy();
     expect(queryByText('test title')).toHaveClass('iui-title');
 
+    const statusIconWrapper = container.querySelector('.iui-svg-icon');
+    expect(statusIconWrapper).toHaveAttribute('data-iui-icon-color', status);
+
     const {
-      container: { firstChild: statusIcon },
-    } = render(StatusIconMap[status]({ className: `iui-status-icon` }));
```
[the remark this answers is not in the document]
mayank99: `<Icon>` handles everything that `iui-status-icon` would have handled

packages/itwinui-react/src/core/utils/components/Icon.tsx
```diff
@@ -0,0 +1,64 @@
...
+import React from 'react';
+import cx from 'classnames';
+import '@itwin/itwinui-css/css/utils.css';
+
+export type IconProps = {
+  /**
+   * Size of the svg.
+   *  - 'auto': scales with text
+   *  - 'medium': works well with most text
+   *  - 'small': smaller than medium
```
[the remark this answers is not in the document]
mayank99: i put `medium` first because the other two sizes are relative to `medium`

packages/itwinui-react/src/core/utils/components/Icon.tsx
```diff
@@ -0,0 +1,64 @@
...
+ *   <SvgStatusError />
+ * </Icon>
+ */
+export const Icon = (props: IconProps) => {
+  const { size = 'auto', fill = 'default', className, ...rest } = props;
+
+  return (
+    <span
+      className={cx('iui-svg-icon', className)}
+      data-iui-icon-size={sizeMap[size] || size}
```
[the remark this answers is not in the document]
mayank99: if the user is not using typescript (or overrides using `any`), this will serve as a nice fallback.

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/// @arg $size - must be one of: 's', 'm', 'l', 'auto'
+@mixin svg-icon-size($size) {
+  --_iui-svg-size: #{map.get($icon-sizes, $size)};
+}
+
+/// @arg $fill - must be one of: 'informational', 'positive', 'warning', 'negative'
+@mixin svg-icon-status-color($fill) {
+  --_iui-svg-fill: var(--iui-color-icon-#{$fill});
```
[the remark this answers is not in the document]
mayank99: no because we are setting a css variable here, whereas that mixin sets `fill` directly

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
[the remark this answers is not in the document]
rohan-kadkol: Is updating the dom of `ExpandableBlock.tsx` a breaking change? E.g. if users have selectors with the old dom tree.

If so, can we update `ExpandableBlock.tsx` to use `<Icon/>` instead of `.iui-status-icon` in a minor release (according to the [changeset message](https://github.com/iTwin/iTwinUI/pull/854/files#diff-858caa1acb0daf69f63eb77bf057bf99303b567ab8f8d43d376f8838ddfeb503R2))?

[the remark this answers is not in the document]
mayank99: well, the itwinui-react API is unaffected. Users should not be targeting internal classes produced by our react components

whereas for itwinui-css, we are expected to keep supporting the old class names because someone might be consuming them directly.

mayank99: we cannot remove the old class names but we could support both use cases. i'm not sure of the benefit though

mayank99:
I've changed the default `size` to `medium` because that is the most common use case. Users will need to opt into autoscaling with text using `size='auto'`.

packages/itwinui-react/src/core/utils/components/Icon.tsx
mayank99: added
https://itwin.github.io/iTwinUI/854/react/?path=/story/utilities-icon

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -0,0 +1,43 @@
...
+  --_iui-svg-fill: var(--iui-color-icon-muted);
+
+  display: inline-flex;
+  flex-shrink: 0;
+
+  &,
+  svg {
+    width: var(--_iui-svg-size);
+    height: var(--_iui-svg-size);
+    fill: var(--_iui-svg-fill);
```
rohan-kadkol, line 24: Is there a way to supply a custom `fill` when using `<Icon>` for just the default sizing? I guess we have to manually overwrite the `--_iui-svg-fill` CSS variable, but is there an easier way?

Because when I tried to supply a custom fill to `<SvgAdd fill='red'/>`, it was not used.

```tsx
<Icon>
  <SvgAdd fill='red' />
</Icon>
```


image

rohan-kadkol: Just realized, we can do that with inline style. To confirm, is this the correct way?

```tsx
<Icon>
  <SvgAdd style={{ fill: 'red' }} />
</Icon>
```

gretanausedaite (approved):
LGTM

mayank99:
so what do we do about this? https://github.com/iTwin/iTwinUI/pull/854#issuecomment-1372811544

> I've also tested in some of our components like `<IconButton>` and it displays fine but the wrapping span ends up increasing in height in a little bit.

gretanausedaite:
> so what do we do about this? [#854 (comment)](https://github.com/iTwin/iTwinUI/pull/854#issuecomment-1372811544)
> 
> > I've also tested in some of our components like `<IconButton>` and it displays fine but the wrapping span ends up increasing in height in a little bit.
> > 

Is it avoidable?

mayank99:
> Is it avoidable?

I haven't been able to figure out why this is happening. need to investigate and play around a bit more. Maybe someone else can figure it out?

packages/itwinui-react/src/core/utils/components/Icon.tsx
```diff
@@ -16,23 +17,37 @@ export type IconProps = {
...
-const sizeMap = {
-  auto: 'auto',
-  small: 's',
-  medium: 'm',
-  large: 'l',
-} as const;
+const getSizeValue = (size: string) => {
+  switch (size) {
+    case 'auto':
+      return 'auto';
```
rohan-kadkol, line 40: Is this case needed? Because it will fall into the default case anyways

```suggestion
```

apps/storybook/src/Icon.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+      options: ['auto', 'small', 'medium', 'large'],
+    },
+    fill: {
+      control: 'radio',
+      options: ['default', 'informational', 'positive', 'warning', 'negative'],
+    },
+  },
+} as Meta<IconProps>;
+
+export const Default: Story<IconProps> = (args) => {
```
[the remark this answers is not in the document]
mayank99: I would like to not advertise this so broadly at first. We don't want people using random values

It's kinda like how we don't show examples of what you can do with custom `className`/`style` tags.

rohan-kadkol (approved):

.changeset/large-chefs-fix.md
```diff
@@ -0,0 +1,14 @@
+---
+'@itwin/itwinui-css': minor
+---
+
+Added new `iui-svg-icon` class for displaying svgs
+- Supports `data-iui-icon-size` attribute for size. Can be one of: 's', 'm', 'l', 'auto'
+- Supports `data-iui-icon-color` attribute for fill. Can be one of: 'informational', 'positive', 'warning', 'negative'
```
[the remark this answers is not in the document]
mayank99: mentioned in the jsdoc. [`9d63366`](https://github.com/iTwin/iTwinUI/pull/854/commits/9d633665157a41983b1df16505df2a08ed20b355)

mayank99:
> > > so what do we do about this? [#854 (comment)](https://github.com/iTwin/iTwinUI/pull/854#issuecomment-1372811544)
> > > > I've also tested in some of our components like `<IconButton>` and it displays fine but the wrapping span ends up increasing in height in a little bit.
> > > > 
> > 
> > Is it avoidable?
> 
> I haven't been able to figure out why this is happening. need to investigate and play around a bit more. Maybe someone else can figure it out?

should be fixed in [`bccf588`](https://github.com/iTwin/iTwinUI/pull/854/commits/bccf588bbe74a377d9f10d9dde2e48596d46b612). does that look good?

packages/itwinui-css/src/button/button.scss
```diff
@@ -33,6 +33,7 @@
 }
 
 @mixin iui-button-icon {
+  &,
```
gretanausedaite, line 36: Adding display prop directly works better. Inherit or inline-flex.

```suggestion
  display: inherit;
```

mayank99: used `inline-flex` in [`e209c85`](https://github.com/iTwin/iTwinUI/pull/854/commits/e209c85691cde4bd1babd3ba88ceb125d562361c).

found test was failing in skip-to-content because class was applied on wrong element.  fixed in [`9dbd5e1`](https://github.com/iTwin/iTwinUI/pull/854/commits/9dbd5e196c6c92cdb96d52ddde57680ca7c259dd), all tests should pass now

gretanausedaite (approved):
Looks good to go.
Maybe add comment why you are leaving `Autoscale` story out of tests.

PR #857, opened 2022-12-15
chore: Update deps

mayank99:
Updated deps in preparation of monorepo merge. It's important not to have conflicts especially when it comes to `prettier` and `husky`. Husky hook format changed (moved from package.json to .husky folder) and prettier resulted in some formatting changes in vite.config.ts and badge.js. Other than that, everything seems to work the same.

In the `resolutions` field, I removed any ambiguity and pointed to exact latest version. This is similar to https://github.com/iTwin/iTwinUI-react/pull/990 which was done to avoid issues in azure pipelines.


package.json
```diff
@@ -34,24 +30,23 @@
...
-    "trim-newlines": "^3.0.1",
-    "minimatch": "^3.0.5",
-    "glob-parent": "^6.0.1",
-    "jpeg-js": "^0.4.3",
-    "ansi-regex": "^5.0.1",
-    "node-fetch": "^2.6.7",
-    "async": "^3.2.2"
+    "hosted-git-info": "6.1.1",
+    "jimp": "0.16.2",
+    "trim-newlines": "3.0.1",
```
rohan-kadkol, line 44: `trim-newlines` latest version is `4.0.2`. Should we use that instead? Any reason why we are sticking with the `3.0.1` tag and not `latest`? ([source](https://www.npmjs.com/package/trim-newlines?activeTab=versions))

image

package.json
```diff
@@ -34,24 +30,23 @@
...
+    "minimatch": "3.1.2",
```
rohan-kadkol, line 45: Any reason we are not going to `latest` tag and sticking with `v3-legacy` instead?

package.json
```diff
@@ -34,24 +30,23 @@
...
+    "glob-parent": "6.0.2",
+    "jpeg-js": "0.4.4",
+    "ansi-regex": "5.0.1",
```
rohan-kadkol, line 48: Latest is `6.0.1`. Maybe we are sticking with v5 to avoid breaking changes?

package.json
```diff
@@ -34,24 +30,23 @@
...
+    "node-fetch": "2.6.7",
```
rohan-kadkol, line 49: Latest is `3.3.0`

package.json
```diff
@@ -34,24 +30,23 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "audit-ci": "^4.0.0",
```
rohan-kadkol, line 37: Any reason why you removed `audit-ci`? If it is not needed, does we need to remove the `audit:ci` script in `packages/itwinui-css/packages.json`?

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
mayank99: this file was auto generated by husky. i'm not entirely sure but i think it will find the first `sh`

itwinui-react has the same: https://github.com/iTwin/iTwinUI-react/blob/main/.husky/pre-commit

package.json
mayank99: i tried to keep all resolutions to the same major version. most of these resolutions were done solely to pass `yarn audit`, so there is no reason to upgrade them to latest and risk breaking changes

package.json
mayank99: see https://github.com/iTwin/iTwinUI/pull/857#discussion_r1050180912

package.json
mayank99: see https://github.com/iTwin/iTwinUI/pull/857#discussion_r1050180912

package.json
mayank99: see https://github.com/iTwin/iTwinUI/pull/857#discussion_r1050180912

also i'm pretty `node-fetch` v3 will not even work because it has dropped support for CJS environments (it is ESM-only)

package.json
mayank99: i removed it because it was not being used anywhere. removed the script in next push as well

rohan-kadkol (approved):

veekeys:
Have you thought about getting rid of husky and instead just use git pre commit hook?

mayank99:
> Have you thought about getting rid of husky and instead just use git pre commit hook?

it is already a git pre-commit hook. husky is just used to install it. git hooks live in the `.git` folder which is not part of version control

gretanausedaite (approved):

PR #860, opened 2022-12-16
feat(Tabs): Right aligned content within Tabs row

### LINKED ISSUE #770: Allow for right aligned content placed within tabs row
#### Feature
A user has requested the ability to place non-tab content right aligned within the tab row.  Original request was shared in iTwinUI-React repo but I believe may need to be resolved in iTwinUI-CSS first: https://github.com/iTwin/iTwinUI-react/discussions/806

#### Examples
image
image
image

rohan-kadkol:
Closes #770 

* To avoid breaking changes, supported this using CSS changes of `display: grid` instead of HTML dom changes
  * To add right/bottom aligned content for horizontal/vertical tabs, we can add it to `div.iui-tabs-actions`
<br/>

* `.iui-tabs-actions` supported types:
  * Only supported for `default` and `borderless` tabs. It is not supported for `pill`.
* `.iui-tabs-actions` supported sizes:
  * `data-iui-size="small"` and `data-iui-size="normal"`

<details>
<summary>See screenshots</summary>

### Default

* Horizontal tabs & `data-iui-size='small'` actions
image

* Horizontal tabs & `data-iui-size='normal'` actions
image

* Horizontal large tabs & `data-iui-size='normal'` actions
image

* Vertical tabs & `data-iui-size='normal'` actions
image

### Borderless

* Horizontal tabs & `data-iui-size='small'` actions
image

* Horizontal tabs & `data-iui-size='normal'` actions
image

* Horizontal large tabs & `data-iui-size='normal'` actions
image

* Vertical tabs & `data-iui-size='normal'` actions
image


packages/itwinui-variables/src/sizes.scss
```diff
@@ -12,6 +12,8 @@
   --iui-size-xl: #{math.div(32, 16)}rem;
   --iui-size-2xl: #{math.div(64, 16)}rem;
   --iui-size-3xl: #{math.div(96, 16)}rem;
+
+  --iui-gap-between-inputs: var(--iui-size-s);
```
mayank99, line 16: i dont think this needs to go in the variable packages (at least not in this PR). it can just be a sass variable in `tabs.scss`

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -2,6 +2,22 @@
...
+  position: relative;
+
+  .iui-tabs-actions {
+    grid-area: tabs-actions;
+    display: flex;
+    justify-content: flex-end;
+    align-items: center;
+  }
+
+  .iui-tabs-content {
```
mayank99, line 16: `iui-tabs-actions` and `iui-tabs-content` don't need to be nested. they can be separate mixins

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -111,9 +128,17 @@
...
+
+  .iui-tabs-actions {
+    column-gap: var(--iui-size-xs);
+    margin-left: var(--iui-gap-between-inputs);
+  }
 }
 
 @mixin iui-tabs-vertical {
+  grid-template-areas: 'tabs tabs-content' 'tabs-actions tabs-content';
+  grid-template-columns: fit-content(1px) auto;
```
mayank99, line 140: what does this line do?

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -136,6 +161,12 @@
       width: var(--iui-size-3xs);
     }
   }
+
+  .iui-tabs-actions {
+    flex-direction: column;
```
mayank99, line 166: i think these should be left aligned when vertically stacked. center alignment looks weird

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -147,7 +178,8 @@
 
   .iui-active::after {
     @media (prefers-reduced-motion: no-preference) {
-      transition: width var(--iui-duration-1) ease, height var(--iui-duration-1) ease;
+      transition: width var(--iui-duration-1) ease,
```
mayank99, line 181: unintentional formatting changes :(

packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -234,6 +249,20 @@ <h2>Default</h2>
             </button>
           </li>
         </ul>
+
+        <div class="iui-tabs-actions">
+          <button
+            class="iui-button iui-button-dropdown"
+            data-iui-size="small"
```
mayank99, line 256: The height of a default button is 36px whereas the height of tabs is 40px, so it should be possible to support the default size, not just small

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -2,6 +2,22 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @import '../utils/index';
 
+@mixin iui-tabs-wrapper {
+  display: grid;
```
mayank99, line 6: There are some random image changes in pill tabs. Might be because of `grid` but can you verify?

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Yes, I noticed that when making this PR ready for review. But instead of reverting it, I thought of keeping it, similar to https://github.com/iTwin/iTwinUI/pull/935#discussion_r1057810187

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: This is the make the left column take only as much width as needed and the right column to take all the remaining space. But if there is a better way to achieve this, I could change it

rohan-kadkol: Without it, the width changes when switching tabs:

[Video6.webm](https://user-images.githubusercontent.com/45748283/210439288-d4f2762f-76a8-4d1e-8b39-489cdbb7d963.webm)

packages/itwinui-variables/src/sizes.scss
rohan-kadkol: Replaced the variables package variable with a CSS variable in `tabs.scss` prefixed with `_`, similar to [`expandable_block.scss (L6)`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/src/expandable-block/expandable-block.scss#L6)

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: This is how it looks with left alignment. Which one looks better @mayank99 @FlyersPh9 ? Because I feel the center one looks better. The left aligned one seems a bit off because it is misaligned with the `+` icon in the bottom most tab that is center aligned.

image

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Moved them to separate mixins in [`7c39650` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/7c39650b69b80c5de6ec603bd8e862005b0aa25f)

packages/itwinui-css/backstop/tests/tabs.html
rohan-kadkol: I tried this out before and saw that the vertical spacing between `iui-tabs-actions` and `iui-tabs-content` is too small.

image

Also, this is similar to what you mentioned in https://github.com/iTwin/iTwinUI/issues/770#issuecomment-1260907982.

But if this small spacing is okay, then yes, we can support the `data-iui-size="normal"` also.

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Those random changes seem to be happening because of the spacing methods I used in `grid-template-columns`. I'm not sure why it caused some spacing issues in storybook, because it worked correctly in CSS.

Changing the spacing methods in `grid-template-columns` in [`9cc7d64` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/9cc7d6498cd07bc4b2aef1316ffbfd72fb1d6abc) got rid of the storybook `Tabs -> Pills` test image change, reduced the number of CSS test image changes, and also made the CSS test image changes cleaner (fewer changes in each image).

packages/itwinui-variables/src/sizes.scss
mayank99: it doesn't need to be a css variable. it is used like a constant value, so sass variable will be fine.

in expandable-block we are assigning it a bunch of times (it's "dynamic"), that's why it's a css variable

packages/itwinui-css/src/tabs/tabs.scss
mayank99: we might want to make those buttons take the full width then

packages/itwinui-css/backstop/tests/tabs.html
mayank99: i think in your screenshot it is center aligned, so the 4px distributed above and below. try aligning it to `start`

packages/itwinui-css/src/tabs/tabs.scss
mayank99: i believe `grid-template-columns: auto 1fr` will achieve that. no need for `fit-content`

packages/itwinui-css/backstop/tests/tabs.html
rohan-kadkol: Aligning it to `start` would almost work for the normal sized button. But for the small sized button, it seems quite misaligned vertically.

image

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Thanks, yes, I realized about `fr` later. So I changed it to `min-content 1fr` as mentioned in https://github.com/iTwin/iTwinUI/pull/860#discussion_r1061025070

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Does this look alright? Apart from `align-items: stretch`, I also added `padding: 0 var(--iui-size-xs)`.

image

packages/itwinui-variables/src/sizes.scss
rohan-kadkol: Ohh, right, changed it to a sass variable [`89bdee6` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/89bdee6c86699095ef8a8ec0a3575f2db8390976)

packages/itwinui-css/src/tabs/tabs.scss
FlyersPh9: I like the full width option.  Any negative impact when using components other than buttons?  Checkbox with a label comes to mind.

packages/itwinui-css/src/tabs/tabs.scss
mayank99: Looking at the docs for [`min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/min-content), it seems like it should fail this situation and cause wrapping, but it seems to work fine, possibly because of `whitespace: nowrap` on the tab.


[`auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns#:~:text=as%20a%20minimum.-,auto,-As%20a%20maximum) would still be more appropriate and resilient, but minor

packages/itwinui-css/src/tabs/classes.scss
```diff
@@ -13,6 +12,14 @@
   &.iui-vertical {
     @include iui-tabs-vertical;
   }
+
+  .iui-tabs-actions {
```
FlyersPh9, line 16: I _think_ you should be able to un-nest `.iui-tabs-actions` & `.iui-tabs-content` to reduce specificity.

```scss
.iui-tabs-wrapper {
  …

  &.iui-horizontal { … }
  &.iui-vertical { … }
}

.iui-tabs-action { … }
.iui-tabs-content { … }
```

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Thanks, changed

packages/itwinui-css/src/tabs/classes.scss
rohan-kadkol: Ohh, thanks, yes, I think it works. Because no other component would have `.iui-tabs-action` because of the "`iui-tabs`" prefix and also all tests passed.

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -111,9 +119,19 @@
       height: var(--iui-size-3xs);
     }
   }
+
+  .iui-tabs-actions {
```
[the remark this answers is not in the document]
rohan-kadkol: Thanks! Changed to option 2, since this reduced the specificity.

packages/itwinui-css/src/tabs/tabs.scss
mayank99: should be revertable now that #950 is merged

packages/itwinui-css/backstop/tests/tabs.html
FlyersPh9: > i expect people will want to place normal size inputs and buttons in the actions area

I do agree with this.

We could increase the height by increasing the padding on tabs or on the right aligned actionable content area.  The one potential issue I see with including the padding on the actionable area is if the content inside appears / disappears based on what tab is active.  This would cause the tabs to grow / shrink in height as content is added / removed.

I'm playing around with it locally now, how about `min-height: calc(var(--iui-size-l) * 2);` on horizontal tabs?  The default button size is 36px & this would make the tabs 48px tall.
Screenshot 2023-01-05 at 1 18 20 PM
Screenshot 2023-01-05 at 1 18 41 PM

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Yes, it was reverted after merging with `main`

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
FlyersPh9: > i could be wrong but i think you actually don't need `@at-root`

I think `@at-root` is required since it is a child of `.iui-tabs-wrapper` & `.iui-horizontal`.
```html
<div class="iui-tabs-wrapper iui-horizontal">
	<ul />
	<div class="iui-tabs-actions" />
	<div class="iui-tabs-content" />
</div>
```

> also we should probably target the tabs wrapper class. `iui-horizontal`/`iui-vertical` are too generic
> and if we want to lower specificity, we can use `:where`

Good point, for safety reasons it makes sense to do this.  Eventually would like to replace these generic classes with `data-iui-orientation` or `aria-orientation`.

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: I tried with a few components other than buttons. I noticed issues in `tag`, `keyboard`, and `badge`.

But I believe this is expected behavior because this happens even without `align-items: stretch`. So the below screenshot happens with or without `align-items: stretch`.

image

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
rohan-kadkol: I implemented the last code block mentioned in https://github.com/iTwin/iTwinUI/pull/860#discussion_r1062703137 but with also including the `&` inside the `:where()` to reduce the specificity even further.

```diff
- :where(.iui-tabs-wrapper.iui-horizontal) & { … }
+ :where(.iui-tabs-wrapper.iui-horizontal &) { … }
```

To confirm, does the second selector above have 0 specificity since everything is in `:where()`?

---

I'm not sure why, but looks like even without `@at-root` works.

I'm curious about the difference between `&` and `#{&}`. Because the generated CSS for both of those are different.

```scss
// Case 1: & (Works)
:where(.iui-tabs-wrapper.iui-vertical &) {} // SCSS
:where(.iui-tabs-wrapper.iui-vertical .iui-tabs-actions) {} // Generated CSS

// Case 2: #{&} (Doesn't work)
:where(.iui-tabs-wrapper.iui-vertical #{&}) {} // SCSS
.iui-tabs-actions :where(.iui-tabs-wrapper.iui-vertical .iui-tabs-actions) {} // Generated CSS

// Case 3: #{&} with @at-root (Works)
@at-root :where(.iui-tabs-wrapper.iui-vertical #{&}) {} // SCSS
:where(.iui-tabs-wrapper.iui-vertical .iui-tabs-actions) {} // Generated CSS
```

For reference, this is how the `.iui-tabs-actions` selector is defined in `tabs/classes.scss`
```scss
.iui-tabs-actions {
  @include iui-tabs-actions;
}
```

packages/itwinui-css/backstop/tests/tabs.html
rohan-kadkol: I added a `min-height: calc(var(--iui-size-l) * 2);` to `.iui-tab` in the mixin `iui-tabs-horizontal`. [`47cb7f4` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/47cb7f47cc5392d4afd10dc6951c252052d5429c)

---

> This would cause the tabs to grow / shrink in height as content is added / removed.

I think this would happen even when applying `min-height` to `.iui-tab`, because if `.iui-tabs-actions` even becomes larger than around `48px` when one of the tabs is selected, then `iui-tab` would match that height.

The below example is with the above mentioned `min-height` on `.iui-tab`. Here, clicking the tab named "Left" sets `data-iui-size="large"` on the dropdown button.

[Video7.webm](https://user-images.githubusercontent.com/45748283/210865807-78ba8eba-6646-49cd-8a9e-63faa7d4a7e4.webm)

But I feel adding a `min-height` to `.iui-tab` is at least more reliable than adding a padding to `.iui-tabs-actions`.

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
mayank99: i don't think we need to make it 0. it's perfectly fine to leave it as `0-1-0` 

the reason it works without `@at-root` is simple: it is officially supported 😄  https://sass-lang.com/documentation/style-rules/parent-selector


case 2 probably doesn't work because it is breaking out of the intended syntax (`.outer-selector &`). when you use `#{}`, it ignores everything and just puts the value there without actually checking what is inside it. my guess: when it first encounters that selector, the sass compiler interprets it as similar to: `:where(.iui-tabs-wrapper.iui-vertical #{something-else})` so it puts the `.iui-tabs-actions` in the beginning as if it was normal nesting.

case 3 works because `@at-root` is specifically used to construct selectors manually, breaking out of implicit nesting.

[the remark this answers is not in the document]
FlyersPh9: Interesting… it looks like nesting `&` breaks it out of the nesting. 🤔
```scss
@mixin iui-tabs-actions {
  …

  .iui-tabs-wrapper.iui-horizontal & {
    color: red;
  }
}
```

I would have expected:
```css
.iui-tabs-actions .iui-tabs-wrapper.iui-horizontal .iui-tabs-actions { color: red; }
```

But instead got:
```css
.iui-tabs-wrapper.iui-horizontal .iui-tabs-actions { color: red; }
```

Crazy that I've made it this far and never knew this, seems pretty fundemental.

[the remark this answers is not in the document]
rohan-kadkol: > i don't think we need to make it 0. it's perfectly fine to leave it as 0-1-0

Removed `&` from the `:where()` to bring the specificity to 0-1-0 [`418cca5` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/418cca506e810c942ea35fffbca106256934e16f)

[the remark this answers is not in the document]
FlyersPh9: Wow, I just opened `stepper.scss` & removed all `@at-root`s and everything is working as expected. 😦 I bet we can remove a LOT of `@at-root` across our repo now knowing this.

packages/itwinui-css/backstop/tests/tabs.html
mayank99: 1. What if we leave default tabs untouched? We can say it works best with small size inputs/buttons in the actions area. The reason: app ui widgets make heavy use of default tabs, so conserving space is probably important there.

2. For borderless tabs, we should find a standard gap between the line and the actions. Lets say 8px. So 36px (button height) + 8px (gap) =`44px` is the minimum height we need to set. We can still keep the actions center aligned vertically (instead of `start`), but ensure that there is always an 8px gap below the actions.

rohan-kadkol: > For borderless tabs, we should find a standard gap between the line and the actions. Lets say 8px. So 36px (button height) + 8px (gap) =44px is the minimum height we need to set. We can still keep the actions center aligned vertically (instead of start), but ensure that there is always an 8px gap below the actions.

To confirm, if we give a height of 44px to `.iui-tab`, then that gives 8px of free space for `.iui-tabs-actions`. But if `.iui-tabs-actions` is aligned centrally (vertically), then the gap between the bottom of the `.iui-tabs-actions` and the `.iui-tabs-content` is 4px, not 8px. Or am I misunderstanding something?

I was thinking of adding a bottom padding to `.iui-tabs-actions` in the `@iui-tabs-borderless` mixin. But I couldn't come up with a selector because `.iui-tabs-actions` and `.iui-tabs.iui-borderless` are siblings.

image

Is there a better way to enforce a minimum of 8px of vertical spacing between `.iui-tabs-actions` and `.iui-tabs-content` when in borderless tabs?

mayank99: flexbox might be able to achieve this

mayank99:
Are you going to handle the react part in this PR too? If not, I suggest making another PR on top of this and testing it before merging this.

Also can you add changesets?

rohan-kadkol:
> Are you going to handle the react part in this PR too? If not, I suggest making another PR on top of this and testing it before merging this.

Sounds good, I will add the react changes in another PR and test that PR before merging this one.

> Also can you add changesets?

Added [`78f4c6c` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/78f4c6c8607c521d79b7d66b982ec6cfda971eed)

.changeset/cool-baboons-crash.md
```diff
@@ -0,0 +1,28 @@
...
+        class="iui-button-icon"
+        aria-hidden="true"
+      ></svg-caret-down-small>
+    </button>
+  </div>
+
+  <div class="iui-tabs-content">...</div>
+  <div class="iui-tabs-content">...</div>
+  <div class="iui-tabs-content">...</div>
+</div>
```
mayank99, line 27: Can be shortened. The content inside tabs-actions doesn't really matter

```suggestion
<div class="iui-tabs-wrapper iui-horizontal">
  <ul class="iui-tabs iui-default">...</ul>
  <div class="iui-tabs-actions">...</div>
  <div class="iui-tabs-content">...</div>
</div>
```

packages/itwinui-css/backstop/tests/tabs.html
rohan-kadkol: Thanks, used these ideas to make it work. I added the following spacing:

* For horizontal tabs: minimum actions-content gap of `var(--iui-size-3xs)` just so that actions will never stick to content ([L215](https://github.com/iTwin/iTwinUI/pull/860/files#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR215))

image

* For horizontal borderless tabs: minimum actions-content gap of `var(--iui-size-xs)` (`8px`) ([L220](https://github.com/iTwin/iTwinUI/pull/860/files#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR220))
(Because of the min spacing, the actions button is slightly above the horizontal centerline)
image

* For vertical tabs: horizontal padding of `var(--iui-size-xs)` ([L216](https://github.com/iTwin/iTwinUI/pull/860/files#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR216))
image

Are these spacings alright? @mayank99 @FlyersPh9

mayank99: what was wrong with my suggested approach? the minimum height of the horizontal tabs has now increased to 48px (it was 44px in my suggestion) because there is an extra 4px above the actions.

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -191,3 +204,58 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+
+  // To always prevent actions from sticking to content
+  --_iui-borderless-horizontal-tabs-actions-content-gap: var(--iui-size-3xs);
+  --_iui-borderless-vertical-tabs-actions-content-gap: var(--iui-size-xs);
+
+  // Add larger min actions-content spacing for borderless horizontal tabs
+  :where(.iui-tabs-wrapper.iui-horizontal) .iui-borderless + & {
```
mayank99, line 219: should not target `iui-borderless` without `iui-tabs`
```suggestion
  :where(.iui-tabs-wrapper.iui-horizontal) :where(.iui-tabs.iui-borderless) + & {
```

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -191,3 +204,58 @@
...
+
+@mixin iui-tabs-actions-wrapper {
+  grid-area: tabs-actions;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+
+  // To always prevent actions from sticking to content
+  --_iui-borderless-horizontal-tabs-actions-content-gap: var(--iui-size-3xs);
+  --_iui-borderless-vertical-tabs-actions-content-gap: var(--iui-size-xs);
```
mayank99, line 216: these can be sass variables

packages/itwinui-css/src/tabs/classes.scss
```diff
@@ -15,6 +14,18 @@
   }
 }
 
+.iui-tabs-actions-wrapper {
+  @include iui-tabs-actions-wrapper;
+
+  .iui-tabs-actions {
```
mayank99, line 20: does this need to be nested?

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -95,13 +102,16 @@
...
-  flex-direction: column;
+  grid-template-areas: 'tabs tabs-actions' 'tabs-content tabs-content';
+  grid-template-columns: 1fr auto;
 
   .iui-tabs {
     display: flex;
   }
 
   .iui-tab {
+    min-height: calc(var(--iui-size-l) * 2);
```
mayank99, line 113: continuing conversation from https://github.com/iTwin/iTwinUI/pull/860#discussion_r1064638779

we can define the 44px at top

```
$tab-borderless-min-height: calc(var(--iui-size-s) * 3 + var(--iui-size-xs)); // 44px
```

and then use it here
```suggestion
    min-height: $tab-borderless-min-height;
```

packages/itwinui-css/src/tabs/tabs.scss
mayank99: > `--_iui-borderless-horizontal-tabs-actions-content-gap` is being conditionally reassigned on [L220](https://github.com/iTwin/iTwinUI/pull/860/files#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR220). So I thought of leaving that as a CSS variable.

but does it need to be reassigned? you can use different sass variables for borderless vs default

packages/itwinui-css/src/tabs/classes.scss
rohan-kadkol: Unnested [`977dbaf` (#860)](https://github.com/iTwin/iTwinUI/pull/860/commits/977dbaf46789dffaf67da1306912536e2c17376e)

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: Implemented the above-mentioned change (https://github.com/iTwin/iTwinUI/pull/860/commits/0dcbbaf20de8ba8cd835a8a1774c54187384f627)

packages/itwinui-css/src/tabs/tabs.scss
rohan-kadkol: I changed the css variables to scss variables (https://github.com/iTwin/iTwinUI/pull/860/commits/409c080d852ee797724d49c449f38e0b9c5e101d).

I had to change the selectors a little. Not sure which version is better. Also, this pattern is slightly different from the pattern followed in `expandable_block.scss` as mentioned in the https://github.com/iTwin/iTwinUI/pull/860#discussion_r1060950174 thread.

Also, not sure if these are the best variable names ([L218-220](https://github.com/iTwin/iTwinUI/pull/860/commits/409c080d852ee797724d49c449f38e0b9c5e101d#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR218-R220)). Do you have any ideas for shorter names?

mayank99: > I had to change the selectors a little. Not sure which version is better. Also, this pattern is slightly different from the pattern followed in `expandable_block.scss` as mentioned in the [#860 (comment)](https://github.com/iTwin/iTwinUI/pull/860#discussion_r1060950174) thread.

ah yeah, so it's a tradeoff between shipping long css variables vs long selectors. i'm also not sure which one is better, but i'm fine with how it currently is.

> Also, not sure if these are the best variable names ([L218-220](https://github.com/iTwin/iTwinUI/pull/860/commits/409c080d852ee797724d49c449f38e0b9c5e101d#diff-0dba596e6381aa5a0bd620f0bb73c94ef6e65fcb3ce323f61bb25704a27a16adR218-R220)). Do you have any ideas for shorter names?

maybe `standard` should be `default` in the horizontal tabs vars. and for vertical it could be removed.
long sass variable names are ok to me, because they get compiled out in the final css.

packages/itwinui-css/src/tabs/tabs.scss
mayank99: any reason to define these in the nested selector instead of at the top of the file?

also, there is an `$iui-actions-gap` at the top, which now looks a bit confusing next `$tab-actions-gap`.

rohan-kadkol: > any reason to define these in the nested selector instead of at the top of the file?

I thought since `$tabs-actions-gap` is used just once, it would be better to keep it closer to its usage, instead of the top of the file.

> also, there is an `$iui-actions-gap` at the top, which now looks a bit confusing next `$tab-actions-gap`.

I went through all the used variable names, made them more self-explanatory and distinct from each other, and added them to the top of the file with comments. (https://github.com/iTwin/iTwinUI/pull/860/commits/7b56399c28519864018bfe2107330eca4cd4cd5b)

rohan-kadkol: As discussed, here are the latest spacing values for horizontal tabs:
* Default horizontal tabs:
  * Min height: `40px` (unchanged, since we want to minimize height [#860 (comment)](https://github.com/iTwin/iTwinUI/pull/860#discussion_r1062889038))
  * Min bottom spacing: `--iui-size-2xs` (`4px`)
* Borderless horizontal tabs:
  * Min height: `44px`
  * Min bottom spacing: `--iui-size-xs` (`8px`)
* Pill horizontal tabs:
  * Min height: `40px` (unchanged)
  * Min bottom spacing: N/A (Pills don't support adding actions)

<details>
<summary>See screenshots</summary>

image
image
image
image

</details>

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -191,3 +219,62 @@
...
+
+    &::before {
+      content: '';
+      flex: 1;
+    }
+
+    &::after {
+      content: '';
+      flex: 1;
+    }
```
mayank99, line 240: can be combined
```suggestion
    &::before,
    &::after {
      content: '';
      flex: 1;
    }
```

mayank99 (approved):
Looks good. Make sure tests are up to date and passing (in both CSS and the follow up React PR) before merging this.

FlyersPh9 (approved):

rohan-kadkol:
I noticed this PR has two approvals. Once I finish the React side of this PR, I will test that React PR using these CSS changes, and then merge this and the React PR. (https://github.com/iTwin/iTwinUI/pull/860#issuecomment-1374017199)

PR #861, opened 2022-12-16
chore: Improve `yarn dev` reliability

### LINKED ISSUE #747: `yarn dev` doesn't always build `global.css`
Sometimes when running `yarn dev`, it complains that it could not find `global.css`.

Screenshot from 2022-08-19 15-57-06

#### Steps to reproduce
1. Be in `mayank/use-variables` branch (at that time, the latest commit was `9759192f06ef310736d058f7b8db158d94f9087b`)
2. Make sure `yarn clean` is run in `packages/itwinui-css`
3. Run `yarn dev` in the root folder

#### Temporary workaround
Run `yarn build` in the root folder before `yarn dev`

#### Additional information
A potential explanation of why this is happening provided by Mayank:
> i don't think this is a documentation issue, i think it's because every component file includes `all.css`, whereas `index.html` includes `global.css`. the sass watch process only outputs `all.css`

rohan-kadkol:
Closes #747

### Issue

* `yarn dev` only creates `all.css`.
* But `index.html` imports `global.css` and `anchor.css`.
* Thus, it gives the file not found error in [#747](https://github.com/iTwin/iTwinUI/issues/747) when running `yarn dev` without `yarn build`.

### PR

* `index.html` imports `all.css` instead of `global.css` + `anchor.css`

<details>
<summary>See old solution description</summary>

* Added `dev:setup`: this builds (not watches) the three CSS files used by the index and component pages:
  * `anchor.css` & `global.css`: used by `index.html`
  * `all.css`: used by component pages
* `yarn dev` runs `yarn dev:setup` before starting the vite server.
* `build:watch:scss` now also listens for changes in `global.scss`

</details>


packages/itwinui-css/package.json
```diff
@@ -49,10 +49,11 @@
     "build": "yarn clean && yarn build:css && node ../shared/copyrightLinter.js --fix css/* && yarn build:html",
     "build:css": "node ./scripts/generateCss.js src css",
     "build:html": "vite build backstop/tests --config vite.config.ts",
-    "build:watch:scss": "sass --watch src/all.scss css/all.css",
+    "build:watch:scss": "sass --watch src/all.scss:css/all.css src/global.scss:css/global.css",
     "createComponent": "node ./scripts/createComponent.js",
     "clean": "rimraf ./css ./scss ./backstop/results/bitmaps_test",
-    "dev": "concurrently -n scss,html -k \"yarn build:watch:scss\" \"yarn vite backstop/tests --config vite.config.ts\"",
+    "dev:setup": "sass src/anchor/classes.scss:css/anchor.css src/all.scss:css/all.css src/global.scss:css/global.css",
+    "dev": "yarn dev:setup && concurrently -n scss,html -k \"yarn build:watch:scss\" \"yarn vite backstop/tests --config vite.config.ts\"",
```
mayank99, line 56: with these changes, there are too many places to keep track of. in the future i we need to add a button to index.html, there is a good chance we might forget to update one of these places.

it might just be simpler to change `index.html` to import `all.css` instead of `global.css` and `anchor.css`. the downside is that it will download more css than is needed, but it's not too much and is an acceptable tradeoff to me for maintainability.

rohan-kadkol: Sounds good, `index.html` now imports `all.css` and I removed the `dev:setup` script.

I initially did not do it this way to reduce the size of the downloaded css. Just for context, these are the values I got in `Inspector -> Network` when `index.html` imports:
* only `anchor.css`:
image
* `all.css`: 
image

mayank99: is that filtered by css? doesn't look right

i expect around <35kB with all.css and <10kB with anchor.css

packages/itwinui-css/backstop/tests/index.html
```diff
@@ -71,7 +71,7 @@
       @import url("@itwin/itwinui-variables");
       @import url("@itwin/itwinui-variables/os.css");
       @import url("@itwin/itwinui-css/global");
```
mayank99, line 73: `global` is not needed if we have `all`

rohan-kadkol: Removed the import of `global.css`. Also removed watching for changes in global.scss in the `build:watch:scss` script since there is no import of `global.css`. [`7b00ed5`](https://github.com/iTwin/iTwinUI/pull/861/commits/7b00ed5f4679fb0f1a357421dcaaf20b59df8bcb)

packages/itwinui-css/package.json
rohan-kadkol: I did not add a CSS filter. When filtering by CSS, it showed `0B` and no items in the list even after a hard refresh. Not sure why that is.

image

mayank99: you need to check the prod build. run `yarn build` followed by `yarn preview` inside `packages/itwinui-css`

mayank99 (approved):

packages/itwinui-css/package.json
rohan-kadkol: It still doesn't show the specific css files when applying the CSS filter. But these are the new values I got when `index.html` imports:

* `all.css`
image

* `global.css` and `anchor.css`
image

So now the size difference between the two imports is smaller, around `28kB transferred | 238 kB resources`

mayank99: ah, that's because the css is getting inlined into a `<style>` tag instead of a separate css file. also the total transferred size might be bloated by any browser extensions you have installed.

so you need to look at the size of the initial document.

| with `global`+`anchor` | with `all` |
| --- | --- |
| 6.3kB | 34.8kB |

packages/itwinui-css/package.json
```diff
@@ -49,10 +49,11 @@
     "build": "yarn clean && yarn build:css && node ../shared/copyrightLinter.js --fix css/* && yarn build:html",
     "build:css": "node ./scripts/generateCss.js src css",
     "build:html": "vite build backstop/tests --config vite.config.ts",
-    "build:watch:scss": "sass --watch src/all.scss css/all.css",
+    "build:watch:scss": "sass --watch src/all.scss:css/all.css src/global.scss:css/global.css",
```
mayank99: Reopening a tangential thread here.

I just realized (as I was doing the monorepo merge) that all our react components rely on individual imports, so we *will* need to build all files if we want the react components to hot reload when changing a css file.

This can be a separate PR/issue or you could incorporate it into the same one because it would allow you to use `global.css` and `anchor.css` in index.

rohan-kadkol: Now the `dev` script runs `yarn build:css` first that builds all the individual CSS files. I reverted `index.html` to import `global.css` + `anchor.css`

mayank99: this will not fully fix the problem. if any of the scss files change, it will not get rebuilt.

we want to incrementally compile all files (e.g. `sass --watch`) so that making a change in one of the `scss` files causes storybook (for example) to hot reload.

rohan-kadkol: I tried using something similar to `sass --watch` but couldn't come up with a proper command. Because the scss files are in subfolders (e.g. `src/alert/alert.scss`, `src/anchor/anchor.scss`) but we want the output CSS files to be directly in the CSS folder (e.g. `css/alert.css`, `css/anchor.css`)

```
sass src:css --watch
```

does this:
* `Compiled src\alert\alert.scss to css\alert\alert.css`
* `Compiled src\anchor\anchor.scss to css\anchor\anchor.css`
* ...

It also transforms the `classes.scss` and `index.scss` files. I guess that is why we were using the `scripts/generateCss.js` script and only sass watching the `all.scss`.

Also, not sure, but I think the `sass` command doesn't accept blob paths.

What command should I use?

gretanausedaite (commented):
No additional comments. 
Maybe this PR could wait after we merge repos, things we need might change?

rohan-kadkol:
> Maybe this PR could wait after we merge repos, things we need might change?

This PR now became a helper for the repo merge PR #868 (https://github.com/iTwin/iTwinUI/pull/861#discussion_r1052725440).

We could still do this PR after the merge repos PR, but would be good to have the hot reload for CSS working when the repos are merged.

packages/itwinui-css/package.json
mayank99: My first thought was to use a node [`child_process`](https://nodejs.org/api/child_process.html) for every file, but i think that is actually unnecessary. Sass already accepts multiple files as inputs and lets us specify the output path for each. So all we really need to do is generate those arguments programmatically and run that single command through `child_process`.

I've created https://github.com/iTwin/iTwinUI/pull/935 with that approach. Can you test if it does what you expect?

gretanausedaite:
This PR is replaced by #935 and can be closed.
@rohan-kadkol

PR #863, opened 2022-12-19
feat(SearchBox): Add Search Box

gretanausedaite:


packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * Example usages go here!
+ */
+export const SearchBox = (props: SearchBoxProps) => {
+  const { size, startAdornment, endAdornment, ...rest } = props;
+  useTheme();
+  return (
+    <div className='iui-input-flex-container' data-iui-size={size}>
+      {startAdornment}
+      <input className='iui-invisible-borders' type='search' {...rest} />
+      {endAdornment}
```
gretanausedaite, line 34: Sizing and state question:
1. Clone provided elements and add size and disabled attributes 
or 
2. Leave size and state for users to handle

packages/itwinui-css/src/utils/mixins.scss
```diff
@@ -142,3 +142,24 @@
...
+  &:focus,
+  &:focus-within,
+  &:hover {
+    outline: none;
+    border: none;
+  }
+}
+
+// Vertical divider for content, eg. input, combobox
+@mixin iui-vertical-divider {
```
gretanausedaite, line 160: 1. Move to utils? 
2. Create React component?

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,39 @@
...
+} & InputProps;
+
+/**
+ * // TODO:
+ * Searchbox component.
+ * Used for searches :)
+ * @example
+ * Example usages go here!
+ */
+export const SearchBox = (props: SearchBoxProps) => {
```
mayank99, line 27: so why is this component called searchbox? i think it can be used in many cases, not just search

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { size, startAdornment, endAdornment, ...rest } = props;
+  useTheme();
+  return (
+    <div className='iui-input-flex-container' data-iui-size={size}>
+      {startAdornment}
+      <input className='iui-invisible-borders' type='search' {...rest} />
```
mayank99, line 33: hmm, I think this `<input>` should be passed by the user so that they have more control over all DOM elements

packages/itwinui-css/src/utils/mixins.scss
```diff
@@ -142,3 +142,24 @@
...
   }
 }
+
+// Hide borders and hover/focus effects, eg. input
+@mixin iui-invisible-borders {
+  border: none;
+  background: none;
+
+  &:focus,
+  &:focus-within,
```
mayank99, line 152: i think `focus-within` will do nothing here because there can be no child of input

packages/itwinui-css/src/input/classes.scss
```diff
@@ -21,3 +25,33 @@
...
+
+  @each $status in positive, warning, negative {
+    &.iui-#{$status} {
+      @include iui-input-status($status: $status, $iconSelector: 'svg', $textSelector: 'input');
+    }
+  }
+}
+
+.iui-invisible-borders {
+  @include iui-invisible-borders;
```
mayank99, line 52: this should be autoapplied on input when used in container. that will make it possible for user to pass a regular `input`

packages/itwinui-css/src/input/classes.scss
```diff
@@ -21,3 +25,33 @@
...
 }
+
+.iui-input-flex-container {
+  @include iui-input-flex-container;
+
+  > input[type='search'] {
+    @include iui-search-box;
+  }
+
+  &[data-iui-size='small'] {
```
mayank99, line 36: could we handle sizing automatically if the user passes, for example, `<Input size='small'>` as child?

packages/itwinui-css/backstop/tests/input.html
```diff
@@ -502,5 +505,118 @@ <h2>No wrapping label</h2>
...
+      <h3>Current method</h3>
+      <div class="iui-input-container iui-inline-icon">
+        <input
+          placeholder="Placeholder…"
+          type="search"
+          class="iui-input"
+          id="test-1"
+        />
+        <svg-search
+          class="iui-input-icon"
```
mayank99, line 523: how will this `iui-input-icon` class be applied in react?

packages/itwinui-css/backstop/tests/input.html
```diff
@@ -502,5 +505,118 @@ <h2>No wrapping label</h2>
...
+          /></span>
+        <input
+          placeholder="Placeholder…"
+          type="search"
+          class="iui-invisible-borders"
+          id="test-1"
+          value="test search input"
+        />
+        <button
+          class="iui-button iui-end-icon"
```
mayank99, line 546: how will `iui-end-icon` be applied in react?

packages/itwinui-css/backstop/tests/input.html
```diff
@@ -502,5 +505,118 @@ <h2>No wrapping label</h2>
...
+      </div>
+      <div
+        class="iui-input-flex-container"
+        disabled
+      >
+        <input
+          placeholder="Placeholder…"
+          type="search"
+          class="iui-invisible-borders"
+          id="test-1"
```
mayank99, line 562: ids are duplicated

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: cloning has caused problems for our users and is discouraged by official react docs: https://beta.reactjs.org/reference/react/cloneElement

`disabled` makes sense for user to handle. `size` would be nice to handle. is it possible to do it without too much effort? if not, we could leave it for #969

packages/itwinui-css/src/utils/mixins.scss
mayank99: makes sense

you can also handle forced-colors
```css
  background-color: CanvasText;
}
```

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Because people want searchbox. This "search" layout could be achieved with no/minimal css changes (input with button) but devs still want separate search component.
Also, we are planning to expand this with functionality, like save 5 last searches.

CSS is not using search in the name, so we could create separate container for this new flex-container css and reuse it in SearchBox component.

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    &:autofill {
+      --_iui-input-background-color: var(--iui-color-background-accent-muted);
+      --_iui-input-border-color: var(--iui-color-border-accent);
+      --_iui-input-text-color: var(--iui-color-text-accent);
+    }
+
+    &[disabled] {
+      --_iui-input-background-color: var(--iui-color-background-disabled);
```
rohan-kadkol, line 56: The disabled searchbox is not getting a darker background color. The background is the same color as the enabled inputs.

packages/itwinui-css/backstop/tests/input.html
```diff
@@ -502,5 +505,118 @@ <h2>No wrapping label</h2>
       </div>
 
     </section>
+
+    <br />
+
+    <h2>Search box</h2>
```
rohan-kadkol, line 511: In the react side, you have examples with `startAdornment`. You could maybe have some examples of `startAdornment` in the css html page too.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * Used for searches :)
+ * @example
+ * Example usages go here!
+ */
+export const SearchBox = (props: SearchBoxProps) => {
+  const { size, startAdornment, endAdornment, ...rest } = props;
+  useTheme();
+  return (
+    <div className='iui-input-flex-container' data-iui-size={size}>
+      {startAdornment}
```
rohan-kadkol, line 32: Would there be styling issues (e.g. incorrect size, padding, etc.) if users pass non-ITwinUI custom elements (e.g. not `<Icon>`)? Do we need to specify a className on the css side that we add to the passed children using `React.cloneElement`?

mayank99: related conversation: https://github.com/iTwin/iTwinUI/pull/863#discussion_r1072277208

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: > Also, we are planning to expand this with functionality, like save 5 last searches.

How will this work? Our components should not contain any application logic, so it sounds a bit out of scope for us.

packages/itwinui-css/backstop/tests/input.html
gretanausedaite: Updated IDs.

packages/itwinui-css/backstop/tests/input.html
gretanausedaite: Removed from html as it is not needed. 
Researching how would work without <Icon> and with other buttons.

packages/itwinui-css/src/utils/mixins.scss
FlyersPh9: We have a vertical divider in the header between the brand button & breadcrumbs.  Can we reuse?

FlyersPh9:
I like that we're targeting `[type='search']` instead of a class modifier or data attribute. 👍

packages/itwinui-css/backstop/tests/input.html
```diff
@@ -502,5 +505,118 @@ <h2>No wrapping label</h2>
...
+      <h3>Flex box</h3>
+      <div
+        class="iui-input-flex-container"
+        data-iui-size="small"
+      >
+        <span>
+          <svg-search
+            class="iui-input-icon"
+            aria-hidden="true"
+          /></span>
```
rohan-kadkol, line 537: nit: Can the closing tag be on a new line? Or maybe you are doing this to save a line?

Same for L549, L552, etc.

mayank99: this syntax is actually invalid. html (unlike jsx) doesn't have self-closing tags, so you need to explicitly write out `<svg-search></svg-search>` even if there is nothing inside.

https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5

packages/itwinui-css/src/input/classes.scss
gretanausedaite: Including in input.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Users can pass input with children.

gretanausedaite:
Is it ok if SearchBox has default "search" button, users provides anything else they want as children:

image

https://user-images.githubusercontent.com/83585998/216359812-7e7b8089-ac7b-49ca-a108-eacbc3717c7d.mov

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+@mixin iui-expandable-searchbox {
+  width: var(--iui-component-height);
+  transition: width var(--iui-duration-3) ease;
+  overflow: hidden;
+
+  &.iui-animate-left {
+    justify-content: flex-end;
```
gretanausedaite, line 75: This is causing a slight shift of icon. Maybe @mayank99 or @FlyersPh9 knows how to fix it?
image

packages/itwinui-css/backstop/tests/input.html
gretanausedaite: Should be repo wide search because we are using self-closing tags a lot. Other PR.

packages/itwinui-css/backstop/tests/input.html
gretanausedaite: This class in not needed.

packages/itwinui-css/src/utils/mixins.scss
gretanausedaite: Removed.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Added animation to searchbox to make it more complex. Created InputFlexContainer for internal use.

packages/itwinui-css/src/input/input-flex-container.scss
FlyersPh9: I'm thinking it's because `.iui-expandable-searchbox` and the nested `.iui-button` are both the same width/height with a border.  If `.iui-expandable-searchbox` is 36x36px with a border that gives `.iui-button` only 34x34px to display (because 2px are used by the border).  I tested by changing `.iui-expandable-searchbox` to a width of 38px which does stop the icon shift.

If you make the border a `::before` I think that will allow you to keep the sizes as is without the icon shift.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,139 @@
...
+    ...rest
+  } = props;
+
+  const animatedVersion = () => {
+    return (
+      <>
+        {animateTo === 'right' && (
+          <IconButton
+            styleType='borderless'
+            onClick={() => onToggle(!isExpanded)}
```
mayank99, line 97: You don't need javascript for this. You can do it fully in CSS if you only use `<input>` (no `<button>`) and then expand on focus (using `:focus-within`).

Try replacing contents of App.tsx in the vite playground with this:
<details>

<summary>App.tsx</summary>

```jsx
import styled from '@emotion/styled';

const App = () => {
  return (
    <>
      <InputWrapper>
        <input />
      </InputWrapper>
    </>
  );
};

const InputWrapper = styled.div`
  display: inline-flex;
  border: 1px solid;
  background: Canvas;
  color: CanvasText;
  cursor: pointer;
  position: relative;
  height: 2.5rem;
  width: 3rem;
  transition: width 0.2s;

  > input {
    all: unset;
    width: 100%;
    padding-inline: 4px;
  }

  &::before {
    content: '';
    position: absolute;
    inset: 0;
    background-color: inherit;
    pointer-events: none;
  }

  &::after {
    --search-svg: url('data:image/svg+xml,<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">\
      <path d="M11 9.7c.7-1 1.1-2.2 1.1-3.5.1-3.5-2.7-6.2-6-6.2C2.7 0 0 2.7 0 6.1s2.7 6.1 6.1 6.1c1.3 0 2.5-.4 3.5-1.1l4.9 4.9 1.4-1.4zm-5 .5c-2.3 0-4.1-1.8-4.1-4.1S3.7 2 6 2s4.1 1.8 4.1 4.1-1.8 4.1-4.1 4.1"/>\
    </svg>');

    content: '';
    position: absolute;
    left: 50%;
    top: 50%;
    background-color: currentColor;
    mask: var(--search-svg) no-repeat center / contain;
    width: 1rem;
    height: 1rem;
    pointer-events: none;
  }

  &:focus-within {
    cursor: unset;
    width: 100%;
    border-color: hotpink;

    &::before,
    &::after {
      content: unset;
    }
  }
`;

export default App;

```
</details>

packages/itwinui-css/backstop/tests/input.html
mayank99: you can still use the correct tags in new code. 😅

mayank99: you can still use the correct tags in new code. 😅

.changeset/brave-otters-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
```
mayank99, line 2: why patch? should be minor

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-input-flex-container {
+  --_iui-input-background-color: var(--iui-color-background);
+  --_iui-input-border-color: var(--iui-color-border-foreground);
+  --_iui-input-text-color: var(--iui-color-text);
+
+  @include iui-reset;
+  @include iui-focus($offset: -2px, $thickness: 2px);
```
mayank99, line 9: iui-focus mixin will not work here because focus is not applied on this element, it is applied on the child

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+  width: 100%;
+  font-family: inherit;
+  font-size: var(--iui-font-size-1);
+  font-weight: var(--iui-font-weight-normal);
+  line-height: var(--iui-size-l);
+  border-radius: var(--iui-border-radius-1);
+  appearance: none;
```
mayank99, line 16: a lot of these styles, like `iui-reset`, `font-family`, `color`, `appearance`, `line-height`, etc are not needed on this element.

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+  height: var(--_iui-input-min-height);
+  padding-block: var(--_iui-input-padding-block);
+  padding-inline: var(--iui-size-s);
```
mayank99, line 19: padding should be set on the children if you want the expansion behavior to work correctly

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+
+  &[disabled],
+  > input[disabled] {
+    --_iui-input-background-color: var(--iui-color-background-disabled);
+    --_iui-input-border-color: var(--iui-color-border-disabled);
+    --_iui-input-with-icon-hover-border-color: var(--iui-color-border-disabled);
+    cursor: not-allowed;
+  }
+
+  > .iui-svg-icon {
```
mayank99, line 64: should not directly target other components here. I also wonder if we need to offer a padded version of the icon.

but for now, you can set the svg inside CSS as mentioned in my [other comment](https://github.com/iTwin/iTwinUI/pull/863#discussion_r1097980519)

packages/itwinui-css/src/utils/mixins.scss
```diff
@@ -142,3 +142,27 @@
     @content;
   }
 }
+
+// Hide borders and hover/focus effects, eg. input
+@mixin iui-invisible-borders {
```
mayank99, line 147: this doesn't need to go in utils.

also it be handled all in one line using `all: unset`

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+
+  right: unset;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 0;
+  margin: 0;
+
+  > input {
+    padding-left: var(--iui-size-s);
```
mayank99, line 34: i think padding value might be wrong (or maybe it's adding to parent's padding). doesn't match our input

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,92 @@
...
+    line-height: var(--iui-size-l);
+
+    &::placeholder {
+      --_iui-input-text-color: var(--iui-color-text-disabled);
+      user-select: none;
+      color: var(--_iui-input-text-color);
+    }
+  }
+
+  &:autofill,
```
mayank99, line 49: `:autofill` wont exist on this element, only on child

packages/itwinui-css/src/input/input.scss
```diff
@@ -81,3 +81,12 @@
     resize: none;
   }
 }
+
+@mixin iui-search-box {
```
mayank99, line 85: this mixin should probably be renamed. it gives the false impression that it contains the code for the whole searchbox component

packages/itwinui-css/src/input/input.scss
```diff
@@ -81,3 +81,12 @@
...
+  &::-webkit-search-decoration,
+  &::-webkit-search-cancel-button,
+  &::-webkit-search-results-button,
+  &::-webkit-search-results-decoration {
+    appearance: none;
```
mayank99, line 90: if we are removing the default browser functionality of clearing the input, then we should show an example of replacing it with a X button that also clears the input

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,139 @@
...
+    );
+  };
+
+  return (
+    <InputFlexContainer
+      className={cx({
+        'iui-expandable-searchbox': expandable,
+        'iui-animate-left': expandable && animateTo === 'left',
+      })}
+      aria-expanded={isExpanded}
```
mayank99, line 130: i think `aria-expanded` is invalid on a generic div. it usually goes on an interactive element (e.g. button)

apps/storybook/src/SearchBox.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+export const Basic: Story<SearchBoxProps> = (args) => {
+  return (
+    <SearchBox expandable {...args}>
+      <input type='search' placeholder='Search...' />
+    </SearchBox>
+  );
+};
+
+export const Static: Story<SearchBoxProps> = (args) => {
```
mayank99, line 40: this is interesting. shouldn't "expandable" be the special option, and the static one should be the default?

gretanausedaite: Probably, will update that.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: You can add a data attribute in addition to `:focus-within`. That will make it possible to make it user controlled.

For icon, I think it's actually fine to not support custom icons. It is possible to achieve through CSS variables, but not very important for initial featureset imo.

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    &::after {
+      content: unset;
+    }
+  }
+}
+
+@mixin iui-search-icon {
+  align-self: center;
+  margin: 0 var(--iui-size-s);
```
mayank99, line 53: this should probably be padding instead of margin, and the `align-self` can be handled in parent (using `align-items`)

.changeset/brave-otters-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
```
mayank99, line 4: missing: jsdocs, changesets, documentation page

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,54 @@
...
+    content: '';
+    position: absolute;
+    inset: 0;
+    background-color: inherit;
+    z-index: 3;
+    pointer-events: none;
+  }
+
+  &::after {
+    --search-svg: url('data:image/svg+xml,<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">\
```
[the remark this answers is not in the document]
gretanausedaite: Prefix'ed with iui

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-expandable-searchbox {
+  overflow: hidden;
+  width: var(--_iui-input-min-height);
+  transition: width var(--iui-duration-3) ease;
+
```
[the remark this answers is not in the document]
gretanausedaite: Added this

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,201 @@
...
+      className={cx({
+        'iui-expandable-searchbox': expandable,
+      })}
+      data-iui-size={size}
+      onFocus={(e) => {
+        onFocusProp?.(e);
+        onToggle?.(true);
+      }}
+      onBlur={() => {
+        onToggle?.(false);
```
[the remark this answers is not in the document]
gretanausedaite: Added onBlur prop

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,201 @@
...
+    inputProps,
+    ...rest
+  } = props;
+
+  return (
+    <InputFlexContainer
+      ref={ref}
+      as={Element}
+      className={cx({
+        'iui-expandable-searchbox': expandable,
```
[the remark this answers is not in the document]
gretanausedaite: Added classname prop

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,201 @@
...
+    isExpanded,
+    children,
+    inputProps,
+    ...rest
+  } = props;
+
+  return (
+    <InputFlexContainer
+      ref={ref}
+      as={Element}
```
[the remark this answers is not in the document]
gretanausedaite: Changed

playgrounds/vite/src/App.tsx
```diff
@@ -1,11 +1,68 @@
...
     <>
-      <Button>Hello world</Button>
+      <InputWrapper>
+        <input />
+      </InputWrapper>
     </>
   );
 };
 
+const InputWrapper = styled.div`
```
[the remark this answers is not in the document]
gretanausedaite: Cleaned up

packages/itwinui-react/src/core/utils/components/VerticalDivider.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const VerticalDivider = React.forwardRef<
+  HTMLDivElement,
+  VerticalDividerProps
+>((props, ref) => {
+  const { className, ...rest } = props;
+
+  useTheme();
+
+  return (
+    <span
```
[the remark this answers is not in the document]
gretanausedaite: Removed

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @supports not selector(:has(+ *)) {
+    &:where(:focus-within) {
+      outline: 2px solid var(--_iui-input-accent-border-color);
+      outline-offset: -2px;
+    }
+  }
+
+  &::before {
```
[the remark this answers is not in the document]
gretanausedaite: Border makes icon not centered when collapsed. Keeping before pseudo element.

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-input-flex-container {
+  --_iui-input-background-color: var(--iui-color-background);
+  --_iui-input-border-color: var(--iui-color-border-foreground);
+  --_iui-input-accent-border-color: var(--iui-color-border-accent);
+  --_iui-input-text-color: var(--iui-color-text);
+
+  &[disabled] {
```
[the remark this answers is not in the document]
gretanausedaite: Moved down

apps/storybook/src/SearchBox.stories.tsx
gretanausedaite: Added stories that should be usefull. Will update website when component is done.

packages/itwinui-css/src/input/input-flex-container.scss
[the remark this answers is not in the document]
mayank99: can you add a code comment then?

gretanausedaite:
TODO:
- [ ] Fix focus on open/close
- [ ] Tests
- [ ] Docs

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,66 @@
...
+  // }
+}
+
+/// Mixin for applying status colors to input container.
+/// @arg status - must be one of: positive, negative, warning
+@mixin iui-flex-input-status($status) {
+  --_iui-input-accent-border-color: var(--iui-color-border-#{$status});
+
+  &::before {
+    border-bottom: var(--iui-size-3xs) solid var(--_iui-input-accent-border-color);
```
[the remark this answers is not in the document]
gretanausedaite: Added

packages/itwinui-css/src/input/input-flex-container.scss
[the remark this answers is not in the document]
gretanausedaite: Removed this code.

packages/itwinui-css/src/input/searchbox.scss
gretanausedaite: Did those changes

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: removed

packages/itwinui-css/src/input/searchbox.scss
mayank99: it's still using margin

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Not relevant because we changed the direction of API.

packages/itwinui-css/src/input/classes.scss
gretanausedaite: Created a context for sizing: when user passes <Searchbox size='small' /> - buttons, inputs and icons are being rendered small.

packages/itwinui-css/backstop/tests/input.html
gretanausedaite: Not relevant because I've created SearchBox.Icon component for users to use.

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: input-flex-container css is reworked to have minimal css as needed.

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: Padding is not applied on the container anymore.

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: Autofill is not used anymore.

packages/itwinui-css/src/input/input.scss
gretanausedaite: Removed this functionality and using type='text' instead.

packages/itwinui-css/src/input/classes.scss
```diff
@@ -21,3 +25,34 @@
...
+    @include iui-input-size(small);
+  }
+
+  &[data-iui-size='large'] {
+    @include iui-input-size(large);
+  }
+
+  @each $status in positive, warning, negative {
+    &.iui-#{$status} {
```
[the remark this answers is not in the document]
gretanausedaite: Using data attributes. Should InputFlexContainer have `aria-disabled`?

.changeset/brave-otters-switch.md
gretanausedaite: Added jsdocs, changeset, will work on docs page.

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: Targeting only input - I think I could create new class for that to apply styles.
Also targeting collapsed/expanded states in searchbox.scss but you'll have to check if this approach is good enough for you - might need to rework that.

packages/itwinui-css/src/input/classes.scss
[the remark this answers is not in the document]
mayank99: no, because the container is a generic div.

`aria-disabled` is only applicable to specific roles listed [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles)

.changeset/brave-otters-switch.md
mayank99: current changeset is too much 😅 it needs to be concise enough to not take more than a few lines

i think the composition example can be removed, and the descriptions for all the subcomponents can also be removed

apps/storybook/src/SearchBox.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Input/SearchBox',
+} as Meta<SearchBoxProps>;
+
+export const Basic: StoryFn<SearchBoxProps> = () => {
+  return (
+    <Flex flexDirection='column'>
+      <SearchBox disabled inputProps={{ placeholder: 'Basic search' }} />
```
mayank99, line 29: can we make basic example actually basic? it's showing too much right now

apps/storybook/src/SearchBox.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+        </SearchBox.Button>
+      </SearchBox>
+    </Flex>
+  );
+};
+
+export const Expandable: StoryFn<SearchBoxProps> = () => {
+  return (
+    <Flex flexDirection='column' alignItems='start'>
+      <SearchBox expandable id='search-input' />
```
mayank99, line 56: no need for two separate searchboxes, every story should only have one

packages/itwinui-css/src/input/input-flex-container.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  input {
+    all: unset;
+    z-index: 2;
+    min-width: var(--_iui-input-min-height);
+    flex: 999;
+    padding-inline-start: var(--iui-size-s);
+
+    &:is(:not(:first-child)) {
+      padding-inline-start: var(--iui-size-2xs);
```
mayank99, line 47: what's the thinking behind modifying padding if not first child?

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-expandable-searchbox {
+  width: var(--_iui-input-min-height);
+  transition: width var(--iui-duration-3) ease;
+  isolation: isolate;
+  overflow: hidden;
+
+  &:has(.iui-searchbox-open-button:focus-visible) {
```
mayank99, line 9: needs fallback for firefox

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,43 @@
...
+    width: 100%;
+    border-color: var(--_iui-input-border-color);
+
+    .iui-searchbox-collapsed {
+      display: none;
+    }
+  }
+}
+
+@mixin iui-searchbox-expanded {
```
mayank99, line 30: this should be a data attribute

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+    >
+      {children ?? <SvgSearch />}
+    </Icon>
+  );
+}) as PolymorphicForwardRefComponent<'span', IconProps>;
+
+// ----------------------------------------------------------------------------
+
+export type SearchBoxInputProps = {
+  label?: string;
```
mayank99, line 217: not necessary as user can just pass `aria-label` and we don't do anything else with `label`

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+const SearchBoxButton = React.forwardRef((props, ref) => {
+  const { children, size, disabled: disabledProp, ...rest } = props;
+  const { size: sizeContext, disabled } = useSafeContext(SearchBoxContext);
+
+  return (
+    <IconButton
+      styleType='borderless'
+      size={size ?? sizeContext}
+      ref={ref}
+      disabled={disabledProp ?? disabled}
```
mayank99, line 270: i think you don't even need to destructure the `size` and `disabled` props, because `...rest` will override the first instance you set from context


same comment in other subcomponents

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+      {children ?? <SvgSearch />}
+    </IconButton>
+  );
+}) as PolymorphicForwardRefComponent<'button', IconButtonProps>;
+
+// ----------------------------------------------------------------------------
+
+/**
+ * SearchBox.Button component to add to your SearchBox.
+ */
```
mayank99, line 282: these jsdoc comments above types are unnecessary

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+    if (idProp && idProp !== inputId) {
+      setInputId(idProp);
+    }
+
+    return (
+      <input
+        id={idProp ?? inputId}
+        aria-label={label}
+        ref={useMergedRefs(ref, inputRef)}
```
mayank99, line 241: this is neat! avoids the need to create an intermediate variable 👍

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+      disabled: disabledProp,
+      ...rest
+    } = props;
+
+    const { inputId, setInputId, disabled, inputRef } =
+      useSafeContext(SearchBoxContext);
+
+    if (idProp && idProp !== inputId) {
+      setInputId(idProp);
+    }
```
mayank99, line 235: this should be moved into useEffect. see #1236

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+        setInputId,
+        openButtonRef,
+      }}
+    >
+      <InputFlexContainer
+        ref={ref}
+        className={cx({ 'iui-expandable-searchbox': expandable }, className)}
+        data-iui-size={size}
+        disabled={disabled}
+        role='searchbox'
```
mayank99, line 155: there should be no `role=searchbox` here, it should go on the `<input>` element

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+    setLocalExpanded(false);
+    onToggle?.(false);
+    openButtonRef.current && openButtonRef.current.focus();
+  };
+
+  const onOpen = () => {
+    setLocalExpanded(true);
+    onToggle?.(true);
+    inputRef.current && inputRef.current.focus();
+  };
```
mayank99, line 135: focus can be fixed using `queueMicrotask`
```suggestion
  const onClose = () => {
    setLocalExpanded(false);
    onToggle?.(false);
    queueMicrotask(() => openButtonRef.current?.focus());
  };

  const onOpen = () => {
    setLocalExpanded(true);
    onToggle?.(true);
    queueMicrotask(() => inputRef.current?.focus());
  };
```

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+export type SearchBoxIconProps = PolymorphicComponentProps<'span', IconProps>;
+
+const SearchBoxIcon = React.forwardRef((props, ref) => {
+  const { size, className, children, ...rest } = props;
+  const { size: sizeContext } = useSafeContext(SearchBoxContext);
+
+  return (
+    <Icon
+      aria-hidden
+      className={cx('iui-svg-icon', 'iui-search-icon', className)}
```
mayank99, line 204: `iui-svg-icon` should not be needed as it's already present in `Icon`

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+  return (
+    <SearchBoxButton
+      ref={ref}
+      aria-label='Close searchbox'
+      size={size ?? sizeContext}
+      disabled={disabledProp ?? disabled}
+      onClick={(e) => {
+        onClickProp?.(e);
+        onClose?.();
```
mayank99, line 311: all of the event handlers need to check for `isDefaultPrevented`

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+              <SearchBoxInput {...inputProps} />
+              <SearchBoxIcon />
+            </>
+          ))}
+        {expandable && (
+          <>
+            <div className='iui-searchbox-collapsed'>
+              {collapsedState ?? <SearchBoxOpenButton {...openButtonProps} />}
+            </div>
+            <div className='iui-searchbox-expanded'>
```
mayank99, line 171: there is no need for `iui-searchbox-collapsed` and `iui-searchbox-expanded` classes. It should be enough to remove/add the actual elements from the DOM like this:

```
!expanded ? collapsedState : expandedState
```

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+   */
+  expandable?: boolean;
+  /**
+   * Searchbox state toggle.
+   */
+  isExpanded?: boolean;
+  /**
+   * Function that is called on toggling (expand, collapse) searchbox.
+   */
+  onToggle?: (isExpanding: boolean) => void;
```
mayank99, line 77: this is probably a nit/subjective but i think instead of `isExpanding: boolean`, we could use a string union `state: 'expanded' | 'collapsed'`

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,398 @@
...
+} from '../utils';
+import type {
+  IconProps,
+  PolymorphicForwardRefComponent,
+  PolymorphicComponentProps,
+  InputFlexContainerProps,
+} from '../utils';
+import { IconButton } from '../Buttons/IconButton';
+import type { IconButtonProps } from '../Buttons/IconButton';
+import '@itwin/itwinui-css/css/input.css';
```
mayank99, line 25: missing searchbox css import

mayank99 (commented):
Looks pretty close! Two important comments about focus and extra divs, and a bunch of minor comments.

packages/itwinui-css/src/input/input-flex-container.scss
gretanausedaite: It seemed a bit too big gap between button and input in not adjusted version:
image

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Thanks! Updated code to use this structure.
Added isExpanded and localExpanded checks for manual control.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Moved to input component.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Using queueMicrotask! That worked really well. Thanks!

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Moved jsdoc comments a bit.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Removed label prop.
We used it for aria-label in case users won't add that on their own. I guess it's ok to trust them to add label or aria label to the searchbox.

packages/itwinui-css/src/input/searchbox.scss
gretanausedaite: removed this mixin

packages/itwinui-css/src/input/searchbox.scss
gretanausedaite: Added fallback. Seems a bit weird behaviour with ::focus-within styles it gets from flex container.

apps/storybook/src/SearchBox.stories.tsx
gretanausedaite: Updated stories to be one searchbox per story.

apps/storybook/src/SearchBox.stories.tsx
gretanausedaite: Updated stories

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Removed this class. You're right - not needed.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Do you have a hard preference on `onOpen` vs `onExpand`?

mayank99: lets do a poll in the team with a few options?

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: you can create a new variable and use that as the source of truth everywhere

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: it's not needed in the components either, because it's already present inside Object.assign at the end

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Team voted Expand/collapse

mayank99: so change buttons to "ExpandButton" and "CollapseButton"?

gretanausedaite: Changed to have `isExpanded`, `onExpand`, `onCollapse`, `SearchBox.ExpandButton` and `SearchBox.CollapseButton`

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Using this magic var.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Removed.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Removed size and disabled props.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: what do you think about making a utility function to combine event handlers like this?

```jsx
<button
  onClick={composeEventHandlers(onClickProp, onClose)}
/>
```
it will call the first argument, then check for `defaultPrevented`, then early return or call the other arguments

packages/itwinui-css/src/input/searchbox.scss
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-expandable-searchbox {
+  width: var(--_iui-input-min-height);
+  transition: width var(--iui-duration-3) ease;
```
FlyersPh9, line 5: In response to https://github.com/iTwin/iTwinUI/pull/863#issuecomment-1535067325, how about this?
```suggestion
  transition: width var(--iui-duration-1) ease-in-out;
```

mayank99: This is serviceable for now, but in the future I would like to use a custom cubic-bezier function. The default ones feel too bland

gretanausedaite: In Toasts we are using `cubic-bezier(0.4, 0, 1, 1)` but that might be too long for quick animations.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: Added to useEffect

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: That would be nice to have. However I don't want to change any other components in this PR. So IF I create this util function, I would only add it to SearchBox.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: > * Did we ever decide on whether default basic state should have search icon at the end or start?

UX team was asked this question. If I remember correctly, no direct answer provided. There are more cases of search icon at the end of the searchbox in products that our company creates, based on research (articles and examples in wide world of internet) - it's better to have search icon at the start. @FlyersPh9 Do you have any thoughts?

> * For expandable search, what is the expected way to set width?

Expandable search takes width of the container or set maxWidth on itself?
With grid:

https://user-images.githubusercontent.com/83585998/236450725-42b62e4b-0870-4d91-9f37-faca45f8e634.mov

With PageLayout:

https://user-images.githubusercontent.com/83585998/236450772-7c9dc09d-d92b-4ebb-b1aa-05eb67204d37.mov

With maxWidth:

https://user-images.githubusercontent.com/83585998/236450796-d19b4d35-9215-4b4e-9d98-6915d7a57d77.mov


> * Should we have any guidance on where/how expandable search should be used? Because if there are items next to the searchbox, then it is going to cause a lot of layout shift and it is also very expensive for the browser to compute.

Yes, we should. I found in old patterns site that expandable searchbox goes OVER content next to it. Will look into it more and add examples of usage in docs website.

> * Do we have any precedent on the animation? Right now the animation feels too slow and linear. Could change the duration and the easing function. (question for @FlyersPh9 i guess)

Answered in FlyersPh9 comment.

> * Are we going to expose InputFlexContainer publicly in v2? Or does it need more work and we should save it for after initial v3 release?

It should be good to be exposed, if we want to expose it.

gretanausedaite:
> * Did we ever decide on whether default basic state should have search icon at the end or start?

UX team was asked this question. If I remember correctly, no direct answer provided. There are more cases of search icon at the end of the searchbox in products that our company creates, based on research (articles and examples in wide world of internet) - it's better to have search icon at the start. @FlyersPh9 Do you have any thoughts?

> * For expandable search, what is the expected way to set width?

Expandable search takes width of the container or set maxWidth on itself?
With grid:

https://user-images.githubusercontent.com/83585998/236450725-42b62e4b-0870-4d91-9f37-faca45f8e634.mov

With PageLayout:

https://user-images.githubusercontent.com/83585998/236450772-7c9dc09d-d92b-4ebb-b1aa-05eb67204d37.mov

With maxWidth:

https://user-images.githubusercontent.com/83585998/236450796-d19b4d35-9215-4b4e-9d98-6915d7a57d77.mov


> * Should we have any guidance on where/how expandable search should be used? Because if there are items next to the searchbox, then it is going to cause a lot of layout shift and it is also very expensive for the browser to compute.

Yes, we should. I found in old patterns site that expandable searchbox goes OVER content next to it. Will look into it more and add examples of usage in docs website.

> * Do we have any precedent on the animation? Right now the animation feels too slow and linear. Could change the duration and the easing function. (question for @FlyersPh9 i guess)

Answered in FlyersPh9 comment.

> * Are we going to expose InputFlexContainer publicly in v2? Or does it need more work and we should save it for after initial v3 release?

It should be good to be exposed, if we want to expose it.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
mayank99: lets do it in a separate PR then

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: This was worked on in pair program session.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,420 @@
...
+        expandable,
+      }}
+    >
+      <InputFlexContainer
+        ref={ref}
+        className={cx({ 'iui-expandable-searchbox': expandable }, className)}
+        data-iui-size={size}
+        isDisabled={isDisabled}
+        data-iui-expanded={isExpanded}
+        aria-label='SearchBox'
```
gretanausedaite, line 158: @mayank99 What do you think about adding this here?

mayank99: I guess just make it say "Search" for now. But in our documentation, we will need to clearly mention all the places we are setting aria-label

mayank99: This is applied on the container, which is not right. The container is a generic `<div>`, so adding `aria-label` to it does nothing.

packages/itwinui-css/src/input/searchbox.scss
gretanausedaite: Added ease-in-out for now.

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: It announces "SearchBox, group" on the first visit.
Without aria-label here:
https://user-images.githubusercontent.com/83585998/236845633-149629c6-8af8-4b93-b4bb-653da0c53c81.mov

With aria-label here:
https://user-images.githubusercontent.com/83585998/236845744-2915f9c4-965e-4e2e-ab53-d743535aa0e2.mov

mayank99: i don't think it will work reliably in all cases and it is also not recommended. https://www.htmhell.dev/adventcalendar/2022/7/#non-interactive-elements

if you want to have an aria-label, try also adding `role=region`; or remove it.

gretanausedaite: Removed it. We can always add it later.

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
```diff
@@ -8,6 +8,7 @@ import type {
   PolymorphicComponentProps,
   PolymorphicForwardRefComponent,
 } from '../props.js';
+import '@itwin/itwinui-css/css/utils.css';
```
mayank99, line 11: This should be `input-container.css`

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -37,3 +37,24 @@
 .iui-input-label {
   @include iui-input-label;
 }
+
+// ----------------------------------------------------------------------------
+
+.iui-input-flex-container {
```
mayank99, line 43: Since this file now includes input-flex-container code, you can modify my [changeset](
https://github.com/iTwin/iTwinUI/blob/main/.changeset/mighty-cycles-sparkle.md) to just say "Reorganized source file structure. Should have no changes in the CSS output".

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
gretanausedaite: Updated

packages/itwinui-css/src/input-container/input-container.scss
gretanausedaite: Updated

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -43,13 +43,19 @@
...
   &[data-iui-size='small'] {
     @include iui-input-flex-size(small);
+    // To match small borderless button left padding + border
+    --_iui-flex-input-icon-padding: 5px;
   }
 
   &[data-iui-size='large'] {
     @include iui-input-flex-size(large);
+    // To match large borderless button left padding + border
+    --_iui-flex-input-icon-padding: 16px;
```
mayank99, line 58: these are only used in `searchbox.scss` so I think should be declared there

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -37,3 +37,30 @@
 .iui-input-label {
   @include iui-input-label;
 }
+
+// ----------------------------------------------------------------------------
+
+.iui-input-flex-container {
+  @include iui-input-flex-container;
```
mayank99, line 44: most of these mixins (`iui-input-flex-container`, `iui--input-flex-size`,...) can be removed and the code can be used directly where the classes are

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-expandable-searchbox {
+  width: var(--_iui-input-min-height);
+  transition: width var(--iui-duration-1) ease-in-out;
+  isolation: isolate;
```
mayank99, line 6: this should go in `input-flex-container.scss` because there is no z-index in this file

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -0,0 +1,43 @@
...
+  overflow: hidden;
```
mayank99, line 7: not needed anymore
```suggestion
```

mayank99 (approved):
LGTM 🚀

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,424 @@
...
+      size?: 'small' | 'large';
+      /**
+       * Context prop for disabling subcomponents
+       */
+      isDisabled?: boolean;
+      /**
+       * Id to pass to input
+       */
+      inputId: string;
+      setInputId: (inputId: string) => void;
```
elephantcatdog, line 41: `setInputId` is missing a comment

apps/storybook/src/SearchBox.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <SearchBox.Button aria-label='Previous result'>
+          <SvgCaretUpSmall />
+        </SearchBox.Button>
+        <SearchBox.Button aria-label='Next result'>
+          <SvgCaretDownSmall />
+        </SearchBox.Button>
+        <Divider orientation='vertical' />
+        <SearchBox.CollapseButton />
+      </SearchBox.ExpandedState>
+    </SearchBox>
```
elephantcatdog, line 85: The composition looks nice in use : )

elephantcatdog (approved):
Looks pretty good to me, just found one thing to fix

packages/itwinui-react/src/core/SearchBox/SearchBox.tsx
gretanausedaite: It is for context so internal use only.
Added comment.

PR #865, opened 2022-12-19
Revert "chore: Remove `index.html` anchor from demo page"

mayank99:
Reverts iTwin/iTwinUI#853

excluding it from the componentList results in the file not even being generated, so https://itwin.github.io/iTwinUI is currently 404ing


rohan-kadkol (approved):
Ohh, good catch. I will try another fix, but for now, we can revert PR #853

PR #875, opened 2022-12-21
ci: Add changesets

mayank99:
https://github.com/changesets/changesets


packages/itwinui-react/CHANGELOG.md
```diff
@@ -67,39 +73,40 @@ iTwinUI no longer supports Internet Explorer. The build output now targets `es20
 
 As for breaking API changes, there is only one:
 
-* **Table:** `columns` prop must now be an array. First level `Header` is no longer required ([#935](https://www.github.com/iTwin/iTwinUI-react/issues/935)) ([83d5cfe](https://www.github.com/iTwin/iTwinUI-react/commit/83d5cfe93980b628c79ec3951d05663a054699fc))
+- **Table:** `columns` prop must now be an array. First level `Header` is no longer required ([#935](https://www.github.com/iTwin/iTwinUI-react/issues/935))
+
 ```diff
-- const columns = { Header: 'Table', columns: [{ accessor: 'name', Header: 'Name' }, … ] };
+- const columns = [{ Header: 'Table', columns: [{ accessor: 'name', Header: 'Name' }, … ] }];
```
mayank99, line 79: this was a typo fix

packages/itwinui-react/CHANGELOG.md
```diff
@@ -2,56 +2,62 @@
...
+- **Table:** Ctrl and checkbox clicks update start row of shift selection ([#889](https://www.github.com/iTwin/iTwinUI-react/issues/889))
 
-* **Table:** Ctrl and checkbox clicks update start row of shift selection ([#889](https://www.github.com/iTwin/iTwinUI-react/issues/889)) ([21900d4](https://www.github.com/iTwin/iTwinUI-react/commit/21900d42900f7a03305bf2040cc64ddb29361b2d))
+## 2.0.4 (2022-12-16)
 
-### 2.0.3 (2022-12-12)
+### Patch changes
 
-### Fixes
+- Use `import type` instead of `import { type }` for supporting older typescript versions ([#996](https://www.github.com/iTwin/iTwinUI-react/issues/996))
```
mayank99, line 38: added 2.0.4. it was missing in the changelog because it was released from a branch
https://www.npmjs.com/package/@itwin/itwinui-react/v/2.0.4

package.json
```diff
@@ -38,18 +38,20 @@
...
     "chokidar-cli": "^3.0.0",
     "concurrently": "^7.6.0",
+    "cross-env": "^7.0.3",
     "husky": "^8.0.2",
     "lint-staged": "^12.5.0",
     "prettier": "2.8.0",
     "rimraf": "^3.0.2",
-    "turbo": "~1.6.3",
-    "cross-env": "^7.0.3"
+    "turbo": "~1.6.3"
```
rohan-kadkol, line 54: This change can be removed

packages/itwinui-react/package.json
```diff
@@ -90,7 +90,6 @@
     "@typescript-eslint/parser": "^5.17.0",
     "babel-loader": "^8.2.2",
     "concurrently": "^5.3.0",
-    "configs": "*",
```
rohan-kadkol, line 93: What is the meaning of `*`? Does it mean any version?

package.json
mayank99: no need. this was automatically sorted when i ran `yarn add` to add changesets

packages/itwinui-react/package.json
mayank99: yes. 

i removed it because changesets was complaining about it being an "ignored" package in the config

gretanausedaite (commented):
> every user-facing PR will need to run yarn changeset locally

Can we automate this? Or is this just something we HAVE to remember to do?

veekeys:
> > every user-facing PR will need to run yarn changeset locally
> 
> Can we automate this? Or is this just something we HAVE to remember to do?

Usually tools have a possibility to check if there was any change done and then require a changeset to be generated, then you could have a pre commit hook and enforce that.
 It is possible to do with [beachball](https://microsoft.github.io/beachball/cli/change.html) by running `check` command.
From what I remember, changesets do not support that (which was the reason I decided not to use that for other project).

gretanausedaite (approved):
No additional comments.
We can always improve later when team tries it.

PR #868, opened 2022-12-19
chore: Merge iTwinUI-react into iTwinUI

mayank99:
This PR is a result of merging the whole iTwinUI-react repo into iTwinUI with `--allow-unrelated-histories`.

There were only 10-15 files with merge conflicts, mostly the shared stuff, like `CONTRIBUTING.md`, `turbo.json`, root `package.json` and github workflows.
   - There was nothing changed inside `packages` until after the merge commit, which means we can be sure the source code is left intact.
   - For most conflicts, i combined file contents in a way that makes sense. There may be some things lost though.
   - I didn't try to combine `release-please.yml` meaningfully and instead just removed it. In a future PR this will be replaced with `changeset`.
   - Stripped down the root README (was previously a symlinked version of the css readme) to just a few sentences.

For reviewers: I recommend visiting the [`monorepo-merge`](https://github.com/iTwin/iTwinUI/tree/monorepo-merge) branch to see what the final result will look like when it's merged into `main`. Also recommend doing a fresh clone and running your usual build/dev/test workflows.

This PR should be merged with a "merge commit" rather than "squash commit" so that history is carried over to the main branch.

---

Got hot reload working with some tweaks (see https://github.com/iTwin/iTwinUI/pull/861#discussion_r1052725440).

https://user-images.githubusercontent.com/9084735/208550797-311a17fb-b74a-4055-987e-c8f84997f6cc.mov

Also got CI organized and passing

---

#### Pending

- [x] Build for astro workspaces (see error in [CI run](https://github.com/iTwin/iTwinUI/actions/runs/3736305852/jobs/6340471650))
   - *Decided to exclude these from CI build for now. Can investigate in a separate PR*
- [x] Visual tests failing because of #846 and #847
    - *Updated images*
- [x] Delete release-please.yml (will otherwise potentially trigger a huge release when merged to main)
- [x] Amend all react commit messages to reference original PR links

PR #929, opened 2022-12-22
chore: Fix errors in astro workspaces

mayank99:
After #868, `astro-playground` and `website` were failing with cryptic errors.

These errors were happening because:
1. we force resolve `node-fetch` to `2.6.7` to fix a vulnerability (#453)
2. astro relies on `node-fetch` version `3` when using node 16. so it was trying to use `node-fetch` v2 as if it was v3, which of course doesn't work.

So to fix this, all I really did is updated a bunch of deps, most important one being `backstopjs`. Previously, we were on backstopjs `6.0.4` which uses puppeteer `12` which has a hard dependency on `node-fetch` version `2.6.5` (hence the need for force resolution). Upgrading backstopjs to `6.0.5` bumps puppeteer to `14` which no longer uses `node-fetch`, so the force resolution is not needed. This means astro can safely use `node-fetch` version `3`.

After this PR, we should be able to run everything on node 16 with no problems.

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -10,7 +10,7 @@
     "clean": "rimraf .turbo && rimraf node_modules && rimraf dist"
   },
   "dependencies": {
-    "@astrojs/mdx": "^0.11.5",
+    "@astrojs/mdx": "^0.13.0",
     "@astrojs/react": "^1.2.2",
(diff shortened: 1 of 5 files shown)


gretanausedaite (approved):
Did not notice any issues. Looks good to me.

PR #931, opened 2022-12-22
chore: Fix jest trying to run transformed test files

### LINKED ISSUE #874: Jest tests failing after `swc` builds esm files
#### Steps to reproduce
1. Run `yarn dev` in `packages/itwinui-react`
1. Then run `yarn test` in `packages/itwinui-react`

#### Actual behavior
In `packages/itwinui-react`, when running `yarn test` after running `yarn dev`, some of the jest tests fail.

image

All of the above failures give an error message similar to the one below

image

Here is the `ColorPicker.test.js` file that failed

image

#### Additional information
This is happening because jest is finding the `<component-name>.test.js` and running them. This is double work since jest also runs the `<component-name>.test.tsx` files with `ts-jest`.

#### Expected behavior
* Should not get the above error when running `yarn test`
* The `<component-name>.test.js` unit test files generated by `swc` should not run. Only the `<component-name>.test.tsx` files should run.
* We could also look into making `swc` not produce the unit test files. More info in [iTwinUI-react PR #969](https://github.com/iTwin/iTwinUI-react/pull/969) and [iTwin/iTwinUI-react@`684456b` (#969)](https://github.com/iTwin/iTwinUI-react/pull/969/commits/684456bf0953be1c582a97d1360b3559fac359a2) commit where `swc` generated the `<component-name>.test.js` units test files too.

#### Workaround

Ignore the test files in the `cjs` and `esm` folders in `jest.config.js`

image

mayank99:
Fixes #874 by ignoring telling jest to ignore the test files built by swc.

Also fixed formatting of jest config to remove extra newlines.

### DIFF
--- a/packages/itwinui-react/jest.config.js
+++ b/packages/itwinui-react/jest.config.js
@@ -2,76 +2,59 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
 // For a detailed explanation regarding each configuration property, visit:
-
 // https://jestjs.io/docs/en/configuration.html
-
 module.exports = {
   // All imported modules in your tests should be mocked automatically
-
   // automock: false,
 
   // Stop running tests after `n` failures
-
   // bail: 0,
 
   // The directory where Jest should store its cached dependency information
-
   // cacheDirectory: "",
 
   // Automatically clear mock calls and instances between every test
-
   // clearMocks: false,
 
   // Indicates whether the coverage information should be collected while executing the test
-
   collectCoverage: true,
 
   // An array of glob patterns indicating a set of files for which coverage information should be collected
-
   // collectCoverageFrom: undefined,
 
   // The directory where Jest should output its coverage files
-
   coverageDirectory: 'coverage',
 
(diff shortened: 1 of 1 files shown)


packages/itwinui-react/jest.config.js
```diff
@@ -209,11 +209,11 @@ module.exports = {
...
 
-  // testPathIgnorePatterns: [
-
-  //   "\\\\node_modules\\\\"
-
-  // ],
+  testPathIgnorePatterns: [
+    '\\\\node_modules\\\\',
+    '\\\\esm\\\\',
+    '\\\\cjs\\\\',
```
rohan-kadkol, line 215: Just curious, what is the reason for the backslashes? I'm guessing in L213 for example becomes `\\node_modules\\`, but not sure why the backslashes are needed.

mayank99: i actually could not even get it to work with that. changing it to `/` works. can you confirm on windows? go inside `packages/itwinui-react` and run `yarn dev` and `yarn test` in parallel terminals

rohan-kadkol: For me on Windows, it worked for both: `\\\\` and `/`

mayank99: ok, lets go with `/` then

rohan-kadkol (approved):

PR #935, opened 2022-12-27
chore: Make sass watcher compile every scss file

### LINKED ISSUE #747: `yarn dev` doesn't always build `global.css`

mayank99:
Updated the `dev` script in itwinui-css to run a custom node script instead of `sass --watch` directly.

This new script reads a list of all component folders and generates a correctly-formatted command to run through the [sass cli](https://sass-lang.com/documentation/cli/dart-sass#many-to-many-mode). This command is then run through a node [`child_process`](https://nodejs.org/api/child_process.html).

Separately, [`chokidar`](https://github.com/paulmillr/chokidar) is used to watch for additions/changes to the compiled css and postprocess it further using [`lightningcss`](https://lightningcss.dev/). Previously we were doing this step in a vite plugin in the vite config for the backstop server, but it is now being done as part of the watch script so that it works throughout the monorepo (i.e. storybook, website, etc).

All of this means that changing a `.scss` file now results in hot reloading react components which rely on individual css imports (unlike our html test pages which rely on `all.css`). This also fixes #747 since all files (including `global.css`) are now built by the dev script.

https://user-images.githubusercontent.com/9084735/208550797-311a17fb-b74a-4055-987e-c8f84997f6cc.mov


packages/itwinui-css/vite.config.ts
```diff
@@ -18,7 +25,7 @@ export default defineConfig({
       },
     },
   },
-  plugins: [lightningCssPlugin(), generateIndex(), addMetaTags(), minifyHtml()],
+  plugins: [generateIndex(), addMetaTags(), minifyHtml()],
```
mayank99, line 28: this is the _only_ change in this file.
```diff
-  plugins: [lightningCssPlugin(), generateIndex(), addMetaTags(), minifyHtml()],
+  plugins: [generateIndex(), addMetaTags(), minifyHtml()],
```

everything else happened automatically through prettier / copyrightLinter.

i think this is because it is now using the shared prettier config whereas previously it was using editorconfig (which was removed in #868). they have different line lengths (#280)

the copyright change makes sense but we might want to introduce a new prettier config for the css workspace to avoid formatting changes like this.

what's also interesting is that we run linter in CI and it was passing, meaning something is not working there. need to investigate this in a separate PR

gretanausedaite (commented):
Is this PR in addition to #861 ? Or we are going to close that one?

mayank99:
> Is this PR in addition to #861 ? Or we are going to close that one?

It replaces that one. I didn't close it because it's @rohan-kadkol's PR

gretanausedaite (approved):
Did not notice any issues with this update.

packages/itwinui-css/scripts/watchScss.mjs
```diff
@@ -0,0 +1,71 @@
...
+  if (event === 'add' || event === 'change') {
+    const { name } = path.parse(changedFilePath); // css/alert.css.map -> alert.css
+    const cssFilePath = path.join(outDir, name);
+    const css = fs.readFileSync(cssFilePath, { encoding: 'utf8' });
+
+    const { code } = lightningCss.transform({
+      minify: false,
+      sourceMap: false,
+      code: Buffer.from(css),
+      filename: `${path.parse(changedFilePath).name}.css`,
```
rohan-kadkol, line 63: Should the `.css` at the end be removed? Because `path.parse(changedFilePath).name` is itself `alert.css`

```suggestion
      filename: path.parse(changedFilePath).name,
```

Also, nit: You could probably reuse the variable on [L55](https://github.com/iTwin/iTwinUI/pull/935/files#diff-b109bb960d69d89dcb141e784a3c37b0958eb76b0735e4501203788b25033b8cR55)

packages/itwinui-css/scripts/watchScss.mjs
```diff
@@ -0,0 +1,71 @@
...
+      targets,
+    });
+
+    fs.writeFileSync(cssFilePath, code, { encoding: 'utf8' });
+
+    console.log(`Postprocessed ${changedFilePath} using lightningcss`);
```
rohan-kadkol, line 69: This gives the message:

```
Postprocessed css\alert.css.map using lightningcss
```

To confirm, are we post-processing the `.css.map` files or the `.css` files?

packages/itwinui-css/scripts/watchScss.mjs
mayank99: no, because the original file name is `alert.css.map`

packages/itwinui-css/scripts/watchScss.mjs
mayank99: changed to `alert.css`

packages/itwinui-css/scripts/watchScss.mjs
rohan-kadkol: Yes, thanks

rohan-kadkol (approved):

PR #936, opened 2022-12-27
chore: Add new portal containing links to all dev servers

mayank99:
Added a barebones `portal` workspace that produces a simple static site using astro.

Running `yarn dev` will now open a single dev server containing links to all the other dev servers. It has a distinct yellow favicon.

In #937, the prod output of this workspace will be hosted on github pages.

I expect we will want to add more things to this page, but that can be done in the future. The purpose of this PR is really to serve as a base for #937 because we *need* to deploy storybook.


FlyersPh9:
What do you think about having the dev favicon apply to all of the dev servers?
Screenshot 2022-12-29 at 12 49 39 PM

apps/portal/src/pages/index.astro
```diff
@@ -0,0 +1,56 @@
+---
+const isDev = import.meta.env.DEV;
+
+const title = isDev ? 'iTwinUI (local)' : 'iTwinUI';
+const faviconHref = isDev ? 'favicon-dev.svg' : 'favicon.svg';
+
+const links = [
+  { name: 'html pages', href: isDev ? 'http://localhost:5173' : 'css' },
```
gretanausedaite, line 8: We could add link to Layouts demo page to have all links in one place.

mayank99:
> What do you think about having the dev favicon apply to all of the dev servers?

Sure we can look into that in the future. It might be a bit tricky though. It was straightforward here because of Astro

apps/portal/src/pages/index.astro
mayank99: can we do that in a future PR? this is blocking #937 which is important for users

mayank99: also layouts doesn't exist in this repo, so we dont know about its dev server

gretanausedaite: Agree with future PR. 
It doesn't matter that layouts does not exist in this repo. You can add github pages link to both cases. It's just a better developer/user experience to have all packages demo pages in one place if this portal will be deployed as a front page for all devs.

gretanausedaite: Same with icons.

gretanausedaite (approved):

Issue #941, opened 2022-12-30
Website: improvement


gretanausedaite:
Issue to track some improvements needed in website:

- [ ] Short description of components needs to be updated. No more than 3 lines without the concat.
- [ ] Split `index.astro` code into more components. Right now it's a mess with everything in one place and a lot of repetitive code.
- [ ] Add version note somewhere in docs.
- [ ] Move card/tiles to React component (for better polymorphism support)
- [x] Add a "See also"/"Related" section with links to the related components ([comment](https://github.com/iTwin/iTwinUI/pull/944#issuecomment-1370099053)) #1185 
- [x] Ability to highlight specific lines in code blocks. Maybe using [`expressive-code`]( https://github.com/expressive-code/expressive-code).
- [x] Fix sitemap
- [ ] Add FAQ (from our wiki)
- [x] Add search functionality: #1365
- [ ] Add theme switcher to examples
- [x] Add "related content" to docs pages
- [ ] Collapsable side nav areas
- [ ] Document exported util functions, components and hooks
- [x] Scrollbar looks wider than in css/storybook examples
- [x] Home screen example be clickable #1191 
- [x] Add link to Figma component library. #1358
- [ ] Mention browser targets. ([comment](https://github.com/iTwin/iTwinUI/pull/1418#discussion_r1269099636))
- [x] Props table styling messed up ([comment](https://github.com/iTwin/iTwinUI/pull/1539#pullrequestreview-1611137568))
- [ ] Missing Props tables for sub components (example: [tabs](https://dev-itwinui.bentley.com/docs/tabs) only lists props for the legacy api)
- [ ] Add a basic usage under each (or most) "Usage" sections on the docs site. E.g. [Input](https://dev-itwinui.bentley.com/docs/input#usage), [ToggleSwitch](https://dev-itwinui.bentley.com/docs/toggleswitch#usage), [Breadcrumbs](https://dev-itwinui.bentley.com/docs/breadcrumbs#usage). ([#1999 (comment)](https://redirect.github.com/iTwin/iTwinUI/pull/1999#discussion_r1568969830))
- [ ] Maybe document how users can extend our theme colors to match their color theme. E.g. how to change their accent color but keep all other colors the same, etc.

Far future:
- [ ] Guide pages (short "tutorials" how to do things)

FlyersPh9:
Do you want help with this:

> Short description of components needs to be updated. No more than 3 lines without the concat.

[...]

PR #944, opened 2023-01-03
website: Combine components

gretanausedaite:
Combined:
- Select and LabeledSelect
- TextArea and LabeledTextArea
- Input and LabeledInput
- ProgressRadial and ProgressLinear


gretanausedaite:
Components that don't have unique thumbnail:
- DropdownButton
- IconButton
- IdeasButton
- SplitButton
- Label

Create thumbnails or combine all buttons into one big buttons doc page?

mayank99:
> Components that don't have unique thumbnail:
> 
> * DropdownButton
> * IconButton
> * IdeasButton
> * SplitButton
> * Label
> 
> Create thumbnails or combine all buttons into one big buttons doc page?

DropdownButton can go together with DropdownMenu.
IconButton and IdeasButton can go in Button.
SplitButton can go either in Button or in DropdownMenu, and then link back to the other.
Label can go with Input or Typography.

In the future we should make everything searchable so that it doesn't matter where what lives

apps/website/src/examples/Input.main.tsx
```diff
@@ -3,12 +3,14 @@
...
 import * as React from 'react';
-import { Input } from '@itwin/itwinui-react';
+import { Input, LabeledInput } from '@itwin/itwinui-react';
 
 export default () => {
   return (
-    <div>
+    <div style={{ display: 'flex', gap: '8px', flexDirection: 'column' }}>
       <Input placeholder='Basic input' />
+      <LabeledInput label='Labeled input' message='Input message' />
```
mayank99, line 12: I think we should only show one example, and only of `LabeledInput`. Other examples can be shown on the in the rest of the page.

apps/website/package.json
```diff
@@ -18,6 +18,7 @@
     "@fontsource/noto-sans": "^4.5.11",
     "@fontsource/noto-sans-mono": "^4.5.11",
     "@itwin/itwinui-react": "*",
+    "@itwin/itwinui-icons-react": "*",
```
mayank99, line 21: is this needed?

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -2,7 +2,7 @@
 title: Text Area
```
mayank99, line 2: This should be called "Textarea", single word.

apps/website/package.json
gretanausedaite: This was an attempt to solve this error:
image

Did not succeed.

mayank99: oh that one is inside the sandbox, so changing package.json will not help.

try adding it to [`LiveExample.tsx`](https://github.com/iTwin/iTwinUI/blob/main/apps/website/src/components/LiveExample.tsx#L88)

apps/website/src/examples/Button.main.tsx
```diff
@@ -3,40 +3,32 @@
...
+    </MenuItem>,
+    <MenuItem key={2} onClick={close}>
+      Item #2
+    </MenuItem>,
+  ];
+
   return (
     <div>
-      <div style={{ display: 'flex', gap: '24px', flexWrap: 'wrap', margin: '12px' }}>
-        <Button size='small'>Default</Button>
```
rohan-kadkol, line 12: I feel the current sandbox example is more intuitive than the changed one. Since the props table below is only for `Button` and not for `SplitButton` or `IconsButton`, displaying `SplitButton` and `IconButton` in the same sandbox can be confusing.

apps/website/src/examples/Input.main.tsx
gretanausedaite: Showing only one example in main example.

apps/website/src/examples/Button.main.tsx
gretanausedaite: Combined docs for buttons, showing different examples for different buttons.

gretanausedaite:
> DropdownButton can go together with DropdownMenu.

I can't think of a page name that fits both. Maybe add DropdowButton to buttons and leave menu separately?

gretanausedaite:
> We could also add a "See also"/"Related" section with links to the related components. 

Added task for this in #941

apps/website/package.json
mayank99: Can you use `latest` for both packages?

gretanausedaite: Updated

mayank99 (approved):

apps/website/src/pages/docs/button.mdx
```diff
@@ -17,6 +26,46 @@ import LiveExample from '~/components/LiveExample.astro';
...
+
+### Props
+
+<PropsTable path={frontmatter.propsPathDefault} />
+
+## Split button
+
+<LiveExample src={frontmatter.exampleCodeFileSplitButton} />
+
+Pressing this left side of the button will commit the action. Pressing the right side of the button will open a menu. Upon selecting an option from the menu, the action immediately takes place, and the left side of the button’s label will change to match the recent selection.
```
rohan-kadkol, line 41: > left side of the button’s label will change to match the recent selection

To confirm, is this valid? Because I don't believe the left button label changes when we click on a menu option.

apps/website/src/components/LiveExample.tsx
```diff
@@ -15,7 +15,11 @@ import { createRoot } from 'react-dom/client';
 import { ThemeProvider } from '@itwin/itwinui-react';
 import App from './App';
 import './styles.css';
-createRoot(document.getElementById('root')).render(<ThemeProvider theme='dark'><App /></ThemeProvider>);
+createRoot(document.getElementById('root')!).render(
```
rohan-kadkol, line 18: Does the `!` have something to do with null safety? Just curious about why is it needed.

apps/website/src/pages/docs/button.mdx
mayank99: i believe the wording was copied over from the old page (splitbutton.mdx). it doesn't really matter for this PR

eventually we will validate the wording of every page

apps/website/src/components/LiveExample.tsx
mayank99: yes, try it in a codesandbox with and without `!`
https://vite.new/react-ts

PR #950, opened 2023-01-04
chore: Fix prettier config in itwinui-css

mayank99:
Before #868, we were relying on `editorconfig` which sets a line length of 120. This is now set in prettier.

This will fix the unintentional formatting changes we saw in https://github.com/iTwin/iTwinUI/pull/935#discussion_r1057810187 and https://github.com/iTwin/iTwinUI/pull/860#discussion_r1060937554

Also noticed and fixed stylelint was missing from fixOnSave (bad merge in #868).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,0 @@
-[*]
-end_of_line = lf
-insert_final_newline = true
-indent_size = 2
-indent_style = space
-max_line_length = 120
-quote_type = single
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -27,7 +27,8 @@
     "selector-max-compound-selectors": 4,
     "selector-max-id": 1,
     "selector-no-qualifying-type": null,
-    "scss/dollar-variable-pattern": null
+    "scss/dollar-variable-pattern": null,
+    "max-line-length": 120
   },
   "ignoreFiles": ["lib/**/*", "**/*.html"],
   "customSyntax": "postcss-scss"
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -27,8 +27,7 @@
     "selector-max-compound-selectors": 4,
     "selector-max-id": 1,
     "selector-no-qualifying-type": null,
-    "scss/dollar-variable-pattern": null,
-    "max-line-length": 120
+    "scss/dollar-variable-pattern": null
   },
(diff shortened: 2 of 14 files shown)


.editorconfig
```diff
@@ -1,7 +0,0 @@
-[*]
```
rohan-kadkol, line 1: How does `.editorconfig` work? Do we have to install any [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)? The [documentation](https://editorconfig.org/#:~:text=To%20use%20EditorConfig%20with%20one%20of%20these%20editors%2C%20you%20will%20need%20to%20install%20a%20plugin.) mentions this, but `.vscode/extensions.json` doesn't include it, so not sure.

rohan-kadkol (commented):
You mentioned something about investigating/changing the CI at the end of https://github.com/iTwin/iTwinUI/pull/935#discussion_r1057810187. A little confused what you meant by that. So just confirming if CI is working as expected.

.editorconfig
mayank99: editorconfig has been removed. previously, we were using it through prettier, not directly. see docs: https://prettier.io/docs/en/options.html#print-width

mayank99:
> You mentioned something about investigating/changing the CI at the end of [#935 (comment)](https://github.com/iTwin/iTwinUI/pull/935#discussion_r1057810187). A little confused what you meant by that. So just confirming if CI is working as expected.

ah, i was referring to the copyright linter, which we run in CI here: https://github.com/iTwin/iTwinUI/blob/main/.github/workflows/build.yml#L90

will be fixed in #952

.editorconfig
```diff
@@ -1,7 +0,0 @@
...
-end_of_line = lf
-insert_final_newline = true
-indent_size = 2
-indent_style = space
-max_line_length = 120
-quote_type = single
```
rohan-kadkol, line 7: To confirm, are all these properties the default in prettier? Or do we need to override any of them in prettier to maintain consistent behavior?

mayank99: some of them are default, and most of them we set in `prettier-config.js` and `pretter-scss-config.js`.

rohan-kadkol (approved):

PR #952, opened 2023-01-04
chore: Fix and combine copyright linters

mayank99:
two commits:
- [`037a0fe`](https://github.com/iTwin/iTwinUI/commit/037a0fe576c9fec5ce079a45e8500117296e1865) moved a combined version of copyrightLinter to `scripts`, deleted old versions from `shared` and `configs`
- [`82bd6b`](https://github.com/iTwin/iTwinUI/commit/82bd6b12fcc82f6ed278857bff89dae7d896b014) ran `yarn lint:copyright` which added missing copyright headers to some `.js` files


FlyersPh9 (approved):

scripts/copyrightLinter.js
```diff
@@ -22,10 +36,10 @@ const copyrightLine2 = `See LICENSE.md in the project root for license terms and
...
 const copyrightBannerHtml = `<!--\n  ${copyrightLine1}\n  ${copyrightLine2}\n-->`;
 const copyrightBannerJs =
-  '/*---------------------------------------------------------------------------------------------\n ' +
-  `* ${copyrightLine1}\n ` +
-  `* ${copyrightLine2}\n ` +
-  '*--------------------------------------------------------------------------------------------*/';
+  '/*---------------------------------------------------------------------------------------------\n' +
+  ` * ${copyrightLine1}\n` +
+  ` * ${copyrightLine2}\n` +
+  ' *--------------------------------------------------------------------------------------------*/';
```
rohan-kadkol, line 42: Are these changes intended?

packages/itwinui-css/scripts/createComponent.js
```diff
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 const fs = require('fs');
-const { copyrightBannerScss, copyrightBannerHtml, copyrightBannerJs } = require('shared/copyrightLinter');
+const { copyrightBannerScss, copyrightBannerHtml, copyrightBannerJs } = require('./copyrightLinter');
```
rohan-kadkol, line 6:
```suggestion
const { copyrightBannerScss, copyrightBannerHtml, copyrightBannerJs } = require('../../../scripts/copyrightLinter');
```

scripts/copyrightLinter.js
mayank99: yes they are. i moved the whitespace around to be more readable. the end result is still the same

packages/itwinui-css/scripts/createComponent.js
mayank99: no, the path is correct. they both live in the same folder

mayank99: ah you're right. we really need to combine these two files 😅

PR #953, opened 2023-01-04
chore: Move `configs` from `packages` to `internal`

mayank99:
After this change, the `packages` folder will only contain packages that are published to the user.

The name `internal` is discussable.


mayank99:
I believe the node modules cache needs to be busted for CI to pass.

gretanausedaite (approved):

Issue #840, opened 2022-12-08
Accessible links with bigger clickable areas


mayank99:
Many of our components are often used with links. Lets look at two specific examples: a `MenuItem` and a `Tile`.

In the case of MenuItem, there is a built-in padding. When adding an anchor element inside it, it makes the clickable area smaller and also brings the default anchor styles.
```html
<li class="iui-menu-item">
  <a href='#'>...</a>
</li>
```
![](https://user-images.githubusercontent.com/9084735/206545436-2a425c7e-a91c-4583-9960-a22fe06ba519.png)

To work around this, users often add `<a>` around the `<li>` but that is no longer valid markup (`<li>` needs to be a direct child of `<ul>`/`<ol>`). It also adds multiple tab stops. 👎

In the case of Tile, users often want the whole tile to be clickable to they add `<a>` around the tile, which also breaks accessibility because screen readers often only read out the first element inside links. 👎

The accessible markup would look more like:
```html
<div class="iui-tile">
  <div class="iui-tile-thumbnail">...</div>
  <div class="iui-tile-content">
    <div class="iui-tile-name">
      <a href='#'>The title of the card</a>
    </div>
    <div class="iui-tile-description">The stuff outside the link</div>
  </div>
</div>
```

This is achievable by using a pseudo-element that increases the clickable portion to be as big as the wrapping component. 👍
```scss
.iui-tile, .iui-menu-item {
  position: relative;

  a::before {
    content: '';
    position: absolute;
    inset: 0;
  }
}
```

Note that this approach should be used with caution and there should be no other interactive elements inside the wrapping component. In the case of menu item we would also need to fix multiple tab stops (in JS).

Further reading: https://inclusive-components.design/cards/

gretanausedaite:
Tile portion was done with #959

mayank99:
List was done with #1106

PR #959, opened 2023-01-06
chore: Tile a11y with anchors

gretanausedaite:
Solving `Tile` part of this issue: #840 

#### Added:
- `iui-link-box` and `iui-link-action` class to add pseudocontent to action component.
- Added LinkBox component for wrapper class
- Added LinkAction component for pseudocontent
- Tile.Action component which should be used in actionable Tiles.

#### Testing:
Unit, visual tests.
Manual testing with screen reader

#### Docs:
Jsdocs,


packages/itwinui-react/src/core/Tile/TileAction.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * Polymorphic action component for actionable Tiles. Should be used in name.
+ * It is `button` by default. Can also be rendered as anchor link.
+ *
+ * @example
+ * <Tile
+ *   isActionable={isActionable}
+ *   name={<Tile.Action onClick={() => doOnClick()}>Tile name</Tile.Action>}
+ * />
+ */
+export const TileAction = <T extends React.ElementType = 'button'>(
```
mayank99, line 35: we have a utility for polymorphic component. you can see usage in `Flex.tsx`

packages/itwinui-react/src/core/Tile/TileAction.tsx
```diff
@@ -0,0 +1,52 @@
...
+  props: TileActionProps<T>,
+) => {
+  const { as: Element = 'button', className, children, ...rest } = props;
+  return (
+    <Element
+      tabIndex={-1}
```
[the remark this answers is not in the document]
mayank99: i'm also wondering if this should be a generic utility component, rather than specific to Tile

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -73,6 +73,25 @@
         transform: translateZ(0) scale(1.1);
       }
     }
+
+    position: relative;
+
+    .iui-tile-content {
+      position: unset;
```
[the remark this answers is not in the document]
mayank99: I think we need to support buttons. It's the reason I abandoned this PR: https://github.com/iTwin/iTwinUI/pull/994/files

packages/itwinui-react/src/core/Tile/TileAction.tsx
gretanausedaite: Using polymorphic util component.

packages/itwinui-react/src/core/Tile/TileAction.tsx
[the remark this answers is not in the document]
gretanausedaite: Moved to generic component. I thought about it too as it might be reused in MenuItem.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -73,6 +73,12 @@
         transform: translateZ(0) scale(1.1);
       }
     }
+
+    position: relative;
+
+    // .iui-tile-content {
+    //   position: unset;
+    // }
```
FlyersPh9, line 81: Don't forget to remove this before merging.

packages/itwinui-css/src/utils/component-action/component-action.scss
```diff
@@ -0,0 +1,18 @@
...
+/// If it contains a focusable element, make sure to revert/exclude these styles.
+@mixin component-action {
+  all: unset;
+  text-decoration: none;
+  color: inherit;
+  user-select: text;
+
+  &::after {
+    content: '';
+    position: absolute;
```
FlyersPh9, line 15: This `@mixin component-action` is an interesting idea, I like it!  Does it make sense to put `position: relative` above the pseudo `::after`?

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
```diff
@@ -0,0 +1,44 @@
...
+  'a',
+  PolymorphicActionOwnProps
+>;
+
+/**
+ * Polymorphic action component.
+ * It is rendered as `a` by default.
+ *
+ */
+export const PolymorphicAction = React.forwardRef((props, ref) => {
```
mayank99, line 29: Need to think about the naming. `PolymorphicAction` is too low-level, and even the JSDoc explains nothing.

Maybe you could list out any other names that you can think of, and do some research into other libraries? One that I've found is `LinkBox` + `LinkOverlay` from chakra-ui. 
https://chakra-ui.com/docs/components/link-overlay

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {
+  PolymorphicComponentProps,
+  PolymorphicForwardRefComponent,
+} from '../index';
+
+type PolymorphicActionOwnProps = {
+  /**
+   * Content in the action element.
+   */
+  children?: React.ReactNode;
```
mayank99, line 16: `children` doesn't need to be a special prop.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
```diff
@@ -0,0 +1,44 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import {
+  PolymorphicComponentProps,
+  PolymorphicForwardRefComponent,
+} from '../index';
```
mayank99, line 10: we should never import anything from `index` to avoid risking circular dependencies.

```suggestion
} from '../props';
```

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -137,148 +138,165 @@ export type TileProps = {
...
+     * Polymorphic action component for actionable Tiles. Should be used in name.
+     * It is `button` by default. Can also be rendered as anchor link.
+     *
+     * @example
+     * <Tile
+     *   isActionable={isActionable}
+     *   name={<Tile.Action onClick={() => doOnClick()}>Tile name</Tile.Action>}
+     * />
+     */
+    Action: PolymorphicAction,
```
mayank99, line 295: does it need to be a subcomponent now? i think users could just pass the generic component directly

packages/itwinui-css/src/utils/component-action/component-action.scss
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+/// Visually hides an element but keeps it accessible to screen readers.
+/// Because it applies some potentially invasive styles, use this mixin on a wrapper element for best results.
+/// If it contains a focusable element, make sure to revert/exclude these styles.
```
mayank99, line 6: these comments have nothing to do with the mixin :/

packages/itwinui-css/src/utils/component-action/component-action.scss
```diff
@@ -0,0 +1,18 @@
...
+@mixin component-action {
+  all: unset;
+  text-decoration: none;
+  color: inherit;
+  user-select: text;
```
mayank99, line 11: does `user-select` do anything here?

packages/itwinui-react/src/core/Tile/TileAction.tsx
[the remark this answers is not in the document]
mayank99: tabindex still needs to be fixed

packages/itwinui-css/src/utils/component-action/component-action.scss
mayank99: `position: relative` will go on the wrapper (Tile or MenuItem), not on the link.

packages/itwinui-react/src/core/Tile/TileAction.tsx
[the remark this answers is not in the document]
mayank99: after that, i expect we will need to update focus styling to use `:has` or `:focus-within`

apps/storybook/src/Tile.stories.tsx
```diff
@@ -197,6 +197,41 @@ Actionable.args = {
...
+  );
+};
+AnchorLink.argTypes = { ...Basic.argTypes };
+AnchorLink.args = {
+  ...Basic.args,
+  description:
+    'If you click on this stadium, it is going to open another page.',
+  isActionable: true,
+  moreOptions: undefined,
+};
```
elephantcatdog, line 234: Could you make this example actually "open" a new page, even if it's just the same page? In this storybook example, the text says "If you click on this stadium, it is going to open another page.", but it seems broken because nothing happens when it's clicked on. It could be like the links in the story fo `SkipToContentLink` or like in `Anchor - Basic`.

elephantcatdog (commented):
Not super related, but are the css visual tests 'with thumbnail' meaningful if we're not testing them with thumbnails? We're hiding the thumbnails because it caused test problems previously.

packages/itwinui-css/src/utils/component-action/component-action.scss
gretanausedaite: Yes, without it text gets selected on click.

packages/itwinui-css/src/utils/component-action/component-action.scss
gretanausedaite: removed

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: importing from props.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: It does not work without children so I kept it in props.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: `ActionOverlay`? I don't want to blindly copy from other libraries. Maybe we need team vote?

packages/itwinui-css/src/utils/component-action/component-action.scss
mayank99: This might be because `iui-tile-name` has `user-select: all`. We should remove that.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
mayank99: what doesn't work without children? if you check `Flex.tsx`, none of them have explicitly defined children.

i think you can just use `{}`

packages/itwinui-css/src/utils/component-action/component-action.scss
gretanausedaite: Removed select all

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: Added two components. LinkBox has tabIndex to skip it when tabbing.

mayank99: The naming is kinda reversed. Box should the outer wrapper, and Overlay should be inner one.

And why do you want to skip it when tabbing?

packages/itwinui-css/src/tile/tile.scss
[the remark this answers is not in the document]
mayank99: #1117 has been merged so this should be possible to achieve now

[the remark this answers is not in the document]
gretanausedaite: Removed tile-content css part.

packages/itwinui-css/src/tile/tile.scss
gretanausedaite: Removed :)

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: Reversed the naming.

In the article linked in the issue, when you tab through the tiles it tabs only once per tile. I think it's more smooth behavior - click on full tile to open a link, not same action in two places. I could be wrong.

apps/storybook/src/Tile.stories.tsx
gretanausedaite: Added link to story in docs tab.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -197,6 +197,41 @@ Actionable.args = {
...
+    thumbnail,
+    isNew,
+    isActionable = true,
+    ...rest
+  } = props;
+
+  return (
+    <Tile
+      name={
+        <Tile.Action onClick={() => console.log('clicked')}>{name}</Tile.Action>
```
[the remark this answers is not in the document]
gretanausedaite: Reworked this part of code.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: Removed tabIndex from Tile, should be good now.

packages/itwinui-react/src/core/Tile/TileAction.tsx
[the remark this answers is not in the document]
gretanausedaite: Updated styles for tile focus when selected state.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: I can use `{}` with eslint disabled

mayank99: `{}` can be assigned to a different type for better readability


this also makes it easier to add more props later if needed

packages/itwinui-css/src/utils/component-action/link-box.scss
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@import '../mixins';
+
+@mixin iui-link-box {
+  position: relative;
+  width: fit-content;
+  height: fit-content;
+  cursor: pointer;
```
mayank99, line 9: `cursor: pointer` should go on link-overlay.

`width:fit-content` and `height: fit-content` shouldn't be necessary. This class should have no opinion on the sizing. but i think you added this because `div`s automatically take up the available width. This can be fixed using the `as` prop.
```
<LinkBox as={Tile}>
```
```
<div class='iui-tile iui-link-box'>
```

however, in #1110, I've also added `position: relative` to the Tile, so maybe there is no need to even wrap the Tile in Linkbox.

packages/itwinui-css/src/utils/component-action/link-box.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  &:where(:focus-within) {
```
mayank99, line 11: the focus styling looks good for the most simple case when tabbing, but it will also show for mouse users. the solution there is to use `:has(:focus-visible)` and then add `:focus-within` fallback for browsers that don't support `:has`.

for more advanced cases, it looks a bit weird. am i focusing the tile (main action) or the secondary action inside it?


this can be fixed using `&:has(.iui-link-overlay:focus-visible)` but the fallback approach for this will still have the same problem. maybe that's ok? firefox will support `:has` by Q2 2023.

packages/itwinui-css/src/utils/component-action/link-box.scss
```diff
@@ -0,0 +1,26 @@
...
+    outline: var(--iui-size-3xs) solid var(--iui-color-border-accent);
+    border-radius: var(--iui-border-radius-1);
```
mayank99, line 13: border-radius doesn't belong here, because similar to `width`, this class should have no opinion on styling.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -181,7 +181,6 @@ export const Tile = (props: TileProps) => {
         className,
       )}
       aria-disabled={isDisabled}
-      tabIndex={isActionable && !isDisabled ? 0 : undefined}
```
mayank99, line 184: This will break keyboard navigation for existing user code because (in your current version) it requires them to add `<LinkOverlay as='button'>`.

Should we automatically do that when `isActionable` is true? we can also handle disabled behavior.

packages/itwinui-react/src/core/utils/components/LinkBox.tsx
```diff
@@ -0,0 +1,44 @@
...
+  'a',
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  {}
+>;
+
+/**
+ * Polymorphic action component.
+ * It is rendered as `a` by default.
+ */
+export const LinkOverlay = React.forwardRef((props, ref) => {
```
mayank99, line 22: will need tests for this.

also nit: i think filename should be `LinkOverlay` because it is the "meaty" part of the component. `LinkBox` is a simple wrapper.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -197,6 +204,46 @@ Actionable.args = {
...
+      <Tile
+        name={
+          <LinkOverlay href='/?path=/docs/core-tile--anchor-link'>
+            {name}
+          </LinkOverlay>
+        }
+        description={description}
+        metadata={metadata}
+        badge={badge}
+        thumbnail={thumbnail}
```
mayank99, line 231: when using arrow key navigation in screen readers, the contents of the thumbnail area (badges, quick options, img alt) will be announced first.

in https://inclusive-components.design/cards/#alternativetext, the suggestion is to use flex order to tackle this problem. can we try that?

packages/itwinui-css/src/utils/component-action/link-box.scss
gretanausedaite: Added suggested css rules.
I think it's ok, users can see the difference in moving focus. Also screen reader does a good job showing which one is selected.

packages/itwinui-css/src/utils/component-action/link-box.scss
gretanausedaite: Removed border radius and width.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -120,6 +131,21 @@ export type TileProps = {
...
+type TileActionOwnProps = {}; // eslint-disable-line
+
+export const TileAction = (
+  props: PolymorphicComponentProps<'a', TileActionOwnProps>,
+) => {
+  const tileContext = useSafeContext(TileContext);
+
+  React.useEffect(() => {
+    console.log('action useefect');
+    !tileContext.isDisabled ? tileContext.setActionable(true) : null;
```
mayank99, line 143: i've realized we can do this with `:has` and avoid an extra rerender (in new chrome/safari) and then use this approach only if `:has` is not supported. see comment here: https://github.com/iTwin/iTwinUI/pull/1127#discussion_r1134494887

also, i think there is no need to check `isDisabled` here. Tile can be both actionable and disabled at the same time, and disabled styling should take preference.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -137,154 +163,180 @@ export type TileProps = {
...
-        isNew={isNew}
-        status={status}
-      />
-
-      <span className='iui-tile-name-label'>{name}</span>
-    </div>
-  );
+    React.useEffect(() => {
+      console.log('tile useeffect');
+      setLocalActionable(isActionable);
```
mayank99, line 200: this use effect is not needed.

you can derive a value like this:
```js
const isActionable = isActionableProp ?? localActionable;
```

see https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -137,154 +163,180 @@ export type TileProps = {
...
-      {buttons && <div className='iui-tile-buttons'>{buttons}</div>}
-    </div>
-  );
-};
+          {buttons && <div className='iui-tile-buttons'>{buttons}</div>}
+        </div>
+      </TileContext.Provider>
+    );
+  },
+  { Action: TileAction },
```
mayank99, line 338: needs jsdoc

packages/itwinui-react/src/core/utils/components/LinkOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+      className={cx('iui-link-overlay', className)}
+      {...rest}
+    />
+  );
+  // eslint-disable-next-line @typescript-eslint/ban-types
+}) as PolymorphicForwardRefComponent<'a', LinkOverlayOwnProps>;
+
+LinkOverlay.displayName = 'LinkOverlay';
+
+export const LinkBox = React.forwardRef((props, ref) => {
```
mayank99, line 43: needs jsdoc

packages/itwinui-react/src/core/utils/components/LinkOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type LinkOverlayOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+type LinkBoxOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+export type PolymorphicLinkOverlayProps = PolymorphicComponentProps<
+  'a',
+  LinkOverlayOwnProps
+>;
+
+export type PolymorphicLinkBoxProps = PolymorphicComponentProps<
```
mayank99, line 20: these names should not have `Polymorphic` in them

packages/itwinui-react/src/core/Tile/Tile.tsx
gretanausedaite: Removed useEffect

packages/itwinui-react/src/core/Tile/Tile.tsx
gretanausedaite: Added check for ":has" functionality.

packages/itwinui-css/src/utils/component-action/link-box.scss
gretanausedaite: I'm not wrapping Tile in LinkBox. 

Tested with MenuItem, should work:
```js
<LinkBox as={MenuItem} key='2'>
  <LinkOverlay as='button' onClick={}>Item #2</LinkOverlay>
</LinkBox>
```

packages/itwinui-react/src/core/Tile/Tile.tsx
gretanausedaite: I think it should be good now. Adding isActionable class / prop.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -22,11 +22,35 @@ import { ProgressRadial } from '../ProgressIndicators';
...
+ * @example
+ * <Tile
+ *   name={<Tile.Action href='/new-page'>Tile name<Tile.Action/>}
+ * />
+ */
+export const TileAction = (
+  props: PolymorphicComponentProps<'a', TileActionOwnProps>,
+) => {
+  const tileContext = useSafeContext(TileContext);
+  const supportsHas = CSS.supports('selector(:has(+ *))');
```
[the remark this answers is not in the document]
gretanausedaite: Doing this

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: Using own props objects.

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: I think everything from this comment is solved.

mayank99: this comment is about naming. Have you tried to research into what other libraries are calling it and/or come up with other names yourself?

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
mayank99: you can remove these eslint comments, they are only needed in one place.

gretanausedaite: Oops

packages/itwinui-react/src/core/utils/components/PolymorphicAction.tsx
gretanausedaite: LinkBox for wrapper and LinkAction for action?
I have almost no ideas on naming. Did not find anywhere else besides chakra-ui.

mayank99: > LinkBox for wrapper and LinkAction for action?
i like it. sounds more intuitive than "overlay".

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -137,154 +173,185 @@ export type TileProps = {
...
+              'iui-new': isNew,
+              'iui-selected': isSelected,
+              'iui-actionable': isActionable,
+              [`iui-${status}`]: !!status,
+              'iui-loading': isLoading,
+            },
+            className,
+          )}
+          aria-disabled={isDisabled}
+          onClick={onClick}
```
mayank99, line 246: don't need this onClick anymore

gretanausedaite: This onClick is "div" event and I don't want to make LinkOverlay as div.

mayank99: i dont think there is much difference in a div event vs button event.

you might able to change the type of `onClick` to something like `React.MouseEventHandler<HTMLElement>`

packages/itwinui-css/backstop/results/bitmaps_reference/iTwinUI_tile_Type_Action_States_dark_0_demo-action-states_0_desktop.png
mayank99, line 1: Some of these images shouldn't have changed. Can you try reverting and see if tests pass?

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -137,154 +174,186 @@ export type TileProps = {
...
+              'iui-folder': variant === 'folder',
+              'iui-new': isNew,
+              'iui-selected': isSelected,
+              'iui-actionable': isActionable,
+              [`iui-${status}`]: !!status,
+              'iui-loading': isLoading,
+            },
+            className,
+          )}
+          aria-disabled={isDisabled}
```
mayank99, line 248: `disabled` (or `aria-disabled` if tackling #391 too) should go on the button, not the tile

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -10,12 +10,48 @@ import {
...
+  PolymorphicComponentProps,
+  useSafeContext,
+  getWindow,
 } from '../utils';
 import '@itwin/itwinui-css/css/tile.css';
 import { DropdownMenu } from '../DropdownMenu';
 import { IconButton } from '../Buttons';
 import { ProgressRadial } from '../ProgressIndicators';
 
+export const TileContext = React.createContext<
```
mayank99, line 23: doesn't need to be eported

packages/itwinui-css/src/utils/component-action/link-overlay.scss
mayank99, line 1: this file and the folder should be renamed to `link-action`

.changeset/lazy-ducks-lick.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+Added `Tile.Action` to take advantage of these new a11y components in Tile.
+
+Usage:
+```js
+<Tile 
+  name={<LinkAction as='button' onClick={() => {/* Do things */}}>Tile name that is also a button</LinkAction>}
+/>
+```
```
mayank99, line 21: I think this Tile.Action part should go in a separate changeset, and the example should show Tile.Action instead of LinkAction

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -169,6 +93,97 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
...
+@mixin iui-selected {
+  @include iui-focus($offset: 0, $thickness: var(--iui-size-2xs));
+  --_iui-tile-status-icon-fill: var(--iui-color-icon-accent);
+
+  outline: solid 2px var(--iui-color-border-accent);
+  background-color: var(--iui-color-background-accent-muted);
+}
+
+@mixin iui-actionable {
+  @include iui-focus($offset: var(--iui-size-3xs), $thickness: var(--iui-size-3xs));
```
mayank99, line 105: this `iui-focus` is no longer necessary as focus is not on the tile

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -169,6 +93,97 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
...
+    outline: var(--iui-size-3xs) solid var(--iui-color-border-accent);
+
+    &.iui-selected {
+      outline-width: var(--iui-size-2xs);
+    }
+  }
+
+  @supports not selector(:has(+ *)) {
+    &:where(:focus-within) {
+      outline: var(--iui-size-3xs) solid var(--iui-color-border-accent);
```
mayank99, line 117: add the `iui-selected` portion in this fallback too. Right now in Firefox, there is no way to tell that a selected tile is focused

packages/itwinui-react/src/core/utils/components/LinkAction.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import {
+  PolymorphicComponentProps,
+  PolymorphicForwardRefComponent,
+} from '../props';
+
```
mayank99, line 11: missing css import

packages/itwinui-react/src/core/Tile/Tile.tsx
gretanausedaite: Tile disabled styling is applied with aria-disabled on Tile.
I can add aria-disabled to the button but I won't remove them from Tile. 
Yes, we can update css, but I think it's future PR problem, this one already has a lot.

packages/itwinui-react/src/core/utils/components/LinkAction.tsx
gretanausedaite: Added css import

packages/itwinui-css/src/tile/tile.scss
gretanausedaite: Added iui-selected portion

packages/itwinui-css/src/tile/tile.scss
gretanausedaite: Removed

packages/itwinui-css/src/utils/component-action/link-overlay.scss
gretanausedaite: renamed

packages/itwinui-react/src/core/Tile/Tile.tsx
gretanausedaite: Is not exported anymore

packages/itwinui-react/src/core/Tile/Tile.tsx
mayank99: right now the clicks are not prevented so yeah you could add `aria-disabled` on the button and conditionally pass `onClick` only when not disabled.

then there is also a bigger problem that `aria-disabled` does nothing on a generic div. you can see which roles it affects here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles
so in the future we can use a data attribute instead.

packages/itwinui-css/backstop/results/bitmaps_reference/iTwinUI_tile_Type_Action_States_dark_0_demo-action-states_0_desktop.png
gretanausedaite: Updated. Was missing outline-offset

elephantcatdog (approved):
Looks good to me!

PR #963, opened 2023-01-06
website: Update avatar documentation & demos

FlyersPh9:
- Add avatar demos for the different options.


FlyersPh9:
Once #1040 merges, this will get unblocked.

apps/website/src/examples/Avatar.picture.tsx
```diff
@@ -0,0 +1,18 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Avatar, getUserColor } from '@itwin/itwinui-react';
+
+export default () => {
+  return (
+    <Avatar
+      abbreviation='GB'
+      backgroundColor={getUserColor('Greg Bentley')}
+      image={<img src='https://www.bentley.com/wp-content/uploads/greg-bentley-hr-profile.jpeg' />}
```
FlyersPh9, line 13: Where can I host an image to use instead of this one from bentley.com?  I don't want to use https://itwin.github.io/iTwinUI/css/assets/user-placeholder.png because it looks too close to the SVG examples.

mayank99: you can put it in the `public` folder under `apps/website` and it should be available at this url: `https://itwinui.bentley.com/<your-file-name.jpg>` (after it's deployed)

if you decided to do this, i would recommend _not_ using a human face (even if it's AI-generated). Instead, you could use something like this: https://stock.adobe.com/search?k=robot+avatar&asset_id=359684982

mayank99: oh i just realized this is for the demos. we probably don't want to show the full domain in the code, so what you really need is to put an image inside the sandbox.

there might be cleaner ways of doing this, but one easy option is to convert it to a [data uri](https://css-tricks.com/data-uris/) and use it in the `src` attribute of the img.
```
<img src='data:image/...'>
```

the url is not the focus, so it gets the point across (that you can use an  tag)

FlyersPh9: Good suggestion, replaced with that robot & used a data URI.

mayank99: oh wow that url is incredibly long. maybe move it out to a variable
```
const imgSrc = `data:image/...`;
```
```
<img src={imgSrc} />
```

apps/website/src/examples/Avatar.icon.tsx
```diff
@@ -0,0 +1,19 @@
...
+import * as React from 'react';
+import { Avatar, getUserColor } from '@itwin/itwinui-react';
+import { SvgBentleySystems } from '@itwin/itwinui-icons-react';
+
+export default () => {
+  return (
+    <Avatar
+      abbreviation='BS'
+      backgroundColor={getUserColor('Bentley Systems')}
+      image={<SvgBentleySystems className='iui-icon' aria-hidden='true' />}
```
mayank99, line 14: - class name should be removed
- a better svg should be used
- `title` should be replaced with more useful name

FlyersPh9: Replaced with `Anonymous user` and `<SvgUser>`.

apps/website/src/examples/Avatar.picture.tsx
mayank99: > move it out to a variable

bump

mayank99 (approved):

apps/website/src/examples/Avatar.picture.tsx
mayank99: looks good

PR #975, opened 2023-01-09
Add `approve:css` and `approve:react` commands to root

mayank99:
Top level package.json now has seprate commands for `approve:css` and `approve:react`, instead of the vaguely named `approve`.


gretanausedaite (commented):
Does it make sense to have one "approve" for both?

mayank99:
> Does it make sense to have one "approve" for both?

I thought about that but didn't do it because it could cause unexpected results when tests are only run for css or react (not both). For example, lets say you run tests for ComboBox react component, then switch branches and run tests for Alert css component. The images from ComboBox are probably still in the cypress folder and would get wrongly approved.

gretanausedaite (approved):

PR #976, opened 2023-01-09
Fix ComboBox maxWidth shifting around when using virtualization

### LINKED ISSUE #955: Combobox : Improvement in dropdown menu width when enabledVirtulization is true
#### Feature
When combobox has enbled virtulization is true then dropdown menu width changes depending on the width of menu. This can be improved by setting dropdownmenu with to combobox width if virtualization is enabled.

#### Examples
You can see in below example dropdown width is changing.
combobox

mayank99:
Fixes #955


packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -61,7 +61,10 @@ const VirtualizedComboBoxMenu = React.forwardRef(
 
     const surfaceStyles = {
       minWidth,
-      maxWidth: `min(${minWidth * 2}px, 90vw)`,
+
+      // set as constant because we don't want it shifting when items are unmounted
+      maxWidth: minWidth,
```
nigtas, line 66: Have you checked how it looks like if virtualization is enabled and item which is not rendered initially is much longer than the width was set before...?

veekeys, line 66: Have you checked how it looks when virtualization is enabled and the item which is not rendered initially is much wider than the calculated width?

veekeys: oh ok.. so it just never recalculates and hardcodes the width as it is calculated from the initial visible list.

FlyersPh9 (approved):

adamhock (approved):

PR #977, opened 2023-01-09
Update pull_request_template.md

mayank99:
## Changes

Trying something new. Instead of a boring checklist which gets ignored in all PRs, I've separated the PR template into three sections (much like this very PR description) and added somewhat helpful comments to each. Got this idea from the astro repo (take a look at some of their [merged PRs](https://github.com/withastro/astro/pulls?q=is%3Apr+is%3Aclosed+) for examples). And if one of the sections does not apply to a PR, then we can write "N/A".

## Testing

N/A

## Docs

N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,14 +1,34 @@
 <!--
 Thank you for your contribution to iTwinUI.
-Please describe your PR here and make sure to complete all of the items below before submitting.
+
+If you are only making changes to the docs site using the "Edit page on GitHub" link,
+then you can ignore most of this template.
 -->
 
-Closes # <!-- Add issue number -->
+Closes #<!-- Add issue number -->
 
-## Checklist
+## Changes
 
-- [ ] Add meaningful unit tests for your component (verify that all lines are covered)
-- [ ] Verify that all existing tests pass
-- [ ] Add component features demo in Storybook (different stories)
-- [ ] Approve test images for new stories
-- [ ] Add screenshots of the key elements of the component
+<!--
+What kind of code changes does this PR include?
+Mention anything that could be helpful for reviewers and include screenshots for visual changes.
+-->
+
+## Testing
+
+<!--
+How did you test your changes?
+If your PR has visual changes, then make sure they are demonstrated in html test pages as well as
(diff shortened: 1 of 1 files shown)


FlyersPh9 (approved):

.github/pull_request_template.md
```diff
@@ -1,14 +1,32 @@
...
+
+<!--
+How did you test your changes?
+If your PR has visual changes, then make sure they are demonstrated in html test pages as well as
+storybook, then approve visual test images for both (`yarn approve:css` and `yarn approve:react`).
+
+If not applicable, you can write "N/A".
 -->
 
-Closes # <!-- Add issue number -->
```
rohan-kadkol, line 6: Should we keep this? Or maybe add a comment about linking the issue? Because maybe external contributors might miss this step.

.github/pull_request_template.md
```diff
@@ -1,14 +1,32 @@
...
+## Changes
+
+<!--
+What kind of code changes does this PR include?
+Mention anything that could be helpful for reviewers and include screenshots for visual changes.
+-->
+
+## Testing
+
+<!--
```
rohan-kadkol, line 17: Maybe we should add about verifying the existing tests pass? Because if we create a `v1` branch like before that does not run CI before the merge into the `v1` branch, failing tests may be merged.

.github/pull_request_template.md
mayank99: I kept it initially but thought about it some more and removed it later in [`c33558a`](https://github.com/iTwin/iTwinUI/pull/977/commits/c33558a3053a0f1e76bd81fd6ccb50a355f5ffec) because:
- 99% of our PRs are from our own team members who will likely not miss step.
- there is not always an issue to link (e.g. quick fixes / docs changes)
- external contributors likely have an issue that they themselves would like to see closed, so they might link it themselves.
- even if it gets missed, it is super easy to link it through github's sidebar.

.github/pull_request_template.md
mayank99: I would like to prevent this situation entirely. Tests should run in CI in all branches, and if they are not running, then we should address that instead of changing the PR template.

rohan-kadkol (approved):

PR #979, opened 2023-01-09
Update CONTRIBUTING.md

mayank99:
## Changes

Updates to contributing guide:
- added prominent note that the guide can be ignored for docs-only contributions
- made the "Adding components" sections smaller, combining the css/react sections.
  - this is because we are not that often adding new components.
- moved the "Visual testing (CSS)" down.
- updated some commands and paths.
- added note about changesets and removed note about conventional commits
- removed note about "Visual Design team" and "specifications". not relevant anymore

## Testing

N/A

## Docs

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -369,6 +369,7 @@ To enable us to quickly review and accept your pull requests, always create one 
 - Tests added for all new code.
   - All existing and new tests should pass.
 - Stories added to demonstrate new features.
+- If your changes are user-facing, then add a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) by running `yarn changeset`.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -60,6 +62,8 @@ e.g. to build all workspaces together, run the following command from the root:
 yarn build
 ```
 
+If you only need to run this task for a specific workspace, you can specify turborepo's `--filter` argument. For example, if you only want to build itwinui-react, you could run `yarn build --filter=itwinui-react`. Note that this will automatically run `build` for any dependencies (e.g. `itwinui-css` and `itwinui-variables`). You can see the pipeline in the `turbo.json` file.
```
FlyersPh9, line 65: Is there any benefit to calling out turborepo specifically vs being non-specific?

FlyersPh9 (approved):

CONTRIBUTING.md
mayank99: i'm not sure what you mean. this `--filter` syntax and the `turbo.json` are specific to turborepo.

also this line is not an addition. it was moved from below.

FlyersPh9: Ok, `--filter` being specific to `turborepo` answers my question. Thanks.

PR #980, opened 2023-01-09
Combine `createComponent` scripts

mayank99:
## Changes

Combined the two `createComponent` script into a single one. Can now be run by simply typing `yarn createComponent` at the root of the repo.

Unrelated: renamed `push-docker` command to `push-docker:css` to be more specific.

## Testing

Ran `yarn createComponent` and tested with four different component names:
- `foo`
- `Foo`
- `foo-bar`
- `FooBar`

## Docs

Already mentioned in CONTRIBUTING.md

https://github.com/iTwin/iTwinUI/blob/8099bf629eda0d3ab3ead75d732f0d07ae45f915/CONTRIBUTING.md?plain=1#L115-L119


scripts/createComponent.js
```diff
@@ -160,56 +161,160 @@ it('should be improved', () => {
...
+`;
+
+  return {
+    path: `${directory}/${componentName}.html`,
+    template: template,
+  };
+};
+
+/** Creates a .js file with a basic backstop visual test scenario */
+const scenarioJsFactory = (directory, componentName) => {
```
rohan-kadkol, line 253: Unused variable

```suggestion
const scenarioJsFactory = (directory) => {
```

mayank99: it's used in line 265

rohan-kadkol (approved):

PR #984, opened 2023-01-11
feat(FileUpload): Standard GUI for single file upload

### LINKED ISSUE #839: File upload after a file has been uploaded
The [file upload component](https://itwin.github.io/iTwinUI/file-upload.html) does not offer a standard GUI after the file is uploaded.  Mockups from the internal discussion & (old) visual standards are below to be used as inspiration, not as strict requirements.

**Single file upload:**
Screenshot 2022-12-08 at 2 07 41 PM

**Multiple file upload:**

https://user-images.githubusercontent.com/849817/206545747-d89a5c4c-ecf3-4d13-ba21-0011907699bf.mp4

**Image upload:**
Screenshot 2022-12-08 at 2 08 26 PM

adamhock:
## Changes

Added a default GUI for when a single file is uploaded using `FileUploadCard`.
Closes #839

## Testing

Uploading a single file in the "Single File Upload" story resulted in the following changes

Before: 
image

After: 
image

## Docs

N/A

## TODO
- [ ] Allow users to change the file they've uploaded using `FileUploadCard`
- [ ] Allow users to customize their `FileUploadCard` using its subcomponents
- [ ] Unit testing


packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -51,11 +68,52 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
...
+    files,
+    fileLabel,
+    fileDescription,
+    svgIcon,
     children,
   } = props;
   useTheme();
 
-  return (
+  const icon = svgIcon ? (
```
gretanausedaite, line 79: I think these icons could be mapped to file type (pdf for pdf, ppt to ppt icon, etc).

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -32,6 +33,22 @@ export type FileUploadTemplateProps = {
    * @default 'or drag & drop it here.'
    */
   subtitle?: React.ReactNode;
+  /**
+   * Files selected. Used for default uploaded file output only when one file is selected.
+   */
+  files?: Array<File>;
```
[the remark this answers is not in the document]
gretanausedaite: Seems like all these new props are targeting single file upload. What happens in a future PRs when we'd add same styles for multiple file upload? Add more props or change these ones (breaking change)?

gretanausedaite (commented):
To me it seems like not finished design. It could have slight background change (eg. surface) to distinguish it from the app background especially because it is built-in template that users can't change. @FlyersPh9 this is a question / design review for you.

Is there a reason why you are only targeting single file upload? It seems like you just run the same code snippet through a loop to create more file uploaded cards.

Should have a way for user to override single file uploaded styles / card.

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
mayank99: it was my suggestion to not handle file extensions because we cannot handle every extension and it would bloat up the component. so it should be left upto the user.

but we should show examples of how the user would achieve it.

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -42,6 +42,34 @@
       margin-bottom: var(--iui-size-s);
     }
 
+    > svg[data-iui-file-uploaded] {
+      @include iui-icon-style(xl);
+      margin-left: var(--iui-size-xs);
+    }
+
+    .iui-template-file-uploaded-text {
```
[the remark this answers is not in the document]
FlyersPh9: I agree.  File upload could benefit from a class name refactor (ex: `.iui-content` becoming `.iui-file-upload-content`), but anything new that we introduce could follow this new name format.

FlyersPh9:
> To me it seems like not finished design. It could have slight background change (eg. surface) to distinguish it from the app background especially because it is built-in template that users can't change. @FlyersPh9 this is a question / design review for you.

Depending on where this is placed in the UI, I could see a bordered & non-bordered state being useful (similar to how tag containers can include the `.iui-visible` modifier or not).  I am confused in the CSS demo page though, because I see one with and one without a border.  When would that bottom example display?
Screenshot 2023-01-11 at 11 21 59 PM

> Is there a reason why you are only targeting single file upload? It seems like you just run the same code snippet through a loop to create more file uploaded cards.

In the issue [I suggested we only target single file upload](https://github.com/iTwin/iTwinUI/issues/839#issuecomment-1346613108), but we should build this in a way to handle multi-file upload eventually.

FlyersPh9:
The file name color doesn't seem as bright as the other blue text but after inspecting I'm not seeing a color difference or opacity applied, so maybe it's an optical illusion.
Screenshot 2023-01-11 at 11 25 44 PM

But the blue file name does make me think I can click it to do something.  If it's not actionable we should probably just use our regular text color.

FlyersPh9:
The gap between file name & description visually looks a little too large.
Screenshot 2023-01-11 at 11 29 48 PM

Can we match the gap used in header?
Screenshot 2023-01-11 at 11 31 02 PM

packages/itwinui-css/backstop/tests/file-upload.html
```diff
@@ -659,5 +659,62 @@ <h2>Not wrapping anything</h2>
...
+      <div class="iui-file-upload">
+        <div class="iui-content">
+          <svg-filetype-pdf
+            class="iui-template-file-uploaded-icon"
+            aria-hidden="true"
+            data-iui-file-uploaded="true"
+          ></svg-filetype-pdf>
+          <span class="iui-template-file-uploaded-text">
+            <span class="iui-template-file-uploaded-label">File title</span>
+            <span class="iui-template-file-uploaded-description">File description</span>
```
FlyersPh9, line 676:
```suggestion
            <span class="iui-template-file-uploaded-description">15KB Wed Jan 11 2023 5:25:59 PM</span>
```
Can we put something more realistic like the file size and date from the Storybook example?

FlyersPh9:
If this file upload was in a long form with inputs above & below it, I think as it is right now after a file is uploaded all content beneath it would shift up because the height is changing.  I'm not sure what we should do, but we should try to prevent content shift when a single file is uploaded.

elephantcatdog:
I don't have any additional comments atm

packages/itwinui-css/backstop/tests/file-upload.html
FlyersPh9: An example I found on another site:
Screenshot 2023-01-12 at 9 35 09 AM

I suppose we should support displaying whatever info the dev user wants:  file size, a message like the example above, something else?

FlyersPh9:
Testing GitHub bot in Microsoft Teams, please ignore this.

FlyersPh9:
Testing once more if GitHub bot in MS Teams is working, please ignore.

FlyersPh9:
Testing, please ignore. 2

packages/itwinui-css/backstop/tests/file-upload.html
adamhock: I have it so that users are able to pass whatever message they want when the file is uploaded. The default is just showing the file name, file size, and the last modified date

The changes to the demo page have been made

adamhock:
> The gap between file name & description visually looks a little too large. Screenshot 2023-01-11 at 11 29 48 PM
> 
> Can we match the gap used in header? Screenshot 2023-01-11 at 11 31 02 PM

These changes have been made

adamhock:
> > To me it seems like not finished design. It could have slight background change (eg. surface) to distinguish it from the app background especially because it is built-in template that users can't change. @FlyersPh9 this is a question / design review for you.
> 
> Depending on where this is placed in the UI, I could see a bordered & non-bordered state being useful (similar to how tag containers can include the `.iui-visible` modifier or not). I am confused in the CSS demo page though, because I see one with and one without a border. When would that bottom example display? Screenshot 2023-01-11 at 11 21 59 PM
> 
> > Is there a reason why you are only targeting single file upload? It seems like you just run the same code snippet through a loop to create more file uploaded cards.
> 
> In the issue [I suggested we only target single file upload](https://github.com/iTwin/iTwinUI/issues/839#issuecomment-1346613108), but we should build this in a way to handle multi-file upload eventually.

This example is what is shown when a file is being dragged into `FileUpload`
FileUploadDragBefore

FileUploadDragAfter

> The file name color doesn't seem as bright as the other blue text but after inspecting I'm not seeing a color difference or opacity applied, so maybe it's an optical illusion. Screenshot 2023-01-11 at 11 25 44 PM
> 
> But the blue file name does make me think I can click it to do something. If it's not actionable we should probably just use our regular text color.

Would you still want the file name to use the regular text color in this case?

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -77,3 +135,19 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
 };
 
 export default FileUploadTemplate;
+
+function toBytes(bytes: number) {
```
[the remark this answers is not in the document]
adamhock: This function is only used in `FileUploadTemplate`, so I moved it to the top of the file

packages/itwinui-css/src/file-upload/file-upload.scss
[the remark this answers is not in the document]
adamhock: The new class names have been moved to separate mixins

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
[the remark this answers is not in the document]
adamhock: I have changed it so that users can only pass in a file into `FileUploadTemplate` if `accessMultiple` is `false`. Once we have a default GUI for multiple files, we can change `files` (which is now called `data` with my recent changes) to be an array of files only when `accessMultiple` is `true`. This should allow us to implement a standard GUI for multiple file uploads down the road without causing any breaking changes

elephantcatdog (commented):
image
The text "Drag files to attach" has a pointer cursor even though there's nothing to click.


The two screenshots for the react visual tests look the same to me. Is one supposed to be without a file having been uploaded?


Also I really like how the "wrapping input" example works; it's cool.

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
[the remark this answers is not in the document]
gretanausedaite: Maybe it makes sense to move 
```
data?: File;
dataLabel?: string;
dataDescription?: string;
dataSvg?: JSX.Element;
```
to new type. Then users could provide single file with metadata or array of files with their metadata. 

Right now, even if you change it to accept Files[], how would user pass label, description and svg for each file?
Although this PR is targeting single file upload only, you should think how it would work with multiple files to not have breaking changes in the future.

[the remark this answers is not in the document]
adamhock: > The two screenshots for the react visual tests look the same to me. Is one supposed to be without a file having been uploaded?
> 

Yes! Good catch. This has been fixed

adamhock:
> The two screenshots for the react visual tests look the same to me. Is one supposed to be without a file having been uploaded?
>

Yes! Good catch. This has been fixed

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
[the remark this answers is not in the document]
adamhock: These props have been moved to its own type. Now when we support multiple file uploads, users can pass an array of `DataProps` which feature their files along with the option to put whatever label, description, and svg they choose for each file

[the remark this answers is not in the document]
gretanausedaite: What do you think about renaming it to `FileDataProps`? A bit more specific than just DataProps.

Also, now that it's moved, inner props name, eg. "dataLabel", seems a bit redundant. It can be just `label` as it is "data label" from design (child of DataProps).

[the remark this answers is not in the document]
adamhock: These names have been changed!

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -51,11 +68,52 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
...
+    <SvgDocument
+      className='iui-icon'
+      aria-hidden
+      data-iui-file-uploaded={true}
+    />
+  );
+
+  return files?.length === 1 ? (
+    <>
+      {icon}
```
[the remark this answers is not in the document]
adamhock: This has been exported into a separate component. I also have the option for users to use the `FileUploaded` component without needing the `FileUpload` or `FileUploadTemplate` components, but a few CSS props are not applied to `FileUploaded` when its not nested within `FileUpload` since they are applied to `iui-content`

```SCSS
    padding: calc(var(--iui-size-s) * 0.5) var(--iui-size-s);
    border: 1px solid var(--iui-color-background-backdrop);
```

Also the additional props for `iui-drag` would not be applied to `FileUploaded` when used as a standalone component

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -74,3 +74,58 @@
...
 }
+
+@mixin iui-file-uploaded-template {
+  width: 100%;
+  height: 100%;
+  line-height: normal;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  background-color: var(--iui-color-background-backdrop);
```
gretanausedaite, line 85: Why did you add background color?
image

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -76,4 +155,36 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
...
+   */
+  label: string;
+  /**
+   * File description for uploaded file output.
+   */
+  description: string;
+  /**
+   * Action element for uploaded file.
+   */
+  action: JSX.Element;
```
gretanausedaite, line 174: I think this prop could be optional. Maybe users don't want any action on uploaded file.

gretanausedaite (commented):
Generally looks good.
- Add changeset.
- Test in one of our test apps to see if typings are ok. 

I think this mandatory one file uploaded style could be a breaking change if users are not expecting anything to change. Maybe if user passes children, we should show the old way to not break apps visually? Or they can simply not pass "data". @mayank99

mayank99:
> I think this mandatory one file uploaded style could be a breaking change if users are not expecting anything to change. Maybe if user passes children, we should show the old way to not break apps visually? Or they can simply not pass "data".

I would have expected this to be an entirely new component that users can choose to use instead of `FileUploadTemplate`. When `FileUploadTemplate` was originally created, it was intentionally made to be a very generic piece that can be swapped out easily. So we can handle both the "before" and "after" states in the new component. This will prevent breaking changes for existing users, and also set the precedent for adding other templates in the future (e.g. multiple file upload perhaps).

Also, looking at the API I think most of those props (except maybe `icon`) should actually be handled through `children` via subcomponents. The current approach doesn't give the user much control over the DOM.

Lastly, I noticed that the height of the "before" and "after" states is different, so it will cause content shift after uploading the file. Maybe the height of the "before" state should be reduced slightly.

packages/itwinui-css/backstop/tests/file-upload.html
```diff
@@ -659,5 +659,67 @@ <h2>Not wrapping anything</h2>
...
+                />
+                Replace
+            </div>
+          </div>
+        </div>
+      </div>
+
+      <br />
+
+      <div class="iui-file-upload iui-drag">
```
elephantcatdog, line 695: This example doesn't look quite right to me/match the others - the blue background doesn't cover the entire area. Is this what it's supposed to look like?
image

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.test.tsx
```diff
@@ -56,3 +56,34 @@ it('should accept input props', () => {
...
+  const label = container.querySelector(
+    '.iui-file-uploaded-template-label',
+  ) as HTMLElement;
+  expect(label).toBeTruthy();
+  expect(label.textContent).toEqual('test.txt');
+
+  const secondary = container.querySelector(
+    '.iui-file-uploaded-template-description',
+  ) as HTMLElement;
+  expect(secondary).toBeTruthy();
```
elephantcatdog, line 81: If you want to test that the text of the description is being formatted correctly, you could mock `toBytes` and `toDate`.

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -3,20 +3,68 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
-import { useTheme, SvgUpload } from '../utils';
+import cx from 'classnames';
+import { useTheme, SvgUpload, SvgDocument } from '../utils';
 import '@itwin/itwinui-css/css/file-upload.css';
 
+const toBytes = (bytes: number) => {
```
elephantcatdog, line 10: This is a neat little function.

packages/itwinui-css/backstop/tests/file-upload.html
adamhock: This was due to a change in the background color that I've made and since reverted. This should be fixed now

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.test.tsx
```diff
@@ -56,3 +56,34 @@ it('should accept input props', () => {
...
+
+  const { container } = render(
+    <FileUploadTemplate
+      onChange={mockedOnChange}
+      acceptMultiple={false}
+      data={{ file: file }}
+    />,
+  );
+
+  const label = container.querySelector(
```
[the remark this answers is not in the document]
adamhock: Added

adamhock:
> > I think this mandatory one file uploaded style could be a breaking change if users are not expecting anything to change. Maybe if user passes children, we should show the old way to not break apps visually? Or they can simply not pass "data".
> 
> I would have expected this to be an entirely new component that users can choose to use instead of `FileUploadTemplate`. When `FileUploadTemplate` was originally created, it was intentionally made to be a very generic piece that can be swapped out easily. So we can handle both the "before" and "after" states in the new component. This will prevent breaking changes for existing users, and also set the precedent for adding other templates in the future (e.g. multiple file upload perhaps).
> 
> Also, looking at the API I think most of those props (except maybe `icon`) should actually be handled through `children` via subcomponents. The current approach doesn't give the user much control over the DOM.
> 
> Lastly, I noticed that the height of the "before" and "after" states is different, so it will cause content shift after uploading the file. Maybe the height of the "before" state should be reduced slightly.

I like this approach of having a new component for this functionality. I will be changing the PR to draft until I have a solution using this approach that is ready for review

adamhock:
> Lastly, I noticed that the height of the "before" and "after" states is different, so it will cause content shift after uploading the file. Maybe the height of the "before" state should be reduced slightly.

I've reduced the size of the upload icon and reset the line height, but it leads to some pretty drastic changes across the board for file upload (see CSS images). Let me know your thoughts on this

packages/itwinui-react/src/core/FileUpload/FileUploadCard.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+          Clear
+        </Button>
+        <FileUploadCard.Input ref={inputRef} />
+      </FileUploadCard.Action>
+    </FileUploadCard>
+  );
+};
+
+it('should render empty FileUploadCard before a file is uploaded', () => {
+  const { container } = render(<FileUploadCard />);
```
adamhock, line 47: All of the unit tests are failing because of this error in `FileUploadCardInput` that I'm not quite sure how to fix yet

image

.changeset/wild-goats-march.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': minor
+---
+
+Added a new FileUploadCard component which serves as a default GUI for when a single file is uploaded using FileUpload
```
gretanausedaite, line 6: Update description with usage example.

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { FileUploadCardLabel } from './FileUploadCardLabel';
+import { FileUploadCardDescription } from './FileUploadCardDescription';
+import { FileUploadCardInput } from './FileUploadCardInput';
+import { FileUploadCardIcon } from './FileUploadCardIcon';
+
+export type FileUploadCardProps = {
+  /**
+   * List of files to pass (only needed when passing a custom action)
+   */
+  files?: File[];
```
gretanausedaite, line 19: Maybe it should be `File` (singular) as this component is for one file.

gretanausedaite (commented):
Drag&drop does not work with new components.

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
elephantcatdog: In the future, would we add another component or parameter for multiple files? If we add onto this component for multiple file support later on, `files` would be better long term.

gretanausedaite: For multiple files we could do something like this:
```js
<FileUploadCardContainer>
  <Card1 />
  <Card2 />
</FileUploadCardContainer>
```
If we want to keep same component for both single and multiple files, we could add a flag "multiple" (or something like that) and add to prop type `files?: File | File[]` based on that flag. 

In both cases we should fix CardLabel and CardDescription components as they are using file[0] to get data.

elephantcatdog (commented):
The image for the visual test "FileUpload.test.ts-Single File Upload Custom.png" looks the same as the one for the non-custom visual test.

packages/itwinui-react/src/core/FileUpload/FileUploadCardAction.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import cx from 'classnames';
+
+export type FileUploadCardActionProps = {
+  /**
+   * Buttons in the dialog bar.
+   */
+  children: React.ReactNode;
```
mayank99, line 8: many of these comments are incorrectly copied, but also, you don't need to explicitly add `children` (in any of the files, not just this one). It will automatically come from `React.ComponentProps`

mayank99: also it might be easier to navigate if you move all subcomponents in the same file, similar to [`Flex`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/utils/components/Flex.tsx)

packages/itwinui-react/src/core/FileUpload/FileUploadCardAction.tsx
```diff
@@ -0,0 +1,21 @@
...
+} & React.ComponentPropsWithRef<'div'>;
+
+export const FileUploadCardAction = React.forwardRef<
+  HTMLDivElement,
+  FileUploadCardActionProps
+>((props, ref) => {
+  const { children, className, ...rest } = props;
+  return (
+    <div className={cx('iui-file-card-action', className)} ref={ref} {...rest}>
+      <label className='iui-anchor'>{children}</label>
```
mayank99, line 18: user needs to have control over the `<label>` element too, so it should probably be another subcomponent

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    <FileUploadCard.Input />
+                    {'Replace'}
+                  </FileUploadCard.Action>
+                </>
+              )}
+            </div>
+          ) : (
+            <>
+              <SvgUpload className='iui-icon' aria-hidden />
+              <div className='iui-template-text'>
```
mayank99, line 86: this initial UI should also be customizable and probably have subcomponents.

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -74,3 +75,56 @@
...
+
+  .iui-browse-input {
+    width: 0.1px;
+    height: 0.1px;
+    opacity: 0;
+    overflow: hidden;
+    position: absolute;
+  }
+
+  .iui-file-card-action {
```
mayank99, line 104: most of these classes should be separate mixins instead of nesting. and then the classes can be defined in `classes.scss`

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -74,3 +75,56 @@
...
+  .iui-file-card-icon {
+    @include iui-icon-style(xl);
+    @include iui-icon-color;
+    margin-top: var(--iui-size-s);
+    margin-bottom: var(--iui-size-s);
+    margin-left: var(--iui-size-xs);
+  }
+
+  .iui-browse-input {
+    width: 0.1px;
```
mayank99, line 97: `iui-browse-input` looks odd. this is a new component, so a new class should be used

it would be better to use visually-hidden styles instead of this code.

playgrounds/vite/src/App.tsx
```diff
@@ -1,11 +1,53 @@
...
+          <FileUploadCard.Action>
+            <Button onClick={(e) => setFiles([])}>Clear</Button>
+          </FileUploadCard.Action>
+        </FileUploadCard>
+      </FileUpload>
     </>
   );
 };
 
+// const WithDiscrd = () => {
```
mayank99, line 37: all changes in App.tsx should be discard instead of being pushed

apps/storybook/src/FileUpload.stories.tsx
```diff
@@ -74,3 +77,54 @@ export const WrappingInput: Story<FileUploadProps> = (args) => {
...
+  );
+};
+
+export const SingleFileUploadCustom: Story<FileUploadProps> = (args) => {
+  const inputRef = React.useRef<HTMLInputElement>(null);
+  const [files, setFiles] = React.useState<File[]>([]);
+  const fileInputId = React.useId();
+
+  return (
+    <FileUpload
```
mayank99, line 100: Have you tested whether it works without `FileUpload`? we need to make it work

also, the `files`/`onFilesChange` props should be added to `FileUpload`. usually in compound components, the component at the very top should control the state

packages/itwinui-react/src/core/FileUpload/FileUploadCardAction.tsx
adamhock: This new subcomponent has been added

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: These classes have been moved to separate mixins

playgrounds/vite/src/App.tsx
adamhock: This has been removed

adamhock:
> The image for the visual test "FileUpload.test.ts-Single File Upload Custom.png" looks the same as the one for the non-custom visual test.

This screenshot has been removed out of redundancy

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: This has been changed to a new component with the visually-hidden styling

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
adamhock: CardLabel and CardDescription have been changed

`files` has been changed to `data` so it works in the singular and plural sense which leaves room for a `multiple` flag that changes the prop type

packages/itwinui-react/src/core/FileUpload/FileUploadCardAction.tsx
adamhock: The comments and the explicitly adding of `children` have been removed. All of the sub-components have been moved to the same file

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
adamhock: An FileEmptyCard has been created

elephantcatdog:
> This screenshot has been removed out of redundancy

Instead of removing it, you could change the custom version so that it looks different.

gretanausedaite (commented):
No additional comments.

Still think it's not a good idea to use `data[0]` to get file data in file card. 
1x`FileUploadCard` = 1x`file`

apps/storybook/src/FileUpload.stories.tsx
adamhock: Outside of the dragging functionality, this component works without `FileUpload`. I changed the CSS for `iui-file-card` and the newly-created `iui-file-card-empty` so that it matches the previous screenshots when outside of `iui-content`. However, I have some CSS issues. 

When `iui-file-card` is wrapped by `iui-content`, extra padding from `iui-content` is added to the padding from `iui-file-card`. This also leads to issues with `focus-within` when comparing the two structures. 

image

I've added the two types of HTML structures for `FileUploadCard` (One standalone and one wrapped by the `FileUpload` component) in the demo pages until I can resolve this issue.
	
As for the other issue, how would I pass the `files` and `onFilesChanged` props from `FileUpload` to `FileUploadCard` if the `children` of `FileUpload` isn't always a `FileUploadCard` component?

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+ *     </FileEmptyCard.Description>
+ *   </FileEmptyCard.Text>
+ * </FileEmptyCard>
+ */
+export const FileEmptyCard = Object.assign(
+  (props: FileEmptyCardProps) => {
+    const { children } = props;
+
+    return (
+      <>
```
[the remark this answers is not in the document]
adamhock: `FileEmptyCard` now has a root element.

apps/storybook/src/FileUpload.stories.tsx
```diff
@@ -74,3 +82,73 @@ export const WrappingInput: Story<FileUploadProps> = (args) => {
...
+    </FileEmptyCard>
+  );
+  return (
+    <FileUpload
+      {...args}
+      onFileDropped={(files) => {
+        action(`${files.length} files uploaded`)();
+      }}
+    >
+      <FileUploadCard
```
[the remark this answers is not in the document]
adamhock: `FileUploadCard.Anchor -> FileUploadCard.InputLabel`
	
`FileUploadCard.Label -> FileUploadCard.Title`

`FileUploadCard.Text -> FileUploadCard.Info` to show that information about the file should be shown here. Not necessarily the best name since users can add custom text to `Title` and `Description` that has nothing to do with the file.
	
I'm having trouble understanding why `FileEmptyCard` should be a subcomponent of `FileUploadCard` if they are two different states?

packages/itwinui-react/src/core/FileUpload/FileUploadCard.test.tsx
adamhock: The issue is that DataTransfer is not in jest, so it must be mocked. I've changed things so that the tests pass, but I still need to properly mock DataTransfer in these unit tests to make sure it's working

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+export type FileEmptyCardAnchorProps = {
+  /**
+   * Anchor label
+   * @default 'Choose a file'
+   */
+  label?: string;
+} & React.ComponentPropsWithRef<'label'>;
+
+const FileEmptyCardAnchor = React.forwardRef<
```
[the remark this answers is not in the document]
adamhock: This name has been changed, and both `FileEmptyCard` and `FileUploadCard` now use the same input label

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const { children } = props;
+
+    return (
+      <>
+        {children ?? (
+          <>
+            <FileEmptyCard.Icon />
+            <FileEmptyCard.Text>
+              <FileEmptyCard.Anchor>
+                <FileUploadCard.Input />
```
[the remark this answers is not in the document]
adamhock: This has been changed

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
```diff
@@ -0,0 +1,323 @@
...
+        }}
+        ref={refs}
+        id={inputId}
+        {...rest}
+      />
+      {children}
+    </>
+  );
+});
+
```
mayank99, line 197: add separator here

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -69,8 +70,120 @@
...
+}
+
+@mixin iui-file-card-action {
+  margin-left: auto;
+  margin-right: var(--iui-size-xs);
+  font-weight: var(--iui-font-weight-bold);
+}
+
+@mixin iui-file-card-input {
+  @include visually-hidden;
```
mayank99, line 143: you can use the `iui-visually-hidden` class directly. no need for creating a new class/mixin

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -69,8 +70,120 @@
...
+  align-items: center;
+  padding: calc(var(--iui-size-s) * 0.5) var(--iui-size-s);
+  border-radius: var(--iui-border-radius-1);
+
+  &:focus-within {
+    outline: 2px solid var(--iui-color-text-accent);
+    outline-offset: -2px;
+  }
+
+  > svg {
```
mayank99, line 96: is this needed? `iui-file-card-icon` should cover this case.

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
```diff
@@ -0,0 +1,323 @@
...
+ *   </FileUploadCard.Action>
+ * </FileUploadCard>
+ */
+export const FileUploadCard = Object.assign(
+  React.forwardRef<HTMLDivElement, FileUploadCardProps>(
+    (props, ref: React.RefObject<HTMLDivElement>) => {
+      const {
+        className,
+        children,
+        data: dataProps,
```
mayank99, line 249: nit: should say `dataProp`

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
```diff
@@ -0,0 +1,323 @@
...
+        onDataChange,
+        emptyCard = <FileEmptyCard />,
+        input,
+        ...rest
+      } = props;
+
+      const [internalData, setInternalData] = React.useState<File[]>();
+      const data = dataProps ?? internalData ?? [];
+      const inputId = React.useId();
```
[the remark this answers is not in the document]
mayank99: also i think you will need to check for user's own `id` and prefer that if passed.

e.g. the user can do this:
```tsx
<FileUploadCard input={<input id='blah'>} />
```

apps/storybook/src/FileUpload.stories.tsx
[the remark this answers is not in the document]
mayank99: > I'm having trouble understanding why FileEmptyCard should be a subcomponent of FileUploadCard if they are two different states?

because `FileEmptyCard` is used inside `FileUploadCard`. they are not alternatives to each other. `FileUploadCard` is the "main" component which holds all the logic. but it's upto you, i can see both options working fine.

---

also this one is still unaddressed:

> FileUploadCard feels wrong because cards are usually elevated and have a background color. Maybe we should add those things?

if not elevation, then a border would also do.

adamhock:
> Drag&drop does not work with new components.

This has been fixed. Drag & Drop only works when `FileUploadCard` is wrapped around `FileUpload` since that is where the dragging logic happens

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
adamhock: This has been changed

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: This has been removed

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: Good catch! This has been changed

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
adamhock: Added

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+  FileEmptyCardDescriptionProps
+>((props, ref) => {
+  const { children, className, ...rest } = props;
+  return (
+    <div className={className} ref={ref} {...rest}>
+      {children}
+    </div>
+  );
+});
+
```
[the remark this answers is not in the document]
adamhock: Added

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+    },
+  ),
+  {
+    Icon: FileEmptyCardIcon,
+    Text: FileEmptyCardText,
+    Description: FileEmptyCardDescription,
+  },
+);
+
+export const FileEmptyCardContext = React.createContext<
```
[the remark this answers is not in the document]
adamhock: This context has been removed

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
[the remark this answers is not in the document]
adamhock: > `useId` doesn't exist before react 18 and we aim to be compatible all the way back to react 16, so you will need to add a fallback.
> 
> you can create a utility like this:
> 
> ```js
> export const useId = React.useId ?? () => {
>   const [id] = React.useState(() => `iui-${getRandomValue(10)}`);
>   return id;
> };
> ```

This has been added. I am still working on the second part of this comment

packages/itwinui-css/src/file-upload/file-upload.scss
mayank99: that didn't really change anything. i'm saying you don't need to make a `iui-file-card-input` because you can use `iui-visually-hidden` directly in the markup

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+export type FileEmptyCardDescriptionProps = React.ComponentPropsWithRef<'div'>;
+
+const FileEmptyCardDescription = React.forwardRef<
+  HTMLDivElement,
+  FileEmptyCardDescriptionProps
+>((props, ref) => {
+  const { children, className, ...rest } = props;
+  return (
+    <div className={className} ref={ref} {...rest}>
```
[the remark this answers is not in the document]
mayank99: but users can pass their own `<div>` and have full control over it

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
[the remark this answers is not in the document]
mayank99: it should go in the hooks folder

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: Even better. This has been changed

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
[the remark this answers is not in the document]
adamhock: This has been moved

packages/itwinui-react/src/core/FileUpload/FileEmptyCard.tsx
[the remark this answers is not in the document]
adamhock: That makes sense. This has been changed

apps/storybook/src/FileUpload.stories.tsx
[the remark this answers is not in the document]
adamhock: Elevation and background color have been added

packages/itwinui-react/src/core/FileUpload/FileUploadCard.test.tsx
adamhock: I worked on this with Mayank, and we came to the conclusion that it's not worth mocking `DataTransfer` in jest just to get tests to pass, so I have written a check for `DataTransfer` in `FileUploadCard` so the tests can still pass

.changeset/wild-goats-march.md
adamhock: Updated. I wasn't sure how in-depth to get with the usage examples, but I wrote a couple

adamhock:
> The image for the visual test "FileUpload.test.ts-Single File Upload Custom.png" looks the same as the one for the non-custom visual test.

This has been added

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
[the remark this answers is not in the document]
adamhock: > also i think you will need to check for user's own `id` and prefer that if passed.
> 
> e.g. the user can do this:
> 
> ```tsx
> <FileUploadCard input={<input id='blah'>} />
> ```

This has been changed, but I would like a second opinion on the implementation. I set the `inputId` using `useId()` first and then check for an `id` prop in `input` since `useId()` cannot be called conditionally.

apps/storybook/src/FileUpload.stories.tsx
adamhock: I was able to get the sizing of the two cards to match by adding a `margin` to `iui-file-card` and removing it if it had a parent of `iui-content`. I have also fixed the double-outline issue, but since the outline goes around the `margin` now, the `focus-within` still just isn't quite right :/

image

I've updated the react screenshots and I will revert the html demo page and upload the css screenshots once this issue is resolved

packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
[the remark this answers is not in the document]
mayank99: i don't think you can read `input.props` like that. you need to read it from inside props in `FileUploadCard.Input` and then set it in context.

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -62,15 +64,150 @@
...
+      .iui-template-icon {
+        fill: var(--iui-color-icon-accent);
+      }
+    }
+  }
+}
+
+@mixin iui-file-card {
+  @include iui-reset;
+  @include iui-surface;
```
mayank99, line 93: Continuing conversation from above:

> > > FileUploadCard feels wrong because cards are usually elevated and have a background color. Maybe we should add those things?
> > 
> > if not elevation, then a border would also do.
> 
> Elevation and background color have been added

Shouldn't empty state also have some kind of border? I'll defer to @FlyersPh9 for final decision on what the UI should look like.

packages/itwinui-css/src/file-upload/file-upload.scss
```diff
@@ -62,15 +64,150 @@
...
+  line-height: normal;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: calc(var(--iui-size-s) * 0.5) var(--iui-size-s);
+  border-radius: var(--iui-border-radius-1);
+  margin: calc(var(--iui-size-s) * 0.5) var(--iui-size-s);
```
mayank99, line 100: Continuing this conversation from [above](https://github.com/iTwin/iTwinUI/pull/984#discussion_r1125097228):

> > Have you tested whether it works without `FileUpload`? we need to make it work
> > 
> > also, the `files`/`onFilesChange` props should be added to `FileUpload`. usually in compound components, the component at the very top should control the state
> 
> I was able to get the sizing of the two cards to match by adding a `margin` to `iui-file-card` and removing it if it had a parent of `iui-content`. I have also fixed the double-outline issue, but since the outline goes around the `margin` now, the `focus-within` still just isn't quite right :/
> 
> image

Why are you setting both `margin` and `padding`? For the outer nodes, you should only set the `padding`. The icon also doesn't need extra margins, everything should be handled in the parent. Doing this will make the behavior more predictable and avoid such issues with focus outlines.

mayank99: Paired with Adam and removed all margins.

packages/itwinui-css/src/file-upload/file-upload.scss
mayank99: Paired with Adam and added border.

.changeset/wild-goats-march.md
```diff
@@ -0,0 +1,20 @@
...
+<FileUploadCard />
+```
+
+```jsx
+const [files, setFiles] = React.useState<File[]>([]);
+
+<FileUpload
+  onFileDropped={console.log()}
+>
+  <FileUploadCard data={files} onDataChange={(files) => setFiles(files)} />
```
mayank99, line 18: update this example to use correct prop names

.changeset/wild-goats-march.md
```diff
@@ -0,0 +1,20 @@
...
+
+```jsx
+<FileUploadCard />
+```
+
+```jsx
+const [files, setFiles] = React.useState<File[]>([]);
+
+<FileUpload
+  onFileDropped={console.log()}
```
mayank99, line 16: would be better to show it updating state rather than console.log

packages/itwinui-css/src/file-upload/file-upload.scss
mayank99: @FlyersPh9 Can you confirm the visuals before this is merged? I'm looking at the linked issue and the example screenshot there has no elevation - it might be used as part of a bigger card.

.changeset/wild-goats-march.md
adamhock: This has been changed

.changeset/wild-goats-march.md
adamhock: This has been changed

packages/itwinui-css/src/file-upload/file-upload.scss
mayank99: also in the linked issue, the whole background/border is blue after file is uploaded, whereas in this PR, it's regular surface colors. not sure which one we want.

FlyersPh9: @adamhock can you remove the drop shadow & leave everything else as is?

FlyersPh9 (approved):

packages/itwinui-css/src/file-upload/file-upload.scss
adamhock: The drop shadow has been removed

PR #995, opened 2023-01-12
website: Updated footer year

gretanausedaite:
## Changes

Updated Footer year to 2023.
image

## Testing

Open home page and scroll to the bottom.

## Docs

N/A


apps/website/src/components/utils/Footer.tsx
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+export default (props: React.ComponentProps<'footer'>) => {
+  const currentYear = new Date().getFullYear();
+  return <footer {...props}>© {currentYear} Bentley Systems, Inc</footer>;
```
mayank99, line 8: does it need to be a react component?

gretanausedaite: Do you want it in Astro or inline?

I did react because it's easy to include variables in html.

mayank99: well, we want components to be astro by default (just as easy to include variables in html), and only use react when astro doesn't work

gretanausedaite: Changed to Astro.

mayank99 (approved):

apps/website/src/components/utils/Footer.astro
```diff
@@ -0,0 +1,14 @@
+---
+const props = Astro.props;
+const currentYear = new Date().getFullYear();
+---
+
+<footer {...props}>© {currentYear} Bentley Systems, Inc</footer>;
```
mayank99, line 6:
```suggestion
<footer {...props}>© {currentYear} Bentley Systems, Inc</footer>
```

adamhock (approved):

PR #997, opened 2023-01-12
website: Update dialog documentation & demos

FlyersPh9:
- Add demos.
- Refine text.
- Create links to other component pages.


rohan-kadkol (commented):
Should we include the similarity between `Modal` and `Dialog`? Because now it looks there is only documentation for `Dialog` but not `Modal`. Also, the props in the props table is from `Modal`, but there is no other mention of `Modal`.

apps/website/src/examples/Dialog.fullpage.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <>
+      <Button styleType='high-visibility' onClick={() => setIsModalOpen(true)}>
+        Open full page dialog
+      </Button>
+      <Modal
+        isOpen={isModalOpen}
+        title={'New message'}
+        styleType='fullPage'
+        onClose={() => closeModal()}
```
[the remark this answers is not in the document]
FlyersPh9: Oh good, I'm glad you caught this and suggested a fix!  Thanks, added this to all of the demos that start with the dialog opened.

apps/website/src/pages/docs/dialog.mdx
```diff
@@ -80,28 +79,17 @@ Don't:
 - Present dialogs unless prompted by the user.
 - Use a login within a dialog windows. You can’t link to them and some password managers can’t pre-fill them because the dialogs are hidden.
 
-### Alternative ways to display the information
-
 Here are other ways to display information if a dialog is deemed inadequate for the context, but an important message still needs to be passed onto the user.
```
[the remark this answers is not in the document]
FlyersPh9: Removed.

apps/website/src/pages/docs/dialog.mdx
```diff
@@ -13,58 +17,53 @@ import LiveExample from '~/components/LiveExample.astro';
...
+
 The two buttons at the bottom right are always pinned in that same area, regardless of content. If there is more content than there is available space in the content area, a scrollbar can be included to make the content area scrollable.
 
 As specified earlier, full-page dialogs will house various settings options as shown in the example below.
 
 ## Usage guidelines
 
-### Dos & Don'ts
-
 Do:
```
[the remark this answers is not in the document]
FlyersPh9: I left Do & Don't as is so they aren't added to the table of contents.  That makes sense to me, but if others feel otherwise I have no issue with changing.

FlyersPh9:
> Should we include the similarity between Modal and Dialog? Because now it looks there is only documentation for Dialog but not Modal. Also, the props in the props table is from Modal, but there is no other mention of Modal.

@rohan-kadkol, I don't know what we should do about Modal vs Dialog.  I think we'll have a hard time getting UX on board with using both terms.

mayank99:
> @gretanausedaite @mayank99, any thoughts on this?

Well, the page is called "dialog". And a modal dialog is still a dialog. So we can update most references to use the term "dialog" and clarify modal vs non-modal in a separate section.

As for the code/api, the `Dialog` component has more features and is more flexible, so most examples should show that. The `Modal` component only exists as a convenient way to show modal dialogs, because `Dialog` can be a bit verbose to assemble.

apps/website/src/pages/docs/dialog.mdx
```diff
@@ -13,58 +17,60 @@ import LiveExample from '~/components/LiveExample.astro';
...
-- Close button: only allowed in dismissible dialogs.
-- Content
-- Actions: see Button standard for best practices.
-- Box & shadow
-- Overlay: Dialog appears above all other UI elements and casts a black overlay as well as a drop shadow on all other content, creating a strong visual focus on the dialog.
+The terms dialog and modal are often used interchangeably. Within iTwinUI we differentiate the two terms like so:
 
-The dialog should be located in the upper half of the screen, because on smaller screens it may get lost. A dialog’s maximum width is 50% of the screen width & the minimum width is 400px. If your content can not fit inside the dialog without scrollbars, you should consider alternative ways to display the information.
+- Dialog: a dialog is the box of content that appears above the rest of the page content. The user can still interact with the page content.
+- Modal dialog: a modal dialog is the combination of a dialog with a backdrop that appears above the rest of the page content. This backdrop disables interaction with the page content until the dialog is dismissed.
```
gretanausedaite, line 31: These changes looks good.
Maybe you could mention that main example is Modal dialog? 
Or change main example to be Dialog with backdrop and add additional example of Modal dialog. 
(I can help you with that if you want)

FlyersPh9: I'd appreciate that, thank you.  Short meeting scheduled for Monday.

gretanausedaite (approved):

apps/website/src/examples/Dialog.dismissible.tsx
```diff
@@ -0,0 +1,43 @@
...
+export default () => {
+  const [isModalOpen, setIsModalOpen] = React.useState(true);
+  const closeModal = () => setIsModalOpen(false);
+
+  return (
+    <>
+      <Button styleType='high-visibility' onClick={() => setIsModalOpen(true)}>
+        Open dismissible dialog
+      </Button>
+      <Modal
```
mayank99, line 24: maybe most of the examples should use `Dialog`, not `Modal`, because it is more flexible and supports all use cases.

and then we can show `Modal` in just one example, saying that it is a "convenience" wrapper over `Dialog`.

@gretanausedaite thoughts?

apps/website/src/pages/docs/dialog.mdx
```diff
@@ -13,58 +18,62 @@ import LiveExample from '~/components/LiveExample.astro';
...
-- Title: must be descriptive of the task at hand. The user must understand the context because they went from the main window to a dialog.
-- Close button: only allowed in dismissible dialogs.
-- Content
-- Actions: see Button standard for best practices.
-- Box & shadow
-- Overlay: Dialog appears above all other UI elements and casts a black overlay as well as a drop shadow on all other content, creating a strong visual focus on the dialog.
+The terms dialog and modal are often used interchangeably. Within iTwinUI we differentiate the two terms like so:
 
-The dialog should be located in the upper half of the screen, because on smaller screens it may get lost. A dialog’s maximum width is 50% of the screen width & the minimum width is 400px. If your content can not fit inside the dialog without scrollbars, you should consider alternative ways to display the information.
+- Dialog: a dialog is the box of content that appears above the rest of the page content. The user can still interact with the page content.
```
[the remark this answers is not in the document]
FlyersPh9: Good suggestion, committed.

apps/website/src/examples/Dialog.dismissible.tsx
FlyersPh9: I have changes local that uses `Dialog` in every demo except for`Dialog.fullpage.tsx` (unsure how to achieve) & `Dialog.modal.tsx`.  Should I push or wait for Greta's response?

mayank99: go ahead and push it, you can always adjust/revert based on Greta's response

FlyersPh9: Pushed!

mayank99 (approved):

apps/website/src/examples/Dialog.fullpage.tsx
```diff
@@ -0,0 +1,44 @@
...
+        setFocus={false}
```
rohan-kadkol, line 29: Don't know if it is just for me for `setFocus` does not seem to exist on `ModalProps`. So VSCode shows an error when passing false to it.

Same in `Dialog.modal.tsx`

rohan-kadkol: Although TS is not allowing to pass `setFocus`, it is being applied using `{...rest}` which is why it does not focus which is good.

Any reason why we did not pick `setFocus` from `DialogMainProps`?

https://github.com/iTwin/iTwinUI/blob/090c60a2a4387e09a7967e471a63a8dbbd4c351d/packages/itwinui-react/src/core/Modal/Modal.tsx#L54

Can we pick `setFocus` and set the default to `true` for `Modal`?

apps/website/src/examples/Dialog.dismissible.tsx
gretanausedaite: I somehow missed that other examples were Modal and not Dialog. Good example of you see what you want to see.
Very much agree with this change.

apps/website/src/examples/Dialog.fullpage.tsx
gretanausedaite: I don't remember why we did not include setFocus. 
We have it added in Modal.tsx line 110 which somehow works.

rohan-kadkol: > We have it added in Modal.tsx line 110 which somehow works.

I believe adding `setFocus` on [L110](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/Modal/Modal.tsx#L110) of Modal.tsx is the shorthand to assign the value of `true` to a boolean prop. So the below two are the same:

```tsx
<>
<Dialog setFocus/>
<Dialog setFocus={true}/>
</>
```

So I guess should we allow to pass a `setFocus` in Modal and pass it down to the Dialog? The default value of the `setFocus` on Modal can be true to keep the same behavior.

mayank99: i believe this was done because Modal should always intercept and trap focus, unlike Dialog which can be non-modal.

PR #1002, opened 2023-01-13
feat(Tabs): Right aligned content within Tabs row (React)

rohan-kadkol:
## Changes

* This is the React implementation of the right aligned content PR #860.
* This adds a new prop called `actions` to Tabs.
  * `actions: React.ReactNode[]` when `type = 'default' | 'borderless'`
  * `actions: undefined` when `type = 'pill'` (since actions are not supported for pill)
* Restructured and renamed `TabsOrientationProps` to `TabsTypeProps`

## Testing

* Added one `size: 'small'` and one normal sized `<Button>` to the `actions` for all Tabs' stories, except for `Pill`.
  * Approved the test images.
* To test the types:
  * I added Tabs to the Next playground (in https://github.com/iTwin/iTwinUI/commit/8c9d574d5078cdf4b14a7507bb659770c0908093). I then tested/confirmed that:
  * Passing `actions` to `<Tabs>` with `type: 'pill'` gives a type error.
  * Passing `type: 'pill'` when `orientation: 'vertical'` gives a type error.

## Docs

* Added JSDocs for the new `actions` prop.
* Updated storybook with examples (more detail in the first point above under testing)


packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -75,12 +82,13 @@ export type TabsProps = {
...
   children?: React.ReactNode;
-} & TabsOrientationProps;
+} & TabsTypeProps;
 
 /**
  * @deprecated Since v2, use `TabProps` with `Tabs`
  */
-export type HorizontalTabsProps = Omit<TabsProps, 'orientation'>;
+export type HorizontalTabsProps = Omit<TabsProps, 'orientation'> &
+  TabsTypeProps;
```
rohan-kadkol, line 91: I'm not sure if we need to add this, since `HorizontalTabsProps` is deprecated. But I had to add this to avoid the TS error on [L358](https://github.com/iTwin/iTwinUI/blob/20b2973655bd87b125bbd220dd9ebb60d0e1629d/packages/itwinui-react/src/core/Tabs/Tabs.tsx#L358)

image

elephantcatdog (commented):
image
I'm guessing this is more of a CSS thing, but since you still have that open, can you make it so that the buttons are aligned on the bottom when the vertical tab section is tall? vs there being a space below the buttons.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Removing `actions?: undefined;` in L41 should get rid of the error without adding TabsTypeProps one more time. It's already included in TabsProps.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -14,24 +14,31 @@ import {
 import '@itwin/itwinui-css/css/tabs.css';
 import { Tab } from './Tab';
 
-type TabsOrientationProps =
+type TabsTypeProps =
```
elephantcatdog, line 17: Isn't changing this prop name a breaking change?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -14,24 +14,31 @@ import {
...
        */
-      type?: 'default' | 'borderless' | 'pill';
+      actions?: React.ReactNode[];
     }
   | {
-      orientation: 'vertical';
-      type?: 'default' | 'borderless';
+      type: 'pill';
+      orientation: 'horizontal';
+      actions?: undefined;
```
elephantcatdog, line 41: Why doesn't `actions` have the same type as above, `React.ReactNode[]`?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: Good point. Since we are not exporting `TabsTypeProps`, I think this should be fine.

Just to be safe and confirm with other reviewers, @gretanausedaite, is this right?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: Because `actions` is not supported for `type: 'pill'`. It is only supported for default and borderless tabs.

I also thought of not adding `actions` on L41, but that gave type errors inside `Tabs.tsx` itself and said that `actions` is not defined in `TabsProps`.

image

So I added `actions` as a prop, but gave it the type of `undefined`.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: We are not using these Props outside this file so should be ok.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: This doesn't make sense. We are omitting "orientation" prop and adding it back with TabsTypeProps. 
Yes, it does get rid of the typescript error but it creates different kind of problem.
Users can override "orientation" prop with vertical: 
image

With current typings typescript thinks that everything with "horizontal" orientation needs to have type "pill". Maybe you can rethink TabsTypeProps?

gretanausedaite (commented):
Don't forget to add changeset file.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: Not defining `actions: undefined` seemed to work. So now no change need to be done on the `HorizontalTabsProps` type. (https://github.com/iTwin/iTwinUI/pull/1002/files/55f64b16bcc8c5bc01ce2823f9f65f49a0ba6185..56b1c540a778ae26c1c327f53986c850c5e45a98)

But since `actions` are not existing when `type: pill`, we cannot get `actions` when destructuring `props`. Instead we have to access `actions` using `props.actions` and also check whether `props.type !== 'pill'` before accessing `props.actions` ([L339](https://github.com/iTwin/iTwinUI/pull/1002/files/55f64b16bcc8c5bc01ce2823f9f65f49a0ba6185..56b1c540a778ae26c1c327f53986c850c5e45a98#diff-b7bdf1756f6685cb3e8678eaedca9cac8289f5e72de3bb3381eb68bd7c31517aR339))

Just for clarity and to avoid logical type-related bugs, I split `TabsTypeProps` into `TabsOrientationProps` (from before this PR) and `TabsTypeProps` (new).

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: After https://github.com/iTwin/iTwinUI/pull/1002#discussion_r1084223719, now `actions` is not included in the type definition at all (not even as `actions?: undefined`). So my above comment of needing to include actions in the type as `actions?: undefined` is outdated.

mayank99 (approved):

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -117,6 +138,14 @@ export type VerticalTabsProps = Omit<TabsProps, 'orientation' | 'type'> & {
  * <Tabs labels={tabsWithIcons} type='pill' />
  */
 export const Tabs = (props: TabsProps) => {
+  // Separate actions from props to avoid adding it to the DOM (using {...rest})
+  let actions: Array<React.ReactNode> | undefined = undefined;
```
mayank99, line 142: it should already default to undefined

```suggestion
  let actions: Array<React.ReactNode>;
```

mayank99: seems you applied the suggestion before my edit. it defaults to undefined but you still need to make TS happy.
```
let actions: Array<React.ReactNode> | undefined;
```
or just use ReactNode
```
let actions: React.ReactNode;
```

elephantcatdog (approved):
Looks good! 🎉

PR #1006, opened 2023-01-17
Add `as` prop to `Anchor`

mayank99:
## Changes

Added type-safe, polymorphic `as` prop to `Anchor`. Same as what `Text` and `Button` already have.

This opens up two possibilities:
- using it as a `<Link>` from react-router, next-router, etc (less important because all of those are already polymorphic)
- using it as a `button` (more important because many UX mockups have this)

To support the latter, I had to undo default button styles.

## Testing

Unit and visual tests added.

Also tested with [`next/link`](https://nextjs.org/docs/api-reference/next/link) and polymorphism works fine there. `href` correctly accepts an object.

## Docs

Story, changeset and JSDoc added.


packages/itwinui-react/src/core/Typography/Anchor/Anchor.tsx
```diff
@@ -4,15 +4,42 @@
...
+  const { as: Element = 'a', className, ...rest } = props;
+
   useTheme();
-  return <a className={cx('iui-anchor', className)} ref={ref} {...rest} />;
-});
+
+  return (
+    <Element className={cx('iui-anchor', className)} ref={ref} {...rest} />
+  );
+}) as AnchorComponent;
```
rohan-kadkol, line 43: nit: instead of `as AnchorComponent`, you could add the type directly on L35 to Anchor (how it is in `Button`). 

```tsx
export const Anchor: AnchorComponent = React.forwardRef((props, ref) => {
```

.changeset/six-turkeys-type.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': patch
+---
+
+`Anchor` can now be rendered as a button using `as` prop.
```
rohan-kadkol, line 6: Do you want to include an example?

packages/itwinui-react/src/core/Typography/Anchor/Anchor.tsx
mayank99: did a bit differently, see [`b07612f`](https://github.com/iTwin/iTwinUI/pull/1006/commits/b07612f3e8a8a7b87637a4c902c8f7672ed5ad32)

.changeset/six-turkeys-type.md
mayank99: added. [`aae2d95`](https://github.com/iTwin/iTwinUI/pull/1006/commits/aae2d9595478affb465a0f268ca10424abb29e21)

rohan-kadkol (approved):

FlyersPh9 (approved):

PR #1007, opened 2023-01-17
Fix broken links on html demo pages

rohan-kadkol:
## Changes

I noticed that some images for [`header.html`](https://itwin.github.io/iTwinUI/css/header.html) were showing as broken. This happened only on the deployed version but not on the localhost version, not sure why. This PR makes small fixes to the image path.

image

I also searched for `../assets/` and found two occurrences in `menu.html`. Corrected the path there too.

## Testing

I verified that the images show correctly in the deploy preview for the PR.

## Docs

N/A


mayank99 (approved):
Looks good, lets merge this

PR #954, opened 2023-01-04
chore(Table): Add expanded state to `Table -> ControlledState` story

rohan-kadkol:
Like @bentleyvk suggested, added expanded state controls in the `Table -> ControlledState` story according to a related [CodeSandbox example](https://codesandbox.io/s/table-control-select-and-expand-state-oju3lx).

## Checklist

- ~[ ]~ Add meaningful unit tests for your component (verify that all lines are covered)
- [x] Verify that all existing tests pass
- ~[ ]~ Add component features demo in Storybook (different stories)
- ~[ ]~ Approve test images for new stories
- ~[ ]~ Add screenshots of the key elements of the component

PR #1008, opened 2023-01-17
`Table -> ControlledState` story without passing `useControlledState`

rohan-kadkol:
## Changes

This is a fix from PR #954 (https://github.com/iTwin/iTwinUI/pull/954#issuecomment-1373974864).

Using `useControlledState` requires us to completely control the state and keep it in sync with the table. This is not very safe and could lead to inconsistent states.

Thus, this PR updates the storybook example to use the `tableInstance` from `stateReducer` instead of using `useControlledState`. (similar to [codesandbox](https://codesandbox.io/s/table-control-select-and-expand-state-oju3lx))

Unrelated -- Removed two unused imports at the top of `Table.stories.tsx`.

## Testing

* Manually clicked the table rows and verified that the checkboxes update appropriately.
* Similarly, clicked checkboxes, confirmed table row update correctly.

## Docs

N/A


elephantcatdog (approved):
Looks okay to me

mayank99 (approved):

apps/storybook/src/Table.stories.tsx
```diff
@@ -1407,27 +1336,31 @@ export const ControlledState: Story<Partial<TableProps>> = (args) => {
...
+          return newState;
+        }, [])}
         isSelectable
+        onSelect={useCallback((selected) => {
+          setSelectedRows(selected ?? []);
+        }, [])}
+        onExpand={useCallback((expanded) => {
+          setExpandedRows(expanded);
+        }, [])}
+        getRowId={useCallback((rowData) => rowData.name, [])} // In real world use instance id
```
bentleyvk, line 1363: Maybe add `id` to data so no one would think of using `name` as a row id and random bugs if there are same names.

bentleyvk (approved):

apps/storybook/src/Table.stories.tsx
rohan-kadkol: Thanks, done (https://github.com/iTwin/iTwinUI/pull/1008/files/772767f26cb6a7601f7f9c1fdd3554935a1edc9e..9fddc890ea2d90cf4c76cd218f6f8f5927ea3a80)

PR #1011, opened 2023-01-17
Add `Flex` component

mayank99:
## Changes

Added `Flex` component and `Flex.Spacer`/`Flex.Item` subcomponents.

### Examples:

"text with icon"
```jsx
<Flex>
  <Icon><SvgPlaceholder /></Icon>
  <Text>Hey</Text>
</Flex>
```

"text with one icon on left + one icon pushed all the way to the end"
```jsx
<Flex>
  <Icon><SvgPlaceholder /></Icon>
  <Text>Hey</Text>

  <Flex.Spacer />

  <Icon><SvgPlaceholder /></Icon>
</Flex>
```

"flex items with different gap for one item"
```jsx
<Flex gap='xl'>
  <Flex.Item>...</Flex.Item>
  <Flex.Item gapBefore='xs'>...</Flex.Item>
  <Flex.Item>...</Flex.Item>
</Flex>
```

## Testing

Tested pretty extensively with various permutations in vite playground. I encourage reviewers to do the same, or at least play with the story controls (https://itwin.github.io/iTwinUI/1011/react/?path=/story/utilities-flex--basic).

Unit/visual tests added.

## Docs

Stories, jsdoc and changesets added. Basic documentation page added to website.


packages/itwinui-css/src/utils/spacing/flex.scss
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+@mixin iui-flex {
+  --iui-flex-display: flex;
+  --iui-flex-direction: row;
+  --iui-flex-gap: var(--iui-size-2xs);
```
mayank99, line 7: i chose a random default value for gap. we might want to adjust this.

also, we need to find designated gap values for common scenarios:
- vertical gap between inputs/buttons (e.g. forms)
- horizontal gap between inputs/buttons (e.g. toolbars, forms, button bars)
- horizontal gap between text and icon
- vertical gap between label and sublabel

packages/itwinui-react/src/core/utils/components/Flex.tsx
```diff
@@ -0,0 +1,271 @@
...
+   * This may need to be adjusted if you're also setting `flex` on
+   * other `Flex.Item`s.
+   *
+   * @default 999
+   */
+  flex?: React.CSSProperties['flex'];
+};
+
+// ----------------------------------------------------------------------------
+// Flex.Item subcomponent
```
mayank99, line 143: I kept everything in the same file and added these dividers.

For me this felt much easier to navigate and read, vs having every subcomponent in a separate file.

I can split into separate files if others feel differently.

packages/itwinui-css/src/utils/spacing/flex.scss
mayank99: @FlyersPh9

FlyersPh9: - **vertical gap between inputs/buttons (e.g. forms)**
We are using `var(--iui-size-s)` for vertical gap between inputs on [input](https://itwin.github.io/iTwinUI/css/input.html), [information panel](https://itwin.github.io/iTwinUI/css/information-panel.html), & [column filter](https://itwin.github.io/iTwinUI/css/column-filter.html).

- **horizontal gap between inputs/buttons (e.g. toolbars, forms, button bars)**
We are using `var(--iui-size-xs)` for horizontal gap between buttons in [dialog](https://itwin.github.io/iTwinUI/css/dialog.html) & [information panel buttons above the table](https://itwin.github.io/iTwinUI/css/information-panel.html).

- **horizontal gap between text and icon**
We are using `var(--iui-size-2xs)` for horizontal gap between icon & text in [input status message](https://itwin.github.io/iTwinUI/css/input), but `var(--iui-size-xs)` for the horizontal gap between icon & text in [table's name column](https://itwin.github.io/iTwinUI/css/table.html).  If we want to standardize I'd lean more towards `var(--iui-size-xs)`, but maybe it's ok for them to be different due to the different font sizes.

- **vertical gap between label and sublabel**
It would appear we have no gap used in [header](https://itwin.github.io/iTwinUI/css/header.html), [menu](https://itwin.github.io/iTwinUI/css/menu.html), & [expandable block](https://itwin.github.io/iTwinUI/css/expandable-block.html).  Altho in the expandable block it looks like we are hard coding a height for the text & not relying on font size to set the height.

I'd say the first 3 seem safe to standardize, but the vertical gap between label & sublabel we should investigate what we should use.

FlyersPh9:
What do you think if we update `yarn createComponent` to create a page within the documentation site?

gretanausedaite:
> But what does that have to do with this PR?

Nothing. Just a discussion. 😃

gretanausedaite (commented):
Code looks good, simple and pretty straight forward.

We should not merge without deciding default vertical gap between label and sublabel. [comment](https://github.com/iTwin/iTwinUI/pull/1011/files#r1082877722)
In FIleUpload PR we are using the same gap as in Header [comment](https://github.com/iTwin/iTwinUI/pull/984#issuecomment-1379800996). Maybe that could be our standard.

mayank99:
> We should not merge without deciding default vertical gap between label and sublabel. [comment](https://github.com/iTwin/iTwinUI/pull/1011/files#r1082877722) In FIleUpload PR we are using the same gap as in Header [comment](https://github.com/iTwin/iTwinUI/pull/984#issuecomment-1379800996). Maybe that could be our standard.

From what I understood, the vertical gap between label and sublabel cannot be easily standardized because we rely on line-height in many cases. Now, we could strip the line-height in all of those components and control the gap explicitly, but that seems a bit out of scope for this PR.

gretanausedaite:
> From what I understood, the vertical gap between label and sublabel cannot be easily standardized because we rely on line-height in many cases. Now, we could strip the line-height in all of those components and control the gap explicitly, but that seems a bit out of scope for this PR.

Can default be no gap / the smallest one and users add one if they need? It would match Header, Menu and FileTemplate scenarios.

packages/itwinui-css/src/utils/spacing/flex.scss
mayank99: I've changed the default gap to `--iui-size-xs`. This will automatically handle cases 2 and 3.

For case 1, I've added `--iui-spacing-form-vertical` to the global tokens. The name and location of this variable is debatable, but as far as the flex component is concerned, it can be used like this:
```jsx
<Flex gap='form-vertical'>
  ...
</Flex>
```

This will still not handle content shift when a status message appears, which makes me think we should exclude this variable from this PR and try to tackle these two things together.

FlyersPh9 (approved):

gretanausedaite (commented):
I agree that we should exclude custom variables from this PR. Leave only spacing tokens and document where to use them (documentation site variables page probably).

mayank99:
i've removed the `form-vertical` value for now. we can revisit in the future.

mayank99:
> Are you planning to add this component to website in other PR?

Added basic `flex.mdx` page.
- Demo will not work until component is released
- Missing thumbnail currently
- We need to find a better way to add decorative styles to demos. Inline styles are not good.

These improvements can be made in future PRs.

PR #1016, opened 2023-01-19
Feat(Multilevel Menu): Added Multilevel Menu

elephantcatdog:

### DIFF
--- a/packages/itwinui-css/backstop/tests/menu.html
+++ b/packages/itwinui-css/backstop/tests/menu.html
@@ -19,6 +19,7 @@
       import "./assets/theme.js";
       import "@itwin/itwinui-icons-elements/checkmark";
       import "@itwin/itwinui-icons-elements/placeholder";
+      import "@itwin/itwinui-icons-elements/caret-left-small";
       import "@itwin/itwinui-icons-elements/caret-right-small";
       import "@itwin/itwinui-icons-elements/imodel-hollow";
       import "@itwin/itwinui-icons-elements/caret-down-small";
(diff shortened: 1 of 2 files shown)


elephantcatdog:
I can't figure out why the skeleton and demo header test images are changing. I undid all of my changes and they still fail.

packages/itwinui-css/backstop/tests/menu.html
```diff
@@ -1892,6 +1893,400 @@ <h3>Borderless button</h3>
...
+        class="iui-menu iui-flat"
+        role="menu"
+      >
+        <li
+          class="iui-menu-item"
+          role="menuitem"
+          tabindex="0"
+          id="test-multilevel"
+        >
+          <span class="iui-content">Option 1</span>
```
rohan-kadkol, line 2176: nit: Should this have a subheading too? Because the unequal heights of the menu items seems a bit off.

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -65,6 +71,13 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
       margin-right: 0;
       margin-left: var(--iui-size-xs);
     }
+
+    &.iui-multilevel-selected {
+      display: flex;
+      justify-content: flex-end;
+      flex-shrink: 0;
```
rohan-kadkol, line 78: Adding `flex-shrink: 0` does not work too well when the `.iui-multilevel-selected` content is too long.

image

Removing this gave the below result:

image

Also, should there be a `min-width` (e.g. 30-70 split) between left and right?

elephantcatdog: Hmm, min-width % seems like a good way to fix it. I added `flex-shrink: 0` because of 
image

packages/itwinui-css/backstop/tests/menu.html
```diff
@@ -1892,6 +1893,400 @@ <h3>Borderless button</h3>
...
+              ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
+              voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+              proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</div>
+            <div class="iui-menu-description">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+              tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+              ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
+              voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+              proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</div>
+          </span>
+          <span class="iui-content iui-multilevel-selected">Option 1A</span>
```
elephantcatdog, line 2202: Should say "Option 2D"

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -193,3 +206,18 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
     margin-left: var(--iui-size-xs);
   }
 }
+
+@mixin iui-multilevel {
+  @include iui-list-menu;
```
mayank99, line 211: This `@include` causes potentially unnecessary duplication of CSS. If you only need a few declarations from the original mixin, it might be better to move those out into a separate intermediate mixin that can be used in both `iui-multilevel` and `iui-list-menu`.

The same comment also applies below to `@include iui-menu-item`.

packages/itwinui-css/backstop/tests/menu.html
```diff
@@ -1892,6 +1893,400 @@ <h3>Borderless button</h3>
...
+        </ul>
+      </div>
+
+      <br />
+
+      <div
+        class="iui-multilevel"
+        role="menuitem"
+        tabindex="0"
+        id="test-multilevel"
```
mayank99, line 2016: This markup is invalid.
- any element with `role-menuitem` must be a child of an element with `role-menu`
- `id`s must be globally unique

Also, `tabindex` should not be `0` because arrow keys (not <kbd>Tab</kbd> jey) are supposed to be used for keyboard navigation. But looks like that's an issue with the rest of the page too

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -32,6 +32,12 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
     overflow-y: overlay;
     --iui-menu-max-height: calc((var(--iui-component-height) - 1px) * 8.5); // Max 8.5 items in .iui-scroll
   }
+
+  &.iui-flat {
```
mayank99, line 36: This `flat` variant looks unrelated to the whole multi-level menu. Might not belong in this PR.

mayank99 (commented):
I took a look at the PR and the visuals look good. I do have a couple of questions.

Right now it's only showing the final, static state. Have you thought about the actual interactions that will needed to be handled in React? Things like:
 - what happens to the main menu after navigating to the next level? Is it kept in the DOM and hidden?
 - how is the menu item in the main menu connected with the next level? something like `aria-haspopup`? and how is the header linked back to the main menu?
 - how do you handle keyboard navigation
   - moving focus between root menu and nested menus
   - moving focus between the multilevel header and the menu below it
 - whether the markup is semantically right, and how it is announced by screen readers
 - how much and what state the user can control

These things can affect the CSS and HTML, so I would suggest handling the React now, and testing in playgrounds. It can be in the same PR or a separate PR that can be opened in parallel and reviewed together with this one. Until then, this PR could be moved back to draft. If you have any specific questions about the HTML/CSS, you could leave comments and `@` team members or set up a meeting.

Other than that, I only have minor comments

elephantcatdog:
> These things can affect the CSS and HTML, so I would suggest handling the React now, and testing in playgrounds. It can be in the same PR or a separate PR that can be opened in parallel and reviewed together with this one. Until then, this PR could be moved back to draft. If you have any specific questions about the HTML/CSS, you could leave comments and @ team members or set up a meeting.

This is a good suggestion; thanks, I'll do that. And thanks for the list of things to think about.

packages/itwinui-css/backstop/tests/menu.html
elephantcatdog: I fixed the `role` (throughout the page) and `id` issues. I left the `tabindex` as-is because it looked like adding arrow-key navigation would require javascript work and I didn't want to add that at this point. If you think it's important, I can make an issue for it.

packages/itwinui-css/src/menu/menu.scss
elephantcatdog: Updated. I tried to make it understandable.

packages/itwinui-css/backstop/tests/menu.html
elephantcatdog: Updated example

packages/itwinui-css/src/menu/menu.scss
elephantcatdog: I didn't use `min-width`, but I did fix this issue by changing the css used for `iui-multilevel-selected`. I also added an example for this scenario. I did a 30/title - 70/selection split, but I can change that to something else if need be.

packages/itwinui-css/backstop/tests/menu.html
elephantcatdog: I added a subheading here and updated the examples around it.

PR #1018, opened 2023-01-20
Fix(Table): Anchors in stories are now keyboard accessible

elephantcatdog:
## Changes
I added `as='button'` to all of the `Anchor`s in the table stories so that they are keyboard accessible.

image

## Testing
I went through all of the table stories and made sure that
  1) anchors are accessible with tabbing and
  2) anchors will fire the `onClick` function when 'enter' is pressed

I updated the test images for the table stories with anchors.

## Docs
N/A (besides changing these table stories)


FlyersPh9:
Do we know why the test images are changing?  I would have thought test images wouldn't change.

elephantcatdog:
Do we know why the test images are changing? I would have thought test images wouldn't change.

It's getting shifted over a bit because of the added CSS for `.iui-anchor:where(button)`, specifically `border: none;`.

image

FlyersPh9 (approved):

adamhock (approved):

adamhock:
I just noticed that not all of the visual storybook tests are passing for `Table`, so make sure those are all squared away before the merge!

elephantcatdog:
@FlyersPh9 @adamhock The anchors were updated by someone such that they don't affect the test images.

PR #1023, opened 2023-01-20
the one which fixes font when anchor is used as button

mayank99:
## Changes

Follow up to #1006.

`<button>` does not automatically inherit surrounding font, so when anchor is used as a button, it ends up looking slightly different. This change fixes that.

## Testing

Tested in storybook.

## Docs

Changeset not needed because #1006 is not yet released.


gretanausedaite (approved):
Love PR name, you might need update visual test image.

rohan-kadkol (approved):

PR #1025, opened 2023-01-23
Toaster: make `closeAll` function `async` to allow `await`ing it

### LINKED ISSUE #1005: Toast is not shown during first trigger
After upgrading to iTwinUI v2, toasts do not show up the first time they are triggered but it works fine from the second time (only in our project). I tried recreating the problem in a new project but it works correctly there.

#### Expected behavior
Toast should be visible the first time it is triggered.

#### Reproduction
Could not be reproduced on a new project.

#### Additional information
iTwinUI - 2.2.0
React - 18

We did not modify our code for toasts after upgrading to v2. Also, the code is executed and toast is being created during the first trigger but it just doesn't show up. Our code is simple:
```
const toastSuccess = (message: string) => {
    toaster.closeAll();
    toaster.positive(message, { type: "temporary" });
  };
```
I guess this is some kind of timing issue, because, for example, if I add a 2 second sleep function between toaster.closeAll() and toaster.positive() it does show up during the first trigger (or calling toaster.closeAll() during the first render of the app).

mayank99:
## Changes

Instead of hoping that promises resolve correctly, users can now `await` close calls to our toaster before opening other toasts.
```jsx
(async () => {
  await toaster.closeAll();
  toaster.informational();
})();
```

Does it fix #1005? Does it break anything? Lets find out!!

Edit: It *does* break things. I'll figure out a different approach. Stay tuned.

Edit 2: Fixed failing tests by not `await`ing creation functions. See https://github.com/iTwin/iTwinUI/pull/1025#discussion_r1084369360.

## Testing

Pending.

## Docs

Pending


packages/itwinui-react/src/core/Toast/Toaster.tsx
```diff
@@ -139,38 +139,41 @@ export default class Toaster {
       },
       ...(this.settings.order === 'descending' ? this.toasts : []),
     ];
+
+    // cannot `await` this, for backwards compat with the return value
     this.updateView();
+
     return { close: () => this.closeToast(currentId) };
```
mayank99, line 146: i wanted to make all methods `await`able, but that would cause breaking changes for users who are using the return value like this:

```jsx
const { close } = toaster.positive();
```

i thought about keeping backwards compatibility by adding a promise as a new property in the return value. but no one has asked for it and i don't see the benefit yet, so i decided against it.

```suggestion
    const updatePromise = this.updateView();
    const done = new Promise<boolean>(async (resolve) => {
      await updatePromise;
      resolve(true);
    });

    return { close: () => this.closeToast(currentId), done };
```

Issue #886, opened 2022-11-15
Table: improve typing for `columns`


mayank99:
https://github.com/iTwin/iTwinUI-react/pull/940 was a hacky workaround to fixing build errors for users, and it even resulted in our own vite playground failing when running `tsc`. Need to investigate this issue and have proper types (without `any`).

#1514:
- Build must pass even if user doesn't have a strong type for `columns`.
- The user must NOT be required to cast their columns using `as Column[]`.

Others:
- Table stories should remove `{...args}` because it hides TS errors. (#1543)

TODOs:
- There must be a proper test for more complicated permutations of the Table in a testing app (e.g. nextjs or vite, via playwright).
- ~~Vite playground build should ideally reintroduce `tsc`~~ (done) and pass (~~playground build currently does not run in CI~~ done).
  - Same with storybook ideally.

rohan-kadkol:
This issue has multiple parts:
1. Give a type error when a top-level header is passed (to reinforce that we don't support multi-level headers) (https://github.com/iTwin/iTwinUI/issues/886#issuecomment-1362950474). I worked mostly on this part.
2. The other parts are listed in the issue description and other comments.

---

I considered a couple of approaches for part `#1`:
* To cut-paste the code from `@types/react-table` to `itwinui-react` (`base-react-table.d.ts`) and remove dep on `@types/react-table`
    * Users' apps don't depend on `react-table`; but they download it since we depend on it. So if we don't depend on it, they won't download it. Thus, `Column` will not be available and they cannot typecast their columns to `Column[]`. Thus, we can only remove this dep when/if we don't need users to use types from `react-table`.
  * Also, if we depend on `@types/react-table`, then users also indirectly depend on it. So when they download the dependency, their version includes the `columns` in `ColumnGroupInterface` even though we exclude it in our version. So I'm not sure how to exclude something that has been declared in the base package. Or maybe I misunderstood how TS works.
* In `react-table-config.ts`, override `columns` to type of `undefined` or similar by overriding `ColumnGroupInterface` inside of `react-table-config.ts`
  * This didn't work since same-name variables should have the same type.
  * ![image](https://user-images.githubusercontent.com/45748283/215583892-c814d567-614d-475d-808a-b964a12e3aaa.png)
* Add `override.d.ts` file
  * This allows to add on to the types in `@types/react-table`. It doesn't allow to exclude declarations already mentioned in `@types/react-table`.
[...]

PR #1026, opened 2023-01-23
Fix typing for Table `columns`

rohan-kadkol:
## Changes

This PR is to fix a portion of #886. The aim is to throw type errors when passing sub-columns. This is since we don't support grouped headers.


rohan-kadkol:
Couldn't get this working. Thought process explained in https://github.com/iTwin/iTwinUI/issues/886#issuecomment-1410585721

PR #1029, opened 2023-01-25
remove all instances of `@at-root`

mayank99:
## Changes

Either removed `@at-root` completely (was redundant) or replaced with `:is` where needed.

This is in preparation of scoping v2 classes, because scoping will rely on nesting, and `@at-root` breaks nesting.

## Testing

All tests should pass.

## Docs

N/A

### DIFF
--- a/packages/itwinui-css/src/input/input.scss
+++ b/packages/itwinui-css/src/input/input.scss
@@ -48,7 +48,7 @@
     cursor: not-allowed;
   }
 
-  @at-root :where(textarea#{&}) {
+  &:where(textarea) {
     @include textarea;
   }
 }
--- a/packages/itwinui-css/src/table/table.scss
+++ b/packages/itwinui-css/src/table/table.scss
@@ -225,8 +225,9 @@
         border-bottom-color: var(--iui-color-border-accent);
       }
 
+      /* stylelint-disable-next-line selector-max-compound-selectors */
       + :not([aria-selected='true']) .iui-table-cell,
-      + :is(.iui-table-expanded-content + .iui-table-row:not([aria-selected='true'])) .iui-table-cell {
+      + .iui-table-expanded-content + .iui-table-row:not([aria-selected='true']) .iui-table-cell {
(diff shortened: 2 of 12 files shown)


packages/itwinui-css/src/input/input.scss
```diff
@@ -48,7 +48,7 @@
     cursor: not-allowed;
   }
 
-  @at-root :where(textarea#{&}) {
+  &:where(textarea) {
```
rohan-kadkol, line 51: Specificity goes up with this. Should we have something like this to keep the old specificity?

```suggestion
  :where(textarea:is(&))
```

```css
/* Old */
:where(textarea.iui-input) {} /* 0,0,0 */
:where(textarea.iui-input)[disabled], :where(textarea.iui-input)[readonly] {} /* 0,1,0 each */

/* New */
.iui-input:where(textarea) {} /* 0,1,0 */
.iui-input:where(textarea)[disabled], .iui-input:where(textarea)[readonly] {} /* 0,2,0 each */
```

packages/itwinui-css/src/table/table.scss
```diff
@@ -146,10 +146,11 @@
     border-radius: inherit;
   }
 
-  @at-root .iui-table-header & {
+  .iui-table-header & {
     flex-grow: 1;
   }
-  @at-root .iui-table-body & {
+
+  .iui-table-body & {
```
rohan-kadkol, line 153: nit: Unnecessary formatting change

packages/itwinui-css/src/input/input.scss
mayank99: i think it doesn't really matter

packages/itwinui-css/src/table/table.scss
mayank99: where? `@at-root` was removed from here.

rohan-kadkol (approved):

gretanausedaite (approved):

FlyersPh9 (approved):

PR #1030, opened 2023-01-25
DropdownButton: Add dropdownMenuProps

### LINKED ISSUE #1028: DropdownButton: Expose popover props to control dropdown position
We should have the option to allow the dropdown button to overflow to the right instead of always to the left. It looks like popover is being used under the hood. If we added a popoverProps option, we could override the default placement.

In this example, we'd like to have the dropdown displayed under the input, but currently there isn't a way.

https://codesandbox.io/s/dropdown-button-to-right-v3got1?file=/src/App.tsx

image

mayank99:
## Changes

Added `dropdownMenuProps` to give user full control over the `DropdownMenu` which wraps the button. Closes #1028.

We are going to get rid of `tippy.js` so this prop will be removed in the next major release, but it's pretty useful to have right now.

## Testing

Basic unit test added. Didn't add visual test. This doesn't need extensive testing because it is not really a new feature, it's just propagating the props.

## Docs

Changeset added.

### DIFF
--- a/.changeset/kind-zoos-boil.md
+++ b/.changeset/kind-zoos-boil.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`DropdownButton` now exposes `dropdownMenuProps` to allow control over all popover props, such as `placement`.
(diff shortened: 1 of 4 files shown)


rohan-kadkol (approved):

gretanausedaite (approved):
We should add example for users how to use this new functionality because most of them won't even realise this is possible. Either in storybook or in website. Probably best would be both.

mayank99:
> We should add example for users how to use this new functionality because most of them won't even realise this is possible. Either in storybook or in website. Probably best would be both.

I left out the story because this is not a new feature. It's the same as what is already possible in all our other popover components.
Agree we should document this on website though.

PR #1031, opened 2023-01-25
scoping for v2

mayank99:
## Changes

Wrapped all css in every component with `:where(.iui-root, [data-iui-theme])` using [`postcss-scoper`](https://github.com/ro1963855/postcss-scoper). This solves half the problem by making sure that v2 styles stay inside the "root" and don't leak out.

The other half is to prevent v1 styles from leaking into v2. This is done by setting `all: revert` on every iui element that is a child of v2 root and also a child of a v1 root. The specificity of this selector is 0, so it relies on #1037 (unlayered styles always take preference over layered styles, regardless of specificity).

Both of these changes together make it possible for apps to incrementally migrate to v2 from the bottom up.

For packages, there is also a new `'inherit'` theme which will use the same theme as a parent ThemeProvider (rather than defaulting to light).

There is technically a breaking change in itwinui-variables - `iui-root` must now always be set instead of relying on variables being automatically added to `:root`. But in practice, this should not affect anyone because we automatically set this class in itwinui-react, and even in the itwinui-css readme, we ask users to set it.

## Testing

Tested a basic version in stackblitz. https://stackblitz.com/edit/vitejs-vite-7qrmeg?file=src%2FApp.jsx

Tested in a real app by releasing these dev versions:
```
"@itwin/itwinui-css": "0.63.2-dev.1",
"@itwin/itwinui-react": "1.48.2-dev.2",
```
```
"@itwin/itwinui-css": "1.5.0-dev.2",
"@itwin/itwinui-react": "2.4.0-dev.3",
"@itwin/itwinui-variables": "2.0.0-dev.1",
```

## Docs

Added changeset. We should probably also document this in the wiki.


packages/itwinui-css/src/anchor/anchor.scss
```diff
@@ -20,7 +20,7 @@
     @include iui-anchor-underline('on-initial');
   }
 
-  [data-iui-contrast='high'] & {
+  :is([data-iui-contrast='high'] &) {
```
mayank99, line 23: without this change, the browser will look for `[data-iui-contrast=high]` inside `:where(.iui-root, [data-iui-theme])` (which is applied through postcss). this doesn't work because both of those are applied on the same element.

with this change, the browser will first resolve `[data-iui-contrast=high] .iui-anchor`, and then separately make sure that `iui-anchor` is also inside `:where(.iui-root, [data-iui-theme])`.

for more details, see https://www.bram.us/2023/01/17/using-is-in-complex-selectors-selects-more-than-you-might-initially-think/

apps/storybook/.storybook/preview.js
```diff
@@ -55,6 +55,7 @@ export const decorators = [
     } = context;
 
     React.useEffect(() => {
+      document.body.classList.toggle('iui-root', true);
```
mayank99, line 58: this resulted in the Icon test image changing ever so slightly. i noticed that the old size was 16x16 whereas now it's 15.99x15.99 for some reason 😄

FlyersPh9 (approved):

FlyersPh9:
Can we do a dev release and show to Caleb or a a team that could test this soon?

gretanausedaite:
> Can we do a dev release and show to Caleb or a a team that could test this soon?

I think we should coordinate with original requestor to test if dev versions works in their app.

mayank99:
I've added a new `'inherit'` theme specifically for packages. The idea is that packages using v2 will wrap everything with `<ThemeProvider theme='inherit'>`. This ensures that the v2 styles stay scoped to the package, and any app that is on v1 will not break and can incrementally adopt the new package. At the same time, this gives apps the ability to control the theme (rather than it being hardcoded in the package).



This is released in the following dev versions:
```
"@itwin/itwinui-css": "1.5.0-dev.2",
"@itwin/itwinui-react": "2.4.0-dev.3",
"@itwin/itwinui-variables": "2.0.0-dev.1",
```

In a future major version, we might want to make `inherit` the default.

apps/storybook/.storybook/preview.js
mayank99: there is another "random" test image change. this one can be explained by a change in font size.


see main branch: https://itwin.github.io/iTwinUI/react/iframe.html?args=&id=utilities-flex--basic
vs this branch: https://itwin.github.io/iTwinUI/1031/react/iframe.html?args=&id=utilities-flex--basic

adamhock (approved):
Looks good to me! 🎉

mayank99:
I've made one more change. If `theme` is set to `'inherit'`, then `applyBackground` will default to `false`. Since `'inherit'` is meant for packages, it should not have any opinion on the styling.

This was done based on testing in a real scenario. Before this change, sync-report was unnecessarily getting a grey background.

gretanausedaite (approved):

PR #1032, opened 2023-01-25
move-on-visible transition in docs website

rohan-kadkol:
## Changes

* Added a util class called `move-on-visible` that animates the tag when it becomes visible
* This effect includes:
  * Movement of element that is controlled by `data-move-on-visible-direction="..."`
    * Possible values: `up`, `down`, `left`, `right`
  * Unblur: `blur(30px)` to `blur(0px)`
  * Opaque: `opacity: 0.5` to `opacity: 1`
* Unrelated change: Added `border-radius` to the `.demo-box` ([L252](https://github.com/iTwin/iTwinUI/pull/1032/files#diff-3ea7493681e865ffe2139eebd0671aa0f6fcee2bef4c1afdefe908cc9e43e292R252))

[Video.webm](https://user-images.githubusercontent.com/45748283/214714270-ed4fb193-27a1-4785-9f7e-5a8910e8360b.webm)

## Testing

Manually tested the docs website to see if the animations work as intended and do not break existing behavior.

## Docs

N/A

## TODO:

* [x] Confirm that animations work correctly on smaller-width screens
  * Change `data-move-on-visible-direction=left or right` to `up` on smaller screens.
  * This is because all `move-on-visible` content is all arranged vertically on smaller screens (no grid). So moving it `up` on visible is better than moving `left/right`.


apps/website/src/pages/index.astro
```diff
@@ -435,4 +445,28 @@ const currentPage = new URL(Astro.request.url).pathname;
...
+      const visibleBottomLine = window.scrollY + window.innerHeight;
+      const demoWrappers = document.querySelectorAll('.move-on-visible');
+      demoWrappers.forEach((demoWrapper) => {
+        if (!(demoWrapper instanceof HTMLElement)) {
+          return;
+        }
+
+        const top = demoWrapper.getBoundingClientRect().top;
+        if (top < visibleBottomLine) {
+          demoWrapper.dataset.visible = 'true';
```
mayank99, line 460: `data-visible` is too generic and might already be used in other places. maybe scope this as `data-move-visible` or something

apps/website/src/pages/index.astro
```diff
@@ -435,4 +445,28 @@ const currentPage = new URL(Astro.request.url).pathname;
...
+        } else {
+          delete demoWrapper.dataset.visible;
+        }
+      });
+    };
+
+    updateVisibility();
+    document.querySelector('body').addEventListener('scroll', () => {
```
mayank99, line 468: have you tried using intersection observer instead of scroll events? it's a lot more efficient

apps/website/src/components/utils/Footer.astro
```diff
@@ -3,7 +3,9 @@ const props = Astro.props;
 const currentYear = new Date().getFullYear();
 ---
 
-<footer {...props}>© {currentYear} Bentley Systems, Inc</footer>
+<footer class='move-on-visible' data-move-on-visible-direction='down' {...props}>
```
mayank99, line 6: do we really need to animate the footer?

apps/website/src/pages/index.astro
```diff
@@ -83,10 +83,10 @@ const currentPage = new URL(Astro.request.url).pathname;
       </div>
     </div>
 
-    <div class='centered-content'>
+    <div class='centered-content move-on-visible' data-move-on-visible-direction='up'>
```
mayank99, line 86: instead of animating in multiple steps, this might be better if the whole section animates together. will look better on taller displays, and less distracting for anyone.

apps/website/src/pages/index.astro
rohan-kadkol: I initially did that, but then it was even unbluring and moving when scrolling from bottom to top. I felt that was too distracting. So with the current approach where it only animates on scrolling from top to bottom.

But now that I think of it, I think if I use an intersection observer itself and check the scroll position in the callback, it will give the same behavior too. I will try that.

apps/website/src/components/utils/Footer.astro
rohan-kadkol: I feel not animating it might feel abrupt. Because if the user sees that most sections or items animate on becoming visible, they would expect the footer to do the same. If the footer was tall (more than just one sentence) and had a background color, then I feel animating it may not be needed.

apps/website/src/pages/index.astro
rohan-kadkol: I did it this way since I thought it looks better. I don't feel it is distracting. On taller displays, since items display one at a time, it might actually make each item stand out. Seeing all at once makes it look like too much content at once.

[Video16.webm](https://user-images.githubusercontent.com/45748283/214939266-b164d1ab-0194-415f-b373-36eaa0ca5b17.webm)

But I can change it if you feel it is distracting.

mayank99: well, of course it is not distracting if you slowly scroll a few pixels at a time, but it will be when you scroll a large amount all at once.

maybe you can add some kind of debouncing, like 50ms, and then decide how much to animate.

apps/website/src/pages/index.astro
rohan-kadkol: Changed `data-visible` to `data-move-visible`

gretanausedaite (commented):
It is a cool effect. 
However it's a bit too slow for my liking - it got me dizzy after some scrolls.
Maybe we could make it move on first scroll only? If you scroll down to the bottom and then scroll up, the bottom content is already "there".
Also what about people with reduced motion preference?

apps/website/src/pages/index.astro
rohan-kadkol: Used `IntersectionObserver` instead. https://github.com/iTwin/iTwinUI/pull/1032/files/0f4e8e178270263fd4366a316a5acb3c66bb582a..c464aa1dc29ddda3ec3606bf6786cacafb5998ab

apps/website/src/pages/index.astro
rohan-kadkol: I'm not sure how to apply debouncing to an animation. Or maybe you meant add debouncing to the `onScroll` listener? (Now I'm using `IntersectionObserver`)

For now, I just made the sections animate as one unit. So I removed the animation from the children of the "Why iTwinUI" and "Our Library" sections.

apps/website/src/pages/index.astro
mayank99: can you increase the [threshold](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#thresholds) a little bit? right now, even if 1% of the section is visible, it will animate.

for example, just the top of the "why itwinui" is visible in this viewport size, so it was animating when the page loads, before even scrolling

apps/website/src/pages/index.astro
mayank99: i meant debouncing where you set the data attributes yeah. but animating the sections as one unit might be enough.

there is also a new [`scrollend`](https://developer.chrome.com/blog/scrollend-a-new-javascript-event/) event, available through a polyfill. it could be useful here

mayank99:
> It is a cool effect. However it's a bit too slow for my liking - it got me dizzy after some scrolls.

Agreed with this. Can we make it less dizzying? Changing it to be less blurry and faster might help.

apps/website/src/pages/index.astro
```diff
@@ -435,4 +436,52 @@ const currentPage = new URL(Astro.request.url).pathname;
       }
     }
   </style>
+
+  <script>
```
mayank99, line 440: all of this could be moved to a separate file

mayank99: it can be a normal JS/TS file, or it can even be a .astro file.

if you go with .astro, you'll be able to move the styles into that file as well

apps/website/src/pages/index.astro
rohan-kadkol: Added a threshold of `[0.3]`

rohan-kadkol:
> > It is a cool effect. However it's a bit too slow for my liking - it got me dizzy after some scrolls.
> 
> Agreed with this. Can we make it less dizzying? Changing it to be less blurry and faster might help.

To make it less dizzying, I removed the blur altogether. I only animated the `opacity` from `0` to `1` and animated the `transform (translate)`. I also reduced this translation offset from `60px` to `30px` so that there is lesser movement. I also removed the animation from the footer.

<details>
<summary>New screenshot</summary>

[Video18.webm](https://user-images.githubusercontent.com/45748283/215130356-5fbc03d3-998a-4f23-9388-cba0298dc928.webm)

</details>

apps/website/src/pages/index.astro
rohan-kadkol: Moved the `<style>` and `<script>` into a separate file called `MoveOnVisible.astro`. I then included used that component in the `_global.astro` file.

apps/website/src/components/utils/MoveOnVisible.astro
```diff
@@ -0,0 +1,99 @@
+---
+---
```
mayank99, line 2: could add some comment what this file is for and how to use it

apps/website/src/pages/index.astro
```diff
@@ -27,7 +27,7 @@ const currentPage = new URL(Astro.request.url).pathname;
   <main>
```
mayank99, line 27: I am noticing a bit of content shift as the scrollbar appears/disappears.

https://user-images.githubusercontent.com/9084735/215138184-c00846fc-42fb-4ff3-884a-43a0e6d2cf73.mov

should be fixable by using [`scrollbar-gutter: stable`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) on the element which scrolls

mayank99:
> To make it less dizzying:
> * Removed the blur altogether
> * Only animated the `opacity` from `0` to `1` and animated the `transform (translate)`
>   * Reduced this translation offset from `60px` to `30px` so that there is lesser movement

Looks pretty good! 👍

apps/website/src/components/utils/MoveOnVisible.astro
rohan-kadkol: I tried looking for a way to add something similar to JSDocs in Astro, so that hovering over `<MoveOnVisible/>` in a code editor would show the docs. But I couldn't find how to do this.

So I just added a normal JSDocs comment with some info. I can remove some parts of the JSDocs if it's unnecessary info.

mayank99: i meant comments about what data attributes to put where. currently those two bullet points dont really add much

rohan-kadkol: Updated [`b4d38c3` (#1032)](https://github.com/iTwin/iTwinUI/pull/1032/commits/b4d38c3d9a8c64cd4f736fd20be3bd336b8e15cf)

apps/website/src/pages/index.astro
rohan-kadkol: I noticed that too before this PR. I feel this is because the scroll bar width somehow changes. This change happens only when the React demos are loaded.

To test this better, I set a width of `100px` to the `::-webkit-scrollbar` and removed `background-clip` from `::-webkit-scrollbar-thumb`.

The scrollbar automatically changes sizes only when including the react demos in the dom.

<details>
<summary>With react demos</summary>

[Video19.webm](https://user-images.githubusercontent.com/45748283/215155873-089fc931-15ca-4644-896b-80e6fd744ae5.webm)

</details>

<details>
<summary>Without react demos</summary>

image

</details>

Is there something unique about the demos that might be causing this?

mayank99: i think this might be because itwinui-react automatically adds global styles? but i'd say dont worry about it in this PR, it's very minor and can be handled separately

mayank99 (approved):

elephantcatdog (approved):
Looks nice!

PR #1040, opened 2023-01-31
`<Avatar/>` supports svgs passed to `image` prop

rohan-kadkol:
## Changes

Avatar's `image` prop can now accept `<svg>` children and display it correctly.

## Testing

Added a unit test to confirm svgs passed to the `image` prop is included in the dom. Added a `With Icon` story and tested the the svg displays properly with the background color.

## Docs

Added a changeset

## TODOs

- [x] Add/approve visual and unit tests


mayank99:
Maybe I missed something, but what is the benefit of using `<Icon>` inside `<Avatar>` if the size is being overridden anyway? The user should just pass an svg directly.

Also, what is the benefit of the deprecation? The same `image` prop could be reused.

rohan-kadkol:
> Maybe I missed something, but what is the benefit of using `<Icon>` inside `<Avatar>` if the size is being overridden anyway? The user should just pass an svg directly.

There's no benefit. I just thought that since we're allowing users to wrap svgs in `<Icon>`, they might do this all the time. So they might also pass `<Icon><SvgAdd/></Icon>` to Avatar. So we should catch this case also.

> Also, what is the benefit of the deprecation? The same `image` prop could be reused.

I could reuse the `image` prop. I thought of changing it to `child` to make it more clear that svgs (or other elements) can be passed.

> The explicit `iui-icon` class also does not need to be supported.

I included that test since that is how icons are added in avatars in the avatar HTML demo page.

image

Also, the styling in avatar.scss targets `.iui-icon` (https://github.com/iTwin/iTwinUI/pull/1040/files#diff-be3c849b1fcfeb4dda01cfe9c31aa0e237ab0e06d71a2838d5ddd77c1700f05aL51-R53)

mayank99:
> There's no benefit. I just thought that since we're allowing users to wrap svgs in `<Icon>`, they might do this all the time. So they might also pass `<Icon><SvgAdd/></Icon>` to Avatar. So we should catch this case also.

If there is no benefit, then lets not add the extra complexity. Currently the icon component is coupled to the avatar for no reason.

> I could reuse the `image` prop. I thought of changing it to `child` to make it more clear that svgs (or other elements) can be passed.

`image` seems fine to me. svgs are technically still images. But if you want to deprecate it, then you could use `children` instead of `child`. it feels more natural in JSX.

> > The explicit `iui-icon` class also does not need to be supported.
> 
> I included that test since that is how icons are added in avatars in the avatar HTML demo page.
> 
> Also, the styling in avatar.scss targets `.iui-icon` ([#1040 (files)](https://github.com/iTwin/iTwinUI/pull/1040/files#diff-be3c849b1fcfeb4dda01cfe9c31aa0e237ab0e06d71a2838d5ddd77c1700f05aL51-R53))

The class names are considered "implementation detail" for React components. It's totally fine to support them in CSS, but React users are never expected to pass the `iui-icon` class. In the future we could rename this class to `iui-avatar-icon` or something, and that should not break react users.

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -48,7 +48,9 @@
     font-size: var(--_iui-avatar-font-size);
   }
 
-  > .iui-icon {
+  > .iui-icon,
+  > svg,
+  > .iui-svg-icon {
```
mayank99, line 53: Now that I think about it, `iui-icon` is actually unnecessary since we can just target `svg` without `>`.
```suggestion
  svg {
```

packages/itwinui-react/src/core/Avatar/Avatar.tsx
```diff
@@ -105,11 +112,10 @@ export const Avatar = (props: AvatarProps) => {
       style={style}
       {...rest}
     >
-      {image ?? (
-        <abbr className='iui-initials' style={{ backgroundColor }}>
-          {abbreviation?.substring(0, 2)}
-        </abbr>
-      )}
+      <abbr className='iui-initials' style={{ backgroundColor }}>
+        {image || child ? '' : abbreviation?.substring(0, 2)}
```
mayank99, line 116: what's the reason for this change? `image` will now be included twice

apps/storybook/src/Avatar.stories.tsx
```diff
@@ -4,7 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 import { Meta, Story } from '@storybook/react';
 import React from 'react';
-import { getUserColor, Avatar, AvatarProps } from '@itwin/itwinui-react';
+import { getUserColor, Avatar, AvatarProps, Icon } from '@itwin/itwinui-react';
+import { SvgAdd } from '@itwin/itwinui-icons-react';
```
mayank99, line 8: could use `SvgUser`
https://itwin.github.io/iTwinUI-icons/#icons-user

packages/itwinui-react/src/core/Avatar/Avatar.tsx
rohan-kadkol: Not sure what do you mean by _included twice_. I did this change since in the HTML demo, the `<abbr>` with empty text is used as the background for images. So if image or child is passed, the text is empty.

apps/storybook/src/Avatar.stories.tsx
rohan-kadkol: Ohh, right. Used `<SvgUser>`

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -48,7 +48,7 @@
     font-size: var(--_iui-avatar-font-size);
   }
 
-  > .iui-icon {
+  svg {
```
FlyersPh9, line 51: Is this considered a breaking change for our CSS pkg?

rohan-kadkol: Ohh, right, it looks like a breaking change. It also breaks the avatar demo html page.

image

I auto-committed the suggestion (https://github.com/iTwin/iTwinUI/pull/1040#discussion_r1092530211), but I think it needs some editing. Will look into it more.

rohan-kadkol: Would this be alright? This keeps the html demos the same, is not a breaking change, and also targets svgs with `.iui-icon` which is likely how icons are passed on the react side.

```suggestion
  > .iui-icon, > svg {
```

packages/itwinui-react/src/core/Avatar/Avatar.tsx
mayank99: ah nvm, it's not included twice.

still, the abbr should not be needed when there is an image. an empty abbr looks wrong to me.

i think you just want the background color, so maybe make it a div when there is an image

packages/itwinui-css/src/avatar/avatar.scss
mayank99: looks like you need to add these three lines to fix the positioning

```
left: 50%;
top: 50%;
```

and then the wrapping section would need `display: flex; align-items: end`

alternatively, `display: grid` could be used instead of `position: absolute` to center the svg. I think this would avoid the need to set styles on the wrapper.

i do not think it would be a breaking change because anyone using `iui-icon` will be unaffected.

rohan-kadkol: Thanks, the first option you mentioned worked. Now targeting the avatar's `svg` instead of `> .iui-icon` ([`7a2f509` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/7a2f509a9d896b0b183b8479771fc19e010b0ba2))

.changeset/chatty-lobsters-drop.md
```diff
@@ -0,0 +1,9 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`<Avatar>`'s `image` now supports passing `<svg>` too
+
+```tsx
+<Avatar child={<SvgUser />} />
```
mayank99, line 8:
```suggestion
<Avatar image={<SvgUser />} />
```

packages/itwinui-react/src/core/Avatar/Avatar.tsx
mayank99: I've also noticed that the default background is `white` is for some reason. This will not work well because we are also setting the svg fill to `white`.

We might want to change default background. @FlyersPh9

rohan-kadkol: > still, the abbr should not be needed when there is an image. an empty abbr looks wrong to me.
> 
> i think you just want the background color, so maybe make it a div when there is an image

To confirm, were you meaning something like [`966544a` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/966544ab0346f92214f0ab11f84481eb03e5badb)?

I thought of adding a proper background like `.iui-avatar-background` instead of re-using `.iui-initials` on the `<div>` background. But I thought that might be a bigger than needed change and so went with the reuse `.iui-initials` approach.

mayank99: yup, that looks fine, although i would have done it like this for better readability:
```jsx
{image ? (
  <div className='iui-initials' style={{ backgroundColor }} />
) : (
  <abbr className='iui-initials' style={{ backgroundColor }}>
    {abbreviation?.substring(0, 2)}
  </abbr>
)}
```

alternatively we could look into moving the backgroundColor to the parent element to avoid the need for this extra div.

rohan-kadkol: > yup, that looks fine, although i would have done it like this for better readability:

I first changed it to the code to the code you showed. [`8bff798` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/8bff798609c9de2dba9db457747a6d11a6127de9)

> alternatively we could look into moving the backgroundColor to the parent element to avoid the need for this extra div.

I then moved the background color to `.iui-avatar` and added a `border-radius: 50%` and also added the `.iui-initials` default background color (`background-color: hsl(72, 51%, 56%);`). [`38c3484` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/38c3484cdb4476abb4e5fb96cb152e50ff86557b)

mayank99: > > I've also noticed that the default background is `white` is for some reason. This will not work well because we are also setting the svg fill to `white`.
> 
> Now `.iui-avatar` has a default `background-color: hsl(72, 51%, 56%);` (olive green)

If you look at line 85, we are setting the default value of the inline style to `white`. This is what needs to change for better support of svgs (hence why i also tagged @FlyersPh9)

packages/itwinui-css/src/utils/icon/icon-component.scss
```diff
@@ -17,11 +17,13 @@ $icon-sizes: (
...
-  &,
+  width: var(--_iui-svg-size);
+  height: var(--_iui-svg-size);
+  fill: var(--_iui-svg-fill);
+
   svg {
-    width: var(--_iui-svg-size);
-    height: var(--_iui-svg-size);
-    fill: var(--_iui-svg-fill);
+    width: 100%;
```
mayank99, line 25: i think changes in icon can be undone

rohan-kadkol: Done

packages/itwinui-react/src/core/Avatar/Avatar.tsx
rohan-kadkol: > https://github.com/iTwin/iTwinUI/pull/1040#discussion_r1095782944: I then moved the background color to `.iui-avatar` and added a `border-radius: 50%` and also added the `.iui-initials` default background color (`background-color: hsl(72, 51%, 56%);`). [`38c3484` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/38c3484cdb4476abb4e5fb96cb152e50ff86557b). 

I think moving the background color from `.iui-initials` to `.iui-avatar` caused 42 test image changes. When checking on GitHub desktop, it shows no changes between the new and old images. But when checking on the GitHub website with the `GitHub PR images comparer settings` extension, it shows a few unrelated changes.

image

I'm not sure which is correct. I approved these images for now and didn't revert moving `background-color` to `.iui-avatar`

[`28e04c9` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/28e04c9451659abd5f7578fe0af43411865ddcd3)

mayank99 (approved):
Looks good. Just need @FlyersPh9's help to find a new default color (https://github.com/iTwin/iTwinUI/pull/1040#discussion_r1093813766)

packages/itwinui-react/src/core/Avatar/Avatar.tsx
FlyersPh9: Was white set as the default background color in React incase someone used a transparent PNG as their avatar photo?  We set a background color on `img` [here in CSS](https://github.com/iTwin/iTwinUI/blob/d6fe794d2e7d4ea7a698f9d7d1b4b82337037030/packages/itwinui-css/src/avatar/avatar.scss#L69), which should fix that issue for transparent PNGs.

I think it should be safe to change from white unless I am misunderstanding why it's being set.  I'd set it to `hsl(72, 51%, 56%)` to match [what we have in CSS](https://github.com/iTwin/iTwinUI/blob/d6fe794d2e7d4ea7a698f9d7d1b4b82337037030/packages/itwinui-css/src/avatar/avatar.scss#L44).

mayank99: Lets unset it in React then. CSS will handle both cases (img and non-img).

FlyersPh9: Agree.

FlyersPh9 (approved):
Good to go once [this change](https://github.com/iTwin/iTwinUI/pull/1040#discussion_r1097937478) is made.

packages/itwinui-react/src/core/Avatar/Avatar.tsx
rohan-kadkol: I removed the default `background-color: white` inline style in React. Now `background-color` inline style is only set when a `backgroundColor` is passed. [`ee84744` (#1040)](https://github.com/iTwin/iTwinUI/pull/1040/commits/ee847440335d571283a31c27909a15e5c1682ecf)

Just confirming before merging; this is what you meant, right? @mayank99 @FlyersPh9

PR #1042, opened 2023-02-01
fix(alert): fix alert button focus color

### LINKED ISSUE #859: Alert close button focus is not matching status color
Screenshot 2022-12-16 at 1 46 15 PM

The focus outline color should match the status color.

LostABike:
- added focus-visible within the iui-alert-category

closes #859


packages/itwinui-css/src/alert/alert.scss
```diff
@@ -72,7 +72,7 @@
     @include iui-anchor-status($statusColor);
   }
 
-  #{$rootSelector}-button {
+  #{$rootSelector}-button:focus-visible {
```
FlyersPh9, line 75: [Test comment] What is the specificity of this?

LostABike: it's [0, 3, 0]

mayank99, line 75: `:focus` will be better than `:focus-visible` because it will work with the fallback for older browsers. (The fallback code is already set inside `iui-alert-button` through `@include iui-button`).

```suggestion
  #{$rootSelector}-button:focus {
```

For more context: this works because `:focus-visible` always implies `:focus` (but not vice versa). And it doesn't have an adverse effect because setting only the color on `:focus` has no effect unless the whole `outline` is also set.

LostABike: That makes sense. I have updated and tested the change and confirm that it still behaves as expected. Thank you.

mayank99 (dismissed):

mayank99:
LGTM 👍

mayank99:
Oh actually changeset is missing, you can add one using `yarn changeset` (PR description should have reminded you, please don't delete the PR description next time).

.changeset/brown-elephants-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
```
mayank99, line 2: should be patch, not major. see https://semver.org for more details

```suggestion
'@itwin/itwinui-css': patch
```

.changeset/brown-elephants-tan.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+added focus to alert button outline for correct specificity order
```
mayank99, line 5: this should describe how it affects the user, rather than implementation details

```suggestion
Alert: close button's focus color now matches the status color.
```

mayank99 (approved):
LGTM 👍

PR #1044, opened 2023-02-02
feat(Surface, Divider): Add surface component with included layout & new divider util

### LINKED ISSUE #798: Make surface component easier to use
Surface component gives the background, shadow, etc., but it should be easier to achieve the common surface appearance seen in apps like Synchro (title section, content, dividing line, padding, etc).
Screenshot 2022-11-07 at 4 15 45 PM

FlyersPh9:
## Changes

- Add surface with layout component:
	- Supports a header & body.
	- Header & body can be padded.
	- Body can be scrollable if height is set on parent.
	- Header & empty state could utilize #1011.
- Add divider util.

Reviving #841, closes #798.  Requires react portion.

## Testing

- BackstopJS

## Docs

N/A


packages/itwinui-css/backstop/tests/surface.html
```diff
@@ -27,16 +29,17 @@
...
-      }
-      .demo-surface-content {
-        padding: var(--iui-size-s);
+      #demo-default {
+        display: grid;
+        grid-template-columns: repeat(
+          auto-fill,
+          minmax(Min(100%, 512px), 1fr)
+        );
+        gap: var(--iui-size-l);
```
FlyersPh9, line 38: Copied from https://github.com/iTwin/iTwinUI-layouts/blob/main/packages/layouts-css/src/fluid-grid.scss

packages/itwinui-css/backstop/tests/surface.html
```diff
@@ -27,16 +29,17 @@
...
+
+        --iui-surface-content-height: 246px;
```
FlyersPh9, line 40: Setting the height of all surfaces nested within.  Try removing this to see all surfaces stretch to fit all content rather than using overflow / scroll.

packages/itwinui-css/backstop/tests/surface.html
```diff
@@ -52,6 +55,35 @@
...
+        content: '';
+        position: absolute;
+        inset: 0;
+      }
+      .demo-mayanks-flex-component {
+        display: flex;
+        flex-grow: 1;
+        justify-content: space-between;
+      }
+      .demo-mayanks-flex-component-2 {
```
FlyersPh9, line 82: @mayank99 I'd like your input on line 82 & 77 being used in here.  I'm suggesting that we do not build this styling into the component and instead expect the user to piece together with your flex component.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,63 @@
...
 // Public facing surface component.
 @mixin iui-surface-component {
+  --iui-surface-padding-inline: var(--iui-size-s);
+  --iui-surface-content-padding: var(--iui-surface-padding-inline);
+
   @include iui-surface;
+  display: flex;
+  flex-direction: column;
+  height: var(--iui-surface-content-height);
+  width: 100%; // Copying line 16 of `fluid-grid.scss` in iTwinUI-layouts, unsure if needed
```
FlyersPh9, line 32: Copied from https://github.com/iTwin/iTwinUI-layouts/blob/10c77cf575d8431d1e4794ba1d48b8aaaa16f634/packages/layouts-css/src/fluid-grid.scss#L16, unsure if needed.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,63 @@
...
+
+@mixin iui-surface-content {
+  display: flex;
+  flex-grow: 1;
+  padding: var(--iui-surface-content-padding);
+  overflow-y: auto;
+  overflow-y: overlay;
+}
+
+@mixin iui-surface-content-empty {
```
FlyersPh9, line 81: This might not be needed at all, might be able to use just #1011 instead.

packages/itwinui-css/backstop/tests/surface.html
mayank99: yup, this component should be as minimal as possible.

packages/itwinui-css/src/surface/surface.scss
mayank99: then remove it 😅

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,63 @@
...
+
+@mixin iui-surface-footer-modifiers {
+  .iui-button {
+    flex-grow: 1;
+    border-radius: 0;
+    @media (forced-colors: active) {
+      border: none;
+    }
+
+    + .iui-button {
```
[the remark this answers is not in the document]
mayank99: this whole mixin is too specific. Have you seen any valid cases for two buttons in the footer area?

packages/itwinui-css/src/surface/surface.scss
mayank99: Does not belong in this component. fluid-grid will set it if needed

packages/itwinui-css/src/surface/classes.scss
```diff
@@ -5,3 +5,20 @@
 .iui-surface {
   @include iui-surface-component;
 }
+
+.iui-surface-header {
+  @include iui-surface-header-footer(header);
```
mayank99, line 10: Might be better to create intermediate mixins so the logic stays out of `classes`.

like this:
```scss
@mixin iui-surface-header {
  @include iui-surface-header-footer(header);
}
```
```scss
@mixin iui-surface-footer {
  @include iui-surface-header-footer(footer);
  @include iui-surface-footer-modifiers;
}
```

mayank99 (commented):
PR name looks wrong (Surface component already exists). Also changeset is missing.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Sounds good, removed.

packages/itwinui-css/src/surface/surface.scss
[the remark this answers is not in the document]
FlyersPh9: I have only seen 1 button so far, I'll remove all of the support for multiple buttons.

Will we expect the user to apply:
- `.iui-button { flex-grow: 1; }`: this one's important, but it's a one liner and I think we can expect devs to do this.
- `border-radius`: visually it's nice if the top corners have 0 & bottom corners match the surface. It isn't a huge problem unless the user changes the default surface border-radius.
- `@media (forced-colors: active) { border: none; }`: prevents a double thick border (1 from the bottom, 1 from the surface).  Forced-colors users are probably slim to none, so we can probably ignore this.

With all that said, I'm thinking remove the whole mixin and the user needs to apply `flex-grow: 1`.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Removed and using a `.demo-flex`.

packages/itwinui-css/src/surface/classes.scss
FlyersPh9: I've done this.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
 
 // Public facing surface component.
 @mixin iui-surface-component {
+  --iui-surface-padding-inline: var(--iui-size-s);
+  --iui-surface-body-padding: var(--iui-surface-padding-inline);
```
mayank99, line 26: Do these need to be configurable? I don't expect people changing this except to remove the body padding entirely.
```jsx
<Surface>
  <Surface.Body padded>...</Surface>
</Surface>
```
```jsx
<Surface>
  <Surface.Body>...</Surface>
</Surface>
```

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
...
+  @if ($type == 'footer') {
+    border-top: 1px solid var(--iui-surface-border-color);
+  }
+}
+
+@mixin iui-surface-header {
+  @include iui-surface-header-footer(header);
+}
+
+@mixin iui-surface-footer {
```
mayank99, line 54: I'm wondering if this footer even needs to exist. It might make more sense to have a `<Surface.Divider>` so that users can do whatever they want within each section.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
...
+  @include iui-surface-header-footer(footer);
+}
+
+@mixin iui-surface-body {
+  display: flex;
+  flex-grow: 1;
+  padding: var(--iui-surface-body-padding);
+  overflow-y: auto;
+  overflow-y: overlay;
```
mayank99, line 63: Might want to swap out `overflow: overlay` with the newer [`scrollbar-gutter: stable`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) which works across all browsers.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
 
 // Public facing surface component.
 @mixin iui-surface-component {
+  --iui-surface-padding-inline: var(--iui-size-s);
+  --iui-surface-body-padding: var(--iui-surface-padding-inline);
+
   @include iui-surface;
+  display: flex;
+  flex-direction: column;
+  height: var(--iui-surface-height);
```
mayank99, line 31: Unless we provide some preset values, this variable doesn't help, because the user can always specify `height` in their own CSS using a custom class.

mayank99 (commented):
This CSS PR will likely need a companion React PR before it can merge.

packages/itwinui-css/src/surface/surface.scss
mayank99: `stable both-edges` would also be a very nice touch for the padded variant.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
 
 // Public facing surface component.
 @mixin iui-surface-component {
+  --iui-surface-padding-inline: var(--iui-size-s);
+  --iui-surface-body-padding: var(--iui-surface-padding-inline);
+
   @include iui-surface;
+  display: flex;
+  flex-direction: column;
```
mayank99, line 30: We cannot make this component flex or grid by default, because it could break apps that already use `<Surface>`. So it needs to go in a modifier class or attribute.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: I've changed `--iui-surface-padding-inline` to `--_iui-surface-padding-inline` & body padding is now set via `data-iui-padded="true"` on `.iui-surface`.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: I was envisioning something like this:
```html
<wrapper style="--iui-surface-height: 100px;">
	<surface />
	<surface />
	<surface />
</wrapper>
```

I think you're proposing this:
```css
.myClass { height: 100px; }
```
```html
<wrapper>
	<surface class="myClass" />
	<surface class="myClass" />
	<surface class="myClass" />
</wrapper>
```
Do you still prefer your suggestion?

packages/itwinui-css/src/surface/surface.scss
mayank99: It doesn't need to be that complicated. Unless there is a strong reason for it, I think we can get rid of both of these CSS variables entirely. If you like the readability benefit, then using Sass variables is perfectly fine (it will get compiled out).

packages/itwinui-css/src/surface/surface.scss
mayank99: yes, because I see no benefit of the CSS variable in this case.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Thanks for sharing this, I love learning about new CSS props like `scrollbar-gutter`!

Right now [our scrollbar width is set to `var(--iui-size-xs)`](https://github.com/iTwin/iTwinUI/blob/a0693430dba7c9e9e52f2b0ca252bdee43fdfbff/packages/itwinui-css/src/utils/mixins.scss#L41) but we've been using `var(--iui-size-s)` for our padded areas.

What I'd like to propose: 1 PR that applies `scrollbar-gutter` everywhere applicable & change the scrollbar to [this old branch](https://github.com/iTwin/iTwinUI/blob/829e18ccb83a04ac799da3040ff7c834c42aa5f6/packages/itwinui-css/src/utils/mixins.scss#L36) that indents the scrollbar from the edge of the area. This would result in the same padding being applied currently. (Image below from a concept image on old documentation site). 
image

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Now using a Sass variable for padding.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Removed the footer & using `:not(:last-child)` to apply a bottom border on `.iui-surface-body`.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Removed the height, am setting the height using `grid-auto-rows` in the demo CSS.

packages/itwinui-css/src/surface/surface.scss
mayank99: `:last-child` won't be reliable, as the user might want to place something else at the end. I really think a `Surface.Divider` component is the most flexible way to do this.

You might even be able to remove the `Surface.Header` if you do this (not tested).
```tsx
<Surface>
  <Surface.Content padded>...header stuff...</Surface.Content>
  <Surface.Divider />
  <Surface.Content padded>...header stuff...</Surface.Content>
  <Surface.Divider />
  <Surface.Content>...footer stuff...</Surface.Content>
</Surface>
```

packages/itwinui-css/src/surface/surface.scss
mayank99: the padding is currently messed up in the changes you pushed.

what i'm suggesting: use `stable both-edges` *along with* a smaller padding.

FlyersPh9: I believe this is fixed now.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: I removed the bottom border and introduced a new `.iui-divider` component.  I did not get rid of `.iui-surface-header` & `.iui-surface-body` for a generic content area because I _think_ the 2 require different CSS:
- Body needs to support overflow.
- Body needs to have `flex-grow: 1`.

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -22,5 +22,43 @@
...
+}
+
+@mixin iui-surface-footer {
+  @include iui-surface-header-footer(footer);
+}
+
+@mixin iui-surface-body {
+  display: flex;
+  flex-grow: 1;
+  padding: var(--iui-surface-body-padding);
```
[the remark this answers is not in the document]
LostABike: Because this review is supposed to be from last week, I think it is no longer applicable now. Resolving this.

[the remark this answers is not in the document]
FlyersPh9: Regarding [the horizontal overflow comment](https://github.com/iTwin/iTwinUI/pull/1044#discussion_r1103251005), I can fix that by adding this:
```diff
@mixin iui-surface-header {
  …

+  > * {
+    overflow: hidden;
+    overflow-wrap: break-word;
+  }
}

@mixin iui-surface-body {
  …
+  overflow-x: hidden;
+
+  > * {
+    overflow: hidden;
+    overflow-wrap: break-word;
+  }

  …
}
```
Screenshot 2023-02-22 at 4 25 47 PM

`overflow-x: hidden;` on the body definitely should be added.  It's the `> *` parts that I am unsure about.  Should that fall on the user?  If it doesn't fall on the user, should the styling be applied via surface CSS or via @mayank99's flex component #1011?

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -18,9 +18,36 @@
...
+
+@mixin iui-surface-header {
+  display: flex;
+  flex-shrink: 0;
+  align-items: center;
+  overflow: hidden;
+  border-bottom: 1px solid var(--iui-surface-border-color);
+  padding: $surface-padding;
+  scrollbar-gutter: stable both-edges;
 }
```
LostABike, line 39: I think everything looks good so I just have question for learning purpose when playing around in this branch :sweat_smile:

I tried to remove the paddings for this gear button because it is not balanced with the left side by adding
``
@mixin iui-button {
  padding: 4px 0;
}
``
here.

But nothing changed. How could I have achieved this?

(also image if you want to see what I'm talking about)
image

packages/itwinui-css/src/surface/surface.scss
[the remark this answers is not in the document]
FlyersPh9: Regarding [the padding comment](https://github.com/iTwin/iTwinUI/pull/1044#discussion_r1103248549), if I understand what you are asking correctly, you're unsure why the padding in red isn't visually the same as the green or the blue.
218197710-765c0f28-c689-413c-9b93-4a35841dfa81

The green side appears to have more padding, but that is only because the text is wrapping unevenly where there's a space in a sentence.  If you right click & inspect the element we can see the padding is equal on both sides despite where the sentence wraps (padding shown in light green).
Screenshot 2023-02-22 at 4 33 45 PM Screenshot 2023-02-22 at 4 34 43 PM

The blue sides, the padding is set to be half of that on the left & right, so it will be smaller.  The padding is only applied at the top and bottom of the element, even if overflowing is present.  So if you scroll to the center, you may experience text touching the very top / bottom borders.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: > I just have question for learning purpose when playing around in this branch 😅

Awesome, that's the best way to learn!

> I tried to remove the paddings for this gear button

If you wanted to make the button thinner but only when placed within the surface header, you'd need to also unset `min-width` which is currently set to be the height of the button (this makes the button 1/1 ratio at it's smallest size) :

```scss
@mixin iui-surface-header {
  … 

  .iui-button {
    padding: 0;
    min-width: auto;
  }
}
```

However, I would suggest against that.  All of our buttons, even [the borderless button](https://itwin.github.io/iTwinUI/css/button.html#demo-borderless), have space around the text / icon to increase the [touch target size](https://www.nngroup.com/articles/touch-target-size/).  A 36x36 px button is easier to tap with a finger than a 16x16 px icon.  [Accessibility best practices](https://www.w3.org/WAI/WCAG21/quickref/#target-size) suggest larger than our default 36px, it's something that we need to look into in the future.

Now, an alternative solution to visually align the icon without reducing the button size could be to set the header's left & right padding separately:
```
padding-left: 12px;
padding-right: 4px;
```
But we're not always 100% sure that the user might place a button in the right hand side.  They might place a checkbox with a label, or a badge:
Screenshot 2023-02-22 at 4 55 21 PM Screenshot 2023-02-22 at 4 55 47 PM

To avoid things getting too complex, we simply keep the padding equal on the left & right.

packages/itwinui-css/src/surface/surface.scss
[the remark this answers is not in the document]
LostABike: Ah I see. Thanks for explaining!

packages/itwinui-css/src/surface/surface.scss
LostABike: Awesome. I just tested your suggestions and they work for what I had in mind. But I also understand we should keep things generic for users to meddle with. Thanks for explaining 😄

LostABike (approved):

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -18,9 +18,37 @@
...
 
-// Public facing surface component.
-@mixin iui-surface-component {
-  @include iui-surface;
+@mixin iui-surface-body {
+  display: flex;
```
mayank99, line 41: does this need to be flex?

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -18,9 +18,37 @@
...
+  }
+}
+
+$surface-padding: var(--iui-size-2xs);
+
+@mixin iui-surface-header {
+  display: flex;
+  flex-shrink: 0;
+  align-items: center;
+  overflow: hidden;
```
mayank99, line 34: does this need overflow hidden?

packages/itwinui-css/backstop/tests/divider.html
```diff
@@ -0,0 +1,65 @@
...
+  </head>
+  <body class="iui-root">
+    <theme-button></theme-button>
+    <h1>Divider</h1>
+    <hr />
+
+    <section id="demo-default">
+
+      <div id="demo-horizontal">
+        <hr class="iui-divider" />
```
mayank99, line 53: I think Divider doesn't need its own html page. We don't have a page for Flex.

We can still show it in storybook. https://itwin.github.io/iTwinUI/react/?path=/story/utilities-flex--basic

packages/itwinui-css/src/divider/divider.scss
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+@mixin iui-divider {
+  background-color: var(--iui-color-border-subtle);
+  flex-shrink: 0;
```
mayank99, line 6: `flex: none` might be better, since we don't want it grow or shrink

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: It is not, removed.

packages/itwinui-css/backstop/tests/divider.html
FlyersPh9: Ok, I've removed the demo page, scenario, and test images for divider.

packages/itwinui-css/src/divider/divider.scss
FlyersPh9: Ooh, good idea, changed!

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Oh wait, it might be needed.  I was using this to center the empty state.  Here's how it looks after removing `display: flex`:

Screenshot 2023-02-23 at 11 08 33 AM

mayank99: well, centering the content should be left up to the user

packages/itwinui-css/src/divider/divider.scss
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+@mixin iui-divider {
```
mayank99, line 4: all divider styles should go under the `utils` folder, next to `flex`

FlyersPh9: Moved to `utils` folder.

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: What about leaving `display: flex` or adding `position: relative` so the user can easily achieve vertical centering?

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: It looks like removing `overflow: hidden` nullifies `scrollbar-gutter: stable both-edges`.  We also risk text leaking out of the container.
Screenshot 2023-02-23 at 11 22 38 AM

cortneysmithbentley (approved):

packages/itwinui-css/src/surface/surface.scss
mayank99: my concern is that `display: flex` can be pretty invasive, and could unnecessarily break styles. same with `positive: relative`, it can create an undesirable containing block.

plus, the user can always easily add custom styles and we even give them a `<Flex>` utility, so i dont think we should do it for them.

packages/itwinui-css/src/surface/surface.scss
mayank99: ok, lets keep it then

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: Ok, I'll leave `display: flex` out of the CSS & will add an inline style to apply `display: flex` just to keep that 1 empty state demo working correctly.

mayank99 (approved):
Looks good! Now onto the react part

apps/storybook/src/Surface.stories.tsx
```diff
@@ -39,3 +50,180 @@ export const Basic: Story<SurfaceProps> = ({ elevation }) => {
 };
 
 Basic.args = { elevation: 1 };
+
+export const Custom: Story<SurfaceProps> = (args) => {
```
adamhock, line 54: I made a story for each of the cases shown in the surface demo pages. Let me know if any of them can be removed. I'm also open to suggestions about these story names

apps/storybook/src/Surface.stories.tsx
```diff
@@ -39,3 +50,180 @@ export const Basic: Story<SurfaceProps> = ({ elevation }) => {
...
+    maxHeight: '300px',
+  };
+  return (
+    <Surface
+      elevation={elevation}
+      style={cardStyle}
+      layout={layout}
+      padded={padded}
+    >
+      <style>
```
adamhock, line 154: I've seen styling added this way in the `CenterContent` story in `Header` [here](https://github.com/iTwin/iTwinUI/blob/main/apps/storybook/src/Header.stories.tsx#L279)

packages/itwinui-react/src/core/Surface/Surface.tsx
```diff
@@ -29,7 +29,47 @@ const getSurfaceElevationValue = (elevation: SurfaceProps['elevation']) => {
...
+      <div className={cx('iui-surface-body', className)} ref={ref} {...rest}>
+        {children}
+      </div>
+    );
+  },
+);
+
 export type SurfaceProps = {
+  /**
+   * TODO: Not sure how to describe this prop
```
adamhock, line 66: I'm still not sure how to describe this prop. I'm open to suggestions

packages/itwinui-css/src/surface/surface.scss
```diff
@@ -18,9 +18,36 @@
...
 
-// Public facing surface component.
-@mixin iui-surface-component {
-  @include iui-surface;
+@mixin iui-surface-body {
+  flex-grow: 1;
+  overflow-x: hidden;
+  overflow-y: auto;
+
+  [data-iui-padded='false'] & {
```
adamhock, line 45: I noticed in the css that `data-iui-padded` only affects `iui-surface-body`, but the demo pages have this data attribute applied to `iui-surface`. Should the data attribute just apply to `iui-surface-body`?

packages/itwinui-react/src/core/utils/components/Divider.tsx
```diff
@@ -0,0 +1,25 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import { useTheme } from '../hooks';
+import '@itwin/itwinui-css/css/utils.css';
+
+export type DividerProps = React.ComponentPropsWithRef<'hr'>;
+
+/**
+ * Shows a divider
```
adamhock, line 13: If anyone has a better way to describe Divider, I'm all ears

adamhock (commented):
React portion of the surface component is ready for review!

packages/itwinui-react/src/core/utils/components/Divider.tsx
gretanausedaite: "element that helps to separate pieces of content into clear sections or parts."

packages/itwinui-react/src/core/Surface/Surface.test.tsx
```diff
@@ -43,3 +43,26 @@ it('should add className and style correctly', () => {
   expect(surface).toBeTruthy();
   expect(surface.style.color).toBe('grey');
 });
+
+it('should render custom surface', () => {
```
gretanausedaite, line 47: Also test data-iui-[**] props you've added.

packages/itwinui-react/src/core/Surface/Surface.tsx
```diff
@@ -29,7 +29,47 @@ const getSurfaceElevationValue = (elevation: SurfaceProps['elevation']) => {
...
+   */
+  layout?: boolean;
```
gretanausedaite, line 68: maybe `useLayout` and `isPadded` are better?
For layout prop you could write "Aligns children vertically when set to true" or smth. Simply describe what it does.

mayank99:
@adamhock Could you please move your changes to a new branch (based on top of this branch) so it can be reviewed separately?

packages/itwinui-css/src/surface/surface.scss
mayank99: @FlyersPh9 Maybe move this attribute to the body? i.e. `&[data-iui-padded='false']`

FlyersPh9: Done, I just pushed this change.

PR #1045, opened 2023-02-02
use node 18

mayank99:
## Changes

Node 16 is no longer the active LTS version. When you go to https://nodejs.org, it prompts you to install Node 18, not 16.

Storybook does not work out of the box with Node 18 because it uses some old version of webpack internally. See issues: https://github.com/storybookjs/storybook/issues/20482 and https://github.com/storybookjs/storybook/issues/20209

This is possible to work around by using `NODE_OPTIONS=--openssl-legacy-provider`.

It's a bit silly that the latest stable version of storybook does not work with the latest stable version of node. Apparently this will be fixed in storybook 7, which won't be released until March this year. This workaround will avoid issues with new contributors in the meantime.

Also, node 16 users are unaffected.

## Testing

Tested locally by switching to node 18.

I've also updated our CI to use node 18 instead of node 16 and tested that CI [fails](https://github.com/iTwin/iTwinUI/actions/runs/4077564555/jobs/7026750726) when not setting NODE_OPTIONS and [passes](https://github.com/iTwin/iTwinUI/actions/runs/4077566304/jobs/7026753079) when setting it.

## Docs

Mentioned Node "16+" (previously was "16") in contributing setup instructions.

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "storybook": "start-storybook -p 6006 --no-open",
-    "build": "dotenv -v NODE_OPTIONS=--openssl-legacy-provider -- build-storybook",
+    "build": "build-storybook",
     "dev": "yarn storybook --quiet",
     "lint": "eslint \"**/*.{ts,tsx}\" --max-warnings=0",
     "test": "node scripts/run-tests.js",
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -19,6 +19,7 @@
     "configs": "*",
     "cypress": "9.6.0",
     "cypress-image-diff-js": "1.18.1",
+    "dotenv-cli": "7.0.0",
     "eslint": "^8.14.0",
     "eslint-config-prettier": "^8.5.0",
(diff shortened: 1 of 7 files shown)


apps/storybook/package.json
```diff
@@ -32,8 +33,8 @@
   },
   "scripts": {
     "storybook": "start-storybook -p 6006 --no-open",
-    "build": "build-storybook",
-    "dev": "yarn storybook --quiet",
+    "build": "dotenv -v NODE_OPTIONS=--openssl-legacy-provider -- build-storybook",
```
aruniverse, line 36: i dont think you need dotenv to set this var, might be wrong tho

mayank99: It's needed to do it in a cross platform way. Without this it will fail on windows.

mayank99: We would, but it might break some of our processes because of incorrectly declared deps.

aruniverse: pnpm should have you covered there too! https://pnpm.io/package_json#pnpmpackageextensions

just an fyi

gretanausedaite (approved):
We should update our website build to not use 16.x.

PR #1046, opened 2023-02-02
Added min-height for draggable dialog

### LINKED ISSUE #965: Resizable dialog should have a `min-height` set
The resizable dialog can be resized to the point where the buttons creep outside of the body.  We should set a `min-height`.

Screen Recording 2023-01-06 at 4 39 49 PM

LostABike:
## Changes

Added min-height variable for draggable dialog, which is a parameter that the user can optionally supply if they want a different min-height. Initialized the variable globally.
Do we want a fixed min-height like this? Or it would be better to calculate off of where the buttons are relative to the dialog window (not sure if possible). Also I have tested all the pre-defined sizes in `sizes.scss` but they are all too small and still cause the buttons to overflow out of the dialog so I used a fixed number instead, would like some opinions on this.

Before:
image
After:
image

closes #965 

## Testing

Ran yarn test, no changes (no tests for this in particular)

## Docs

N/A


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -114,10 +115,11 @@ $iui-dialog-width: 380px;
   }
 }
 
-@mixin iui-dialog-draggable {
+@mixin iui-dialog-draggable($min-height: $iui-dialogue-resizable-min-height) {
```
mayank99, line 118: We only deliver the compiled CSS to our users, not the Sass mixins, so this parameter does nothing.

It can be a bit hard to follow inside the repo, so I suggest browsing the package in isolation here: https://unpkg.com/@itwin/itwinui-css/ This will give you an idea of what is "internal" vs "public".

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -3,6 +3,7 @@
 @import '../utils/index';
 
 $iui-dialog-width: 380px;
+$iui-dialogue-resizable-min-height: 130px;
```
mayank99, line 6: Value looks a bit random. How did you come up with `130px`?

LostABike: Just by testing around and see which ones look like the best min-height on my screen, which I have included a screenshot in the description. Also I thought the width right above looked random too so I thought that was also just from eye-gauging. This is something that I wanted input on per the PR description since hard coded value is always risky considering it doesn't take into account different screen, browsers or OS.

packages/itwinui-css/src/dialog/dialog.scss
LostABike: So is the only way to make it an optional variable that users can supply is through React?

elephantcatdog:
> Also I have tested all the pre-defined sizes in sizes.scss but they are all too small and still cause the buttons to overflow out of the dialog so I used a fixed number instead, would like some opinions on this.

Sometimes we multiply the pre-defined sizes by another number to get a less "random" number for something else. There's examples of this if you search-all in the code base by "calc(var(--iui-size", or if you look in the `sizes.scss` file you mentioned, on line 24-26:
image

You could do `calc(var(--iui-size-2xl) * 2)` to get ~128px.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: Through React props (for react users) or through CSS variables (for CSS users) or a combination of both in many cases. You can see an example in Badge:
https://github.com/iTwin/iTwinUI/blob/fd2e52397764d16ddd3132d9958817517016e086/packages/itwinui-css/src/badge/badge.scss#L21
https://github.com/iTwin/iTwinUI/blob/fd2e52397764d16ddd3132d9958817517016e086/packages/itwinui-react/src/core/Badge/Badge.tsx#L74

But for dialog min-height, I think there is no need to make it configurable. It should just be a sensible default value. Users always have the ability to override using their own custom CSS.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: Let's look at it mathematically then. If we add up all the individual parts that make up the UI, then get somewhere around ~124px, which can be converted to rem to get 7.75rem.


```suggestion
$iui-dialogue-resizable-min-height: 7.75rem;
```

It would be a bit verbose to do all those calculations in code so you could just add a comment.

```suggestion
// 7.75rem = 124px = title bar height + margins + button height + partially visible text
$iui-dialogue-resizable-min-height: 7.75rem;
```

LostABike: That makes sense. My question would be why do we want to have the leftover part be 16px? Is it universal rule to have that be less than line-height? Or a range of line-height > x > 0 where x is leftover would be adequate?

mayank99: We want it to be less than the line-height so that the text is partially visible, which serves an additional indicator that the region in scrollable. We use a similar approach in menu-items

.changeset/six-radios-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Draggable dialog now have a min height set so that the dialog can't be minimized more than where the dialog buttons are. It can also be optionally customized using the `$min-height` variable
```
mayank99, line 5:
```suggestion
Resizable dialog now has a min height set so that it always includes the button bar.
```

mayank99 (approved):

FlyersPh9 (approved):
Looks good to me once you make [the change suggested by Mayank](https://github.com/iTwin/iTwinUI/pull/1046/files#r1096150094).

mayank99:
@LostABike Why was this merged before updating the changeset?

LostABike:
@mayank99 That was an oversight on my part. I have edited the branch and recreated a [PR](#1051) to update that.

Issue #1053, opened 2023-02-07
Table pager buttons cause page to refresh when inside form tag


mattl-bentley:
#### Describe the bug (actual behavior)
Clicking the table pager buttons causes the page to refresh when the table is inside a form tag.

#### Expected behavior
Clicking the page buttons changes pages without refreshing the page.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/paging-page-refresh-w0bpoh?file=/App.tsx:2606-2617

##### Steps to reproduce
1. When the page loads, click page "2" and notice the page refreshes instead of changing pages

#### Additional information
This issue exists in itwinui-react@2.4, but not in 1.x. Comparing the html output, 2.x does not include type="button" for the paging buttons.

![image](https://user-images.githubusercontent.com/110254711/217288089-e605f5ea-76c9-46a4-8c53-c9ca85faf36f.png)

![image](https://user-images.githubusercontent.com/110254711/217287996-abb3e802-07e5-4b6f-b2e1-692980dde2ee.png)

mayank99:
Hey @mattl-bentley. That makes sense the button needs `type='button'`, although I would question putting the whole table in a form; it can cause issues with some screen readers)

Do you want to contribute by creating a pull request? The change would go here: https://github.com/iTwin/iTwinUI/blob/4e43e5ca488d998bd7920c9b30fa85e82d65eeb5/packages/itwinui-react/src/core/Table/TablePaginator.tsx#L173-L175

mayank99:
this is now released in [2.4.1](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%402.4.1) 🎉

mattl-bentley:
Great, thank you!

PR #1056, opened 2023-02-07
Specify type='button' on paginator buttons

mattl-bentley:
## Changes

Specify type='button' to prevent postbacks when pager is used in forms.
Fixes issue #1053 

## Testing

Copy/pasted sample from https://codesandbox.io/s/paging-page-refresh-w0bpoh?file=/App.tsx in vite playground to verify pager buttons no longer postback.

## Docs

N/A


mayank99 (approved):
Thanks for the contribution, looks good!

PR #1058, opened 2023-02-07
fix(Table): Expandable row content displays when table has a fixed height

### LINKED ISSUE #473: Last row in table when expanded then need to be scrolled

FlyersPh9:
## Changes

If a height is set on the table expandable rows do not display.  Adding `flex-shrink: 0;` fixes this.

**Before:**
table-before

**After:**
table-after

If `.iui-scroll-snapping` is applied on `.iui-table-body` it will auto-scroll to display the bottom expanded row content.
table-scroll

Closes #473.

## Testing

How I tested this:
1. Go to [the expandable table Storybook demo](https://itwin.github.io/iTwinUI/react/iframe.html?args=&id=core-table--expandable&viewMode=story).
2. Set `height: 200px;` on `.iui-table`.
3. Try to expand a row and it won't display.

I think the user was asking for a solution to the 3rd GIF in the description which we already offered a solution to.  While looking into the original issue, I discovered this bug.

## Docs

N/A


packages/itwinui-css/src/table/table.scss
```diff
@@ -201,6 +201,7 @@
       border-left: 1px solid transparent;
       border-right: 1px solid transparent;
       border-bottom: 1px solid var(--iui-color-border);
+      flex-shrink: 0;
```
mayank99, line 204: How did you test these changes? How do I reproduce your gifs? (please dont clear the PR description template)

Why does it work and will it have any adverse affect in edge cases like virtualization?

FlyersPh9: How I tested this:
1. Go to [the expandable table Storybook demo](https://itwin.github.io/iTwinUI/react/iframe.html?args=&id=core-table--expandable&viewMode=story).
2. Set `height: 200px;` on `.iui-table`.
3. Try to expand a row and it won't display.

I think the user was asking for a solution to the 3rd GIF in the description which we already offered a solution to.  While looking into the original issue, I discovered this bug.

mayank99: Dont forget changeset.

mayank99 (approved):

LostABike (approved):
I tested this change too. Looks good to me.

packages/itwinui-css/src/table/table.scss
FlyersPh9: Changeset added.

PR #1060, opened 2023-02-08
feat(Tabs): Added `overflowButton` property

### LINKED ISSUE #921: HorizontalTabs: overflow handling
Maybe it is valid use case. We could just use `useOverflow` hook as in breadcrumbs, to have something like this
image

elephantcatdog:
## Changes

image
image
image
image

## Testing
- Added visual test for storybook
- Added unit tests
- Manually tested:
  - Made sure that clicking on tabs switches tabs opens the corresponding content

## Docs
- Added horixzontal example to website
- Added horizontal and vertical stories to storybook
- Added changeset

-----------------------------------------------------------Previous Version------------------------------------------------------------------
## Changes
- Added class `iui-overflow` to add styling for overflow tabs
  - `overflow: auto` prevents tabs from overflowing the tabs wrapper & allows the function `useOverflow` to work correctly
    - * Could be more specific - `overflow: none`?
  - `white-space: nowrap` prevents the text in tabs from wrapping (this looks better than allowing it to wrap when using the overflow button)
- Added property `overflowButton` that enables tabs to be truncated and an overflow button to be shown
  - expects a function that takes the number of items that are visible and returns the 'ReactNode' to render
- `onTabClick` was made into a callback because it's used as a parameter for the memo `displayTabs`
- Added function `displayTabs` to only show visible tabs when the overflow button is passed in; without an overflow button tabs are displayed as normal

image
image

## Testing
- Added visual test for storybook
- Added unit tests
- Manually tested:
  - Made sure that clicking on tabs switches tabs and opens the corresponding content
  - Made sure that clicking the button opens the dropdown menu
  - Made sure that clicking an option in the dropdown menu opens the corresponding content and shows that no tabs are selected/open
  - Made sure that resizing works correctly, both visually and in the dropdown menu content

## Docs
- Added example to website
- Added story to storybook
- Added changeset

closes #921

## To Do:
[X] Fix test images
[X] Fix keyboard accessibility
[X] Test with disabled tabs
[X] Fix breadcrumbs tests ?????
[ ] Add tests for children observer
[ ] Maybe change keyboard accessibility


elephantcatdog:
What happens if it's used with vertical tabs?

apps/website/src/examples/Tabs.overflow.tsx
```diff
@@ -0,0 +1,95 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Tabs, Tab, DropdownMenu, MenuItem, IconButton } from '@itwin/itwinui-react';
+import SvgMoreSmall from '@itwin/itwinui-icons-react/cjs/icons/MoreSmall';
```
FlyersPh9, line 7: Can you share a screenshot of what it looks like with `SvgMore`, not the small variant?  I think it might look more proportional.

apps/website/src/examples/Tabs.overflow.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  return (
+                    <MenuItem key={index} onClick={onClick}>
+                      Item {index}
+                    </MenuItem>
+                  );
+                })
+            }
+          >
+            <IconButton
+              style={{ paddingTop: '12px', margin: '4px', height: 'auto' }}
```
FlyersPh9, line 81: `paddingTop` can be removed, we'll want the icon vertically aligned, it doesn't need to align with the bottom of the text.

FlyersPh9, line 81: I think you're using `margin` here to vertically center the button and give it some gap between the last tab and the overflow button.

I did not test this, but you might be able to do something like this instead:
```scss
.iui-overflow li:last-of-type {
  align-self: center;
  margin-left: 4px;
}
```

Although that might negatively impact a tablist with that class that is not currently overflowing. 🤔

apps/website/src/examples/Tabs.overflow.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  };
+                  return (
+                    <MenuItem key={index} onClick={onClick}>
+                      Item {index}
+                    </MenuItem>
+                  );
+                })
+            }
+          >
+            <IconButton
```
FlyersPh9, line 80: All of the `.iui-tab` have a `role="tab"`.  I'm unsure if the overflow button needs a `role` set, but probably something we should investigate.

FlyersPh9:
In Storybook, when I select a tab from the overflow dropdown menu it should have a checkmark showing which menu item is active.
Screenshot 2023-02-08 at 11 57 19 AM

Another thought I'm just thinking outloud, unsure if this is a good idea or not.  From a UX perspective, if I select a tab from the overflow dropdown, I have no visual indicator what tab is active without opening the dropdown menu again.  What if an active tab from the overflow is displayed as the last visible tab?  Mockup below:
Screenshot 2023-02-08 at 12 00 16 PM

LostABike:
Seconding @FlyersPh9 on visibility of the currently selected tab, which would give users' visual feedback of what tab they are on.
Also it looks like for smaller screens like on phone, the entire tab seems to be minimized more than it should, at least compared to the default tab:
image
image

I'm just wondering if we want the tab behavior for smaller screens to be consistent across different tab implementations, or if the difference is expected behavior?

elephantcatdog:
> Also it looks like for smaller screens like on phone, the entire tab seems to be minimized more than it should, at least compared to the default tab:

It looks like this because I applied `maxWidth: 25%` to the surrounding div so that the example would be truncated. I didn't just set a width so that users can play with resizing the div.

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: I'm currently leaving it up to the user as to what the button looks like. They might want it touching the other tabs or be bigger or something.

elephantcatdog: But when I did this with Breadcrumbs we wanted it to align with the bottom of the text...

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: I did the Breadcrumbs version with SvgMore and I was asked to make it the small version.

Here is what it looks like with the big version, vertically centered:
image

elephantcatdog: Here's the Breadcrumbs version I did before: 
image
 I'm noticing now that it's not exactly lined up with the text, but I tried.

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: We do this in several places, where we leave the exact implementation of a particular thing up to the users (overflowButton in ButtonGroup, menuItems in Header, thumbProps in Slider). The alternative would be to enforce what the button/component looks like in this particular instance, but users might want to put something else here besides a dropdown & button.

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: Added

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -54,6 +55,76 @@ type TabsTypeProps =
...
+      type?: 'default' | 'borderless';
+    }
+  | {
+      orientation: 'vertical';
+      type?: 'default' | 'borderless';
+    }
+  | {
+      type: 'pill';
+      orientation?: 'horizontal';
+    };
```
elephantcatdog, line 126: This isn't quite right. I get errors on the 'PillTabs' and 'Vertical' stories:
image
Does anyone know what the issue is? @gretanausedaite @mayank99 @adamhock

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: No, the ellipsis are clickable. See https://github.com/iTwin/iTwinUI-react/pull/814
image
At one point the '...' icon was blue (so that it looked like a button), so I'm not sure why it's black now.

elephantcatdog: But, look, I'll change both to be SvgMore and centered. But dang, it's frustrating to get the exact opposite feedback on something so similar, haha

FlyersPh9: Oh wow, I was unaware.  The [breadcrumb HTML demos](https://itwin.github.io/iTwinUI/css/breadcrumbs.html) are using the ellipsis text, not an SVG.  If we're using an SVG in storybook, we should probably have a demo using the SVG in our HTML demos as well.

elephantcatdog: Update on the blue/not blue situation: It's blue in react v1, and black in react v2. Did all the button colors change or something?

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: Updated

apps/website/src/examples/Tabs.overflow.tsx
elephantcatdog: Updated

elephantcatdog:
@LostABike I updated the tabs (and breadcrumbs) overflow stories so that the area won't get too small or too big.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Do you get the same error in test app? I think this can be just in storybook (similar errors in combobox).

elephantcatdog:
> In Storybook, when I select a tab from the overflow dropdown menu it should have a checkmark showing which menu item is active.

> Another thought I'm just thinking outloud, unsure if this is a good idea or not. From a UX perspective, if I select a tab from the overflow dropdown, I have no visual indicator what tab is active without opening the dropdown menu again. What if an active tab from the overflow is displayed as the last visible tab? Mockup below:

@FlyersPh9 I added logic to make the last tab the visible tab. I also updated the dropdown menu so that it only shows items that aren't currently visible: 
image

LostABike:
I like how everything works and looks right now so I'm just manually stretch testing things. What is the expected behavior when the tab name is a little longer?
image
image
Maybe we want to truncate or phased out words to keep the same tab length? Like how Chrome has it here:
image

elephantcatdog:
> I like how everything works and looks right now so I'm just manually stretch testing things. What is the expected behavior when the tab name is a little longer?

@LostABike 
Did you add extra text from Chrome inspect? The useOverflow function can't work correctly in that case because it doesn't have all the information about the tabs (how long they are) before determining how many tabs can be visible, because you're changing that information after it's decided how many tabs can be shown. If the text is longer from the story it would look like this:
image
But it doesn't work super great when the last tab is replaced with something smaller:
image
or larger:
image
image

Hm, I'll look into fixing that issue. 

> Maybe we want to truncate or phased out words to keep the same tab length? Like how Chrome has it here:

We don't have any considerations for this on the regular tabs. I feel like we should do it for all types of tabs if we do it. I'm not opposed to the idea, but I'm not sure about it because I feel like that's one of the pros of using an overflow - you can see the entire label even when there's not room for all of them. @FlyersPh9 ?

FlyersPh9:
> > Maybe we want to truncate or phased out words to keep the same tab length? Like how Chrome has it here:
> 
> We don't have any considerations for this on the regular tabs. I feel like we should do it for all types of tabs if we do it. I'm not opposed to the idea, but I'm not sure about it because I feel like that's one of the pros of using an overflow - you can see the entire label even when there's not room for all of them. @FlyersPh9 ?

Let's release without that & see what kind of feedback we get.  Ideally tabs should never be that long & UX team should help guide devs to use shorter labels.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
elephantcatdog: I don't get the same errors, and yeah, I see the errors in ComboBox. Thanks Greta!

gretanausedaite (commented):
No additional comments.
Making users add margin to our IconButton looks a bit strange. Maybe we could consider adding dedicated Tabs.OverflowButton (or smth) in a future for default cases if simple IconButton does not work well.

apps/website/src/examples/Tabs.overflow.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <Tab key={7} label='Founders' />,
+    <Tab key={8} label='History' />,
+  ];
+
+  return (
+    <div style={{ maxWidth: 425, border: '1px solid lightpink', padding: 8 }}>
+      <Tabs
+        labels={labels}
+        overflowButton={(visibleCount: number) => (
+          <DropdownMenu
```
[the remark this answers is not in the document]
elephantcatdog: Updated

LostABike:
> 

Nice! I think everything works and looks good now 😸

LostABike (approved):

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -99,6 +99,14 @@ $borderless-horizontal-tab-min-height: calc(
     }
   }
 
+  &.iui-overflow {
```
FlyersPh9, line 102: Can we use a data attribute instead?  Next major release I'd like to replace most, if not all, class modifiers with data attributes.
```suggestion
  &[data-iui-overflow='true'] {
```

elephantcatdog:
Overflow tabs don't work with keyboard well.

elephantcatdog:
Need to test with disabled tabs.

apps/website/src/pages/docs/tabs.mdx
```diff
@@ -20,6 +20,16 @@ import * as AllExamples from '~/examples';
   <AllExamples.TabsMainExample client:load />
 </LiveExample>
 
+## Usage
+
+### Overflow
+
+If the available space does not allow all the tabs to be displayed, add an overflow button to allow access to all tabs.
+
+<LiveExample src='Tabs.exampleCodeOverflow.tsx'>
```
gretanausedaite, line 29:
```suggestion
<LiveExample src='Tabs.overflow.tsx'>
```

gretanausedaite (commented):
Why did breadcrumbs images change?

No comments on code. Looks good.

elephantcatdog:
> Why did breadcrumbs images change?

@gretanausedaite Jon requested that the '...' be the large version for tabs, so I went back to breadcrumbs and made it match. It had the small '...' before.

apps/website/src/pages/docs/tabs.mdx
elephantcatdog: Updated

packages/itwinui-css/src/tabs/tabs.scss
elephantcatdog: Updated

.changeset/lovely-poems-remember.md
```diff
@@ -0,0 +1,44 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': minor
+---
+
+- Added data attribute 'data-iui-overflow' - when true it adds styling for overflow tabs
+- Added property 'overflowButton' - enables tabs to be truncated and an overflow button to be shown; expects a function that takes the number of items that are visible and returns the 'ReactNode' to render
+
+```typescript
```
FlyersPh9, line 9: Did you mean to put this code in the changeset?  I haven't seen us do this before, should we?

apps/storybook/src/Breadcrumbs.stories.tsx
```diff
@@ -20,7 +20,7 @@ import {
 import {
   SvgChevronRightDouble,
   SvgFolder,
-  SvgMoreSmall,
+  SvgMore,
```
FlyersPh9, line 23: Can we revert this change in breadcrumbs?

When the breadcrumbs were created in HTML/CSS first, I was using the `…` text symbol.  [I never noticed until recently](https://github.com/iTwin/iTwinUI/pull/1060/files#r1101599521) that Storybook had an example using `SvgMoreSmall`.  I would almost rather see breadcrumbs use the `…` text symbol rather than any SVG, but for sake of this keeping this PR scoped I would say just leave it as is.  You can close #1151 as well.
What I would prefer to do rather than use an SVG

The difference I see between these 2 scenarios:
- Breadcrumbs is using `…` express there content that is being truncated, just like when setting `text-overflow: ellipsis;` in CSS.
- Tabs is using `SvgMore` as a symbol to show there is a menu, similar to in tiles component.  Tabs could very well be using a different looking symbol, it just so happens that we're using 3 dots to express this here as well.
Screenshot 2023-03-22 at 3 44 08 PM

Maybe the better thing for us to consider would be replacing `SvgMore` with some other symbol?  @VincentChabot1, something to keep an eye out for in other design systems for inspiration.

.changeset/lovely-poems-remember.md
elephantcatdog: We do: https://github.com/iTwin/iTwinUI/pull/984/files

I think it's not often necessary when fixing bugs.

apps/storybook/src/Breadcrumbs.stories.tsx
FlyersPh9: Had a conversation with Leah about this icon.  The icon is not a default, it's only a sample, so leaving the icon change is fine.  Regarding #1151, I think we need to have a bigger discussion re: consistency.  We've added an item to our internal document for next major release planning to consider any changes here.

elephantcatdog:
I'm undoing the breadcrumbs changes and putting them in [this other PR](https://github.com/iTwin/iTwinUI/pull/1151/files) because I cannot get the tests to work, and it's probably better to be in its own PR anyway; scope and stuff.

elephantcatdog:
image
😑

I don't know why this is failing. I undid the breadcrumbs changes. Whyyyyyyyyy

elephantcatdog:
Two known issues: 
1. On storybook, clicking on a tab breaks the page. This is specific to storybook and doesn't happen in the doc site example.
2. The horizontal autoscrolling does not work correctly if you refresh when on the horizontal overflow example in storybook. This occurs because there are passive events running after the initial render that cause the font to change. The calculated distance to scroll happens before the passive events (in a useEffect, which is supposed to happen only after the render) when the font is different. When the font changes, it's larger and pushes the tabs over. Refreshing on a different page and then switching to the horizontal tabs shows how it should act.

I also need to confirm that auto scrolling works when the active tab is to the left of the visible tabs.

elephantcatdog:
Oh! I just thought of another issue. In the doc site, it's hard to scroll horizontal with vertical mouse wheel because the page also scrolls, so your mouse cursor leaves the box and you can no longer scroll.

gretanausedaite (commented):
I think we need more indication for scrolling. Maybe we could add drop-shadow/fade at the ends:
image

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -15,6 +15,24 @@ import {
 import '@itwin/itwinui-css/css/tabs.css';
 import { Tab } from './Tab';
 
+export type OverflowOptions = {
+  /**
+   * Whether to allow tabs list to scroll when there is overflow,
+   * i.e. when there is not enough space to fit all the tabs.
+   *
+   * Not applicable to `type: 'pill'`.
```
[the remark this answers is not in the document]
elephantcatdog: Pill tabs take up the entire width of the given space. Buttons aren't applicable to pill type so I assumed there was a reason for that (that pill tabs are a specific use case and should have a specific look? and/or that it's too difficult programmatically?) so I thought the same is true for overflow.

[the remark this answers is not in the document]
gretanausedaite: I think it's a valid use case to have overflowing pill tabs. That fact that they can't have extra buttons doesn't mean there can't be a lot of tabs.

gretanausedaite (commented):
Auto scroll looks better.
Still need to fix styles - add fade/shadow to the end to make it more clear that these tabs are scrollable.
Also maybe add scrollbar to be visible on hover.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -279,9 +464,15 @@ export const Tabs = (props: TabsProps) => {
       case 'Enter':
       case ' ':
       case 'Spacebar': {
-        event.preventDefault();
-        if (focusActivationMode === 'manual' && focusedIndex !== undefined) {
-          onTabClick(focusedIndex);
+        if (
```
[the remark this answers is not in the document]
elephantcatdog: Reverted

elephantcatdog:
Added fade to end and scrollbar on hover

elephantcatdog:
image
Maybe an alternative. Uses `clip-path: polygon(10px 0%, 100% 0%, 100% 100%, 10px 100%, 0% 50%);` but would need to be adjusted to angle both sides, and probably add a border or a background.

elephantcatdog:
https://stackoverflow.com/questions/72329195/react-component-getting-wrong-ref-dimensions-in-useeffect for my meh fix for intial placement of active tab.

elephantcatdog:
Active tab never sits in fade:
https://user-images.githubusercontent.com/37936169/235053098-2d6e5a15-1221-4f27-9959-401c0b7937fa.mp4

LostABike (commented):
When I make the tab name a little longer (via `labels` prop not browser inspection), and in smaller screen like mobile or tablet, the text does not get truncated/faded at the end but you can scroll past the limit
image

I think this is tricky to design, do we want to be able to (in this case) horizontal scroll on each tab label? Which would be tacky. Or is it a better practice to limit the users on how long label should be, since they are just labels after all. Or we just leave it at that and note to users they can implement something like a tooltip to view the full label if it is too long (like browsers' tabs).

Either ways I think it would still be a good idea to have it truncated/faded out correctly like on browser so behavior is unified.

elephantcatdog:
@LostABike I'm just going to make it match what the vertical tabs do without overflow. It just disallows the tabs to be shrunk past the length of the tabs.

elephantcatdog:
Need to update test images

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -99,6 +99,24 @@ $borderless-horizontal-tab-min-height: calc(
     }
   }
 
+  &[data-iui-overflow='true'] {
+    .iui-tab {
+      white-space: nowrap;
+    }
+
+    &.iui-borderless {
```
gretanausedaite, line 107: Remove custom scrollbar styling. We should keep it consistent across all components.

.changeset/lovely-poems-remember.md
```diff
@@ -0,0 +1,20 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': minor
+---
+
+- Added data attribute 'data-iui-overflow' - when true it adds styling for overflow tabs
+- Added property 'overflowOptions' - contains `useOverflow`, which when true enables tabs to scroll if there's overflow
+
+```typescript
+const [index, setIndex] = React.useState(0);
```
gretanausedaite, line 10: Keep this example as simple as possible. I don't think users need useState or return statements.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -11,10 +11,47 @@ import {
...
+  /**
+   * Whether to allow tabs list to scroll when there is overflow,
+   * i.e. when there is not enough space to fit all the tabs.
+   *
+   * Not applicable to type `pill`.
+   *
+   * @default false
+   * @example <caption>Enables scrolling for overflowing tabs</caption>
+   *  return (
+   *    <Tabs useOverflow={true}>
```
gretanausedaite, line 29: Remove example.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -146,6 +184,13 @@ export const Tabs = (props: TabsProps) => {
     props = { ...props };
     delete props.actions;
   }
+  // Separate overflowOptions from props to avoid adding it to the DOM (using {...rest})
+  let overflowOptions: OverflowOptions | undefined;
+  if (props.type !== 'pill' && props.overflowOptions) {
```
gretanausedaite, line 189: Maybe we should not allow scrolling on borderless tabs as it covers border which indicates selected tab.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -218,13 +263,215 @@ export const Tabs = (props: TabsProps) => {
...
+      return 0; // tab is visible
+    } else if (tabStart < visibleStart + fadeBuffer) {
+      return -1; // tab is before visible section
+    } else {
+      return 1; // tab is after visible section
     }
-    setCurrentActiveIndex(index);
   };
 
+  function easeInOutQuad(
```
gretanausedaite, line 326: Why not arrow function?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -218,13 +263,215 @@ export const Tabs = (props: TabsProps) => {
...
+    }, 50);
+  }, [
+    overflowOptions?.useOverflow,
+    currentActiveIndex,
+    focusedIndex,
+    orientation,
+    scrollToTab,
+  ]);
+
+  const [scrollingPlacement, setScrollingPlacement] = React.useState('none');
```
gretanausedaite, line 417: Having 'none' initial value adds this prop to ALL tabs, even not overflowing ones. Change to undefined.
```suggestion
  const [scrollingPlacement, setScrollingPlacement] = React.useState('none');  const [scrollingPlacement, setScrollingPlacement] = React.useState<
    string | undefined
  >(undefined);
```

packages/itwinui-react/src/core/Tabs/Tabs.test.tsx
```diff
@@ -76,6 +76,63 @@ it('should render vertical tabs', () => {
...
+        <Tab key={9} label='Label 9' />,
+      ]}
+    >
+      Test content
+    </Tabs>,
+  );
+
+  const tabContainer = container.querySelector('.iui-tabs') as HTMLElement;
+  expect(tabContainer).toBeTruthy();
+  expect(tabContainer).toHaveAttribute('data-iui-overflow', 'true');
```
gretanausedaite, line 130: You are not testing placement prop or mouse wheel scroll.

gretanausedaite (commented):
Pill overflow image change seems to be obsolete.
Add overflowing tabs example in css html pages.

packages/itwinui-css/src/tabs/tabs.scss
elephantcatdog: If I do that, it covers up the highlighted line.

gretanausedaite: That's ok because of this [comment](https://github.com/iTwin/iTwinUI/pull/1060#discussion_r1188446917)

LostABike (commented):
For very long tab content, the wrapper just increased in length instead of staying the same and make content scrollable? That is not the right behavior right?
image

LostABike (commented):
And with vertical tab, content just flows out of the wrapper altogether
image

gretanausedaite:
> For very long tab content, the wrapper just increased in length instead of staying the same and make content scrollable? That is not the right behavior right?

It have the same behaviour right now in main.

> And with vertical tab, content just flows out of the wrapper altogether

overflow: auto added to vertical tabs fixes that, but then it gets two scrollbars. I guess it's better than overflowing content.
image

LostABike (approved):

PR #1062, opened 2023-02-08
fix VirtualScroll minHeight update for react 17

### LINKED ISSUE #1050: ComboBox: List container of virtualized boxes not shrinking to content after a change to emptyStateMessage
A MenuItemSkeleton is visible then replaced by a smaller message. A small gap can be seen below the message item. This gap disappears after additional interactions with the ComboBox.

#### Expected behavior
A MenuItemSkeleton is visible then replaced by a smaller message. No gaps can be seen around the smaller message.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/itwinui-react-example-forked-804s68?file=/src/App.tsx

##### Steps to reproduce
1. Open ComboBox dropdown and observe the MenuItemSkeleton element
1. Wait 1.5 seconds for the delayed update
1. Observe that the "No Results." message is smaller than the previous MenuItemSkeleton element resulting in a gap beneath

#### Additional information
This bug only occurs to ComboBoxes with `enableVirtualization` set to true. Specifically, it is the result of the containing `iui-surface` div added for virtualization not rerendering and resizing to the new content.

mayank99:
## Changes

React 18 automatically batches all updates, but React 17 doesn't. We can opt into this by using `unstable_batchedUpdates`, which fixes #1050.

Yes, this is an "unstable" API, but it will be around for the forseeable future and I've added a noop fallback just in case. See https://github.com/reactwg/react-18/discussions/21

## Testing

Tested by manually replicating this sandbox in vite playground: https://codesandbox.io/s/itwinui-react-example-forked-qk7dyt?file=/src/index.tsx

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/utils/components/VirtualScroll.tsx
+++ b/packages/itwinui-react/src/core/utils/components/VirtualScroll.tsx
@@ -3,11 +3,15 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
+import * as ReactDOM from 'react-dom';
 import {
   mergeRefs,
   useResizeObserver,
   useIsomorphicLayoutEffect,
 } from '../hooks';
+
+const unstable_batchedUpdates =
+  ReactDOM.unstable_batchedUpdates ?? ((cb: () => void) => void cb());
 
 const getScrollableParent = (
   element: HTMLElement | null,
@@ -224,9 +228,10 @@ export const useVirtualization = (props: VirtualScrollProps) => {
       if (height > 0) {
         setIsMounted(true);
       }
-      setScrollContainerHeight(height);
-
-      updateChildHeight();
+      unstable_batchedUpdates(() => {
+        setScrollContainerHeight(height);
+        updateChildHeight();
+      });
     },
     [updateChildHeight],
   );
--- a/.changeset/clever-moons-clap.md
+++ b/.changeset/clever-moons-clap.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue in ComboBox where the height of empty state content was not getting updated when using `enableVirtualization`.


FlyersPh9 (approved):

packages/itwinui-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -3,12 +3,16 @@
...
 import React from 'react';
+import * as ReactDOM from 'react-dom';
 import {
   mergeRefs,
   useResizeObserver,
   useIsomorphicLayoutEffect,
 } from '../hooks';
 
+const unstable_batchedUpdates =
+  ReactDOM.unstable_batchedUpdates ?? ((cb: () => void) => void cb());
```
LostABike, line 14: So if I understand this correctly, if we're in React 17 and there is no `ReactDOM.unstable_batchedUpdates`, then we use a callback function that executes statements immediately instead - in this case updating the VirtualScroll height. Then what was originally React 17 behavior that caused the update to not work as expected?

mayank99: no, this is for future React versions which might not have `ReactDOM.unstable_batchedUpdates`.


we want our library to continue to work with future react versions, so this code is basically replacing it with a function that executes the callback immediately.

in react 17, it will still use `ReactDOM.unstable_batchedUpdates`, which is why it fixes the bug.

LostABike: I see. Would not having a fallback function at all would also work because React 18 and beyond do batching automatically? I.e use `unstable_batchedUpdates` if it exists, otherwise don't use anything

mayank99: > use unstable_batchedUpdates if it exists, otherwise don't use anything

that's what the fallback is doing though. if you're suggesting something else, maybe you can show some pseudo-code of how i would do it without the fallback

LostABike: No yea that's right. I was just a bit confused about that fallback function but it essentially does nothing. Anyway I have also manually tested this change and it works.

LostABike (approved):

PR #1066, opened 2023-02-08
Modal: fix overflow on mobile when content too large

### LINKED ISSUE #425: Modal appears larger than phone display

LostABike:
## Changes

Added max-height to `iui-dialog-content` because the content was going out of screen/overflow for smaller screens.

image

This is not a complete PR. The issue with setting max-height for dialog content is that in the case of full screen, it would affect the position of the buttons, because they are placed below the content and not relative to the full page (I might be overthinking this and that it is on the user to place their buttons to the right position...)

image

Alternatively, I tried setting max-height to inherit instead, which I think makes sense to have content min-height inherit from parents which is the dialog box that contains it, and that works for full screen. However because the content position is a bit below the dialog box, this is what happens

image

So a possible solution would be to do something like max-height: inherit - 20px, basically says content max-height is whatever the parent (dialog container) max-height is, but a bit smaller. But `inherit - something` is not a thing. Is something equivalence to that?

closes #425


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -171,6 +171,7 @@ $iui-dialogue-resizable-min-height: 7.75rem; // 7.75rem = 124px = title bar heig
   padding: 0 var(--iui-size-m);
   overflow-y: auto;
   overflow-y: overlay;
+  max-height: 70vh;
```
mayank99, line 174: i think instead of this, it might make more sense to set `max-height: 100%` on the dialog itself

LostABike: it seems like we already have a max-height set for the dialog. However the buttons would still be unreachable. I'm not sure why but I think probably because they are flex relative to the dialog.
image

Which is why I thought it'd be better to have max-height equal 100%/inherit minus [buttons height]? But setting smaller dialog max-height would cause the content to flow out like this 
image

Which circle back to why I was messing with the content max-height instead of the dialog 🤔

.changeset/six-radios-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Draggable dialog now have a min height set so that the dialog can't be minimized more than where the dialog buttons are. It can also be optionally customized using the `$min-height` variable
```
mayank99, line 5: This changeset looks unintentional. I think it might be leftover from your old branch.

packages/itwinui-css/src/dialog/dialog.scss
mayank99: possibly related: i noticed the same issue happens with width too.



whereas full page dialog looks fine: 


so we might want to use whatever approach full page dialog is using.

mayank99: The width issue is fixed by changing line 53:

```diff
-  min-width: $iui-dialog-width;
+  min-width: Min(100%, $iui-dialog-width);
```

For the height, I just realized we didn't add `display: flex` on the regular dialog to keep backwards compatibility with an older version where `Dialog.Content` was optional.

mayank99: Try something like this on the dialog-content:
```css
max-height: calc(100% - 7.5rem);
```

you'll have to calculate the correct value manually.

LostABike: dialog-content max-height as a percentage do not seem to have any effect (also tried calc too)
image
image

but vh is good
image

LostABike: I think we could just do 100vh - [buttons size] ?

FlyersPh9:
I pulled this branch locally and ran it in the iOS Simulator to investigate.

I think part of the reason we're facing this issue is because the default dialog does not have `display: flex` set.  After adding these 2 lines & removing 1 line, it appears to fix the issue:
```diff
@mixin iui-dialog {
+ display: flex;
+ flex-direction: column;
  ...
}

@mixin iui-dialog-content {
  …
- max-height: 70vh;
}
```
If we add that, we could then remove those duplicated lines within `@mixin iui-dialog-full-page` & `@mixin iui-dialog-draggable`.

However, I think there was a reason we didn't do this.  Would changing the default dialog from `display: block` to `display: flex` be considered a breaking change?

mayank99 (approved):

FlyersPh9 (approved):

PR #1071, opened 2023-02-10
Modified padding for table cell based on different density settings to adjust horizontal spacing

### LINKED ISSUE #252: Condensed table has too big spacing
I think we could check up the horizontal spacing too, cause now it seems to big:
image

LostABike:
## Changes

~~Updated padding-left for expander cell based on the density setting: 16, 12 and 8px for `default`, `condensed` and `extra-condensed` respectively. Also adjusted individual left-padding for all cells (not just expander) also based on density.~~

Latest update:
- Reverted my initial change where I added left-padding for general table cells which messed up other tables. Instead, I only targeted the width of selection column and change that based on density, as well as the specific paddings for expandable rows. See images in testing section.

Closes #252
## Testing

For reviewers, you can now test this change on Storybook by toggling the density in `Table.stories.tsx`. You can also test it via Storybook control but that is a bit wonky and you will need to click another table and go back to see the horizontal spacing change.

Testing I did:
- Ran all tests for React and CSS
- Visual testing on various Storybook tables:

Default:
image

Condensed:
image

Extra-condensed:
image


mayank99:
@LostABike The main task was to reduce horizontal spacing in the existing densities.

Vertical spacing is also something that we could tackle, but that was a secondary task.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -219,7 +219,7 @@ export type TableProps<
    * Modify the density of the table (adjusts the row height).
    * @default 'default'
    */
-  density?: 'default' | 'condensed' | 'extra-condensed';
+  density?: 'default' | 'condensed' | 'extra-condensed' | 'mega-condensed';
```
FlyersPh9, line 222: I vote `most-condensed` and we make the table as condensed as possible so there can't be an even more condensed version.

In [our next major release](https://github.com/iTwin/iTwinUI/milestone/2) we can consider setting the size via a data attribute and use consistent size naming as other components (small, medium, large maybe).

LostABike: Oops yea let me clean that up

packages/itwinui-css/src/table/sizes.scss
```diff
@@ -17,8 +21,22 @@
     .iui-table-header .iui-table-cell {
       padding-block: calc(var(--iui-size-s) * 0.25);
     }
+
+    .iui-table-cell {
+      padding-left: calc(var(--iui-size-2xs));
```
FlyersPh9, line 26: I think we'll want the spacing to the left & right of columns to be the same.  On the far right side of the table, if a scrollbar is present, [the scrollbar is `var(--iui-size-xs)` wide](https://github.com/iTwin/iTwinUI/blob/f8a0c8aee62cc848094d3efa4302415d159a4756/packages/itwinui-css/src/utils/mixins.scss#L41).  I'm unsure if we'll want the left/right padding to be any less than that.

FlyersPh9:
Looks like we have some overlapping on the smallest size where the new file green dot is clashing with the filetype icon:
Screenshot 2023-02-10 at 3 39 16 PM

packages/itwinui-css/src/table/sizes.scss
```diff
@@ -17,8 +22,23 @@
...
       padding-block: calc(var(--iui-size-s) * 0.25);
     }
+
+    .iui-table-cell:not(:first-child) {
+      padding-left: calc(var(--iui-size-2xs));
+      padding-right: calc(var(--iui-size-2xs));
+    }
   }
 
+  //will probably remove this in the final edit because the assignment was to adjust horizontal spacing, not vertical
```
mayank99, line 32: remove all of this 😅

packages/itwinui-css/src/table/sizes.scss
```diff
@@ -8,6 +8,11 @@
   &[data-iui-size='condensed'] {
     --_iui-table-header-size: calc(var(--iui-size-s) * 4);
     --_iui-table-row-size: calc(var(--iui-size-s) * 4 + 2px);
+
+    .iui-table-cell:not(:first-child) {
+      padding-left: calc(var(--iui-size-xs));
```
mayank99, line 13: Is only changing the padding of the cell enough?

I believe one of the main concerns in the original issue was that row expanders were taking up too much space.

mayank99 (commented):
I was starting to review this and left some comments on the code, but then realized I don't know how to test most of these changes. 😅 I see no change in horizontal spacing in any stories when I toggle between the densities. Could you add some instructions and point out where exactly to look?

LostABike:
> 

You can now test the changes on Storybook, specifically for expander table. However, I must be missing a dependency for a hook somewhere because you have to click on another table and then go back to see the change after toggling the density. Please help :(

LostABike:
> I was starting to review this and left some comments on the code, but then realized I don't know how to test most of these changes. 😅 I see no change in horizontal spacing in any stories when I toggle between the densities. Could you add some instructions and point out where exactly to look?

You can now test the changes on Storybook, specifically for expander table. However, I must be missing a dependency for a hook somewhere because you have to click on another table and then go back to see the change after toggling the density. Please help :(

LostABike:
Updated description.

gretanausedaite (commented):
No additional comments.
Agree with [this one](https://github.com/iTwin/iTwinUI/pull/1071#discussion_r1107492165)

LostABike (commented):
test

.changeset/pretty-zebras-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
```
mayank99, line 2: should be `patch`, not `major`.
```suggestion
'@itwin/itwinui-react': patch
```

when you run `yarn changeset`, it first asks you "which packages have MAJOR changes?" you can skip that by pressing <kbd>Enter</kbd> (without selecting anything with <kbd>Space</kbd>).

packages/itwinui-css/src/table/sizes.scss
mayank99: i think it can be lowered even more. this is what i see in the current version:


it's only a difference of 8px from the live version, probably because you only changed the "constant" part of the equation.

try changing the "dynamic" part as well

LostABike: So I changed the multiplier for this based on the density. original 35 for default, 32 for condensed and 28 for extra-condensed. I saw the original author comment on how they got to 35 but I couldn't find where the expander_width + margin was set? Or how to manually calculate it? So I chose arbitrary multipliers that still look reasonably good/condensed.

mayank99: the items are misaligned now 


you might need to adjust the expander button's margin (yellow)


also it looks like the size of the first column is not going down (it's 48 regardless of density) :/

gretanausedaite (commented):
Code looks good.
Could you expand "Condensed" table story with subrows to have padding visual tests?
Also it would be awesome to have unit tests for padding calculation.

packages/itwinui-css/src/table/sizes.scss
LostABike: Regarding the first column/selection column not changing despite different density, it is working on my end by either: adding density to the return Table (in this specific case it looks like table Full 2 in Storybook), or by changing the density in Storybook control. How did you reproduce it?
image

mayank99: see https://itwin.github.io/iTwinUI/1071/react/?path=/story/core-table--full-2&args=density:condensed

maybe you forgot to push some changes?

LostABike: It is working for me there too. But as mentioned, I have to click on another table and return back to Full 2 table to see the change, if you are using the control to toggle density.

LostABike: I have updated to the suggested padding, which fixes misalignments issue. Also added expandability to condensed table and accompanied Storybook visual test, as well as Table unit test per @gretanausedaite suggestion.

packages/itwinui-react/src/core/Table/TableCell.tsx
```diff
@@ -40,10 +42,23 @@ export const TableCell = <T extends Record<string, unknown>>(
...
+      densityToPadding[density as keyof typeof densityToPadding] ?? 16;
+
+    const densityMultiplier = density === 'default' ? 34 : 30;
+
+    const paddingLeft =
+      densityPadding +
+      (cell.row.depth + (cell.row.canExpand ? 0 : 1)) * densityMultiplier;
+
+    return { paddingLeft };
   };
```
mayank99, line 62: Just one more suggestion. Currently this is a bit hard to understand, and the comment is disconnected with the code. if-conditions might be more readable

```suggestion
    const dynamicMargin = cell.row.depth + (cell.row.canExpand ? 0 : 1);

    let cellPadding = 16;
    let expanderMargin = 8;

    if (density === 'condensed') {
      cellPadding = 12;
      expanderMargin = 8;
    } else if (density === 'extra-condensed') {
      cellPadding = 8;
      expanderMargin = 8;
    }

    const multiplier = 27 + expanderMargin;

    return {
      paddingLeft: cellPadding + dynamicMargin * multiplier,
    };
```

mayank99 (approved):
Looks good!

mayank99:
it looks like some of the Table visual tests are failing though

gretanausedaite (approved):

packages/itwinui-react/src/core/Table/TableCell.tsx
```diff
@@ -42,24 +42,24 @@ export const TableCell = <T extends Record<string, unknown>>(
...
+      expanderMargin = 8;
+    } else if (density === 'extra-condensed') {
+      cellPadding = 8;
+      expanderMargin = 8;
+    }
 
-    const paddingLeft =
-      densityPadding +
-      (cell.row.depth + (cell.row.canExpand ? 0 : 1)) * densityMultiplier;
+    const multiplier = 27 + expanderMargin;
```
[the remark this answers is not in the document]
LostABike: Thanks for reviewing and suggesting/contributing 😸

Issue #887, opened 2022-11-14
Table: Grouped Header support


adamhock:
#### Feature
After iTwin/iTwinUI-react#153 was closed, `Table` now displays a header when a first-level `Header` is specified. However, support from grouped headers needs to be added

#### Examples
react-table Sandbox: https://codesandbox.io/s/github/tanstack/table/tree/main/examples/react/grouping?from-embed
iTwin/iTwinUI-react#153 Sandbox: https://codesandbox.io/s/shy-wind-1sp5q?file=/src/App.js

mayank99:
This will not be possible to achieve after #1072 and no one has requested it so I will close this. We can revisit after the upgrade to tanstack/react-table v8.

PR #1072, opened 2023-02-10
`Table`: allow passing top-level `Header` again

mayank99:
## Changes

I spent a few days trying to improve the typings for `columns` but I am convinced now that this is somewhat of an impossible task to do in a backwards compatible way. I tried a few different ideas and felt like I was making progress, but all of them were producing at least one type error in some scenario. The main difficulty here is that our types need to be compatible with the `Column` that the user might import from `react-table`.

So I've instead decided to reintroduce compatibility with the top-level dummy `Header`, effectively undoing the breaking change from https://github.com/iTwin/iTwinUI-react/pull/935 while still supporting the new shape added in that PR. In dev mode, there will now be runtime warning when the legacy columns shape is used, but it will not break anything (meaning we officially support both shapes in v2). As an added benefit, this change actually improves compatibility between v1 and v2, thus making migration easier.

I do not think this is worth spending any more time on, as we are going to migrate to react-table v8 in the next major.

Also, this change locks us out of supporting grouped column headers in v2 (#887), which was the original reason for the breaking change, but I do not think we would have done that in v2 anyway because no one has requested it.

## Testing

Tested in vite playground. When passing a top-level Header, there is no TS error, the UI looks correct, and there is a console warning only in dev mode.

Also added unit test for top-level Header.

## Docs

Added JSDoc, changeset, and console warning (in dev environment) linking to the migration guide.


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -597,6 +596,19 @@ export const Table = <
...
     setGlobalFilter,
   } = instance;
+
+  let headerGroups = _headerGroups;
+
+  if (columns.length === 1 && 'columns' in columns[0]) {
+    headerGroups = _headerGroups.slice(1);
+    if (isDev && !didLogWarning) {
+      console.warn(
+        `Table's \`columns\` prop should not have a top-level Header or sub-columns. It is only kept for backwards compatibility.\n See https://github.com/iTwin/iTwinUI/wiki/iTwinUI-react-v2-migration-guide#breaking-changes`,
```
gretanausedaite, line 606: Message should be the same as in config jsdocs.

packages/itwinui-react/src/types/react-table-config.ts
```diff
@@ -95,8 +95,11 @@ declare module 'react-table' {
       UseSortByOptions<D> {
     /**
      * List of columns.
+     *
+     * Should not have a top-level `Header` or a `columns` sub-property. They are allowed to be passed for backwards compatibility but are not officially supported.
+     * See https://github.com/iTwin/iTwinUI/wiki/iTwinUI-react-v2-migration-guide#breaking-changes`
      */
-    columns: Array<Column<any>>; // eslint-disable-line @typescript-eslint/no-explicit-any
+    columns: Array<Column<D>>; // eslint-disable-line @typescript-eslint/no-explicit-any
```
gretanausedaite, line 102: No need to disable eslint anymore

```suggestion
    columns: Array<Column<D>>;
```

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -597,6 +596,19 @@ export const Table = <
     visibleColumns,
     setGlobalFilter,
   } = instance;
+
+  let headerGroups = _headerGroups;
+
+  if (columns.length === 1 && 'columns' in columns[0]) {
```
gretanausedaite, line 602: Is there a way to test if this is added only in dev environment?

gretanausedaite (commented):
Looks good to me.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: 1. add an example in vite playground that uses top-level HeaDer
2. when you run `yarn dev` and open `localhost:1701`, you should see the log.
3. stop the dev servers and run `yarn dev --filter=vite-playground`
4. now run `yarn workspace vite-playground preview`, then open `localhost:4173`

packages/itwinui-react/src/types/react-table-config.ts
mayank99: hmm, yes i will remove. but now i actually can't remember why we made it `any` in the first place. need to investigate

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: updated a little bit. is it better now?

for users who don't know how it was before major release, they will use the new columns type anyway, so they won't see the message

packages/itwinui-react/src/core/Table/Table.tsx
gretanausedaite: I meant like unit tests for that. But I guess it's not that bad if it's not showing for some reason (if we break some code in the future or something).

gretanausedaite: Storybook also shows this warning.

packages/itwinui-react/src/types/react-table-config.ts
mayank99: Found this PR but the description is not helpful :/
https://github.com/iTwin/iTwinUI-react/pull/940

packages/itwinui-react/src/types/react-table-config.ts
```diff
@@ -190,7 +191,7 @@ declare module 'react-table' {
     /**
      * Filter component used as a column filter. Should use filters from `tableFilters`.
      */
-    Filter?: Renderer<FilterProps<D>>;
+    Filter?: Renderer<FilterProps<D>> | Renderer<TableFilterProps<D>>;
```
mayank99, line 194: this fixes https://github.com/iTwin/iTwinUI/issues/886#issuecomment-1362950469

mayank99:
@adamhock @LostABike Can you review this?

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -292,15 +300,11 @@ export type TableProps<
 } & Omit<CommonProps, 'title'>;
 
 // Original type for some reason is missing sub-columns
```
adamhock, line 302: I think this comment was for `ColumnType` which has been removed. Other than that, things look good to me!

adamhock (approved):

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: Removed

gretanausedaite (approved):

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -4580,3 +4580,51 @@ it('should navigate through table filtering with the keyboard', async () => {
...
+
+it('should ignore top-level Header if one is passed', async () => {
+  const data = mockedData();
+  const { container } = render(
+    <Table
+      data={data}
+      emptyTableContent='nothing to see here'
+      columns={[
+        {
+          Header: 'Header name',
```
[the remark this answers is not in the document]
mayank99: Yes, "top-level Header` refers this line and it should not be rendered.

To fully understand this behavior, you need to go all the way back to 2021. In iTwinUI v1, this bug was reported: https://github.com/iTwin/iTwinUI-react/issues/153

That bug was a _mistake_ in the original implementation, and it blocked us from adding proper support for grouped column headers.

In 2022, when we released iTwinUI v2, we wanted to fix this mistake and unblock the path for adding grouped column headers in the future. So this PR was merged: https://github.com/iTwin/iTwinUI-react/pull/935

That PR caused lots of typescript issues, so a temporary hacky fix was deployed at the very last minute (https://github.com/iTwin/iTwinUI-react/pull/940) and an issue was created (https://github.com/iTwin/iTwinUI/issues/886) to improve the typings later.

But after weeks/months of trying, it seems a bit impossible to improve these types in a non-breaking way, so my current PR is basically undoing all of the above (see PR description). In the future, we are going to migrate away from react-table v7, so the code will change significantly and therefore it is not worth spending any more time on this now.

Does that answer your question?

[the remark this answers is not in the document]
LostABike: Okay I can see that, just the _top level_ (:smirk:) understanding but I get the gist of it. Thank you for explaining.

LostABike (approved):

Issue #843, opened 2022-12-08
List component


jzbuchalski:
#### Feature
A simple list component (div or ul) with child list items (li or div) that will handle the background colors for hover, selected, disabled, etc?

For most of our list components, we'd rather avoid using variables, and have a base component handle the styling.  The less customization at our end, the better :-)

#### Examples

![image](https://user-images.githubusercontent.com/17307464/206556598-bb7a50fe-7d1c-4007-a101-95f3620416d2.png)

mayank99:
I agree we should have a lower primitive for lists, but if you need it quick, `iui-menu-item` will do the job in the meantime:
```html
<ul>
  <li class="iui-menu-item">Item 1</li>
  <li class="iui-menu-item iui-active">Item 2 (active)</li>
  <li class="iui-menu-item iui-disabled">Item 2 (disabled)</li>
</ul>
<style>
  ul {
    /* these should probably be in your reset */
    padding: 0;
    margin: 0;
  }
</style>
```

jzbuchalski:
LOL, we have been using iui-menu-item, but not overly intuitive :-)

jzbuchalski:
As you can imagine:

<List>
    <ListItem>
       My content
    </ListItem>
</List>

jzbuchalski:
I guess the selected state would be complicated because of single versus multiple selection.  Maybe that could be left out.

FlyersPh9:
@jzbuchalski can you share more information on how this list component will be used?

jzbuchalski:
In most (if not all) of our widgets, we display list of items.  The list is either a tree (coming from app ui, actually viewer components react), or a flat list.  For the flat lists, we are creating out own components, either a div or ul, with child divs (or li's).  The challenge is getting the different hover, disabled, selected states to match.  If we could get, say, a list item primitive, we would eliminate different groups having to track down how to handle the different item states.  I am open to discussion?

mayank99:
So your list is supposed to have interactable elements inside it, like `<button>`/`<a>`? Because that somewhat complicates things - the list item padding would make the clickable portion smaller (detailed in #840).

PR #1073, opened 2023-02-10
feat(TransferList): Add transfer list component

### LINKED ISSUE #634: Add listbox / transfer list component
We should add a [listbox](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html) / [transfer list](https://mui.com/material-ui/react-transfer-list/) like component.  I think we should be able to use the styling present in `menu.scss` for the list boxes & our buttons for the left / right buttons.

Requested by Vikram More / SYNCHRO.

FlyersPh9:
## Changes

- Introduce new `TransferList` component which is used to move one or more items between lists.
```jsx
    <TransferList>
      <TransferList.ListboxWrapper>
        <TransferList.ListboxLabel>Listbox 1</TransferList.ListboxLabel>
        <TransferList.Listbox>
          <TransferList.Item>Item 1</TransferList.Item>          
          <TransferList.Item>Item 2</TransferList.Item>          
          <TransferList.Item>Item 3</TransferList.Item>
        </TransferList.Listbox>
      </TransferList.ListboxWrapper>
      <TransferList.Toolbar>
        <IconButton />
        <IconButton />
        <IconButton />
        <IconButton />
      </TransferList.Toolbar>
      <TransferList.ListboxWrapper>
        <TransferList.ListboxLabel>Listbox 2</TransferList.ListboxLabel>
        <TransferList.Listbox>
          <TransferList.Item>Item 4</TransferList.Item>          
          <TransferList.Item>Item 5</TransferList.Item>          
          <TransferList.Item>Item 6</TransferList.Item>
        </TransferList.Listbox>
      </TransferList.ListboxWrapper>
    </TransferList>
  );
```
- Allows for use to place anything they want above / below each transfer list box (example has search & checkbox).
- Using `.iui-list` from #1106.

Closes #634.

## Testing

- New unit tests and HTML demo page
- Referenced [w3 listbox article](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/) for accessibility best practices.

## Docs

- New changesets, website documentation, and stories added


FlyersPh9:
In [the W3 example](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/) when hitting the tab key the focus goes to the listbox, not the individual listbox items.  The user can then use the keyboard up / down keys to select an item.  I think this would be preferred over how it is functioning now.  For example: if the left listbox had 100+ items, it would be painful to have to tab through all 100 items before focus moves over to the right listbox.

I imagine to make that work I'll need to make these changes in this PR:
- Add `@include iui-focus` to `@mixin iui-menu-listbox`.
- Add `tabindex="0"` to `<ul class="iui-menu" role="listbox">`.
- Remove `tabindex="0"` from all `<li class="iui-menu-item" role="option">`

And the rest of the functionality would be added in JS / React.  Is that a correct assumption?

packages/itwinui-css/src/utils/input-container/input-container.scss
```diff
@@ -35,7 +35,8 @@
 
   .iui-input,
   .iui-input-with-icon,
-  .iui-input-group {
+  .iui-input-group,
+  .iui-menu {
```
FlyersPh9, line 39: Maybe this should be `.iui-menu[role='listbox']` instead?

FlyersPh9:
I've re-run tests on GitHub a couple of times and I keep getting 1 failure.  I'm not getting the failure when running tests locally.  Unsure how to get this passing.

mayank99:
I re-ran the tests, and CSS is passing entirely. I think the failures in React are legitimate. They are happening in the Select component, which is likely affected by your changes.

elephantcatdog:
The open combobox and select tests are failing because of the missing shadow/border at the bottom.
image
image

packages/itwinui-css/src/menu/classes.scss
```diff
@@ -4,6 +4,10 @@
 
 .iui-menu {
   @include iui-list-menu;
+
+  .iui-listbox-wrapper & {
```
FlyersPh9, line 8: Thank you @elephantcatdog for the tip as to what was failing.  This change fixed the failed test images:
```diff
- &:where([role='listbox'])
+ .iui-listbox-wrapper & {
```

But depending on how #843 pans out, this may not be needed.  @mayank99

mayank99:
@FlyersPh9 I've created #1106, which has very minimal changes, but will allow you to use `iui-list`/`iui-list-item` classes. Take a look and then we can think on how much styling belongs in which component.

I also see some screenshots in #634 which I feel would need a bit more work.

packages/itwinui-css/backstop/tests/listbox.html
```diff
@@ -0,0 +1,197 @@
...
+  </head>
+  <body class="iui-root">
+    <theme-button></theme-button>
+    <h1>Listbox</h1>
+    <hr />
+
+    <section id="demo-default">
+
+      <div class="iui-listbox-wrapper">
+        <div class="iui-input-container iui-with-message">
```
mayank99, line 43: _input_-container is meant for _inputs_. we do not expose it publicly (it's marked internal) and i dont think we should complicate this component even more by trying to force it into listbox.

packages/itwinui-css/src/listbox/listbox.scss
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+@mixin iui-listbox-wrapper {
+  --iui-menu-listbox-height: calc((var(--iui-component-height) - 1px) * 4.5);
```
mayank99, line 5: why not just set `height`?

packages/itwinui-css/src/listbox/listbox.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  display: flex;
+  gap: var(--iui-size-xs);
+
+  > :where(:first-child, :last-child) {
+    flex: 1;
```
mayank99, line 11: `:first-child`/`:last-child` selectors are not very reliable. maybe `grid` will be better as it will let you bypass the need for these.

FlyersPh9: Yeah, `grid` might be a better choice here.  Using `grid` now.

packages/itwinui-css/src/listbox/listbox.scss
FlyersPh9: Trying to refresh my memory on this one, it's been a while.

I think I set the height on the `class="iui-list" role="listbox"` to avoid inconsistent heights if the listboxes contained a different amount of items:
Screenshot 2023-04-06 at 4 43 29 PM

And if `flex: 1` was applied but there was other stuff surrounding it:
Screenshot 2023-04-06 at 4 44 28 PM

Pink background applied to `.iui-listbox-wrapper`.

mayank99: Let me rephrase the question. Why not set `height: calc((var(--iui-component-height) - 1px) * 4.5)` directly on the listbox rather than set a CSS variable here first and use it in the listbox?

I don't see how having different number of items affects it.

adamhock:
The React portion of the PR is updated and ready for review.

packages/itwinui-css/backstop/tests/listbox.html
```diff
@@ -0,0 +1,213 @@
...
+    <theme-button></theme-button>
+    <h1>Listbox</h1>
+    <hr />
+
+    <section id="demo-default">
+
+      <div class="iui-listbox-wrapper">
+
+        <div class="iui-listbox-area">
+          <div class="iui-listbox-label">Options</div>
```
adamhock, line 47: I placed `iui-listbox-label` inside of `iui-listbox-area` instead of listing them all at the top. This allowed me to get rid of an empty `<div>` that was placed between the labels and used to correct the formatting. The test images have changed slightly as a result

apps/website/src/pages/docs/listbox.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: Listbox
+description: TODO Write Listbox description
```
adamhock, line 3: There is still some TODOs for the documentation of this component. I'm not exactly sure how to describe this component without repeating the `List` documentation

packages/itwinui-css/src/listbox/listbox.scss
adamhock: Height is now being set directly in listbox and not through the use of a variable

packages/itwinui-css/backstop/tests/listbox.html
adamhock: Search box has been removed

packages/itwinui-css/src/listbox/listbox.scss
mayank99: yes, thank you!

mayank99:
Currently it is missing all important interactions, such the ability to select items and navigate through them using keyboard. I'd imagine we'd also need to maintain state of currently selected items and provide props for user to control them.

Also I think the component should be renamed. "Listbox" by itself is only a small part of this whole pattern, and `role=listbox` is also used in other components (such as ComboBox). A better name might be "TransferList" (stolen from [mui](https://mui.com/material-ui/react-transfer-list/) / [mantine](https://mantine.dev/core/transfer-list/)).

apps/storybook/src/Listbox.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/Listbox',
+} as Meta<ListboxProps>;
+
+export const Basic: Story<ListboxProps> = (args) => {
+  return (
+    <Listbox {...args}>
+      <Listbox.Area>
```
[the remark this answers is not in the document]
elephantcatdog: Why can't you remove it? It looks the same without lines 32 & 41 to me.

apps/storybook/src/Listbox.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <IconButton styleType={'borderless'}>
+          <SvgChevronDown />
+        </IconButton>
+        <IconButton styleType={'borderless'}>
+          <SvgChevronLeft />
+        </IconButton>
+        <IconButton styleType={'borderless'}>
+          <SvgChevronRight />
+        </IconButton>
+      </Listbox.Toolbar>
```
elephantcatdog, line 93: This toolbar is centered based on the `Listbox` instead of between the two `Listbox.Areas`.
image
I think it should be centered between the two areas if it's meant to move items between them.

elephantcatdog: Oh, I see it's centered in the html page.

apps/website/src/pages/docs/listbox.mdx
elephantcatdog: Maybe use something like "The Listbox component is used to display a `ListItem`s within the confines of a box"? I would just sprinkle "in a box" around haha

packages/itwinui-react/src/core/index.ts
```diff
@@ -139,6 +139,9 @@ export type {
   ListItemDescriptionProps,
 } from './List';
 
+export { Listbox } from './Listbox';
```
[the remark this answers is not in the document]
elephantcatdog: I put "listbox" in google and it seemed split on "Listbox" vs "ListBox". There was also "List Box/list box" in there. Seems fine either way to me.

adamhock:
> Currently it is missing all important interactions, such the ability to select items and navigate through them using keyboard. I'd imagine we'd also need to maintain state of currently selected items and provide props for user to control them.
> 
> Also I think the component should be renamed. "Listbox" by itself is only a small part of this whole pattern, and `role=listbox` is also used in other components (such as ComboBox). A better name might be "TransferList" (stolen from [mui](https://mui.com/material-ui/react-transfer-list/) / [mantine](https://mantine.dev/core/transfer-list/)).

Implemented functionality for selecting items and navigating through them using the keyboard. The unit test for focusing in `List.test.tsx` is failing :/

I will push rename changes shortly

mayank99:
> Implemented functionality for selecting items and navigating through them using the keyboard. The unit test for focusing in `List.test.tsx` is failing :/

There should be little to no changes in List and ListItem components. Both of those are generic low-level components that can be used in many places, so we cannot handle specific TransferList behaviors in there.

adamhock:
> > Implemented functionality for selecting items and navigating through them using the keyboard. The unit test for focusing in `List.test.tsx` is failing :/
> 
> There should be little to no changes in List and ListItem components. Both of those are generic low-level components that can be used in many places, so we cannot handle specific TransferList behaviors in there.

This has been changed

adamhock:
> The unit test for focusing in `List.test.tsx` is failing :/
> 

This unit test has been fixed

mayank99:
> > > Implemented functionality for selecting items and navigating through them using the keyboard. The unit test for focusing in `List.test.tsx` is failing :/
> > 
> > 
> > There should be little to no changes in List and ListItem components. Both of those are generic low-level components that can be used in many places, so we cannot handle specific TransferList behaviors in there.
> 
> This has been changed

I see selection and arrow keys work internally, but currently there is no way for the user to access it. You need to expose the selection state and provide an easy way to implement a fully functional transferlist, such that items can be moved back and forth between the left and right listboxes and the state can be maintained by the user.

packages/itwinui-react/src/core/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export type ListboxProps = {
+  /**
+   * Content in the listbox.
+   */
+  children: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * The Listbox container allows content to appear elevated through the use of a drop shadow
```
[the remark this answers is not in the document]
adamhock: This has been changed

adamhock:
> > > > Implemented functionality for selecting items and navigating through them using the keyboard. The unit test for focusing in `List.test.tsx` is failing :/
> > > 
> > > 
> > > There should be little to no changes in List and ListItem components. Both of those are generic low-level components that can be used in many places, so we cannot handle specific TransferList behaviors in there.
> > 
> > 
> > This has been changed
> 
> I see selection and arrow keys work internally, but currently there is no way for the user to access it. You need to expose the selection state and provide an easy way to implement a fully functional transferlist, such that items can be moved back and forth between the left and right listboxes and the state can be maintained by the user.

The selection state is exposed and the toolbar story is now fully functional

apps/storybook/src/TransferList.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+    { name: 'Item 3', active: false },
+    { name: 'Item 4', active: false },
+    { name: 'Item 5', active: false },
+    { name: 'Item 6', active: false },
+  ]);
+
+  return (
+    <TransferList {...args}>
+      <TransferList.Area>
+        <TransferList.List role={'listbox'}>
```
gretanausedaite, line 43: Why are you adding role='listbox'? I think that is already added in TransferList?

apps/storybook/src/TransferList.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+                }}
+              >
+                {item.name}
+              </TransferList.ListItem>
+            );
+          })}
+        </TransferList.List>
+      </TransferList.Area>
+      <TransferList.Toolbar>
+        <IconButton
```
gretanausedaite, line 182: I think these buttons should be TransferList subcomponents.

packages/itwinui-css/src/transfer-list/transfer-list.scss
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@use '../input-container/mixins';
+@use './mixins' as *;
+
+.iui-transfer-list-wrapper {
+  display: grid;
+  grid-template-columns: 1fr auto 1fr;
```
gretanausedaite, line 8: CSS suggestion: Maybe redo grid to have more than one row? Right now it messes up vertical centering when TransferList does not have message undeneath it.

packages/itwinui-react/src/core/List/List.tsx
```diff
@@ -12,20 +12,26 @@ import type {
 import '@itwin/itwinui-css/css/menu.css';
 
 export const List = React.forwardRef((props, ref) => {
-  const { as: Element = 'ul', className, ...rest } = props;
+  const { as: Element = 'ul', className, role = 'list', ...rest } = props;
```
mayank99, line 15: undo changes in this file. `role` prop should automatically be respected as it will be part of `...rest`.

packages/itwinui-react/src/core/List/ListItem.tsx
```diff
@@ -33,6 +33,7 @@ const ListItemComponent = React.forwardRef((props, ref) => {
       data-iui-size={size === 'large' ? 'large' : undefined}
       data-iui-actionable={actionable ? 'true' : undefined}
       data-iui-focused={focused ? 'true' : undefined}
+      tabIndex={disabled ? undefined : -1}
```
mayank99, line 36: undo changes in this file. `tabIndex` should be managed by the consumer of this component.

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+export type TransferListAreaProps<T extends React.ElementType = 'div'> =
+  PolymorphicComponentProps<T, TransferListAreaOwnProps>;
+
+const TransferListArea = React.forwardRef((props, ref) => {
+  const { as: Element = 'div', children, className, ...rest } = props;
+
+  return (
+    <Element
+      className={cx('iui-transfer-list-area', className)}
```
mayank99, line 34: is it possible to remove this extra wrapper div?

ideally, the component would look like this:

```
<TransferList>
  <TransferList.Listbox>
   <TransferList.Item>...</TransferList.Item>
   <TransferList.Item>...</TransferList.Item>
   <TransferList.Item>...</TransferList.Item>
  </TransferList.Listbox>

  <TransferList.Toolbar>...</TransferList.Toolbar>

  <TransferList.Listbox>
   <TransferList.Item>...</TransferList.Item>
   <TransferList.Item>...</TransferList.Item>
  </TransferList.Listbox>
</TransferList>
```

apps/storybook/src/TransferList.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+export default {
+  component: TransferList,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/TransferList',
+} as Meta<TransferListProps>;
+
+export const Basic: Story<TransferListProps> = (args) => {
```
mayank99, line 30: Does it make sense to show examples of transferlist with only one listbox? I'd imagine this component is pretty much always used with two listboxes.

apps/storybook/src/TransferList.stories.tsx
adamhock: This has been fixed

packages/itwinui-css/src/transfer-list/transfer-list.scss
adamhock: Genius. This has been added and the toolbar has been centered properly as a result

packages/itwinui-react/src/core/List/List.tsx
adamhock: Fixed

packages/itwinui-react/src/core/List/ListItem.tsx
adamhock: `tabIndex` has been moved from `ListItem` to `TransferList.ListItem`, but I think it's still necessary to define it for the keyboard navigation. Correct me if I am wrong on that

mayank99: yes, `TransferList.ListItem` is the "consumer" of this component, so that's the right place to set tabindex

mayank99: I think we will want to use `-1` for all items though, including disabled items.

Disabled items need to be focusable and should have `aria-disabled`, but clicking or pressing <kbd>Enter</kbd> while focusing them should do nothing.

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -0,0 +1,276 @@
...
+ *       <TransferList.ListItem>Item 3</TransferList.ListItem>
+ *       <TransferList.ListItem>Item 4</TransferList.ListItem>
+ *       <TransferList.ListItem>Item 5</TransferList.ListItem>
+ *       <TransferList.ListItem>Item 6</TransferList.ListItem>
+ *     </TransferList.List>
+ *   </TransferList.Area>
+ * </TransferList>
+ */
+export const TransferList = Object.assign(
+  React.forwardRef(
```
mayank99, line 236: move this component to a local variable, preferably near the top

also make it polymorphic

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -0,0 +1,276 @@
...
+    }
+  };
+
+  return (
+    <ListItem
+      ref={ref}
+      onClick={onClickEvents}
+      onKeyDown={onKeyDown}
+      active={active}
+      actionable={actionable}
```
mayank99, line 158: i think this should `actionable` by default

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+export type TransferListListItemProps<T extends React.ElementType = 'li'> =
+  PolymorphicComponentProps<T, TransferListListItemOwnProps>;
+
+const TransferListListItem = React.forwardRef((props, ref) => {
+  const { actionable, disabled, onActiveChange, children, active, ...rest } =
+    props;
+
+  const onClickEvents = () =>
+    actionable && onActiveChange && onActiveChange(active ?? false);
```
mayank99, line 135: why is `onActiveChange` called with the stale state? I would expect it to be called with the next expected state

packages/itwinui-css/src/menu/mixins.scss
```diff
@@ -187,3 +187,10 @@ $iui-active-outline: 1px solid var(--iui-color-border-accent);
     fill: var(--_iui-list-item-icon-fill, var(--iui-color-icon));
   }
 }
+
+@mixin iui-transfer-list {
```
mayank99, line 191: lets move this to `transfer-list.scss` and create a new class, e.g. `iui-transfer-list-listbox`

packages/itwinui-css/src/transfer-list/mixins.scss
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+@mixin iui-transfer-list($area: null) {
```
mayank99, line 3: Can we move all this code into `transfer-list.scss`? No need to create mixins when we can just put the code directly under the classes

packages/itwinui-react/src/core/TransferList/TransferList.tsx
mayank99: two things came to mind:
1. css grid could be used to keep the layout and get rid of the extra div
2. right now, it looks like the label is not even associated with the listbox (missing `aria-labelledby`), so maybe the `label` should be a prop on `TransferList.Listbox`.

packages/itwinui-css/src/transfer-list/transfer-list.scss
gretanausedaite: Is it valid case for users to add message under the second column? Is it possible to do that easily with current API?
image

packages/itwinui-react/src/core/List/ListItem.tsx
adamhock: This has been fixed

packages/itwinui-react/src/core/TransferList/TransferList.tsx
adamhock: Fixed

packages/itwinui-css/src/menu/mixins.scss
adamhock: Fixed

packages/itwinui-css/src/transfer-list/mixins.scss
adamhock: Fixed

packages/itwinui-css/src/menu/mixins.scss
mayank99: `role=listbox` needs to be brought back. It is needed, just not for styling.

packages/itwinui-css/src/transfer-list/transfer-list.scss
adamhock: This is easily possible to do with our API as long as there is an empty `div` element to take the place of the middle column

image
image

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -156,7 +162,9 @@ const TransferListListItem = React.forwardRef((props, ref) => {
       onKeyDown={onKeyDown}
       active={active}
       actionable={actionable}
-      tabIndex={disabled ? undefined : -1}
+      tabIndex={-1}
+      aria-disabled={disabled}
```
mayank99, line 177: Similar to `role=listbox` on the list, each of the listitems should have `role=option`. Also should have `aria-selected` when it's active.

See APG: https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/

After adding those, please also test in NVDA to confirm it's working.

adamhock: This has been added, but unit tests were not passing, so I still need to add those

packages/itwinui-css/src/menu/mixins.scss
adamhock: Fixed

packages/itwinui-react/src/core/TransferList/TransferList.tsx
adamhock: 1. I tried using `grid-template-area` and I don't think we can get the desired layout since the grid involves more than one element with the same class name. The grid would have to be constructed like so:

`'label . label'`
`'listbox toolbar listbox'`

Correct me if I'm wrong on this since I'm not too familiar with `display: grid`

2. If `TransferList.Label` is replaced with a `label` prop, that would lead to the label not being polymorphic. Is that an issue? Can we add a second prop to change the label's element type?

mayank99: if the css grid is too complicated, then lets keep the extra divs, but we still need to associate the label using `aria-labelledby`. context seems like the right solution for this.
```jsx
// this will set id in context
<TransferList.Area>

  // this will read id from context, and set `id` attribute
  <TransferList.Label label='Blah' />

  // this will read id from context, and set `aria-labelledby`
  <TransferList.List>...</TransferList.List>
</TransferList.Area>
```

also renaming suggestions:
- TransferList.Area -> TransferList.ListboxWrapper
- TransferList.List -> TransferList.Listbox
- TransferList.Label -> TransferList.ListboxLabel
- TransferList.ListItem -> TransferList.Item

packages/itwinui-react/src/core/TransferList/TransferList.tsx
adamhock: The unit test has been added

packages/itwinui-react/src/core/TransferList/TransferList.tsx
adamhock: Context is now used to pass a `labelId` to the label and listbox subcomponents

These components have also been renamed

apps/storybook/src/Listbox.stories.tsx
adamhock: This centering issue has been fixed

adamhock:
I will be making changes shortly that address the use of `polymorphic` and `useGlobals`

mayank99:
Reposting https://github.com/iTwin/iTwinUI/pull/1073#discussion_r1203745795 for visibility:

> Both of ListboxLabels gets the same labelId and ListBoxes are not described correctly. Example I'm using:
> image
> 
> "Avalable" listbox
> image
> 
> "Selected" listbox
> image

adamhock:
> Reposting [#1073 (comment)](https://github.com/iTwin/iTwinUI/pull/1073#discussion_r1203745795) for visibility:
> 
> > Both of ListboxLabels gets the same labelId and ListBoxes are not described correctly. Example I'm using:
> > image
> > "Avalable" listbox
> > image
> > "Selected" listbox
> > image

This has been fixed

.changeset/ninety-keys-double.md
```diff
@@ -0,0 +1,18 @@
+---
+'@itwin/itwinui-css': minor
+'@itwin/itwinui-react': minor
+---
+
+Added `TransferList` component.
+
+```jsx
```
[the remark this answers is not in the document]
adamhock: A description has been added. Let me know if the description is too wordy

adamhock:
> It seems that I can't reach second ListBox with keyboard (Tab).

I have fixed this by adding `tabIndex={0}` to `TransferList.Listbox`

apps/storybook/src/TransferList.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+              >
+                {item.name}
+              </TransferList.Item>
+            );
+          })}
+        </TransferList.Listbox>
+      </TransferList.ListboxWrapper>
+      <TransferList.Toolbar>
+        <IconButton
+          styleType={'borderless'}
```
gretanausedaite, line 229: nit: It would be nice if you added label to tool buttons. It would set an example for users.

packages/itwinui-react/src/core/TransferList/TransferList.tsx
```diff
@@ -0,0 +1,282 @@
...
+   * 	TransferList listbox label subcomponent
+   */
+  ListboxLabel: TransferListListboxLabel,
+  /**
+   * 	TransferList toolbar subcomponent
+   */
+  Toolbar: TransferListToolbar,
+});
+
+export const TransferListContext = React.createContext<
```
gretanausedaite, line 268: Is it a valid use case for ListBox to have disabled state (not individual list items, but full box disabled)?

.changeset/ninety-keys-double.md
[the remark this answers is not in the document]
gretanausedaite: A bit wordy. A simple comma separated list of subcomponents should be enough.

apps/storybook/src/TransferList.stories.tsx
adamhock: Labels have been added for each of the buttons

.changeset/ninety-keys-double.md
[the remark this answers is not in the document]
adamhock: Kept the `TransferList` description and then listed the new subcomponents

packages/itwinui-react/src/core/TransferList/TransferList.tsx
mayank99: Not that I can think of, but we can always revisit later if someone requests it.

packages/itwinui-react/src/core/TransferList/TransferList.tsx
adamhock: TransferListNVDA

Here is the NVDA test. Is it supposed to tell me when an item is selected?

mayank99: yes. I think you're missing `aria-multiselectable` (check the APG link above).

here's voiceover on safari with `aria-multiselectable`:

https://github.com/iTwin/iTwinUI/assets/9084735/d138346f-3594-4a4a-8be4-49612de67dcc

adamhock: `aria-multiselectable` has been added. I still need to apply `aria-activedescendant` to the listbox though

TransferListNVDAWithMultiSelect

mayank99: `aria-activedescendant` is useful when the focus is on a different element, e.g. in the case of ComboBox, the focus is on input rather than the options. In this case, the focus is actualy on the options, so it should be ok.

You can compare by testing that demo from APG using NVDA as well. The results should be very similar.

adamhock: TransferListNVDAW3Example

The results are very similar

mayank99 (approved):
LGTM

PR #1078, opened 2023-02-14
fix(Table): Fixed reordering column behavior when columns are hidden

### LINKED ISSUE #892: Table: columns become sticky right when added with column manager
After reordering columns and selecting not visible columns in column manager, they become sticky right.
#### Expected behavior
Columns should not become sticky after reordering and selecting not visible columns in column manager..

#### Reproduction
https://codesandbox.io/s/itwinui-react-example-forked-g9xroe?file=/src/App.tsx

##### Steps to reproduce
1. Add `sticky: right` column manager and enable column reordering.
2. In column manager unselect a few columns
3. Reorder existing columns. For example, drag the first column onto the second
4. In column manager select any not selected column
5. Selected column becomes sticky right

#### Additional information

`"itwinui-react": 1.47.1`

https://user-images.githubusercontent.com/75395229/196682818-296306eb-6805-467f-8135-acd1525220e4.mp4

adamhock:
## Changes

Fixed a bug which gave users an unexpected column order when columns were hidden, reordered, and then shown
Fixes #892 

## Testing
Before:
ColumnManagerBeforeReorderSolution

After:
ColumnManagerReorderSolution

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Table/hooks/useColumnDragAndDrop.tsx
+++ b/packages/itwinui-react/src/core/Table/hooks/useColumnDragAndDrop.tsx
@@ -98,7 +98,7 @@ const defaultGetDragAndDropProps =
       event.preventDefault();
       setOnDragColumnStyle(event);
 
-      const columnIds = instance.flatHeaders.map((x) => x.id);
+      const columnIds = instance.allColumns.map((x) => x.id);
       const srcIndex = instance.state.columnReorderStartIndex;
       const dstIndex = columnIds.findIndex((x) => x === header.id);
 
--- a/.changeset/brave-dragons-sparkle.md
+++ b/.changeset/brave-dragons-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed bug which gave an unexpected column order once columns were hidden, reordered, and then shown


packages/itwinui-react/src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -98,7 +98,7 @@ const defaultGetDragAndDropProps =
       event.preventDefault();
       setOnDragColumnStyle(event);
 
-      const columnIds = instance.flatHeaders.map((x) => x.id);
+      const columnIds = instance.allColumns.map((x) => x.id);
```
adamhock, line 101: Wasn't sure if it was better to use `allColumns` or `headers` here. `flatHeaders` was only returning the visible headers which was leading to the reordering issue on hidden columns

mayank99 (approved):

packages/itwinui-react/src/core/Table/hooks/useColumnDragAndDrop.tsx
elephantcatdog: I'm not sure either; if it's just getting the ids from them and those are the same it shouldn't matter thought.

elephantcatdog (approved):
Looks good to me, good job!

Issue #43, opened 2021-04-26
Tooltip bugs: blurring and shadow


mayank99:
After adding the frosted glass effect in #19, three different visual bugs show up:
1. [ ] Tooltip text is blurry if using CSS transforms (workaround in https://github.com/iTwin/iTwinUI-react/pull/25)
2. [ ] Tooltip target element is also blurry on hover (when tooltip shows) similar to #44
3. [ ] There is a very noticeable shadow covering the target element

All 3 are noticeable in this gif:
![tooltip](https://user-images.githubusercontent.com/9084735/116114421-544c1480-a687-11eb-987d-0d6bb4e713fa.gif)

**Environment info**: Windows 10, Chrome 89, itwinui-css@0.9.0, itwinui-react@1.1.0

PR #1082, opened 2023-02-15
blurred background on tooltip

mayank99:
## Changes

I was trying out something in a sandbox, when I noticed that the tooltip opacity makes it really hard to read when there is text behind the tooltip. So I'm making the background blurry.

## Testing

Looks perfectly fine in my testing on M1 Macbook and personal Windows PC.

I believe we had this 2 years ago but removed it because it was causing some blurry text (in react, not css). See #43. So I would like other Windows users to verify.

## Docs

Added changeset.

### DIFF
--- a/.changeset/violet-ligers-itch.md
+++ b/.changeset/violet-ligers-itch.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Tooltip's background now has a blur filter to make text more readable.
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):
No blur on macOS Brave or iOS Safari.

macOS:
Screenshot 2023-02-16 at 3 19 33 PM

iOS:
Screenshot 2023-02-16 at 3 19 40 PM

mayank99:
Note: In 2021, the blur was only present on react, so make sure you test the storybook.

elephantcatdog:
Before: 
image
After:
image

elephantcatdog:
No text it looks the same:
Before:
image

After: 
image

elephantcatdog:
Closeup @ 150%:
Before: 
image
After: 
image

It looks a little less crisp to me, but I think it's fine. It's not very noticeable, and is better than having text show through.

elephantcatdog (approved):

PR #1088, opened 2023-02-22
replace sandpack with static examples

mayank99:
## Changes

Sandpack comes with a huge bundle and takes forever to load, so I'm replacing it with statically rendered, hydrated examples (they load instantly and are also interactive). For editing, they can be opened in codesandbox externally.

In the process, I also improved a lot of the example code for readability/usability.

## Testing

Tested in local dev server and prod build.

Tested deploying on azure too. See https://dev-itwinui.bentley.com/

## Docs

N/A.

The approach for including code examples has now changed. Examples will first need to be exported in `src/examples/index.tsx` and then used in mdx files like this:

```tsx
import * as AllExamples from '~/examples';

<LiveExample src='Tabs.main.tsx'>
  <AllExamples.TabsMainExample client:load />
</LiveExample>
```


mayank99:
this PR is now ready for review. there are always improvements to be made, but it's in a good enough state now.

biggest issue i've noticed is that there is undesirable horizontal scrolling on mobile, but it might be unrelated to this PR

gretanausedaite (approved):
Code looks good. 🍰 
New approach is better than previous but now I'm a bit worried these code samples takes too much space in docs (eg. alert). Especially if only one prop value is changing.

FlyersPh9 (approved):
Wow so fast, much speed, such amaze!

But in all honesty, this performance is so much better than before.  I shared the dev link with @jbwriter to see if he has any thoughts on how the code is displayed.

jbwriter:
Wow. So much faster! Don't even notice them loading now. Great improvement!

mayank99:
Awesome, lets merge this then! There will always be improvements to be made, so we can revisit in the future based on more feedback/experience.

PR #1092, opened 2023-02-23
feat(ToggleSwitch): Add small size

### LINKED ISSUE #1091: Smaller size for toggle switch
Hey guys, 
we need a smaller size toggle switch as this one doesn't look good/proportional with base text. It's not good either in smaller panels.

image

FlyersPh9:
## Changes

- Swap necessary SASS variables for CSS variables.
- Introduce `data-iui-size="small"` and bump CSS variable values down by 1.
- Hide icon when small size is active.
- Swap hard coded margin values with CSS variables in demo page.

Screenshot 2023-02-23 at 9 49 07 AM

- Added `switchSize` prop to ToggleSwitch type in React, which changes `data-iui-size` depending on whether it is `small` or `default`. Default is `default` if nothing is specified.
- Added "Small" under Toggle Switch in storybook to demonstrate the smaller toggle switch based on the passed in `switchSize` prop.

image
image

~~Requires React portion,~~ closes #1091.

## Testing

- BackstopJS

## Docs

N/A


.changeset/eleven-mangos-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': minor
```
mayank99, line 2:
```suggestion
'@itwin/itwinui-css': minor
'@itwin/itwinui-react': minor
```

apps/storybook/src/ToggleSwitch.stories.tsx
```diff
@@ -23,6 +23,19 @@ export const Basic: Story<ToggleSwitchProps> = (args) => {
   return <ToggleSwitch defaultChecked {...args} />;
 };
 
+export const Small: Story<ToggleSwitchProps> = (args) => {
+  const { switchSize = 'small', ...rest } = args;
+  return (
+    <div>
```
mayank99, line 29: shouldn't need `div`

.changeset/eleven-mangos-mix.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add small size toggle switch option.
```
mayank99, line 5: you can add a tiny code snippet here showing how to use it

```tsx
<ToggleSwitch switchSize='small' />
```

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -26,6 +26,11 @@ export type ToggleSwitchProps = {
    * Icon inside the toggle switch. Shown only when toggle is checked.
    */
   icon?: JSX.Element;
+  /**
+   * Size of the toggle switch.
+   *  @default 'default'
+   */
+  switchSize?: 'default' | 'small';
```
mayank99, line 33: The `switchSize` name is a bit weird. Maybe we should use `size`? Only concern there is potential clash with the [native `size`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size) attribute, but that is not applicable to checkbox inputs, so we should be fine.

.changeset/eleven-mangos-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-css': minor
+'@itwin/itwinui-react': minor
+---
+
+Add small size toggle switch option using `size` prop as follow: `<ToggleSwitch size='small' />`
```
mayank99, line 6:
```suggestion
Add small size toggle switch option using `size` prop as follows: `<ToggleSwitch size='small' />`
```

mayank99 (approved):
Looks good 🚀

.changeset/eleven-mangos-mix.md
LostABike: Hold on, changing variable name to `size` was breaking my code which was why I named it `switchSize`. However it is not anymore but it is breaking the unit test. Do you happen to know why?

`
  ● Test suite failed to run

    src/core/ToggleSwitch/ToggleSwitch.test.tsx:35:46 - error TS2322: Type 'string' is not assignable to type 'undefined'.

    35   const { container } = render(<ToggleSwitch size='small' />);
                                                    ~~~~

      src/core/ToggleSwitch/ToggleSwitch.tsx:33:3
        33   size?: 'default' | 'small';
             ~~~~
        The expected type comes from property 'size' which is declared here on type 'IntrinsicAttributes & { label?: ReactNode; labelPosition?: "left" | "right" | undefined; setFocus?: boolean | undefined; icon?: Element | undefined; size?: "small" | ... 1 more ... | undefined; } & Omit<...> & RefAttributes<...>'
        `

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -26,6 +26,11 @@ export type ToggleSwitchProps = {
    * Icon inside the toggle switch. Shown only when toggle is checked.
    */
   icon?: JSX.Element;
+  /**
+   * Size of the toggle switch.
+   *  @default 'default'
+   */
+  size?: 'default' | 'small';
 } & Omit<React.InputHTMLAttributes<HTMLInputElement>, 'type'>;
```
mayank99, line 34: This should fix your test failure mentioned in https://github.com/iTwin/iTwinUI/pull/1092#discussion_r1123425633
```suggestion
} & Omit<React.InputHTMLAttributes<HTMLInputElement>, 'type' | 'size'>;
```

LostABike: Oh interesting! Thanks 😸

gretanausedaite (approved):

PR #1096, opened 2023-02-23
fix(Checkbox): Focus outline appears above checkbox border

### LINKED ISSUE #844: Radio / Checkbox border appears above focus outline
**How it currently looks:**
Screenshot 2022-12-08 at 2 54 53 PM

**How it should look:**
Screenshot 2022-12-08 at 2 54 05 PM

Moving the border styling out of `::before` onto the parent appears to fix this.  Double check that the solution does not regress #406.

## Changes
- Changed checkbox / radio border from `border` to `outline`.
- Moved checkbox / radio border from pseudo element `::before` to `.iui-checkbox` /

idakukimiya:
## Changes

- Changed checkbox / radio border from `border` to `outline`.
- Moved checkbox / radio border from pseudo element `::before` to `.iui-checkbox` / `.iui-radio`.

Closes #844.

## Testing

- Visually confirmed `:focus` outline no longer appears under checkbox / radio border in browser.
- Ran visual tests which passed.

## Docs

N/A


packages/itwinui-css/src/checkbox/checkbox.scss
```diff
@@ -72,17 +72,8 @@
...
-    position: absolute;
-    inset: 0;
-    transition: border-color var(--iui-duration-1) ease-out;
-    border-radius: inherit;
-    border-style: solid;
-    border-width: 1px;
-    border-color: var(--_iui-checkbox-border-color);
-  }
+  transition: border-color var(--iui-duration-1) ease-out;
+  border: solid 1px var(--_iui-checkbox-border-color);
```
FlyersPh9, line 76: If you check the bottom of the PR you'll see that visual tests are failing.  If you click the circled link it will take you [here](https://itwin.github.io/iTwinUI/1096/results/html_report/) where you can see what visuals changed (under the 64 FAILED tab).
Screenshot 2023-02-24 at 11 34 51 AM

So it would appear that adding a border is making the checkmark symbol smaller.  I think a way to fix this would be to use `outline` instead of `border` like so:
```diff
- transition: border-color var(--iui-duration-1) ease-out;
+ transition: outline-color var(--iui-duration-1) ease-out;
- border: solid 1px var(--_iui-checkbox-border-color);
+ outline: solid 1px var(--_iui-checkbox-border-color);
+ outline-offset: -1px;
```

I tested this locally and test images are passing with this proposed change.

mayank99 (commented):
Change looks good. Please fill out missing info in the PR description, add changeset, and sign CLA.

packages/itwinui-css/src/checkbox/checkbox.scss
mayank99: I believe we went with a psuedo-element because this is how it looks in Safari:


but that will fix automatically in Safari 16.4, so this is fine imo

elephantcatdog:
@idakukimiya Here's an example of what the PR description should look like: https://github.com/iTwin/iTwinUI/pull/1086
The preformatted text in the PR description also guides you on what to add to each section. Adding this information makes it easier to review, especially when the PRs get bigger.

.changeset/fluffy-cobras-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
```
FlyersPh9, line 2:
```suggestion
'@itwin/itwinui-css': patch
```
Following [semantic versioning](https://semver.org/), I believe this will be a patch.

.changeset/fluffy-cobras-smile.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+The issue was once the user use tab to select the checkbox, there would be a combination of blue and grey border. To resolve the issue is to make the blue border primary and not have the grey border visable. Originally, the css code was in psudo-element ::before. We took it out the psudo-element and add the code to parent with a few tweeks.
```
FlyersPh9, line 5: I don't think we have any guidelines on what the message should read like, but I would suggest something like:
```suggestion
Fixed issue where checkbox / radio border appeared above `:focus` outline.
```

mayank99 (approved):
Looks good

PR #1097, opened 2023-02-23
doc(Anchor): Added description

jbwriter:

### DIFF
--- a/apps/website/src/pages/docs/anchor.mdx
+++ b/apps/website/src/pages/docs/anchor.mdx
@@ -1,6 +1,6 @@
 ---
 title: Anchor
-description:
+description: The anchor component works like a simple hyperlink, targeting a specific URL. Additionally, it can also target a named location within a page. For example, you might want to target a content area that appears below the fold so the user does not have to scroll to discover it.
 layout: ./_layout.astro
 propsPath: '@itwin/itwinui-react/src/core/Typography/Body/Body.tsx'
 exampleCodeFile1: Anchor.main.tsx


apps/website/src/pages/docs/anchor.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Anchor
-description:
+description: The anchor component works like a simple hyperlink, targeting a specific URL. Additionally, it can also target a named location within a page. For example, you might want to target a content area that appears below the fold so the user does not have to scroll to discover it.
```
mayank99, line 3: This is a bit too wordy. Descriptions need to be short enough to fit in 3 small lines

elephantcatdog:
@elephantcatdog Update text to "an anchor"

apps/website/src/pages/docs/anchor.mdx
jbwriter: Valid. Didn't realize that the "Description" field is used in the tile synopsis. There are plenty reasons to provide more clarifying text (I think "wordy" might be a bit harsh ;) ), but I will include extra information about the component below the first live example and not in the "Description" field. First timer here.

apps/website/src/pages/docs/anchor.mdx
```diff
@@ -15,6 +15,8 @@ import LiveExample from '~/components/LiveExample.astro';
 
 <LiveExample src={frontmatter.exampleCodeFile1} />
 
+Consider using a named anchor when you want the user to land not only on a specific page, but to actually arrive at a content area that may be below the fold so the user does not have to scroll to discover it.
```
mayank99, line 18:
```suggestion
Consider using a named anchor when you want the user to land not only on a specific page, but also to arrive at a content area that may be below the fold so the user does not have to scroll to discover it.
```

apps/website/src/pages/docs/anchor.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Anchor
-description:
+description: The anchor component works like a simple hyperlink, targeting a specific URL or named anchor within a page.
```
mayank99, line 3:
```suggestion
description: The anchor component creates a hyperlink targeting a specific location within the same page or a different one.
```

apps/website/src/pages/docs/anchor.mdx
mayank99: Definitely, there is a lot to document even for a very simple component like this one (for instance, anchors can link to not just pages but also to files). If you would like to further expand on the description, feel free to do so before the paragraph you just added.

PR #1098, opened 2023-02-24
feat: Links to github on doc website open a new tab

elephantcatdog:
## Changes
Added `target='_blank' to force the links to open a new tab. This will benefit UXers that are updating the doc pages so that they can reference the current page easily, as the preview on Github is not very good.

## Testing
I clicked the links and they opened new tabs.

## Docs
N/A


mayank99:
Automatically opening links in new tab is bad from a usability/accessibility perspective. It breaks "back" button, it's jarring on mobile, and it offers no feedback for screen-readers.

If the user wants to open in a new tab, they can currently right click / middle click the link (or tap and hold on touch devices). It's better this way because they have the choice.

elephantcatdog:
Ah, I see. Thanks for the info.

Issue #961, opened 2023-01-06
Website: Review documentation of each component


FlyersPh9:
In prep for the new documentation site going live, we need to review the content that we have so far.  Here is what we need to do:

- Review the text that was copied over from the internal documentation site.
	- Only document what a component is capable of doing now, remove any documentation on features that we do not yet support.
	- Some documentation mentions restrictions that iTwinUI does not enforce. Reword this from a restriction to a suggestion. Example: "A button can only have 20 characters" 👉 "We suggest a button only have 20 characters"
	- If a component page mentions another component, the first instance of that other component being mentioned should link to that page. Example: [see tags link on line 20 of badges PR](https://github.com/iTwin/iTwinUI/blob/a98c981bfaab77b8e49cc620c0d8b03d274760f1/apps/website/src/pages/docs/badge.mdx?plain=1#L20).
	- Remove any mention of visual composition as the design system handles this.  Example of text that could be removed: "A button is 36px tall and has 4px rounded corners"
	- Consistent `H2` sections. As of right now, I think we should aim for each page to have Appearance, Usage, & Props.
- Introduce more demos.
	- Each page should have a generic top of page demo.
	- Each section under ~~Appearance~~ Variants should have a demo.
	- Each section should be ordered:
		1. Title
		3. Paragraph(s)
		2. Demo
		5. Paragraph
		6. Demo
		7. etc
	- Demos should be named like `exampleCodeCTA: Button.CTA.tsx`, do not use generic names like `exampleCode1: Button.1.tsx`.

To do:

Some of the components which don't have more than one PR needs review.

- [ ] Variables
    - [x] #1334
    - [ ] Rework variables page to show all themes
- [ ] Components
	- [x] Alert #1656 
	- [x] Anchor #1206 (add this: https://github.com/iTwin/iTwinUI/issues/961#issuecomment-1954478799 )
	- [x] Avatar #963 #1667 
	- [x] Badge #956 
	- [x] Blockquote #1260 
	- [x] Breadcrumbs #957
	- [x] Button #1158
	- [x] Button group #1162
	- [x] Carousel #1165 #1675 
	- [x] Checkbox #1160
	- [x]  Code #1766 
	- [x] Color picker #1189
	- [x] ComboBox
	- [x] DatePicker #1326
	- [x] Dialog
	- [x] Dropdown menu #1207 
	- [x] Expandable block #1184 
	- [x] Fieldset #1168
	- [x] File upload #1552 
[...]

PR #1099, opened 2023-02-24
feat: Temporary placeholders for unfinished pages on doc website

elephantcatdog:
## Changes
I thought it would be helpful to have placeholders for unfinished pages on the doc website so that people going here or being redirected here from ux.bentley.com aren't only getting a bunch of mostly empty pages with little information on them.

I want to add it to all the unfinished pages, but wanted to see if this format/wording is good before copy/pasting to all the other pages. I added the lines before and after so that it's a bit more noticeable. Also I'm not sure if it's better to have the full link be visible vs just having 'storybook' with a link.

image

## Testing
N/A

## Docs
N/A


gretanausedaite (commented):
Very good idea!
Maybe we could add some kind of banner/alert instead of simple text? You could create a component and pass down all mdx pages that need this info. Also, you could create new styling for this component, it doesn't have to be iTwinUI - you can use our theme gradients or something.

mayank99:
Agreed with @gretanausedaite

elephantcatdog:
I'm not sure why, but the grid template isn't working correctly for the alert I added: 
image

This was supposed to be a quick addition, so I didn't make the alert look fancy or anything.

gretanausedaite (commented):
Grid does not work because Alert does not get "placeholder" class. 
Add classname prop (eg. Card.astro) or move styles to Placeholder.astro.
Also I think it'd look better if alert had space before title (bottom padding) (eg. var(--space-5)). 

You sure you don't want to play around with styles? 
image

apps/website/src/components/Placeholder.astro
```diff
@@ -0,0 +1,19 @@
+---
+import { SvgInfoCircular } from '@itwin/itwinui-icons-react';
+---
+
+<div class='iui-alert' data-iui-status='informational'>
+  <SvgInfoCircular className='iui-alert-icon' aria-hidden='true' />
+  <span class='iui-alert-message'
+    >Some pages have not yet been finished and are being worked on. In the meantime, visit <a
+      class='iui-alert-link'
+      href='https://itwin.github.io/iTwinUI/react/?path=/story/overview--page'>storybook</a
```
gretanausedaite, line 10: It would be cool if we routed users to same component in storybook. Can be added later.

mayank99:
I think it might be better to control this message at a page-level rather than adding to the grid on all pages.

This will also allow customizing the message/link for each page.

The implementation could look like one of these two options:

 - A simple component that takes the component name (or url) as prop.
   ```mdx
   ---
   title: Alert
   ...
   ---
   ...
   <PlaceholderMessage component='alert' />
   ```

 - Alternatively, use a frontmatter property to indicate that the page should have a message.
   ```mdx
   ---
   title: Alert
   ...
   wip: true
   ---
   ...
   ```
   After this, you'll be able to access this property (among other properties such as `title`) in the [layout](https://docs.astro.build/en/core-concepts/layouts/#markdownmdx-layouts) and decide what to show.

elephantcatdog:
@mayank @gretanausedaite 

Thank you both for your input! I've updated the placeholder. Now it's got a property for the end of the storybook url, and is placed on an individual component's page vs the layout. I also updated the visuals to something similar to what you showed, Greta, because I thought that looked nice.

If this looks good, then I'll put the placeholder on all the pages that need it. I have it on 'Avatar' and 'Carousel' right now (and will remove it from 'Avatar').

image

apps/website/src/components/Placeholder.astro
```diff
@@ -0,0 +1,39 @@
+---
+import { SvgInfoCircular } from '@itwin/itwinui-icons-react';
+
+export type Props = {
+  /** ending of the url for a specific storybook page */
+  componentPageName: string;
+};
+
+const { componentPageName } = Astro.props;
```
mayank99, line 9: you might want to take the rest props and add them to the root element
```
const { componentPageName, ...rest } = Astro.props;
---
<div {...rest}>
```

apps/website/src/components/Placeholder.astro
```diff
@@ -0,0 +1,39 @@
...
+  <span class='iui-alert-message'
+    >This page has not yet been finished and is being worked on. In the meantime, visit <a
+      class='iui-alert-link'
+      href=`https://itwin.github.io/iTwinUI/react/?path=/story/${componentPageName}`>storybook</a
+    > to see more examples.
+  </span>
+</div>
+
+<style lang='scss'>
+  .alert {
```
mayank99, line 23: styles are automatically scoped to the component, so you can just use `div` here instead of `.alert`

apps/website/src/components/Placeholder.astro
```diff
@@ -0,0 +1,39 @@
...
+  /** ending of the url for a specific storybook page */
+  componentPageName: string;
+};
+
+const { componentPageName } = Astro.props;
+---
+
+<div class='alert' data-iui-status='informational'>
+  <SvgInfoCircular className='iui-alert-icon' aria-hidden='true' />
+  <span class='iui-alert-message'
```
mayank99, line 14: we are trying to maintain a distinct design language for the website, so let's not use anything from iTwinUI (no `iui-` classes / itwinui-icons).

you don't need to make anything fancy, but would be nice if you can use spacing/color values from `__global.astro` (line 116 onwards).

for icons, you can grab something from https://icones.js.org/collection/all and either add it to the `website/src/icons/` folder or inline it in CSS.

mayank99: here is an example implementation you can build upon:
```astro
---
export type Props = {
  /** ending of the url for a specific storybook page */
  componentPageName: string;
};

const { componentPageName, ...rest } = Astro.props;

const storyUrl = `https://itwin.github.io/iTwinUI/react/?path=/story/${componentPageName}`;
---

<div {...rest}>
  <p>
    This page has not yet been finished and is being worked on. In the meantime, please visit
    <a href={storyUrl}>storybook</a> to see more examples.
  </p>
</div>

<style lang='scss'>
  div {
    border-block: 1px solid var(--color-line-1);
    padding-block: var(--space-3);
    display: flex;
    align-items: center;
    gap: var(--space-2);

    &::before {
      content: '';
      flex-shrink: 0;
      display: inline-block;
      width: 1.5em;
      height: 1.5em;

      --_svg: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 24 24"><path fill="currentColor" d="M13 9a1 1 0 1 1-2 0a1 1 0 0 1 2 0Zm-1 2.75a.75.75 0 0 1 .75.75v5a.75.75 0 1 1-1.5 0v-5a.75.75 0 0 1 .75-.75Z"/><path fill="currentColor" fill-rule="evenodd" d="M14.27 3.993a2.749 2.749 0 0 0-4.54 0l-.432.632a75.95 75.95 0 0 0-6.944 12.563l-.09.208a2.511 2.511 0 0 0 2.024 3.497a69.43 69.43 0 0 0 15.424 0a2.511 2.511 0 0 0 2.024-3.497l-.09-.208a75.951 75.951 0 0 0-6.944-12.563l-.432-.632Zm-3.302.846a1.25 1.25 0 0 1 2.064 0l.432.632a74.444 74.444 0 0 1 6.806 12.315l.09.208a1.011 1.011 0 0 1-.814 1.408c-5.015.56-10.077.56-15.092 0a1.011 1.011 0 0 1-.815-1.408l.09-.208a74.45 74.45 0 0 1 6.807-12.315l.432-.632Z" clip-rule="evenodd"/></svg>');
      -webkit-mask-image: var(--_svg);
      mask: var(--_svg) no-repeat center / contain;
      background-color: currentColor;
    }
  }
</style>

```

mayank99 (commented):
usage looks good so i think it can be added to all pages as-is, but i think the styling could be toned down to be less distracting.

apps/website/src/components/Placeholder.astro
elephantcatdog: Updated

apps/website/src/components/Placeholder.astro
elephantcatdog: Updated

elephantcatdog:
image

apps/website/src/components/Placeholder.astro
elephantcatdog: Updated

apps/website/src/components/Placeholder.astro
```diff
@@ -0,0 +1,37 @@
...
+  div {
+    display: flex;
+    align-items: center;
+    gap: var(--space-3);
+    padding: var(--space-3);
+    border: solid 1px #adb7f5;
+    border-radius: var(--border-radius-1);
+
+    svg {
+      color: #adb7f5;
```
mayank99, line 32: can you use one of the color variables defined in `_global.astro` (or add a new one if you think it makes sense)?

veekeys:
> Updated look: image

Would it be possible to get rid of any CONNECT related references in public docs site? I do understand initially we might have copied some of the old content, but in this particular case (hopefully not that many in all docs site) it is something we do not want to even mention to anyone anymore.

mayank99:
> Would it be possible to get rid of any CONNECT related references in public docs site? I do understand initially we might have copied some of the old content, but in this particular case (hopefully not that many in all docs site) it is something we do not want to even mention to anyone anymore.

Agreed we should remove that.

Almost every single page had content copied over from the old docs, so there might be more weird stuff. #961 is tracking the status of pages that have been further reviewed/adjusted after the old content was moved.

FlyersPh9 (approved):
Once [this comment](https://github.com/iTwin/iTwinUI/pull/1099/files#r1128205630) is resolved, looks good to me!

elephantcatdog:
Okay, I added the placeholder to all the unfinished pages, and confirmed that the links go where they're supposed to. I have a question about one of the pages though - 'Typography' is a page that has 'text' and 'label' on it, but we have separate pages for the other types of typography. Why does this page exist as it is, and not either include all types of typography or have separate pages for all?

apps/website/src/components/Placeholder.astro
elephantcatdog: Updated

gretanausedaite:
> I have a question about one of the pages though - 'Typography' is a page that has 'text' and 'label' on it, but we have separate pages for the other types of typography. Why does this page exist as it is, and not either include all types of typography or have separate pages for all?

What are "other types of typography"?

mayank99 (approved):

elephantcatdog:
> What are "other types of typography"?

These are all listed under "Typography" in storybook:
image

gretanausedaite:
> These are all listed under "Typography" in storybook: image

They are under "typography" in side nav.
image

We've just combined text and label because it would be hard to separate them visually in all components page
image

elephantcatdog:
Hm, I see, thanks Greta

PR #1106, opened 2023-03-01
List component

mayank99:
## Changes

- Added `iui-list` and `iui-list-item` classes that share base styling with `iui-menu` and `iui-menu-item`respectively.
- Uses data attributes and respective props for controlling actionable/active/disabled/focused/size attributes.
- Moved a lot of other styles out into separate mixins, because their class name structure is not great.

## Testing

Unit tests added. New stories and visual tests added.

Also tested manually in playground.

## Docs

Added changeset. Needs some more docs/examples probably.


mayank99:
This PR is still in very early stages, looking for feedback on what separates it from MenuItem, what should be reused between this and Menu/Listbox, how much we should tackle in initial release, etc.

gretanausedaite:
> This PR is still in very early stages, looking for feedback on what separates it from MenuItem, what should be reused between this and Menu/Listbox, how much we should tackle in initial release, etc.

In my mind menuItem is clickable, selectable, etc. Users can do actions with it.
ListItem displays information, maybe has hover/focus state, users can reorder listItems or drag to another list (eg. from one listBox to another). I like how chakra-ui separates them.

In my opinion initial List should not have any styling (background color, drop shadow) - we have ListBox (or Surface) for that.

mayank99:
> In my mind menuItem is clickable, selectable, etc. Users can do actions with it.
ListItem displays information, maybe has hover/focus state, users can reorder listItems or drag to another list (eg. from one listBox to another). I like how chakra-ui separates them.

@gretanausedaite What do you think of having a prop in `ListItem` that adds hover/focus states? So far I've added `active` and `disabled` props which add styling for those things.

Eventually I'd like for `MenuItem` to internally use `ListItem`, but we can't do that right now as it would be a bit of a breaking change.

> In my opinion initial List should not have any styling (background color, drop shadow) - we have ListBox (or Surface) for that.

Then do we even need a `List` component? The important part there is that it removes default browser list styling (`list-style-type`), but that's only present for `<ul>`/`<ol>` whereas currently I'm using `<div role='list'>`.

gretanausedaite:
> @gretanausedaite What do you think of having a prop in `ListItem` that adds hover/focus states? 

I think something like `actionable` prop would work.

> Then do we even need a `List` component? The important part there is that it removes default browser list styling (`list-style-type`), but that's only present for `<ul>`/`<ol>` whereas currently I'm using `<div role='list'>`.

So with transfer list (ListBox) component in mind, we would have something like that? 
```typescript
<TransferList>
  <List 1>
    <ListItem />
    <ListItem />
    <ListItem />
   </List>
   <List 2>
     <ListItem />
     <ListItem />
     <ListItem />
   </List>
</TransferList>
```

Or use ListItems directly? 

Maybe having List component is useful after all.

mayank99:
I've kept the List component but removed surface styles from it. I've also added `actionable` and `focused` props.

I'd imagine a `TransferList` to have two separate `ListBox`es, and every item within it would need `role='option'` (which is not the default role for a listitem), so I'd imagine we'll need a `ListBox.Item` component on top of `ListItem` as well.

packages/itwinui-react/src/core/List/ListItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+  useTheme();
+
+  return (
+    <Element
+      className={cx('iui-list-item', className)}
+      data-iui-active={active ? 'true' : undefined}
+      data-iui-disabled={disabled ? 'true' : undefined}
+      data-iui-size={size === 'large' ? 'large' : undefined}
+      data-iui-actionable={actionable ? 'true' : undefined}
+      data-iui-focused={focused ? 'true' : undefined}
```
gretanausedaite, line 35: Does users need to add their own focus functionality?

gretanausedaite (commented):
Looks good.
We also need website docs for this. Are you planning to add them here or in different PR?

packages/itwinui-react/src/core/List/ListItem.tsx
mayank99: This is for custom focus management (like we do in Select/Combobox).

In the css I have other selectors too, so if the listitem is focusable or contains a LinkAction, then it will have focus styling.

packages/itwinui-react/src/core/List/List.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  );
+  expect(screen.getByRole('list')).toHaveClass('iui-list');
+  const items = container.querySelectorAll('.iui-list-item');
+  expect(items).toHaveLength(3);
+  items.forEach((item, index) => {
+    expect(item).toHaveTextContent(`item ${index + 1}`);
+  });
+});
+
+describe('ListItem', () => {
```
elephantcatdog, line 27: It'll probably be easier for the future if the 'ListItem' tests are in their own file.

apps/storybook/src/List.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </ListItem>
+      <ListItem {...args} disabled>
+        Item 3
+      </ListItem>
+    </List>
+  );
+};
+Actionable.args = {
+  actionable: true,
+};
```
elephantcatdog, line 99: Maybe add an argTypes here so that the actionable example can't become unactionable?
`Actionable.argTypes = {
  actionable: { control: { disable: true } },
};`

elephantcatdog (commented):
Seems like you're aware of it, but the new components and props need comments/descriptions.

packages/itwinui-react/src/core/List/List.test.tsx
mayank99: moved

apps/storybook/src/List.stories.tsx
mayank99: done

packages/itwinui-react/src/core/List/ListItem.tsx
gretanausedaite: Should actionable list items be automatically focusable?

mayank99: What does "automatically focusable" mean?

gretanausedaite: Right now I can't tab through actionable list items. Or did you do this on purpose for users to add all functionality themselves? (Comparing to MenuItem)

mayank99: We can't simply add `tabindex` as it is not enough by itself. Listitems are generally meant for displaying data rather than for interaction. When it needs to be interactable, that usually happens through either a child element (e.g. a link) or entirely custom focus and role management (e.g. MenuItem has `role=option`, roving tabindex, and arrow key handling).

In a way, you can look at `ListItem` as a low-level primitive for more advanced interactive cases. I expect we will use it inside `Listbox` and in the future also inside `Menu`/`Select`/`ComboBox`.

Maybe the name "actionable" is misleading then. But I did mention in the jsdoc that it is only for styling. And I will have more details in the documentation page as well.

gretanausedaite (approved):
Documentation promised to be done in different PR.

mayank99:
Blocked. Waiting for #1197 to be merged and released first because that is a critical fix that can go in a patch.

PR #1109, opened 2023-03-02
add `as` prop to HeaderLogo

### LINKED ISSUE #920: HeaderLogo hover state is applicable only when onClick is defined
On our page we use links for navigation, so we have no point of defining onClick for HeaderLogo, but w/o it HeaderLogo has no hover indication. Maybe there might be another option having HeaderLogo hover indication w/o adding hacky/useless onClick? 

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/itwinui-react-header-logo-hover-state-issue-57yjb

@veekeys might add additional info if needed.

mayank99:
## Changes

Closes #920

`as` prop will allow rendering the HeaderLogo as a link, with correct hover states. Also removed some extra code that is not necessary - previously we were always rendering a `<div>` and doing extra work to turn it into button when `onClick` was passed.

The `as` prop will now default to `div` when no `onClick` passed, and `button` when one is passed.

## Testing

Updated existing unit tests and added new unit test for `as` prop.

Also inspected dom in stories/playground and it was rendering the correct element.

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
+++ b/packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
@@ -57,7 +57,7 @@ it('renders with as prop correctly', () => {
       hello
     </HeaderLogo>,
   );
-  const link = container.querySelector('h1');
+  const link = container.querySelector('a');
   expect(link).toHaveClass('iui-header-brand');
   expect(link).toHaveAttribute('href', 'https://www.example.com/');
   link?.click();
--- a/packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
+++ b/packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { SvgMore as SvgPlaceholder } from '../utils';
 
 import HeaderLogo from './HeaderLogo';
@@ -13,9 +13,9 @@ it('renders default correctly', () => {
     <HeaderLogo logo={<SvgPlaceholder />}>Application</HeaderLogo>,
   );
 
-  const root = container.querySelector('.iui-header-brand') as HTMLDivElement;
+  const root = container.querySelector('div') as HTMLDivElement;
   expect(root).toBeTruthy();
-  expect(root.getAttribute('role')).toBeNull();
+  expect(root).toHaveClass('iui-header-brand');
 
   const {
     container: { firstChild: placeholderIcon },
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
```diff
@@ -38,34 +38,30 @@ it('renders with onClick correctly', () => {
...
 });
 
-it('handles keypress with onClick correctly', () => {
+it('renders with as prop correctly', () => {
   const onClickMock = jest.fn();
   const { container } = render(
-    <HeaderLogo logo={<SvgPlaceholder />} onClick={onClickMock}>
-      Application
+    <HeaderLogo
+      as='a'
```
gretanausedaite, line 52: In this case clicking on item opens a link or does onClick action? Or both?

packages/itwinui-react/src/core/Header/HeaderLogo.tsx
```diff
@@ -33,25 +38,25 @@ export type HeaderLogoProps = {
  * <HeaderLogo logo={<img src='image.png' />} />
  * <HeaderLogo logo={<img src='data:image/png;base64,...' />}>Downloaded Image</HeaderLogo>
  */
-export const HeaderLogo = (props: HeaderLogoProps) => {
-  const { className, children, logo, onClick, ...rest } = props;
-  const keyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
-    if (
-      onClick &&
-      (e.key === 'Enter' || e.key === ' ' || e.key === 'Spacebar')
```
gretanausedaite, line 41: Why removed keyboard functionality?

packages/itwinui-react/src/core/Header/HeaderLogo.tsx
```diff
@@ -6,24 +6,29 @@
...
-  /**
-   * Expects the app name, is put on the right of the logo.
-   */
-  children?: React.ReactNode;
-} & CommonProps;
+  onClick?: (e: unknown) => void;
+};
+
+export type HeaderLogoProps = PolymorphicComponentProps<
+  'div',
```
gretanausedaite, line 29: This should break if HeaderLogo is rendered as button?

packages/itwinui-react/src/core/Header/HeaderLogo.test.tsx
mayank99: both. `onClick` is in addition to the default action

packages/itwinui-react/src/core/Header/HeaderLogo.tsx
mayank99: because it is all handled automatically by `<button>`

mayank99: from PR description:
> removed some extra code that is not necessary - previously we were always rendering a <div> and doing extra work to turn it into button when onClick was passed.

you can try it in [storybook](https://itwin.github.io/iTwinUI/1109/react/?path=/story/core-header--full0. inspect the logo to see that it is a `<button>`, then try tabbing to it and press Enter or Space to see logs in Actions tab.

packages/itwinui-react/src/core/Header/HeaderLogo.tsx
```diff
@@ -60,8 +65,8 @@ export const HeaderLogo = (props: HeaderLogoProps) => {
           })
         : undefined}
       {children && <span className='iui-header-brand-label'>{children}</span>}
-    </div>
+    </Element>
   );
-};
+}) as PolymorphicForwardRefComponent<'div', HeaderLogoOwnProps>;
```
LostABike, line 70: What does this mean or do? Why does it have type 'div' in there?

gretanausedaite (approved):

packages/itwinui-react/src/core/Header/HeaderLogo.tsx
mayank99: You can check out the comments at: https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/utils/props.ts#L42-L53

It's advanced typescript stuff and it's a bit hard to explain tbh. It is a generic type (think of it like a function) which takes two parameters - the first param is the default element name (so we want it to have all `'div'` props by default) and the second param is the component's own props (e.g. `logo`) to be added to the div props.

The result of this is that you get correctly typed and autocompletable attributes. For example, when I use `as='a'`, it automatically knows about [`href`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href).




You can also ignore all the typescript stuff and review only the javascript parts - the code is identical to what I suggested in https://github.com/iTwin/iTwinUI/issues/920#issuecomment-1362952436.

LostABike: Thanks. That's helpful

LostABike (approved):

PR #1110, opened 2023-03-02
update tile thumbnail hover effect for svgs

### LINKED ISSUE #910: Tile: `isActionable` does not scale svg
When Tile has isActionable set, it does not scale svg type (iTwinUI v1.37.1)
image

mayank99:
## Changes

Closes #910

I tried making svgs scale, but it was causing unpleasant blurriness during the transition, so I instead went with a rotate effect.

I also increased the intensity of the shadow on the tile itself but it was not very noticeable, so I then added a translate effect to the whole tile.

## Testing

No new tests needed.

https://user-images.githubusercontent.com/9084735/222790186-e900013f-78a1-49c8-bf81-468d424e7c85.mov

https://user-images.githubusercontent.com/9084735/222789831-d5296869-6b61-492b-8e64-6d3448759968.mov

## Docs

Added changeset.


mayank99:
@FlyersPh9 what do you think of spicing up the hover effect a little more?

https://user-images.githubusercontent.com/9084735/222568618-3782c419-16a6-4ef9-9a87-18615cb1f3bc.mov

FlyersPh9:
> @FlyersPh9 what do you think of spicing up the hover effect a little more?

As long as it doesn't interfere with other tiles when placed in [the iTwinUI-layouts fluid grid](https://itwin.github.io/iTwinUI-layouts/#/fluid-grid), I'm open to it.  Let's look at it together tomorrow in a call.

mayank99:
To make it less jarring, I've decreased the distance and added a 50ms delay.

https://user-images.githubusercontent.com/9084735/222783071-397141d5-062e-476b-8228-e1b8418b3e48.mov


Try it: https://itwin.github.io/iTwinUI/1110/css/tile.html

FlyersPh9:
When hovering a tile, the thumbnail appears to visually glitch out around the edges of the thumbnail area / tile.
https://user-images.githubusercontent.com/849817/222796920-4c9df753-f87f-47ea-93a9-003509df0307.mov

This is also present on our current tiles, so it's not a regression.  Want to think about an alternative for that as well?

mayank99:
@FlyersPh9 i've replaced `absolute` positioning with `display: grid` and i think that fixed both the glitching and the blurriness. here's the commit: [`e50b62f`](https://github.com/iTwin/iTwinUI/pull/1110/commits/e50b62fbe4d156f070cdeef81dd50ebb5565d4ef). i can move it to a separate PR if it's too much.

mayank99:
there is another small problem - when hovering near the bottom edge of the tile, the transition will trigger and then untrigger because mouse is no longer hovering the tile.

https://user-images.githubusercontent.com/9084735/223181716-30976bb4-58d6-468c-9ea3-9a70e0d731d8.mov

i can fix this using an absolutely positioned pseudo-element (see my [codepen](https://codepen.io/MrRoboto/pen/vYjPzvP)) but that will require changing some other code to use grid, so i will create a separate PR with that and [`e50b62f`](https://github.com/iTwin/iTwinUI/pull/1110/commits/e50b62fbe4d156f070cdeef81dd50ebb5565d4ef). as an added benefit, it will also unblock https://github.com/iTwin/iTwinUI/pull/959#discussion_r1113147711.

FlyersPh9 (approved):
Really like the hover effect where the tile moves up.  The icon tilt, I'm ok with.

mayank99:
this PR is good to go now. the psuedo-element trick i mentioned above has been added.

PR #1112, opened 2023-03-03
Add a hook for `iui-animated` in Tab

### LINKED ISSUE #911: Tabs: inconsistent stripe width on first mount
On the _first visit_ to a page that uses our borderless and pill tabs, I noticed an inconsistency in the stripe width. When I visit the page, the stripe width looks wrong. When I refresh it, it corrects itself.

In borderless and pill tabs, we rely on client-side JavaScript to set the width and position of tabs. There is no way around that, which is fine.

This also causes some flakiness in our react visual tests as the screenshot is taken before the transition completes.

This could possibly be because in our CSS we apply a transition on the `width` and `left` properties, which causes a bit of delay/content shift after the JavaScript runs.

https://github.com/iTwin/iTwinUI/blob/6c94ffbb7e93049dc6d1c81049ddf55e57418c68/src/tabs/tabs.scss#L202-L204

Further investigation is needed. This issue can be moved to iTwinUi-react if the fix is entirely JavaScript.

LostABike:
## Changes

Added a hook `useIsClient`  in util and used in `Tabs.tsx` to change `iui-not-animated` to `iui-animated` after first mount.

Closes #911

## Testing

- Vigorously refreshing Storybook's tabs to see if the change fixed the issue 🤣 
- Deployed in Nextjs playground for visual testing

## Docs

- Added changeset

N/A


packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -208,6 +208,11 @@ export const Tabs = (props: TabsProps) => {
     }
   }, [focusedIndex]);
 
+  const [isAnimated, setIsAnimated] = React.useState(false);
+  React.useEffect(() => {
+    setIsAnimated(true);
+  }, []);
```
mayank99, line 214: This can actually be a generic hook called `useIsClient`.

```js
const useIsClient = () => {
  const [isClient, setIsClient] = React.useState(false);

  React.useEffect(() => {
    setIsClient(true);
  }, []);

  return isClient;
};
```

you can add it under `utils/hooks/` and then use it like this:
```js
const isClient = useIsClient();
```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -299,7 +304,8 @@ export const Tabs = (props: TabsProps) => {
           `iui-${type}`,
           {
             'iui-green': color === 'green',
-            'iui-animated': type !== 'default',
+            [isAnimated ? 'iui-animated' : 'iui-not-animated']:
+              type !== 'default',
```
mayank99, line 308: this can be simplified like this (make sure to test it first):
```suggestion
            'iui-animated': type !== 'default' && isClient,
            'iui-not-animated': type !== 'default' && !isClient,
```

mayank99 (commented):
For testing these changes, I would suggest adding the code to nextjs playground and then running `yarn build --filter=next-playground && cd playgrounds/next && yarn start`. You will likely need to stop your dev server first.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
LostABike: Thanks for the suggestion. I was stuck for a while on this thinking how to not make it look so tacky 🤣 Adjusted

packages/itwinui-react/src/core/Tabs/Tabs.tsx
LostABike: Interesting. I have added this hook into util and use it in Tabs based on your suggestion. I have also added the nextjs playground code for testing. In this playground, I don't see the bug anymore. But I still see it in Storybook I feel like. It's so subtle so really hard to tell O_O

LostABike: I left the playground code too for reviewers to test. But will try to remember to remove it 🤔

mayank99: remove it and add changeset. otherwise this PR looks good

LostABike: Thanks. Done.

mayank99 (approved):

adamhock (approved):
LGTM 🎉

PR #1114, opened 2023-03-03
Annie/modal display overflow

LostABike:
## Changes

** Don't review this yet, I'm making a PR to test on my phone atm.


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -171,7 +171,7 @@ $iui-dialogue-resizable-min-height: 7.75rem; // 7.75rem = 124px = title bar heig
   padding: 0 var(--iui-size-m);
   overflow-y: auto;
   overflow-y: overlay;
-  max-height: calc(100vh - 6.75rem); // 6.75rem = 108px = button height + paddings + bottom bar
+  max-height: calc(100svh - 6.75rem); // 6.75rem = 108px = button height + paddings + bottom bar
```
mayank99, line 174: `svh` was added pretty recently to browsers: https://caniuse.com/viewport-unit-variants

so you could add a `vh` fallback like this:
```
max-height: calc(100vh - 6.75rem);
max-height: calc(100svh - 6.75rem);
```

in browsers that support `svh`, the last declaration of `max-height` will be used. in browsers that don't, the previous declaration will be used.

apps/storybook/src/Dialog.stories.tsx
```diff
@@ -120,7 +120,98 @@ export const Modal: Story = () => {
             aliquip ex ea commodo consequat. Duis aute irure dolor in
             reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
             pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
-            culpa qui officia deserunt mollit anim id est laborum.
+            culpa qui officia deserunt mollit anim id est laborum. Lorem ipsum
```
mayank99, line 123: you'll need to update visual test images if youre changing the story

packages/itwinui-css/src/dialog/dialog.scss
LostABike: Thanks for the info. I updated this.

apps/storybook/src/Dialog.stories.tsx
LostABike: I have reverted the change altogether

packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -171,7 +171,8 @@ $iui-dialogue-resizable-min-height: 7.75rem; // 7.75rem = 124px = title bar heig
   padding: 0 var(--iui-size-m);
   overflow-y: auto;
   overflow-y: overlay;
-  max-height: calc(100vh - 6.75rem); // 6.75rem = 108px = button height + paddings + bottom bar
+  max-height: calc(100vh - 6.75rem); //6.75rem = 108px = button height + paddings + bottom bar
+  max-height: calc(100svh - 6.75rem);
```
mayank99, line 175: actually i think `dvh` will be better than `svh` because it will always be correct regardless of whether the browser ui is present

.changeset/pretty-zebras-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+The different density settings for table will now also affect horizontal spacing.
```
mayank99, line 5: this changeset is wrong

packages/itwinui-css/src/dialog/dialog.scss
LostABike: Do you mean that we should use only `dvh`? Or changing `svh` to `dvh` and still keep the `vh`?

mayank99: the latter.

we need the fallback because `dvh` has same browser support as `svh`. https://caniuse.com/viewport-unit-variants

LostABike: Thanks, updated. Also tested the change like you suggested and it seems to be working. I don't see that bug anymore.

.changeset/pretty-zebras-pull.md
LostABike: Oops. Removed and added the right changeset.

mayank99 (approved):
🚀

gretanausedaite (approved):

mayank99:
> This is what I see in iOS simulator using this link: [rq3h6g.csb.app](https://rq3h6g.csb.app/)

That is the repro link from the issue, so of course it is broken. Are you able pull this branch and open localhost in that simulator?

FlyersPh9:
> > This is what I see in iOS simulator using this link: [rq3h6g.csb.app](https://rq3h6g.csb.app/)
> 
> That is the repro link from the issue, so of course it is broken. Are you able pull this branch and open localhost in that simulator?

My mistake, pulled the branch and default dialog in HTML demo is looking correct!
Screenshot 2023-03-08 at 1 51 23 PM

FlyersPh9 (approved):

mayank99:
Awesome! Lets merge it then

Issue #1115, opened 2023-03-03
ComboBox with Multiple Select crashes if value given is not in the list of options


orlandoortegajr:
#### Describe the bug (actual behavior)
When a value is passed to the component with the "multiple" prop set to true, if this value is not within the list of options, the component will crash.

#### Expected behavior
The component does not crash and instead should not show the option selected. This would be particularly useful in cases were you have the value available on initial render, but you have to make a request to an API to retrieve the list of options asynchronously, once you get the full list of options and the state variable holding these options is updated, the component should then show the initial value given as selected (as long as it is within the updated list of options).

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/dreamy-http-dn5zow?file=/src/App.tsx

##### Steps to reproduce
1. Create a state variable that will hold the values to be passed to the ComboBox.
2. Create a state variable that will hold a list of options, these options should not include the option assigned to the value state variable.
3. Pass them to the ComboBox component.
4. Set the multiple prop for the ComboBox to true.
5. Run it and observe it crash

gretanausedaite:
Hey! Thanks for the bug report.
We fixed this issue with #1116 and it should be available in the next release.

orlandoortegajr:
Awesome, thanks!!

PR #1116, opened 2023-03-06
Update combobox to not crash

gretanausedaite:
Solves #1115 

## Changes

Updated `getSelectedIndexes` to check if user provided index is valid (exists in options).

## Testing

To test this copy-paste user provided example and run our test app.
Added unit test for this case.

## Docs

N/A


mayank99 (approved):
add changeset

gretanausedaite:
> add changeset

Oops, added

PR #1117, opened 2023-03-06
use grid layout in Tile

mayank99:
## Changes

Extracted out of #1110. Tile now uses `display: grid` instead of `position: relative`/`position: absolute` for all the overlapping parts, including:
- thumbnail (both picture and svg)
- quick actions (buttons on the top left/right)
- badges
- more-options button

This fixes some weird visual glitches in thumbnail (when hovering) and also unblocks https://github.com/iTwin/iTwinUI/pull/1110#issuecomment-1456559935 and https://github.com/iTwin/iTwinUI/pull/959#discussion_r1113147711.

## Testing

No new tests needed.

Compare the difference in the [live version](https://itwin.github.io/iTwinUI/css/tile.html) vs [this PR](https://itwin.github.io/iTwinUI/1117/css/tile.html) (most noticeable in chrome).

Other than the hover effect, there is a small visual difference in the positioning of the more-options button. I think it's better aligned now, but that's probably subjective.
| Before | After |
| --- | --- |
| image | image |

## Docs

N/A. Added changeset.


FlyersPh9 (approved):
Much better than before, nice fix!

gretanausedaite (approved):

PR #1120, opened 2023-03-07
Make icon not appear when size is small for toggle switch in React

LostABike:
## Changes

@gretanausedaite brought up making this change per Vilius suggestion in the call
image
And recommended that I follow example in `Tabs.tsx` 

So I split `ToggleSwitchProps` into two with the new type only consisting of `icon` and `size` prop. Makes `icon` not part of the prop if `size` is `small`. Also added condition to display icon to not have size small.

## Testing

- Tested different cases in Storybook: when icon is set and size is not set, or set to default or small. When icon is not set and size is default, not set, or small.
- Added unit test

## Docs

- Added changeset


veekeys:
I think `small` size of `ToggleSwitch` is already released. If it is, technically this is a breaking change :)

gretanausedaite:
> I think `small` size of `ToggleSwitch` is already released. If it is, technically this is a breaking change :)

It is released 15 hours ago (yesterday). Does that really count? 😄 

To not "break" the API, we could keep props as is and add condition on displaying icon. (Keeping the description on icon prop that it is not shown when size is small).

```ts
icon && size !== 'small'
```

mayank99:
> technically this is a breaking change

Yes, but almost every change, no matter how small, has the potential to break something. I would consider this to be a correction in the API. or an improvement in the types.

> add condition on displaying icon.

We could do both.

---

> not sure how to unit test this. Because I can't set icon when size is set to small. 

@LostABike That _is_ kinda the test (the fact that TS errors). You can simply try to render that and use `// @ts-expect-error`.

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -53,10 +61,15 @@ export type ToggleSwitchProps = {
  */
 export const ToggleSwitch = React.forwardRef(
   (props: ToggleSwitchProps, ref: React.RefObject<HTMLInputElement>) => {
+    let icon: JSX.Element | undefined;
+    if (props.size !== 'small') {
+      icon = props.icon;
+      props = { ...props };
+      delete props.icon;
+    }
```
mayank99, line 69: i wonder if we should make a utility function to hide this mess

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -7,6 +7,22 @@ import cx from 'classnames';
 import { useMergedRefs, useTheme } from '../utils';
 import '@itwin/itwinui-css/css/toggle-switch.css';
 
+type ToggleSwitchSizeProps =
```
mayank99, line 10: is this intermediate type necessary? could be inlined into `ToggleSwitchProps`

mayank99 (approved):
Functionality looks good.
Couple minor comments above improving readability

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
LostABike: I actually don't know. I was referring to this example in `Tabs.tsx` :
image
And the way I understand it (correct me if I'm wrong) is that this intermediate typing would separate icon from different sizing only. But if we combine it into one type, then "small" size will be separated from all the other props too not just icon. Actually let me test that rn 🤔

LostABike: ohh never mind, it works the same with inline typing! updated

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
gretanausedaite: Maybe we should. There are a lot of places where we do this. Maybe in next PR to not make unrelated changes in this one.

gretanausedaite (approved):
Looks good

Issue #1113, opened 2023-03-03
Carousel: Warning when navigating with arrow keys in React 16.X or below


6ar8nas:
#### Describe the bug (actual behavior)
When a carousel is being used with React 16.X or below, a warning after using arrow keys for navigation through the carousel is visible in the console.
![image](https://user-images.githubusercontent.com/86420873/222811693-6bb32733-86dd-4627-a2fd-85c81b15fe04.png)

#### Expected behavior
Warning in the console does not appear.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/angry-dew-1phjuu

##### Steps to reproduce
1. Hold right arrow key

#### Additional information
Looking into the warning, the potential issue is located at https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/Carousel/Carousel.tsx#L102, while someone's suggested solution to this can be found at https://stackoverflow.com/a/56629758.

gretanausedaite:
Hi!
Thank you for the bug report and initial investigation.

Would you be interested in contributing to our library?  
The best place to start is [Contributing guide](https://github.com/iTwin/iTwinUI/blob/main/CONTRIBUTING.md).

6ar8nas:
Sure, will further investigate it and try creating a pull request myself.

mayank99:
🚀 This is now available in 2.7.0

PR #1123, opened 2023-03-08
Fix carousel arrow navigation warning

6ar8nas:
## Changes

Pull request includes only the change suggested in the related issue (moving necessary event's information to a local variable and referencing that variable instead - https://stackoverflow.com/questions/49500255/warning-this-synthetic-event-is-reused-for-performance-reasons-happening-with/56629758#56629758), which was sufficient enough to solve the problem.
Fixes iTwin/iTwinUI#1113

## Testing

Probably not really applicable, because it was done manually, but the branch contains the changes that were done in the `vite` playground to make sure everything works as expected. 
<details>
  <summary>I first reproduced the issue locally</summary>
  
</details>
<details>
  <summary>and then fixed it with the following commit and reassured I can no longer reproduce it</summary>
  
</details>

## Docs

N/A


packages/itwinui-react/src/core/Carousel/Carousel.tsx
```diff
@@ -98,8 +98,9 @@ export const Carousel = Object.assign(
         return;
       }
 
-      if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
-        setKeysPressed((old) => ({ ...old, [event.key]: true }));
+      const { key } = event;
```
gretanausedaite, line 101: I think you could write this as:
```ts
const key = event.key;
```

We do this in more places than destructing event props.

gretanausedaite (approved):

mayank99 (approved):
Looks great, thanks

PR #1125, opened 2023-03-09
update Tile order so name comes first

mayank99:
## Changes

Before this change, when using arrow key navigation in screenreaders, all the content in the thumbnail area would be announced before the name of the Tile. This is not ideal.

So now, `iui-tile-name` will now be the first thing inside tile (in the DOM) with no changes to visuals. This is done using flex `order`.

I left the folder tile unchanged because it was getting complicated and it's less important - there isn't much stuff there in the thumbnail area.

## Testing

No new tests needed. Existing tests all pass because there are no visual changes.

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/Tile/Tile.tsx
+++ b/packages/itwinui-react/src/core/Tile/Tile.tsx
@@ -166,6 +166,19 @@ export const Tile = (props: TileProps) => {
   const showMenu = React.useCallback(() => setIsMenuVisible(true), []);
   const hideMenu = React.useCallback(() => setIsMenuVisible(false), []);
 
+  const tileName = (
+    <div className='iui-tile-name'>
+      <TitleIcon
+        isLoading={isLoading}
+        isSelected={isSelected}
+        isNew={isNew}
+        status={status}
+      />
+
+      <span className='iui-tile-name-label'>{name}</span>
+    </div>
+  );
+
   return (
     <div
       className={cx(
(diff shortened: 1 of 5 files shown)


packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -184,6 +197,8 @@ export const Tile = (props: TileProps) => {
       tabIndex={isActionable && !isDisabled ? 0 : undefined}
       {...rest}
     >
+      {variant !== 'folder' ? tileName : null}
```
adamhock, line 200: nit:

```jsx
{variant !== 'folder' && tileName}
```

adamhock (commented):
Should we do the same thing for tile descriptions?

packages/itwinui-react/src/core/Tile/Tile.tsx
mayank99: this was intentional because ternary is always clearer and there is no chance of accidentally rendering `false` as a string

adamhock (approved):

gretanausedaite (approved):
Looks good, a lot of css changes and only one `order` added.

mayank99:
> Looks good, a lot of css changes and only one `order` added.

There were a few more changes. You can "hide whitespace" in PR diff to see those more clearly. They were needed to preserve old spacing.

PR #1126, opened 2023-03-09
fix Table header misalignment in firefox

mayank99:
## Changes

Noticed an issue where the Table header was misaligned with Table body in firefox. We actually already had a workaround for Firefox in place, but it was only partially working because it was occupying invisible space. So I've updated it to have a background-color.

## Testing

Before this change, this is how it looks in Firefox:

image

After this change, this is how it looks in Firefox:
image

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-css/src/table/table.scss
+++ b/packages/itwinui-css/src/table/table.scss
@@ -15,7 +15,8 @@
   overflow: hidden;
   // Fix for Firefox columns misalignment
   @supports not (overflow: overlay) {
-    overflow-y: scroll;
+    scrollbar-gutter: stable;
+    background-color: var(--iui-color-background);
   }
   display: flex;
   flex-shrink: 0;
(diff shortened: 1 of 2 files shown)


packages/itwinui-css/src/table/table.scss
```diff
@@ -15,7 +15,8 @@
   overflow: hidden;
   // Fix for Firefox columns misalignment
   @supports not (overflow: overlay) {
-    overflow-y: scroll;
+    scrollbar-gutter: stable;
```
LostABike, line 18: Interesting. Does this affect behaviors in other browsers though?

mayank99: it's inside `@supports not (overflow: overlay)`. Chrome and Safari do support `overflow: overlay` so it will not affect them.

LostABike: Oh wow that's another interesting thing to know. Thanks

LostABike (approved):

FlyersPh9 (approved):

PR #1129, opened 2023-03-13
fix(Stepper): Remove track behind last active item

### LINKED ISSUE #1090: Stepper css issue with last element
When using the Stepper component, the last element stepper.css clashes with other css defined in stepper.css. The result is a not transparent line that appears on the last step of the Stepper component.
image

#### Expected behavior
The transparent line at the last step of Stepper.

##### Steps to reproduce
1. Use the Stepper component
1. Reach the final step
1. Additional line appears after the last step

#### Additional information
package version: "@itwin/itwinui-react": "^2.2.0",
Possible solution: fix the clashing css
image

FlyersPh9:
## Changes

- Moved the `:first-of-type` & `:last-of-type` beneath the styling for `.iui-current`.  Closes #1090.

## Testing

- Introduced 2 new visual tests to check when the first or last item is current.

## Docs

N/A


LostABike (approved):
This looks good, and the additional test images are definitely helpful and important test cases 😸

gretanausedaite (approved):
Fix looks good.

Why can't we focus on current step (using tab)? Is this expected usability for keyboard nav?

PR #1130, opened 2023-03-13
fix toaster breaking v1 styles

### LINKED ISSUE #1128: iTwinUI V2 `Toaster` breaks V1 styles [Incremental migration]
Every time we trigger a toast that has a close button, parent app using itwin v1 styles break.

#### Expected behavior
Toasts don't assign `.iui-root` class to `body` element

#### Reproduction
<https://stackblitz.com/edit/vitejs-vite-gknvln?file=src%2FV2.jsx>

##### Steps to reproduce
1. Parent app uses itwin V1
2. Add <ThemeProvider/> from V2
3. Render components from V2 (all ok)
4. Trigger Toast with options `{hasCloseButton: true}`.
5. Parent app using V1 styles break.  `.iui-root` was added to `body` element

#### Additional information
"@itwin/itwinui-react@1": "1.48.2",
"@itwin/itwinui-css@0": "0.63.2",
"@itwin/itwinui-react@2": "2.6.0",
"@itwin/itwinui-css@1": "1.7.0",

I assume problem is because toasts are rendered via `ReactDom.createRoot` and not `React.Portal`. Whole provider tree that was before it means nothing to it. It lives in it's own react world. And when time comes to `useTheme` it doesn't find anything up the tree, so it assigns new theme to `document.body`
It will probably ignore dark theme from the provider too

@mayank99 This might be major block for incremental migration <https://github.com/iTwin/iTwinUI/wiki/iTwinUI-react-v2-migration-guide#incremental-migration>

mayank99:
## Changes

Added a manual check for v1 classes (`iui-body`/`iui-theme-*`) and early return from `useTheme` if found. This fixes #1128.

## Testing

Tested in vite playground. See commit: [`7198dc6`](https://github.com/iTwin/iTwinUI/pull/1130/commits/7198dc69daa2607759268c33feb214b4c85d1b9a)

Also added unit test.

## Docs

N/A. Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/utils/hooks/useTheme.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useTheme.ts
@@ -59,6 +59,14 @@ export const useTheme = (
       return;
     }
 
+    // do nothing if using v1 for the main page (incremental migration)
+    if (
+      ownerDocument.documentElement.className.includes('iui-theme-') ||
+      ownerDocument.body.classList.contains('iui-body')
+    ) {
+      return;
+    }
+
     ownerDocument.body.classList.toggle('iui-root', true);
 
     switch (theme) {
--- a/.changeset/dull-pumas-explode.md
+++ b/.changeset/dull-pumas-explode.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue with incremental migration where adding a close button to v2 Toaster was breaking v1 styles for the whole page.
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -171,7 +171,8 @@ describe('Fallback (without children)', () => {
   };
 
   afterEach(() => {
-    document.body.classList.remove('iui-root');
+    document.documentElement.className = '';
+    document.body.className = '';
     document.documentElement.removeAttribute('data-iui-theme');
     document.documentElement.removeAttribute('data-iui-contrast');
     window.matchMedia = originalMatchMedia;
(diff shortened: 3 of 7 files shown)


packages/itwinui-react/src/core/utils/hooks/useTheme.ts
```diff
@@ -59,6 +59,14 @@ export const useTheme = (
...
 
+    // do nothing if using v1 for the main page (incremental migration)
+    if (
+      ownerDocument.documentElement.className.includes('iui-theme-') ||
+      ownerDocument.body.classList.contains('iui-body')
+    ) {
+      return;
+    }
+
     ownerDocument.body.classList.toggle('iui-root', true);
```
LostABike, line 70: If I understand this correctly, the addition of `iui-root` here and/or the rest of the code is what is breaking v1 style in the presence of v1 classes? Why is that the case?

mayank99: Yes, it is both. Here's how it works:
1. in v1, when using a component directly, the `useTheme` hook will add `iui-body` to `<body>` and `iui-theme-*` classes to the document element (`<html>`).
  
2. in v2, when using a component directly, the `useTheme` hook will add `iui-root` to `<body>` and `data-iui-*` attributes to `<html>`.
   
3. in v2, when wrapping the component tree in `<ThemeProvider>`, the `useTheme` hook will do nothing because `<ThemeProvider>` automatically adds those attributes/classses. This ensures that nothing outside ThemeProvider is affected (it is considered "scoped").

For this bug specifically, instead of 1 and 3 in the above list, 1 and 2 are happening. 1 and 3 are compatible but 1 and 2 are not.

Does that help?

LostABike: Yes that makes sense. Thanks for explaining 😸

LostABike (approved):

PR #1131, opened 2023-03-13
feat(Badge): Add predefined status variants

### LINKED ISSUE #964: Differentiate status badges from soft badges
To help make the status badges stand out from the soft badges we should consider a slightly different visual treatment.  I started [a badge refactor](https://github.com/iTwin/iTwinUI/tree/jon/refactor-badge) over 2 months ago that does this & makes use of `data-iui-variant` which would require a major version bump.  Pushed the work to a branch & filed an issue to be considered at a later date.

Screenshot 2023-01-06 at 4 15 43 PM
Screenshot 2023-01-06 at 4 15 59 PM

FlyersPh9:
## Changes

- Add `data-iui-status` to badge.
- Previous method of applying colors is unchanged / still valid.
- Removed badges from most tile demos to reduce future test image changes.

Closes #964.

Screenshot 2023-03-14 at 1 19 21 PM

## Testing

- Ran backstop.
- Confirmed that previous method is uneffected.

## Docs

N/A


packages/itwinui-css/src/badge/badge.scss
```diff
@@ -27,4 +27,15 @@ $iui-badge-horizontal-padding: calc(var(--iui-size-2xs) * 1.5);
   overflow: hidden;
   text-overflow: ellipsis;
   font-weight: var(--iui-font-weight-semibold);
+
+  &[data-iui-status] {
+    outline: 1px solid;
+    outline-offset: -1px;
```
mayank99, line 33: i think it would be better to use border. i know it occupies space and could cause a size inconsistency between status vs other badges, so my suggestion is to add border to all badges like this:
```css
border: 1px solid transparent;
```
and then inside the `iui-badge-status` mixin, you can set `border-color: currentColor`.

as a result, the code is simpler, works in safari and also results in better support for forced-colors.

mayank99 (approved):
Looks good apart from one small suggestion. Will also need the react portion implemented in parallel.

LostABike (approved):
LGTM 👍

packages/itwinui-css/src/badge/badge.scss
FlyersPh9: Replaced `outline` with `border`.  It's currently setting `border` on all badges.  Non-status badges have `--border-color: var(--background-color)`, then the status badges define background & border separately.  The border change did create a lot of image changes but `forced-color` support was something i forgot to consider & I think is more important.

mayank99: oh dang. we might want to remove badges from those tile tests - dont want another 100 images changing when we rework badges again

FlyersPh9: How about we test badges within tiles in only 1 sample?  Right now we have badges within 10 tile samples.

mayank99: yeah, feel free to experiment. we can even have multiple Tiles with badges as long as we don't use those same tiles for the more granular tests (hover/focus).

FlyersPh9: Reduced badge usage within tile demos to just 1 new tile.

idakukimiya (approved):

PR #1132, opened 2023-03-14
fix grid positioning in safari

mayank99:
## Changes

In #1117, we introduced css grid. This looked fine in chrome/firefox but was breaking in safari. Affected elements: thumbnail buttons and more-options button.

After investigating, I found the cause to be the use of `position: absolute` together with grid. These were left behind because we don't want the "overlaying" parts to occupy space. I was able to easily remove it from thumbnail buttons, but removing it from more-options was causing layout shift (because it now occupies space) so I had to get a bit creative.

The more-options button now uses `grid-row: -1` together with the metadata (both are considered "last row") and also uses `display: grid` to get rid of any excess line height. And for folder tiles, this is not enough for (there is no metadata region there) so I've removed it from the grid and reverted back to `position: absolute`.

## Testing

Tested successfully in Safari.

| Before | After |
| --- | --- |
| image | image |

| Before | After |
| --- | --- |
| image | image |

No visual changes in Chrome/Firefox.

## Docs

N/A. No changeset needed because #1117 is not released yet.

### DIFF
--- a/packages/itwinui-css/src/tile/tile.scss
+++ b/packages/itwinui-css/src/tile/tile.scss
@@ -136,6 +136,13 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
       margin-bottom: 0;
       -webkit-line-clamp: 2;
     }
+
+    .iui-tile-more-options {
+      position: absolute;
+      margin: 0;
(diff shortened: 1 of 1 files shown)


gretanausedaite (approved):
🌵

LostABike (approved):

PR #1133, opened 2023-03-14
fix(Table): "WithManualPaginator" story filtering fix

### LINKED ISSUE #1043: Table Column Filtering only apply to the current displayed page.
When applying a column filter in the table, only the displayed page is filtered. 

Here I get the message that tells me that no result contains "1009" in the description.
image
But when going to a different page, "1009" is present in a result. 
image

#### Expected behavior
The filtering should be applied on all the data in the source.  

#### Reproduction
Link to storybook table (02/02/2023) : 
https://itwin.github.io/iTwinUI/react/?path=/story/core-table--with-manual-paginator
1. Navigate to the storybook link shared above - ^
2. Apply a column filter that includes some data from a different page. (Ex: On page one, filter Description = "1009")
3. Notice that the data does not appear in the table.

#### Additional information
A possible resolution would be to fetch the data again, applying a filtering query.

elephantcatdog:
## Changes

Added manual filtering alongside the manual paginator in the storybook example "With Manual Paginator" (now "With Manual Paginator And Filter").

## Testing

- Updated visual test name to "With Manual Paginator And Filter" and ran all tests (passed).
- Played around with it on storybook and tested:
  - Filtering on name
  - Filtering on description
  - Filtering on name and description
  - Clearing filter(s)
  - Filtering and going to another page
  - Filtering and changing the number of items on a page
  - Filtering something with no result

## Docs
N/A

closes #1043


LostABike (changes requested):
I'm a little confused on the expected behavior of the filtering. When we filter both name and description, then it only shows results that match both, and hide everything that doesn't match, as a result the pages are reduced significantly:
image
But when I filter only for name (or description), only the first page shows the results with matching result, but the table still displays the remaining pages with non-matching results. And when you click on the other pages and come back to first page (filter still on), first page matching result is also no longer there? What should we expect?

Steps:
- Filter name (i.e 44):
image
- Go to page 2 (or any other pages):
image
No more off "44" results?
- Go back to page 1:
image

elephantcatdog:
@LostABike 

Whoops, I added some more specific code at the end and didn't run it. It's fixed now.

LostABike (approved):

LostABike:
@elephantcatdog yea my bad I wasn't sure what that does, tested it and couldn't figure out how to revert it lol. But I approved the changes so I think that should does it?

elephantcatdog:
@LostABike 
Yep, thanks!

adamhock (commented):
Looks good to me! I think the "With Manual Paginator" screenshot needs to be removed. Also, "With Manual Paginator" needs to change to "With Manual Paginator And Filter" in Table.test.ts

elephantcatdog:
@adamhock Updated!

PR #1138, opened 2023-03-15
fix grid positioning in safari (again)

mayank99:
## Changes

This PR basically undoes #1132 because it was causing layout shift in Chrome/Firefox when Tile didn't have metadata.  It is now replaced with hardcoded absolute positioning specifically for safari (as a final resort after spending way too much time trying to make it work in a cross-browser way).

This is achieved by targeting safari using `@supports (-apple-pay-button-style: inherit)` which i've put in a mixin.

## Testing

Before:

https://user-images.githubusercontent.com/9084735/225394169-1582488d-9a45-4f8e-a9ba-1e7bdc6c52b5.mov

After:

https://user-images.githubusercontent.com/9084735/225395385-18c73166-fa3a-4c16-a866-32b6330087ea.mov

Additionally, I've also tried to go through and hover over every single Tile to make sure there is nothing left.

## Docs

N/A

### DIFF
--- a/packages/itwinui-css/src/utils/mixins.scss
+++ b/packages/itwinui-css/src/utils/mixins.scss
@@ -142,3 +142,9 @@
     @content;
   }
 }
+
+@mixin safari-only {
+  @supports (-apple-pay-button-style: inherit) {
+    @content;
+  }
+}
--- a/packages/itwinui-css/src/tile/tile.scss
+++ b/packages/itwinui-css/src/tile/tile.scss
@@ -136,13 +136,6 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
       margin-bottom: 0;
       -webkit-line-clamp: 2;
     }
-
-    .iui-tile-more-options {
-      position: absolute;
-      margin: 0;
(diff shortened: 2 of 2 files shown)


LostABike (approved):

packages/itwinui-css/src/utils/mixins.scss
```diff
@@ -142,3 +142,9 @@
     @content;
   }
 }
+
+@mixin safari-only {
+  @supports (-apple-pay-button-style: inherit) {
```
FlyersPh9, line 147: I have not heard of [`-apple-pay-button-style`](https://css-tricks.com/snippets/css/apple-pay-buttons-in-css/) before.  Just to be sure I understand what is happening here, I'll recap from my perspective:

We're checking if the browser supports `-apple-pay-button-style`.  That only works in Safari, so it's a way of applying styling only for Safari.  We're not actually doing anything with the `-apple-pay-button-style`, we're just using it to target Safari only.

Clever. 🙂

FlyersPh9 (approved):
I pulled and hovered over all of the tiles, I did not see anything incorrect.  Interesting solution.

PR #1139, opened 2023-03-15
remove manual tooltip state from slider

mayank99:
## Changes

There is no need to manually maintain tooltip state because tippy already has a way of handling this. Hopefully should fix https://github.com/iTwin/iTwinUI/discussions/1122

## Testing

Tested in existing stories.

Updated unit test to use userEvent.

also tested in iTwin.js sandbox: see https://github.com/iTwin/iTwinUI/discussions/1122#discussioncomment-5326566

## Docs

Added changeset.

### DIFF
--- a/.changeset/lazy-buses-cross.md
+++ b/.changeset/lazy-buses-cross.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue where Slider's tooltip was still visible after unmounting the component.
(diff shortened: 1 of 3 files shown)


LostABike:
Do you mind sharing the sandbox link lol I just want to test quick test too. I was looking for it but I think you didn't paste

LostABike (approved):
Tested on Storybook, seems like it's working good 👍

gretanausedaite (approved):

PR #1140, opened 2023-03-15
feat(Badge): React portion of badge status variants

adamhock:
## Changes

Added react portion for data attributes for badge status

React code for #1131 

## Testing

Storybook test images and unit tests have been updated

## Docs

Changeset has been added


packages/itwinui-react/src/core/Badge/Badge.tsx
```diff
@@ -45,6 +23,7 @@ export type BadgeProps = {
    */
   backgroundColor?:
     | 'primary'
+    | 'informational'
```
adamhock, line 26: I added `'informational'` as a possible option for `backgroundColor` since the data attribute doesn't recognize `'primary'` as a status option. I think we should replace `'primary'` with `'informational'` in our next major release

packages/itwinui-react/src/core/Badge/Badge.tsx
```diff
@@ -69,15 +48,34 @@ export const Badge = (props: BadgeProps) => {
...
+    reducedBackgroundColor && statuses.includes(reducedBackgroundColor);
+
+  const _style =
+    reducedBackgroundColor && !isStatus
+      ? {
+          '--iui-badge-background-color': isSoftBackground(
+            reducedBackgroundColor,
+          )
+            ? SoftBackgrounds[reducedBackgroundColor]
+            : reducedBackgroundColor,
```
mayank99, line 67: i dont think it is necessary to pass status colors as css variables. they should be automatically set when using the data attribute

so most of this logic can be simplified and moved into two simple helper functions that you can call like this:

```
'--iui-badge-background-color': getBadgeColorValue(backgroundColor)
```
```
data-iui-status={getStatusValue(backgroundColor)}
```

.changeset/great-jobs-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Adding a data attribute for badge status
```
mayank99, line 5: This changeset can be improved.

```suggestion
Updated the visuals for status colors in Badge.
```

adamhock: This has been changed

packages/itwinui-react/src/core/Badge/Badge.tsx
adamhock: This has changed. I have it so that the background color is set in style regardless of whether `backgroundColor` is a status or not.

packages/itwinui-react/src/core/Badge/Badge.tsx
```diff
@@ -17,22 +17,26 @@ import '@itwin/itwinui-css/css/badge.css';
  * Helper function that returns one of the preset badge color values.
  */
 const getBadgeColorValue = (color: BadgeProps['backgroundColor']) => {
+  const statuses = ['positive', 'negative', 'warning', 'informational'];
```
adamhock, line 20: I wasn't able to use `getStatusValue` here because the return type includes the status strings which aren't able to be passed into `isSoftBackground`. If there is a way around this, please let me know

mayank99 (approved):
🚀

packages/itwinui-react/src/core/Badge/Badge.tsx
adamhock: Of course! Thanks for the explanation. This has been changed

as for `reducedBackgroundColor`, `'primary'` is not an option for `data-iui-status`, so the change from `'primary'` to `'informational'` is still necessary

LostABike (approved):

PR #1141, opened 2023-03-16
fix z-index issues with actionable tiles

mayank99:
## Changes

In #959, `z-index: 1` was added on the tile-name to allow clicking on the thumbnail region to activate the action. But this broke all the other buttons on the tile, such as the quick-action/type-indicator (on the top-left/top-right of the tile) and the more-options (bottom-right).

So I've changed tile-name to have `z-index: 0`, and everything that needs to be on top of it to have `z-index: 1`. This ensures that clicking on *most* of the tile will perform the default action (such as navigation or selecting the Tile) and clicking on these specific regions will perform the contextual action (such as opening a dropdown).

## Testing

Tested working correctly by clicking/hovering on all the different parts of tiles in the html test pages (under the "Default" section) as well as storybook (the "Actionable" and "Anchor Link" stories).

## Docs

N/A. Changeset not needed because #959 is not released yet.

### DIFF
--- a/packages/itwinui-css/src/tile/tile.scss
+++ b/packages/itwinui-css/src/tile/tile.scss
@@ -225,7 +225,7 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
   &,
   > .iui-link-action {
     width: 100%;
-    z-index: 1;
+    z-index: 0;
     overflow: hidden;
     white-space: nowrap;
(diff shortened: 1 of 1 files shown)


FlyersPh9 (approved):
I ran Storybook locally & tried clicking on various buttons within the tiles, keeping an eye on the Actions tab.  Everything appears to be working correctly.

adamhock (approved):
LGTM

PR #1142, opened 2023-03-16
feat(Menu): Omit border-radius on adjacent active menu items

### LINKED ISSUE #837: ComboBox: Adjacent selected `iui-menu-item`s could omit border radius
#### Feature
Currently when selecting multiple `iui-menu-item`s in the `ComboBox` with `multiple=true` (Multiple select), each item has a border radius. It might look better to omit the appropriate border radius sides when two selected `iui-menu-item`s are adjacent.

image

Something similar to how adjacent buttons don't have border radius in ButtonGroup

image

#### Additional information
The associated React PR that added multiple select support: [React PR #830](https://github.com/iTwin/iTwinUI-react/pull/830)

FlyersPh9:
## Changes

- When multiple menu items are selected next to each other, the interior items omit `border-radius`.  Is wrapped within `@supports` so for [browsers that don't support `:has`](https://caniuse.com/css-has) there will be no change.  **BackstopJS tests using Chromium 103, `:has` requires 105+ so you won't see these changes in test images.**  Closes #837.

Screenshot 2023-03-16 at 3 38 34 PM

## Testing

- Update menu tests to include a scenario with adjacent active menu items.

## Docs

N/A


packages/itwinui-css/src/menu/menu.scss
```diff
@@ -100,6 +100,26 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
...
+
+      // Styles first active item with an active item following it
+      &:has(+ .iui-active) {
+        border-bottom-left-radius: 0;
+        border-bottom-right-radius: 0;
+      }
+
+      // Styles last active item
+      &:has(+ :not(.iui-active)),
+      &:last-child {
```
rohan-kadkol, line 117: Just wondering, is the `:last-child` selector needed? Is the `&:has(+ :not(.iui-active))` not enough?

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -100,6 +100,26 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
...
+        border-bottom-left-radius: var(--iui-border-radius-1);
+        border-bottom-right-radius: var(--iui-border-radius-1);
```
rohan-kadkol, line 119: Instead of initially resetting all the border radius on [L105-107](https://github.com/iTwin/iTwinUI/pull/1142/files#diff-a340fb8a57004b5349e2d15e6cce890a749c4afd443d3323981302b08a2a77b0R105-R107) and readding it for this one case, would it be safer code to not reset the border radius but only remove the border radius in certain cases?

* first item: remove only bottom border radii
* middle items: remove top and bottom border radii
* last item: remove top border radii

This is just a suggestion. Because for example, if we ever change the border radius on [L17](https://github.com/iTwin/iTwinUI/blob/47d994cf3d58119c3d63efe60f909733c7e5d7a9/packages/itwinui-css/src/menu/menu.scss#L17) but forget to change it here, then there could be inconsistencies in the border radius specification.

rohan-kadkol (approved):

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -100,6 +100,26 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
   }
 
   &.iui-active {
+    @supports (selector(.iui-active:has(+ .iui-active))) {
```
mayank99, line 103: this check doesn't need any specific selectors. should just be:
```suggestion
    @supports selector(:has(+ *)) {
```

packages/itwinui-css/src/menu/menu.scss
```diff
@@ -100,6 +100,26 @@ $iui-active-outline: thin solid var(--iui-color-border-accent);
...
+      // Styles active items between the first & last
+      + .iui-active {
+        border-radius: 0;
+      }
+
+      // Styles first active item with an active item following it
+      &:has(+ .iui-active) {
```
mayank99, line 110: `iui-active` is not enough by itself, we usually want `iui-menu-item.iui-active` together.

you can write it like this to keep specificity flat and avoid the need to even reference any selector directly:
```scss
    &:where(:has(+ #{&})) {
      border-bottom-left-radius: 0;
      border-bottom-right-radius: 0;
    }

    + :where(#{&}) {
      border-top-left-radius: 0;
      border-top-right-radius: 0;
    }
```

packages/itwinui-css/src/menu/menu.scss
FlyersPh9: I knew there was a better way to do this, I just couldn't piece it together.  Suggestions here are much better, changes made!

packages/itwinui-css/src/menu/menu.scss
FlyersPh9: Much cleaner, change made.

mayank99 (approved):

packages/itwinui-css/src/menu/menu.scss
FlyersPh9: Changes from below made the `:last-child` no longer necessary.

PR #1143, opened 2023-03-16
Adjust min-height for VirtualizedScroll 

### LINKED ISSUE #1107: Virtualized ComboBox miscalculates height of options
The bottom of the options list shows a gap below the last option when using a virtualized combobox where the **second option** has a sublabel and a sufficient number of options without sublabels.

#### Expected behavior
The bottom of the options list stops at the last option.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/gracious-pond-ur3c5g?file=/src/App.tsx

##### Steps to reproduce
1. Open the combobox
1. Observe gap below the last option

#### Additional information
This occurs in both scrollable and unscrollable option lists.
It appears the gap is the result of the scrollable distance being calculated as if every option has a sublabel, accumulating all the height differences from all options without sublabels at the bottom.
Testing seems to suggest this only occurs when the second option has a sublabel.

LostABike:
## Changes

ComboBox min-height calculation (line 420 `VirtualScroll.tsx`) is no longer accurate, because it assumes all middle child height are the same which is not always true as there can be child with sublabel and child without. I made a new function to calculate each individual heights and use that to set min-height instead.

But side note, because it is a "min-height", I feel like it might be better to calculate shortest node multiply by the amount of node instead? Since min-height is supposed to be the floor.

Closes #1107 

## Testing

- Test different ComboBox child elements in vite playground and container's min-height
Code I use to test in vite: https://codesandbox.io/s/dawn-sea-tjzr2j?file=/src/App.tsx


elephantcatdog (commented):
It looks like `VirtualScroll` is a utility that can be and is used for more than just ComboBox; did you take that into consideration?

~Also, I'm a bit uncertain on what problem you're trying to solve with this. Can you send me the code you used to test it?~ Nevermind I see the linked issue now.

veekeys:
> > It looks like `VirtualScroll` is a utility that can be and is used for more than just ComboBox; did you take that into consideration?
> > ~Also, I'm a bit uncertain on what problem you're trying to solve with this. Can you send me the code you used to test it?~ Nevermind I see the linked issue now.
> 
> @elephantcatdog You're right, it is probably better to calculate the height of each node individually instead, which I just updated. I also included sandbox in the description for the testing code that can be copied into vite. Also my bad I forgot to add the "closes x" or linking the problem 😅

`VirtualScroll` DOES NOT support dynamic height elements (I think it was documented somewhere). That basically means, if list consists of items of different height, miscalculations are expected. So to fix these, need to add support of dynamic heights elements, which is not that straight forward. I did have a branch with trying to add this feature. I will try to find time and see if I can revert that stash from somewhere.

elephantcatdog:
@LostABike You've got some visual storybook tests failing for combobox and table.

veekeys:
Ok, I pushed a VERY OLD code here
https://github.com/iTwin/iTwinUI/tree/vyki/dynamic-virtual-scroll
If you somehow make the build pass (mine failed on storybook, could not figure why), you should be able to see this working with table rows of different heights.
Again, would need to take the logic out and apply to current VS component, but hope this code will be useful for you guys to work on.
I might check it up myself, but currently have a lot of different stuff on my plate, so just physically cannot do it rn.
Please tag me on the PR, if you apply some of these and make it work, would be really interesting to see!

PR #1147, opened 2023-03-20
fix z-index issues with actionable tiles (again)

mayank99:
## Changes

#1141 was only working on firefox, so this change makes it work on chrome too.

## Testing

same as #1141 but made sure to click on thumbnail area in chrome and safari too.

Before:

https://user-images.githubusercontent.com/9084735/226354345-48920c2d-ec5e-4c93-beb6-192a1ea3e413.mov

After:

https://user-images.githubusercontent.com/9084735/226354490-76c2ec04-e27b-400e-b9a3-251a7f0b22aa.mov

(results might vary by operating system)

## Docs

N/A

### DIFF
--- a/packages/itwinui-css/src/tile/tile.scss
+++ b/packages/itwinui-css/src/tile/tile.scss
@@ -225,7 +225,7 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
   &,
   > .iui-link-action {
     width: 100%;
-    z-index: 0;
+    z-index: 1;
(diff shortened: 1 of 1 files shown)


gretanausedaite (approved):
Works for me (in chrome and safari)


> (results might vary by operating system)

Make sure we have that in writing in docs.

gretanausedaite:
> why do we want that in docs? i'm saying my screen recordings might not match the system of the PR reviewer. i.e. the "Before" and "After" might be the same on windows

Just joking.

Issue #922, opened 2021-10-12
Handle Alt key in keyboard navigation


bentleyvk:
Handle Alt key in keyboard navigation. Example: [howto-tabs](https://developers.google.com/web/fundamentals/web-components/examples/howto-tabs#:~:text=don%E2%80%99t%20handle%20modifier%20shortcuts%20typically%20used%20by%20assistive%20technology.)

mayank99:
More detail on what this task entails:
1. Find components where we do custom keyboard handling (using `onKeyDown`).
2. Early return if <kbd>alt</kbd> key is pressed.
3. Possibly move the early return check into a utility that can be reused.

PR #1149, opened 2023-03-20
Handle-alt-key-in-keyboard-navigation

idakukimiya:
## Changes

-Updated all files with either const handleKeyDown or cont onKeyDown with if (event.altKey).  

Closing #922

## Testing

Each file or section in the storyboard should be able to use the alt-tab key and the arrow key to move around.

## Docs

N/A


mayank99:
This PR does *nothing* because:
- the utility function simply returns without a value
- it is not even used anywhere

Did you try testing in any of the components? Perhaps you are missing some changes or forgot to push them?

I also think there might not be a need to create a utility function if it is only going to check for `altKey`. When I initially brought up the idea, I was thinking it could check for all common modifier keys, similar to how it's done in Carousel.

https://github.com/iTwin/iTwinUI/blob/f16d71bf2b09b732f2f3103a5d7aa03baa3da0b6/packages/itwinui-react/src/core/Carousel/Carousel.tsx#L97

LostABike:
It is like @mayank99 mentioned, you don't really need to make a utility function for this, but simply just look for anywhere we implement keys handling via `onKeyDown` function, and if it doesn't have alt key handling yet, add ` if (event.altKey) { return; }` in that function near the top.

You can check out an example of this in `MenuItem.tsx` line 109

I have mentioned all these in the meeting but just want to iterate again so you can refer back if you need to.

apps/website/src/env.d.ts
idakukimiya, line 1: I deleted this file.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
idakukimiya, line 1: This PR push is to see if I am going in the right direction.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -106,7 +106,7 @@ export const MenuItem = React.forwardRef<HTMLLIElement, MenuItemProps>(
     const [isSubmenuVisible, setIsSubmenuVisible] = React.useState(false);
 
     const onKeyDown = (event: React.KeyboardEvent<HTMLLIElement>) => {
-      if (event.altKey) {
+      if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
```
LostABike, line 109: What you're doing here is adding handling CTRL, meta key and SHIFT key. What we need to do is adding **only** ALT key handling. Meaning this was already handled before your edit, or it was already good/how it should be. So you should revert or remove your change here. Instead, find any `onKeyDown` function that **do not** have ALT key handling yet, and then add ALT key handling to it.

For example, check out the `onKeyDown` function of file/component `StepperStep.tsx` which is defined on line 65. You can see that within this function, there is no ALT key handling:

```
  const onKeyDown = (e: React.KeyboardEvent) => {
    if (!isClickable) {
      return;
    }

    if (e.key === 'Enter' || e.key === 'Space' || e.key === ' ') {
      onCompletedClick();
    }
  };
 ```
 
 You can add ALT key handling into this function by adding condition to check for ALT key as follows:
 ```
   const onKeyDown = (e: React.KeyboardEvent) => {
    if (e.altKey){
      return;
    }
    
    if (!isClickable) {
      return;
    }

    if (e.key === 'Enter' || e.key === 'Space' || e.key === ' ') {
      onCompletedClick();
    }
  };
  ```
 
And that is all you need to do. You should look for all `onKeyDown` functions that do not have ALT key handling yet, and add the ALT key handling into it like my example above. To start with, you can check out `StepperStep.tsx` and imitate my example.

**Comment by LostABike** [comment]
The alt key handlings that you added look good! 🎉

There are some places that we override default alt key handling behavior using defined `handleKeyDown()` like `ColorPalette.tsx` or straight up inline function like in `TimePicker.tsx` line 626. Should we add alt key handling in these situations? @mayank99

**Comment by mayank99** [comment]
> There are some places that we override typical keys handling behavior using defined `handleKeyDown()` like `ColorPalette.tsx` or straight up inline function like in `TimePicker.tsx` line 626. Should we add alt key handling in these situations? 

Yeah, makes sense to me

**Comment by LostABike** [comment]
The update is good. There are still many more places that have custom keys handling, you should find all of these using the search feature in VSCode, look for `onKeyDown`. They will either be passed in a different function like `handleKeyDown` so you can add alt key handling via those functions (an example of this is in `ColorBuilder.tsx`), or inline handling like `Breadcrumbs.folder.tsx`. My examples are not comprehensive, try to look for all of them.

**Comment by mayank99** [comment]
@idakukimiya Is it clear what needs to be done? Do you need any help updating the remaining instances? I've found only two more files that need to be changed:
- ColorBuilder.tsx
- DatePicker.tsx

**Comment by idakukimiya** [comment]
@mayank99 I have added the alt key to the 2 files: 
-ColorBuilder.tsx
-DatePicker.tsx

File: .vscode/settings.json [hunk 2]
```diff
@@ -46,5 +46,6 @@
     "*.astro": "${capture}.tsx, ${capture).css",
     "*.tsx": "${capture}.ts, ${capture}.css",
     "*.stories.tsx": "${capture}.test.ts"
-  }
+  },
+  "cSpell.words": ["dont", "itwin", "itwinui", "Tbgr"]
```
mayank99, line 50: undo this change. If you want to add some terms to your dictionary, you can always do it for your "User settings" rather than "Workspace settings" in the dropdown menu

apps/website/src/pages/docs/tooltip.mdx
```diff
@@ -9,13 +9,10 @@ group: core
 
 import PropsTable from '~/components/PropsTable.astro';
 import LiveExample from '~/components/LiveExample.astro';
-import Placeholder from '~/components/Placeholder.astro';
```
mayank99, line 12: please undo all changes in tooltip.mdx as well

mayank99 (approved):
Changes look good. Currently missing changeset, and PR description about how you tested these changes.

FlyersPh9 (approved):

PR #1150, opened 2023-03-21
doc(Breadcrumbs): Add custom overflow stories

elephantcatdog:
## Changes
I added the two custom overflow breadcrumbs stories to the website. The two examples are for a back button and a dropdown. I also added some text as explanation.

image

## Testing
I made sure the website examples worked as intended.

## Docs
N/A


apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <div style={{ maxWidth: '50%', border: '1px solid lightpink', padding: 8 }}>
+      <Breadcrumbs
+        overflowButton={(visibleCount: number) => {
+          const previousBreadcrumb =
+            visibleCount > 1 ? items.length - visibleCount : items.length - 2;
+          return (
+            <Tooltip content={`Item ${previousBreadcrumb}`} placement='bottom'>
+              <IconButton
+                onClick={() => {
+                  // open previous breadcrumb
```
mayank99, line 24: Did you forget to push some code? This example is not very helpful if nothing happens on click.

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <div style={{ maxWidth: '50%', border: '1px solid lightpink', padding: 8 }}>
+      <Breadcrumbs
+        overflowButton={(visibleCount: number) => {
+          const previousBreadcrumb =
+            visibleCount > 1 ? items.length - visibleCount : items.length - 2;
+          return (
+            <Tooltip content={`Item ${previousBreadcrumb}`} placement='bottom'>
+              <IconButton
```
mayank99, line 22: buttons must have an [accessible name](https://www.tpgi.com/what-is-an-accessible-name/) which you can specify using `aria-label` when there is no text inside the button

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Breadcrumbs, Button, Tooltip, IconButton } from '@itwin/itwinui-react';
+import { SvgMore } from '@itwin/itwinui-icons-react';
+
+export default () => {
+  const items = Array(10)
+    .fill(null)
+    .map((_, index) => <Button key={index}>Item {index}</Button>);
+
+  return (
+    <div style={{ maxWidth: '50%', border: '1px solid lightpink', padding: 8 }}>
```
mayank99, line 15: a bigger maxWidth value will work better on smaller screens

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
elephantcatdog: So this example is supposed to be a 'back' button to the last hidden breadcrumb. Nothing happens when I click a regular breadcrumb, so why should something happen when this button is clicked? Should something be happening when breadcrumbs are clicked? None of our examples are interactive like that.

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
elephantcatdog: Updated

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
elephantcatdog: Updated to match the overflow example already there

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
mayank99: Because you are specifically trying to show a case where something different should happen for this overflowButton (otherwise why even show the example?)

There is always a fine line between overdocumenting and not documenting enough. So you could enhance this example to maintain the "current" state which changes when you click on the different buttons. But I think in this case it might be enough to not even show an example.

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
mayank99: it's better but still doesn't look great on small screen widths
image

the same problem is also present in the original example
image

you can guard against this using something like `maxWidth: 'min(90%, 425px)'`, or just a larger percentage in this case (`maxWidth: 70%`)

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
elephantcatdog: I removed it

apps/website/src/examples/Breadcrumbs.customOverflowBackButton.tsx
elephantcatdog: Updated the truncation examples: 
image

LostABike (approved):

PR #1151, opened 2023-03-21
fix(Breadcrumbs): Update breadcrumbs stories to follow tabs

elephantcatdog:
## Changes

I just moved the code related to breadcrumbs from the [tab overflow PR](https://github.com/iTwin/iTwinUI/pull/1060) to this one. It changes the story formatting to:
- Use `SvgMore` instead of `SvgMoreSmall`
- Center the above icon
- Changes the style to improve resizing sizes (won't get too big or too small when changing the width of the page)

This PR no longer changes the color of button icons in breadcrumbs.

image

## Testing
Updated storybook test images

## Docs
N/A

~## Changes
I updated the CSS for breadcrumbs so that button icons (eg. overflow button icons) in the list would match the color of the text. I think it makes it look more 'clickable'. It used to match but it got changed in the major update.~

~Before:~

~After:~

~## Testing
Updated storybook test images~

~## Docs
Added changesest~


FlyersPh9:
I left [a comment](https://github.com/iTwin/iTwinUI/pull/1060/files#r1145328566) in #1060 that kind of relates to this PR.  If we go with what I suggest in that comment thread, I think this PR can be closed.

gretanausedaite (commented):
Could you also update CSS demo to have them matching?

@FlyersPh9 Check if visuals are what we want.

elephantcatdog:
@gretanausedaite Added to css demo

packages/itwinui-css/backstop/tests/breadcrumbs.html
```diff
@@ -354,6 +355,96 @@ <h2>Breadcrumbs button</h2>
 
     </section>
 
+    <h2>Overflow button</h2>
```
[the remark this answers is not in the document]
elephantcatdog: There was previously no example for an overflow button in the CSS demo. Only the regular overflow '...' text. In react I'm updating the example with the button also, not the examples with the text '...'.

packages/itwinui-css/backstop/tests/breadcrumbs.html
```diff
@@ -354,6 +355,96 @@ <h2>Breadcrumbs button</h2>
...
+    <section id="demo-buttons-2">
```
gretanausedaite, line 359:
```suggestion
    <section id="demo-buttons-3">
```

packages/itwinui-css/backstop/tests/breadcrumbs.html
```diff
@@ -354,6 +355,96 @@ <h2>Breadcrumbs button</h2>
...
+          <li
+            class="iui-breadcrumbs-separator"
+            aria-hidden="true"
+          >
+            <svg-chevron-right aria-hidden="true"></svg-chevron-right>
+          </li>
+          <li class="iui-breadcrumbs-item">
+            <button
+              class="iui-breadcrumbs-button"
+              id="test-7"
```
gretanausedaite, line 410: Delete this id.
```suggestion
```

packages/itwinui-css/backstop/tests/breadcrumbs.html
```diff
@@ -354,6 +355,96 @@ <h2>Breadcrumbs button</h2>
...
```
gretanausedaite, line 393:
```suggestion
              id="overflow-hover"
```

packages/itwinui-css/backstop/tests/breadcrumbs.html
```diff
@@ -354,6 +355,96 @@ <h2>Breadcrumbs button</h2>
...
+          <li
+            class="iui-breadcrumbs-separator"
+            aria-hidden="true"
+          >
+            <svg-chevron-right aria-hidden="true"></svg-chevron-right>
+          </li>
+          <li class="iui-breadcrumbs-item">
+            <button
+              class="iui-breadcrumbs-button"
+              id="test-8"
```
gretanausedaite, line 424:
```suggestion
              id="overflow-focus"
```

gretanausedaite (commented):
You did not add a new test to CSS scenarios.
Do that by adding this code to breadcrumbs.js in screnarios folder
```js
  scenario('State hover and focus overflow buttons', {
    actions: [hover('#overflow-hover'), focus('#overflow-focus')],
    selectors: ['#demo-buttons-3'],
  }),
```

In HTML test file you did not change the IDs of the test, therefore it "worked" for you and created the images which are copy of breadcrumbs buttons test (you can check the name of the files). I suggest changing the IDs in your newly added test.
Run the tests again to generate images and then clean the old ones.

packages/itwinui-css/backstop/tests/breadcrumbs.html
elephantcatdog: Updated

packages/itwinui-css/backstop/tests/breadcrumbs.html
elephantcatdog: Updated

packages/itwinui-css/backstop/tests/breadcrumbs.html
elephantcatdog: Updated

gretanausedaite (approved):

elephantcatdog:
> Everything looks good for me. I just have a logic question but I think it's irrelevant to this PR 🤔 But for the overflow back button, how would users go back further than the one immediately before that? I.e 2->5 is hidden but button only shows 5, can't go back to 4. I'm guessing we left that up to the users to implement the specific logic.

Yeah, the specifics are left up to users. In that case you'd have to click to go back multiple times. It's probably going to be more usual for users to make a dropdown menu, I'd think. But maybe users just want a breadcrumbs button that always goes back to a 'landing page' for a complicated process or something. I wanted to give more examples to give users more options.

PR #1152, opened 2023-03-21
Add iui-dialog-full-page into full page modal

### LINKED ISSUE #1022: Full page modal: not opening properly in iTwinUI storybook
When clicking "Full page modal" and going to "Canvas". Modal slides up and disappears from screen:

https://user-images.githubusercontent.com/122555873/213699629-04af2b07-88b8-440e-89ff-88aadb0eca9c.mp4

When switching to "Docs" the animation is a bit jumpy and slides in from top of the screen instead of bottom:

https://user-images.githubusercontent.com/122555873/213699620-21457c09-7d47-4182-852c-a62ccbfd92f4.mp4

Browser: Chrome Version 109.0.5414.75 (Official Build) (64-bit)
Laptop: HP ZBook 15 _(system information will follow)_

#### Expected behavior
Note that the full page modal is working correctly in the CSS iTwinUI library for me:

https://user-images.githubusercontent.com/122555873/213700695-ea489165-b751-4f79-8e2d-704c2568c2b5.mp4

##### Steps to reproduce
1. Open: https://itwin.github.io/iTwinUI/react/?path=/story/core-modal--full-page-modal
1. Make sure the switch in top left is on "Canvas"
1. Click on button to open full page modal

#### Additional information
--

LostABike:
## Changes

The full page dialog animation in HTML is working ok, but Storybook and doc isn't.
I think this is because `iui-dialog-visible` is added *at the same time* as `iui-dialog-animation`, so the `enter` and `enter-active` animation is happening at the same time, while `iui-dialog-visible` should be slightly slower than the `iui-dialog-animation` so that this code has time to take effect:

```
  &:not(.iui-dialog-visible) {
    transform: translateY(100%);
  }
 ```
So I added a small delay of 1 millisecond before applying the `iui-dialog-visible` class so the code above have time to take effect. Surprisingly that does fix the glitchy animation and it now correctly slides from bottom up 🤞 

This feels hacky but it works, I can't think of a better way to do this..(also spent too much time staring at this thing..)

Closes #1022

## Testing

- Visual test in Storybook and doc website.
- Ran all tests: Approved react visual test and updated failing unit tests. The unit tests were looking for `.iui-dialog` class that includes `.iui-dialog-visible` but because the dialog is not always visible (I'm assuming they are still there) until animation fired, I've removed the extra class and the tests now pass.

## Docs

- Added changeset

### CODE REVIEW

File: packages/itwinui-react/src/core/Modal/Modal.tsx
Comment (mayank99) [inline packages/itwinui-react/src/core/Modal/Modal.tsx:1]: Try checking `DialogMain.tsx`. In there we already have the div which has `iui-dialog` and `iui-dialog-full-page` classes. The animation is handled by the `WithCSSTransition` helper which uses [`react-transition-group`](https://reactcommunity.org/react-transition-group/css-transition). This helper will handle the entry/exit animations while also making sure that the component gets removed from the DOM after the animation is complete. You can read the docs, play with the durations and inspect the DOM to get a better understanding of how it works. Let me know if you have any questions.

Other than that, I should mention that in the html test pages, it stays in the DOM even when not visible, but we don't want that behavior in React. We want to unmount the component when it's not visible, while still having correct animations.

Comment (LostABike) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: Hmm I think the issue is when the dialog is not open or visible, the component isn't there. However the CSS code relies on the fact that the component **is** there and condition that dialog is not yet visible in order to apply this styling:
image

Which is why hacking to have the component exists before opening the dialog seems to fix the issue. But I don't know how I can achieve that in React, or like you said, probably shouldn't do that anyway. So could this be something we need to change with the CSS code?

Comment (mayank99) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: You're right, that line might be messing with the animation. But there is also CSS specifically for the enter/exit classes too.
https://github.com/iTwin/iTwinUI/blob/31a02321644f694cd26b13bc9309620b15efb51a/packages/itwinui-css/src/dialog/dialog.scss#L125-L135

These classes are toggled by the WithCSSTransition helper. This is why I suggested playing with the duration and inspecting the DOM. In this recording you can see the `iui-dialog-animation-enter`, `iui-dialog-enter-active` and `iui-dialog-enter-done` classes at various stages of the animation.

https://user-images.githubusercontent.com/9084735/226726336-9eec116c-858a-4f6e-a1d9-ef77712fe89a.mov

Ultimately you will just need to make sure that the classes are applied in the right order, and you might need to modify both CSS and React for that.

Comment (LostABike) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: I am trying to review your feedback but the video is no longer available. Do you mind reposting it?

Comment (LostABike) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: Updated! It's working now. Although code might be questionable 🤣

Comment (LostABike) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: Wow 🙏 Yes that seems like to proper way to do this. Thanks for the help. Updated. Some of the React tests are failing though. I'm looking into that.

Comment (LostABike) [reply packages/itwinui-react/src/core/Modal/Modal.tsx:1]: Updated unit tests and PR description for more details.

File: packages/itwinui-react/src/core/Dialog/DialogMain.tsx [hunk 1]
```diff
@@ -227,8 +226,15 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
...
         in={isOpen}
-        classNames='iui-dialog-animation'
+        classNames={{
+          enter: 'iui-dialog-animation-enter',
+          enterActive: 'iui-dialog-animation-enter-active',
+          enterDone: 'iui-dialog-visible',
+        }}
         timeout={{ exit: 600 }}
+        onEntered={() => {
+          setTimeout(() => dialogRef.current?.focus(), 300);
```
mayank99, line 236: @LostABike Are you sure this `setTimeout` is needed? In my testing it was working fine even without it, because `onEntered` is only called after the animation is done.

On a different note, you only want to call this code when the `setFocus` prop is set to `true`.

Also, previously this focus behavior was handled in useEffect, near line 87. I think you can remove much of that code now. The `previousFocusedElement` behavior could be handled through one of the `onExit` calbacks instead.

mayank99: Oh I see, so the focus works, but the animation is glitchy.

I believe this is because focused elements are automatically scrolled into view by the browser. You can override this behavior using [`preventScroll`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#parameters).

```js
dialogRef.current?.focus({ preventScroll: true })
```

LostABike: Oh man...today I learn 😅 Yea that'd do it

LostABike: Updated, removed `setTimeout` and the hook that was used to handle focusing, and added focus handling via callbacks.

LostABike: Okay I think this is what you mean. Updated.

packages/itwinui-react/src/core/Dialog/DialogMain.test.tsx
```diff
@@ -64,15 +63,14 @@ it('should render with custom style and className', () => {
...
-  ) as HTMLElement;
+  const dialog = container.querySelector('.iui-dialog') as HTMLElement;
   expect(dialog).toBeTruthy();
   expect(dialog.classList.contains('test-class')).toBeTruthy();
   expect(dialog).toHaveStyle('color: red');
 });
 
-it('should close on Esc click and move focus back', () => {
+it('should close on Esc click and move focus back', async () => {
+  config.disabled = true;
```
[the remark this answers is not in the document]
LostABike: that is so weird I could have sworn it was failing without, because after a while I asked Mayank if there's anything else we could do....but I also just tested removing it and now it passed..? So I guess we don't need it lol. But basically it's to disable the react transition time for animation, in this case dialog enter and exit animation time.

PR #1156, opened 2023-03-23
Add check for undefined header in Column resizer

### LINKED ISSUE #1004: Table: improve resizing for action column
Closed the following issues in favor of this issue:
* #2555

---

When table contains column manager and I try to resize a column (in expand mode), the app crashes.

#### Expected behavior
Column properly expands/shrinks.

#### Reproduction
`columnResizeMode: "expand"`
[Link to a minimal repro: 
](https://codesandbox.io/s/itwinui-react-example-forked-jpt6vg?file=/src/App.tsx)

##### Steps to reproduce
1. Just try to resize last column :D

https://user-images.githubusercontent.com/75395229/212881490-58b2a8b3-ef19-44e4-b9a5-a42dd842cea1.mp4

#### Additional information
`iTwinUI-react: 1.48.1`

LostABike:
## Changes

When resizing the last column that is column manager in `expand` mode, the table crashes with error "cannot read properties of undefined" which seem to occur from using `getHeaderWidth` function. So I added a check within that function to see whether if `header` is undefined and return 0. 

Closes #1004

## Testing

- Tested column manager table in Storybook with extra prop `isResizable=true` and `columnResizerMode='expand'`

## Docs

- Added changeset


elephantcatdog (commented):
So this doesn't show the error anymore, but it doesn't allow me to actually resize the column. Try to get that working. If you need help, let me know and I can help you figure it out.

LostABike:
@elephantcatdog the storybook story works. 

> So this doesn't show the error anymore, but it doesn't allow me to actually resize the column in the example the user sent in. Does it work for you in the storybook story?

The storybook story works-ish..? I'm not too confident about the expected behavior though:
column resize test
ontent.com/22462084/227565395-2e4b4453-878b-4770-a974-6f48db36fee1.gif)

I was thinking that we disable resizing for the the selection/first column, should we do the same for column manager column? That would keep the table width the same too. But I'm not sure what the expected behavior should be.

elephantcatdog:
@LostABike Oooh, I see now. I was trying to make the column manager column bigger, but I actually don't see a reason for that to be bigger IMO. In that case, I tested it and this looks good to me.

LostABike:
> It fixes the error but now I can make columns smaller than table. image

I can't seem to reproduce this 😮 I used the user's example to run locally in vite, which I believe was what you did as well. Could you elaborate on how you were able to reproduce that bug?

gretanausedaite:
> Could you elaborate on how you were able to reproduce that bug?

https://user-images.githubusercontent.com/83585998/229789879-fdb1976a-87ea-4752-89fd-d990a4901636.mov

I think it's something with how we calculate min width of the column based on table width.

gretanausedaite (commented):
I think we can merge this as a hotfix for crashing. Table changing size is better that app not working.
We just need to add a new issue for resizing bugs.

@LostABike @mayank99 What do you think?

mayank99:
Sounds good to me. Lets merge this PR and reopen the issue.

gretanausedaite (approved):

PR #1157, opened 2023-03-23
fix site url for seo

mayank99:
## Changes

Astro requires the [`site`](https://docs.astro.build/en/reference/configuration-reference/#site) property to be specified in config to generate correct sitemaps. This property is then also used to generate canonical urls for SEO.

## Testing

Tested correct urls showing up in build output. (Previously it was not even generating sitemap and was using localhost in some of the meta tags)

## Docs

N/A


FlyersPh9 (approved):

elephantcatdog (approved):
Looks good to me

PR #1158, opened 2023-03-27
website: Buttons documentation update

gretanausedaite:
## Changes

Updated documentation of buttons.
Button part of #961

## Testing

N/A

## Docs

N/A


apps/website/public/favicon.ico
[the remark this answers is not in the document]
gretanausedaite: Because it is loaded in prod 
image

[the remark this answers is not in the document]
gretanausedaite: At least we should update that to new favicon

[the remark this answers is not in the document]
mayank99: handle it in a different PR then.

we need an `.ico` version of the new favicon to use as fallback for when svg is not supported

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,32 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
+
+## Usage
+
+### Style types
+
+- Use default button just about everywhere. The default button can display a text label, an icon, or a text label & icon in combination.
+
+- Use high visibility Button when you want to draw the user’s attention to a specific action applicable to a section of a UI or “Page”. Remember, there should be only 1 high visibility button per section (e.g. in a modal dialog, or above the table). The high visibility button can display a text label, or a text label & icon in combination. Icon buttons should not have this styletype.
+
+- The call to action button is the primary function on the page. There should only ever be one CTA on a page at a time. The call to action always appears within a dedicated page footer so as to make it very clear that the application is awaiting this main input before it can switch modes or views. The call to action button can display a text label, or a text label & icon in combination. Icon buttons should not have this styletype.
```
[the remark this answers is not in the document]
mayank99: Make it shorter then. The "can disaplay text label" etc can be removed.

Also borderless button is missing from list

apps/website/public/favicon.ico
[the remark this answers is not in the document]
gretanausedaite: Added back

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,31 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
 ### Props
 
 <PropsTable path={frontmatter.propsPathIdeasButton} />
+
+## Usage
+
+### Style types
+
+- Use default button just about everywhere.
+- One high visibility button per section (e.g. in a modal dialog, or above the table).
```
FlyersPh9, line 133:
```suggestion
- One high visibility button per section such as within a [dialog](dialog), or above a [table](table).
```

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,31 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
+- One CTA button on a page at a time and should always appear within a dedicated page footer.
```
FlyersPh9, line 134: We define CTA above, but if someone scrolled past it and missed it, CTA might be confusing.

The old site used [`<abbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) on acronyms.  What do you makes the most sense?  We could consider this in a separate PR just to get this merged quicker.

mayank99: I would just put "(blue)" and "(green)" in parentheses.

`<abbr>` has known accessibility issues.

mayank99: and of course, we can just spell "call-to-action" too.

In addition to that, we can link back to that the sections above.

apps/website/src/pages/docs/button.mdx
```diff
@@ -18,51 +18,92 @@ import * as AllExamples from '~/examples';
...
 
+Pressing this left side of the button will commit the action. Pressing the right side of the button will open a menu.
+
 <LiveExample src='SplitButton.main.tsx'>
   <AllExamples.SplitButtonMainExample client:load />
 </LiveExample>
 
-Pressing this left side of the button will commit the action. Pressing the right side of the button will open a menu. Upon selecting an option from the menu, the action immediately takes place, and the left side of the button’s label will change to match the recent selection.
-
 ### Props
```
FlyersPh9, line 86: I started using `Properties` instead of `Props`.  @jbwriter asked how props might translate, so I figured it would be best to use the full word.

mayank99: "Props" is not just a shorthand, it's the canonical term and we should not deviate from it.
https://react.dev/learn/passing-props-to-a-component

FlyersPh9: Ok, I will revert my branch to `Props` and will share this with Jay.

apps/website/src/pages/docs/button.mdx
FlyersPh9: I'd prefer to either link to the CTA section, completely spell out call-to-action, or both (no preference between these).  I would prefer to not say blue or green.

gretanausedaite: Using call-to-action, still have CTA definition because we use that in props.

FlyersPh9 (approved):

apps/website/src/pages/docs/button.mdx
```diff
@@ -18,51 +18,92 @@ import * as AllExamples from '~/examples';
...
+<LiveExample src='Button.cta.tsx'>
+  <AllExamples.ButtonCTAExample client:load />
+</LiveExample>
+
+### Borderless
+
+The borderless button is useful in tighter areas where using a default button would visibly cause too many boxes.
+
+The borderless button is often used by actionable icons in relation to another component.
+They are commonly used within [table](table) rows and widgets. A good example would be the close icon within a modal, toast notification, or alert.
```
adamhock, line 58: Should there be links to Modal, Toast, and Alert pages?

apps/website/src/pages/docs/button.mdx
```diff
@@ -18,51 +18,92 @@ import * as AllExamples from '~/examples';
...
+  <AllExamples.ButtonSizeExample client:load />
+</LiveExample>
+
+## Button props
 
 <PropsTable path={frontmatter.propsPathDefault} />
 
 ## Split button
 
+Pressing this left side of the button will commit the action. Pressing the right side of the button will open a menu.
```
adamhock, line 80:
```suggestion
Pressing the left side of this button will commit the action. Pressing the right side of the button will open a menu.
```

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,32 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
+
+- Use default button just about everywhere.
+- One high visibility button per section such as within a [dialog](dialog), or above a [table](table).
+- One call-to-action button on a page at a time and should always appear within a dedicated page footer.
+- Borderless buttons should be used in places where a default and icon button could cause too much visual clutter.
+- Icon buttons should not have be high visibility or call-to-action styletypes.
+
+### Labels
+
+Strive for a single action word with a well-understood meaning. Verbs do the best. Remember that all labels will need translation, so if the button absolutely requires a longer string, adhere to the _twentycharacterlimit_ rule.
```
adamhock, line 140: Nitpicking here

```suggestion
Strive for a single action word with a well-understood meaning such as a verb. Remember that all labels will need translation, so if the button absolutely requires a longer string, adhere to the _twentycharacterlimit_ rule.
```

apps/website/src/pages/docs/button.mdx
```diff
@@ -18,51 +18,92 @@ import * as AllExamples from '~/examples';
...
+### High visibility
+
+This is the button to emphasize an action applicable to an area of the page or a [dialog](dialog). High visibility buttons are used to draw the user’s attention to the main action of the page.
+There should only be one high visibility button per area of the user interface.
+
+<LiveExample src='Button.highvisibility.tsx'>
+  <AllExamples.ButtonHighVisibilityExample client:load />
+</LiveExample>
+
+### Call to action
```
adamhock, line 44:
```suggestion
### Call-to-action
```

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,32 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
 
 <PropsTable path={frontmatter.propsPathIdeasButton} />
+
+## Usage
+
+### Style types
+
+- Use default button just about everywhere.
+- One high visibility button per section such as within a [dialog](dialog), or above a [table](table).
+- One call-to-action button on a page at a time and should always appear within a dedicated page footer.
```
adamhock, line 134: Nitpicky

```suggestion
- One call-to-action button on a page at a time, and it should always appear within a dedicated page footer.
```

apps/website/src/pages/docs/button.mdx
```diff
@@ -72,10 +113,32 @@ Pressing this button will always open a menu. Upon selecting an option from the
...
+- Borderless buttons should be used in places where a default and icon button could cause too much visual clutter.
+- Icon buttons should not have be high visibility or call-to-action styletypes.
```
adamhock, line 136:
```suggestion
- Icon buttons should not have high visibility or call-to-action styletypes.
```

adamhock (commented):
I've proofread the documentation and left a few suggestions, but other than that, this looks good to me!

apps/website/src/pages/docs/button.mdx
gretanausedaite: Removed this part as we are handling close buttons for the user.

adamhock (approved):

PR #1159, opened 2023-03-27
expose VisuallyHidden component

mayank99:
## Changes

Improved the CSS for visually-hidden component, added focus behavior using data attribute, removed the "revert" mixin (was not used anywhere), and publicly exposed react component so users can consume it (previously it was internal).

## Testing

Added examples and tested in voiceover+safari.

image

## Docs

Added changeset, jsdocs, and a new docs page with multiple examples.


gretanausedaite (approved):
Very useful!

I could not get screen reader to read "3 stars out of 5". Do you have any tips how to achieve that?

mayank99:
> Very useful!
> 
> I could not get screen reader to read "3 stars out of 5". Do you have any tips how to achieve that?

Tried it in playground just now and it seems to work.

FlyersPh9 (approved):

apps/website/src/pages/docs/visuallyhidden.mdx
```diff
@@ -0,0 +1,33 @@
...
+---
+
+import PropsTable from '~/components/PropsTable.astro';
+import LiveExample from '~/components/LiveExample.astro';
+import Placeholder from '~/components/Placeholder.astro';
+import * as AllExamples from '~/examples';
+
+The `VisuallyHidden` utility is used for hiding content from sighted users, while still keeping it in the accessibility tree for screen readers and others assistive technologies. This is achieved using a [very specific ruleset](https://www.tpgi.com/the-anatomy-of-visually-hidden/) that has been tested across a variety of browsers and assistive technologies.
+
+One common use for this component is to provide a text alternative for icons and symbols. In the following example, sighted users will see the star visual reprentation, while screen reader users will hear "3 stars out 5".
```
elephantcatdog, line 17: reprentation -> representation

apps/website/src/pages/docs/visuallyhidden.mdx
```diff
@@ -0,0 +1,33 @@
...
+  <AllExamples.VisuallyHiddenIconExample client:load />
+</LiveExample>
+
+Another use for this component is to add additional text for screen readers to make it easy to differentiate between a set of similar buttons/links, without interfering with the design for sighted users. In the following example, the buttons will be announced by screen readers as "Delete item 1", "Delete item 2", etc.
+
+<LiveExample src='VisuallyHidden.moretext.tsx'>
+  <AllExamples.VisuallyHiddenMoreTextExample client:load />
+</LiveExample>
+
+By default, if an interactive element is passed as children and it gets focused, then it will "unhide". This avoids confusion for keyboard users for whom the focus might otherwise get "lost". If your design already has alternative means to indicate focus, then you can disable this behavior using `unhideOnFocus={false}`.
```
elephantcatdog, line 29: The plurality of "an interactive element" and "children" doesn't seem to match. You could change it to "if an interactive element is passed as a child" or "if interactive elements are passed as chidren".

elephantcatdog (commented):
I understand why you didn't make any visual tests, because it's not supposed to be visible, but maybe having a test that confirms that's the case for regression purposes would be good?

Also I tested it with windows voiceover and it works. 👍

mayank99:
> I understand why you didn't make any visual tests, because it's not supposed to be visible, but maybe having a test that confirms that's the case for regression purposes would be good?

It is already used in other components (e.g. Tile, FileUpload and MenuItemSkeleton), so we have indirect visual tests. And having examples in the docs page is a good place to verify it as well. So I don't think a story is necessary.

elephantcatdog (approved):

PR #1160, opened 2023-03-27
website: Update checkbox documentation & demos

FlyersPh9:
- Add checkbox demos for the different options.

### DIFF
--- a/apps/website/src/examples/Checkbox.indeterminate.tsx
+++ b/apps/website/src/examples/Checkbox.indeterminate.tsx
@@ -8,10 +8,16 @@ import { Checkbox, Flex } from '@itwin/itwinui-react';
 export default () => {
   return (
     <Flex flexDirection='column' alignItems='flex-start'>
-      <Checkbox label='Option 1' defaultChecked />
-      <Checkbox label='Option 2' />
-      <Checkbox label='Option 3' defaultChecked disabled />
-      <Checkbox label='Option 4' disabled />
+      <Checkbox label='Option 1' indeterminate defaultChecked />
+      <Flex
+        flexDirection='column'
+        alignItems='flex-start'
+        style={{ marginLeft: 'var(--iui-size-l)' }}
+      >
+        <Checkbox label='Option 1.1' />
+        <Checkbox label='Option 1.2' defaultChecked />
+        <Checkbox label='Option 1.3' />
+      </Flex>
     </Flex>
   );
 };
--- a/apps/website/src/examples/Checkbox.indeterminate.tsx
+++ b/apps/website/src/examples/Checkbox.indeterminate.tsx
@@ -8,23 +8,10 @@ import { Checkbox, Flex } from '@itwin/itwinui-react';
 export default () => {
   const [option1, setOption1] = React.useState(false);
   const [option2, setOption2] = React.useState(true);
-  const [allOptions, setAllOptions] = React.useState(false);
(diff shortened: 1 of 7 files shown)


apps/website/src/examples/Checkbox.indeterminate.tsx
FlyersPh9, line 1: How can I make it so if the user checks/unchecks the parent becomes indeterminate?  Should I do that for this demo?

FlyersPh9: Thank you @gretanausedaite for the help here!

apps/website/src/examples/Checkbox.visibility.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Checkbox, Flex } from '@itwin/itwinui-react';
+
+export default () => {
+  const [option1, setOption1] = React.useState(false);
+  const [option2, setOption2] = React.useState(true);
+  const [allOptions, setAllOptions] = React.useState(false);
+  const [isIndeterminate, setIsIndeterminate] = React.useState(true);
```
mayank99, line 12: This one doesn't need to be a state.

```

const isIndeterminate = !(option1 && option2) && (option1 || option2);
```

mayank99 (approved):

apps/website/src/examples/Checkbox.indeterminate.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Checkbox, Flex } from '@itwin/itwinui-react';
+
+export default () => {
+  const [option1, setOption1] = React.useState(false);
+  const [option2, setOption2] = React.useState(true);
+  const [allOptions, setAllOptions] = React.useState(false);
```
mayank99, line 11: I've just realized that allOptions also doesn't need to be a state.

You can derive it from option1 and option2:
```tsx
const allOptions = option1 && option2;
```

Doing this will also allow you remove the whole useEffect from lines 14 to 24.

Same comment in Checkbox.visibility

apps/website/src/examples/Checkbox.visibility.tsx
```diff
@@ -0,0 +1,65 @@
...
+        label='Option 1'
+        onChange={(event) => onAllChange(event.target.checked)}
+        indeterminate={isIndeterminate}
+        checked={allOptions}
+        variant='eyeball'
+      />
+      <Flex
+        flexDirection='column'
+        alignItems='flex-start'
+        style={{ marginLeft: 'var(--iui-size-l)' }}
```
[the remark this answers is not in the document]
mayank99: Ah you're right, the parent flex direction is column.

My concern was that it is now referencing a CSS variable that does not come direclty from itwinui-react and could potentially be renamed in the future. But it's fine for now

apps/website/src/examples/Checkbox.indeterminate.tsx
FlyersPh9: Change is made in both indeterminate & visibility demos.

PR #1164, opened 2023-03-28
website: Update radio documentation & demos

idakukimiya:
- Add radio demos for the different options.


apps/website/src/examples/Radio.main.tsx
```diff
@@ -8,8 +8,10 @@ import { Flex, Radio } from '@itwin/itwinui-react';
 export default () => {
   return (
     <Flex flexDirection='column' alignItems='flex-start'>
-      <Radio label='Apple' value='apple' name='fruit' />
-      <Radio label='Orange' value='orange' name='fruit' />
+      <Radio label='Option 1' defaultChecked />
+      <Radio label='Option 2' />
+      <Radio label='Option 3' defaultChecked disabled />
+      <Radio label='Option 4' disabled />
```
FlyersPh9, line 14:
```suggestion
      <Radio name='Options' label='Option 1' defaultChecked />
      <Radio name='Options' label='Option 2' />
      <Radio name='Options' label='Option 3' disabled />
```
A few changes suggested here:
1. Because radios can only have 1 checked item, we can delete line 13 because line 11 is already set active.
2. Added `name` so that these 3 radios are treated as a group, forcing only 1 radio to be checked. This was not documented previously, so is a nice catch & inclusion for this demo!

apps/website/src/pages/docs/radio.mdx
```diff
@@ -20,11 +20,17 @@ import * as AllExamples from '~/examples';
...
+A radio is not interchangeable with a [checkbox](checkbox). A radio must have at minimum 2 choice options and only one option can be checked. A checkbox, however, can have as many or as few items checked, and can be displayed as standalone or within a group of two or more.
 
-In the event when by default, there are no buttons selected in a list when a user enters a page, selecting one would make restoring to the empty default impossible. Then the list of radio buttons would behave like it normally would.
+### Statuses
+
+<LiveExample src='Radio.statuses.tsx'>
+  <AllExamples.RadioStatusesExample client:load />
+</LiveExample>
+
+If you ever need to display the status or severity of a single checkbox, you can apply a status on the checkbox. Before doing so, consider if using an [input group](inputgroup) with a status message makes more sense.
```
FlyersPh9, line 33: I think you missed some text to replace:
```suggestion
If you ever need to display the status or severity of a single radio, you can apply a status on the radio. Before doing so, consider if using an [input group](inputgroup) with a status message makes more sense.
```

FlyersPh9 (approved):

apps/website/src/pages/docs/radio.mdx
```diff
@@ -14,17 +14,21 @@ import * as AllExamples from '~/examples';
 
 <p>{frontmatter.description}</p>
 
-<Placeholder componentPageName='input-radio' />
```
FlyersPh9, line 17: Since removing this, we can safely remove the import on line 12:
`import Placeholder from '~/components/Placeholder.astro';`

PR #1165, opened 2023-03-28
doc(Carousel): Updated carousel website page

elephantcatdog:
## Changes
I added two missing examples from storybook - controlled and only dots. I also added some text to explain them.

Something new that I added to this one is the `Subcomponents` section. I think this is important to have because the subcomponents aren't well documented anywhere except in our code. I think we should have this section on all pages for components that have subcomponents like header and table. One of my coworkers was talking about how great iTwinUI has gotten but how the website needed to be faster and hooray it is, and how documentation is needed for info about subcomponents and their properties.

The properties tables still aren't working for carousel, or for its subcomponents.

## Testing

I checked that the examples worked correctly on the page. In the `onlyDots` example I had to tweak the values for `diff` to account for the sidenav, scrollbar, and page reorg for resizing. `clientX` is the x for the whole page on the doc website rather than for just the component/iframe as in storybook.

## Docs
N/A


mayank99 (approved):

elephantcatdog:
I looked into updating react-docgen to the most recent version (v6 beta5) to see if it would fix the issues with properties tables, but it didn't have any effect.

apps/website/src/pages/docs/carousel.mdx
```diff
@@ -25,8 +25,54 @@ located below the slides, consisting of "dots" and "previous"/"next" buttons, us
...
+
+<PropsTable path={frontmatter.propsPath} />
+
+## Subcomponents
+
+This component uses a composition approach so it should be used with the provided subcomponents.
+
+### Slider
+
+`CarouselSlider` is the scrollable list that should consist of `CarouselSlide` components.
```
gretanausedaite, line 54: Shouldn't these components be `Carousel.Slider` because users see them this way?

gretanausedaite (approved):

apps/website/src/pages/docs/carousel.mdx
elephantcatdog: Yep, thank you!

PR #1169, opened 2023-03-29
fix(Toast): Allow toast message to break

elephantcatdog:
## Changes
Added css that allows toast messages to break rather than overflow as in this example: https://stackblitz.com/edit/react-ts-tyyqu4?file=App.tsx

Before: 
image
After: 
image

## Testing

Tested that long words and short words behave as expected in the toast. The long words should break if necessary but short words should not unecessarily.

## Docs
Added changeset

### DIFF
--- a/.changeset/weak-carrots-cheat.md
+++ b/.changeset/weak-carrots-cheat.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Allows toast messages to break when words are long.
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):
Wow, surprised this hasn't come up yet!  It looks like the same issue might be present in Alert as well.

elephantcatdog:
> It looks like the same issue might be present in Alert as well.

@FlyersPh9 Made a PR for that too: https://github.com/iTwin/iTwinUI/pull/1172

gretanausedaite (approved):

PR #1170, opened 2023-03-29
feat(Table): New story for custom filters

elephantcatdog:
## Changes
I added a story that uses a custom column filter. A user asked if this was possible so I went ahead and made an example for them and anyone in the future interested in it.

image

*This was the alternative method I initially used to make a manual filter to go alongside the manual paginator when I updated the manual paginator story, so I'm happy that a version of it is going to be useful. This version is simplified and doesn't have a paginator. Users can combine it with the manual paginator story if they need that.*

## Testing
Added a new test image.

I clicked around on the example and made sure that everything worked as I intended.

## Docs
Added a new storybook.


mayank99:
The user asked for something else, so this story is no longer necessary.

elephantcatdog:
> The user asked for something else, so this story is no longer necessary.

@mayank99 True, but it's still an example for something we don't already have.

mayank99 (approved):

gretanausedaite (approved):

PR #1177, opened 2023-03-30
Virtual scroll update

LostABike:
## Changes

Make virtual scroll work by incorporating @veekeys logic plus some debugging, also with help from @gretanausedaite. Aside from the logic, I think the glitchy scroll bug was because we used `childHeight.current.middle` to calculate visible nodes, and this once again assumes all nodes are of the same height. So I replaced that with the `getAverageChildHeight()` function, I'm not sure if this is the right approach, need sanity check on the whole logic of this code as it is a lot.

## Testing

- Manual testing of functionality in Storybook:
   +  Lazy loading table
   +  Virtualization table
   +  Virtualized ComboBox


veekeys:
I have pushed some changes. Works OK with that example of lazy load table we have. Of course, need better testing with more scenarios.
Some points to mention:

- `onResize` is being called on scroll. That should not happen, so need to investigate, because it results on double calculations. I think it happens because `updateChildHeight` changes ref and observer gets recreated.. that should not happen that way
- I havent implemented heights map for `scrollToIndex` calculations, but there logic is same - just take original heights from items OR take the average one
- There is MAJOR issue, if you scroll slower you will see there is little blinking/rerendering happening whenever first visible row top border touches the body top (basically first start index changes). But I have noticed, that this is happening already in deployed version! From little debugging, I havent noticed anything, which could be impacting this (I tried commenting out resize observer, did not help). With that little blink you can notice, that some other elements were rendered at first and then it renders the correct ones. So there is no miscalculations, but something strange is happening. 

I think, we should concentrate on finding when that bug started to appear and try to find the place impacting it, cause last time I was working on VS, I do not remember this thing being present.

Oh, the idea about dynamic heights in short:
We are collecting all rendered items heights into the map and apply actual row heights on all the calculations - so that is not a guess, but the correct math. The issue though is that there are a lot of items, which we havent rendered, as a result, we have no idea what the height of the item is going to be. So for these, we are taking all known heights and getting the average value which we apply. So on every scroll, our calculations gonna improve as we get new rendered rows.

PR #1178, opened 2023-03-31
website: Textarea

gretanausedaite:
## Changes

Textarea of #961

## Testing

N/A

## Docs

N/A


FlyersPh9:
Why does the resize handle / scrollbar look so different between the docs site vs. the Storybook & HTML demo?

[HTML demo](https://itwin.github.io/iTwinUI/css/textarea.html):
Screenshot 2023-03-31 at 2 19 59 PM

[Storybook](https://itwin.github.io/iTwinUI/react/?path=/story/input-textarea--basic):
Screenshot 2023-03-31 at 2 22 13 PM

Docs site:
Screenshot 2023-03-31 at 2 20 07 PM

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -24,6 +21,36 @@ Text areas (or multi-line fields) are larger text fields that can accommodate mo
...
+
+## Variants
+
+### Status
+
+<LiveExample src='Textarea.status.tsx'>
+  <AllExamples.TextareaStatusExample client:load />
+</LiveExample>
+
+If status messages are required, the styling is once again similar to [inputs](input). For the default textarea, the bottom border gains the color of the alert level, as well as the corresponding icon and message.
```
elephantcatdog, line 38: You wrote "once again" here but I don't see a first mention of how the styling is similar to inputs? I think it would make sense to omit that and just write "the styling is similar to input".

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -24,6 +21,36 @@ Text areas (or multi-line fields) are larger text fields that can accommodate mo
 
 It used to be possible to freely resize the text area horizontally in addition to vertically; this pattern is now deprecated, and text areas only allow vertical resizing for better predictability. The resize handle is optional; it may be omitted if resizing of the text area is not desired.
```
elephantcatdog, line 22: How long ago was `Textarea` able to be resized horizontally? This part seems unneccessary to me.

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -24,6 +21,36 @@ Text areas (or multi-line fields) are larger text fields that can accommodate mo
...
+
+If status messages are required, the styling is once again similar to [inputs](input). For the default textarea, the bottom border gains the color of the alert level, as well as the corresponding icon and message.
+
+### Inline
+
+<LiveExample src='Textarea.inline.tsx'>
+  <AllExamples.TextareaInlineExample client:load />
+</LiveExample>
+
+It’s possible to use textareas inline, in forms for example. he status displays a bit differently. For these textareas the status border remains the same, but the icon displays horizontally centered inside the area, at the far right.
```
elephantcatdog, line 46: Is this level of depth needed for our documentation site, the describing how it looks? Because users no longer have to make their stuff look how we want; it already does. But if you leave it, there's a "T" missing at the beginning of the second sentence.

But it could be something like "It's possible to use text areas inline. This could be useful in styling forms."

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -24,6 +21,36 @@ Text areas (or multi-line fields) are larger text fields that can accommodate mo
...
+
+## Usage
+
+Use multi-line text areas when a form requires extensive information to be entered by the user. Example: Incident reports, daily logs, details of events, etc.
```
elephantcatdog, line 50: I think the examples would be better written as a sentence like "Examples include incident reports, daily logs, and details of events."

apps/website/src/examples/Textarea.scroll.tsx
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Label, Textarea } from '@itwin/itwinui-react';
+
+export default () => {
+  const [value, setValue] = React.useState(
+    'If the area is shrunk to a size smaller than the amount of text within, a scrollbar appears to keep navigation possible. This behavior is especially important for text areas without a resizing handle, to avoid making text editing tedious.'
```
elephantcatdog, line 10: It seems a bit repetitive to me to have this explanation inside the text box and outside of it, instead of having filler text inside the box. I'm curious why you did it like this?

elephantcatdog (commented):
In some places "textarea" is used and in others "text area" is used. I think we should do it the same way throughout the page. I prefer with the space, but whatever is the right way if that's already set.

apps/website/src/pages/docs/textarea.mdx
gretanausedaite: Removed the once again.

apps/website/src/pages/docs/textarea.mdx
gretanausedaite: Updated that sentence.

apps/website/src/pages/docs/textarea.mdx
gretanausedaite: Removed whole thing

apps/website/src/examples/Textarea.scroll.tsx
gretanausedaite: I don't think anyone reads what's inside the example.

I can update that to lorem ipsum or some other random text, if you prefer that way.

gretanausedaite:
> Why does the resize handle / scrollbar look so different between the docs site vs. the Storybook & HTML demo?

Maybe we are overriding scrollbar width somewhere in the app. It looks wider everywhere. Added an issue in #941

apps/website/src/examples/Textarea.scroll.tsx
elephantcatdog: That makes sense; I was looking at it really closely to review it.

apps/website/src/pages/docs/textarea.mdx
```diff
@@ -9,20 +9,45 @@ group: inputs
...
+
+## Variants
+
+### Status
+
+<LiveExample src='Textarea.status.tsx'>
+  <AllExamples.TextareaStatusExample client:load />
+</LiveExample>
+
+If status messages are required, the styling is similar to [inputs](input). For the default textarea, the bottom border gains the color of the alert level, as well as the corresponding icon and message.
```
elephantcatdog, line 36: This comment is just a reminder about "textarea" vs "text area".

gretanausedaite: Updated "textarea" to be "text area"

FlyersPh9 (approved):

elephantcatdog (commented):
Should the title and thumbnail (lines 2 & 6 of textarea.mdx) be "Text area"?

gretanausedaite:
> Should the title and thumbnail (lines 2 & 6 of textarea.mdx) be "Text area"?

Thumbnail "TextArea" is the name of the image. Will not change that.
Title Textarea matches our component name - leaving that as well.

mayank99:
fyi, "textarea" is the canonical name, all one word. you can see the [MDN docs on textarea](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea), it does not use "text area" or "TextArea" anywhere.

elephantcatdog (approved):
> fyi, "textarea" is the canonical name, all one word. you can see the MDN docs on textarea, it does not use "text area" or "TextArea" anywhere.

I didn't know this. Greta, if you think it's better to have the text be "textarea" that's fine with me. I just mostly wanted it to be consistent.

PR #1181, opened 2023-04-03
fix changesets format

mayank99:
## Changes

Updated changesets config to use a custom script instead of [`@changesets/changelog-git`](https://github.com/changesets/changesets/blob/main/packages/changelog-git/src/index.ts).

Two main changes:
- will show PR number instead of commit number
- hopefully will not show duplicate entries for updated dependencies
 
## Testing

Tested by running `yarn changeset version` before today's release.

## Docs

N/A

### DIFF
--- a/.changeset/changelog-github.cjs
+++ b/.changeset/changelog-github.cjs
@@ -1,0 +1,56 @@
+/** @type {import('@changesets/types').ChangelogFunctions} */
+module.exports = {
+  getReleaseLine: async ({ commit, summary }, type, options) => {
+    checkOptionsRepo(options);
+
+    const [firstLine, ...futureLines] = summary
+      .split('\n')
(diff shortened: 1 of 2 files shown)


gretanausedaite (approved):
Could you add some comments on what is happening? Might be easier to understand at a first glance.

mayank99:
What part is not clear? It is almost identical to the default [`changelog-git`](https://github.com/changesets/changesets/blob/main/packages/changelog-git/src/index.ts) script, other than the two main changes listed in the description

gretanausedaite:
Yeah, it is almost identical. If we add script to our repo, I think it would be beneficial to our team members to have comments (even though there are no comments in the original script) to understand which line does what without having to analyse it and put a lot of effort into understanding.
If you don't want to, that's ok.

PR #1185, opened 2023-04-05
website: Add related components section

gretanausedaite:
## Changes

Added component to show related components cards list. Currently added only to ButtonGroup.

image

## Testing

Visual testing, also voiceOver testing.

## Docs

N/A

### DIFF
--- a/apps/website/src/components/RelatedComponents.astro
+++ b/apps/website/src/components/RelatedComponents.astro
@@ -1,0 +1,71 @@
+---
+import Card from './Card.astro';
+
+export type Props = {
+  components: { title: string; url: string }[];
+};
+
+const { components } = Astro.props;
+---
+
+<ul>
+  {
+    components.map((page) => (
+      <li class='card'>
+        <h3>
+          <a href={page.url}>{page.title}</a>
+        </h3>
+      </li>
+    ))
+  }
+</ul>
+
+<style lang='scss'>
+  ul {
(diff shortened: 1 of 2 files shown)


apps/website/src/components/RelatedComponents.astro
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const { components } = Astro.props;
+---
+
+<ul>
+  {
+    components.map((page) => (
+      <li class='card'>
+        <h3>
```
mayank99, line 15: it doesn't need to be a heading

apps/website/src/components/RelatedComponents.astro
```diff
@@ -0,0 +1,71 @@
...
+import Card from './Card.astro';
+
+export type Props = {
+  components: { title: string; url: string }[];
+};
+
+const { components } = Astro.props;
+---
+
+<ul>
```
mayank99, line 11: forward the rest props here

gretanausedaite: Forwarding rest props there.

apps/website/src/components/RelatedComponents.astro
gretanausedaite: Removed a heading.

mayank99 (approved):

FlyersPh9 (approved):

PR #1186, opened 2023-04-05
docs: Update node version

veekeys:
## Changes

 Updated contributing guide and package.json with node version. I was trying to run project locally with node16 and could not run it, getting some random storybook errors. Took time to guess its node version issue.


mayank99 (approved):
I think you should be able to run it with Node 16 if you use `NODE_OPTIONS=--openssl-legacy-provider`. But yeah, no reason not to bump minimum. 👍

mayank99:
> > I think you should be able to run it with Node 16 if you use `NODE_OPTIONS=--openssl-legacy-provider`. But yeah, no reason not to bump minimum. 👍
> 
> Thats the point, I could not. image
> 
> We do have this option provided in scripts already.

Oh, so you *need* `NODE_OPTIONS=--openssl-legacy-provider` for making storybook 6 work with node 18, but you *can't* even run it in node 16 if you use flag.

Ideally we should not even use this flag so that it works in both node 16 and 18, but that will require an upgrade to storybook 7. But like I said, I'm fine with bumping the minimum to 18 because it is the current LTS.

gretanausedaite (approved):

package.json
```diff
@@ -73,8 +73,8 @@
     "autoprefixer": "10.4.13"
   },
   "engines": {
-    "npm": ">=7.0.0",
-    "node": ">=16.0.0"
+    "npm": ">=9.0.0",
+    "node": ">=18.0.0"
```
mayank99, line 77: Looks like the tests are not being run anymore
image

Probably will nee to update dockerfile and publish new image. (or we could leave it as node 16 engine in package.json but keep 18 in contributing.md)

veekeys: ahh, its docker change :O
And i wanted to collect easy contribution :D
well ok, what you want me to do?

mayank99: Either: update node version in dockerfile and publish new images (could be done in a separate PR)

Or: revert the package.json change, and leave the contributing.md change.

Take your pick :)

mayank99:
When #1188 gets merged, this PR can probably be closed as we won't require Node 18 anymore.

veekeys:
> When #1188 gets merged, this PR can probably be closed as we won't require Node 18 anymore.

Well, after verification, will need to update docs at least anyway

package.json
mayank99: I'd suggest the second option:
> revert the package.json change, and leave the contributing.md change.

veekeys: done, but ideally, would be cool to have here also correct versions, if node 16 fails (errors of those usually are unclear whereas having engine set correctly would deliver understandable error)

PR #1189, opened 2023-04-05
doc(ColorPicker): Updated documentation page

elephantcatdog:
## Changes

- Updated color picker documentation page.
- Added two properties to `LiveExample`
  - `hasDemoCode` allows you to remove the demo code from the page. The example and code can still be opened in a new sandbox page.
    - I wanted to remove the code from the examples I made for the subcomponents as they're just small pieces. I mostly wanted a picture for them but don't want to actual pictures as they could change, etc.
  - `maxHeight` allows you to change the max height if necesasry
    - I wanted to change the max height so that the 'advanced' example wouldn't have a vertical scroll bar
    - Before: 
image

It might be better to just increase the max height, but I thought I'd put this together to see. Maybe add an option to decrease the min height also? Might be good for small 'pictures' like the subcomponents section.

## Testing

Tested that:
- Examples work correctly
- New properties for LiveExample work correctly

## Docs

N/A


apps/website/src/components/LiveExample.astro
```diff
@@ -16,9 +16,21 @@ export type Props = {
...
+   * whether the live example will have a visible code snippet
+   * @default true
+   */
+  hasDemoCode?: boolean;
+
+  /**
+   * use to set the max height if it needs to be taller than the default
+   * @default '400px'
+   */
+  maxHeight?: string;
```
mayank99, line 30: I don't think this needs to be a prop. You should be able to increase the max-height in the CSS for all demos to 500px.

apps/website/src/pages/docs/colorpicker.mdx
```diff
@@ -14,16 +14,82 @@ import * as AllExamples from '~/examples';
...
+
+## Subcomponents
+
+The color picker is extremely flexible; pieces can be used based on product requirements. If certain parts are not required for your context, they can be left out. There are no hard guidelines for what an advanced color picker contains. It should remain usable and align with the user’s workflow.
+
+### Color Builder
+
+`ColorBuilder` consists of two parts: a color canvas to adjust saturation and lightness values, and a set of sliders to adjust hue and alpha values.
+
+<LiveExample src='ColorPicker.builder.tsx' hasDemoCode={false}>
```
mayank99, line 61: I don't think it makes sense to have demos for all subcomponents. These subcomponents are not meant to be used by themselves. So the `hasDemoCode` prop is also not necessary.

What would be more helpful: an annotated screenshot of a ColorPicker, visually mapping the different parts of the UI to the names of the subcomponents.

elephantcatdog: Yeah, but I thought we weren't putting screenshots on our site because they can't keep up with updates, while the demoes do.

apps/website/src/components/LiveExample.astro
elephantcatdog: Updated

apps/website/src/pages/docs/colorpicker.mdx
mayank99: Then just remove these subcomponent demos for now. This decision can be made at a later point and shouldn't block this PR.

elephantcatdog: Updated

mayank99: the prop can be removed too

apps/website/src/pages/docs/colorpicker.mdx
```diff
@@ -14,16 +14,66 @@ import * as AllExamples from '~/examples';
...
+
 ## Props
 
 <PropsTable path={frontmatter.propsPath} />
+
+## Subcomponents
+
+The color picker is extremely flexible; pieces can be used based on product requirements. If certain parts are not required for your context, they can be left out. There are no hard guidelines for what an advanced color picker contains. It should remain usable and align with the user’s workflow.
+
+### Color Builder
```
mayank99, line 57: Maybe a nit: I think using these headings should match the exact names of the components, i.e. `ColorBuilder` not "Color Builder"

mayank99 (approved):

apps/website/src/pages/docs/colorpicker.mdx
```diff
@@ -14,16 +14,66 @@ import * as AllExamples from '~/examples';
...
 <p>{frontmatter.description}</p>
 
-<Placeholder componentPageName='core-colorpicker' />
-
 <LiveExample src='ColorPicker.main.tsx'>
   <AllExamples.ColorPickerMainExample client:load />
 </LiveExample>
 
-The color picker is extremely flexible. Bits and pieces can be used based on product requirements. If certain parts are not required for your context, they can be left out.
+Color pickers allow users to browse and apply a color to a selected item. The color picker is extremely flexible. Bits and pieces can be used based on product requirements. If certain parts are not required for your context, they can be left out.
```
adamhock, line 21: This section is written twice: once here and then again in the Subcomponent section

apps/website/src/pages/docs/colorpicker.mdx
```diff
@@ -14,16 +14,66 @@ import * as AllExamples from '~/examples';
...
+
+### Advanced
+
+The advanced color picker lets the user pick any color. You may display user saved colors or recently used colors if needed. Because the colors are immensely flexible, colors are labeled with HEX, RGB/RGBA, or HSL/HSLA color code values. A clickable icon in the text field allows the user to rotate through the different color code types.
+
+<LiveExample src='ColorPicker.advanced.tsx'>
+  <AllExamples.ColorPickerAdvancedExample client:load />
+</LiveExample>
+
+The color code can be displayed in editable fields for precise color selection. There are two ways to show this code: Within the panel itself as seen in the example above, or in a dropdown button/input field combo, like below. If the color code is shown outside of the panel, do not include the color code within the panel at the same time.
```
adamhock, line 43:
```suggestion
The color code can be displayed in editable fields for precise color selection. There are two ways to show this code: within the panel itself as seen in the example above, or in a dropdown button/input field combo, like below. If the color code is shown outside of the panel, do not include the color code within the panel at the same time.
```

apps/website/src/pages/docs/colorpicker.mdx
```diff
@@ -14,16 +14,66 @@ import * as AllExamples from '~/examples';
...
+
+`ColorPalette` is used to show a group of `ColorSwatch` components.
+
+<PropsTable path={'@itwin/itwinui-react/esm/core/ColorPicker/ColorPalette.d.ts'} />
+
+### Color Swatch
+
+`ColorSwatch` is a component to display within a color palette.
+
+<PropsTable path={'@itwin/itwinui-react/esm/core/ColorPicker/ColorSwatch.d.ts'} />
```
adamhock, line 79: The props for these subcomponents can be removed.

elephantcatdog: I think these props will be helpful (when they work)

apps/website/src/pages/docs/colorpicker.mdx
elephantcatdog: Fixed

apps/website/src/pages/docs/colorpicker.mdx
elephantcatdog: Removed

PR #1193, opened 2023-04-07
improve website typography

mayank99:
## Changes

- Increased the spacing between sections by using a slightly larger value and by using `em` instead of `rem`.
- Fixed the colors so that heading text looks brighter than paragraph text.
- Updated the heading font size to use the fluid typography scale that we already have in place.

All of these changes help address https://github.com/iTwin/iTwinUI/issues/941#issuecomment-1500391356

## Testing

Tested locally.

## Docs

N/A


FlyersPh9 (approved):

elephantcatdog (approved):
Looks great!

PR #1210, opened 2023-04-13
remove default tooltip role from popover

mayank99:
## Changes

Noticed that popovers were incorrectly getting `role='tooltip'` which is what Tippy.js defaults to, so removed it. Role must be manually specified for each case.

## Testing

Tested by inspecting the DOM.

## Docs

N/A

### DIFF
--- a/.changeset/cuddly-moons-invite.md
+++ b/.changeset/cuddly-moons-invite.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Removed incorrect `role=tooltip` from Popover, DropdownMenu, Select and ComboBox.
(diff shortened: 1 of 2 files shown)


LostABike (approved):

FlyersPh9 (approved):
LGTM.  Should `role=tooltip` be added to [the tooltip HTML demo page](https://github.com/iTwin/iTwinUI/blob/909fdbc91cd878cdc7b8856539aa2898f404b48f/packages/itwinui-css/backstop/tests/tooltip.html#L69)?

mayank99:
> LGTM. Should `role=tooltip` be added to [the tooltip HTML demo page](https://github.com/iTwin/iTwinUI/blob/909fdbc91cd878cdc7b8856539aa2898f404b48f/packages/itwinui-css/backstop/tests/tooltip.html#L69)?

The semantics of the HTML demo pages don't matter for the most part, because they are only meant to show styling. We do have `role=tooltip` in the actual react component.
https://github.com/iTwin/iTwinUI/blob/909fdbc91cd878cdc7b8856539aa2898f404b48f/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx#L45-L48

gretanausedaite (approved):

PR #1211, opened 2023-04-14
fix(Table): Date filter checks

adamhock:
## Changes

The date range picker used in date filters now prevents users from picking a start date later than the end date or picking an end date earlier than the start date. When selecting a `From` date, any dates after a given `To` date are disabled. When selecting a `To` date, any dates before a given `From` date are disabled. 

## Testing

1. Open the `Filters` story in the Table storybook
2. Click the date filter button from the `Start date` or `End date` columns
3. Set the `From` date
4. Open the `DatePicker` for `To` date and notice how dates before the `From` date are disabled
5. Set the `To` date
6. Open the `DatePicker` for `From` date and notice how dates after the `To` date are disabled

## Docs

Changeset added


mayank99:
The idea makes sense, but I think the experience is a bit confusing because it doesn't respect the date that is actually being selected.

https://user-images.githubusercontent.com/9084735/232584061-72e8f72c-4339-456f-95f6-9038acb27765.mov

It might be better to instead use the `isDateDisabled` prop to disable the out-of-bounds dates.

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -117,25 +118,33 @@ export const DateRangeFilter = <T extends Record<string, unknown>>(
...
-        label={translatedStrings.to}
-        date={to}
-        onChange={onToChange}
-        formatDate={formatDate}
-        parseInput={parseInput}
-        onKeyDown={onKeyDown}
-        placeholder={placeholder}
-      />
+      <DateRangeFilterContext.Provider
+        value={{ datePicker: 'from', selectedDate: to }}
```
mayank99, line 122: why not just pass these as props instead of context?

mayank99: also unit tests are failing because of useSafeContext

adamhock: This has been fixed

mayank99 (approved):

mayank99:
add unit tests too

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -13,10 +13,34 @@ export type DatePickerInputProps = {
   onChange: (date?: Date) => void;
   parseInput: (text: string) => Date;
   formatDate: (date: Date) => string;
+  /**
+   * Decides if this component is used for the 'from' or 'to' date
+   */
+  datePicker: 'from' | 'to';
```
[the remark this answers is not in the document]
adamhock: `datePicker` has changed to `isFromOrTo` (I'm open to name suggestions)

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.test.tsx
```diff
@@ -127,3 +127,47 @@ it('should call onChange with undefined when input field is cleared', async () =
...
+it('should disable dates before "from" date when using "to" date picker', async () => {
+  const fromDate = new Date(2023, 3, 22);
+  const onClick = jest.fn();
+  const { container, getByText } = renderComponent({
+    isFromOrTo: 'to',
+    selectedDate: fromDate,
+  });
+
+  const iconButton = container.querySelector(
+    '.iui-input-icon.iui-button[data-iui-variant="borderless"]',
```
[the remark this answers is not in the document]
adamhock: Fixed

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.test.tsx
```diff
@@ -127,3 +127,47 @@ it('should call onChange with undefined when input field is cleared', async () =
...
+  ) as HTMLButtonElement;
+  expect(iconButton).toBeTruthy();
+
+  await userEvent.click(iconButton);
+  const calendar = container.querySelector('.iui-date-picker');
+  expect(calendar).toBeTruthy();
```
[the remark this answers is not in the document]
adamhock: Fixed

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -13,10 +13,34 @@ export type DatePickerInputProps = {
   onChange: (date?: Date) => void;
   parseInput: (text: string) => Date;
   formatDate: (date: Date) => string;
+  /**
+   * Decides if this component is used for the 'from' or 'to' date
+   */
+  isFromOrTo?: 'from' | 'to';
```
gretanausedaite, line 19: Is this done for a reason or just because?
I think `isStartDate = true` for "from" and false for "to" could be enough.

gretanausedaite (approved):

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
mayank99: see history: https://github.com/iTwin/iTwinUI/pull/1211#discussion_r1172602567

there are many ways to do this. one option is to have two props `upperBound`/`lowerBound` that simply accept dates.
```
<DatePickerInput lowerBound={new Date()} />
```
```
<DatePickerInput upperBound={new Date()} />
```
or even combine the two props into one (questionable lol)
```
<DatePickerInput bound={{ type: 'lower', date: new Date() }} />
```

gretanausedaite: Maybe `useAs? : 'from' | 'to'` sound better than isFromOrTo?

mayank99: - too vague
- `use` prefix is for hooks
- `as` is generally used to control the element which it is rendered as (polymorphism)

ultimately it doesn't matter because this component is private, so i'm fine with all of the suggestions. My top preference is for `upperBound`+`lowerBound` (remove `isFromOrTo` and `selectedDate`).

@gretanausedaite Maybe you can also list your top preference, and then we can let @adamhock make the final decision?

gretanausedaite: My preferences:
- leave as is
- Mayank's suggestion also sounds good.

PR #1213, opened 2023-04-17
patch storybook vite builder for windows

mayank99:
## Changes

Used `patch-package` to add `.replaceAll('\\', '/')` in one of the imports in `@storybook/builder-vite`. This *might* fix windows issues.

## Testing

On windows, pull this branch and run `yarn`. You should see this message:
```
$ patch-package
patch-package 6.5.1
Applying patches...
@storybook/builder-vite@7.0.2 ✔
```

Then run storybook as normal.

## Docs

N/A


LostABike (approved):
I tested this on windows and everything appears to be OK:
image

I am able to run storybook locally as well

mayank99:
> I am able to run storybook locally as well

@LostABike Just to double check that this fix actually works, can you confirm that you cannot run storybook outside this branch? (Switch to `main` branch, then run `yarn`, then run storybook)

LostABike:
> > I am able to run storybook locally as well
> 
> @LostABike Just to double check that this fix actually works, can you confirm that you cannot run storybook outside this branch? (Switch to `main` branch, then run `yarn`, then run storybook)

Yea it appears I cannot run Storybook on main branch, with the same error that @elephantcatdog posted. So I think this fix should work.

PR #1228, opened 2023-04-20
multiselect announcements

mayank99:

### DIFF
--- a/packages/itwinui-css/backstop/tests/assets/selectTag.js
+++ b/packages/itwinui-css/backstop/tests/assets/selectTag.js
@@ -16,7 +16,7 @@ class SelectTag extends HTMLElement {
     const dismissible = this.hasAttribute('dismissible');
 
     const innerHtml = `
-    <span
+    <li
       class="iui-select-tag"
     >
       <span
@@ -32,7 +32,7 @@ class SelectTag extends HTMLElement {
           </button>`
           : ''
       }
-    </span>
+    </li>
     `;
     this.parentElement.insertAdjacentHTML('beforeend', innerHtml);
     this.remove();
--- a/packages/itwinui-css/src/select/select.scss
+++ b/packages/itwinui-css/src/select/select.scss
@@ -69,6 +69,8 @@
 }
 
 @mixin iui-select-tag-container {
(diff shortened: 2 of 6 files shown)


packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -411,6 +414,14 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
         const newArray = selectedChangeHandler(__originalIndex, actionType);
         dispatch({ type: 'multiselect', value: newArray });
         onChangeHandler(__originalIndex, actionType, newArray);
+
+        // update live region
+        setLiveRegionSelection(
```
gretanausedaite, line 419: Would it be usefull to add 'Selected items' for screen reader to announce before announcing a list of items? It might be confusing why it is listing random numbers.

gretanausedaite (commented):
Looks awesome!

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: maybe. but then it would need to be localized (one more thing devs have to worry about)

packages/itwinui-react/src/core/Select/Select.test.tsx
```diff
@@ -514,3 +514,41 @@ it('should use custom render for selected item (multiple)', async () => {
...
+          setSelected((prev) =>
+            type === 'added'
+              ? [...prev, value]
+              : prev.filter((v) => v !== value),
+          )
+        }
+      />
+    );
+  };
+  const { container } = render(<MultiSelectTest />);
```
mayank99, line 540: It was way harder than it should have been to write this test. Our Select doesn't maintain its own state - we should change that in the future

gretanausedaite: I think we have an issue to update tests.

mayank99: it's not a problem with tests though - it's the component that needs to be improved

PR #1232, opened 2023-04-24
set aria-pressed on IconButton when isActive is true

mayank99:
## Changes

`aria-pressed` is designed specifically for toggle buttons so it makes sense to use it in icon buttons.

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

## Testing

tested in voiceover, it's announced as "selected, toggle button"

tested in NVDA, it's announced as "toggle button pressed"

also added unit test

## Docs

Not sure if we need to document this.


LostABike (approved):
It is working correctly on Chrome. Everything looks good to me.

gretanausedaite (approved):
Why only for IconButtons?

PR #1235, opened 2023-04-25
use implicit id association in LabeledInput & LabeledTextarea

mayank99:
## Changes

Changed the internal DOM structure of `LabeledInput` and `LabeledTextarea` to prefer explicit association over implicit. The `<label>` is now associated with the input using `htmlFor`/`id` and the container is a generic div. To achieve this, I had to add some checks inside `InputContainer` and a prop to pass the id. I also left out `LabeledSelect` because there is no "input" in there to associate the label with; this can be addressed in a separate PR.

This change improves accessibility, with no API changes and no effect on visuals.

In the next major version, we can further improve `InputContainer` to reuse `Label` and `StatusMessage` components, and remove any duplicate CSS rules.

## Testing

Tested by going through all stories and clicking on them and inspecting the dom.

## Docs

N/A

### DIFF
--- a/.changeset/witty-ties-dream.md
+++ b/.changeset/witty-ties-dream.md
@@ -1,0 +1,8 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': patch
+---
+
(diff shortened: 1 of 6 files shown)


adamhock (approved):
LGMT

gretanausedaite (approved):

PR #1238, opened 2023-04-25
change Select trigger to use role=combobox

mayank99:
## Changes

- Changed `Select` to render a `<div role=combobox>` instead of a div. This greatly improves semantics so that everything is announced correctly.
- Added new `triggerProps` for passing misc attributes to this div.
- Changed `LabeledSelect` to associate the label (which is also rendered as a `<div>`) with this `<div>` using `aria-labelledby`.

This unblocks #1228 as well.

## Testing

Tested in voiceover and NVDA. The label *and* the selected value are both correctly announced.

The screen recording below is from Voiceover + Safari on macOS.

https://user-images.githubusercontent.com/9084735/235247366-034f7f7a-802f-45f1-abcb-103dee41af87.mov

And this one is from NVDA + Firefox on Windows.

https://user-images.githubusercontent.com/9084735/235254144-cb78ec4f-daf9-4208-af2d-ffe9f3eef906.mov

## Docs

N/A


packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -228,18 +235,16 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
     ...rest
   } = props;
 
   useTheme();
 
-  const [isOpen, setIsOpen] = React.useState(popoverProps?.visible ?? false);
-  React.useEffect(() => {
-    setIsOpen((open) => popoverProps?.visible ?? open);
-  }, [popoverProps]);
+  const [isOpenState, setIsOpen] = React.useState(false);
```
adamhock, line 244: Nitpick, but maybe change `setIsOpen` to `setIsOpenState`

adamhock (approved):
Just one nitpick. Other than that, this looks good to me

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: This is done a bit intentionally because `isOpen` is defined right below it, so `isOpenState` is not used anywhere else in the code

gretanausedaite (approved):

PR #1245, opened 2023-05-03
increase checkbox touch target size

mayank99:
## Changes

Used `::before` element with negative inset to increase checkbox to 24x24 by default. Can be adjusted using the CSS variable `--iui-checkbox-target-size`.

In table, I'm using this variable to make it 32x32. This will prevent accidentally clicking on the row instead of the checkbox.

No visual changes.

## Testing

Tested in demo pages / storybook.

See branch preview: https://itwin.github.io/iTwinUI/1245/react/?path=/story/core-table--selectable-multi

## Docs

N/A

### DIFF
--- a/packages/itwinui-css/src/table/table.scss
+++ b/packages/itwinui-css/src/table/table.scss
@@ -384,6 +384,9 @@
     justify-content: center;
     align-items: center;
     flex-basis: calc(var(--iui-size-l) * 2);
+
+    // make the checkbox bigger (32x32) to prevent accidentally clicking on the row
+    --iui-checkbox-target-size: 32px;
   }
 
   &.iui-table-cell-sticky {
--- a/.changeset/sour-garlics-lay.md
+++ b/.changeset/sour-garlics-lay.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/itwinui-css/src/table/table.scss
```diff
@@ -384,6 +384,9 @@
     justify-content: center;
     align-items: center;
     flex-basis: calc(var(--iui-size-l) * 2);
+
+    // make the checkbox bigger (32x32) to prevent accidentally clicking on the row
+    --iui-checkbox-target-size: 32px;
```
FlyersPh9, line 389: Why hardcoded values and not size vars from iTwinUI-Variables?

mayank99: I could use `iui-size-l` and `iui-size-xl` but i decided against it.

The base checkbox size is already set to `iui-size-m` (1rem) so it will still scale with font size. For example, if the user increases their OS font size to 200%, then the checkbox target size will be: 32px (visible 1rem * 200%) + 16px (invisible negative inset) = **46x46px**, which still meets the 32x32 _minimum_. 

On the other hand, if I use variables/rem, then it will be 32px (visible 1rem * 200%) + 32px (invisible 1rem * 200%) = **64x64px**, which could be _too_ big. We don't want users to accidentally click on the checkbox (inverse of the problem we are currently trying to solve); here is a fine balance with using this negative inset hack.

Also, I find that the code is more clear / less confusing this way

gretanausedaite (approved):

PR #1249, opened 2023-05-04
fix(Breadcrumbs): Remove legacy support for buttons & anchors

elephantcatdog:
## Changes

- Removed mixins `iui-breadcrumbs-text` and `iui-breadcrumbs-button` that were for explicitly classnaming anchor text in a breadcrumb and buttons in a breadcrumb
- Removed related html examples `Breadcrumbs anchor` and `Breadcrumbs button`
- Renamed existing examples from `Generic anchor` to `Breadcrumbs with anchor` and from `.iui-button; overrides` to `Breadcrumbs with buttonss`
- Updated html example `Overflow button` and renamed to `Breadcrumbs with overflow button`
- Removed scenarios for removed examples for css visual tests
- Removed css visual tests

## Testing

- Updated css visual tests
- Ran all tests
- Confirmed that storybook, html examples, and doc site still behave correctly

## Docs

- need to add changeset


mayank99:
This PR is doing the opposite of what it should be doing. `iui-breadcrumbs-item-overrides` is the legacy code that we want to remove.

In the new API, we should only have React subcomponents.
```jsx
<Breadcrumbs>
  <Breadcrumbs.Item>...</Breadcrumbs.Item>
  <Breadcrumbs.Item>...</Breadcrumbs.Item>
  <Breadcrumbs.Item>...</Breadcrumbs.Item>
</Breadcrumbs>
```

PR #1251, opened 2023-05-04
remove extra scss files and migrate to `@use`

mayank99:
## Changes

Repo-wide change in prep for v3. 

- Removed all `index.scss` and `classes.scss` files
  - Most simple components (e.g. badge) have the entire source in a single file
  - Some complex components (e.g. button/table) still have multiple files
  - Components whose mixins are used in other components have one extra file (I would like to reduce such cases in the future)
- Created new `mixins.scss` for all reusable mixins.
  - Originally, the `utils` folder was intended to be used like this, but it started getting cluttered when we put our public-facing code in that folder.
- Changed all `@import`s to `@use`/`@forward`
- No real changes in actual CSS output - mostly just moved things around in source.

## Testing

Went through every single output file to verify contents. Also visual tests pass.

## Docs

N/A


gretanausedaite (commented):
> in prep for v3.

This change will be released as v2.x?

mayank99:
> This change will be released as v2.x?

yes, that's why i'm doing it now - trying to minimize differences between v2 and v3 codebase.

gretanausedaite (approved):

FlyersPh9:
I think `/packages/itwinui-css/src/table/table-densities.scss` can be deleted, I don't see those SASS vars being used anywhere.

PR #1253, opened 2023-05-04
feat(Breadcrumbs): Remove legacy support for button/anchor

elephantcatdog:
## Changes
- Removed mixin `iui-breadcrumbs-item-overrides`
- Removed related html examples `.iui-button; overrides` and `Generic anchor`
- Removed scenarios for removed examples for css visual tests

- Added composition component `Breadcrumbs.Item` for `Breadcrumbs`
  - Has property `type` with options `button`, `anchor`, and `text` for different styles
  - Has property `href` for use with type anchor
  - Has property `onClick` for use with type button
- Removed uses of `iui-breadcrumbs-item-overrides` in Breadcrumbs.tsx

## Testing
- Added unit tests for Breadcrumbs.Item
- * need to check unit tests for Breadcrumbs
- Removed css visual tests for old html examples `.iui-button; overrides` and `Generic anchor`
- Updated breadcrumbs stories with new composition
- Confirmed that react visual tests pass
- Confirmed that react unit tests pass
- Confirmed that css visual tests pass

## Docs
- added changeset
- * need to update doc site


packages/itwinui-react/src/core/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+   */
+  type?: 'anchor' | 'button' | 'text';
+  /**
+   * Include an href for anchor type breadcrumbs.
+   */
+  href?: string;
+  /**
+   * Include an onClick function for button type breadcrumbs.
+   */
+  onClick?: () => void;
```
elephantcatdog, line 65: I think that I should be able to remove the three properties `type`, `href`, and `onClick` here because they're defined above (like in Tabs.tsx). I get an error below when defining the props though that it can't find those properties. Not sure what I'm doing wrong. I think the way that I'm currently doing it isn't good because it doesn't warn that you can't use href with buttons, etc.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -110,74 +111,79 @@ export type BreadcrumbsProps = {
...
+                  </React.Fragment>
+                );
+              })}
+          </ol>
+        </nav>
+      );
+    },
+  ),
+  {
+    Item: BreadcrumbsItem,
```
mayank99, line 186: needs jsdoc here

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -110,74 +111,79 @@ export type BreadcrumbsProps = {
...
-    const {
-      children: items,
-      currentIndex = items.length - 1,
-      separator,
-      overflowButton,
-      className,
-      ...rest
-    } = props;
+export const Breadcrumbs = Object.assign(
+  React.forwardRef(
```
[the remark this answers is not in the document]
mayank99: also should have `displayName`

PR #1255, opened 2023-05-08
Tile composition

LostABike:
## Changes

Convert most Tile's props into polymographic subcomponents with the following structure:
```
    <Tile>
      <Tile.ThumbnailArea>
        <Tile.ThumbnailPicture/>
        <Tile.Badge/>
        <Tile.TypeIndicator/>
        <Tile.QuickAction/>
      </Tile.ThumbnailArea>
      <Tile.Name/>
      <Tile.ContentArea>
        <Tile.Description />
        <Tile.Metadata/>
        <Tile.MoreOptions/>
      </Tile.ContentArea>
      <Tile.Buttons/>
    </Tile>
```

## Testing

- Updated all Storybook.
- Updated all unit tests.

## Docs

- Included comments for each subcomponents export. Some have example.
- Created example code, but documentation page .mdx might need to be in a separate PR since behaviors differ from Storybook on certain code.


LostABike:
@mayank99 can you briefly check if I'm going the right direction? before I try to complete the other subcomponents. Also I made a different file to work on incrementally rather than modifying existing Tile, easier for me this way. Once everything works right I'll deleted old Tile and rename new Tile

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,177 @@
...
+      {...rest}
+    />
+  );
+}) as PolymorphicForwardRefComponent<'div', TileOwnProps>;
+
+type TileOwnProps = {
+  /**
+   * Status of the tile.
+   */
+  status?: 'positive' | 'warning' | 'negative';
```
mayank99, line 62: The main Tile needs to have all the high-level state props, including (in addition to `status`): `isSelected`, `isNew`, `variant`, `isActionable`, `isLoading` and `isDisabled`

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,177 @@
...
+        })
+      ) : React.isValidElement(thumbnail) ? (
+        React.cloneElement(thumbnail, {
+          className: cx('iui-thumbnail-icon', thumbnail.props.className),
+        })
+      ) : (
+        thumbnail
+      )}
+      {leftIcon &&
+        React.cloneElement(leftIcon as React.ReactElement, {
```
mayank99, line 118: most of these dom elements should be subcomponents too. I expect the final markup to look something like:
```
<Tile>
  <Tile.ThumbnailArea>
    <Tile.Thumbnail />
    <Tile.QuickAction>⭐️</Tile.QuickAction>
    <Tile.QuickAction>💬</Tile.QuickAction>

    <Tile.Badges>
     <Badge>badge 1</Badge>
     <Badge>badge 2</Badge>
    </Tile.Badges>
  </Tile.ThumbnailArea>

  <Tile.Name>Hello</Tile.Name>

   ...
</Tile>
```

in general, we want to avoid `cloneElement` and prefer making new subcomponents

LostABike: I'm unsure how we can avoid `cloneElement` in cases where the Element supplied by user have many props or components within it. For example a thumbnail (this is made up, not real use of these components):
```
    <Avatar
      size='x-large'
      status='online'
      abbreviation='TR'
      backgroundColor={getUserColor('Terry Rivers')}
      image={
        <img src='https://itwinplatformcdn.azureedge.net/iTwinUI/user-placeholder.png' />
      }
      title='Terry Rivers'
        <IconButton styleType='borderless'>
            <SvgInfo />
        </IconButton>
    />
```
And this could be something the user made just for the thumbnail picture, which I think `cloneElement` is the best use here anyway?

LostABike: So based on the above, my thumbnail subcomponent now looks like this:

```
const TileThumbnail = React.forwardRef((props, ref) => {
  const { as: Element = 'div', className, thumbnail, ...rest } = props;
  return (
    <Element
      className={cx('iui-tile-thumbnail', className)}
      ref={ref}
      {...rest}
    >
      {thumbnail && typeof thumbnail === 'string' ? (
        <div
          className='iui-tile-thumbnail-picture'
          style={{ backgroundImage: `url(${thumbnail})` }}
        />
      ) : thumbnail && (thumbnail as JSX.Element).type === 'img' ? (
        React.cloneElement(thumbnail as JSX.Element, {
          className: 'iui-tile-thumbnail-picture',
        })
      ) : React.isValidElement(thumbnail) ? (
        React.cloneElement(thumbnail, {
          className: cx('iui-thumbnail-icon', thumbnail.props.className),
        })
      ) : (
        thumbnail
      )}
    </Element>
  );
}) as PolymorphicForwardRefComponent<'div', TileThumbnailOwnProps>;
```

Do you think it should be broken down even more? But either ways where it is a custom element, there would be a `cloneElement` regardless, no?

LostABike: So even if we make a new subcomponent for QuickAction, it would also return a `cloneElement` of whatever is passed in, like how we have it originally. Or if you can show me an example of how to break that down without cloning? a use case: 
```
      <Tile.QuickAction
          leftIcon={
            <IconButton styleType='borderless'>
              <SvgInfo />
            </IconButton>
          }
       />
```

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Added these

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Turned everything into subcomponents

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,698 @@
...
+      ref={ref}
+      {...rest}
+    >
+      {children}
+    </Element>
+  );
+}) as PolymorphicForwardRefComponent<'div', TileThumbnailAreaOwnProps>;
+TileThumbnailArea.displayName = 'TileNew.ThumbnailArea';
+
+type TileThumbnailAreaOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
```
mayank99, line 213:
```suggestion
const TileThumbnailArea = polymorphic.div('iui-tile-thumbnail');
TileThumbnailArea.displayName = 'TileNew.ThumbnailArea';
```

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,698 @@
...
+  );
+}) as PolymorphicForwardRefComponent<'div', TileDescriptionOwnProps>;
+TileDescription.displayName = 'TileNew.Description';
+
+type TileDescriptionOwnProps = {
+  /**
+   * Description text of the tile.
+   * Gets truncated if it can't fit in the tile.
+   */
+  description?: React.ReactNode;
```
mayank99, line 411: doesn't need to be a separate prop, can just use `children`

mayank99:
> The folder variant is relying on the name prop to make its own iui-tile-name within iui-tile-content, should we still keep the name prop and pass it around the subcomponents? Or restructure css to have to have name outside of content even for folder variant?

We can restructure the html/css. It was kept this way in #1125 for backwards compatibility.

packages/itwinui-react/src/core/index.ts
```diff
@@ -295,6 +295,22 @@ export type { TextareaProps } from './Textarea/index.js';
 export { Tile } from './Tile/index.js';
 export type { TileProps } from './Tile/index.js';
 
+export { TileNew } from './Tile/index.js';
+export type { TileNewProps } from './Tile/index.js';
```
adamhock, line 299: All of these types can go under the same set of braces

packages/itwinui-react/src/core/Tile/index.ts
```diff
@@ -5,3 +5,4 @@
 export { Tile } from './Tile.js';
 export type { TileProps } from './Tile.js';
 export default './Tile';
+export * from './TileNew';
```
adamhock, line 8: Changing this to `'./TileNew.js'` should fix the `yarn lint` issue in the PR

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import React from 'react';
+import cx from 'classnames';
+import {
+  useSafeContext,
+  // useTheme,
```
adamhock, line 10: This can be removed. Also, you can import `useGlobals` and call that within `TileComponent`

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+  SvgNew,
+  LinkAction,
+  supportsHas,
+  polymorphic,
+} from '../utils/index.js';
+import '@itwin/itwinui-css/css/tile.css';
+import { ProgressRadial } from '../ProgressIndicators/index.js';
+import { DropdownMenu } from '../DropdownMenu/index.js';
+import { IconButton } from '../Buttons/index.js';
+// import { ProgressRadial } from '../ProgressIndicators';
```
adamhock, line 25: These commented imports can be removed

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+       * Whether the tile is "new". Tile name becomes bold and gets a new status icon.
+       */
+      isNew?: boolean;
+      /**
+       * Default tile variant or the folder layout.
+       * @default 'default'
+       */
+      variant?: 'default' | 'folder';
+      /**
+       * Whether the tile is expected to be interactable (i.e. `onClick`).
```
adamhock, line 46: interactable wasn't coming up as a word in the spellchecker. I think it can be changed to `interactive`. There's another instance in this file where this change can be made as well

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+
+//<Tile>
+//  <Tile.Thumbnail>
+//  <Tile.Name>
+//  <Tile.Content>  //has meta data
+//  <Tile.Buttons>
+//</Tile>
+
+// ----------------------------------------------------------------------------
+// Main Tile component
```
adamhock, line 102: This should go after context and before `Tile.Action`

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+       */
+      isDisabled?: boolean;
+      setActionable: React.Dispatch<React.SetStateAction<boolean>>;
+      onClick?: React.MouseEventHandler<HTMLElement>;
+    }
+  | undefined
+>(undefined);
+TileContext.displayName = 'TileContext';
+
+type TileActionOwnProps = {}; // eslint-disable-line
```
adamhock, line 71: The comment on this line should be changed to the following 

`// eslint-disable-line @typescript-eslint/ban-types`

Also, I think it makes sense for the component props to come before the composition like it does here. Could you change this for the other components which list the props after the component?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+ * Renders `a` element by default.
+ * @example
+ * <Tile
+ *   name={<Tile.Action href='/new-page'>Tile name<Tile.Action/>}
+ * />
+ */
+export const TileAction = (
+  props: PolymorphicComponentProps<'a', TileActionOwnProps>,
+) => {
+  const tileContext = useSafeContext(TileContext);
```
adamhock, line 84: This can be changed to the following code 

```tsx
const {setActionable} = useSafeContext(TileContext)
```

That way, `useEffect` only depends on `setActionable` instead of the entire context

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+TileThumbnailPicture.displayName = 'TileNew.TileThumbnailPicture';
+
+type TileThumbnailPictureOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+// ----------------------------------------------------------------------------
+// Tile.ThumbnailAvatar component
+
+const TileThumbnailAvatar = React.forwardRef((props, ref) => {
+  const { as: Element = 'div', className, children, ...rest } = props;
+  if (React.isValidElement(children)) {
```
adamhock, line 234: I'm not exactly sure what this check is for. Do you mind leaving a comment in the code for what this check is for?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+
+// ----------------------------------------------------------------------------
+// Tile.QuickAction component
+
+const TileQuickAction = React.forwardRef((props, ref) => {
+  const { as: Element = 'button', className, children, ...rest } = props;
+  return (
+    //todo: styling is all off, need to fix after
+    <Element
+      className={cx('iui-button', 'iui-tile-thumbnail-quick-action', className)}
```
adamhock, line 260: This could be fixed when styling issues are fixed, but I'll comment anyway

If `Element` isn't a `button`, do we still want to add `iui-button` styling? The same goes for `TypeIndicator`

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+const TileBadge = polymorphic.div('iui-tile-thumbnail-badge-container');
+TileBadge.displayName = 'Tile.Badge';
+
+type TileBadgeOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+// ----------------------------------------------------------------------------
+// Tile.Name component
+
+const TileName = React.forwardRef(
+  ({ as: Element = 'div', children, ...rest }, ref) => {
```
adamhock, line 310: I'd prefer if `props` were extracted within the function like all of the other components. Just replace this list with `props` and then extract these in the next line

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+
+type TileMetadataOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+// ----------------------------------------------------------------------------
+// Tile.MoreOptions component
+
+const TileMoreOptions = React.forwardRef((props, ref) => {
+  const { as: Element = 'div', className, children, ...rest } = props;
+  const [isMenuVisible, setIsMenuVisible] = React.useState(false);
+  const showMenu = React.useCallback(() => setIsMenuVisible(true), []);
```
adamhock, line 380: nitpicky but can these calls just immediately go inside of `onShow` and `onHide`, leaving no need to make these separate variables?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,646 @@
...
+  const hideMenu = React.useCallback(() => setIsMenuVisible(false), []);
+
+  return (
+    children && (
+      <DropdownMenu
+        onShow={showMenu}
+        onHide={hideMenu}
+        menuItems={(close) =>
+          children.map((option) =>
```
adamhock, line 389: Would it make sense to have `menuItems` be a prop here that's passed into `TileMoreOptions`? That was you can verify that it's of type `JSX.Element[]` and can use `map` properly?

adamhock (commented):
This is great work!

Don't forget to leave a note somewhere to update unit tests and add a changeset.

packages/itwinui-react/src/core/index.ts
LostABike: I think I can't because exporting type requires `export type` ?
image

Unless you mean for TileNew stuff to go inside the same braces as original `Tile` right above, then I plan to rename `TileNew` to `Tile` soon and remove one of the set :d

packages/itwinui-react/src/core/Tile/index.ts
LostABike: Nice! Thanks for the tip. I updated this.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Right yea. I forgot to clean up comment. Updated

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Cleaned up!

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Thanks for catching. I updated both spots.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Makes sense. I moved Tile up.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Yea this makes sense. I moved props to above components. Also updated comment.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Thanks. I updated this.

mayank99:
> moving name out of content for folder variant was getting more complicated than I would like to so I keep the original logic and have `name` under `content` as before, and made specific comment about that for `<Tile.Name>` so user knows where to put it in which case. I could try to do grid if we are keen on moving it out of content, though 😅

@LostABike Can you elaborate on what difficulty/complexity you're facing? It kinda makes sense to use different grids for folder/non-folder tiles.

Keeping `name` and `content` as before would make the whole api inconsistent because some parts are passed through `children` and others through props.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Actually want to pull @mayank99 in on this one to ask because I copied the logic from original Tile, which is that the `thumbnail` or `children` in this case is a custom React Element for thumbnail that the user passed in, and we cloned it to render 
image

So I think the check is for whether what is passed in `thumbnail` is a custom React Element. As opposed to being just a string to a thumbnail image.

But also, I circumvented `React.cloneElement` by using `React.createElement` in order to keep the same logic but without using `cloneElement`. Should we do it like that? Or if not, what else can we do here?
image

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Yea that makes sense. I updated this.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Yup I moved these functions to inline :d

packages/itwinui-react/src/core/Tile/TileNew.tsx
mayank99: Please avoid using deprecated apis such as `cloneElement` or `createElement`. Writing such code increases complexity and is prone to bugs, while providing little benefit.

Ultimately, what we want to achieve is pretty simple: for both `Tile.ThumbnailPicture` and `Tile.ThumbnailAvatar`, we only need to support _two_ cases. The developer should be able to pass either:
- a string containing the url to the image, or
- a component that renders an `` or `<svg>` or `<div>` with background-image

You can use a [discriminated union](https://www.totaltypescript.com/discriminated-unions-are-a-devs-best-friend) type to narrow the prop types (similar to #1120). Something like:
```
{
  url: string
} |
{
  children: string
}
```

After that, you can render a regular div with your css class (whether it's `iui-tile-thumbnail-picture` or `iui-tile-thumbnail-avatar` or whatever).
```jsx
<div
  className={cx('iui-whatever', className)}
  style={{ backgroundImage: `url()`, ...style }}
  {...rest}
/>
``` 
```jsx
<div
  className={cx('iui-whatever', className)}
  {...rest}
>
  {children}
</div>
```

After that, you can handle all cases (svg vs img vs div) inside the css code.

LostABike: Oop I did not realize `createElement` is deprecated 😬 but thanks for the direction 😅

packages/itwinui-react/src/core/index.ts
adamhock: image
I was referring to all of these types to be in the same braces since they're all types and exported from the same file

packages/itwinui-react/src/core/Tile/TileNew.tsx
adamhock: Make sure to call `useGlobals` in the `TileComponent`

packages/itwinui-react/src/core/index.ts
LostABike: Ah yea that's right. Updated

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Oop right. Updated :D

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: I updated this. Working as it should at least on Storybook 🤔 also I removed the use of ThumbnailAvatar altogether and only use ThumbnailPicture to handle both cases. The user will have to use a `url` prop to pass in the link for their image, or they could pass in their own Avatar/SVG. Did not clean up Avatar code yet, trying to see if this is what we want.

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: Hm yea I was having errors with these. I have instead require the type of `children` to be `JSX.Element[]`. Not sure if this is preferable or not?

LostABike:
> > moving name out of content for folder variant was getting more complicated than I would like to so I keep the original logic and have `name` under `content` as before, and made specific comment about that for `<Tile.Name>` so user knows where to put it in which case. I could try to do grid if we are keen on moving it out of content, though 😅
> 
> @LostABike Can you elaborate on what difficulty/complexity you're facing? It kinda makes sense to use different grids for folder/non-folder tiles. I'd expect you'd be able to use something like this for folder tiles:
> 
> ```css
> grid-template-areas:
>   'thumbnail name'
>   'thumbnail content';
> ```
> 
> Keeping `name` and `content` as before would make the whole api inconsistent because some parts are passed through `children` and others through props. And there is also an accessibility benefit to moving name before thumbnail (see #1125).

I was just trying to achieve this without grid, which is possible just too extra. But yea grid makes sense. I updated the css. Folder looks correct now and Name is moved to outside of Content :D  
image

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import React from 'react';
+import cx from 'classnames';
+import {
+  useSafeContext,
+  SvgMore,
+  type PolymorphicComponentProps,
+  type PolymorphicForwardRefComponent,
```
mayank99, line 12: can you move this into a separate `import type` statement?

see https://github.com/iTwin/iTwinUI-react/pull/996

mayank99:
> yea grid makes sense. I updated the css. Folder looks correct now and Name is moved to outside of Content :D
image

Make sure to update html files to reflect the new markup.

Also I think it might need some spacing adjustments. Maybe @FlyersPh9 can help with the exact values. First thing that comes to mind is this:
```css
grid-template:
  'thumbnail name' auto
  'thumbnail content' 1fr / 1fr auto;
row-gap:  calc(var(--iui-size-s) * 0.5);
```

If there's margin set on the tile-name, it might need to be removed.

mayank99:
> Quick Action and Type Indicator icons look different than in original Tile?
image

This is something fishy going on with the markup. There is a button inside another button.


Can you fix that?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+type TileMoreOptionsOwnProps = {
+  children: React.ReactNode[];
+};
+
+const TileMoreOptions = React.forwardRef((props, ref) => {
+  const { as: Element = 'div', className, children, ...rest } = props;
+  const [isMenuVisible, setIsMenuVisible] = React.useState(false);
+
+  return (
+    <DropdownMenu
```
gretanausedaite, line 400: Real use case we are missing: We need to allow users define their own DropdownMenu for moreOptions (At least expose onClick which expands menu).

mayank99: do you mean the button that opens the menu?

gretanausedaite: Yes, one team wants to load custom options on moreOptions button click so they need onClick event exposed.

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+// Tile.QuickAction component
+
+type TileQuickActionOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+const TileQuickAction = React.forwardRef((props, ref) => {
+  const { as: Element = 'button', className, children, ...rest } = props;
+  return (
+    //todo: styling is all off, need to fix after
+    <Element
+      className={cx('iui-button', 'iui-tile-thumbnail-quick-action', className)}
```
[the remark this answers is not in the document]
LostABike: Do you mean that we don't use `<Element/>` at all but instead use `<IconButton/>`?

packages/itwinui-react/src/core/index.ts
adamhock: Closing since types do not have to be exported anymore

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+
+  return (
+    <Element className='iui-tile-name' ref={ref} {...rest}>
+      <TitleIcon
+        isLoading={isLoading}
+        isSelected={isSelected}
+        isNew={isNew}
+        status={status}
+      />
+      <span className='iui-tile-name-label'>
```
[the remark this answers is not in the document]
LostABike: > every subcomponent should try to render one element

Yes I see that. But it looks like icon is controlled by other props that the user passed into the *main* Tile component, like `isLoading`, `isNew` etc, and is not currently something we offer customizability on but instead provide our own icons for whichever cases, which is why it just renders along with name.

Are you proposing that we create a separate subcomponent for icon and then also offer customizability when none of the icon related prop is passed into main Tile? i.e the props (isNew, isLoading,...) take precedent to render even if the user passed in a customized SVG for icon.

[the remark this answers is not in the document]
mayank99: I'm suggesting kinda the opposite. By default the icons will be the ones that make sense for `isLoading`, `isNew`, etc. But if the user passes a custom svg, then it should override the default.

packages/itwinui-react/src/core/index.ts
```diff
@@ -295,6 +295,24 @@ export type { TextareaProps } from './Textarea/index.js';
 export { Tile } from './Tile/index.js';
```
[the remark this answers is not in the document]
LostABike: Yup, updated!

packages/itwinui-react/src/core/Tile/index.ts
```diff
@@ -5,3 +5,4 @@
 export { Tile } from './Tile.js';
 export type { TileProps } from './Tile.js';
 export default './Tile';
+export * from './TileNew.js';
```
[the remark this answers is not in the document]
LostABike: Thanks. Updated.

apps/storybook/src/Tile.stories.tsx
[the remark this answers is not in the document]
LostABike: Wait do I still need this? I think I removed it somehow

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+    isSelected,
+    isLoading,
+    isActionable,
+    isDisabled,
+    setActionable,
+    onClick,
+    ...rest
+  } = props;
+
+  useGlobals();
```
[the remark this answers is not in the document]
LostABike: Yup. Updated to use `Box` and removed `useGlobals`

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+            url && typeof url === 'string' ? `url(${url})` : undefined,
+        }}
+        ref={ref}
+        {...rest}
+      />
+    );
+  }
+
+  if (React.isValidElement(children) && children.type == 'SVG') {
+    console.log('true');
```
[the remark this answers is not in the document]
LostABike: Odd, I thought I corrected this before push but yea, removed.

packages/itwinui-react/src/core/Tile/TileNew.tsx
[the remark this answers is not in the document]
LostABike: I return `Box` for this now, but did you mean to instead return `<IconButton/>` instead?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+
+type TileTypeIndicatorOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+const TileTypeIndicator = React.forwardRef((props, ref) => {
+  const { as: Element = 'button', className, children, ...rest } = props;
+  return (
+    //todo: styling is all off, need to fix after
+    <Element
+      className={cx(
+        'iui-button',
```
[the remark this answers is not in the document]
LostABike: I return `Box` for this now, but did you mean to instead return `<IconButton/>` instead?

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+    return (
+      <ProgressRadial
+        className='iui-tile-status-icon'
+        aria-hidden
+        indeterminate
+      />
+    );
+  }
+  if (isSelected) {
+    return <SvgCheckmark className='iui-tile-status-icon' aria-hidden />;
```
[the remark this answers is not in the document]
LostABike: Updated

packages/itwinui-react/src/core/Tile/TileNew.tsx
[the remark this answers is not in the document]
LostABike: Thanks for clarification. I have separated `Name` and `Icon`, should work as you mentioned now.

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+}) as PolymorphicForwardRefComponent<'div', TileThumbnailPictureOwnProps>;
+
+TileThumbnailPicture.displayName = 'TileNew.TileThumbnailPicture';
+
+// ----------------------------------------------------------------------------
+// Tile.ThumbnailAvatar component
+
+type TileThumbnailAvatarOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+const TileThumbnailAvatar = React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
LostABike: Yup removed.

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+    </Element>
+  );
+}) as PolymorphicForwardRefComponent<'button', TileTypeIndicatorOwnProps>;
+TileTypeIndicator.displayName = 'Tile.TypeIndicator';
+
+// ----------------------------------------------------------------------------
+// Tile.Badge component
+type TileBadgeOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
+
+const TileBadge = polymorphic.div('iui-tile-thumbnail-badge-container');
```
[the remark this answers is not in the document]
LostABike: Makes sense. I updated the name.

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+   * Dropdown menu containing `MenuItem`s.
+   */
+  MoreOptions: TileMoreOptions,
+  /**
+   * Upto two buttons shown at the bottom of the tile.
+   */
+  Buttons: TileButtons,
+});
+
+export type TileNewProps = PolymorphicComponentProps<'div', TileOwnProps>;
```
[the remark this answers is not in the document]
LostABike: Removed/

packages/itwinui-react/src/core/Tile/TileNew.tsx
```diff
@@ -0,0 +1,657 @@
...
+  if (isNew) {
+    return <SvgNew className='iui-tile-status-icon' aria-hidden />;
+  }
+  if (StatusIcon) {
+    return <StatusIcon className='iui-tile-status-icon' />;
+  }
+  return null;
+};
+
+export const TileNew = Object.assign(TileComponent, {
```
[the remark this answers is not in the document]
LostABike: Oops, updated

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: I see. Updated

LostABike:
> > Quick Action and Type Indicator icons look different than in original Tile?
> > image
> 
> This is something fishy going on with the markup. There is a button inside another button. 
> 
> Can you fix that?

The styling issue and button you mentioned have been fixed when I used a `div` (`Box`) instead, so I think this is good now? I also left a comment in respond to your feedback on those buttons.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -2,111 +2,73 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import * as React from 'react';
+
+import React from 'react';
```
mayank99, line 6: can you try to minimize the diff?


other than order, the big thing is `import * as React` (which should be preferred over `import React`).

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -148,209 +106,289 @@ type TileProps = {
...
-          )}
-        </Box>
-      </Box>
-    );
+//<Tile>
+//  <Tile.Thumbnail>
+//  <Tile.Name>
+//  <Tile.Content>  //has meta data
+//  <Tile.Buttons>
+//</Tile>
```
mayank99, line 118: remove this comment

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -367,25 +405,174 @@ const TitleIcon = ({
...
+
   if (isLoading) {
-    return (
-      <ProgressRadial
-        className='iui-tile-status-icon'
-        aria-hidden
-        indeterminate
-      />
+    icon = (
+      <ProgressRadial style={{ height: '100%' }} aria-hidden indeterminate />
```
mayank99, line 412: why inline style `height: 100%`?

apps/storybook/src/Tile.stories.tsx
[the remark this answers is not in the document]
mayank99: yes you still need it because `TileProps` can no longer be imported from `@itwin/itwinui-react`

packages/itwinui-react/src/core/Tile/TileNew.tsx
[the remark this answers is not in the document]
mayank99: that works, but now I'm thinking this needs another subcomponent `Tile.IconButton` (name debatable) for the custom styling that's applied on top of `IconButton`.

would be used like this:
```
<Tile.QuickAction>
  <Tile.IconButton>⭐️</Tile.IconButton>
</Tile.QuickAction>
```

packages/itwinui-react/src/core/Tile/TileNew.tsx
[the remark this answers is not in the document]
mayank99: It's missing the convenience option 


Also the code from the private `TitleIcon` can be moved inside `TileNameIcon`

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: I added an `onClick` prop, is that what you mean?

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Noted, I reorganized the imports

LostABike:
> > yea grid makes sense. I updated the css. Folder looks correct now and Name is moved to outside of Content :D
> > image
> 
> Make sure to update html files to reflect the new markup.
> 
> Also I think it might need some spacing adjustments. Maybe @FlyersPh9 can help with the exact values. First thing that comes to mind is this:
> 
> ```css
> grid-template:
>   'thumbnail name' auto
>   'thumbnail content' 1fr / 1fr auto;
> row-gap:  calc(var(--iui-size-s) * 0.5);
> ```
> 
> If there's margin set on the tile-name, it might need to be removed.

When you say to update the html file to reflect the new markup, you mean https://itwin.github.io/iTwinUI/css/tile.html right? I think barely any css was changed so everything is still looking like it should. But let me know if you notice any discrepancies.

Also about the folder name spacing, I added a margin-top instead to get it to have the same spacing as before, seems ok?

image

packages/itwinui-react/src/core/Tile/TileNew.tsx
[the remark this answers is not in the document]
LostABike: I added another subcomponent `ButtonIcon` 👀 how's this?

apps/storybook/src/Tile.stories.tsx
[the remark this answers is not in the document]
LostABike: Okay I added this back

packages/itwinui-react/src/core/Tile/TileNew.tsx
LostABike: This has been updated, it is now whatever <Box/> is and no more `iui-button` classname, there is a new subcomponent for icon which is polymographically default to be `button` with`iui-button-icon`, seems to be working as intended.

gretanausedaite (commented):
Code has some build errors

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -148,244 +106,479 @@ type TileProps = {
...
-type TitleIconProps = {
-  isLoading?: boolean;
-  isSelected?: boolean;
-  isNew?: boolean;
-  status?: 'positive' | 'warning' | 'negative';
+// ----------------------------------------------------------------------------
+// Tile.Action component
+
+export const TileAction = (
+  props: PolymorphicComponentProps<'a', TileActionOwnProps>,
```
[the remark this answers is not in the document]
LostABike: Yup I have updated this

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -148,244 +106,479 @@ type TileProps = {
...
+        )}
+        ref={ref}
+        {...rest}
+      >
+        <IconButton
+          as='button'
+          styleType='borderless'
+          size='small'
+          aria-label='More options'
+          onClick={onClick}
```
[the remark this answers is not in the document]
LostABike: Updated name

LostABike:
Updated PR description. I think I went over all and responded/resolved all of them on my end, so now waiting for reviewers' response next. Build failures fixed. Unit tests updated. HTML markup updated. CSS updated and looks good on my end but might be unconventional/not optimal (folder grid, margin etc). Documentation code added, but documentation itself might need to be in a separate PR.

.changeset/polite-walls-rush.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': major
+'@itwin/itwinui-css': major
+---
+
+Updated Tile to be composition of customizable subcomponents
```
gretanausedaite, line 6: Can you separate changeset files for css and react changes? This way you could be more descriptive with what is changed, eg. in css the main change is folder tile changing to display: grid.

apps/website/src/pages/docs/tile.mdx
```diff
@@ -14,14 +14,26 @@ import * as AllExamples from '~/examples';
 
 <p>{frontmatter.description}</p>
 
-<Placeholder componentPageName='core-tile' />
+{/* <Placeholder componentPageName='core-tile' /> */}
```
gretanausedaite, line 17: Remove if not needed anymore or add back if you're not ready to update docs page just yet.

gretanausedaite (commented):
Generally looks good

.changeset/polite-walls-rush.md
LostABike: I created 2 separate changesets

apps/website/src/pages/docs/tile.mdx
LostABike: Oops, actually not meant to push this file, I have reverted the changes.

.changeset/polite-walls-rush.md
gretanausedaite: changeset description should be useful

LostABike:
Should we add `aria-label` for ALL returning `<Box/>` of each subcomponent? I only see it for certain specific components in our code base like `<IconButton/>`

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: What should we do about this now?

mayank99:
> Should we add `aria-label` for ALL returning `<Box/>` of each subcomponent? I only see it for certain specific components in our code base like `<IconButton/>`

No, `aria-label` is generally only needed on icon buttons that are hardcoded by us (e.g. the more options button which otherwise shows up as three dots).

packages/itwinui-react/src/core/Tile/Tile.tsx
mayank99: For now, you'll need to adjust the CSS to fix it. I think changing `svg` to `svg:not(.iui-radial)` might do it?

I'm working on changes to `<ProgressRadial>` which will avoid the need for such hacks.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -41,97 +40,109 @@ export default {
...
-      leftIcon={leftIcon}
-      rightIcon={rightIcon}
-      moreOptions={moreOptions}
-      {...rest}
-    />
+    <Tile {...args}>
+      <Tile.Name>
+        <Tile.NameIcon />
+        <Tile.NameLabel>Stadium</Tile.NameLabel>
+      </Tile.Name>
```
mayank99, line 49: when the label is a simple string and when there is no custom icon, we can just show `<Tile.Name name='Stadium'>`

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+  return (
+    <Box
+      className={cx('iui-tile-thumbnail-quick-action', className)}
+      ref={ref}
+      {...rest}
+    >
+      {children}
+    </Box>
+  );
+}) as PolymorphicForwardRefComponent<'div'>;
```
mayank99, line 233:
```suggestion
const TileQuickAction = polymorphic.div('iui-tile-thumbnail-quick-action');
```

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+// Tile.Name component
+type TileNameOwnProps = {
+  name?: string;
 };
 
+const TileName = React.forwardRef((props, ref) => {
+  const { className, children, name, ...rest } = props;
+  return (
+    <Box className={cx('iui-tile-name', className)} ref={ref} {...rest}>
+      {name ?? children}
```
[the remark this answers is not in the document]
mayank99: yes

apps/storybook/src/Tile.stories.tsx
LostABike: I think for Storybook it is better to keep it this way so that users can toggle the icons, since if only `name` prop is passed in, there will be no icons, and if there is NameIcon subcomponent, it overrides the `name` (I talked about this in another comment below)

mayank99: can we change the website examples and at least the basic story?

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+    <TileContext.Provider
+      value={{
+        status,
+        variant,
+        isNew,
+        isSelected,
+        isLoading,
+        isActionable,
+        isDisabled,
+        setActionable: () => {},
```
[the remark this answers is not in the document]
LostABike: I was getting this error
image

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
 
-type TitleIconProps = {
-  isLoading?: boolean;
-  isSelected?: boolean;
-  isNew?: boolean;
-  status?: 'positive' | 'warning' | 'negative';
+// ----------------------------------------------------------------------------
+// Tile.Action component
+
+export const TileAction = (props: PolymorphicForwardRefComponent<'a'>) => {
```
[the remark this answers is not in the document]
LostABike: o right, I updated this

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -12,100 +13,63 @@ import {
...
   supportsHas,
+  polymorphic,
   Box,
 } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { DropdownMenu } from '../DropdownMenu/index.js';
 import { IconButton } from '../Buttons/index.js';
 import { ProgressRadial } from '../ProgressIndicators/index.js';
 
+type TileProps = React.ComponentProps<typeof Tile>;
```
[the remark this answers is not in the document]
LostABike: done

apps/storybook/src/Tile.stories.tsx
```diff
@@ -41,97 +40,109 @@ export default {
...
+      <Tile.ThumbnailArea>
+        <Tile.BadgeContainer>
+          <Badge backgroundColor='hsl(197, 71%, 83%)'>Badge</Badge>
```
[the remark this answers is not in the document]
LostABike: oh cool, updated to 'skyblue' now

apps/storybook/src/Tile.stories.tsx
LostABike: yea makes sense. I updated this.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -22,12 +23,10 @@ import SvgInfo from '@itwin/itwinui-icons-react/cjs/icons/Info';
 import SvgStar from '@itwin/itwinui-icons-react/cjs/icons/Star';
 import SvgTag from '@itwin/itwinui-icons-react/cjs/icons/Tag';
 
-type TileProps = React.ComponentProps<typeof Tile>;
```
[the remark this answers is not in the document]
LostABike: Done

apps/storybook/src/Tile.stories.tsx
```diff
@@ -22,12 +23,10 @@ import SvgInfo from '@itwin/itwinui-icons-react/cjs/icons/Info';
...
-
 export default {
   component: Tile,
   argTypes: {
-    className: { control: { disable: true } },
+    classNameIcon: { control: { disable: true } },
```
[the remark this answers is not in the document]
LostABike: Oops yea, fixed.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+  const { status, isLoading, isSelected, isNew } = useSafeContext(TileContext);
+
+  type TitleIconProps = {
+    isLoading?: boolean;
+    isSelected?: boolean;
+    isNew?: boolean;
+    status?: 'positive' | 'warning' | 'negative';
+  };
+
+  const TitleIcon = ({
```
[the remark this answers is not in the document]
LostABike: I'm getting this build error
```
src/core/Tile/Tile.tsx(281,15): error TS2604: JSX element type 'StatusIcon' does not have any construct or call signatures.
```

So doing this instead:
```
const TileNameIcon = React.forwardRef((props, ref) => {
  const { children, ...rest } = props;
  const { status, isLoading, isSelected, isNew } = useSafeContext(TileContext);

  const StatusIcon = !!status && StatusIconMap[status];
  let icon;

  if (isLoading) {
    icon = <ProgressRadial aria-hidden indeterminate />;
  }
  if (isSelected) {
    icon = <SvgCheckmark aria-hidden />;
  }
  if (isNew) {
    icon = <SvgNew aria-hidden />;
  }
  if (StatusIcon) {
    icon = <StatusIcon aria-hidden />;
  }

  return children || icon ? (
    <Box className='iui-tile-status-icon' ref={ref} {...rest}>
      {children ?? icon}
    </Box>
  ) : null;
}) as PolymorphicForwardRefComponent<'div'>;
TileNameIcon.displayName = 'Tile.NameIcon';
```
also returning empty if nothing is passed in so that there is no empty space/box in front of name when the subcomponent is present (especially on Storybook)

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+// ----------------------------------------------------------------------------
+// Tile.Metadata component
+
+const TileMetadata = polymorphic.div('iui-tile-metadata');
+TileMetadata.displayName = 'Tile.Metadata';
+
+// ----------------------------------------------------------------------------
+// Tile.MoreOptions component
+type TileMoreOptionsOwnProps = {
+  menuOnClick?: React.MouseEventHandler<HTMLElement>;
```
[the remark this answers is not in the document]
LostABike: updated to `onButtonClick`

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
mayank99: StatusIcon should come before the other conditions

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,244 +107,499 @@ type TileProps = {
...
+
+// ----------------------------------------------------------------------------
+// Tile.ButtonIcon component
+
+const TileButtonIcon = React.forwardRef((props, ref) => {
+  const { className, children, ...rest } = props;
+  return (
+    <Box
+      as='span'
+      className={cx('iui-button-icon', className)}
```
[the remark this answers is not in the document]
mayank99: it's different
```js
<Tile.QuickAction>
  <Tile.IconButton>...</Tile.IconButton>
</Tile.QuickAction>
```
vs
```js
<Tile.QuickAction>
  <Tile.ButtonIcon>
    <IconButton>...</IconButton>
  </Tile.ButtonIcon>
</Tile.QuickAction>
```

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
mayank99: that's because you removed the code that was passed here before.

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: so are we returning the component `IconButton` inside the Tile Subcomponent `<Tile.IconButton>` then?

apps/storybook/src/Tile.stories.tsx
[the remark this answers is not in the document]
LostABike: Done, and also simplified most of them to use `name` prop as well since can't toggle icon on website

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: ohhh. I moved these logic back.

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: Done

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: Right, I moved it up.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -9,11 +9,11 @@ import {
   Badge,
   Button,
   getUserColor,
-  IconButton,
   MenuItem,
   Tag,
   TagContainer,
   Tile,
+  TileProps,
```
gretanausedaite, line 16: Not exported anymore

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+  const {
+    className,
+    status,
+    variant,
+    isNew,
+    isSelected,
+    isLoading,
+    isActionable: isActionableProp,
+    isDisabled,
+    onClick,
```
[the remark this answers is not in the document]
mayank99: cool, so lets remove `onClick` from context, and remove any special handling for `onClick`, then update any leftover actionable examples to explicitly use `Tile.Action` instead of `isActionable` prop.

should we also remove `isActionable` prop?

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -2,6 +2,7 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
```
adamhock, line 5: nitpicky, but this newline can be removed. The fewer line changes the better

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -21,91 +23,51 @@ import { ProgressRadial } from '../ProgressIndicators/index.js';
 
 const TileContext = React.createContext<
   | {
+      status: 'positive' | 'warning' | 'negative' | undefined;
```
adamhock, line 26: nitpicky, but you can change this to `status?` and removed the `undefined` option

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+      />
+    </TileContext.Provider>
+  );
+}) as PolymorphicForwardRefComponent<'div', TileOwnProps>;
+TileComponent.displayName = 'Tile';
+
+// ----------------------------------------------------------------------------
+// Tile.Action component
+
+const TileAction = React.forwardRef((props, forwardedRef) => {
```
adamhock, line 168: nitpicky, but `forwardedRef` can change to `ref` for consistency

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+  }, [setActionable]);
+
+  return <LinkAction ref={forwardedRef} {...props} />;
+}) as PolymorphicForwardRefComponent<'a'>;
+TileAction.displayName = 'Tile.Action';
+
+// ----------------------------------------------------------------------------
+// Tile.ThumbnailArea component
+
+const TileThumbnailArea = polymorphic.div('iui-tile-thumbnail');
```
adamhock, line 183: You can replace `polymorphic.div(` with `polymorphic(` since `div` is the default. This change can be made across the entire file

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
-          </Box>
-
-          {buttons && <Box className='iui-tile-buttons'>{buttons}</Box>}
-        </Box>
-      </TileContext.Provider>
+      <Box
+        className={cx('iui-tile-thumbnail-picture', className)}
+        style={{
+          backgroundImage:
+            url && typeof url === 'string' ? `url(${url})` : undefined,
```
adamhock, line 204: `url &&` can be removed since there already is a check for if `url` exists

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+          aria-disabled={isDisabled}
+        >
+          {children}
+        </LinkAction>
+      ) : (
+        children
+      )}
+    </Box>
+  );
+}) as PolymorphicForwardRefComponent<'div'>;
```
adamhock, line 334: This component should be forwarded as a `span` instead of a `div`

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+          {
+            'iui-visible': isMenuVisible,
+          },
+          className,
+        )}
+        ref={ref}
+        {...rest}
+      >
+        <IconButton
+          as='button'
```
adamhock, line 393: This `as` statement can be removed since the default type for `IconButton` is a `button`

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
+  IconButton: TileIconButton,
+  /**
+   * `Name` subcomponent under thumbnail or top of the Tile if no thumbnail present.
+   */
+  Name: TileName,
+  /**
+   * `NameIcon` next to name of the tile. Goes under <Tile.Name>
+   * @example
+   * <Tile>
+   * <Tile.Name>
```
adamhock, line 497: nitpicky, but the tabbing is off slightly in this example and in the `NameLabel` example

adamhock (commented):
I have only reviewed the `Tile.tsx` and `Tile.test.tsx` files. I will review the rest of the files and possibly leave more comments shortly

Overall, this is impressive! Great job

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: @mayank99 I believe this is what you mean? I moved the action logic out of `Tile.NameLabel` into `Tile.Action`, removing the `onClick` context, so now custom `onClick` if any is passed into `Tile.Action` instead (also pushed):

image

packages/itwinui-react/src/core/Tile/Tile.tsx
mayank99: forwardedRef is the new preferred name

https://github.com/iTwin/iTwinUI/blob/ab07a844f5c2bca1c255adc5502e572c34fc8ede/scripts/createComponent.js#L71

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
 
-type TitleIconProps = {
-  isLoading?: boolean;
-  isSelected?: boolean;
-  isNew?: boolean;
-  status?: 'positive' | 'warning' | 'negative';
+// ----------------------------------------------------------------------------
+// Tile.Name component
+type TileNameOwnProps = {
+  name?: string;
```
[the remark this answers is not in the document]
mayank99: this is done for convenience so the user doesn't always have to manually pass `TileNameIcon`

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
mayank99: yes, basically that. You're missing rest props though.

Also, we can set default `as` prop based on `href` vs `onClick`, similar to how it's done in `Alert.Action`

https://github.com/iTwin/iTwinUI/blob/ab07a844f5c2bca1c255adc5502e572c34fc8ede/packages/itwinui-react/src/core/Alert/Alert.tsx#L94-L102

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Done :D

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Updated!

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: okay, so I updated to `forwardedRef` for all of them

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: oooh thanks! I updated all of the similar instances.

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: I see. Removed.

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
LostABike: hmm actually wouldn't having the string as children achieve the same result as having the prop though?

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,239 +105,463 @@ type TileProps = {
...
-    return <SvgNew className='iui-tile-status-icon' aria-hidden />;
+    icon = <SvgNew aria-hidden />;
   }
-  if (StatusIcon) {
-    return <StatusIcon className='iui-tile-status-icon' />;
-  }
-  return null;
+
+  return children || icon ? (
+    <Box className='iui-tile-status-icon' ref={ref} {...rest}>
```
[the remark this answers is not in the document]
LostABike: thanks for catching. I updated this!

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Done

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Thanks! removed

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: Updated :D

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
mayank99: `children` is currently rendered as-is. That means `<Tile.Name>some string</Tile.Name>` doesn't render an icon. I think we want to preserve that behavior, otherwise it would be confusing why `<Tile.Name><>string in a fragment</></Tile.Name>` suddenly starts adding an icon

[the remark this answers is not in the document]
mayank99: on that note, you can make this a react node
```suggestion
  name?: React.ReactNode;
```

packages/itwinui-react/src/core/Tile/Tile.tsx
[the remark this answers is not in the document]
gretanausedaite: > should we also remove `isActionable` prop?

I don't remember if there is a use-case for actionable tile without click action. Maybe users could want that hover effect when clicking on more options menu? We could mark that as deprecated and keep only for classname toggle.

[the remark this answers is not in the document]
LostABike: marked @deprecated

FlyersPh9:
On the HTML demo page, the folder that is loading appears to have some issues:
1. The more options button is misaligned to the right.
2. The title & spinner appears to be lower than the rest. 
Screenshot 2023-06-13 at 4 02 11 PM

LostABike:
> On the HTML demo page, the folder that is loading appears to have some issues:
> 
> 1. The more options button is misaligned to the right.
> 2. The title & spinner appears to be lower than the rest.
>    Screenshot 2023-06-13 at 4 02 11 PM

For the options button, it seems to be because I didn't have `iui-tile-description` under `iui-tile-content` like the other demos, so I added that, although it might be more optimal that it should look right regardless whether that class is added, which-

brings us to the main issue, is that I think the way I made my grid portion is....not optimal and very hacky/hard-coded? Currently this setup would make current demos look fine but I'm not sure for universal usages...Maybe you can help?
image

LostABike:
> The video tiles with the play icon overlaid above a thumbnail are using the incorrect color.
> 
> **Before:** Screenshot 2023-06-13 at 4 05 11 PM
> 
> **After:** Screenshot 2023-06-13 at 4 05 35 PM

Should be fixed now. Removed an unnecessary css hack.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,240 +104,459 @@ type TileProps = {
...
+ *      <Tile.Description />
+ *      <Tile.Metadata/>
+ *      <Tile.MoreOptions/>
+ *    </Tile.ContentArea>
+ *    <Tile.Buttons/>
+ * </Tile>
+ */
+export const Tile = Object.assign(TileComponent, {
+  /**
+   *ThumbnailArea subcomponent that contains `ThumbnailPicture`, `QuickAction`, `TypeIndicator` or `BadgeContainer`
```
adamhock, line 432: I'd add a space before `ThumbnailArea` on this line

adamhock (approved):
One nitpicky comment. Other than that, make sure the tests pass before you merge!

Fantastic job! 🎉

packages/itwinui-react/src/core/Tile/Tile.tsx
LostABike: done 😄 , and thanks!

LostABike:
@FlyersPh9 do you want to make a last check on the css/html of this? before I merge (if I manage to pass those image tests that is 😂 )

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -150,9 +154,12 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
...
 @mixin iui-folder {
-  display: inline-flex;
-  flex-direction: row;
-  flex-wrap: nowrap;
+  display: grid;
+  grid-template-areas:
+    'thumbnail name'
+    'thumbnail content';
+  grid-template-columns: 3fr 7fr;
+  grid-template-rows: 1.5fr 2.5fr;
```
mayank99, line 162: Since @FlyersPh9 is out, here's my next best suggestion for the grid:
```suggestion
  grid-template-columns: 1fr 2fr;
  grid-template-rows: auto 1fr;
```

the columns preserve the 1:2 ratio from before, while the rows will prefer to keep the "name" part fixed and let the "content" part expand.

you will likely need to add some gap between the name and content. I think there was a `margin-bottom` before which you removed. Reintroducing that might fix the visuals.

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -147,240 +104,459 @@ type TileProps = {
    * @default false
    */
   isDisabled?: boolean;
-  onClick?: React.MouseEventHandler<HTMLElement>;
+  setActionable?: React.Dispatch<React.SetStateAction<boolean>>;
```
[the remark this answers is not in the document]
LostABike: Thanks. I removed this

packages/itwinui-css/src/tile/tile.scss
LostABike: This is working now, as in no more visual issues. Thanks for the suggestion

mayank99 (approved):
LGTM! Nice work

PR #1257, opened 2023-05-08
revert v1 styles for psuedo elements too

mayank99:
## Changes

After #1245, a user reported some unexpected visible borders around checkboxes.

The border comes from v1 styles which were not being correctly targeted in the place where we globally revert all v1 styles, so I explicitly added selectors for `::before`/`::after` in that rule.

## Testing

Tested with this code:
```jsx
import { Global } from '@emotion/react';
import { Checkbox, ThemeProvider } from '@itwin/itwinui-react';

export default () => {
  return (
    <div className='iui-body'>
      <Global
        styles={`
          @layer test {
            .iui-checkbox::before {
              border: 1px solid hotpink;
            }
          }
        `}
      />
      <ThemeProvider theme='dark'>
        <Checkbox label='Hello' />
      </ThemeProvider>
    </div>
  );
};
```

Without this change:
pink outline

With this change:
no visible outline

## Docs

N/A

### DIFF
--- a/.changeset/wet-eyes-drop.md
+++ b/.changeset/wet-eyes-drop.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue where pseudo elements inside a v2 boundary were not reverting v1 styles.
--- a/packages/itwinui-css/src/global.scss
+++ b/packages/itwinui-css/src/global.scss
@@ -5,7 +5,11 @@
 // reset anything that may have been set by v1
 :where(.iui-body, [class*='iui-theme-']) :where(.iui-root, [data-iui-theme]) {
   :where([class*='iui-']:not(.iui-root)) {
-    all: revert;
+    &,
+    &::before,
+    &::after {
+      all: revert;
+    }
   }
 }
 


FlyersPh9 (approved):

HaveSpacesuit:
Are we expected to add the `.iui-body` class to our project, or should the iTwin-react package be doing that for us? I've noticed that my app doesn't have `iui-body` on it, so this fix doesn't work for me unless I add that in.

mayank99:
> Are we expected to add the `.iui-body` class to our project, or should the iTwin-react package be doing that for us? I've noticed that my app doesn't have `iui-body` on it, so this fix doesn't work for me unless I add that in.

Can you give more details about your setup? Where are the v1 styles coming from? And are you able to reproduce the issue in a codesandbox/stackblitz environment? Try forking https://stackblitz.com/edit/vitejs-vite-jxx5iu?file=src%2FApp.jsx

In short, `iui-body` should be automatically set by any components from `@itwin/itwinui-react@1`. See relevant code here (L22): https://unpkg.com/browse/@itwin/itwinui-react@1.48.2/esm/core/utils/hooks/useTheme.js

HaveSpacesuit:
Ah, that makes sense! We are importing some functionality that comes along with an iTwin 1 component. We only display that component on one part of the app. Once I visit that part, the style overrides start working.

I think the best fix for us for now is to just add the `iui-body` manually, with a comment that when that component is upgraded we can remove the class.

PR #1258, opened 2023-05-08
update cypress to v12 and use prebuilt docker image

mayank99:
## Changes

- Updated from cypress 9.6.0 to 12.11.0 (latest)
- Now using official prebuilt `cypress/included` docker image.
- Removed custom docker image setup (Dockerfile and publish script).
- Test images updated due to rendering differences (now using Chromium 106 vs previously 94)

## Testing

N/A

## Docs

N/A


FlyersPh9 (approved):

gretanausedaite (approved):

apps/storybook/src/Breadcrumbs.test.ts
```diff
@@ -20,7 +20,8 @@ describe('Breadcrumbs', () => {
       cy.visit('iframe', { qs: { id } });
 
       if (testName === 'Custom Overflow Dropdown') {
-        cy.get('.iui-button').eq(1).click();
+        cy.get('[aria-label="Dropdown with more breadcrumbs"').click();
+        cy.wait(1000); // wait for dropdown to open?
```
mayank99, line 24: It seems like despite all these changes, it still fails 2 out of 3 times because the dropdown doesn't even open.

If it continues to be flaky, I think we should exclude this particular story.

PR #1264, opened 2023-05-09
move `cra` out of `playgrounds/`

mayank99:
## Changes

After #1233, `yarn dev` was no longer launching the dev servers for playgrounds because Turborepo doesn't like this line:
```
"playgrounds/!(cra)"
```

So I reverted that line, moved the `cra` folder out of `playgrounds`, and updated codesandbox links.

## Testing

`yarn dev` works correctly now.

## Docs

N/A


gretanausedaite (commented):
So CRA is only used for online sandboxes? Could CRA folder be renamed to better show its purpose? Right now it's a bit confusing what is CRA and why do we need that.

mayank99:
> So CRA is only used for online sandboxes?

Yes, currently. Although it could also be used for testing bugs in CRA locally - just would need to run `npm install` and `npm start` manually inside the folder.

> Could CRA folder be renamed to better show its purpose? Right now it's a bit confusing what is CRA and why do we need that.

Added `description` field to package.json explaining what it's for. Open to naming suggestions as well.

gretanausedaite:
> Open to naming suggestions as well.

`cra-sandbox` ?

Issue #969, opened 2023-01-09
Future: Scoped sizing using Context


FlyersPh9:
Add the ability to set size on the entire page & all components match.  Unsure if this will be a breaking change or not.

GerardasB:
It might be useful to display elements in a container component consistently. I.e. a toolbar component defines the size, every toolbar item is rendered with the same size.

FlyersPh9:
From our iTwinUI community meeting:
> localize “Loading.” text in progress indicator

@veekeys

veekeys:
> From our iTwinUI community meeting:
> 
> > localize “Loading.” text in progress indicator
> 
> @veekeys

Probably not only in progress indicator. Can review all the "visually hidden" strings throughout the components.
Next stage would be thinking if extracting all localisable strings into single place would make sense? (I am not sure yet about this one, especially how this would be setup in multiple packages using iTwinUI)

veekeys:
I had one more thought on this..
Actually, if there are strings which should be just localised (same word kept, just different language), those should be localised in iTwinUI itself. That way it could be handled from centralised place rather than each product localising same string all over again.
I am not sure now if there are strings in iTwinUI, which could be changed? 
Datepicker: for sure only localisation
Loading: I guess also just localisation?
Close icons: Localisation

Need to review cases with hardcoded strings. 
That would mean adding some library (react-i18next) and then preparing the library for localisation (moving strings to separate file). And then in case there is any product requiring some language, iTwinUI would need to work with localisation team on this (probably some integration of process in github might be a challenge... )

mayank99:
> That would mean adding some library (react-i18next) and then preparing the library for localisation (moving strings to separate file). And then in case there is any product requiring some language, iTwinUI would need to work with localisation team on this (probably some integration of process in github might be a challenge... )

Is there a defacto library that is used across all applications?

I don't think we should add a dependency on a specific i18n library. The design system should be compatible with whatever library the application prefers to use. If we add a dependency on one i18n library and the application uses another, it could technically still work but would cause unnecessary duplication in the client JS bundle.

FlyersPh9:
iTwinUI is now in maintenance mode per [the support policy](https://github.com/iTwin/iTwinUI/wiki/Support-policy), with [StrataKit](https://github.com/iTwin/stratakit) replacing it as the primary design system.

PR #1265, opened 2023-05-09
replace useTheme with internal useGlobals

mayank99:
## Changes

- Removed `useTheme` hook, which means `ThemeProvider` is now required.
- `ThemeProvider` must always be used with `children`.
- Removed the deprecated `ownerDocument` option in ThemeProvider.
- Added new `useGlobals` hook in every component. Will be used for global setup (e.g. context for #969 and useToaster) and side-effects (e.g. css imports and console warnings).

## Testing

Everything seems to work same as before when app is wrapped in `<ThemeProvider>`.

When not wrapped, the warning is correctly displayed in dev environment.

## Docs

Changeset added. Pending changes in migration guide.


mayank99:
There are some issues in storybook currently because it is not using ThemeProvider with the `theme` prop.

It needs to use the theme from `storybook-dark-mode` and it also needs to use the background from `@storybook/addon-backgrounds`.

mayank99:
Everything looks good except this one story "select--with-selected-value".


@gretanausedaite Maybe you want to investigate?

gretanausedaite:
> @gretanausedaite Maybe you want to investigate?

Controlled tooltip is also broken :)

It appears that Popover does not get `rootRef` from ThemeProvider therefore adds content to owerDocument.body.
I added `id='iui-root'` in theme provider for testing purposes:
image

And added a query for that in Popover:
image

And now Select with controlled Popover and Controlled Tooltip works:
image

My guess this is the same reason why in some cases Popover does not append itself to our theme provider root. (eg. recent ColorPicker issue, AppUI select issue)

mayank99:
> Controlled tooltip is also broken

Both fixed. It looks like `rootRef.current` was undefined on first re-render and correctly set on second render, so when using `visible: true`, it was appending to body. Fixed by using `key`. See https://github.com/iTwin/iTwinUI/pull/1265/commits/d3c5c9e8fd2e27b032a7c09c77565a6143fbe2a5

LostABike (approved):
Ok I have tried to test and toggle globals for everything that has more than a name change, and some random ones with only name change and everything seems to work fine for me (Windows + Chrome). Have also read the actual code changes.

gretanausedaite (approved):
Toasts get theme warning in Storybook. Probably because it's storybook.

LGTM.

mayank99:
> Toasts get theme warning in Storybook. Probably because it's storybook.

I think it's expected, because we are calling a separate `createRoot` inside Toaster. Should be fixed when we provide new `useToaster` api.

PR #1269, opened 2023-05-10
feat(Dialog): All variants use `flex`

### LINKED ISSUE #1268: Dialog should use `flex` on all variants

FlyersPh9:
## Changes

- `.iui-dialog` has `display: flex;` & `flex-direction: column;` rather than the class modifiers.
- Removed `max-height` from `.iui-dialog-content`, added a `max-height` using `dvh` to `.iui-dialog-default` in response to https://github.com/iTwin/iTwinUI/pull/1269#issuecomment-1542670933.

Closes #1268.

## Testing

- Ran visual tests.
- Checked in to ensure overflow is working as expected in browser & iOS Simulator.

## Docs

N/A


.changeset/thirty-lemons-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
+---
+
+All dialog variants have `flex` applied.
```
mayank99, line 5:
```suggestion
All dialog variants have `flex` applied by default. This means the content should be wrapped with `Dialog.Content` or `ModalContent` for optimal layout.
```

.changeset/thirty-lemons-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
```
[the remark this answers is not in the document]
FlyersPh9: Ok, suggestion applied.

.changeset/thirty-lemons-tie.md
FlyersPh9: Ok, suggestion applied.

mayank99 (changes requested):
LGTM. Lets wait for pending v2 PRs to release (#1250) before merging this.

(Blocking accidental merge using "Request changes".)

mayank99:
@FlyersPh9 I vaguely recall there were some hacks we needed to do because `display: flex` was not always present. Think you'll be able to dig through the history and find those?

Here's one I found: https://github.com/iTwin/iTwinUI/pull/1066#issuecomment-1424330775

We might want to make some changes in the dialog code now that we can assume `display: flex` is always present.

FlyersPh9:
> @FlyersPh9 I vaguely recall there were some hacks we needed to do because `display: flex` was not always present. Think you'll be able to dig through the history and find those?
> 
> Here's one I found: [#1066 (comment)](https://github.com/iTwin/iTwinUI/pull/1066#issuecomment-1424330775)
> 
> We might want to make some changes in the dialog code now that we can assume `display: flex` is always present.

Done, added a bullet to the PR description with more deets.

PR #1270, opened 2023-05-10
Unitless `line-height`

FlyersPh9:
## Changes

- Change `line-height` to `1.5` unitless.
- Remove `line-height` or set it to `inherit` where applicable.
- Change `min-height` of tile description to fix overflow issue introduced by `line-height` change.

## Testing

Visual tests

## Docs

None yet.


packages/itwinui-css/backstop/tests/side-navigation.html
FlyersPh9, line 1: Added additional lines to re-introduce a scrollbar due to more dense line height.

packages/itwinui-css/src/button/base.scss
```diff
@@ -12,7 +12,6 @@
   justify-content: center;
   position: relative;
   border-radius: var(--iui-border-radius-1);
-  line-height: 1.2;
```
FlyersPh9, line 15: Might need to revert this, some buttons (but not all) look like the label & icon are not vertically centered after this change.

packages/itwinui-css/src/input/input.scss
```diff
@@ -75,6 +74,7 @@
 }
 
 @mixin textarea {
+  line-height: inherit;
```
FlyersPh9, line 77: An alternative might be to set `all: unset;` instead.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -237,7 +237,7 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
   margin-bottom: var(--iui-size-s);
   font-size: var(--iui-font-size-1);
   height: 100%;
-  max-height: calc(var(--iui-size-s) * 6);
+  max-height: 4.5em;
```
FlyersPh9, line 240: `max-height` changed to `em`.

1.5 line height * 3 desired lines of text = 4.5em.

FlyersPh9:
Wondering if I should remove all image changes for code review & then add images?  1,150 changes makes my browser lag when opening the [files changed](https://github.com/iTwin/iTwinUI/pull/1270/files?file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.md&file-filters%5B%5D=.mdx&file-filters%5B%5D=.scss&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true).

packages/itwinui-css/src/tile/tile.scss
mayank99: Would `lh`/`rlh` be useful? https://css-tricks.com/lh-and-rlh-units/

Can use it with fallback for unsupported browsers:
```suggestion
  max-height: 4.5em;
  max-height: 3lh;
```
https://caniuse.com/mdn-css_types_length_lh

packages/itwinui-css/src/button/base.scss
FlyersPh9: Reverted.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Ooh, either I didn't know of or forgot about `lh`!  Added in 2 places within `tile.scss`.

packages/itwinui-css/src/global.scss
```diff
@@ -16,7 +16,7 @@
 :where(.iui-root) {
   color: var(--iui-color-text);
   font-size: var(--iui-font-size-1);
-  line-height: var(--iui-size-l);
+  line-height: 1.5;
```
LostABike, line 19: If I understand the change correctly, all `line-height` for specific components are removed and they now all inherit from this global value off root? Also, what is unitless? Is that implied to be pixel? And shouldn't we still have a variable for this too?

FlyersPh9: Here's an article on it which will explain better than I can: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values

My basic explanation is the line height will adjust based on the font size.  12px text will have small line height, 32px text will have larger small height.

mayank99 (approved):

LostABike (approved):

PR #1275, opened 2023-05-12
add exception for svg when reverting v1 styles

mayank99:
## Changes

Problem: https://stackblitz.com/edit/vitejs-vite-12mizp?file=src%2FApp.jsx

This was happening because the properties that are set on the internal svg elements (through html attributes) were being reverted.

Solution: Added an exception for `svg *` when reverting.

## Testing

Before:
no v2 spinner

After:
v2 spinner!

## Docs

N/A

### DIFF
--- a/.changeset/fluffy-dogs-vanish.md
+++ b/.changeset/fluffy-dogs-vanish.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue where `ProgressRadial` was not correctly showing when used inside a v2 boundary within a v1 page.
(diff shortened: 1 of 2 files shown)


gretanausedaite (approved):

mayank99:
Initially I was a bit hesitant but now I feel confident about this approach after looking at [The New CSS Reset](https://elad2412.github.io/the-new-css-reset/) which also adds an except for `svg *`.


The other interesting part in that selector is `img` which can have similar issues as `svg`. But no one has reported any problems so far; if someone does, then we can think about adding an exception for `img` in the future.

veekeys:
> Initially I was a bit hesitant but now I feel confident about this approach after looking at [The New CSS Reset](https://elad2412.github.io/the-new-css-reset/) which also adds an exception for `svg *`. 
> 
> The other interesting part in that selector is `img` which can have similar issues as `svg`. But no one has reported any problems so far; if someone does, then we can think about adding an exception for `img` in the future.

looking at this, do you think it makes sense for you to tackle svg itself too, not only svg children?

mayank99:
> looking at this, do you think it makes sense for you to tackle svg itself too, not only svg children?

I don't think so. We target `svg` in lots of places, so we _do_ want to revert v1 styles for them.

PR #1276, opened 2023-05-12
fix Popover controlled mode during first render

mayank99:
## Changes

Ported from https://github.com/iTwin/iTwinUI/pull/1265#issuecomment-1545755572 for release in v2. It could fix some of the issues with our Popover.

## Testing

Fixes the weird edge cases that I was facing in #1265, and doesn't cause any existing tests to break, which is good.

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/utils/components/Popover.tsx
+++ b/packages/itwinui-react/src/core/utils/components/Popover.tsx
@@ -8,7 +8,7 @@ import cx from 'classnames';
 import Tippy from '@tippyjs/react';
 import type { TippyProps } from '@tippyjs/react';
 import type { Placement, Instance } from 'tippy.js';
-import { useMergedRefs } from '../hooks/useMergedRefs.js';
+import { useMergedRefs, useIsClient } from '../hooks/index.js';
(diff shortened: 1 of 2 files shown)


mayank99:
This could maybe help with https://github.com/iTwin/appui/issues/271 (unsure).

LostABike (approved):
This is the same change in #1265 , wouldn't that just override it?

mayank99:
> This is the same change in #1265 , wouldn't that just override it?

This PR is meant to be released in the next v2 release (#1250), whereas #1265 is a breaking change that won't be released until v3 (a few months from now).

packages/itwinui-react/src/core/utils/components/Popover.tsx
```diff
@@ -156,3 +161,9 @@ export const hideOnEscOrTab = {
 };
 
 export default Popover;
+
+const useIsDomAvailable = () => {
```
gretanausedaite, line 165: Does it belong in hooks folder?

gretanausedaite (approved):
LGTM

PR #1282, opened 2023-05-15
set docker user only in CI

mayank99:


mayank99:
@adamhock @LostABike Please test this.

LostABike (approved):
I tested this: switched to your branch and ran `yarn test` in .\packages\iTwinUI-css and it works. Tests are being run and no more error messages.

PR #1287, opened 2023-05-17
Search feature for documentation site

rohan-kadkol:
## Changes

This PR adds a chatbot (powered by OpenAI) to help users on the iTwinUI documentation site. All the code is in `/features/search`. I added a `README.md` inside that folder.

I worked mainly on the logic but less on the UI. When the logic starts working well, I will give more time to the UI.

There still is some testing to do to find out which queries provide incorrect results or even correct but poor results. Thus, your feedback after testing would be greatly appreciated 🙂

## Logic/workflow

* `create_chunks/create_chunks.js` reads all `.mdx` files in `apps/website/src/pages/docs` and creates `docs_chunks.json` with an array of chunks for each new heading level. This step can run in the CI/CD after any change to any docs mdx file.
  * ```jsx
    // Sample input:
	lines = [
	 'Alert is an element',
	 '## Appearance',
	 'Alert should be concise',
	 '### Informational',
	 'Default style',
	 '## Placement',
	 'Either inline or sticky',
	]
	title = 'Alert'

	// Sample output:
	[
	 {
	   "header": ['Alert'],
	   "content": 'Alert is an element',
	 },
	 {
	   "header": ['Alert', 'Appearance'],
	   "content": 'Alert should be concise',
	 },
	 {
	   "header": ['Alert', 'Appearance', 'Informational'],
	   "content": 'Default style',
	 },
	 {
	   "header": ['Alert', 'Appearance', 'Placement'],
	   "content": 'Either inline or sticky',
	 },
	]
  	```
* We then generate embeddings for each chunk ([info](https://openai.com/blog/introducing-text-and-code-embeddings))
* When the user types a query, we encode the query too.
  * We then compare the query vector with all the saved vectors for each chunk.
  * We give the N (some number) most similar chunks to OpenAI's text generation API to summarize the chunks to answer the user's query.


apps/website/src/pages/_global.astro
```diff
@@ -81,10 +81,10 @@ const { content = {} } = Astro.props;
...
       }
 
-      :where(ul:not(.content-wrapper *)) {
-        list-style: none;
-        padding: 0;
-      }
+      // :where(ul:not(.content-wrapper *)) {
+      //   list-style: none;
+      //   padding: 0;
+      // }
```
rohan-kadkol, line 87: I just commented this out to make the markdown bullet points in the search's response appear correctly. Not sure if this breaks anything

rohan-kadkol:
Superseded by #1414

PR #1296, opened 2023-05-18
remove deprecated typography components

mayank99:
## Changes

_yeet!_

I'm working on some repo-wide changes and these were getting in the way.

## Testing

N/A

## Docs

Added changeset.


gretanausedaite:
> I'm working on some repo-wide changes and these were getting in the way.

What was the problem?

gretanausedaite (approved):
Do we have any recommendation for users how to use Text component in their apps? How to add correct padding?
eg. wrap in div or span or something.

gretanausedaite:
> Oh, do you mean the `iui-text-spacing` class? iirc we were not happy with the spacing it adds because it is undesirable in most cases and even when it's desirable, it can be too large (constant 22px). But I don't think we have any recommendations

Yeah, when user wants to add padding at the bottom of `<Text variant='heading'/>`, what is the recommended way to do that. We definitely should document which spacing variables to use with which variant and show some examples how to do that in our docs site.

mayank99:
It's a bit difficult to even come up with a recommendation. We can't just say something like "add `margin-bottom: var(--iui-size-m)`". The desired value will be different in every case.

Maybe we can just say "use margin or gap with one of the values from itwinui-variables".

gretanausedaite:
> Maybe we can just say "use margin or gap with one of the values from itwinui-variables".

Lets move the discussion out of PR. It's documentation issue, not this PR.

PR #1300, opened 2023-05-19
add portal container to context

mayank99:
## Changes

A simple `<div>` will be created under `ThemeProvider` and its ref will be stored in context. Modal/Dialog/Popover will read this ref from context and allow using `createPortal` to render those elements out of the main tree.

Removed old apis from Modal and replaced with new `portal` prop. Defaults to `true` in Modal and `false` in Dialog.

In future PRs, new Tooltip/Popover and Toast apis can also make use of `portalContainerRef`.

## Testing

Tested in storybook by inspecting dom.

Also updated failing unit tests.

## Docs

Added changeset.


gretanausedaite (approved):

LostABike (approved):
I reviewed code, LGTM 👍

PR #1302, opened 2023-05-23
css modules

mayank99:
## Changes

Created a `styles.js` folder inside `src` that uses Vite to do CSS-modules processing. After processing, the `iui-` prefix becomes `_iui3-` (for example `iui-button` becomes `_iui3-button`). Not including hash yet to ease migration, but it will be a one-line change to add it in v4.

Vite will output two files: `styles.css` and  `style.js`, both of which get copied into the published output of `@itwin/itwinui-react`. The `styles.css` file will need to be imported manually by users, and the `styles.js` file will be used inside `polymorphic` to get the transformed class names.

Made some tweaks in `@itwin/itwinui-css` to remove previous scoping mechanism (from #1031). Added `:where` to `@itwin/itwinui-variables` and required `iui-root` in all global styles.

Pending issues:
- [x] selectors in unit tests. fixed using jest mock
- [x] some styles broken. fixed by updating component code.
  - ~~added exception for Table because it will be rewritten~~
  - also should improve after pending PRs (#1295, #1255)

See branch preview: https://itwin.github.io/iTwinUI/1302/react/

## Testing

Tested running dev and prod build of playgrounds and storybook. More extensive testing needed.

## Docs

Updated starter templates and "getting started" pages. Added multiple changesets.

PR #1295, opened 2023-05-18
v3: Menu is using ListItem

gretanausedaite:
## Changes

<!--
WIP

PR #1304, opened 2023-05-23
remove use of `cloneElement` for adding classes

mayank99:
*Follow-up to #1298. Needed for #1302.*

## Changes

Removed all* instances of `cloneElement` that were used to set `iui-` classes. Replaced them all with `<Box>` wrapper and made necessary changes in CSS. Most of them required targeting the child element, while some of them needed different styles on the wrapper and child.

*didn't touch:
- MenuItem, ComboBox: will be addressed in #1295
- Tile: will be addressed in #1255

---

Reviewing recommendation:
- first review component `.tsx`/`.html` files
- then review `.scss`
- then review tests and anything else remaining

## Testing

Updated selectors in unit tests.

All visual tests should pass. For some reason had to update Header images in react (tiny 0.1% change in logo).

## Docs

Added changeset.


packages/itwinui-react/src/core/Header/HeaderBasicButton.tsx
```diff
@@ -33,15 +33,9 @@ export const HeaderBasicButton = React.forwardRef((props, ref) => {
       style={style}
       {...rest}
     >
-      {startIcon &&
-        React.cloneElement(startIcon, {
-          className: startIcon.props.className,
-        })}
+      {startIcon}
```
LostABike, line 36: What is the difference? i.e why was `cloneElement` used in the first place? Does it mean that we could achieve the same thing but with CSS adjustments?

mayank99: if you're asking only about `HeaderBasicButton`, i think in this case it did nothing because we are simply assigning the classname that is already there

LostABike (approved):
I checked the code and tested relevant changes on Storybook and HTML pages, everything seems to be working as intended so I think this is good.

gretanausedaite (approved):
Looks good

PR #1309, opened 2023-05-26
fix: `FileUpload` missing styles

veekeys:
## Changes

- Replaced `label` element with `Anchor` component, so css gets imported for it.
- Import missed css in `FileUploadCard`
- Use `VisuallyHidden` instead of class. Probably, this was also broken.

## Testing

Before my change, this is the outcome for `FileUpload`
image

After:
image

Can use this code
```
 <FileUpload onFileDropped={() => {}}>
       <FileUploadTemplate>nothing</FileUploadTemplate>
 </FileUpload>
```

Important thing I have noticed:
dev build in Vite (at least) includes all the css from all available components. I was pretty sure that was not the case some time ago... That means, that testing environment with dev build is not really isolated.
As a result, the same happens in all the code sandboxes in examples..
Storybook also has all the css included, so you would never reproduce it there.

This was not a bug in V1, because back then `iui-anchor` style was included in global ( I think).
I would suggest reviewing your testing environments to ensure no stuff from outside ends up included.

### DIFF
--- a/.changeset/spicy-tomatoes-leave.md
+++ b/.changeset/spicy-tomatoes-leave.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+`FileUpload`: fix missing anchor style
--- a/packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
+++ b/packages/itwinui-react/src/core/FileUpload/FileUploadCard.tsx
@@ -10,10 +10,12 @@ import {
   useSafeContext,
   useId,
   mergeEventHandlers,
+  VisuallyHidden,
 } from '../utils/index.js';
 import cx from 'classnames';
 import { FileEmptyCard } from './FileEmptyCard.js';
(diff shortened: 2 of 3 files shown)


gretanausedaite (approved):

mayank99:
> dev build in Vite (at least) includes all the css from all available components. I was pretty sure that was not the case some time ago... Storybook also has all the css included, so you would never reproduce it there.

I think it has always been the case, at least in the last year or so. I've noticed CRA also does the same in dev server.

> I would suggest reviewing your testing environments to ensure no stuff from outside ends up included.

But it's okay, because in v3, we will bundle all CSS in a single file and ask users to manually import. Should behave more predictably then.

mayank99 (approved):
LGTM. Will merge into `main` first and then `dev`.

mayank99:
Released in [`2.11.3`](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%402.11.3). Thanks @veekeys

PR #1311, opened 2023-05-29
Using FloatingUI with Tooltips

gretanausedaite:
## Changes

Added `FloatingUI` and used it in Tooltip:
- Updated `Tooltip` to use floatingUI.
- Tooltip uses portal
- Updated reference functionality to use callback and set ref to user provided element
- Added pointer events to the Tooltip to allow hovering and selection

## Testing

- [x] Unit tests
- [x] Visual tests


gretanausedaite:
@mayank99 Checkpoint if the approach is correct:
Added useTooltip hook and NewTooltip React component that mimics almost all functionality previous tooltip had. I think we could make useTooltip hook to be more generic and add more options to it for customization (Tooltip vs Dialog vs Menu). 
I think it's ok if we don't have same functionality as previously on Tooltips, eg. trigger prop.

mayank99:
> I think we could make useTooltip hook to be more generic and add more options to it for customization (Tooltip vs Dialog vs Menu).

On the contrary, I think this hook should be moved inside the Tooltip file. Tooltip is very different from Menu, so it would be better not to share code.

---

In general looks ok so far, other than three comments.

1. is it possible to avoid the extra div here? i am going to add an extra div in #1300 so it would be nice if we could just use that instead.


2. we will definitely need a bunch of customization options. three important ones: `appendTo` equivalent, ability to turn off modifiers, ability to change between `aria-describedby`, `aria-labelledby`, and nothing.

3. we need to try to make this tooltip more accessible than the previous one. for example, hovering away from the trigger and onto the tooltip should not dismiss the tooltip. and pressing Escape should dismiss it. See https://sarahmhigley.com/writing/tooltips-in-wcag-21/

gretanausedaite:
> 1. is it possible to avoid the extra div here? i am going to add an extra div in [add portal container to context #1300](https://github.com/iTwin/iTwinUI/pull/1300) so it would be nice if we could just use that instead.

That div comes from FloatingPortal I am using to portal tooltip on given element. I think we could use their hook if we want to have just one portal https://floating-ui.com/docs/FloatingPortal#usefloatingportalnode

> 2. we will definitely need a bunch of customization options. three important ones: `appendTo` equivalent, ability to turn off modifiers, ability to change between `aria-describedby`, `aria-labelledby`, and nothing.

We can easily override aria props. I added appendTo which goes to portal element.
TooltipContent:
image

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+    whileElementsMounted: (referenceEl, floatingEl, update) =>
+      autoUpdate(referenceEl, floatingEl, update, {
+        animationFrame: followTrigger,
+      }),
+    middleware: [offset(5), flip(), shift()],
+  });
+
+  const context = data.context;
+
+  const hover = useHover(context);
```
gretanausedaite, line 89: Should be allow users to customise interactions? eg. disable tooltip on focus.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+
+  const data = useFloating({
+    placement,
+    open,
+    onOpenChange: setOpen,
+    whileElementsMounted: (referenceEl, floatingEl, update) =>
+      autoUpdate(referenceEl, floatingEl, update, {
+        animationFrame: followTrigger,
+      }),
+    middleware: [offset(5), flip(), shift()],
```
gretanausedaite, line 84: Allow users to customise middleware?

gretanausedaite (commented):
@mayank99 I think this is in a good shape for review.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
- * <Tooltip content='tooltip text' reference={buttonRef} />
- */
-export const Tooltip = (props: TooltipProps) => {
-  const { content, children, className, style, visible, ref, id, ...rest } =
-    props;
+const useTooltip = ({
+  placement = 'top',
+  visible: controlledOpen,
+  toggleVisible: setControlledOpen,
+  followTrigger = false,
```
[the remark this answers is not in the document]
gretanausedaite: It's suggested to use this option sparingly so I chose opt in approach: https://floating-ui.com/docs/autoUpdate#animationframe

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: yes. see https://github.com/iTwin/iTwinUI-react/issues/783 and https://github.com/iTwin/iTwinUI-react/issues/947

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
    */
-  children?: JSX.Element;
-} & Omit<PopoverProps, 'className'> &
-  Omit<CommonProps, 'title'>;
+  children?: React.ReactNode;
+  /**
+   * Element to append tooltip to.
+   * Appends to ThemeProvider portalContainerRef by default.
+   */
+  appendTo?: HTMLElement;
```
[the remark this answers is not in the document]
gretanausedaite: renamed to portal. Not adding boolean version for now.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+      )
+    : null;
+  // eslint-disable-next-line @typescript-eslint/ban-types
+}) as PolymorphicForwardRefComponent<'div', {}>;
+
+const TooltipContent = React.forwardRef((props, propRef) => {
+  const { children, className, ...rest } = props;
+  const context = useTooltipContext();
+  const themeInfo = React.useContext(ThemeContext);
+  const ref = useMergeRefs([context.refs.setFloating, propRef]);
```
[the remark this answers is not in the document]
gretanausedaite: Oops, accidentally used FloatingUI useMergeRefs functions instead of ours.
Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+        children,
+        context.getReferenceProps({
+          ref,
+          ...rest,
+          ...children.props,
+        }),
+      )
+    : null;
+  // eslint-disable-next-line @typescript-eslint/ban-types
+}) as PolymorphicForwardRefComponent<'div', {}>;
```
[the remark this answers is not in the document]
gretanausedaite: Updated

packages/itwinui-react/package.json
```diff
@@ -82,6 +82,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.12.10",
+    "@floating-ui/react": "^0.24.1",
```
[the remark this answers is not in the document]
gretanausedaite: Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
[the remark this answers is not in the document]
mayank99: why not? that's the whole point of that prop - it will give a simple way to toggle between `appendTo={portalContainer}` and `appendTo='parent'`.

without boolean, there is no easy way to append to parent.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +54,141 @@ export type TooltipProps = {
...
+
+  const data = useFloating({
+    placement,
+    open,
+    onOpenChange: setOpen,
+    whileElementsMounted: (referenceEl, floatingEl, update) =>
+      autoUpdate(referenceEl, floatingEl, update, {
+        animationFrame: followTrigger,
+      }),
+    middleware: [offset(5), flip(), shift()],
```
mayank99, line 84: i think we were using `4` as offset before

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +54,141 @@ export type TooltipProps = {
...
+  });
+
+  const context = data.context;
+
+  const hover = useHover(context);
```
mayank99, line 89: hover should have some delay and also use safePolygon.
currently the tooltip closes as soon as you leave the reference element, so it fails [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html) (see "Hoverable" part)

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +54,141 @@ export type TooltipProps = {
...
+      open,
+      setOpen,
+      ...interactions,
+      ...data,
+    }),
+    [open, setOpen, interactions, data],
+  );
+};
+
+const TooltipContext = React.createContext<
```
mayank99, line 111: if we are not exposing subcomponents, then do we need context? would be simpler to do it all in one component.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Maybe adding prop would be ok?
```ts
middleware?: {
  offset?: number;
  flip?: boolean;
  ..etc;
}
```

mayank99: Yeah, lets control all the names now. Other option would be to allow user to pass any and all middleware, but then they might start doing weird things

I also think it should be possible for the user to toggle `autoUpdate`

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Updated, will add user overrides in future commit

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: For future, I think we want to have subcomponents for Tooltip so it would be easy transition from props to that.
Keeping props for now for easier transition.
*** 
Though I don't know, users would need to update props anyways so not sure of the benefit to keep old API (which is only content and children props).

mayank99: the benefit of `children` and `content` is convenience.

It is easier to do this:
```jsx
<Tooltip content='close'>
 <Button>❌</Button>
</Tooltip>
```
than this:
```jsx
<Tooltip>
 <Tooltip.Trigger>
  <Button>❌</Button>
 </Tooltip.Trigger>
 <Tooltip.Content>
  close
 </Tooltip.Content>
</Tooltip>
```

whereas I'm not sure about the benefit of subcomponents here because the trigger does not render anything, it does a cloneElement, so there is only one "real" element.

also, right now we are using `children` as trigger, so it locks us out of adding subcomponents inside `<Tooltip>`. which means if we actually want subcomponents, we might need to add a `<Tooltip.Wrapper>`. see https://github.com/iTwin/iTwinUI/pull/1328#issuecomment-1582974467

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: it would be nice to give control, but i fear that developers will start removing focus as trigger, making tooltips inaccessible

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+  if (!context.open) {
+    return null;
+  }
+
+  const portalRoot = context.appendTo ?? themeInfo?.portalContainerRef?.current;
+
+  return portalRoot ? (
+    ReactDOM.createPortal(contentBox, portalRoot)
+  ) : (
+    <FloatingPortal>{contentBox}</FloatingPortal>
```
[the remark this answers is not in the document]
gretanausedaite: Removed.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,192 @@
...
+   */
+  followTrigger?: boolean;
+};
+
+type TooltipOwnProps = {
   /**
    * Content of the tooltip.
    */
-  content: React.ReactNode;
+  content?: React.ReactNode;
```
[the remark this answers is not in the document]
gretanausedaite: Added back

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +63,116 @@ export type TooltipProps = {
...
+      : portal
+      ? context?.portalContainerRef?.current ?? getDocument()?.body
+      : null;
+
+  const contentBox = (
+    <Box
+      className={cx('iui-tooltip')}
+      ref={tooltip.refs.setFloating}
+      style={tooltip.floatingStyles}
+      {...tooltip.getFloatingProps()}
```
mayank99, line 146: missing forwardRef, cx, rest props, and need to merge style prop too

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +63,116 @@ export type TooltipProps = {
...
+      {React.isValidElement(children)
+        ? React.cloneElement(
+            children,
+            tooltip.getReferenceProps({
+              ref: tooltip.refs.setReference,
+              ...children.props,
+            }),
+          )
+        : null}
+      {tooltip.open
```
mayank99, line 163: Do we need to add an extra `useClient` check (like in Dialog) or can we trust this `tooltip.open` state? Try testing in nextjs playground with `portal={false}` and make sure there is no [hydration error](https://nextjs.org/docs/messages/react-hydration-error) in console.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +63,116 @@ export type TooltipProps = {
...
 };
 
-export default Tooltip;
+/**
+ * Basic tooltip component to display informative content when an element is hovered or focused.
+ * Uses [FloatingUI](https://floating-ui.com/).
+ * @example
+ * <Tooltip content='tooltip text' placement='top'>Hover here</Tooltip>
+ */
+export const Tooltip = TooltipComponent;
```
mayank99, line 178: not needed i think. `TooltipComponent` can be renamed to `Tooltip` and exported directly.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,10 +4,57 @@
...
+  /**
+   * Use this if you want Tooltip to follow moving trigger element
+   * @default false;
+   */
+  followTrigger?: boolean;
+  /**
+   *
+   */
+  middleware?: {
+    offset?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
mayank99, line 47: Can be booleans. Or an array of string literals if you prefer.

Also should be renamed from `middleware` to `modifiers` or `options`? Because it includes more than just middleware

mayank99: also it's not used anywhere yet 😅

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: also missing controlled `enabled` (similar to `useFocus`)

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -16,46 +71,148 @@ export type TooltipProps = {
...
-        </Box>
-      }
-      offset={[0, 4]}
-      ref={ref}
+  const portalTo =
+    typeof portal !== 'boolean'
+      ? portal.to
+      : portal
+      ? context?.portalContainerRef?.current ?? getDocument()?.body
+      : null;
```
LostABike, line 184: This is a little hard for me to understand. Can you explain what this means and what is it for?

apps/storybook/cypress-visual-screenshots/baseline/AvatarGroup.test.ts-With Tooltip.png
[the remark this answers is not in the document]
gretanausedaite: Fixed the reference!

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Fixed this one.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Added

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Fixed

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,10 +4,57 @@
...
+   */
+  placement?: Placement;
+  /**
+   * Property for manual visibility control
+   */
+  visible?: boolean;
+  /**
+   * Function that allows users to control Tooltip visibility
+   */
+  toggleVisible?: (open: boolean) => void;
```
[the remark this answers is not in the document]
gretanausedaite: Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: prop `portal` can be boolean (true or false) or object with `to` value.

If portal is not boolean, we add it to the user defined element.
If portal is true - add it to our div that is set in theme provider or document body (if portalContainer is not there).
If portal is false - add it the tooltip next to the element.

Play around with these settings and see in dev tools where tooltip is added.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: We don't need useClient check. Next playground runs without any errors.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,250 @@
...
+  placement?: Placement;
+  /**
+   * Property for manual visibility control
+   */
+  visible?: boolean;
+  /**
+   * autoUpdate options
+   * https://floating-ui.com/docs/autoUpdate#options
+   */
+  updateOptions?: {
```
mayank99, line 44: should be called `autoUpdateOptions` maybe?

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,250 @@
...
-    </Popover>
+      {content}
+    </Box>
   );
-};
+
+  const childrenRef =
+    React.isValidElement(children) &&
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useMergedRefs(
```
mayank99, line 227: shouldn't call hooks conditionally. Try using `mergeRefs` functions instead?

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -4,58 +4,250 @@
...
-          role='tooltip'
-          id={id}
-        >
-          {content}
-        </Box>
-      }
-      offset={[0, 4]}
-      ref={ref}
+  React.useEffect(() => {
+    setReference && setReference(tooltip.refs.setReference);
```
mayank99, line 202: should we instead ask for a `reference` and set it ourselves?

e.g.
```js
const { reference, ... } = props;


React.useEffect(() => {
  if (reference) {
    tooltip.refs.setReference(reference);
  }
}, [reference]);
```

gretanausedaite: Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Updated

mayank99: > By default everything is false.

This is not true. Some of these are true by default in floating-ui. https://floating-ui.com/docs/autoupdate#options

Also missing `layoutShift` option.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: update all leftover jsdocs to say `reference` instead of `setReference`

mayank99: and why not include correct deps in useEffect instead of ignoring eslint?


similarly on line 128, `any` can be removed.

apps/storybook/cypress-visual-screenshots/baseline/Breadcrumbs.test.ts-Overflow.png
mayank99, line 1: Possible to revert image changes?

packages/itwinui-react/setupTests.ts
```diff
@@ -6,6 +6,11 @@ import '@testing-library/jest-dom';
 
 window.HTMLElement.prototype.scrollIntoView = () => {};
 window.HTMLElement.prototype.scrollTo = () => {};
+window.ResizeObserver = jest.fn().mockImplementation(() => ({
```
mayank99, line 9: This could be useful to mention in migration guide. Just a short note "Since floating-ui uses `ResizeObserver`, you might need to add a mock in your tests.".

gretanausedaite: Will mention this once this is merged.

apps/storybook/cypress-visual-screenshots/baseline/Breadcrumbs.test.ts-Overflow.png
gretanausedaite: I don't know why it still shows the image change. I tried to revert it.

mayank99: `git checkout origin/dev apps/storybook/cypress-visual-screenshots/baseline/Breadcrumbs.test.ts-Overflow.png`

gretanausedaite: Should be fixed

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: autoUpdateOptions were added.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Not giving control for now. We can add that later.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Oops, added

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Updated

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
gretanausedaite: Update seems to fix the issue

mayank99 (approved):
LGTM. 🚀

One more thing I would like to see is the ability to switch between `aria-describedby` vs `aria-labelledby` vs no aria. This will allow us to use `<Tooltip>` inside `<IconButton>`. But it can be done in a future PR.

Another thing we should do is add [`FloatingDelayGroup`](https://floating-ui.com/docs/floatingdelaygroup) inside `ButtonGroup`. Again can be done in a future PR.

LostABike (commented):
I think because now all tooltips default to `top` placement, the demos in Storybook will look weird, like this:

image

Also the behavior has changed too: before you can just hover on the element to see tooltip, but now we have to click on it. I'm unsure if this is intended and we want to change default behavior to be that way, if so ignore :D

mayank99:
works fine for me fwiw. it defaults to `top` but should automatically flip to bottom when there is no space.

https://github.com/iTwin/iTwinUI/assets/9084735/44ecd2c7-e1c2-467d-b0ca-2919514ee209

LostABike:
> works fine for me fwiw. it defaults to `top` but should automatically flip to bottom when there is no space.
> 
>  Screen.Recording.2023-06-29.at.2.21.46.PM.mov

ah you're right. I just re-installed everything and it works correctly for me now 😵

LostABike (approved):

.changeset/silly-roses-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+- Removed Tippy.js as dependency in Tooltip. Using FloatinUI instead.
```
LostABike, line 5: nit 🐱 

```suggestion
- Removed Tippy.js as dependency in Tooltip. Using FloatingUI instead.
```

LostABike (approved):

PR #1319, opened 2023-06-01
fix(DatePicker): allow month and year change even if date is disabled

veekeys:
## Changes

I have removed disabled state from month and year. This should be allowed in general as this is view only and only the day selection should be disabled.
Valid case:

image

Select previous day than today is disabled. But currently selected date is in the disabled range. In this case, all navigation to next month is disabled (but you need to choose a new date which is not in the past)

### DIFF
--- a/packages/itwinui-react/src/core/DatePicker/DatePicker.test.tsx
+++ b/packages/itwinui-react/src/core/DatePicker/DatePicker.test.tsx
@@ -491,9 +491,32 @@ it('should prevent selecting disabled dates', async () => {
   await userEvent.keyboard('{Enter}');
   expect(onClick).not.toHaveBeenCalled();
 
-  // next month/year button should be disabled
-  expect(screen.getByLabelText('Previous year')).not.toBeDisabled();
-  expect(screen.getByLabelText('Previous month')).not.toBeDisabled();
-  expect(screen.getByLabelText('Next month')).toBeDisabled();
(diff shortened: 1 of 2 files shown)


mayank99:
I'm not sure I understand the valid case for it. Can you try to elaborate/rephrase?

veekeys:
> I'm not sure I understand the valid case for it. Can you try to elaborate/rephrase?

Sure.
There is an assignment which has due date. It had expired, but user is able to set a new due date, which cannot be in the past. We do show currently selected date in the date picker and if it is not in current month, you wont be able to go and set anew date (all buttons are disabled).

.changeset/orange-cats-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+DatePicker: allow year and month navigation in disabled range
```
mayank99, line 5:
```suggestion
DatePicker will no longer disable month and year navigation when the next/previous months are disabled.
```

mayank99 (approved):
Okay yeah, that makes sense.

One suggestion for changeset:

.changeset/orange-cats-yell.md
veekeys: Done, I think

mayank99:
Released in 2.11.4! Thanks

PR #1321, opened 2023-06-01
Add "disabled" prop to ExpandableBlock component

xman343:
## Changes

Added disabled prop to the ExpandableBlock component that activates 'disabled' CSS styling and makes it unresponsive to onClick events when prop is set to true.

## Testing

Visual tests and unit tests were performed. Visual test assessed whether the 'disabled' styling was displayed properly. Unit tests ensured that the disabled styling was activated properly and that the ExpandableBlock was unresponsive to onClick events when disabled.

## Docs

Update included in changeset and stories.
-->


packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -57,7 +58,12 @@ type ExpandableBlockProps = {
    * @default 'default'
    */
   styleType?: 'default' | 'borderless';
-};
+  /**
+   * Disables ExpandableBlock.
+   * @default false
+   */
+  disabled?: boolean;
+} & Omit<CommonProps, 'title'>;
```
mayank99, line 66: `CommonProps` are not necessary here

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -129,8 +141,9 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
       <Box
         aria-expanded={expanded}
         className='iui-header'
+        aria-disabled={disabled}
```
mayank99, line 144: `aria-disabled` (and `aria-expanded`) is [not supported](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles) on a generic role-less div.

Suggestion: add `role='button'` to this element (for now) and then test that it works in [NVDA](https://webaim.org/articles/nvda/).

In a future PR, we can improve this component to use `<button>` instead of `<div role='button'>`. Further reading: https://adrianroselli.com/2020/05/disclosure-widgets.html

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -129,8 +141,9 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
         tabIndex={0}
-        onClick={handleToggle}
+        onClick={() => !disabled && handleToggle()}
```
mayank99, line 146: This `disabled` check could be done inside `handleToggle`

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -119,6 +130,7 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
           'iui-expanded': expanded,
           'iui-small': size === 'small',
           'iui-borderless': styleType === 'borderless',
+          'iui-disabled': disabled,
```
mayank99, line 133: `iui-disabled` is not present in the CSS so it does nothing and can be removed.

.changeset/happy-walls-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added "disabled" prop to ExpandableBlock component. When activated, it triggers the "disabled" CSS styling for the ExpandableBlock and disables user interaction with it. This allows developers to easily disable ExpandableBlock component interactivity when appropriate.
```
mayank99, line 5: Changeset can be made less verbose.

```suggestion
Added `disabled` prop to ExpandableBlock component. When activated, it triggers the "disabled" CSS styling for the ExpandableBlock and disables user interaction with it.
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: added role='button', seems to register as a button on NVDA

xman343: https://github.com/iTwin/iTwinUI/assets/44583237/4fbf04b7-ec38-40b7-9104-732136fd2bd6

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: fixed

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: fixed

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: moved from onClick to handleToggle

mayank99: instead of ternary, it would be better to use an if statement and early return.

```js
if (disabled) {
  return
}

setExpanded(!expanded);
onToggle?.(!expanded);
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.test.tsx
```diff
@@ -180,3 +180,33 @@ it('should render borderless', () => {
...
+    >
+      test content
+    </ExpandableBlock>,
+  );
+
+  const header = container.querySelector('.iui-title') as HTMLElement;
+  expect(header).toBeTruthy();
+  fireEvent.click(header);
+  expect(onToggleMock).toHaveBeenCalledTimes(0);
+});
```
mayank99, line 212: suggestions to improve unit test:
- combine these two tests
- instead of querying by classname, try querying by role or text
- check for `aria-disabled` instead of `iui-disabled`
- also check for `aria-expanded`

```js
it('should respect disabled prop', () => {
  const onToggleMock = jest.fn();
  const { getByRole } = render(
    <ExpandableBlock title='test title' onToggle={onToggleMock} disabled>
      test content
    </ExpandableBlock>,
  );

  const header = getByRole('button');
  expect(header).toBeDisabled();
  fireEvent.click(header);
  expect(onToggleMock).not.toHaveBeenCalled();
  expect(header).toHaveAttribute('aria-expanded, 'false');
});
```

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Open).png
mayank99, line 1: none of the existing test images should have changed. Can you try reverting them and re-running tests (inside docker)?

apps/storybook/src/ExpandableBlock.test.ts
```diff
@@ -11,6 +11,7 @@ describe('ExpandableBlock', () => {
     'Small',
     'Status Icon',
     'With Caption',
+    'Disabled',
```
mayank99, line 14: for disabled block, there is no "open" state to test, so you can add an if condition around lines 22-23

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.test.tsx
xman343: What method do I use to check for aria-disabled?

mayank99: `toBeDisabled()` should work, but if not, then you can use `toHaveAttribute('aria-disabled', 'true')`

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: done

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.test.tsx
xman343: Added suggested unit test

mayank99: unfortunately it is failing so you'll have to change it to `toHaveAttribute`. https://github.com/iTwin/iTwinUI/actions/runs/5148140855/jobs/9269498639

make sure it passes locally before committing

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Open).png
xman343: Images have been reverted to their proper state

apps/storybook/src/ExpandableBlock.test.ts
xman343: image
Does something like this work?

mayank99:
would be good to mention something in the documentation (source: [`expandableblock.mdx`](https://github.com/iTwin/iTwinUI/blob/dev/apps/website/src/pages/docs/expandableblock.mdx), preview: https://localhost:1700/docs/expandableblock), but it can be done in a separate PR if you want

apps/storybook/src/ExpandableBlock.test.ts
mayank99: yup, that's good. just use `!==` instead of `!=`

mayank99: this image can be deleted now. https://github.com/xman343/iTwinUI/blob/xander/expandable-block-disabled-/apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Disabled%20(Open).png

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -101,6 +110,10 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
       return;
     }
 
+    if (disabled) {
```
adamhock, line 113: This if statement can be combined with the if statement above it

adamhock (commented):
I think there should be a website example for this

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
xman343: done

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -127,10 +140,12 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
       {...rest}
     >
       <Box
+        role='button'
         aria-expanded={expanded}
         className='iui-header'
+        aria-disabled={disabled}
         tabIndex={0}
-        onClick={handleToggle}
+        onClick={() => handleToggle()}
```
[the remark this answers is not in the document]
xman343: It was left over from when I was moving the disabled check from onClick to the handleToggle function. 

I cleaned up the line so that it just calls {handleToggle} like in the original version.

adamhock (approved):
The first of many! 🎉

Issue #115, opened 2021-05-26
UserIcon in IconButton is not properly sized in some scenario.


raplemie:
#### Environment
- __Package version(s)__: 0.16.0

Google Chrome | 90.0.4430.212&nbsp;(Official Build)&nbsp;(64-bit)&nbsp;(cohort: Stable)
-- | --
Revision | e3cd97fc771b893b7fd1879196d1215b622c2bed-refs/branch-heads/4430@{#1429}
OS | Windows&nbsp;10 OS Version 1909 (Build 18363.1556)

#### Code Sandbox
Could not create a repo as this issue is not occuring in there, not sure how to control that.

#### Steps to reproduce
```
<IconButton styleType={"borderless"}>
  <UserIcon
    size={"medium"}
    abbreviation={"IU"}
    backgroundColor={getUserColor("IU")}
  />
</IconButton>
```
#### Actual behavior
UserIcon with size "medium" is overriden by default ".iui-button>.iui-icon" **in some scenario** (which I could not reproduce in code sandbox) when the build output changes the css order...

This is what I get in my application:
![image](https://user-images.githubusercontent.com/1904889/119720329-d6bb2600-be37-11eb-8428-8db137982120.png)

This is what I get in Storybook or Codesandbox examples:
![image](https://user-images.githubusercontent.com/1904889/119718592-bf7b3900-be35-11eb-92a9-03e4fadc7a13.png)

#### Expected behavior
I expect the UserIcon to have the specified size even if I put it in a IconButton, so I guess the `.iui-user-icon.iui-user-icon` should be made even more specific so order does not impact the output.

#### Additional information
The specificity of the 2 height/width classes are the same (2 classes), so it boils down to the order in which the classes are loaded in the browser. I dont know why my build is putting the `.iui-user-icon.iui-user-icon` before the `.iui-button>.iui-icon`, but it does...
I'm using a package that uses iTwinui-react components and @bentley/react-scripts, I cant test that in codesandbox as those packages are not publicly available at the moment.

In my test, I was also inside a Header (not set to slim mode), but that didnt seem to add anything regarding the classes.
Not that when in slim mode, the usericon actually get to the correct 24px size, instead of the 16px size of iui-icon.

(I must admit that I initially expected the size to be handled by the header in non slim mode too, but it went to the default small size.)

mayank99:
Hi Raph. Thanks for the issue.

[...]

PR #1322, opened 2023-06-01
remove css imports from components

mayank99:
## Background

Up until now, we were importing `.css` files inside our React components. Such imports are not valid javascript syntax, and we have been relying on bundlers to magically take care of it. In hindsight, this was a mistake, just like automatically setting theme on the document was a mistake (eventually fixed in https://github.com/iTwin/iTwinUI-react/pull/825 and #1265).

Some of the many issues with using such non-standard imports:
- Doesn't work consistently in frameworks where it's supported (see https://github.com/iTwin/iTwinUI/discussions/1194 and https://github.com/iTwin/iTwinUI/pull/1309).
- Requires hacks in some frameworks (e.g. [`next-global-css`](https://www.npmjs.com/package/next-global-css) in older nextjs versions, [`transpilePackages`](https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages) in nextjs 13)
- Requires hacks in some testing scenarios (e.g. jest mocks).
- Doesn't work at all in some frameworks (remix).
- Can lead to lots of duplicate css sometimes (I have personally seen a dozen different versions of iTwinUI styles being included in a production app).
- Can cause issues with CSS ordering (see https://github.com/iTwin/iTwinUI/issues/115; there was another user who recently reached out to me privately with a similar issue).

## Changes

Removed all css imports from every component and created a `styles.css` file that can be manually imported. (This change was initially part of #1302 but pulled out for ease of review.)

## Testing

All tests/playgrounds been updated to explicitly import the css.

## Docs

Updated readme/md pages and added changeset.


mayank99:
> So now we will always import all the css even if we use few components from iTwinUI ? :(

Yes. This is actually a requirement for CSS modules. We cannot scope every component separately, as the same class when used in two different css files would get different hashes.

If you're concerned about bundle size, it's only 30KB gzipped and easily cacheable.

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -15,24 +15,27 @@
 }
 
 .iui-expandable-searchbox {
-  width: var(--_iui-input-min-height);
-  transition: width var(--iui-duration-1) ease-in-out;
+  // increase specificity to override input-flex-container
+  &#{&} {
```
mayank99, line 19: perfectly demonstrates why having one stylesheet is better. previously we were relying on order of imports, `input-container.css` needs to come before `searchbox.css` - this is brittle because user app could additionally import `input-container.css` later.

"correct" solution would be to control the ordering/layering in `all.css`, but for now i'm increasing the specificity here

LostABike (approved):
I think this is good now. I have checked the visuals and they look good/not broken anymore. Also checked code changes and Overview page 👍

gretanausedaite (approved):
Lets do dev release with this and css modules and try to migrate real app (eg. AppUI with its test apps)

gretanausedaite:
> Do you mean from a branch? Because we have changesets set up, so merging the PR will also allow doing a dev release.
No, with changesets. Just saying it's a good idea to test this in real app.

PR #1330, opened 2023-06-07
Deprecate FileUploadTemplate

LostABike:
## Changes

- Removed `FileUploadTemplate` references from:
 + indexes imports
 + Removed the Default example from Storybook because it is the same as the Default File Upload Card example, but with `FileUploadTemplate` instead of `FileUploadCard`.
 + Doc Website (replaced with `FileUploadCard`)
 + marked deprecated in jsdoc
 + updated relevant jsdocs in `FileUpload`

- Did not remove (yet?):
 + `FileUploadTemplate.tsx` itself
 + `FileUploadTemplate.test.tsx`

Do we remove those right now? ^

## Testing

- Tested the changes in vite when replacing `FileUploadTemplate` with `FileUploadCard`, also make sure main example on website works as intended. It doesn't but should work after #1329 merged because I independently tested this fix on this branch.

## Docs

- Changes in jsdocs
- Updated website main example
- Brief description in migration guide

### DIFF
--- a/apps/storybook/src/FileUpload.stories.tsx
+++ b/apps/storybook/src/FileUpload.stories.tsx
@@ -11,7 +11,6 @@ import {
   FileUploadCard,
   FileEmptyCard,
   FileUploadProps,
-  FileUploadTemplate,
   LabeledInput,
   IconButton,
 } from '@itwin/itwinui-react';
@@ -33,26 +32,6 @@ export default {
   },
   title: 'Core/FileUpload',
 } as Meta<FileUploadProps>;
-
-export const Default: Story<FileUploadProps> = (args) => {
-  const [files, setFiles] = useState<Array<File>>([]);
-
-  return (
-    <FileUpload
-      {...args}
-      onFileDropped={(files) => {
-        setFiles(Array.from(files));
-        action(`${files.length} files uploaded`)();
-      }}
-    >
-      <FileUploadTemplate
-        onChange={(e) => setFiles(Array.from(e.target.files || []))}
-      >
-        {files.map((f) => f.name).join(', ')}
-      </FileUploadTemplate>
-    </FileUpload>
-  );
-};
 
 export const WrappingInput: Story<FileUploadProps> = (args) => {
   const [files, setFiles] = useState<Array<File>>([]);
--- a/apps/website/src/examples/FileUpload.main.tsx
+++ b/apps/website/src/examples/FileUpload.main.tsx
@@ -3,20 +3,19 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { FileUpload, FileUploadTemplate } from '@itwin/itwinui-react';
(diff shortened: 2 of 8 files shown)


apps/website/src/examples/FileUpload.main.tsx
```diff
@@ -3,20 +3,19 @@
...
-  const [files, setFiles] = React.useState([]);
-
+  const [files, setFiles] = React.useState<Array<File>>([]);
   return (
     <FileUpload
+      style={{ position: 'absolute' }}
       onFileDropped={(files) => {
-        setFiles(Array.from(files));
+        setFiles(files);
+        action(`${files.length} files uploaded`)();
```
gretanausedaite, line 15: I don't think action is valid in website example.

packages/itwinui-react/src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -41,10 +41,7 @@ export type FileUploadTemplateProps = {
 } & React.ComponentProps<'div'>;
 
 /**
- * Default template to be used with the `FileUpload` wrapper component.
```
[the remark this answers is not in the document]
LostABike: I brought back description.

apps/website/src/examples/FileUpload.main.tsx
LostABike: Right. I removed action

LostABike:
> Deprecated apis need to be kept in index, otherwise it is the same as removal
> 
> Also i think your branch is several commits behind `dev`, so needs to be updated

Ok they are back in index.

gretanausedaite (commented):
Should be good to go with couple changes:
- Move back FileUploadTemplate to line 59 (will undo all changes in index file)
- You deleted Default story, so test images should be cleaned up.

LostABike:
> Should be good to go with couple changes:
> 
> * Move back FileUploadTemplate to line 59 (will undo all changes in index file)
> * You deleted Default story, so test images should be cleaned up.

Thanks. I updated these.

apps/storybook/src/FileUpload.stories.tsx
```diff
@@ -10,7 +10,7 @@ import {
   FileUpload,
   FileUploadCard,
   FileEmptyCard,
-  FileUploadTemplate,
+  FileUploadProps,
```
mayank99, line 13: remove

LostABike: FileUploadProps? removed

examples/FileUpload.main.tsx
```diff
@@ -3,22 +3,21 @@
...
-import { FileUpload, FileUploadTemplate } from '@itwin/itwinui-react';
+import { FileUpload, FileUploadCard } from '@itwin/itwinui-react';
 
 export default () => {
-  const [files, setFiles] = React.useState([]);
-
+  const [files, setFiles] = React.useState<Array<File>>([]);
   return (
     <FileUpload
+      style={{ position: 'absolute' }}
```
mayank99, line 12: why position: absolute?

PR #1335, opened 2023-06-07
fix website styles

mayank99:
## Changes

After #1302, some website demos broke, so fixing them.

In a future PR, i might move all iTwinUI-specific stuff into shadow dom to prevent conflicts with the page.

## Testing

All pages look fine now.

## Docs

N/A


gretanausedaite (approved):

LostABike (commented):
dropdown menus all missing styles/background I think?
image

LostABike:
> > dropdown menus all missing styles/background I think?
> 
> good catch. I could fix this using `appendTo='parent'`, but we don't want our users to start adding that everywhere, so it's better to leave the snippets unchanged.
> 
> I think it might be fixed by #1300 though. Could you review that? We can merge and then update this branch to see if it's fixed.

ok I'll take a look

mayank99:
Unfortunately #1300 didn't fix it, so I've made use of the internal `_iui3-root` class for now to fix demos. [`aade838`](https://github.com/iTwin/iTwinUI/pull/1335/commits/aade838840a2592a3a785e0c22c0365ddf600950)

Will investigate some more later.

LostABike (approved):
I *think* I checked everything, and they all work as intended now 😅

PR #1341, opened 2023-06-08
fix(Table): fix slot column alignment

### LINKED ISSUE #1340: Table: slot columns are moving when using global filter and data has subrows
div with `iui-table-header-actions-container` class is added, but its empty..
`canFilter` on the column is true, but no idea how..
This hook is the issue for it `useSubRowFiltering`.

I wanted to contribute, but so far could not override it..Only fix I can think of is to have same condition as for actual Filter component, which is not being rendered.

https://github.com/iTwin/iTwinUI/assets/81580355/1be3d993-f882-4869-b3a2-2c86fcf0bc79

veekeys:
Closes #1340 

## Changes

I have added missing check in rendering whole actions container, so it is same as the Filter button itself.
Updated full story with global search as it was not really available anywhere in combo with more features.

## Testing

Checked and linked bug is not reproducible anymore.


apps/storybook/src/Table.stories.tsx
```diff
@@ -1702,21 +1702,44 @@ export const Full2: Story<Partial<TableProps>> = (args) => {
...
+      <div style={{ display: 'flex', margin: '8px 0' }}>
+        <Input
+          placeholder='Search...'
+          value={globalFilterValue}
+          onInput={(e: React.ChangeEvent<HTMLInputElement>) =>
+            setGlobalFilterValue(e.target.value)
+          }
+        />
+      </div>
+      <div
```
mayank99, line 1720: it's a lot of extra divs. any chance to reduce the number?

or even just revert the story change.

veekeys: Just copy pasted from the story before..
Do we really care about divs in here..?

mayank99: this code is visible to users, so we do care a bit. but minor

veekeys: ah ok.. one div can be removed, mb

mayank99:
(Closed and reopened to force CI to run after changing branch)

Will be merged and released as `2.11.6` in the next hour or so.

apps/storybook/cypress-visual-screenshots/baseline/Table.test.ts-Full 2.png
mayank99, line 1: visual tests failing

veekeys: no that i have no idea.. might have broken something.>
pushed updated ones, but not sure why it is influenced...

mayank99: Seems like previously it was adding empty div which was occupying space.

Before:
empty div

After:
no empty div

Shouldn't make a practical difference.

PR #1351, opened 2023-06-12
replace `toaster` with new `useToaster` api

mayank99:
## Changes

The `Toaster` class instance been replaced with a `useToaster` hook which returns an object that has properties matching the public functions in the old api.

Only api change needed from user side is:
```diff
- import { toaster } from '@itwin/itwinui-react';
+ import { useToaster } from '@itwin/itwinui-react';

  export default () => {
+   const toaster = useToaster();

    return <button onClick={() => toaster.positive('hello')}>show</button>
  };
```

### How it works

- The `useToaster` hook simply dispatches reducer actions.
- The reducer state is stored in `<ToastProvider>`, which wraps the entire tree inside `<ThemeProvider>`.
- `<ThemeProvider>` renders a `<Toaster>` which reads the state and renders `<Toast>`s accordingly.

~~TODO: Will work on unit tests~~ Fixed.

## Testing

Tested in vite playground and storybook.

Everything seems to work well. 🎉

## Docs

Added changeset.

~~Need to fix website example.~~ Fixed via #1353.


packages/itwinui-react/src/core/Toast/Toaster.tsx
```diff
@@ -38,140 +35,141 @@ export type ToastOptions = Omit<
...
   };
+};
 
-  /**
-   * Set global Toaster settings for toasts order and placement.
-   * Settings will be applied to new toasts on the page.
-   */
-  public setSettings(newSettings: ToasterSettings) {
-    newSettings.placement ??= this.settings.placement;
-    newSettings.order ??= newSettings.placement?.startsWith('bottom')
```
gretanausedaite, line 96: Was this functionality removed on purpose?

packages/itwinui-react/src/core/Toast/Toaster.tsx
```diff
@@ -38,140 +35,141 @@ export type ToastOptions = Omit<
...
-  }
+// ----------------------------------------------------------------------------
 
-  public warning(content: React.ReactNode, options?: ToastOptions) {
-    return this.createToast(content, 'warning', options);
-  }
+export const ToastProvider = ({ children }: { children: React.ReactNode }) => {
+  const [toasterState, dispatch] = React.useReducer(toastReducer, {
+    toasts: [],
+    placement: 'top',
```
gretanausedaite, line 91: Why placement is repeated two times?

packages/itwinui-react/src/core/Toast/Toaster.tsx
```diff
@@ -18,13 +15,13 @@ export type ToasterSettings = {
 
    * When `placement` is set to a top value, order defaults to 'descending', otherwise 'ascending'.
    */
-  order?: 'descending' | 'ascending';
+  order: 'descending' | 'ascending';
```
gretanausedaite, line 18: Both setting props are mandatory now?

gretanausedaite (commented):
Couple questions. Other than that component looks much simpler!

packages/itwinui-react/src/core/Toast/Toaster.tsx
mayank99: removed

packages/itwinui-react/src/core/Toast/Toaster.tsx
mayank99: no, it's using `Partial` in the dispatch action (line 165)

packages/itwinui-react/src/core/Toast/Toaster.tsx
mayank99: hmm, not intentionally. But I'm thinking about it - previously we allowed `order` to be optional, whereas now it gets a default value. We could introduce a third option `auto` and make that the default?

gretanausedaite (approved):
LGTM.

PR #1352, opened 2023-06-12
Component accessibility testing

xman343:
## Changes

Adding a Cypress script that tests all the components in the `examples` folder for accessibility violations using [`cypress-axe`](https://github.com/component-driven/cypress-axe).

## Testing

To run all tests headlessly: `yarn test --filter=a11y` from root.

To run interactively inside browser (locally): `yarn workspace a11y open`.

## Docs

N/A


package-lock.json
mayank99, line 1: i think you ran `npm install` instead of `yarn install`. Delete this file, delete node_modules, then run `yarn install`.

while you're at it, can you also move the playwright stuff out of `packages/itwinui-react` into a new workspace (similar to `examples`)?

xman343: Done

mayank99: also:
- move the other changes from `packages/itwinui-react` into `playwright-testing/`. i'm referring to changes in `package.json` and `.gitignore`
- update branch and resolve conflicts in `yarn.lock`

xman343: done

mayank99: There are still some things left in `packages/itwinui-react`, whereas there should be absolutely no changes there. Take a look at the diff: https://github.com/iTwin/iTwinUI/pull/1352/files

- remove the leftover playwright folders
- move the `test-ct` script

mayank99:
I thought about it some more and here's what I'd suggest:
- Rename the `playwright-testing` workspace to `a11y` and move it into a new folder called `testing`. After that, update the root `package.json` workspaces to use `testing/*`. The idea behind this is that we'll be able to add more types of testing in the future (likely using playwright too).
- Instead of manually listing out all components and examples, lets change the approach to loop over all examples. So you will have a single test file that looks something like:
  ```js
  import { test, expect } from '@playwright/experimental-ct-react';
  import AxeBuilder from '@axe-core/playwright';

  test.describe('Should have no WCAG violations', () => {
    Object.entries(allExamples).forEach(([name, Component]) => {
      test(name, async ({ mount, page }) => {
        await mount(<Component />);
        const results = await new AxeBuilder({ page }).analyze();
        expect(results.violations).toHaveLength(0);
      });
    });
  });
  ```
  The benefit of this is less manual work, both upfront and ongoing. Any future examples will be automatically tested without needing to add new tests.
- Use a [github action workflow](https://playwright.dev/docs/ci-intro) to automatically runs these tests in CI on every push. You can add new job in [`build.yml`](https://github.com/iTwin/iTwinUI/blob/dev/.github/workflows/build.yml).

mayank99:
I looked into this some more and it appears that "component testing" in playwright is [very limited](https://playwright.dev/docs/next/test-components#under-the-hood). Playwright needs to know all the components beforehand, so my above suggestion about looping will not work. Related issue: https://github.com/microsoft/playwright/issues/18057

So your options are:
- either list out everything manually, or
- get rid of all "component testing" configuration, and instead use regular ["end to end testing"](https://playwright.dev/docs/writing-tests). This will require manually wiring up all the boilerplate needed to mount the components, but it should be doable.

I would suggest going with the latter. It can be achieved in three steps:
1. Create a simple isolated "test app", similar to the `playwright/index.html` that currently exists. You can use either [Vite](https://vitejs.dev/) or [Astro](https://astro.build/) for this.
2. Set up the test app such that it renders all the examples at different urls. For example, `localhost:6666/AlertMainExample`, `localhost:6666/AlertInlineExample`, etc.
   - If using astro, [dynamic routes](https://docs.astro.build/en/core-concepts/routing/#dynamic-routes) might come in handy.
   - You'll have to research and experiment yourself. It doesn't matter how you do this - all that matters is that there is a different url for each example.
5. Within playwright, loop over all of those examples again, but instead of mounting the component, use `page.goto` to visit all of those urls and run Axe against them.

mayank99:
Another option you might want to explore is [Cypress component testing](https://docs.cypress.io/guides/component-testing/overview). It is considered stable, unlike Playwright component testing (which is still experimental), so it's possible that you might be able to loop over components in cypress just fine. Definitely worth experimenting.  If you run into similar limitations, then you can carry on with my suggestion above about using playwright e2e testing instead.

Note: You'll have to use [`cypress-axe`](https://github.com/component-driven/cypress-axe) instead of `@axe-core/playwright`.

mayank99:
Here's the code I used to get cypress ct working:

<details>
<summary>cypress.config.ts</summary>

```js
import { defineConfig } from 'cypress';
import react from '@vitejs/plugin-react';
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);
const axeCorePath = require.resolve('axe-core');

export default defineConfig({
  component: {
    devServer: {
      framework: 'react',
      bundler: 'vite',
      viteConfig: {
        plugins: [react()],
      },
    },
    env: { axeCorePath },
    video: false,
    screenshotOnRunFailure: false,
    setupNodeEvents(on, config) {
      on('task', {
        log(message) {
          console.log(message);
          return null;
        },
        table(message) {
          console.table(message);
          return null;
        },
      });
    },
  },
});


```
</details>

<details>
<summary>cypress/support/component.ts</summary>

```ts
import { mount } from 'cypress/react18';

declare global {
  namespace Cypress {
    interface Chainable {
      mount: typeof mount;
    }
  }
}

Cypress.Commands.add('mount', mount);

import 'cypress-axe';
```
</details>

<details>
<summary>cypress/support/component-index.html</summary>

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta
      name="color-scheme"
      content="dark"
    />
    <meta
      name="viewport"
      content="width=device-width,initial-scale=1.0"
    >
    <title>a11y testing</title>
  </head>
  <body data-iui-theme>
    <main>
      <h1>a11y testing</h1>
      <div data-cy-root></div>
    </main>
  </body>
</html>
```
</details>

<details>
<summary>all.cy.tsx</summary>

```tsx

describe('Should have no WCAG violations', () => {
  Object.entries(allExamples).forEach(([name, Component]) => {
    it(name, () => {
      cy.mount(
        <ThemeProvider theme='dark' style={{ height: '100vh' }}>
          <Component />
        </ThemeProvider>,
      );
      cy.injectAxe({
        axeCorePath: Cypress.env('axeCorePath'),
      });
      cy.checkA11y(undefined, undefined, (violations) => {
        const violationData = violations.map(
          ({ id, impact, description, nodes }) => ({
            id,
            impact,
            description,
            nodes: nodes.length,
          }),
        );
        cy.task('table', violationData);
      });
    });
  });
});
```

</details>

<details>
<summary>package.json</summary>

```json
{
  "name": "a11y",
  "version": "0.0.0",
  "private": true,
  "type": "module",
  "scripts": {
    "test": "cypress run --component",
    "open": "cypress open --component",
    "clean": "rimraf .turbo && rimraf node_modules"
  },
  "dependencies": {
    "@itwin/itwinui-react": "3.0.0-dev.3",
    "axe-core": "^4.7.2",
    "cypress": "^12.14.0",
    "cypress-axe": "^1.4.0",
    "examples": "*"
  }
}
```

</details>

xman343:
Added Mayank's code to repo, a11y testing the `examples` now works

testing/a11y/cypress/support/commands.js
mayank99, line 1: this file can be deleted

testing/a11y/cypress/support/component.ts
```diff
@@ -0,0 +1,35 @@
...
+//
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
```
mayank99, line 20:
```suggestion
```

testing/a11y/src/Component.cy.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
```
mayank99, line 1:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
import * as React from 'react';
```

mayank99:
I think we should run these tests in CI as well.

You can add a new job in `build.yml` that looks something like this:
```yml
  a11y:
    name: Test for a11y violations
    needs: install
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3

      - name: Use Node 18.X
        uses: actions/setup-node@v3
        with:
          node-version: 18.x

      - name: Cache node_modules
        uses: actions/cache@v3
        id: cache-node-modules
        with:
          path: |
            node_modules
            packages/*/node_modules
            apps/*/node_modules
            playgrounds/*/node_modules
          key: modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}

      - run: yarn test --filter=a11y
```

testing/a11y/cypress/support/commands.js
xman343: Done

testing/a11y/cypress/support/component.ts
xman343: Done

testing/a11y/src/Component.cy.tsx
xman343: Done

testing/a11y/src/Component.cy.tsx
```diff
@@ -0,0 +1,33 @@
...
+      });
+      cy.checkA11y(undefined, undefined, (violations) => {
+        const violationData = violations.map(
+          ({ id, impact, description, nodes }) => ({
+            id,
+            impact,
+            description,
+            nodes: nodes.length,
+          }),
+        );
```
mayank99, line 28: Lets make the table more concise and useful.

```suggestion
        const violationData = violations.map(({ id, help }) => ({
          Component: name,
          'Rule ID': id,
          Description: help,
        }));
```

xman343:
> I think we should run these tests in CI as well.
> 
> You can add a new job in `build.yml` that looks something like this:
> 
> ```yaml
>   a11y:
>     name: Test for a11y violations
>     needs: install
>     runs-on: ubuntu-latest
> 
>     steps:
>       - uses: actions/checkout@v3
> 
>       - name: Use Node 18.X
>         uses: actions/setup-node@v3
>         with:
>           node-version: 18.x
> 
>       - name: Cache node_modules
>         uses: actions/cache@v3
>         id: cache-node-modules
>         with:
>           path: |
>             node_modules
>             packages/*/node_modules
>             apps/*/node_modules
>             playgrounds/*/node_modules
>           key: modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
> 
>       - run: yarn test --filter=a11y
> ```

Added!

testing/a11y/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "type": "module",
+  "scripts": {
+    "test": "cypress run --component",
+    "open": "cypress open --component",
+    "clean": "rimraf .turbo && rimraf node_modules"
+  },
+  "dependencies": {
+    "@itwin/itwinui-react": "3.0.0-dev.3",
+    "axe-core": "^4.7.2",
+    "cypress": "^12.14.0",
```
mayank99, line 14: Looks like CI is failing. This could potentially be because of multiple versions of cypress being installed (one in storybook).

Could you change all of them to be `12.16.0` and run `yarn`?

Files you need to change:
1. `apps/storybook/package.json`
2. `apps/storybook/scripts/run-tests.js`
3. `testing/a11y/package.json`

xman343: Done

.github/workflows/build.yml
```diff
@@ -376,3 +376,28 @@ jobs:
...
+        id: cache-node-modules
+        with:
+          path: |
+            node_modules
+            packages/*/node_modules
+            apps/*/node_modules
+            playgrounds/*/node_modules
+          key: modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+
+      - run: yarn test --filter=a11y
```
mayank99, line 403: Looks the problem was because of caching.


Try using this:
```suggestion
      - uses: cypress-io/github-action@v5
        with:
          build: yarn build --filter=itwinui-react
          working-directory: testing/a11y
```

xman343: Committed the suggestion

mayank99: looks like that didn't work. Try this:

```
      - yarn build --filter=itwinui-react

      - uses: cypress-io/github-action@v5
        with:
          working-directory: testing/a11y
```

xman343: done

mayank99: sorry, typo. the first line should be prefixed with `run: ` like this:
```yml
      - run: yarn build --filter=itwinui-react
```

xman343: Fixed!

mayank99: Looks like it is correctly getting to that step now, but it is trying to run `e2e` tests instead of `component` tests.


Try adding `component: true` at the end
```
      - uses: cypress-io/github-action@v5
        with:
          working-directory: testing/a11y
          component: true
```

xman343: Added

mayank99: Works correctly now. https://github.com/iTwin/iTwinUI/actions/runs/5412682827/jobs/9837126124

Just missing [one suggestion above](https://github.com/iTwin/iTwinUI/pull/1352#discussion_r1245535046) and then should be good to approve.

testing/a11y/src/Component.cy.tsx
xman343: done!

mayank99 (approved):
LGTM 🚀 working great both locally and in [CI](https://github.com/iTwin/iTwinUI/actions/runs/5412836563/jobs/9837511253).

Now lets start fixing all reported violations in future PRs.

LostABike (commented):
Testing/playing around with this PR. If I understand correctly, we are running a pre-written script axe-core on all of our components? I am able to run the test, and it would tell me there are components that fail accessibility but doesn't specify how. Does it simply test for `aria-label` on all components? @xman343 @mayank99

xman343:
> Testing/playing around with this PR. If I understand correctly, we are running a pre-written script axe-core on all of our components? I am able to run the test, and it would tell me there are components that fail accessibility ~but doesn't specify how. Does it simply test for `aria-label` on all components?~ nevermind I just saw the exact test rules and descriptions! @xman343 @mayank99

Anything I should add, or does the code look good to you?

LostABike:
> > Testing/playing around with this PR. If I understand correctly, we are running a pre-written script axe-core on all of our components? I am able to run the test, and it would tell me there are components that fail accessibility ~but doesn't specify how. Does it simply test for `aria-label` on all components?~ nevermind I just saw the exact test rules and descriptions! @xman343 @mayank99
> 
> Anything I should add, or does the code look good to you?

I'm not too well-versed in configurating modules and scripts so I'm just learning from your code :D But I have read them and I think they look good. Also tested functionality and seems to be working as expected 👍

LostABike (approved):

apps/storybook/cypress-visual-screenshots/baseline/Breadcrumbs.test.ts-Overflow.png
mayank99, line 1: fyi i updated this because visual tests were failing even after retrying 7 times.

PR #1353, opened 2023-06-13
wrap all website examples in ThemeProvider

mayank99:
## Changes

**What I did**: Changed `examples/index.tsx` so that all exports are wrapped in `<ThemeProvider>` using a HOC `withThemeProvider`. Also some examples were using (nonexistent) `args`, so removed those too.

**Why**: In #1351, toasts were not working on website because context was not being found somehow. I figured out it's because context [doesn't work](https://docs.astro.build/en/core-concepts/sharing-state/) in Astro between islands. (Previously, `ThemeProvider` and the examples were rendered in separate islands). Within the same island, context works fine.

## Testing

Tested against #1351 and toasts are now correctly showing up!

https://github.com/iTwin/iTwinUI/assets/9084735/7c051747-caca-4548-8b0a-7b516960d03e

## Docs

N/A

### DIFF
--- a/examples/index.tsx
+++ b/examples/index.tsx
@@ -2,258 +2,1083 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-export { default as AlertMainExample } from './Alert.main';
-export { default as AlertInlineExample } from './Alert.inline';
-export { default as AlertInformationalExample } from './Alert.informational';
-export { default as AlertPositiveExample } from './Alert.positive';
-export { default as AlertNegativeExample } from './Alert.negative';
-export { default as AlertStickyExample } from './Alert.sticky';
-export { default as AlertWarningExample } from './Alert.warning';
+import * as React from 'react';
+import { ThemeProvider } from '@itwin/itwinui-react';
 
-export { default as AvatarMainExample } from './Avatar.main';
-export { default as AvatarGroupExample } from './Avatar.group';
-export { default as AvatarGroupAnimatedExample } from './Avatar.groupanimated';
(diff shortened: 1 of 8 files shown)


LostABike (commented):
I think this search bar is supposed to be longer? (the rest of the bars in the [SearchBox doc](http://localhost:1700/docs/searchbox) is also short)
image

Or did we update all input boxes to have a smaller size at some point? 🤔

mayank99:
@LostABike I've fixed the styles issue. It was because there was an additional wrapper element. It's a bit hacky but works. We'll need to find a more robust solution in the future.

The problem was that #1300 introduced an extra `<div>` in ThemeProvider. If the parent is a flexbox, then this div ends up occupying space. This is making me think it could also cause problems in user apps.

examples/index.tsx
```diff
@@ -2,258 +2,1083 @@
...
-export { default as TooltipPlacementExample } from './Tooltip.placement';
-
-export { default as TransferListMainExample } from './TransferList.main';
-export { default as TransferListWithLabelExample } from './TransferList.withlabel';
-
-export { default as TreeMainExample } from './Tree.main';
-
-export { default as VisuallyHiddenIconExample } from './VisuallyHidden.icon';
-export { default as VisuallyHiddenMoreTextExample } from './VisuallyHidden.moretext';
+import * as React from 'react';
```
gretanausedaite, line 5: That's a lot of lines 👀

gretanausedaite (commented):
Something weird with backdrop and side nav: 
image

examples/index.tsx
mayank99: yeah, but it's a lot fewer lines than adding `<ThemeProvider></ThemeProvider>` manually to every single example usage - especially when you consider that we will start using examples in other places (like #1352)

gretanausedaite (approved):

PR #1354, opened 2023-06-14
Expandable Block composition

### LINKED ISSUE #1292: Expandable Block Component: New props to style header and content container
#### Feature
New properties to handle styles for the Header and Content container. 

#### Examples

Ideally, the idea is to have two separate props for handling styles, one for Header and one for the Content container. 

I want to implement background color, borders, and other styles in each part of the Expandable Block component. 
I'm trying to reach something like that: 
image

And currently, I am just achieving this:
image

LostABike:
## Changes

- Converted `ExpandableBlock` component to be a composition of subcomponents as follows:

```
   <ExpandableBlock>
     <ExpandableBlock.Header>
       <ExpandableBlock.LabelArea>
         <ExpandableBlock/>
         <ExpandableBlock.Caption/>
       </ExpandableBlock.LabelArea>
       <ExpandableBlock.EndIcon/>
     </ExpandableBlock.Header>
     <ExpandableBlock.Content/>
   </ExpandableBlock>
```

- ~~Changed `iui-header` class name to `iui-expandable-header`~~ reverted this cuz bug and to work on other stuff first, will come back.
- Updated css to apply @FlyersPh9 's solution for user's issue where expander icon changes for nested block when it shouldn't be. Verified working!
- Updated Storybook to use composition.
- HTML looks fine.

Closes #1292 

## Testing

- Manually tested all cases in Vite and Storybook, verified working as old component.
- Updated unit tests.

## Docs

- Added jsdocs for all subcomponents
- TBD: ~~Website examples~~ + docs update


packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +179,146 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
+      role='button'
+      aria-expanded={isExpanded ?? !disabled}
+      aria-disabled={disabled}
+      tabIndex={0}
+      onClick={handleToggle}
+      onKeyDown={onKeyDown}
+      ref={forwardedRef}
       {...rest}
     >
+      <SvgChevronRight className='iui-icon' aria-hidden />
```
mayank99, line 192: if this icon is not customizable, then maybe it should be moved into the css. 🤔

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99, line 1: any chance we can change the css class names to be more specific? e.g. `iui-header` -> `iui-expandable-block-header`. `iui-icon` -> `iui-expandable-block-icon`, etc

apps/storybook/src/ExpandableBlock.stories.tsx
```diff
@@ -29,29 +28,38 @@ export default {
 
 export const Basic: Story<Partial<ExpandableBlockProps>> = (args) => {
   return (
-    <ExpandableBlock title='Basic Block' {...args}>
-      {args.children}
+    <ExpandableBlock {...args}>
+      <ExpandableBlock.Header>
+        <ExpandableBlock.LabelArea>
```
mayank99, line 33: this is a lot of code for a basic block. Maybe a convenience `label` prop in `ExpandableBlock.Header` will help (similar to `Tile.Name`)

```jsx
<ExpandableBlock>
	<ExpandableBlock.Header label='Blah' />
	<ExpandableBlock.Content>
	 ...
	</ExpandableBlock.Content>
</ExpandableBlock>
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +182,150 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
+        ref={forwardedRef}
+        {...rest}
       >
-        <SvgChevronRight className='iui-icon' aria-hidden />
-        <Box as='span' className='iui-expandable-block-label'>
-          <Box className='iui-title'>{title}</Box>
-          {caption && <Box className='iui-caption'>{caption}</Box>}
-        </Box>
-        {icon && <Icon fill={status}>{icon}</Icon>}
+        <Box>{children}</Box>
```
gretanausedaite, line 264: Why do you have this additional Box for children?

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -10,28 +10,58 @@ import {
...
-type ExpandableBlockProps = {
-  /**
-   * The main text displayed on the block, regardless of state.
-   */
-  title: React.ReactNode;
-  /**
-   * Small note displayed above title, regardless of state.
-   */
-  caption?: React.ReactNode;
+const ExpandableBlockContext = React.createContext<
```
gretanausedaite, line 18: You can reuse ExpandableBlockOwnProps to not repeat all of the already existing props.

gretanausedaite (commented):
Can you do a11y testing?

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: How do you mean?

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Because of this in the css file:
image

apps/storybook/src/ExpandableBlock.stories.tsx
LostABike: Done. I used the other subcomponent to keep the class structure the same.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Yea I was also thinking about making it customizable. Updated. But then it also made the code too extra since it is part of header (aka if header has children then all the label/title code must also be added). So I decided to make `expandIcon` a convenience prop of `Header`, and if it is passed in then we would just provide the default icon. Otherwise user can override with the `ExpandIcon` subcomponent to use customize the icon they want.

LostABike:
> Can you do a11y testing?

Did you mean manually test NVDA on components with `aria-`?

gretanausedaite:
> Did you mean manually test NVDA on components with `aria-`?

NVDA screen reader and keyboard navigation

packages/itwinui-css/src/expandable-block/expandable-block.scss
FlyersPh9, line 1: This looks good to me, matches what is in https://github.com/iTwin/iTwinUI/pull/1376/files#diff-7139d3558365a97d23934152a27c2e4c8faba3aaeb01912529f15c90f764b7a2.

LostABike:
> > Did you mean manually test NVDA on components with `aria-`?
> 
> NVDA screen reader and keyboard navigation

Tested and verified everything working! One thing though, is that the status `Icon` doesn't have an `aria-label`, and I notice that only `IconButton` have those. Should we add label for `Icon` in this case? Since they convey a status but not necessarily a button.

image

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
gretanausedaite: const ExpandableBlockContext = ExpandableBlockOwnProps & {[props that are missing from main prop type]}

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +183,174 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
+const ExpandableBlockLabelArea = polymorphic.span('iui-expandable-block-label');
+ExpandableBlockLabelArea.displayName = 'ExpandableBlock.LabelArea';
+
+// ----------------------------------------------------------------------------
+// ExpandableBlock.Title component
+
+const ExpandableBlockTitle = React.forwardRef((props, forwardedRef) => {
+  const { className, children, ...rest } = props;
+  return (
+    <Box className={cx('iui-title', className)} ref={forwardedRef} {...rest}>
```
gretanausedaite, line 238: Maybe it's possible to change classNames to be unique? Eg. `iui-expandable-block-title`

playgrounds/vite/src/App.tsx
```diff
@@ -1,10 +1,17 @@
-import { Button } from '@itwin/itwinui-react';
+// import { SvgAdd } from '@itwin/itwinui-icons-react';
+import { ExpandableBlock } from '@itwin/itwinui-react';
 
 const App = () => {
+  // const onToggleMock = jest.fn();
```
gretanausedaite, line 5: You can remove playground code from PR

gretanausedaite:
> Should we add label for `Icon` in this case? Since they convey a status but not necessarily a button.

Icons are decorative elements that have aria-hidden option added.

LostABike:
Changing class names appear to cause visual tests to fail on PR but not locally?

LostABike:
I'm not sure why changing class names are causing the tests to mess up, seem to look fine locally in Storybook

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Updated!

playgrounds/vite/src/App.tsx
LostABike: Oops, reverted :D

LostABike:
Updated everything in response to all comments and more: 
- Changed class names to be more specific
- Tested accessibility
- Updated unit tests and html accordingly
- Shortened props declaration in context
- Make expandable icon into subcomponent
- Updated Website examples to use subcomponents (manually tested and verified working as before)

Side note: I believe the actual documentation text on website does not need to be updated since they describe usage more than implementations of the component?

To be updated:
- Migration guide
- Changesets: currently having trouble with this due to the following error:
image

gretanausedaite (approved):
LGTM

mayank99:
> > > Did you mean manually test NVDA on components with `aria-`?
> > 
> > 
> > NVDA screen reader and keyboard navigation
> 
> Tested and verified everything working!

Can you post a screen recording testing all states (expanded, collapsed, disabled)? See related thread: https://github.com/iTwin/iTwinUI/pull/1321#discussion_r1213522621

---

>     * Changesets: currently having trouble with this due to the following error:
>       image

Are you still facing this issue? Should have been fixed after `main` was merged into `dev` on Monday.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +154,178 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
-        'iui-expandable-block',
-        {
-          'iui-with-caption': !!caption,
-          'iui-expanded': expanded,
-          'iui-small': size === 'small',
-          'iui-borderless': styleType === 'borderless',
-        },
-        className,
+      className={cx('iui-expandable-header', className)}
+      role='button'
```
mayank99, line 158: this might be a good time to use a button instead of a div. See https://github.com/iTwin/iTwinUI/pull/1321#discussion_r1213349992

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +154,178 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
+        ref={forwardedRef}
+        {...rest}
       >
-        <SvgChevronRight className='iui-icon' aria-hidden />
-        <Box as='span' className='iui-expandable-block-label'>
-          <Box className='iui-title'>{title}</Box>
-          {caption && <Box className='iui-caption'>{caption}</Box>}
-        </Box>
-        {icon && <Icon fill={status}>{icon}</Icon>}
+        <Box>{children}</Box>
```
mayank99, line 259: lets add `innerProps` if user wants to add a class to this div

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -120,43 +154,178 @@ export const ExpandableBlock = React.forwardRef((props, ref) => {
...
+      aria-disabled={disabled}
+      tabIndex={0}
+      onClick={handleToggle}
+      onKeyDown={onKeyDown}
+      ref={forwardedRef}
+      {...rest}
+    >
+      {children ?? (
+        <>
+          {expandIcon && <ExpandableBlock.ExpandIcon />}
```
mayank99, line 169: typo?

```suggestion
          {expandIcon ?? <ExpandableBlock.ExpandIcon />}
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: what about `iui-small` and `iui-borderless`? those could be data attributes

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: `??` doesn't work I think because there might not be any value but just a prop pass in like this 
image

You can see that it doesn't show up here https://itwin.github.io/iTwinUI/1354/react/?path=/story/core-expandableblock--basic

mayank99: But why would you want to pass it without a value? it's supposed to be a reactnode, not a boolean.


On that note, I'm unsure if we even need this prop. Is there a case where you're trying to render the header without ExpandIcon or with a custom one?

LostABike: Yea I think we talked about this
image

So I tried to make the icon customizable, but that makes the code too extra so I tried to add a convenience option as well

LostABike: Yea I think we talked about this
image

So I tried to make the icon customizable, but that makes the code too extra so I tried to add a convenience option as well

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: where is `innerProps` passed in from? Do you mean like this? 
```
const ExpandableBlockContent = React.forwardRef((props, forwardedRef) => {
  const { className, children, innerProps, ...rest } = props;
  ```

**Comment by LostABike** [comment]
> > > > Did you mean manually test NVDA on components with `aria-`?
> > > 
> > > 
> > > NVDA screen reader and keyboard navigation
> > 
> > 
> > Tested and verified everything working!
> 
> Can you post a screen recording testing all states (expanded, collapsed, disabled)? See related thread: [#1321 (comment)](https://github.com/iTwin/iTwinUI/pull/1321#discussion_r1213522621)
> 
> > ```
> > * Changesets: currently having trouble with this due to the following error:
> >   image
> > ```
> 
> Are you still facing this issue? Should have been fixed after `main` was merged into `dev` on Monday.

expandable block test

File: packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx (hunk 7 above)
Comment (LostABike) [reply packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx]: I updated this to a button and removed the `role='button'`, assuming it's not needed anymore

File: packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx (hunk 9 above)
Comment (mayank99) [reply packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx:165]: Oh I see. So you can keep the `expandIcon: React.ReactNode` if you want. But _imo_, it is also ok to leave it - it will still be possible to customize it using children. It is not that often (if at all) that you want to change the icon, so extra code is ok. (In my original comment, what I really meant by "customizable" is adding a custom css classname, not changing the icon itself.)

File: packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx (hunk 7 above)
Comment (mayank99) [reply packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx]: yes, this is good. Some additional notes:
1. it should be possible for it to be both expanded and disabled at the same time.
2. tabindex is not needed because `<button>`s are already focusable.
3. onKeyDown is not needed because it is already handled when using `<button>` with `onClick`.
4. user's `onClick` prop needs to be combined with your handler.
5. add `type='button'` to prevent accidental form submissions

```jsx
  const { className, children, label, onClick: onClickProp, ...rest } = props;
  const { isExpanded, setExpanded, disabled, onToggle } = useSafeContext(
    ExpandableBlockContext,
  );

  return (
    <Box
      as='button'
      type='button'
      className={cx('iui-expandable-header', className)}
      aria-expanded={isExpanded}
      aria-disabled={disabled}
      onClick={mergeEventHandlers(onClickProp, () => {
        if (disabled) {
          return;
        }
        setExpanded(!isExpanded);
        onToggle?.(!isExpanded);
      })}
      ref={forwardedRef}
      {...rest}
    >
      {children ?? (
        <>
          <ExpandableBlock.ExpandIcon />
          <ExpandableBlock.LabelArea>
            <ExpandableBlock.Title>{label}</ExpandableBlock.Title>
          </ExpandableBlock.LabelArea>
        </>
      )}
    </Box>
  );
```

after all this, test again in NVDA to confirm behavior is unchanged.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
mayank99: yes, and used here:

```jsx
<Box {...innerProps}>{children}</Box>
```

passed by user like this:

```jsx
<ExpandableBlock.Content innerProps={{ className: 'my-inner' }}>
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Thanks for the suggestion. I added these changes and re-tested behavior + NVDA, working as expected still

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Ok I added innerProps

mayank99: This will be better:
```
innerProps: React.ComponentPropsWithoutRef<'div'>;
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
mayank99: tabindex needs to be removed

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
mayank99: If you're keeping `expandIcon: React.ReactNode`, then all stories and examples need to be changed to remove it, because it should not be passed without a value.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -63,45 +63,79 @@ type ExpandableBlockProps = {
...
     disabled = false,
     ...rest
   } = props;
 
-  const icon = endIcon ?? (status && StatusIconMap[status]());
-
   const [expanded, setExpanded] = React.useState(isExpanded);
   React.useEffect(() => {
     setExpanded(isExpanded);
   }, [isExpanded]);
```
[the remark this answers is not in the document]
mayank99: Unresolving this comment because I think you've reverted part of this change.

The logic here is that `isExpanded` (prop) should always be preferred over the internal state. If the user wants default uncontrolled behavior, then `isExpanded` prop should not be passed.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: but isn't that how it is used? to be passed in so that default icon will display? Or should we change it to be `boolean` instead?

mayank99: no, basic usage should be like this:

```jsx
<ExpandableBlock>
  <ExpandableBlock.Header label='Expandable block' />
  <ExpandableBlock.Content>Content in block!</ExpandableBlock.Content>
</ExpandableBlock>
```

LostABike: I'm still confused, then `expandIcon` is not used at all?

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -63,100 +64,259 @@ type ExpandableBlockProps = {
...
+// ----------------------------------------------------------------------------
+// ExpandableBlock.EndIcon component
+
+const ExpandableBlockEndIcon = React.forwardRef((props, forwardedRef) => {
+  const { className, children, ...rest } = props;
+  const { status } = useSafeContext(ExpandableBlockContext);
+
+  const icon = children ?? (status && StatusIconMap[status]());
+  return (
+    <Icon className={className} fill={status} ref={forwardedRef} {...rest}>
```
mayank99, line 223: if there is no additional class here, then `className` doesn't need to be destructed, it will be used automatically through `...rest` props.

also, `iui-status-icon` can be removed from scss/html code

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: bump

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
mayank99: It should only be used if the icon needs to be changed. That's what I've been trying to say in this whole thread. See above: https://github.com/iTwin/iTwinUI/pull/1354#discussion_r1247110364

Custom usage for a ReactNode would be like this:
```jsx
<ExpandableBlock>
  <ExpandableBlock.Header label='Expandable block' expandIcon={'➡️'} />
  <ExpandableBlock.Content>Content in block!</ExpandableBlock.Content>
</ExpandableBlock>
```

And by default, it should use the caret icon (this is achieved using `??`). It doesn't make sense to ask users to always pass `expandIcon` for the default caret icon; we should include it by default because that's what is desired in 99% of cases.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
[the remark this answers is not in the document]
LostABike: hmm not sure I understand correctly. do you mean that the user controls the states themselves via `isExpanded` prop and their own `onClick`? 
the reason I reverted this was because the expanding and collapsing logic relies on the variable controlled by `useState` (which is `expandedState` in the suggested update). Leaving them as `expanded` won't change inner logic when clicking on header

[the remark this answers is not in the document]
mayank99: > the user controls the states themselves via `isExpanded` prop and their own `onClick`?

Yes but only if they want to opt into controlled mode. By default it will work correctly as uncontrolled.

> Leaving them as expanded **won't** change inner logic when clicking on header

It **will** change if `isExpanded` prop is not passed.


You can test this snippet in vite playground (note there is no `isExpanded` prop passed):

```jsx
<ExpandableBlock>
  <ExpandableBlock.Header label='Expandable block' />
  <ExpandableBlock.Content>Content in block!</ExpandableBlock.Content>
</ExpandableBlock>
```

---

I think in storybook you are explicitly setting value of `isExpanded` in args. That needs to be removed if you want correct uncontrolled behavior.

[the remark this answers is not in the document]
LostABike: Ok yea those controls made me think the logic was off. I reverted this change back. Also, removed `isExpanded` toggle logic for Accordion Story since that seems to not be needed anymore?

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: removed

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: thanks! updated

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: ok yea that makes sense. I have updated all stories and website example and tested working as expected

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: thanks, updated and removed

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -63,100 +64,259 @@ type ExpandableBlockProps = {
...
+        children,
+      }}
+    >
+      <Box
+        className={cx(
+          'iui-expandable-block',
+          {
+            'iui-expanded': expanded,
+            'iui-small': size === 'small',
+            'iui-borderless': styleType === 'borderless',
```
mayank99, line 103: `iui-expanded`, `iui-small` and `iui-borderless` could be data attributes

https://github.com/iTwin/iTwinUI/pull/1354/files/924a3064d78e60073950372b6138023367570f5d..001d046d306da6846b67f7765cf7573715ce348f#r1246683447

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: updated! changeset as well

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: these class names have been updated :d

mayank99: i was thinking converting them to data attributes would be better than class names, like so:

```html
<div class='iui-expandable-block' data-iui-size='small' data-iui-expanded='false' data-iui-disabled='true'>
```

you can see an example in `button.scss`+`Button.tsx`

LostABike: ohhh o_o

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: I am having trouble passing the `should trigger onToggle when clicked with Enter or Spacebar` unit test now because we eliminated the keyDown handler. I tried to add it back in this code but `mergeEventHandlers` would only accept MouseEvent argument and not KeyBoard, what do? It's worth noting that I have manually tested keyboard events (space, enter) on the header with a custom `onToggle` function and can verify that `onToggle` does get triggered, so I'm not sure why this test doesn't pass :-s

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: Done! had some trouble with fixing tests (that also came from other changes not just this in particular), but @adamhock helped me :D

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -135,7 +135,7 @@ const ExpandableBlockHeader = React.forwardRef((props, forwardedRef) => {
       as='button'
       type='button'
       className={cx('iui-expandable-header', className)}
-      aria-expanded={isExpanded}
+      aria-expanded={isExpanded ?? !disabled}
```
mayank99, line 138: why?

```suggestion
      aria-expanded={isExpanded}
```

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -63,100 +64,254 @@ type ExpandableBlockProps = {
...
+type ExpandableBlockContentOwnProps = {
+  innerProps?: React.ComponentPropsWithoutRef<'div'>;
+};
+
+const ExpandableBlockContent = React.forwardRef((props, forwardedRef) => {
+  const { className, children, innerProps, ...rest } = props;
+  const { isExpanded } = useSafeContext(ExpandableBlockContext);
+
+  return (
+    <WithCSSTransition in={isExpanded}>
```
mayank99, line 239: Tangential to API changes: looks like animation is not working, because the `WithCSSTransition` component doesn't take #1302 into consideration. Might be time to use [pure CSS transitions](https://css-tricks.com/css-grid-can-do-auto-height-transitions/).

It can be done in a separate PR but you might want to start working on that PR *now* in case it requires any API/dom structure changes. You can create a chained PR by creating a branch on top of `expandable-block-composition` and using it as the base branch for your PR.

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: I was trying to pass this unit test 

```
it('should respect disabled prop', () => {
  const onToggleMock = jest.fn();
  const { getByRole } = render(
    <ExpandableBlock onToggle={onToggleMock} disabled>
      <ExpandableBlock.Header label='test title' />
      <ExpandableBlock.Content>test content</ExpandableBlock.Content>
    </ExpandableBlock>,
  );
  const header = getByRole('button');
  expect(header).toHaveAttribute('aria-disabled', 'true');
  fireEvent.click(header);
  expect(onToggleMock).not.toHaveBeenCalled();
  expect(header).toHaveAttribute('aria-expanded', 'false');
});

```

Since passing `disabled` only adds `aria-disabled` attribute and not `aria-expanded`, I try to explicitly add it this way when `isExpanded` is not present

mayank99: discussed offline and fixed in [`65ba13b`](https://github.com/iTwin/iTwinUI/pull/1354/commits/65ba13b0dc9d475ec2a034b264f83186e7aa0362)

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -166,15 +166,16 @@ ExpandableBlockHeader.displayName = 'ExpandableBlock.Header';
 const ExpandableBlockExpandIcon = React.forwardRef((props, forwardedRef) => {
   const { className, children, ...rest } = props;
   return (
-    <Box
+    <Icon
       className={cx('iui-expandable-block-icon', className)}
+      size='medium'
```
mayank99, line 171: i think this size prop can be removed

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -125,8 +125,6 @@
 
 /// @arg $type Can be one of: expander, status
 @mixin iui-expandable-block-header-icons($type) {
-  @include mixins.iui-icon-style('m');
```
mayank99, line 128: below, instead of setting `inline-size`, `block-size` and `fill`, you can set these css variables.

```scss
--iui-svg-size: var(--_iui-expandable-block-expander-icon-size);
--iui-svg-fill: var(--_iui-expandable-block-expander-icon-fill);
```

packages/itwinui-css/backstop/tests/expandable-block.html
```diff
@@ -341,9 +372,9 @@ <h2>Borderless</h2>
     </section>
 
     <script>
-      document.querySelectorAll('.iui-header:not([aria-disabled="true"])').forEach((el) => {
+      document.querySelectorAll('.iui-expandable-header:not([aria-disabled="true"])').forEach((el) => {
         el.addEventListener('click', (event) => {
-          event.currentTarget.parentElement.classList.toggle('iui-expanded');
+          event.currentTarget.parentElement.classList.toggle('[data-iui-expanded="true"]');
```
mayank99, line 377: this code doesnt work. you need to set the attribute, not class.

something like this might work (untested):
```suggestion
        const prevExpanded = event.currentTarget.parentElement.dataset.iuiExpanded;
        event.currentTarget.parentElement.dataset.iuiExpanded = !prevExpanded;
```

another option is to use `setAttribute`

packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx
LostABike: removed

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: thanks! this works great for adjusting the icon size based on attribute. I was going to go `:where` route..

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -148,77 +133,65 @@
...
   transition: color var(--iui-duration-1) ease;
   @media (forced-colors: active) {
     transition: none;
   }
+}
 
-  @if $type == 'sublabel' {
-    font-size: var(--iui-font-size-0);
-    color: var(--_iui-expandable-block-header-caption-color, var(--iui-color-text-muted));
+.iui-expandable-block-caption {
```
mayank99, line 146: some of these styles are the same as above and can be combined, e.g.:

```scss
:is(.iui-expandable-block-title, .iui-expandable-block-caption) {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  transition: color var(--iui-duration-1) ease;
    transition: none;
  }
}

.iui-expandable-block-caption {
  font-size: var(--iui-font-size-0);
  color: var(--_iui-expandable-block-header-caption-color, var(--iui-color-text-muted));
}
```

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Closed).png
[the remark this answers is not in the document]
LostABike: manually adding `line-height: inherit` to header does bring header height back to be the same as dev's but content is still shorter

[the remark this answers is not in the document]
LostABike: so I updated `font-family` to `font` and it does help keeping the header same height, not content.

according to the computed layout it tells me that `line-height` is where the discrepancy is:
dev:
image
local:
image

everything else remains consistent.
so I added `line-height` to content and that seems to adjust the height to be same as on dev o_O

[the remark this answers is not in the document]
LostABike: that is odd. without that `line-height` content is shorter than `dev` for me, but let me try that again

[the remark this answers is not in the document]
mayank99: this is the url for `dev` (v3): https://itwin.github.io/iTwinUI/dev/react/?path=/story/core-expandableblock--basic

this is `main` (v2): https://itwin.github.io/iTwinUI/dev/react/?path=/story/core-expandableblock--basic

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: still waiting for this

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Closed).png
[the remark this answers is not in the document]
LostABike: wow. thanks for those! I've been just trying to re-run the tests and hope something click

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: updated!

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Closed).png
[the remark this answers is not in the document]
LostABike: ya new block's content is slightly shorter than current test images. approve?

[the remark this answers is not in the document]
mayank99: weird that it's only in css, but yeah approve.

before approving, can you make sure all headers in the html file use `<button>`? i'm seeing some of them are still using `<div>`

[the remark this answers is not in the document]
LostABike: updated. and css tests are still passing at least locally. lets see

mayank99 (approved):
LGTM

PR #1359, opened 2023-06-15
use only data attribute for variables

mayank99:
## Changes

Removed `iui-root` class from variables because the class will change between major versions, whereas the data attribute won't.

This change will help iTwinUI-variables v3 to be compatible with components from iTwinUI-react v2, v3, v4.

## Testing

Visited portal, html pages, storybook, docs site. Everything looks the same as before. All existing tests pass.

## Docs

added changeset.

### DIFF
--- a/apps/portal/src/pages/index.astro
+++ b/apps/portal/src/pages/index.astro
@@ -32,7 +32,7 @@ const links = [
     </style>
   </head>
 
-  <body data-iui-theme>
+  <body class='iui-root' data-iui-theme>
     <h1 class='iui-text-title'>
       <span class='iui-visually-hidden'>iTwinUI</span>{titleSuffix}
     </h1>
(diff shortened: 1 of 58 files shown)


apps/portal/src/pages/index.astro
```diff
@@ -32,7 +32,7 @@ const links = [
     </style>
   </head>
 
-  <body class='iui-root'>
+  <body class='iui-root' data-iui-theme>
```
LostABike, line 35: So can `class='iui-root'` here and in the html pages be removed if it is not needed anymore?

mayank99: it is still needed for the global css. just not for variables.

gretanausedaite (approved):

LostABike (approved):

PR #1363, opened 2023-06-15
allow setting icon size and fill from outside

mayank99:
## Changes

Restructured the icon css slightly, adding new public-facing `--iui-svg-size` and `--iui-svg-fill` vars. These new vars will be preferred over the internal versions (prefixed with _), allowing easier modifications. The benefit of this would be that we wouldn't need to duplicate icon styling in a bunch of components, as we could simply add `iui-svg-icon` in the html and then specify these css vars inside the component css.

## Testing

All* existing usage is unchanged. Setting size and fill from outside is reflected correctly.

*had to change expandable-block.html to add missing class.

## Docs

added changeset.


LostABike (approved):
Everything icons seem to be working correctly/as before 👍

gretanausedaite (approved):

PR #1365, opened 2023-06-16
Added Algolia Docsearch to docs website

rohan-kadkol:
## Changes

Added [Algolia Docsearch](https://docsearch.algolia.com/) to the `/docs` pages of the documentation website.

The hosted website was indexed automatically by the [crawler](https://www.algolia.com/doc/tools/crawler/getting-started/overview/). I did not change the default parameters set in Algolia. But with time, I hope to fine-tune those parameters for better search results.

<details>
<summary>Old screenshot</summary>
[Video31.webm](https://github.com/iTwin/iTwinUI/assets/45748283/2c5ae2a1-3aa0-4e13-a611-7b18424b5eb8)
</details>

New screenshot:

[Video33.webm](https://github.com/iTwin/iTwinUI/assets/45748283/3d509258-5f7f-4626-a98b-e40a015f196e)

## Testing

I tested it with a few simple queries. Later on, I hope to try more complex search queries and also try to see what search queries users are searching for to improve the results for those types of queries.

## Docs

Added an Algolia search button to `Header.astro`


mayank99:
in general, it looks good apart. few minor comments:
- i would like to make sure the mobile layout looks good.
- instead of linking the algolia js/css from a cdn, we can self-host it
- can we use a custom search button that opens the algolia dialog? the algolia search button doesn't fit the look&feel.
- can we make the colors inside the dialog more accessible?

rohan-kadkol:
> * i would like to make sure the mobile layout looks good.

Here is how it looks on smaller screen widths:

[Video32.webm](https://github.com/iTwin/iTwinUI/assets/45748283/de3f2b17-e054-4804-ac7b-0bbaf67cb6e8)

> * instead of linking the algolia js/css from a cdn, we can self-host it

I changed it to use yarn packages instead of the CDN url

> * can we use a custom search button that opens the algolia dialog? the algolia search button doesn't fit the look&feel.

I was thinking the default button might seem more intuitive for some users since they might used the default button on other doc websites that use Algolia DocSearch too. But I can change it if you feel so.

> * can we make the colors inside the dialog more accessible?

I changed the default Algolia colors to use the colors in `_global.astro`. (https://github.com/iTwin/iTwinUI/pull/1365/files#diff-f58c340d3e32a221844059fe1b76fc412a6668e448c2b16b3b93e93e228907f1R219-R225). Also, is there any alternative to using `!important`? If so, I can replace the `!important` with that.

apps/website/src/pages/_global.astro
```diff
@@ -10,6 +10,8 @@ import '@fontsource/noto-sans/600.css';
 import '@fontsource/noto-sans/300.css';
 import '@fontsource/noto-sans/800.css';
 
+import '@docsearch/css';
```
mayank99, line 13: if you import this inside a style tag, you will be able to use cascade layers like this:

```css
@import '@docsearch/css' layer algolia;
```

this will avoid the need for `!important` when overriding variables

apps/website/.gitignore
```diff
@@ -1,3 +1,6 @@
 # build output
 dist/
 .output/
+
+# Environment variables
+.env
```
mayank99, line 6: maybe should be moved to the root gitinore? we always want to ignore `.env`

apps/website/src/components/Header.astro
```diff
@@ -86,10 +88,6 @@ const { pathname } = Astro.url;
     gap: var(--space-2);
     align-items: center;
     margin-left: auto;
-
-    @media (max-width: 50em) {
-      display: none;
```
mayank99, line 91: why was this removed?

apps/website/src/pages/_global.astro
```diff
@@ -250,5 +260,16 @@ const { content = {} } = Astro.props;
         element.setAttribute('tabindex', '0');
       });
     </script> -->
+    <script>
```
mayank99, line 263: would it make sense to move all the algolia stuff (html, css, js) into a separate astro component?

mayank99 (commented):
i'm ok with leaving the search button styling unchanged in the initial PR but i *would* like to change it at some point.

also, could you target `dev` branch instead of `main`?

apps/website/src/components/Header.astro
rohan-kadkol: This was because on small screen widths, we still want to show the search icon. But with this style, the nav bar does not appear.

image

That's why I removed this style to get:

image

apps/website/src/pages/_global.astro
rohan-kadkol: Ohh, right, made a separate component called AlgoliaSearch. [`3f5a944` (#1365)](https://github.com/iTwin/iTwinUI/pull/1365/commits/3f5a944a61a28319b1f319510c649c2860e4d67b)

apps/website/.gitignore
rohan-kadkol: Sounds good, added this to the root .gitignore [`a16f453` (#1365)](https://github.com/iTwin/iTwinUI/pull/1365/commits/a16f453437d50ee185ff0bef1b0a1ff26c373caf)

rohan-kadkol:
> i'm ok with leaving the search button styling unchanged in the initial PR but i _would_ like to change it at some point.

Great, then I will focus on some other PRs and come back to the button styling a little later.

> also, could you target `dev` branch instead of `main`?

Sounds good, changed the target branch to `dev`

mayank99:
One more thing i noticed the search results always go to the prod url. Can we make the urls use the current context? e.g. localhost vs dev url vs prod url

mayank99:
Cross posting Greta's comment from https://github.com/iTwin/iTwinUI/pull/1414#pullrequestreview-1534753758:

> I have my OS in light mode so Algolia search button is light (probably does not belong in this PR but adding it anyways) 
Screenshot 2023-07-18 at 12 05 19

xman343 (commented):
Looks good. 

Seconding Mayank, it would be better if the search results directed you to the localhost page instead of to product.

apps/website/src/pages/_global.astro
rohan-kadkol: Thanks, I am now using `body` and I don't need `!important` anymore. The Algolia styles still appear outside the cascade layer, but the `<style>` that contains `body` (instead of `:root`) appears higher in the layer specificity list. Just wondering, why could that be the case?

Here is the screenshot where the first layer is the Algolia class, the second layer with `body` are my classes, the third layer with `:root` is the duplicate layer, and the fourth layer are the algolia styles inside the cascade layer.

image
image

rohan-kadkol:
> One more thing i noticed the search results always go to the prod url. Can we make the urls use the current context? e.g. localhost vs dev url vs prod url

I added the `transformItems` prop to remove the base url. (https://github.com/iTwin/iTwinUI/pull/1365/commits/3236957e9355b9a2af749766383d251bab3e0a23)

rohan-kadkol:
> Cross posting Greta's comment from [#1414 (review)](https://github.com/iTwin/iTwinUI/pull/1414#pullrequestreview-1534753758):
> 
> > I have my OS in light mode so Algolia search button is light (probably does not belong in this PR but adding it anyways)
> > Screenshot 2023-07-18 at 12 05 19

I fixed the styling issues so that it displays correctly when `prefers-color-scheme` is `light` and also when it is `dark`. (https://github.com/iTwin/iTwinUI/pull/1365/commits/7c9e00d4fcf30db81dda5b4ef93f71eb51fb6d6c)

apps/website/src/pages/_global.astro
mayank99: are you asking why `body` variables take preference over `:root`? that's because they are lower in the dom tree, so they override anything above it, regardless of specificity.

the layer issue is also just a dev-only thing. i'd suggest inspecting prod build to confirm everything is working
```
yarn build --filter=website
```
```
cd apps/website && yarn preview
```

rohan-kadkol: > are you asking why body variables take preference over :root? that's because they are lower in the dom tree, so they override anything above it, regardless of specificity.

`body` has a specificity of `(0, 0, 1)` and `:root` has a specificity of `(0, 1, 0)`. So I first was a bit confused regarding why `body` took preference over `:root`.

Upon looking in the inpect panel, I saw that the layer with `body` is higher list of layers in the inspect panel. But I am not sure why did the `body` layer appear higher than the `:root` layer and thus override the styles in the layer with `:root`.

> the layer issue is also just a dev-only thing. i'd suggest inspecting prod build to confirm everything is working

Sounds good, will inspect prod version.

PR #1369, opened 2023-06-20
feat(Breadcrumbs): Removed legacy support for button and link

### LINKED ISSUE #1242: v3: Remove legacy support for button / link in breadcrumbs

adamhock:
## Changes

- Removed `iui-breadcrumbs-item-overrides` mixin
- Removed HTML demos for overrides and Generic anchor
- Added `Breadcrumbs.Item` subcomponent which can be used to customize children for `Breadcrumbs`

Closes #1242 

## Testing

- Updated CSS images
- Added unit testing to check for `Breadcrumbs.Item` rendering with different types (`a`, `button`, and `span`)

## Docs

- Changesets added
- Website and Storybook examples updated


.changeset/soft-crabs-flow.md
```diff
@@ -0,0 +1,19 @@
+---
+'@itwin/itwinui-react': major
+---
+
+Added new `Breadcrumbs.Item` subcomponent to use for custom `a` and `button` elements within the `Breadcrumbs`
+
+```jsx
```
gretanausedaite, line 7: Remove example, simply list components that are available. We have documentation site for examples.

gretanausedaite (approved):
Looks good!

adamhock:
image

I keep getting this storybook image mismatch where the Link text is shifted up slightly, but it seems to be centered with the other items, so I'll push this image change

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -210,4 +204,57 @@ const Separator = ({ separator }: Pick<BreadcrumbsProps, 'separator'>) => (
...
+    itemType = 'a';
+  } else if (!!props.onClick) {
+    return (
+      <Box
+        as={'button' as 'a'}
+        className={cx('iui-breadcrumbs-button', className)}
+        ref={forwardRef}
+        {...rest}
+      >
+        <Box as='span' className={cx('iui-breadcrumbs-text', className)}>
```
mayank99, line 224: any chance we can get rid of this extra span? 

it should be possible to use the same single class for both anchors and buttons, because we already do this in other places (such as iui-anchor and iui-alert-link and iui-toast-anchor).

adamhock: Each of the examples you've listed (iui-anchor, iui-alert-link, and iui-toast-anchor) do not apply different styling for button vs anchor. Should iui-breadcrumbs-action have the same styling for both button and anchor elements?

mayank99: it is very similar though.


for the parts that are different, you can apply specific styles targeting `a` or `button`. e.g.:
```scss
.iui-breadcrumbs-action {
  // ...shared styles

  &:where(a) {
    // ...anchor-specific
  }

  &:where(button) {
    // ...button-specific
  }
}
```

both anchor and toast do this to a lesser extent.

https://github.com/iTwin/iTwinUI/blob/92f81adaf94d6874b710de026606cddc779a443e/packages/itwinui-css/src/anchor/mixins.scss#L5-L6

https://github.com/iTwin/iTwinUI/blob/92f81adaf94d6874b710de026606cddc779a443e/packages/itwinui-css/src/toast/toast.scss#L153-L154

adamhock: This has been updated. I had to use `display: inline-block` for the `<button class='iui-breadcrumbs-action'>Previous</button>` element since there is no way to set `text-overflow` on a `button` element that has `display: flex` (to my knowledge). As a result, the icon for the overflow button has been shifted up slightly

image

My suggestion on how to fix this is adding an extra class name or data attribute to `iui-breadcrumbs-action` called `iui-empty` which is true if the button doesn't have any children elements besides plain text. Is there a better solution for this?

packages/itwinui-css/src/breadcrumbs/mixins.scss
```diff
@@ -33,75 +33,39 @@
...
+  padding: 0 var(--iui-size-xs);
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  line-height: var(--iui-component-height);
 }
 
-@mixin iui-breadcrumbs-text {
+@mixin iui-breadcrumbs-action {
   padding: 0 var(--iui-size-xs);
```
adamhock, line 45: Since these attributes shared between both `a` and `button` should be used on any element regardless of type, should we use `iui-breadcrumbs-action` for the `<span>` examples and just get rid of `iui-breadcrumbs-text` entirely? The only issue I see with this is that the name `iui-breadcrumbs-action` would be misleading if the class was applied to a `span`.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: How it normally works in a button is it uses a `<span>`	inside the `<button>` for the text. Maybe we should use the same setup? (We could even reuse `<Button styleType='borderless'>` since it's already set up, but that would get more complicated.)

packages/itwinui-css/src/breadcrumbs/mixins.scss
mayank99: we could change the class name, so that's not an issue. But i feel like `<span>` has different styling - it is not blue. So we shouldn't use same class. But we can reuse styles like this:

```scss
.iui-breacrumbs-action, .iui-breadcrumbs-text {
  padding: 0 var(--iui-size-xs);
```

adamhock: This has been updated. I took a slightly different approach. I used `@include iui-breadcrumbs-text` inside of the `iui-breadcrumbs-action` mixin

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
adamhock: This has been changed. `<span>` is used inside of `<button>` for the button text, but its original class name of `iui-breadcrumbs-text` has been removed

LostABike (approved):
I have read the code and tested functionality. Everything seems to be working as intended 😸

packages/itwinui-css/src/breadcrumbs/mixins.scss
mayank99: wait, so if `iui-breadcrumbs-text` is entirely duplicated inside, then we should be able to combine the two classes, no? It will greatly simplify both the css and the react code.

Try this:

```scss
.iui-breadcrumbs-item-content {
  padding: 0 var(--iui-size-xs);
  overflow: hidden;
  white-space: nowrap;
  text-overflow: ellipsis;
  line-height: var(--iui-component-height);

  &:where(a) {
    // ...
  }

  &:where(button) {
    // ...
  }
}
```

```jsx
  const { children: childrenProp, className, ...rest } = props;

  const defaultAs = !!props.href ? 'a' : !!props.onClick ? 'button' : 'span';
  const children =
    defaultAs === 'button' ? <span>{childrenProp}</span> : childrenProp;

  return (
    <Box
      as={defaultAs as 'a'}
      className={cx('iui-breadcrumbs-item-content', className)}
      ref={forwardedRef}
      {...rest}
    >
      {children}
    </Box>
  );
```

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Looks much better with the shared class between `a` and `button`! I wish we could avoid the extra span, but it's a necessary evil because of the overflow, so it's ok - at least it doesn't have a class. I have left another suggestion in https://github.com/iTwin/iTwinUI/pull/1369#discussion_r1247032801

packages/itwinui-css/src/breadcrumbs/mixins.scss
adamhock: This has been changed

mayank99 (approved):

xman343:
Tested it out myself, it works perfectly!

mayank99:
Some css tests might be failing as a result of #1311; it's weird that the PR still passed. Feel free to update images if necessary.

adamhock:
This branch didn't merge because one of the CSS visual tests failed in `Table` even though my CSS image tests pass locally and in the CI currently. I'm going to try and merge one more time 🤞

PR #1356, opened 2023-06-14
refactor `ProgressRadial` to use pseudo-elements

mayank99:
## Changes

Instead of using an svg, the radial progress indicator will now use pseudo-elements with a combination of conic and linear gradients. Based on this example: https://css-tricks.com/single-element-loaders-the-spinner/#aa-more-loader-examples

- Both the HTML and CSS is greatly simplified.
- The size/status is set using data attributes.
- For the determinate variant, the percentage value will be set using a CSS variable `--iui-progress-percentage`.

## Testing

Html preview: https://itwin.github.io/iTwinUI/1356/css/progress-indicator.html

https://github.com/iTwin/iTwinUI/assets/9084735/56787f7a-9f3c-48f9-9729-1af8507ae562

Storybook preview: https://itwin.github.io/iTwinUI/1356/react/?path=/story/progressindicators-progressradial--indeterminate

https://github.com/iTwin/iTwinUI/assets/9084735/1a6d5692-88ff-4643-b7c3-c4f85e88921f

Unit/visual tests updated.

## Docs

Changeset added.

PR #1328, opened 2023-06-06
feat(Overlay): new React Overlay component

marashad001:
#924 Overlay component

## Changes
- Created new React `Overlay` component from `Progress Indicator` html demo page.

## Testing

- Tests if the component is visible on the page and blurs the appropriate content wrapped within it.

## Docs

Added changeset `sharp-pianos-destroy.mdx` and updated `progress-indicator.mdx`.

PR #1370, opened 2023-06-20
refactor `ProgressLinear` to use pseudo-elements

FlyersPh9:
## Changes

Expands upon #1356 for linear progress indicators.

- HTML and CSS have been greatly simplified to be made of a single div.
- Track fill is set using a linear-gradient on the `::before` pseudo-element.
- Labels require an extra div inside.
- `<Icon>` must now be explicitly used in labels. Fill will be automatically set. 
- All states are set through data attributes.
- ~~All progress indicators use the same class & are changed between radial / linear using `data-iui-variant`.~~

**(description edited by @mayank99)**

## Testing

- Cleaned up html test to exclude `<h3>` & `<p>`.
- Tested working correctly in html pages and storybook.
- Confirmed working in `forced-colors` mode (emulation).
- Updated unit tests.

## Docs

Changesets added.


xman343:
What work needs to be done on the React side of things?

packages/itwinui-css/backstop/tests/checkbox.html
```diff
@@ -571,7 +571,9 @@ <h3>Checkbox on left</h3>
             tabindex="-1"
           />
           <div
-            class="iui-progress-indicator-radial iui-indeterminate"
+            class="iui-progress-indicator"
+            data-iui-variant="radial"
+            data-iui-indeterminate="true"
```
xman343, line 576: I'm liking this new setup for the progress indicator display. Allowing cutsomization through `data` types instead of just a `class` will make it a lot more efficient to switch between different display varieties.

xman343 (commented):
Looking good so far!

packages/itwinui-css/backstop/tests/checkbox.html
FlyersPh9: I do much prefer using data-attributes over class modifiers. ☺️

mayank99:
i've updated the branch (had conflicts from #1328 and #1372), and reorganized the files a little bit. no other code changes.

i plan to make some changes after this, so this comment serves as a sort of a "checkpoint".

mayank99:
i've refactored the code to remove an extra wrapping div, and separated radial and linear into two classes again because there is almost no overlap between the two. also i'm now using manipulating background-image only; this is more performant than animating the width.

LostABike:
in `ErrorPage.tsx` I think ProgressRadial import needs to be updated to `import { ProgressRadial } from '../ProgressIndicators/index.js';`

mayank99:
good catch, updated!

packages/itwinui-css/src/progress-indicator/progress-indicator.scss
```diff
@@ -1,28 +1,200 @@
...
+  &:where([data-iui-animated='true']:not([data-iui-indeterminate='true'])) {
+    &::before {
+      transition: background-size 1s ease-in-out;
+    }
+  }
+
+  &:where([data-iui-indeterminate='true']) {
+    &::before {
+      $_keyframe-name: iui-#{string.unique-id()};
+      animation: $_keyframe-name 1.5s infinite linear;
```
gretanausedaite, line 180: Maybe we could use same cubic-bezier function as in radial?

gretanausedaite (approved):
Code looks good to me.

packages/itwinui-css/src/progress-indicator/progress-indicator.scss
mayank99: maybe. the difference is that the keyframes are already weirdly spaced out (0%, 40%, 100%, with different sizes in each) so i tried to keep it similar to how it felt before

if you want to play with the easing function and keyframes, feel free to suggest something specific

LostABike (approved):
Not sure why I can't add comments on unchanged parts anymore, but this looks good and works good. Just jsdocs need to be updated

LostABike:
> > Just jsdocs should to be updated
> 
> Do you mean this or something else? [`44ef656`](https://github.com/iTwin/iTwinUI/pull/1370/commits/44ef656a89748377a9078291466670a269d1a3b8)

Actually my bad, I misread some stuff. I don't think any jsdocs need update. Everything is good

packages/itwinui-css/src/progress-indicator/progress-indicator.scss
mayank99: @gretanausedaite lmk if you have a specific suggestion for changing this animation or if we can merge this as-is

gretanausedaite: Merge as is.

Issue #1368, opened 2023-06-20
Adding `[x]Props` to every component


gretanausedaite:
The goal is to allow adding a custom className (or data attribute) to every single dom node.
For example labelProps to Button, wrapperProps to Checkbox, etc. 

To do:
- [ ] Go through each component and check if you can customise every part of it. Eg. Checkbox wrapper, checkbox and checkbox label needs to be customizable.
- [ ] Add a new prop to allow that customisation where needed. Name it like [x]Props, eg. WrapperProps in Checkbox.

Juliakas:
Since migrating to iTwinUI 3.x we replaced all our usages of `.iui-` class styling in favor of passing `className`s into these additional props you provided. So far everything is great, we only noticed that we are missing those:

1. Providing extra props to `Tag` component - namely tag text and close button as right now we are using `span` and `button` css selectors to style them - not sure if its a good solution now.
2. Providing extra props to `Modal` component - we can utilize `ModalContent` and `ModalButtonBar` and style them properly - but we still can't style title bar, X close button, modal backdrop and modal wrapper. Modal wrapper especially since sometimes other packages have a higher z-index than Modal component and we have to fight over who gets display priority. Right now we workaround this by having our own modal portal in the root of our application that we can style and use this portal for every single Modal component. So ability to provide more props to Modal would be great - or at least we could have a composable Modal structure like
```jsx
<Modal.Wrapper>
  <Modal.Backdrop/>
  <Modal.Box>
   ...
  </Modal.Box>

</Modal.Wrapper>
```

mayank99:
@Juliakas 
>     1. Providing extra props to `Tag` component - namely tag text and close button as right now we are using `span` and `button` css selectors to style them - not sure if its a good solution now.

Could you share your use case for customizing `Tag`? We can add `textProps` and `closeButtonProps`, but it should also be fine to do this:

```jsx
<Tag className='my-tag' />
```
```css
.my-tag > span {}
.my-tag > button {}
```

>     2. Providing extra props to `Modal` component - we can utilize `ModalContent` and `ModalButtonBar` and style them properly - but we still can't style title bar, X close button, modal backdrop and modal wrapper. Modal wrapper especially since sometimes other packages have a higher z-index than Modal component and we have to fight over who gets display priority. Right now we workaround this by having our own modal portal in the root of our application that we can style and use this portal for every single Modal component. So ability to provide more props to Modal would be great - or at least we could have a composable Modal structure like

[...]

PR #1371, opened 2023-06-20
Adding props to Checkbox component

xman343:
## Changes

Adding the option for customizable wrapper props to be added to the Checkbox component. Issue: https://github.com/iTwin/iTwinUI/issues/1368
## Testing

React unit testing and visual testing undertaken.
## Docs

Docs will be updated to outline wrapperProps utility.


packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -92,18 +109,24 @@ export const Checkbox = React.forwardRef((props, ref) => {
             'iui-checkbox-visibility': variant === 'eyeball',
             'iui-loading': isLoading,
           },
-          className && { [className]: !label },
+          boxClassName && { [boxClassName]: !label },
```
gretanausedaite, line 112: I think you got confused a bit: this is the main Checkbox component - the little square, therefore, it should get "main" className, style, etc. props. 
Undo this change and check if you are able to customise `label`.

xman343: Isn't label already being customized by WrapperProps?   wrapperProps?: React.ComponentProps<'label'>;
Or do you mean the label as in the text?

xman343:
Just remembered I need to update the entry in Docs - I'll do that very soon!

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -104,6 +109,12 @@ export const Checkbox = React.forwardRef((props, ref) => {
     </>
   );
 
+  const {
+    className: wrapperClassName,
+    style: wrapperStyle,
```
adamhock, line 114: Since we're not doing any manipulation to `wrapperStyle` before we add it to the wrapper element, this doesn't need to be extracted. `style` customization will be included when `...restWrapperProps` is added to the wrapper element

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -116,9 +127,10 @@ export const Checkbox = React.forwardRef((props, ref) => {
           [`iui-${status}`]: !!status,
           'iui-loading': isLoading,
         },
-        className,
+        wrapperClassName,
```
adamhock, line 130: Now that the class name for the label comes from `wrapperClassName`, there is a check in the `checkbox` element that only adds `className` to the checkbox when there is no label. This check can be removed and `className` can be added directly to the checkbox element regardless of if there is a label or not

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -116,9 +127,10 @@ export const Checkbox = React.forwardRef((props, ref) => {
...
       )}
-      style={style}
+      style={wrapperStyle}
+      {...restWrapperProps}
     >
       {checkbox}
       {label && (
```
adamhock, line 136: @gretanausedaite Should we do the same thing here for label as we did for the wrapper, or is passing a custom `React.ReactNode` for `label` enough customization?

packages/itwinui-react/src/core/Checkbox/Checkbox.test.tsx
```diff
@@ -187,6 +189,19 @@ it('renders correctly with visibility checkbox', () => {
   expect(container.querySelector('.iui-checkbox-visibility')).toBeTruthy();
 });
 
+// wrapperProps test
```
adamhock, line 192: This comment can be removed

adamhock (commented):
Make sure you add a changeset and link the issue number that this PR is referencing in the PR description

xman343:
> Make sure you add a changeset and link the issue number that this PR is referencing in the PR description

Done

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -86,14 +90,10 @@ export const Checkbox = React.forwardRef((props, ref) => {
...
-            'iui-checkbox-visibility': variant === 'eyeball',
-            'iui-loading': isLoading,
-          },
-          className && { [className]: !label },
-        )}
+        className={cx('iui-checkbox', {
+          'iui-checkbox-visibility': variant === 'eyeball',
+          'iui-loading': isLoading,
+        })}
         style={!label ? style : undefined}
```
gretanausedaite, line 97:
```suggestion
        style={style}
```

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
gretanausedaite: I think it's enough. label className simply sets line-height, everything else is up to the user.
@mayank99 agree?

mayank99: no. every single element needs to be customizable.

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
adamhock: `className` is still needed for users to add a custom class name to `iui-checkbox`. Only the check for `label` needed to be removed

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -36,6 +36,14 @@ type CheckboxProps = {
...
    */
   isLoading?: boolean;
+  /**
+   * Passes properties for checkbox wrapper.
+   */
+  wrapperProps?: React.ComponentProps<'label'>;
+  /**
+   * Passes properties for checkbox label.
+   */
+  labelProps?: React.ComponentProps<'label'>;
```
adamhock, line 46: Since these props are added to a `span` element, make sure this prop's type changes from `<'label'>` to `<'span'>`

adamhock (commented):
Once we get visual and unit tests to pass, I will approve

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
xman343: Done!

xman343:
Image changes were added. Visual React test in CI won't pass without them, but I don't think the image changes originated from my PR. Any advice? 
87e89bc0-1d6e-494c-a7e4-61b87a3aa479

.changeset/empty-rockets-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+Added wrapperProps to allow for styling of CheckBox wrapper.
```
gretanausedaite, line 5: Update changeset with all changes you've made

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -116,13 +130,20 @@ export const Checkbox = React.forwardRef((props, ref) => {
...
+      {...restWrapperProps}
     >
       {checkbox}
       {label && (
-        <Box as='span' className='iui-checkbox-label'>
+        <Box
+          as='span'
+          className={
+            (cx('iui-checkbox-label'),
+            labelClassName ? labelClassName : className)
```
gretanausedaite, line 143: With this code, if user has labeled checkbox and adds `custom-classname`, but not labelClassName/wrapperClassName, html structure would be this:
image

.changeset/empty-rockets-try.md
xman343: Done

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
xman343: I wrote it like that so that the `label` and `wrapper` would share a className with the parent component by default, unless the user goes out of their way to specify custom classNames for `label` and `wrapper`. Should their classNames both be different by default?

gretanausedaite: Why they need to share className?

xman343: They don't have to, I just figured some users might want them to. If they shouldn't share classNames by default, then I'll change the code so they don't

gretanausedaite: If user wants to add styles to certain part of component, they need to use props for that part of component.
Remove this check.

xman343: Removed the checks.

Two unit tests are now failing again. I think it's because they're assuming that setting the `className` variable for the component would still assign it to both the `input` and the `label`, when this isn't the case anymore. I could rewrite the tests so that they make sure the label *isn't* automatically assigned the same `className` as the `input` when only the `input` is given the assignment. 
image
image

mayank99: i think those two tests can be adjusted/removed because the behavior that they were testing has been changed significantly.

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
```diff
@@ -116,13 +130,17 @@ export const Checkbox = React.forwardRef((props, ref) => {
...
       )}
-      style={style}
+      {...restWrapperProps}
     >
       {checkbox}
       {label && (
-        <Box as='span' className='iui-checkbox-label'>
+        <Box
+          as='span'
+          className={(cx('iui-checkbox-label'), labelClassName)}
```
mayank99, line 141: there is a huge (yet subtle) typo in this code

```suggestion
          className={cx('iui-checkbox-label', labelClassName)}
```

this should fix your issue above with tests and you should be able to revert images
> Image changes were added. Visual React test in CI won't pass without them, but I don't think the image changes originated from my PR. Any advice?

this also makes me realize you are missing a unit test for this important functionality.

xman343: Good eye! Fixed the typo. 

> this also makes me realize you are missing a unit test for this important functionality.

What sort of unit test should I ad?

mayank99: same as the one for wrapperProps. test the behavior you are adding.

xman343: Added.

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
xman343: Modified` "Should isomorphically apply class on %s"` to only test `input`, and also removed the final failing test.

mayank99: it is no longer isomorphic so it's not testing anything special and can be removed

imo you should have a single unit test that tests adding separate custom classes all three elements (`className`, `labelProps`, `wrapperProps`)

xman343: Done

mayank99: you're only testing two cases in there, missing a test for `className` which gets applied on the input.

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
mayank99: now update images too

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
xman343: Added

packages/itwinui-react/src/core/Checkbox/Checkbox.tsx
xman343: Done!

packages/itwinui-react/src/core/Checkbox/Checkbox.test.tsx
```diff
@@ -115,6 +114,7 @@ it.each(['label', 'input'] as const)(
     const { container } = render(
       <Checkbox
         label={el === 'label' ? 'Some label' : undefined}
+        wrapperProps={{ style: { color: 'blue' } }}
```
adamhock, line 117: I think instead of having this `it.each` statement, this test can be written similarly to how your test for `className` is written. Just apply different stylings to `wrapperProps`, `labelProps`, and the input, and check for those three stylings within this test.

It might even be possible to test for `style` and `className` within the same test labeled `'correctly passes props through wrapperProps and labelProps`.

Also, I would make sure you apply different colors to each style instead of `blue` across the board.

adamhock (commented):
A few nitpicky comments about the unit tests, but other than that, things look good to me!

packages/itwinui-react/src/core/Checkbox/Checkbox.test.tsx
xman343: Done!

.changeset/empty-rockets-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+Added wrapperProps and labelProps to allow for styling of CheckBox wrapper and label elements.
```
mayank99, line 5: Worth mentioning the breaking change in className/style behavior.

```suggestion
Checkbox: `className` and `style` will now always be applied on the checkbox input element. Added new `wrapperProps` and `labelProps` to allow for styling of wrapper and label elements.
```

mayank99 (approved):

adamhock (approved):
Once the changeset description changes, it's ready to go! 🎉 Great work

xman343:
> Once the changeset description changes, it's ready to go! 🎉 Great work

done!

PR #1377, opened 2023-06-21
Trying to fix ellipses

gretanausedaite:
## Changes

Updated Tags basic test to not have truncation.
Added separate section for truncated tags that is not tested in css.

## Testing

N/A

## Docs

N/A


LostABike (commented):
it still doesn't pass the test. Would this need the test image approved/updated?

LostABike (approved):

packages/itwinui-css/backstop/scenarios/tag.js
```diff
@@ -9,9 +9,6 @@ module.exports = [
   scenario('Type Default', {
     selectors: ['#demo-default'],
   }),
-  scenario('Type Basic', {
```
mayank99, line 12: Instead of removing the basic test, can we change it so that the basic demo does not have truncation? Then we can add a separate truncation demo that we can exclude from tests.

gretanausedaite: Removed truncation from the basic test and added new one with truncation.

mayank99 (approved):
LGTM

PR #1389, opened 2023-06-30
upgrade to TS 5.1 and use proper ESM

mayank99:
## Changes

- Updated to latest TypeScript 5.1. Also updated any surrounding dependencies, e.g. ts-jest, prettier, eslint, react types, etc.
  - This led to some build failures because of `isValidElement`/`cloneElement` types, so added cast to `as JSX.Element` in those places. This change is invisible to users.
- Added [`tslib`](https://github.com/microsoft/tslib) dependency to itwinui-react, and used it with `importHelpers` for more concise CJS output. This package was already included through a transitive dep, so it doesn't add any extra bundle weight.
- `importsNotUsedAsValues` is not supported in TS 5.X so changed to `verbatimModuleSyntax`, but this only works in ESM, so I decided to add `"type": "module"` to `itwinui-react/package.json`.
  - Originally I had planned to do ESM conversion in a separate PR but this gave me an excuse to do it now.
  - Had to add a dummy package.json inside `cjs/` using `postBuild.mjs` script. This is the same approach as https://github.com/iTwin/iTwinUI-icons/pull/49 which was tested thoroughly in multiple environments.
- (Unrelated) Since we have a postBuild script now, I used this opportunity to also add prettier to this script for better debugging experience. Hopefully this prevents users from asking about sourcemaps.

## Testing

Existing build and tests pass!

Inspected build output to verify all intended changes are correctly reflected.

~~Further testing needed in more setups.~~ Tested local build against different bundlers. See https://github.com/iTwin/iTwinUI/pull/1389#issuecomment-1622584272

## Docs

Changesets added.

The typescript change should be non-breaking and invisible, but the `"type": "module"` change could be breaking.


packages/itwinui-react/src/core/Tabs/Tab.tsx
```diff
@@ -80,4 +80,4 @@ export const Tab = React.forwardRef((props, forwardedRef) => {
   );
 }) as PolymorphicForwardRefComponent<'button', TabProps>;
 
-export default TabProps;
+export default Tab;
```
mayank99, line 83: caught this horrible mistake thanks to `verbatimModuleSyntax`

gretanausedaite (commented):
No comments at the moment.

LostABike (approved):
no issues that I could find. also builds and runs on Windows 👍

mayank99:
Tested a local build using `npm pack` to create a simple hello world app in several environments. Working correctly with:
- ✅ Vite 4.3.9 (ESM)
- ✅ Nextjs 13.4.8 App Router / webpack 5 (ESM)
- ✅ Nextjs 13.4.8 Pages Router / webpack 5 (ESM)
- ✅ Astro 2.7.4 (ESM)
- ✅ CRA 5.0.1 / webpack 5 (CJS)
- 🟡 CRA 4.0.3 / webpack 4 (CJS): *Prod build passes but dev server fails because of outdated dependencies, such as postcss and babel, which both fail when they encounter modern syntaxes in our files (unrelated to this PR).* imo this is ok - i expect users to eject out of cra 4 and update old deps.

gretanausedaite (approved):
Code looks good.
Do users needs to know about this change (do we add that to migration wiki)?

mayank99:
> Do users needs to know about this change (do we add that to migration wiki)?

I've included changesets for user-facing parts. I'll mention it in the "general" section of migration guide, because I'm not sure what other migration notes to put. I don't expect anything to break.

PR #1392, opened 2023-07-05
FloatingUI: Menus

gretanausedaite:
## Changes

- Created `PopoverCopy` to work with new popover incrementally. After all functionality is replicated in new api, old popover can be replaced with new one.
- Updated components that use Popover to use PopoverCopy. Changed their code just to make it build, some functionality and edge cases needs to be reviewed.

TODO:
- [ ]  Go through items to make it match functionality-wise (mainly onShow / onHide)
- [ ]  Portal: to: ‘parent’ functionality is missing
- [ ]  On outside click functionality is missing
- [ ]  Aria props / overrides
- [ ] DropdownButton arrow not flipping ???

## Testing

WIP

## Docs

WIP


mayank99:
just a heads up, Dialog should not use floating-ui

packages/itwinui-react/package.json
```diff
@@ -72,7 +72,6 @@
     "dev:styles": "concurrently \"yarn build:styles --watch\" \"yarn copy-styles\""
   },
   "dependencies": {
-    "@floating-ui/react": "^0.24.3",
```
marashad001: Where does the FloatingUI come from if it's not being used in this package?

packages/itwinui-react/src/core/utils/components/PopoverCopy.tsx
```diff
@@ -0,0 +1,171 @@
+/*---------------------------------------------------------------------------------------------
```
marashad001, line 1: How did you decide to make a copy of `Popover.tsx`?

mayank99:
@marashad001 This PR is still in draft, you might want to review something else.

packages/itwinui-react/src/core/utils/components/PopoverCopy.tsx
```diff
@@ -143,6 +220,12 @@ export const PopoverCopy = React.forwardRef((props, ref) => {
     </Box>
   );
 
+  React.useEffect(() => {
+    if (reference) {
+      popover.refs.setReference(reference.current);
```
mayank99, line 225: learned recently that this useEffect can be avoided because there is a simpler way to set `reference` too
https://floating-ui.com/docs/react#external-elements

gretanausedaite:
TODO:
- [ ] Go through components that uses Popover to make it match to previous
- [ ] On outside click feature
- [ ] Props review
- [ ] Aria props / overrides for both, tooltip and popover

rohan-kadkol:
> just a heads up, Dialog should not use floating-ui

Just curious, what is the reason for that?

PR #1396, opened 2023-07-06
Css transition for expandable block

LostABike:
## Changes

- Removed `WithCSSTransition` for `iui-expandable-content`
- Handle content transitioning in css code
- Removed all mixins

## Testing

- Manual testing on storybook
- Ran `yarn test` (approval TBD)

## Docs

- Added changeset


packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -195,6 +195,21 @@
...
+    min-height: 0;
+    visibility: hidden;
+    @include iui-expandable-block-content;
+  }
+
+  &[data-iui-expanded='true'] .iui-expandable-content {
+    grid-template-rows: 1fr;
+    visibility: visible;
     @include iui-expandable-block-content;
   }
```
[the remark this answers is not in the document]
mayank99: I've found a way to keep padding without needing two wrapping divs - it will require the use of `transform: translateY(-100%)` combined with `z-index` to hide the leftover padding under the header only when it's closed.

Also, because this requires changes in a few different elements, it would be best to clean up the CSS to remove the use of all mixins, and use classes directly. The mixins are confusing and unnecessary.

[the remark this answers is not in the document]
LostABike: like..this? it works but not smooth. also removed mixin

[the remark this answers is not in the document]
mayank99: it's not smooth because:
(1) you have removed padding from the closed state.
(2) you are not transitioning `transform`.

this should work:
```scss
    display: grid;
    grid-template-rows: 0fr;
    overflow: hidden;
    z-index: -1;
    transition: grid-template-rows 0.2s, visibility 0.2s, transform 0.2s;

    > div {
      min-height: 0;
      visibility: hidden;
      transition: visibility 0.2s;
      padding: var(--_iui-expandable-block-content-padding);
    }
  }

  &[data-iui-expanded='true'] .iui-expandable-content {
    grid-template-rows: 1fr;

    > div {
      visibility: visible;
    }
  }

  &:not([data-iui-expanded='true']) .iui-expandable-content {
    transform: translateY(-100%);
  }
```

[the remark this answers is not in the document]
mayank99: as for the mixins, i really meant removing *all* mixins, not just `iui-expandable-content`.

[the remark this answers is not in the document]
LostABike: Ok now the content is not visible altogether :-s I tried tweaking around too, setting visible variables and transformY on various divs/classes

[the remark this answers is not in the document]
mayank99: can you try to paste the exact code i wrote above? it works fine for me.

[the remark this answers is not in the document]
LostABike: Yea using exact code caused the other issue of content now showing when expanded for me which is why I tweaked around (it is the one being pushed just now, you can check [here](https://itwin.github.io/iTwinUI/1396/react/?path=/story/core-expandableblock--basic)
expandable block

I think is because z-index is still -1 from default. making it not -1 when `data-iui-expanded='true'` would show the content. But like I mentioned, animation is still not smooth cuz collapsing is instant:
expandable block z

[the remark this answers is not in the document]
mayank99: try adding `isolation: isolate` to `iui-expandable-block` (root)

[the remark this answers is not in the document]
LostABike: sorcery?? thanks! that does it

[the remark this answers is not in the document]
mayank99: the problem was that `z-index: -1` places it underneath the background, so `isolation: isolate` will create a new stacking context which tells it to only place it underneath the elements within this context.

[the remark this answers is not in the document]
LostABike: ohhh that makes sense!

apps/storybook/cypress-visual-screenshots/baseline/ExpandableBlock.test.ts-Accordion (Closed).png
mayank99, line 1: looks like it is occupying some extra space in collapsed state :(

LostABike: ya I think it is because of the content's padding. I moved the padding to when expanded = true only. Also its bttom border was showing so I moved that as well. I think it looks correct now

mayank99: it's causing some pretty unpleasant content shift. try splitting the padding so that `padding-inline` is always applied, and `padding-block` is only applied when open.

LostABike: Updated to your suggested change and test image. Seems ok now?

.changeset/itchy-poems-shout.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': minor
+---
+
+Expandable block animation now uses pure css for transitioning and `WithCSSTransition` has been removed.
```
mayank99, line 6: should be a patch change and should not mention internal names like "WithCSSTransition"

```suggestion
---
'@itwin/itwinui-react': patch
'@itwin/itwinui-css': patch
---

Expandable block animation now uses pure css instead of javascript for transitioning.
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
+  --_iui-expandable-block-header-padding: var(--iui-size-s);
 
   display: flex;
   flex-direction: column;
+  isolation: isolate;
   width: 100%;
   border-radius: var(--iui-border-radius-1);
 
   &[data-iui-expanded='true'] {
     --_iui-expandable-block-expander-icon-transform: rotate(90deg);
```
mayank99, line 35: this can be combined with the data-iui-expanded rule above (line 21)

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
+    background-color: var(--_iui-expandable-block-background-color);
+    border-right: var(--_iui-expandable-block-border);
+    border-left: var(--_iui-expandable-block-border);
+    border-bottom-left-radius: inherit;
+    border-bottom-right-radius: inherit;
+
+    display: grid;
+    grid-template-rows: 0fr;
+    overflow: hidden;
+    z-index: -1;
```
mayank99, line 120: i think lets remove this and instead use `z-index: 1` on the expandable-block-header.

because the user might have some content that uses z-index and `-1` could potentially cause unexpected issues with it.

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
+    transition: grid-template-rows 0.2s, visibility 0.2s, transform 0.2s;
```
mayank99, line 121: lets use `--iui-duration-1` instead of `0.2s` everywhere in this file

also add some easing curve. `transition-timing-function: ease-out;` might be enough

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
+
+    > div {
+      min-height: 0;
+      visibility: hidden;
+      padding-inline: var(--_iui-expandable-block-content-padding);
+      transition: visibility 0.2s ease-out;
+      align-self: start;
```
mayank99, line 128: this might be subjective but it looks slightly better with `end` (could be an illusion though)

```suggestion
      align-self: end;
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
-  }
-
-  &:where(:hover) {
-    --_iui-expandable-block-expander-icon-fill: var(--iui-color-icon-hover);
-    --_iui-expandable-block-header-caption-color: var(--iui-color-text-hover);
-    --_iui-expandable-block-header-label-color: var(--iui-color-text-hover);
-
-    background-color: var(--_iui-expandable-block-header-background-color-hover);
 
+  > .iui-expandable-header {
```
mayank99, line 38: almost all of the classes in this file can be _unnested_ because they have descriptive unique names now

packages/itwinui-css/src/expandable-block/expandable-block.scss
```diff
@@ -2,219 +2,195 @@
...
-  transition: color var(--iui-duration-1) ease;
-  @media (forced-colors: active) {
-    transition: none;
-  }
-
-  @if $type == 'sublabel' {
-    font-size: var(--iui-font-size-0);
-    color: var(--_iui-expandable-block-header-caption-color, var(--iui-color-text-muted));
+    > .iui-expandable-block-icon {
+      @include mixins.iui-icon-style('m');
```
mayank99, line 98: **_unrelated_** to this PR. You should be able to remove this mixin and instead use `<Icon>` in react. This change can be done in #1354

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: also we are missing a check for [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion), so that will need to be added to all transitions

mayank99 (commented):
generally looks good so far! just a few minor comments

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: Removed nesting

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: Like this?

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: Moved to header

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: Right, combined

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: as in wrap the element in `<Icon/>` instead of `<Box/>` and remove the css class?

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: it's not removed though.

```scss
// original
.iui-expandable-block {
  // ...

  > .iui-expandable-header {
    // ...
  }
}
```

```scss
// expected
.iui-expandable-block {
  // ...
}
.iui-expandable-header {
  // ...
}
```

```scss
// what you did (currently)
.iui-expandable-block {
  // ...

  .iui-expandable-header {
    // ...
  }
}
```

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: should be similar to `ExpandableBlockEndIcon`

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: hmm so I moved `iui-expandable-content` and `iui-expandable-header` out (haven't pushed), but for other things like `title` and `caption` under `label`, should we move those out too? I just mainly thought it looks more organized in a hierarchal way as is
image

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: On Windows I can toggle animation (I think we did this before) and I think it works as intended. No transition animation when it is off.

packages/itwinui-css/src/expandable-block/expandable-block.scss
mayank99: > for other things like title and caption under label, should we move those out too?

yes, nesting should be avoided as much as possible because it leads to hard-to-maintain code

LostABike: okay. I un-nested everything and tested functionality to make sure nothing broke :x

mayank99: for these two cases, you should be able to move the code back where it was and do nesting _opposite_ direction.
image

something like this:

```scss
.iui-expandable-content {
  // ...

  :where(.iui-expandable-block[data-iui-expanded='true']) & {
    // ...
  }

  :where(.iui-expandable-block:not([data-iui-expanded='true'])) & {
    // ...
  }
}
```

this ensures that all the code that is applied on `iui-expandable-content` is grouped together

LostABike: oh cool! I was trying to achieve something similar a while back! thanks! updated

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: okay I have updated this (might have forgotten to push yesterday) + resolved the merge conflicts. also re-tested to make sure things are still working as before

packages/itwinui-css/src/color-picker/color-picker.scss
mayank99, line 1: dont think these are intentional, should be reverted.

try running `yarn` so your dependencies are up-to-date, then save this file and push it.

packages/itwinui-css/src/toggle-switch/toggle-switch.scss
mayank99, line 1: these don't look intentional either.

packages/itwinui-css/src/expandable-block/expandable-block.scss
LostABike: hold on css is messed up I'm looking into that

packages/itwinui-css/src/color-picker/color-picker.scss
LostABike: ya weird, I reverted

packages/itwinui-css/src/toggle-switch/toggle-switch.scss
LostABike: reverted

LostABike:
@mayank99 added `flex-shrink: 0` to icon because it was getting squished within header container when title is too long. Also, positioning of icon was weird on html:
image

so I removed `block-size: var(--_iui-expandable-block-expander-icon-size);`, which doesn't seem to do anything to other cases and work to fix small icon positioning

also fixed script within html, wasn't working correctly before.

Issue #1148, opened 2023-03-20
Testing accessibility for all components


mayank99:
All components, especially the more complex ones, need to be tested for accessibility. The most basic form of this involves running automated tools (like axe) which could even be made of part of CI. A more useful form of this will involve testing with (at least) NVDA+Firefox/Chrome on Windows and VoiceOver+Safari on macOS, and also with Talkback+Android and VoiceOver+iOS where possible.

This issue can be used as a checklist to track the status of all components.

#### Automated (using axe)

_(already passing if not mentioned)_

- [x] Avatar #1397
- [x] ComboBox #1456
- [x] Tooltip 
- [x] Button group #1404
- [x] Carousel #1626
- [x] Color picker #1618, #1621
- [x] DatePicker #1486 #1581
- [x] Header #1495
- [x] Side navigation #1496
- [x] Surface #1500
- [x] Table #1505
- [x] Tabs #1548 
- [x] Tile #1493, #1631
- [x] Tree #1492 
- [x] Button
  - [x] #1532
  - [x] #1533
- [x] Input: #1536
- [x] Input group #1482 
- [x] Select #1479
- [x] Slider #1474
- [x] Textarea https://github.com/iTwin/iTwinUI/pull/1540
- [x] TransferList https://github.com/iTwin/iTwinUI/pull/1541

#### Manual accessibility audit

Pending.

#### Automated screen-reader testing

Pending.

See [guidepup](https://www.guidepup.dev/).

maliksenpai:
Currently I'm working on it. I'm doing this tests with this library, is it okay? https://github.com/nickcolley/jest-axe @mayank99

mayank99:
> I'm doing this tests with this library, is it okay? [nickcolley/jest-axe](https://github.com/nickcolley/jest-axe)

jest-axe is okay, but the problem is it runs in jsdom, so it won't be able to catch many of the issues such as color contrast. A better option would be to run aXe inside cypress/playwright or through [storybook](https://storybook.js.org/blog/automate-accessibility-tests-with-storybook#catch-accessibility-regressions-automatically).

And of course that is only the first step. After that, we also need to do manual testing and document the results.

maliksenpai:
So I will do it with jest-axe firstly, after that I will add more test with cypress.

jest-axe does not guarantee that what you build is accessible but still jest-axe is good.

PR #1397, opened 2023-07-06
Bring failing Avatar examples to a11y compliance

xman343:
## Changes

Editing the `Avatar` examples that are violating a11y rules to make them compliant. Component-level changes may be needed.

Addresses issue [1148](https://github.com/iTwin/iTwinUI/issues/1148).

## Testing

Accessibility tested thru Cypress a11y tests. Unit and visual tests run as well.

## Docs

N/A


examples/Avatar.main.tsx
```diff
@@ -36,7 +36,7 @@ export default () => {
       <Avatar
         abbreviation='RR'
         backgroundColor={getUserColor('Robbie Robot')}
-        image={<img src={imgSrc} />}
+        image={<img src={imgSrc} alt='Profile picture of Robbie Robot.' />}
```
mayank99, line 39: empty alt text should suffice because the user name is already exposed
```suggestion
        image={<img src={imgSrc} alt='' />}
```

same comment in other examples

packages/itwinui-react/src/core/Avatar/Avatar.tsx
```diff
@@ -104,6 +104,7 @@ export const Avatar = React.forwardRef((props, ref) => {
       {status && (
         <Box
           as='span'
+          role='status'
```
mayank99, line 107: `role=status` creates a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), which is meant for notifications and other updates, so it is the wrong role in this case.

I think there is no special role needed here, and that we should actually remove the aria-label.

We could add the status text as a separate [visually-hidden](https://itwinui.bentley.com/docs/visuallyhidden) span in a future PR if desired.

testing/a11y/cypress.config.js
```diff
@@ -20,6 +20,7 @@ export default defineConfig({
     },
     env: { axeCorePath },
     video: false,
+    numTestsKeptInMemory: 205,
```
mayank99, line 23: is this change needed?

xman343: It keeps snapshots from every component test; by default, it only keeps fifty of them. Without this change, you would only be able to view the snapshots from the last fifty examples tested.

mayank99: then lets pick a higher number and guard it under a CI check (there is limited memory in CI)


something like this might work (untested)
```suggestion
    numTestsKeptInMemory: process.env.CI ? 500 : undefined,
```

examples/Avatar.main.tsx
xman343: Done!

testing/a11y/cypress.config.js
mayank99: whoops, the condition is flipped. should be:

```
    numTestsKeptInMemory: !process.env.CI ? 500 : undefined,
```

("when not CI, use 500, otherwise leave it undefined")

make sure to test that it works as expected locally

xman343: It worked locally, pushed

packages/itwinui-react/src/core/Avatar/Avatar.tsx
xman343: Done

.changeset/stale-scissors-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
```
mayank99, line 2: Should be a patch change, not major. Please read https://semver.org/

```suggestion
'@itwin/itwinui-react': patch
```

xman343: done!

mayank99: forgot to push?

xman343: pushed now

mayank99 (approved):

LostABike (approved):
LGTM 👍

PR #1400, opened 2023-07-10
use `React.lazy` in ErrorPage and make illustrations optional

mayank99:
## Changes

Instead of all illustrations always getting included in the bundle, they will now be lazy loaded when `ErrorPage` is rendered. Devs will also need to opt into this by manually installing `@itwin/itwinui-illustrations-react`.

This is basically reintroducing the first iteration of https://github.com/iTwin/iTwinUI-react/pull/924 which was failing during Next.js hydration but working in client-side apps. Because we now have `NonIdealState` as a preferred alternative, this can always be used if the `React.lazy` approach doesn't work for any of our users.

Note that I would have liked to remove the `ErrorPage` component as it's deprecated, but did not do so because it has [convenient functionality](https://github.com/iTwin/iTwinUI-react/pull/715#issuecomment-1179071257) that our users find desirable and that's missing from `NonIdealState`. So this PR is the best of both worlds: remove the huge illustrations dependency while having at least one component that's suitable in every environment, at the cost of manually having to install `@itwin/itwinui-illustrations-react`.

### Important notes

I had to add a `dynamicImport` utility function because I encountered very strange issues with typescript and with jest as well (see links in code comments).

Uses changes from https://github.com/iTwin/iTwinUI-icons/pull/70.

## Testing

Tested working correctly in vite, next and astro playgrounds.

## Docs

Changeset added.


packages/itwinui-react/src/core/utils/functions/import.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Wrapper around native `import()` with workarounds for:
+ *   1. typescript trying to transpile dynamic import in CJS environment (https://github.com/microsoft/TypeScript/issues/43329)
+ *   2. jest/node causing segfault when encountering dynamic import (https://github.com/nodejs/node/issues/35889)
+ */
+export const dynamicImport =
```
aruniverse, line 11: its not really dynamic import tho is it because youre not awaiting it right?

mayank99: it is just a wrapper over `import()`, but its main use case is dynamic imports.

the `await`ing happens in the place where it is called.

aruniverse: yea exactly so the name is misleading, just a nit

mayank99: Is it misleading though? For who? 😄 

From MDN:

> The import() syntax, commonly called dynamic import [...]

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

aruniverse: wow; TIL!

aruniverse (approved):

gretanausedaite (approved):

PR #1401, opened 2023-07-10
feat(Tabs): v3 API Composition

### LINKED ISSUE #970: v3: Tabs Composition API
Here we need to rethink and improve the whole API.

- Split into separate components: TabsWrapper, TabPanel.
- Rename `labels` to `tabs` or even accept as children.

adamhock:
### CSS

- The wrapper element for the tab label and description has been removed
- The tab label and tab description have changed from `<div>` elements to `<span>` elements
- The `iui-tab-label` class now applies to the tab label

```diff
- <ul class="iui-tabs">
-   <li>
-     <button class="iui-tab">
-       <span class="iui-tab-label">
-         <div>Active</div>
-         <div class="iui-tab-description">Sublabel</div>
-       </span>
-     </button>
-   </li>
- </ul>

+ <div class="iui-tabs">
+   <button class="iui-tab">
+     <span class="iui-tab-label">Active</span>
+     <span class="iui-tab-description">Sublabel</span>
+   </button>
+ </div>
```

### React
- Tabs composition has been updated such that it is now made up of customizable subcomponents.
- The `actions`, `labels`, `tabsClassName`, `contentClassName`, and `wrapperClassName` props have been removed.
- The exported `Tab` class has been removed

```jsx
  <Tabs>
    <Tabs.TabList>
      <Tabs.Tab value='tab1'>
        <Tabs.TabLabel>Label 1<Tabs.TabLabel>
      </Tabs.Tab>
      <Tabs.Tab value='tab2'>
        <Tabs.TabLabel>Label 2<Tabs.TabLabel>
      </Tabs.Tab>
      <Tabs.Tab value='tab3'>
        <Tabs.TabLabel>Label 3<Tabs.TabLabel>
      </Tabs.Tab>
    </Tabs.TabList>
    <Tabs.ActionsWrapper>
      <Tabs.Actions>
        <Button>
          Button 1    
        </Button>
        <Button>
          Button 2
        </Button>
      </Tabs.Actions>
    </Tabs.ActionsWrapper>
    <Tabs.Panel value='tab1'>
      Panel 1
    </Tabs.Panel>
    <Tabs.Panel value='tab2'>
      Panel 2
    </Tabs.Panel>
    <Tabs.Panel value='tab3'>
      Panel 3
    </Tabs.Panel>
  </Tabs>
```

Closes #970.

## Testing

- Unit tests have been updated

## Docs

- Changeset added
- Website documentation updated


packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -230,6 +146,75 @@ export const Tabs = (props: TabsProps) => {
...
+      currentActiveIndex !== activeIndex &&
+      setCurrentActiveIndex
+    ) {
+      setCurrentActiveIndex(getBoundedValue(activeIndex, 0, items.length - 1));
+    }
+  }, [activeIndex, currentActiveIndex, items.length]);
+
   const [focusedIndex, setFocusedIndex] = React.useState<number | undefined>();
   React.useEffect(() => {
     if (tablistRef.current && focusedIndex !== undefined) {
```
adamhock, line 220: This hack needs to be addressed in this PR. Any suggestions on how to fix this hack would be appreciated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +511,360 @@ export const Tabs = (props: TabsProps) => {
...
+    <Box
+      as='ul'
+      className={cx(
+        'iui-tabs',
+        `iui-${type}`,
+        {
+          'iui-green': color === 'green',
+          'iui-animated': type !== 'default' && isClient,
+          'iui-not-animated': type !== 'default' && !isClient,
+          'iui-large': hasSublabel,
```
adamhock, line 524: Didn't know where to leave the comment, but our method for getting `hasSublabel` is also a hack that needs to be addressed in this PR

apps/storybook/src/Tabs.stories.tsx
```diff
@@ -32,210 +25,285 @@ export default {
...
 DefaultTabs.args = {
   type: 'default',
-  labels: [
-    <Tab key={1} label='Item1' />,
-    <Tab key={2} label='Item2' />,
-    <Tab key={3} label='Item3' />,
-  ],
 };
 
 export const BorderlessTabs: Story<Partial<TabsProps>> = (args) => {
```
adamhock, line 94: Some of the visual tests aren't passing due to the stripe not extending all the way to the edge of the tab on the initial render. Seems to happen with all borderless examples, but most noticeable in these two stories. This needs to be fixed. 

image

image

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +511,360 @@ export const Tabs = (props: TabsProps) => {
...
+}) => {
+  const { currentActiveIndex } = useSafeContext(TabsContext);
+
+  const items = Array.isArray(children) ? children : [children];
+
+  return (
+    <>
+      {items.map((item, index) => {
+        return (
+          <TabsContext.Provider
```
gretanausedaite, line 686: Why do you have two TabsContext.Providers?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +511,360 @@ export const Tabs = (props: TabsProps) => {
...
+      {children}
+    </Box>
+  );
+}) as PolymorphicForwardRefComponent<'button', TabsTabOwnProps>;
+TabsTab.displayName = 'Tabs.Tab';
+
+// ----------------------------------------------------------------------------
+// Tabs.TabIcon component
+
+const TabsTabIcon = polymorphic.span('iui-tab-icon', {
```
gretanausedaite, line 599: You should rething Icon, Info, Label and Description components. The current approach is very verbose and a lot of code needed for simple stuff. Maybe we could achieve something like this:
```
<Tab label='Label 1' />
```
```
<Tab>
  <Tab.Icon>☆</Tab.Icon>
  <Tab.Label>Label 2</Tab.Label>
  <Tab.Description>Description</Tab.Description>
</Tab>
```

apps/storybook/src/Tabs.stories.tsx
gretanausedaite: Agree, this needs to be fixed.

adamhock: This has been fixed by adding a timeout of 50ms so that the animation can finish before the stripe is applied. Let me know if there's a better solution for this

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +511,360 @@ export const Tabs = (props: TabsProps) => {
...
+// ----------------------------------------------------------------------------
+// Tabs.TabDescription component
+
+const TabsTabDescription = polymorphic('iui-tab-description');
+TabsTabDescription.displayName = 'Tabs.TabDescription';
+
+// ----------------------------------------------------------------------------
+// Tabs.Actions component
+
+type TabsActionsOwnProps = {}; // eslint-disable-line @typescript-eslint/ban-types
```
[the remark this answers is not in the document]
adamhock: Removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: The context provider for the wrapper is the main one that passes different variables to `Tabs.TabList`.

The context provider for the tabs list passes an updated `currentActiveIndex`, an `onClick` function that was created in `Tabs.TabList`, a `focusedIndex` variable created in `Tabs.TabList`, and an `index` variable to each of the `Tabs.Tab` children so that each child can be indexed properly. The same thing happens when passing `index` from `Tabs.Panels` to `Tabs.Panel`.

Is there any issues I'm not aware of that occur when nesting context providers this way?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: I agree that creating basic tabs with no description and no icon should use the first example, but I think `Tab.Label` and `Tab.Description` need to be wrapped within a `div` element with the class name `iui-tab-label` so that the label and the description both get the `display: block` property.

mayank99:
@adamhock You need to remove all the `querySelector` hacks as described in https://github.com/iTwin/iTwinUI/issues/970#issuecomment-1572525585. You can do this by moving some of the code from `Tabs` into `Tab` and also through the use of context. The purpose of using `querySelector` is to get a reference to the element, but you can already get a reference using `ref`, so `querySelector` is not needed. And you can call `setHasSublabel` from inside `Tabs.Description` since it is its own component now.

Also the setTimeout hack you added in [`e2d7f38`](https://github.com/iTwin/iTwinUI/pull/1401/commits/e2d7f382ed71b168428428e03a39c7aaf473b704) should not be needed because https://github.com/iTwin/iTwinUI/pull/1112 already had a fix for stripe width on initial render.

apps/storybook/src/Tabs.stories.tsx
adamhock: As @mayank99 mentioned in [this comment](https://github.com/iTwin/iTwinUI/pull/1401#issuecomment-1640609214), setting a timeout is not the proper solution here.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: This hack has been addressed by using `tablistRef` and `focusedIndex` to locate the correct tab to focus.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: This hack has been updated. Adding a `Tab.Description` subcomponent changes the value of `hasSublabel` if the tab type is not a pill.

apps/storybook/src/Tabs.stories.tsx
adamhock: This issue has now been fixed by setting the stripe properties inside of `Tabs.TabList` and using `setStripeProperties` to set the properties in `Tabs`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: I've added a `label` prop to the `Tabs.Tab` subcomponent so simple tabs can be less verbose.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,399 @@ export const Tabs = (props: TabsProps) => {
...
+    return <></>;
+  }
+}) as PolymorphicForwardRefComponent<'div'>;
+TabsPanel.displayName = 'Tabs.Panel';
+
+/**
+ * Tabs organize and allow navigation between groups of content that are related and at the same level of hierarchy.
+ * @example
+ * <Tabs>
+ *   <Tabs.Tab>
```
LostABike, line 748: missing `<Tabs.TabList>` for this example to work

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,399 @@ export const Tabs = (props: TabsProps) => {
...
+  /**
+   *  Tab icon subcomponent
+   */
+  TabIcon: TabsTabIcon,
+  /**
+   * 	Tab info subcomponent
+   */
+  TabInfo: TabsTabInfo,
+  /**
+   * 	Tab label subcomponent
```
LostABike, line 809: nit: maybe slightly more descriptive jsdocs?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,399 @@ export const Tabs = (props: TabsProps) => {
...
+
+const TabsTabIcon = polymorphic.span('iui-tab-icon', {
+  'aria-hidden': true,
+});
+TabsTabIcon.displayName = 'Tabs.TabIcon';
+
+// ----------------------------------------------------------------------------
+// Tabs.TabInfo component
+
+const TabsTabInfo = polymorphic.span('iui-tab-label');
```
LostABike, line 613: since we have the actual label be under `TabInfo` now, maybe this should be named to `iui-tab-info` or similar, and the actual label will have `iui-tab-label` as class name?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,399 @@ export const Tabs = (props: TabsProps) => {
...
+            key={index}
+          >
+            {item}
+          </TabsContext.Provider>
+        );
+      })}
+    </>
   );
 };
+TabsPanels.displayName = 'Tabs.Panels';
```
LostABike, line 717: This is for content within the Tab right? I think `Tabs.Content` would be more a informative name

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,399 @@ export const Tabs = (props: TabsProps) => {
...
+        {...rest}
+      >
+        {children}
+      </Box>
+    );
+  } else {
+    return <></>;
+  }
+}) as PolymorphicForwardRefComponent<'div'>;
+TabsPanel.displayName = 'Tabs.Panel';
```
LostABike, line 742: ~~This is for content within the Tab right? I think Tabs.Content would be more a informative name~~

Now that I think about it more, I was expecting to be able to add tab's content under `TabInfo`, along with label, and not in a separate subcomponent. But the original implementation also had them separate :thinking: Regardless I still think `Tabs.Content` would be a clearer name here

adamhock: I like this idea, but does this also mean the `Tabs.Panels` name has to change as well? Perhaps to `Tabs.Contents`?

There's also the idea of keeping `Tabs.Panel` and replacing the wrapper `Tabs.Panels` name with `Tabs.Content`. What are your thoughts @LostABike @gretanausedaite

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: Good catch! This has been fixed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: `iui-tab-label` has changed to `iui-tab-info`

I did not add the `iui-tab-label` class to `Tabs.TabLabel` since there are no selectors which require that element's class name. CSS attributes are passed using the `> *` selector from the `iui-tab-info` mixin

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: why do we want a generic, undescriptive name like "content"?

"tabpanel" is the canonical, widely-understood term matching the aria role. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role

LostABike: I was just thinking back to Tile's `ContentArea` coupled with my own expectation when using this Tab's component, but wasn't aware of the canonical naming 😅 But TDIL. This is fine then

packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -62,7 +62,7 @@ <h2>Default</h2>
                 class="iui-tab-icon"
                 aria-hidden="true"
               ></svg-placeholder>
-              <span class="iui-tab-label">
+              <span class="iui-tab-info">
                 <div>Active</div>
                 <div class="iui-tab-description">Sublabel</div>
               </span>
```
mayank99, line 68: Can we explore this markup? 

```html
<button class="iui-tab">
  <svg-x class="iui-tab-icon">...</svg-x>
  <span class="iui-tab-label">The label</span>
  <span class="iui-tab-description">The sublabel</span>
</button>
```

Should be achievable using css grid.
```css
.iui-tab {
  display: grid;
  grid-template-columns: auto minmax(0, 1fr);
  grid-template-areas:
    'icon label'
    'icon description';
}
```

adamhock: I tried and pushed this implementation, but I'm having trouble fixing this overflow issue without the use of a wrapper `div` element for the tab label and description. Any suggestions would be appreciated!

image

This new DOM structure is present in the second tab of the third example and the first three tabs in the fourth example in the HTML demo pages for now. I didn't want to apply this change to the rest of the examples or React until I'm certain this implementation works.

mayank99: you can try this:
```diff
- grid-template-columns: auto minmax(0, 1fr)
- grid-template-columns: auto minmax(min-content, 1fr)
```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Try to use ThemeProvider only once.
You are just looping the value in Panel:
image

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: Due to the changes made in the DOM structure, `TabInfo` has been removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: I agree with Mayank here that the `Panel` name is more descriptive and more associated with tabs.

`TabInfo` and `iui-tab-info` have been removed

packages/itwinui-css/backstop/tests/tabs.html
adamhock: > edit: `grid-template-areas` hardcodes everything, so will not work because our icons and descriptions are optional. However, an implicit grid could work if the areas are defined by children.
> 
> ```css
> .iui-tab {
>   display: grid;
>   grid-template-columns: auto minmax(0, 1fr);
> }
> .iui-tab-icon {
>   grid-column: 1;
>   grid-row: 1 / span 2;
> 
>   &:not(:last-child) {
>     margin-inline-end: var(--iui-size-xs);
>   }
> }
> .iui-tab-label {
>   grid-area: 1 / 2;
> }
> .iui-tab-description {
>   grid-area: 2 / 2;
> }
> ```

I did not see this edit until just now! I ended up finding a different CSS solution that works for all cases, but if this solution is better than the one I've pushed please let me know

packages/itwinui-css/backstop/results/bitmaps_reference/iTwinUI_tabs_Type_borderless_dark_0_demo-borderless_0_.png
adamhock, line 1: The reason for these visual changes is because the `max-inline-size` for the wrapper class was set to `60ch`. Since the wrapper is removed, I've set `max-inline-size` for both `iui-tab-label` and `iui-tab-description` to `60ch`. 

Because `iui-tab-description` has a smaller font size and the value for `max-inline-size` is character-based, this value has decreased. I believe this image change is for the better if we're going to use character size to determine the `max-inline-size` value, but if these need to be reverted, let me know.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: `Panel` only gets the `currentActiveIndex` value from its grandparent (`Tabs`) if the parent (`Panels`) doesn't use a context provider. Since `Panels` needs to use a context provider to pass `index` to `Panel`, the `currentActiveIndex` must be passed through this context as well.

Can you be more specific about your ThemeProvider comment?

mayank99: i believe Greta meant TabsContext.Provider, not ThemeProvider

---

here are my general thoughts:
- it doesn't make sense to override context value completely. If the two instances of the context serve different purposes, then you probably need two different contexts
- it's probably not a good idea to create several different contexts inside a loop when most of the values are shared, which brings me to my next point:
- We might want to move a lot of the code and props from `TabsComponent` to `Tabs.Tab`. Each tab knows about itself, so less code will be necessary to make it do similar work. For example, the `activeIndex` prop could be removed from the top level component and instead passed to the individual tabs.
  ```jsx
  <Tabs>
    <Tabs.Tab>...</Tabs.Tab>
    <Tabs.Tab isActive>...</Tabs.Tab>
    <Tabs.Tab>...</Tabs.Tab>
  </Tabs>
  ```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +503,383 @@ export const Tabs = (props: TabsProps) => {
...
+      data-iui-overflow={overflowOptions?.useOverflow}
+      data-iui-scroll-placement={scrollingPlacement}
+      role='tablist'
+      onKeyDown={onKeyDown}
+      ref={refs}
+      {...rest}
+    >
+      {items.map((item, index) => {
+        return (
+          <li key={index}>
```
mayank99, line 529: do we need to use `<ul>`/`<li>` elements? i think divs should work just fine, and we'll be able to remove this extra element completely.

```diff
- <ul role=tablist>
+ <div role=tablist>
-   <li>
      <button role=tab>...</button>
-   </li>
- </ul>
+ </div>
```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: I get why multiple contexts should be used for this component, and I agree that all of the props from `Tabs` aside from `orientation` can at least be passed into `Tabs.TabList` from the user. I'll try your approach with using `isActive` in `Tabs.Tab` to get the active tab.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: The sizing for the vertical tabs is messed up when the `li` wrapper is removed.

image

This is the only part of the tabs css that targets `li` elements, but the same properties applied to it are also applied to `iui-tab`

image

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: After looking into this further, I'm not sure if this is a better approach for determining which tab is active. Would each tab have it's own `onActiveChange` prop? This method also allows users to have more than one active tab at a time. Also, `Tabs.TabList` still needs to know which tab is active relative to the other tabs for keyboard handling, so I think it's best if `activeIndex` and `onTabSelected` are props for `Tabs.TabList`.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: have you tried investigating why and fixing it? maybe there's an `li` somewhere else, or it relies on some parent-child relationship

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: > Would each tab have it's own `onActiveChange` prop?

yes

> This method also allows users to have more than one active tab at a time

user-error 🤷‍♂️ i'd expect users to know how to correctly maintain state, especially when we'll have correct examples.

> Also, `Tabs.TabList` still needs to know which tab is active relative to the other tabs for keyboard handling

that code can also be moved into `Tabs.Tab`, and would likely be better that way

gretanausedaite: > i believe Greta meant TabsContext.Provider, not ThemeProvider

Oops.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: The `ul` element has been replaced with `div` and the `li` wrapper has been removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: - `TabsContext` has been broken up into `TabsTabListContext` and `TabsPanelsContext` so that each context only uses `.Provider` once
- `onTabSelected` and `activeIndex` props have been removed
- `isActive` and `onActiveChange` props have been added to `Tabs.Tab`
- All keyboard navigation is now handled through `Tabs.Tab`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,78 +85,147 @@ export const Tabs = (props: TabsProps) => {
...
+    children,
     orientation = 'horizontal',
-    tabsClassName,
-    contentClassName,
-    wrapperClassName,
+    type = 'default',
+    ...rest
+  } = props;
+
+  const [currentActiveIndex, setCurrentActiveIndex] = React.useState(0);
```
gretanausedaite, line 95: Maybe we could add a prop for users to choose initial active index? In case they want to show something else than Tab0.

examples/Tabs.overflow.tsx
```diff
@@ -3,50 +3,93 @@
...
+                  });
+                  setActiveIndex(index);
+                }}
+              />
+            );
+          })}
+        </Tabs.TabList>
+
+        <Tabs.Panels>
+          <Tabs.Panel>
```
gretanausedaite, line 53: How users know which TabPanel goes with which Tab?

gretanausedaite (commented):
Current naming is a bit confusing, added some suggestions.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -451,36 +425,118 @@ export const Tabs = (props: TabsProps) => {
...
+      )}
+      data-iui-overflow={overflowOptions?.useOverflow}
+      data-iui-scroll-placement={scrollingPlacement}
+      role='tablist'
+      ref={refs}
+      {...rest}
+    >
+      {items.map((item, index) => {
+        return (
+          <TabsTabListContext.Provider
```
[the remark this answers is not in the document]
adamhock: This context provider passes props that are consumed by both `Tab` and `TabDescription`. I figured the context was named according to where it's be provided instead of where it's being consumed.

Does this mean I should wrap the items in 2 sets of Contexts?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: [This comment](https://github.com/iTwin/iTwinUI/pull/1401#discussion_r1275298092) asked for the removal of these two props. Instead, if users want to set an initial active index, they can pass the `isActive` prop to the tab which essentially nullifies the use of `currentActiveIndex`. Both of the overflow examples in Storybook are examples of this method in action

examples/Tabs.overflow.tsx
adamhock: I figured users would list their `Panel` elements in the same order as their `Tab` elements in order to associate the two.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: i think we should do away with the concept of indices. users want to select the tab, not the index.

i expect controlled usage to look like this:
```jsx
const [current, setCurrent] = React.useState('apple');

<Tabs>
  <Tabs.TabList>
    {['apple', 'banana', 'orange'].map(label => (
      <Tabs.Tab
       key={label}
       label={label}
       isActive={current === label}
       onActiveChange={() => setCurrent(label)}
      />
    ))}
  </Tabs.TabList>
</Tabs>
```

examples/Tabs.overflow.tsx
mayank99: on the outside, the association might be based on order, but on the inside, it needs to be explicit through the use of `id`/`aria-controls`. see APG: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ (you can inspect the examples)

it might make sense to ask for a unique value from the users, to explicitly connect the tabs and panels from the outside. not sure about the api 🤔

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
mayank99: > I figured the context was named according to where it's be provided instead of where it's being consumed.

the name should be based on where it is used, yes, but in this case you are using it inside a loop, so it's for individual tabs, not the whole tabslist.

---

but i still think the [comment from earlier](https://github.com/iTwin/iTwinUI/pull/1401#discussion_r1275298092) is not completely addressed - most things inside this context should *not* be necessary, except for `hasSublabel`/`setHasSublabel`. and even in that case, you only need one context, not a loop of several contexts.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -451,36 +425,118 @@ export const Tabs = (props: TabsProps) => {
...
+  disabled?: boolean;
+  /**
+   * Flag whether the tab is active.
+   * @default 'false'
+   */
+  isActive?: boolean;
+  /**
+   * Callback fired when the isActive prop changes.
+   */
+  onActiveChange?: () => void;
```
mayank99, line 490: `onActiveChange` makes it sound like it would also be called when a tab 

what do you think of changing it to `onActivated`/`onActive`?

alternatively, both `isActive`+`onActiveChange` -> `isSelected`+`onSelected`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +586,319 @@ export const Tabs = (props: TabsProps) => {
...
-      {children && (
-        <Box
-          className={cx('iui-tabs-content', contentClassName)}
-          role='tabpanel'
-        >
-          {children}
-        </Box>
-      )}
-    </Box>
+const TabsActions = React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
adamhock: This has been changed

`TabsAction` has also changed to `TabsActions` since all action buttons are supposed to be wrapped in `Tabs.Actions`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,78 +85,147 @@ export const Tabs = (props: TabsProps) => {
...
+   * @default 'auto'
+   */
+  focusActivationMode?: 'auto' | 'manual';
+  /**
+   * Tab items.
+   */
+  children: React.ReactNode[] | React.ReactNode;
+};
+
+const TabsTabList = React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
adamhock: I've changed this to `TabList` instead of `TabsList`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: I've changed this prop to `onActivated`

examples/Tabs.overflow.tsx
adamhock: The `value` prop has been added to `Tab` and `Panel` subcomponents. As a result, an `index` value is no longer passed through context meaning `TabsPanelsContext` has been removed and `TabListContext` does not wrap each `Tab` element!

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: This is a much cleaner usage! I've changed the stories and website to reflect this

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
adamhock: After the addition of the `value` prop in the `Tab` and `Panel` subcomponents, there is no longer a need to loop the `TabListContext` for each individual tab.

Outside of `hasSublabel` / `setHasSublabel`, here are the remaining props I am still passing through context:

- `onTabClick`: A function that clicks a tab given its index. Since tab focusing is done through `Tab`, I needed a way to click on the next available tab given its index when `focusActivationMode` is set to `auto`.
- `setFocusedValue`: Called in the `onClick` and `focusTab` functions to set the new focused value.
- `focusActivationMode`: Prop in `Tabs.TabList` used to check if the tab is clicked when focusing (`'auto'`) or when pressing `Enter` or `Spacebar` (`'manual'`).

examples/Tabs.overflow.tsx
gretanausedaite: Maybe value is not the most intuitive name for this functionality.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
adamhock: These have been updated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,85 +605,283 @@ export const Tabs = (props: TabsProps) => {
...
+}) as PolymorphicForwardRefComponent<'div', TabsPanelOwnProps>;
+TabsPanel.displayName = 'Tabs.Panel';
+
+/**
+ * Tabs organize and allow navigation between groups of content that are related and at the same level of hierarchy.
+ * @example
+ * <Tabs>
+ *   <Tabs.TabList>
+ *     <Tabs.Tab value='tab1'>
+ *       <Tabs.TabLabel>Label 1</Tabs.TabLabel>
```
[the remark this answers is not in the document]
adamhock: This example has been changed

packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -46,85 +46,69 @@ <h2>Default</h2>
...
+          >
+            <svg-add-circular
+              class="iui-button-icon"
+              aria-hidden="true"
+            ></svg-add-circular>
+          </button>
+        </div>
 
         <div class="iui-tabs-actions-wrapper">
           <div class="iui-tabs-actions">
```
adamhock, line 114: I want to open a discussion about `iui-tabs-actions-wrapper` and `iui-tabs-actions`.

In all of these HTML examples and in the old React API for Tabs, `iui-tabs-actions-wrapper` only has one child which is `iui-tabs-actions`. When you place multiple `iui-tabs-actions` elements in the wrapper, this is the result for a horizontal and vertical tab

image
image

If the horizontal example is intended where adding a new `iui-tabs-actions` element creates a new row of action buttons, then this needs to be true for creating a new column of action buttons in the vertical example and the spacing needs to be fixed

If this isn't intended and the wrapper is only supposed to wrap around a single `actions` element, can we look into removing `iui-tabs-actions-wrapper` altogether?

mayank99: i believe the wrapper was needed for grid placement. it might not be possible to remove the class, but you can remove the react component and add `wrapperProps` to `Tabs.Actions`.

PR #1402, opened 2023-07-11
Website Docs Visual Separation

marashad001:
## Changes

#941 - Adds visual separation in between different sections of demos to make them more distinct

Before:
Screenshot 2023-07-12 105400
Screenshot 2023-07-12 105926

After:
Screenshot 2023-07-12 111601
Screenshot 2023-07-12 111536

## Testing

N/A (yet, maybe)

## Docs

N/A (we're just making the docs look cleaner)


LostABike (commented):
Does this change affect all the other pages? Or just the getting started page?

Also, just personal opinion but I think in this case h2 and h3 would look better with a smaller gap, something like this
image

marashad001:
> Does this change affect all the other pages? Or just the getting started page?
> 
> Also, just personal opinion but I think in this case h2 and h3 would look better with a smaller gap, something like this image

Yes, it affects the entire doc website. I agree, I was having trouble coming up with a way to narrow the gap between `h2` and `h3`. Do you have any suggestions for how this could be implemented?

LostABike:
https://github.com/iTwin/iTwinUI/blob/13518490f2bb926f4d0b3f2eba45d8fab85c02f1/apps/website/src/pages/docs/_layout.astro#L214C1-L216

I found the related padding in this file, maybe try meddling with this number? also make sure all other pages still look ok too

mayank99:
what problem is this PR actually trying to solve? because the previous spacing looks better to me.

marashad001:
> what problem is this PR actually trying to solve? because the previous spacing looks better to me.

The spacing between the different sections of the documents are too close together, they need more visual separation  

https://github.com/iTwin/iTwinUI/issues/941#issuecomment-1500391356

The new styling does seperate the sections, but it does need refining to keep the website looking clean and polished.

mayank99:
i believe that comment was already addressed in https://github.com/iTwin/iTwinUI/pull/1193

marashad001:
> i believe that comment was already addressed in #1193

OK. I'll close the PR, it seems redundant to try to fix a resolved comment.

mayank99:
is there some spacing you personally think still needs to be improved? because maybe that PR didn't fully solve the problem.

marashad001:
> is there some spacing you personally think still needs to be improved? because maybe that PR didn't fully solve the problem.

The main thing I notice is the lack of space between the `<h2>` headers and the other blocks of content. Since they are their own section with their own `<h3>` headers within them, they should have a little bit more space in between.

apps/website/src/pages/_global.astro
```diff
@@ -74,6 +74,15 @@ const { content = {} } = Astro.props;
         font-size: var(--type-3);
       }
 
+      :where(h2:not(.astro-4EWMQMN4):not(left-sidebar)) {
```
mayank99, line 77: 1. why are you targeting this hashed class name? it's not meant to be used in source code as it can change at any time.
2. why are you adding this style to the global reset?

it will affect random parts of the website and break at any time

mayank99: the correct place to add this style is inside `.content-wrapper` in `_layout.astro`. you can add it after [line 108](https://github.com/iTwin/iTwinUI/blob/a4e00f411f3f3be5cfd75267ae06ebb1484fd670/apps/website/src/pages/docs/_layout.astro#L108) for example

apps/website/src/pages/_global.astro
```diff
@@ -74,6 +74,15 @@ const { content = {} } = Astro.props;
...
+        border-top: 1px solid var(--color-line-2);
+        inline-size: 100%;
+        border-bottom: 1px solid rgb(245 245 245 / 0%);
+        border-left: 1px solid rgb(245 245 245 / 0%);
+        border-right: 1px solid rgb(245 245 245 / 0%);
```
mayank99, line 82: most of these declarations are unnecessary, you only need `border-block-start` and `padding-block-start`.

```suggestion
        border-block-start: 1px solid var(--color-line-2);
```

apps/website/src/pages/_global.astro
marashad001: fixed

apps/website/src/pages/_global.astro
marashad001: fixed

apps/website/src/pages/docs/_layout.astro
```diff
@@ -106,6 +106,11 @@ const currentPage = new URL(Astro.request.url).pathname;
       grid-area: h1;
     }
 
+    :global(h2:not(left-sidebar)) {
```
mayank99, line 109: i don't think you need the `:not(left-sidebar)` part because this is already nested inside `.content-wrapper`

```suggestion
    :global(h2) {
```

apps/website/src/pages/docs/_layout.astro
```diff
@@ -106,6 +106,11 @@ const currentPage = new URL(Astro.request.url).pathname;
...
+      border-top: 1px solid var(--color-line-2);
```
mayank99, line 110: logical properties should be preferred

```suggestion
      border-block-start: 1px solid var(--color-line-2);
```

marashad001: gotcha

mayank99 (approved):

LostABike (approved):
LGTM 👍

PR #1403, opened 2023-07-11
Dropdown API children improvement

LostABike:
## Changes

- Accept additional type of `JSX.Element[]` and `JSX.Element` for `menuItems` prop

## Testing

- Added unit tests

## Docs

- Added changeset


mayank99 (approved):
looks good. i like that it's backwards compatible!

don't forget to add changeset and unit test.

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
```diff
@@ -44,6 +44,7 @@ export type DropdownMenuProps = {
  *     Item #3
  *   </MenuItem>,
  * ];
+ *
```
adamhock, line 47: Nitpicky, but I would remove this added line just to keep the changes as concise as possible

adamhock (approved):
Minor suggestion, but other than that it looks good to go

PR #1407, opened 2023-07-12
Adding Props to `NonIdealState` Component

marashad001:
## Changes

#1368 - Adding the ability for each node element in `NonIdealState` to have a custom `className`.

## Testing

Not there yet

## Docs

N/A

PR #1406, opened 2023-07-12
Removing setFocus from components

gretanausedaite:
## Changes

Most of components have refs. Therefore custom setFocus functionality is not needed anymore - users can set focus themselves using said ref.

Removed `setFocus` from:
- Checkbox
- ColorPicker
- ComboBox
- Input
- LabeledInput
- LabeledTextarea
- Radio
- RadioTile
- Select
- Slider
- ToggleSwitch

Added setFocus functionality to:
- TextFilter
- NumberRangeFilter
- DateRangeFilter

Keeping setFocus on (because of more complex use case):
- DatePicker
- Dialog
- Menu
- TimePicker

## Testing

- Updated unit tests to manually focus 
- Visual tests pass

## Docs

- [x] changeset
- [ ] wiki

PR #1409, opened 2023-07-13
Improvements to RadioTile DOM Node customization

xman343:
## Changes

Creating `props` variable for each DOM node of the `RadioTile` component. 

Addresses issue #1368.

## Testing

Will be tested through unit & visual testing.

## Docs

TBD


xman343:
Changelog will be created shortly.

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -28,29 +57,82 @@ type RadioTileProps = {
...
+  } = props;
 
   const inputElementRef = React.useRef<HTMLInputElement>(null);
   const refs = useMergedRefs<HTMLInputElement>(inputElementRef, ref);
 
+  React.useEffect(() => {
+    if (inputElementRef.current && setFocus) {
+      inputElementRef.current.focus();
+    }
+  }, [setFocus]);
```
mayank99, line 81: why did you add this `useEffect` and the `setFocus` prop? they were intentionally removed in #1406.

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -28,29 +57,82 @@ type RadioTileProps = {
...
+
+  const { className: tileClassName, ...restTileProps } = tileProps;
+  const { className: inputClassName, ...restInputProps } = inputProps;
+  const { className: contentClassName, ...restContentProps } = contentProps;
+  const { className: iconClassName, ...restIconProps } = iconProps;
+  const { className: labelClassName, ...restLabelProps } = labelProps;
+  const { className: subLabelClassName, ...restSubLabelProps } = subLabelProps;
```
mayank99, line 88: could you use the less verbose approach by _not_ destructuring? see [`fc3e8d6`](https://github.com/iTwin/iTwinUI/pull/1407/commits/fc3e8d6bb4888ad5a27fc633b24803d61a3857e0)

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -28,29 +57,82 @@ type RadioTileProps = {
...
+
   return (
-    <Box as='label' className={cx('iui-radio-tile', className)} style={style}>
+    <Box
+      as='label'
+      className={cx('iui-radio-tile', tileClassName)}
+      style={style}
+      {...restTileProps}
```
mayank99, line 95: i think `tileProps` is unnecessary. we should forward the top-level `...rest` props here, and use the top-level `className` as well.

also i believe we should use the forwarded ref here, rather than on the input.

also pinging @gretanausedaite for a second opinion

packages/itwinui-react/src/core/RadioTiles/RadioTile.test.tsx
```diff
@@ -45,16 +45,87 @@ it('should render radio tile with icon', () => {
   expect(container.querySelector('.iui-radio-tile-icon')).toBeTruthy();
 });
 
-it('should take class and style', () => {
+it('should take tile class and style', () => {
```
adamhock, line 48: I think there could be one test that covers all the possible custom props that can be added instead of a separate test for each element. Thoughts @mayank99 ?

.changeset/lazy-waves-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Props added to RadioTile to individually customize the tile, input, content, icon, label, and sublabel parts of the component.
```
adamhock, line 5: Nitpicky, but I'd mention the new props by name.

adamhock (commented):
I agree with Mayank's suggestions

I also don't think this PR should result in image changes in storybook.

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: I think I added that by mistake, I'll remove it.

xman343: It has been removed

mayank99: needs to be removed from the `type` as well

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
adamhock: https://github.com/iTwin/iTwinUI/pull/1409/files#diff-c1ed339f177f865b3908b978bff96791d3171cbe4bb8f0ddbca5cd73ca77d46aL31

They were defined in this line when setting the props

xman343: Added it back, now it looks like this: 
image

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: done

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: done

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: Fixed!

packages/itwinui-react/src/core/RadioTiles/RadioTile.test.tsx
mayank99: agreed

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
mayank99: lines 65-66 can also be removed. because we are not doing anything special with it, we don't need to create or combine refs, we can simply pass the forwarded ref to the label element

xman343: Done!

mayank99: looks good. i'll leave this comment open for @gretanausedaite to verify

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
mayank99: only the `className` prop needs to be moved to the end, the others can stay as before.

for example, in this case, `as='input'` and `type='radio'` should be moved to the lines before `...inputProps`. this will ensure that `inputProps` can override them.
image

same thing should be done for all elements.

packages/itwinui-react/src/core/RadioTiles/RadioTile.test.tsx
xman343: completed the refactor

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
gretanausedaite: Looks good to me.

packages/itwinui-react/src/core/RadioTiles/RadioTile.test.tsx
mayank99: why not mount a single component that has all of those props? no need for creating a complex loop

xman343: done

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: Done

mayank99: `disabled` should also be moved up

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -20,6 +20,30 @@ type RadioTileProps = {
    * Additional description, if needed.
    */
   description?: React.ReactNode;
+  /**
+   *  Assigns disabled styling to component.
+   */
+  disabled?: boolean;
```
mayank99, line 26: i don't think `disabled` should be added here.

all input props (`disabled`, `checked`, `defaultChecked`, `onChange`) should now be passed through `inputProps` because `...rest` is not applied on the input

mayank99: but this would be kind of an annoying breaking change.

(this is a result of the other suggestion from https://github.com/iTwin/iTwinUI/pull/1409#discussion_r1267015293.)

now i'm thinking it would be better to go the other way - remove `inputProps` and add `wrapperProps`. it would still be a breaking change, but would only break custom styles (rather than breaking functionality)

pinging @gretanausedaite again for second opinion, as i'm unsure.

@xman343 you can work on a different PR in the meantime.

xman343: When I remove `disabled` from here, it no longer recognizes the property, even when I pass it through `inputProps`. 
disabledError1
disabledError2

gretanausedaite: After deeper investigation of RadioTile my official opinion is:
- We should to move `background-color` and `outline` from `iui-radio-tile-content` to `iui-radio-tile`. This way styling wrapper makes the most sense as currently users can't do much when styling it.
- Keep `ref` and `disabled` on `input` - focus styles are on input and it makes the most sense for user to have this ref.
- Have `style` and `classname` props on wrapper
- Have inputProps for styles and classnames (mainly for other input props) for input but not sure users can do much with input component inside radio-tile.

So RadioTile props looks like this:
``` ts
{
  disabled?: boolean;
  inputProps?: inputStuff;
  <...>
}
```

Component:
``` js
<Box className={cx('iui-radio-tile', className)} style={style} {...rest}>
  <Box as='input' ref={ref} disabled={disabled} {...inputProps} />
  <...>
</Box>
```

mayank99: >     * Keep `ref` and `disabled` on `input` - focus styles are on input and it makes the most sense for user to have this ref.
> 
>     * Have `style` and `classname` props on wrapper

I have to disagree with these conflicting ideas put together. It is incredibly confusing for some things (like `ref` and `disabled`) to go on the input and the others (like `className`/`style`/`onClick`) to go on the wrapper and yet others (like `onChange`/`checked` to go inside `inputProps`).

What's the issue with going with the same approach as Checkbox (#1371)?

gretanausedaite: The current RadioTile setup can only be customized by changing `iui-radio-tile-content` part. 
There is no issue with going with the same approach as Checkbox, just I'm not sure if it's useful at all. Even if we decide to add wrapperProps, I'm still suggesting to move background and outline to the wrapper.

So yes, we can do wrapperProps. Just make sure users can actually customise it.

mayank99: Update: I opened https://github.com/iTwin/iTwinUI/pull/1420 to remove content div and move relevant styles to wrapper div.

@xman343 You might want to change your base branch to `mayank/radio-tile-refactor`.

xman343: Am I just focusing on that branch now, or am I rebasing this branch onto `mayank/radio-tile-refactor`? How should I do this in VSCode? (I can ask someone else for more detailed instructions if that's preferable.)

And should I remove `inputProps` and add `wrapperProps` before or after I switch base branches?

mayank99: you will still work in your own branch.

1. change the base branch to `mayank/radio-tile-refactor` using github ui
   

2. run `git merge origin/mayank/radio-tile-refactor` locally and resolve any merge conflicts, then push.

3. remove `inputProps` after that.

xman343: Should I accept "incoming" or "current" changes? 
image

xman343: I'm pretty sure it's incoming, since that's coming from the refactor, but it's kind of hard for me to tell whether that accounts for all the changes we want in the resultant file

mayank99: you need to resolve merge conflicts in a way that makes sense. usually it will be a combination of Incoming and Current. it doesn't have to be perfect as you can always make any corrections in subsequent commits.

xman343: Ok, completed the merge & removed `inputProps`. Placing `wrapperProps` is next.

xman343: Added wrapperProps

mayank99: your entire PR is wrong at this point.

currently:
- `className`/`style` are being set on the wrapper, rather than the input.
- none of the class names are being merged. user-passed classes are completely replacing iui classes.

xman343: Replacing it with the version from `radio-tile-refactor`. I think it'll be easier to fix if I start with that and go from there.

xman343: Moved `style` to `input`

xman343: Just to clarify: Do we want `className` and `style` to be on the input, like in `Checkbox`, or do we want them to be on the wrapper, per Greta's suggestion? https://github.com/iTwin/iTwinUI/pull/1409#discussion_r1269139797

xman343: I read the whole thread. It was just hard to tell from context, and I wanted to be completely sure I was doing the right thing, so that I don't make any further mistakes.

xman343: `style` & `className` now in input

xman343: Should I place `iconProps`, `labelProps`, and `subLabelProps` back into `RadioTile`, or should we keep those out? I got the impression from the thread that we're just doing `disabled` and `wrapperProps` now, but I just want to be 100% positive before I update the changeset.

mayank99: > Should I place `iconProps`, `labelProps`, and `subLabelProps` back into `RadioTile`

yes.

also you are still not correctly merging the classes on the wrapper

xman343: Done

mayank99: classes are still not being merged correctly.

The final resulting class must consist of both the original `iui-` className as well as the user's custom passed class. If that sounds confusing, you can try it out in the playground and see for yourself to get a better idea. The unit test should check for the presence of both classes too.

xman343: Should be fixed now. All the classes merge properly and the unit test evaluates for both the `iui-` and custom className parts of the className.

.changeset/lazy-waves-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+'wrapperProps', 'iconProps', 'labelProps', 'subLabelProps' added to RadioTile to individually customize each part of the component.'
```
mayank99, line 5: last comment: changeset should mention the breaking change where some props have moved

```suggestion
---
'@itwin/itwinui-react': major
---

`RadioTile`'s `className` and `style` props will now be applied on the `<input>` element instead of the wrapper. Added `wrapperProps`, `iconProps`, `labelProps`, `subLabelProps` to individually customize each part of the component.
```

xman343: Done

mayank99 (approved):

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -28,33 +44,64 @@ type RadioTileProps = {
...
+          {...iconProps}
+          className={cx('iui-radio-tile-icon', iconProps?.className)}
+        >
           {icon}
         </Box>
       )}
-      {label && <Box className='iui-radio-tile-label'>{label}</Box>}
+      {label && (
+        <Box
+          as='div'
```
adamhock, line 90: `as='div'` is the default for `<Box>`, so it doesn't need to be defined here

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
```diff
@@ -28,33 +44,64 @@ type RadioTileProps = {
...
+          {...labelProps}
+          className={cx('iui-radio-tile-label', labelProps?.className)}
+        >
+          {label}
+        </Box>
+      )}
       {description && (
-        <Box className='iui-radio-tile-sublabel'>{description}</Box>
+        <Box
+          as='div'
```
adamhock, line 99: Same fix here:

`as='div'` is the default for `<Box>`, so it doesn't need to be defined here

adamhock (commented):
Just two quick comments

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
xman343: done with both

packages/itwinui-react/src/core/RadioTiles/RadioTile.tsx
mayank99: it appears necessary because of some typescript shenanigans. https://github.com/iTwin/iTwinUI/actions/runs/5649482386/job/15303947305

it might be possible to change `Box`/`polymorphic` type definitions to fix this, but i do not have time to investigate and it's easier to just add `as='div'`.

adamhock (approved):

PR #1410, opened 2023-07-13
constrain tabs height to let content grow

mayank99:
## Changes

this should partially fix the user-reported issue where tabs-list was growing when tabpanel content was not big enough.

this is the same as what we do in vertical tabs.

https://github.com/iTwin/iTwinUI/blob/d90b55239b5a14c6c2173ff11ec61597196893b6/packages/itwinui-css/src/tabs/base.scss#L181-L184

## Testing

tested with a few different versions of custom css in codesandbox. also verified in user's app.

## Docs

added changeset

### DIFF
--- a/.changeset/wet-planes-dance.md
+++ b/.changeset/wet-planes-dance.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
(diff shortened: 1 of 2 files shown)


gretanausedaite (approved):
Seems to work for user-reported issue

PR #1412, opened 2023-07-13
wrap all styles in layer

mayank99:
## Changes

All itwinui-react styles will now be under a [cascade layer](https://developer.mozilla.org/docs/Learn/CSS/Building_blocks/Cascade_layers) named `itwinui.v3`. This will prevent any and all specificity battles with application styles, thus removing the need for artificially inflating specificity or using `!important`.

The `.` in the layer name results in a nested layer, meaning if the user wants to control the final layer order, they can reference the `itwinui` layer without messing with versions. For example, they can add a reset before itwinui styles.
```
@layer reset, itwinui, anything-else;
```

itwinui-css and itwinui-variables are currently unchanged; only itwinui-react's bundled stylesheet is affected. I was investigating adding layers to those too, but Sass makes it a bit difficult so I decided to leave it out. Imo this is ok because it keeps the base stylesheets flexible and the developer always has the ability to import into a layer manually.

I will create separate PRs to add `v2` and `v1` layers to older versions.

## Testing

All existing tests pass because there are no code changes.

Confirmed layers are working by inspecting build output and also by inspecting dev tools.

Tried adding low-specificity rules outside the layer and confirmed they override the higher-specifity rules from itwinui.

## Docs

Added changeset. Pending docs.


package.json
```diff
@@ -67,11 +67,11 @@
     "async": "3.2.4",
     "trim": "0.0.3",
     "set-value": "4.1.0",
-    "postcss": "8.4.20",
+    "postcss": "8.4.25",
```
gretanausedaite, line 70: Does this change means that users would need to update postcss/sass/etc too? Or it's just us?

packages/itwinui-react/src/styles.js/styles.module.css
```diff
@@ -2,5 +2,8 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-@import '@itwin/itwinui-variables';
-@import '@itwin/itwinui-css';
+
+@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3;
```
gretanausedaite, line 6: Why both are there: itwinui-v1 and itwinui.v1?

gretanausedaite (commented):
Is it possible to have some kind of instruction set for the users to correctly use layers in their apps?

packages/itwinui-react/src/styles.js/styles.module.css
mayank99: `itwinui-v1` is currently used, i plan to change it to `itwinui.v1` in a future PR. we need to support both

package.json
mayank99: i was updating all these packages when trying to get layers working with itwinui-css/itwinui-variables, but then decided not to, mainly because Sass doesn't make it easy.

these are dev dependencies and we deliver compiled CSS, so the user shouldn't need to update any of these. One exception is some postcss plugins might not understand `@layer`, but currently v1 already uses layers and we haven't heard any reports.

mayank99:
> Is it possible to have some kind of instruction set for the users to correctly use layers in their apps?

Yes i plan to write something up soon.

packages/itwinui-react/src/styles.js/styles.module.css
```diff
@@ -2,5 +2,8 @@
...
+
+@import '@itwin/itwinui-variables' layer(itwinui.v3);
+@import '@itwin/itwinui-css' layer(itwinui.v3);
```
LostABike, line 9: What does this syntaxes mean/do? Like specifying that the import is part of itwinui.v3 layer?

LostABike:
> Tried adding low-specificity rules outside the layer and confirmed they override the higher-specifity rules from itwinui.

I'm still kinda confused how layer work? Isn't this already how it is? Like user's defined styles usually override itwinui's? Or am I misunderstanding something

gretanausedaite (approved):
No comments from my side

mayank99:
> > Tried adding low-specificity rules outside the layer and confirmed they override the higher-specifity rules from itwinui.
> 
> I'm still kinda confused how layer work? Isn't this already how it is? Like user's defined styles usually override itwinui's? Or am I misunderstanding something

No, usually it depends on [specificity](https://web.dev/learn/css/specificity/#visualizing-specificity) and order of appearance. For example, if an itwinui selector is `.iui-expandable-block[data-iui-size='small'`, it has a specificity of 0-2-0, so if the user adds a custom class `.my-expander`, then it will _not_ override itwinui's higher-specificity styles. Layers prevent this issue by bypassing specificity entirely.

You can read more here: https://css-tricks.com/css-cascade-layers/

packages/itwinui-react/src/styles.js/styles.module.css
mayank99: yes, it is importing the stylesheet into the layer: https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules_into_a_cascade_layer

additionally, vite is inlining the contents of the stylesheet using [`postcss-import`](https://github.com/postcss/postcss-import#postcss-import), so the end result looks kinda like this:

```css
@layer itwinui.v3 {
  // ... everything inside itwinui-variables
}
@layer itwinui.v3
  // ... everything inside itwinui-css
}
```

to verify this, you can inspect `styles.css` that gets created after build.

mayank99:
i've added a changeset, so this is good to merge. still working on proper documentation about layers, will come in a separate PR

PR #1413, opened 2023-07-14
Add temporary variables page in html tests

gretanausedaite:
## Changes

Added a page with most of the variables listed so our designer could have an easy access deployed page to look at when adding and editing colours in Figma design library.

This page can be removed later.

## Testing

N/A
## Docs

N/A

### DIFF
--- a/packages/itwinui-css/backstop/tests/variables.html
+++ b/packages/itwinui-css/backstop/tests/variables.html
@@ -21,7 +21,7 @@
     <style>
       @import url("./assets/demo.css") layer(demo);
       @import url('@itwin/itwinui-variables') layer(variables);
-      @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
+      @import url("@itwin/itwinui-css/css/global.css") layer(itwinui);
 
       section[id^='demo'] {
         margin: 0 -24px -24px -24px;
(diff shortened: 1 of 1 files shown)


mayank99:
but this already exists: https://dev-itwinui.bentley.com/docs/variables

gretanausedaite:
> but this already exists: [dev-itwinui.bentley.com/docs/variables](https://dev-itwinui.bentley.com/docs/variables)

Dark mode only. We need light, dark and hc hex values.

packages/itwinui-css/backstop/tests/variables.html
```diff
@@ -0,0 +1,560 @@
...
+      import "@itwin/itwinui-icons-elements/info-circular";
+      import "@itwin/itwinui-icons-elements/status-success";
+      import "@itwin/itwinui-icons-elements/status-warning";
+      import "@itwin/itwinui-icons-elements/status-error";
+      import "@itwin/itwinui-icons-elements/status-rejected";
+    </script>
+    <style>
+      @import url("./assets/demo.css") layer(demo);
+      @import url('@itwin/itwinui-variables') layer(variables);
+      @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
```
mayank99, line 31: component css doesn't need to be included i guess. only variables.

mayank99 (approved):

packages/itwinui-css/backstop/tests/variables.html
gretanausedaite: Removed not needed imports for components and icons

mayank99: this broke some styles! i think `global.css` is needed

gretanausedaite: Should be fixed. all.css was needed

PR #1415, opened 2023-07-14
wrap v2 styles in layer

mayank99:
## Changes

Similar to #1412, except it adds the styles from JS at runtime rather than importing them. This allows the flexibility of switching layers on or off (default on).

This is done in three steps:
1. `generateStyles.mjs` script will generate a `styles.js` file which exports the contents of itwinui-variables and itwinui-css as a plain javascript string.
2. The `useStyles` hook, which gets called in `ThemeProvider` (and also in `useTheme`), will import this string and decide whether or not the final stylesheet should be wrapped in `@layer` (depending on the user-controlled `withLayer` option).
3. The final stylesheet is added to the document using [constructable stylesheets](https://web.dev/constructable-stylesheets/) (which is supported in all browsers, but will also fall back to using a `<style>` tag in case it is not supported somewhere). After this, a flag `iuiv2Loaded` is set on the document to avoid re-adding these styles.

Tangential changes:
- `styles.tsx` was removed, since there is a `styles.js` file similar to #1302. dependency on itwinui-css/itwinui-variables also removed.
- css imports were removed, similar to #1322, because there is a single stylesheet inside JS now.
- expandable-searchbox needed increased specificity because utils was overriding it. [`6426a1d`](https://github.com/iTwin/iTwinUI/pull/1415/commits/6426a1de999d236c1b0647e5d5c519723ff39600?w=1)

## Testing

Tested in playground and storybook. Verified build output and inspected browser devtools to make sure constructable stylesheet is created at runtime. Also verified fallback `<style>` tag is created.

Tested toggling `withLayer` prop in playground, working correctly.

Not sure if it makes sense to test this in jest as it would require mocking everything, but all visual tests pass, so that serves as a sanity check at least.

## Docs

**Pending**.

Because this code diverges significantly from the `dev` branch, I'm not sure if this should be merged right now or after v3 is stable.


gretanausedaite (commented):
`build:styles` command is getting 'operation not permitted' error:
image

mayank99:
> `build:styles` command is getting 'operation not permitted' error

i think it was because `styles.js` empty directory (from switching branches) gets copied by swc. should be fixed now.
 
> > Because this code diverges significantly from the dev branch, I'm not sure if this should be merged right now or after v3 is stable.
> 
> Are these changes needed in v3? I thought it's only for v2 -> no need to merge with 'dev' branch because v3 changes takes care of this in different way.

we have been merging `main` back into `dev` so far. this way `dev` stays up to date and can be easily merged into `main` when it's time for stable release.

still unsure about this, but if we do merge it first, then i guess i'll resolve merge conflicts when updating `dev` and take care of any accidents in subsequent PRs.

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Dynamically loads the iTwinUI styles as a constructed stylesheet,
+ * falling back to a regular `<style>` tag in `<head>`.
+ */
+export const useStyles = (options?: {
+  withLayer?: boolean;
+  document?: () => Document | undefined;
+}) => {
+  const { withLayer = true, document = () => getDocument() } = options ?? {};
```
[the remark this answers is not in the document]
mayank99: by default, v2 _will_ be added to layer. you can inspect any of the components.

to turn layers off (in case something is breaking and impossible to fix), user can do this:
```jsx
<ThemeProvider styleOptions={{ withLayer: false }}>
```

(suggestions for better prop names welcome)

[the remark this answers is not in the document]
LostABike: Does it mean that users can cherry pick what components use which layer?

LostABike (commented):
I'm also getting this despite force build

image

mayank99:
> I'm also getting this despite force build

Try deleting the `styles.js` folder from `src` before running build/dev.

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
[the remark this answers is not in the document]
mayank99: no, it's all or nothing, since there is a single stylesheet.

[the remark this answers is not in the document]
gretanausedaite: I was just thinking, if the current approach is no layer at all, then maybe `withLayer` prop should be 'false' by default. Adding styles to a layer if user does not expect layers could break things (I think).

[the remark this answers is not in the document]
gretanausedaite: ok, makes sense

gretanausedaite (approved):
LGTM 🌵

LostABike (approved):
No other comments. I played around with this and it is working as expected. Why the test images change though? I think those are common failure images from testing flakiness.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -61,9 +60,22 @@ type ThemeProviderOwnProps = Pick<RootProps, 'theme'> &
...
     | {
         themeOptions?: RootProps['themeOptions'];
+        styleOptions?: {
+          /**
+           * If true, all styles will be wrapped in a cascade layer named `itwinui`.
+           * This helps avoid specificity battles with application styles.
+           *
+           * @default true
+           */
+          withLayer?: boolean;
```
mayank99, line 70: @gretanausedaite @LostABike Do you have a better naming suggestion?

I was thinking of renaming `styleOptions` to `autoImportStyles` and also allowing boolean. The boolean would be useful if somebody wants to import `@itwin/itwinui-css` manually.

```jsx
<ThemeProvider autoImportStyles={{ withLayer: false }}>
```
```jsx
<ThemeProvider autoImportStyles={false}>
```

Alternatively, keep `styleOptions` and rename `withLayer` to just `layer` or `layers`. Can also add `disabled` property if manually importing.
```jsx
<ThemeProvider styleOptions={{ layer: false }}>
```
```jsx
<ThemeProvider styleOptions={{ disabled: true }}>
```

LostABike: I feel `styleOptions` -> `styleImports` is a bit more concise than `autoImportStyles`. Allowing boolean would be nice.

Also I think this
```
<ThemeProvider autoImportStyles={false}>
```
is more intuitive than 
```
<ThemeProvider styleOptions={{ disabled: true }}>
```
at least for me.

marashad001:
For this PR, this is changes the way that styling is used and shared between `itwinui-css` and `itwinui-react`, correct? If a user were to disable the `withLayer` option, would the code still have some styling or just be an empty shell?

packages/itwinui-react/jest.config.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
 
   // A map from regular expressions to module names or to arrays of module names that allow to stub out resources with a single module
   moduleNameMapper: {
-    '\\.(css|less|sass|scss)$': '<rootDir>/src/__mocks__/styleMock.js',
+    'styles.js$': '<rootDir>/src/__mocks__/styleMock.js',
```
marashad001, line 86: This changes the way the css is imported into React right? Through `styles.js` file instead of the files in the old folder?

mayank99: no, this file is only used for configuring `jest` unit tests.

specifically this line is used to return a dummy `styleMock.js` instead of the real `styles.js` file (which doesn't exist in source code).

you can read about this option here: https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
```diff
@@ -0,0 +1,75 @@
...
+  const loaded = _React.useRef(false);
+
+  useIsomorphicInsertionEffect(() => {
+    if (!loaded.current) {
+      loadStyles({ withLayer, document });
+      loaded.current = true;
+    }
+  }, [document, loaded, withLayer]);
+};
```
marashad001, line 35: Does this append the new layer onto the current styling?

mayank99:
> For this PR, this is changes the way that styling is used and shared between `itwinui-css` and `itwinui-react`, correct?

yes and no. this PR doesn't change `itwinui-css` at all; it only changes `itwinui-react` so that it doesn't explicitly depend on `itwinui-css`.

> If a user were to disable the `withLayer` option, would the code still have some styling or just be an empty shell?

it will have styling without layers. you can try this yourself in the vite playground.

> Would this feature be so the user could create and customize their own styling for their application?

yes, it makes it easier to customize our components.

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
mayank99: not sure i understand the question, can you rephrase?

packages/itwinui-react/src/core/utils/hooks/useTheme.ts
```diff
@@ -6,8 +6,7 @@ import * as React from 'react';
 import { getDocument, getWindow } from '../functions/index.js';
 import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect.js';
 import { useIsThemeAlreadySet } from './useIsThemeAlreadySet.js';
-import '@itwin/itwinui-css/css/global.css';
```
marashad001, line 9: If we're not using the imports anymore, will they disappear from the repo? They're still being used by the html pages I assume.

mayank99: please read the PR description. this is the same change as #1322

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
marashad001: I re-read some more of the PR, never mind. Thanks

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: another option: `includeCss`

```jsx
<ThemeProvider includeCss={{ withLayer: false }}>
```
```jsx
<ThemeProvider includeCss={false}>
```

gretanausedaite: > The boolean would be useful if somebody wants to import @itwin/itwinui-css manually.

Why would they want it?

> autoImportStyles={false}

Users might think that `autoImportStyles={false}` will not import styles at all and not just disables layers.

I like `styleImports` with options like `layers` (or `useLayers`) and `disabled` to disable all style imports.

mayank99: > > The boolean would be useful if somebody wants to import @itwin/itwinui-css manually.
> 
> Why would they want it?

it can be useful if they want more control over the css. For example, inlining the styles into html or running postcss transformations.

> > autoImportStyles={false}
> 
> Users might think that `autoImportStyles={false}` will not import styles at all and not just disables layers.

yes that's what it does. it will not include styles at all.

> I like `styleImports` with options like `layers` (or `useLayers`) and `disabled` to disable all style imports.

`use` prefix is meant for hooks and i would like to avoid naming properties with this prefix.

gretanausedaite: > <ThemeProvider includeCss={{ withLayer: false }}>

I like this one then ^

LostABike: ^ agree with @gretanausedaite, wording for both is pretty clear in that case

mayank99: Renamed!

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    loaded.current = true;
+  }, [document, loaded, context]);
+};
+
+// ----------------------------------------------------------------------------
+
+const layers = `@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3;`;
+
+const loadStyles = ({ withLayer = true, document = () => getDocument() }) => {
```
gretanausedaite, line 56: Do we need tests for this?

PR #1416, opened 2023-07-18
Xander/carousel accessibility compliance

xman343:
## Changes

Bringing `Carousel` component to accessibility compliance by distributing proper roles and labels within the component. 

Addresses #1148.
## Testing

Evaluated using Cypress a11y tests. Unit & visual testing conducted as needed. 

## Docs

N/A


packages/itwinui-react/src/core/Carousel/CarouselNavigation.tsx
xman343, line 1: I'm not 100% sure about these button labels, I just figured "Previous" and "Next" would be the best way to describe their function in relation to the carousel for screenreader users. If there's a better way to label them, please let me know.

mayank99: i would use `aria-label`, as a tooltip is unnecessary in this case

packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
```diff
@@ -51,7 +51,6 @@ export const CarouselSlide = React.forwardRef((props, ref) => {
     <Box
       as='li'
       className={cx('iui-carousel-slider-item', className)}
-      role='tabpanel'
```
mayank99, line 54: i believe you need `role='group'` when using `aria-roledescription`

examples/Carousel.onlyDots.tsx
```diff
@@ -46,8 +46,7 @@ export default () => {
         {gradients.map(({ from, to }, index) => (
           <li
             key={index}
-            role='tabpanel'
-            id={`${id}-slide-${index}`}
+            id={`${id}--slide-${index}`}
```
mayank99, line 49: why was this id changed?

xman343: There was originally a typo that caused one of the examples to fail, but I should change this back to `-slide-` and then change the example id to match that.

xman343: There was a typo in the example that caused it to throw an error. It was supposed to be `--slide-`, which is the format used in the `CarouselDotsList` component. 
image

packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
mayank99: that's interesting, because [APG](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) suggests using `role=group` (check the heading "Basic carousel elements")

i believe it might be because we are using `li`. we could make it a generic `div`. if we do that, the parent `ol` would also need to be a `div`.

examples/Carousel.onlyDots.tsx
mayank99: hmm i think this behavior is actually wrong. `CarouselDotsList` is the component, whereas `Carousel.onlyDots` is the example; the latter should not rely on implementation details of the former.

i wonder if we even need `id` and `aria-controls`. since we removed `role=tabpanel` in https://github.com/iTwin/iTwinUI/pull/1416#discussion_r1268630982, i think `role=tab` can also be removed from `CarouselDot`. i will double check tomorrow

packages/itwinui-react/src/core/Carousel/CarouselNavigation.tsx
xman343: Done!

packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
xman343: Done

xman343:
There's a unit test that's failing, so I'm going to work on that too.

xman343:
> 

Done, just had to update a querySelector (ul -> div) and a role (tabpanel -> group)

examples/Carousel.onlyDots.tsx
xman343: I removed `role=tab` from `CarouselDot` myself, and it generated these violations. I think we have to  keep this role, unless something else can be changed that would allow us to remove it. 
image

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
```diff
@@ -86,7 +86,7 @@ export const CarouselSlider = React.forwardRef((props, ref) => {
 
   return (
     <Box
-      as='ol'
+      as='div'
```
LostABike, line 89: Might not be needed because `<Box>` default is `div`, but other than that LGTM

LostABike (approved):

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
xman343: You were right! I removed `div` and it still works perfectly

examples/Carousel.onlyDots.tsx
mayank99: so i've been looking into this and need more time. I think fixing Carousel accessibility would require some bigger changes across the whole component, so you could move this PR to draft and work on a different PR in the meantime.

i will pick up Carousel later

PR #1418, opened 2023-07-19
bump lightningCss targets

mayank99:
## Changes

Old targets were producing suboptimal output full of unnecessary fallbacks. Bumped these up for make the output leaner.

## Testing

Diffed the build output (before vs after). Most changes are only in logical vs physical properties.

`appearance: none` and `:autofill` are used without `-webkit` fallback (support added in safari 15.x)

image

## Docs

Not sure this needs to be mentioned in changelogs. #1373 is not released yet.


adamhock (approved):
Ran locally. LGTM

packages/itwinui-css/scripts/lightningCssSettings.js
```diff
@@ -3,9 +3,9 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 const targets = {
-  chrome: (90 << 16) | (0 << 8), // chrome 90.0
-  firefox: (90 << 16) | (0 << 8), // firefox 90.0
-  safari: (14 << 16) | (0 << 8), // safari 14.0
+  chrome: (105 << 16) | (0 << 8), // chrome 105.0
```
gretanausedaite, line 6: I think this needs to be mentioned in our docs (or migration guide) that this package works best with these browser versions.

gretanausedaite (approved):
🌵

packages/itwinui-css/scripts/lightningCssSettings.js
mayank99: it's interesting, because we don't really target specific browser versions on purpose. maybe we could mention a moving target like "it works with last two major versions of evergreen browsers". see example: https://shoelace.style/index.html#browser-support

gretanausedaite: "Design for ..." or "Best experience with ..."

PR #1419, opened 2023-07-19
fix(Input): Added `htmlSize` prop to set native `size` attribute

### LINKED ISSUE #973: v3: Input size prop overrides native `size` attribute
native input has size prop, which we drop out. Need to rename our prop and return the native prop.

adamhock:
## Changes

Added `htmlSize` prop to the `Input` component which handles the native `size` attribute in `<input>`.

Closes #973 

## Testing

Added new unit test

## Docs

Updated documentation to explain new prop
Changeset added

### DIFF
--- a/.changeset/shaggy-hornets-enjoy.md
+++ b/.changeset/shaggy-hornets-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+The Input component's `size` prop has been renamed to `localSize` so it doesn't conflict with the native `size` prop found in `<input>`
--- a/.changeset/shaggy-hornets-enjoy.md
+++ b/.changeset/shaggy-hornets-enjoy.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': major
+'@itwin/itwinui-react': minor
 ---
 
 Added `htmlSize` prop to the `Input` component which handles the native `size` attribute in `<input>`.
--- a/.changeset/shaggy-hornets-enjoy.md
+++ b/.changeset/shaggy-hornets-enjoy.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': major
 ---
 
-The Input component's `size` prop has been renamed to `localSize` so it doesn't conflict with the native `size` prop found in `<input>`
+The Input component's `size` prop has been renamed to `size` so it doesn't conflict with the native `size` prop found in `<input>`
--- a/.changeset/shaggy-hornets-enjoy.md
+++ b/.changeset/shaggy-hornets-enjoy.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': major
 ---
 
-The Input component's `size` prop has been renamed to `size` so it doesn't conflict with the native `size` prop found in `<input>`
+Added `htmlSize` prop to the `Input` component which handles the native `size` attribute in `<input>`.
--- a/packages/itwinui-react/src/core/Input/Input.tsx
+++ b/packages/itwinui-react/src/core/Input/Input.tsx
@@ -11,7 +11,7 @@ export type InputProps = {
   /**
    * Modify size of the input.
    */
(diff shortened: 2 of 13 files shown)


packages/itwinui-react/src/core/Input/Input.tsx
```diff
@@ -11,26 +11,26 @@ export type InputProps = {
   /**
    * Modify size of the input.
    */
-  size?: 'small' | 'large';
+  localSize?: 'small' | 'large';
```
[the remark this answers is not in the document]
gretanausedaite: `inputSize` | `boxSize` ? localSize sounds like internal variable.

mayank99:
maybe we can leave `size` as-is and then add `htmlSize` for native attribute? we can explain the difference in the jsdoc and our documentation site.

reason: it is way more common to use the `size` prop to change height, than it is to use the native `size` attribute to constrain width.

adamhock:
> maybe we can leave `size` as-is and then add `htmlSize` for native attribute? we can explain the difference in the jsdoc and our documentation site.
> 
> reason: it is way more common to use the `size` prop to change height, than it is to use the native `size` attribute to constrain width.

Great idea! This has been updated.

LostABike (commented):
Everything looks fine, I just have a question. The native `size` attribute of `<input>` is to determine how many characters will be visible right? However, because we have `inline-size = 100%` within our `<Input/>` component styling, setting this `size` (or `htmlSize`) will not do anything unless user specifically override our component `inline-size`. Is this intended?

.changeset/shaggy-hornets-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
```
gretanausedaite, line 2: I believe this is not major change? We are adding attribute, not changing current API.

packages/itwinui-react/src/core/Input/Input.tsx
```diff
@@ -12,6 +12,10 @@ export type InputProps = {
    * Modify size of the input.
    */
   size?: 'small' | 'large';
+  /**
+   * Modify the `size` attribute of the `<input>` element.
```
gretanausedaite, line 16: Maybe mention it's for native attr.

gretanausedaite (approved):
LGTM.

adamhock:
> > The native `size` attribute of `<input>` is to determine how many characters will be visible right? However, because we have `inline-size = 100%` within our `<Input/>` component styling, setting this `size` (or `htmlSize`) will not do anything unless user specifically override our component `inline-size` styling. Is this intended?
> 
> It is intended that our input fills up the width of the container, because otherwise it would have some weird default width. So yeah, to make use of this `size` (`htmlSize`) attribute, the developer would need to use `width: unset` or `inline-size: unset`. We could mention this in the jsdoc comment for our prop.

Could we apply `width: unset` or `inline-size: unset` stylings to `<Input>` if an `htmlSize` prop is passed?

packages/itwinui-react/src/core/Input/Input.tsx
adamhock: Fixed

.changeset/shaggy-hornets-enjoy.md
adamhock: Updated

mayank99:
> Could we apply `width: unset` or `inline-size: unset` stylings to `<Input>` if an `htmlSize` prop is passed?

Overengineering it imo. I do not expect this attribute to be used commonly enough to justify the extra code. If somebody finds the current behavior confusing, we can always revisit in the future.

Oh and slight typo in [`6ef5d74`](https://github.com/iTwin/iTwinUI/pull/1419/commits/6ef5d740ebd9c98e8aed4dca406f6c44d05b671c) btw. Should say "properties" instead of "attributes".

LostABike (approved):
LGTM :D

PR #1420, opened 2023-07-20
remove radio-tile-content div

mayank99:
## Changes

Slight refactor to remove `<div class="iui-radio-tile-content">`.

```diff
 <label class="iui-radio-tile">
   <input class="iui-radio-tile-input" type="radio" />
-  <div class="iui-radio-tile-content">
     <svg class="iui-radio-tile-icon">...</svg>
     <div class="iui-radio-tile-label">...</div>
     <div class="iui-radio-tile-sublabel">...</div>
-  </div>
 </label>
```

I believe it was needed mainly to make `+` selector work with the input states (`:checked`/`:disabled`). Previously, the input was 0x0 whereas now it covers the entire size of the tile, so we can apply focus styles directly to it, and still use `+`/`~` to target sibling elements. Only problem I ran into is with disabled background-color which is now set on the parent, so I used `:has` for that and added a data attribute as fallback for firefox.
```diff
 <label
  class="iui-radio-tile"
+ data-iui-disabled="true"
 >
   <input class="iui-radio-tile-input" type="radio" disabled />
   ...
 </label>
```

Also used `display: grid` so the content is correctly centered now (previously it was off-center vertically).

## Testing

Verified html pages and storybook. Updated visual test images.

## Docs

Added changesets.


mayank99:
@xman343 Please don't push commits to others' branches without permission.

gretanausedaite (approved):

xman343:
> @xman343 Please don't push commits to others' branches without permission.

my mistake

adamhock (approved):

PR #1422, opened 2023-07-20
v2+v1: use `all: revert-layer` instead of `all: revert`

mayank99:
## Changes

`all: revert` causes too many issues which can be avoided by using `revert-layer`. This works because we only care about reverting the styles that come from v1.

## Testing

Tested by installing itwinui v1 in the vite playground as an npm alias, then using this code:

<details>
<summary>main.tsx</summary>

```jsx
import { createRoot } from 'react-dom/client';
import { Button, ThemeProvider } from '@itwin/itwinui-react';
import { Button as V1Button, UserIcon as V1UserIcon } from 'itwinui-v1';

const Shell = () => {
  return (
    <div>
      <V1Button>hello v1</V1Button>
      <V1UserIcon
        size='small'
        abbreviation='V1'
        backgroundColor='green'
        title='v1 outside iui-root'
      />

      <ThemeProvider>
        <V1Button>v1 inside iui-root</V1Button>

        <V1UserIcon
          size='small'
          abbreviation='V1'
          backgroundColor='green'
          title='v1 inside iui-root'
        />

        <Button>v2</Button>
      </ThemeProvider>
    </div>
  );
};

createRoot(document.getElementById('root')!).render(
  <React.StrictMode>
    <Shell />
  </React.StrictMode>,
);

```

</details>

Result:

Two cases
- `iui-button` is shared by v1 and v2, so using v1 button inside v2 boundary results in v1 styles being reverted but button still looking fine because of v2 styles.
- `iui-user-icon` is present in v1 but removed in v2, so using v1 UserIcon inside v2 boundary correctly removes all styles.

## Docs

Added changesets. Working on "proper" docs.


packages/itwinui-react/src/core/utils/hooks/useStyles.ts
```diff
@@ -11,6 +11,14 @@ import * as React from 'react';
 // @ts-ignore
 import rawCssText from '../../../styles.js';
 
+/** rule that reverts all v1 styles when used inside a v2 boundary */
+const revertV1Styles = (() => {
+  // using :not(#\#) to artifically inflate the specificity of * selector
+  const allSelector = `:where(.iui-body, [class*='iui-theme-']) :where(.iui-root) :not(#\\#)`;
+  const revertRule = `${allSelector}, ${allSelector}::before, ${allSelector}::after { all: revert-layer; }`;
```
gretanausedaite, line 18: Seems like revert-layer is not widely supported yet. https://developer.mozilla.org/en-US/docs/Web/CSS/revert-layer#browser_compatibility
I think we need more testing before adding this.

mayank99: the data is wrong. it is widely supported everywhere that cascade layers are supported.
https://wpt.fyi/results/css/css-cascade?label=master&label=stable&product=chrome&product=firefox&product=safari&aligned&q=revert-layer

gretanausedaite (approved):

packages/itwinui-css/src/revert-v1.scss
mayank99, line 1: @gretanausedaite @adamhock I've pushed another update. Before the latest batch of changes, these revert-v1-styles were only present in the JS (react), whereas now they are also present in the css in case somebody is using iTwinUI without react.

It required some more code but hopefully looks cleaner than trying to write css inside a javascript string. Lmk what you think.

gretanausedaite: Code does look cleaner! 👍

adamhock (approved):

PR #1423, opened 2023-07-20
v1: use nested `.v1` layer name instead of `-v1` suffix

mayank99:
## Changes

Follow-up to #1412 and #1415. Renamed the layer and also specified layer order upfront (in `global.css`) to ensure correct order is used.

## Testing

N/A

## Docs

updated changelog.

### DIFF
--- a/packages/itwinui-css/src/style/global.scss
+++ b/packages/itwinui-css/src/style/global.scss
@@ -3,7 +3,9 @@
 @import './index';
 @import './anchor';
 
-@layer itwinui-v1 {
+@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3;
+
+@layer itwinui.v1 {
   html {
     @include theme-variables(light);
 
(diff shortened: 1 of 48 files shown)


packages/itwinui-css/src/style/global.scss
```diff
@@ -3,7 +3,9 @@
 @import './index';
 @import './anchor';
 
-@layer itwinui-v1 {
+@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3;
```
adamhock, line 6: Can the `itwinui-v1` layer be removed?

adamhock (commented):
The two visual tests that failed in the CI also failed when building locally

packages/itwinui-css/src/style/global.scss
mayank99: no, we need both. https://github.com/iTwin/iTwinUI/pull/1412#discussion_r1263716496

users might have the old stylesheet included somewhere

mayank99:
> The two visual tests that failed in the CI also failed when building locally

unrelated to this PR and it looks fine, so i'm not worried about the 0.5% difference. this is the old v0 css codebase that will likely never be updated again.

adamhock (approved):

gretanausedaite (approved):

PR #1427, opened 2023-07-21
fix: Add minWidth to resizable Dialog

gretanausedaite:
## Changes

Add new class `iui-dialog-resizable` to fix resize bug.

Before:

https://github.com/iTwin/iTwinUI/assets/83585998/23744a6d-712e-407a-9b43-500cc3fab363

After:

https://github.com/iTwin/iTwinUI/assets/83585998/23770bb5-f2d5-4711-9f19-2eb733217e9e

## Testing

WIP

## Docs

WIP

### DIFF
--- a/packages/itwinui-css/src/dialog/dialog.scss
+++ b/packages/itwinui-css/src/dialog/dialog.scss
@@ -4,6 +4,7 @@
 
 $iui-dialog-width: 380px;
 $iui-dialogue-resizable-min-height: 7.75rem; // 7.75rem = 124px = title bar height + margins + button height + partially visible text
+$iui-dialogue-resizable-min-width: 14.5rem; // 14.5rem = 232px = margins + button width x 2
 
 .iui-dialog-wrapper {
   position: fixed;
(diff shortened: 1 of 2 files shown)


packages/itwinui-css/src/dialog/dialog.scss
```diff
@@ -59,7 +59,7 @@ $iui-dialog-min-height: 7.75rem; // 7.75rem = 124px = title bar height + margins
 
 .iui-dialog-default {
   max-width: Max(50%, $iui-dialog-width);
-  min-width: Min(100%, $iui-dialog-width);
```
mayank99, line 62: this `Min` function was added intentionally, because we don't want the dialog to overflow the container or the screen (mobile phones can be narrower than 380px).

adamhock (commented):
The image differences that showed up in the storybook CI also showed up for me locally, so a new test image may need to be pushed as a result of this change

Other than that, I have no additional comments

.changeset/kind-wasps-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Fixed the issue where Dialog moved across the screen when resizing.
```
mayank99, line 6: changeset could be removed. there is no change in css, and there is already a react changeset from #1424.

gretanausedaite: Changeset removed

mayank99 (approved):

PR #1436, opened 2023-07-24
Adding Props to `Slider` Component

marashad001:
## Changes - #1368 

Adding the ability for every DOM node to have a custom className passed to it

## Testing 

After adding the new props into `Slider.tsx`, a specific prop can be passed into the return() code for `Slider` that includes the new custom className.

## Docs

N/A


packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -152,6 +152,26 @@ export type SliderProps = {
...
+   */
+  sliderMinProps?: React.ComponentProps<'span'>;
+  /**
+   * Allows props to be passed for slider-max
+   */
+  sliderMaxProps?: React.ComponentProps<'span'>;
+  /**
+   * Allows props to be passed for slider-container
+   */
+  sliderContainerProps?: React.ComponentProps<'div'>;
```
mayank99, line 166: there is already `railContainerProps`, so `sliderContainerProps` is not needed

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -152,6 +152,26 @@ export type SliderProps = {
    * Additional props for container `<div>` that hold the slider rail, thumbs, and tracks.
    */
   railContainerProps?: React.HTMLAttributes<HTMLDivElement>;
+  /**
+   * Allows props to be passed for slider-min
+   */
+  sliderMinProps?: React.ComponentProps<'span'>;
```
mayank99, line 158: lets remove `slider` prefix from all of these to make the names consistent

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -490,6 +529,7 @@ export const Slider = React.forwardRef((props, ref) => {
           );
         })}
         <Track
+          {...sliderTrackProps}
```
mayank99, line 532: should go at the end. only `className` prop (if present) should go after it

packages/itwinui-react/src/core/Slider/Slider.tsx
mayank99, line 1: missing props for ticks area

marashad001: I'm on it

packages/itwinui-react/src/core/Slider/Slider.tsx
marashad001: fixed

packages/itwinui-react/src/core/Slider/Slider.tsx
marashad001: fixed

packages/itwinui-react/src/core/Slider/Slider.tsx
marashad001: fixed

packages/itwinui-react/src/core/Slider/Slider.tsx
mayank99: you still need to fix the existing `railContainerProps` to actually merge the classes correctly

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -412,9 +437,18 @@ export const Slider = React.forwardRef((props, ref) => {
...
+          className={cx('iui-slider-ticks', tickProps?.className)}
+        >
           {tickLabels.map((label, index) => (
-            <Box as='span' key={index} className='iui-slider-tick'>
+            <Box
+              as='span'
+              {...tickProps}
+              key={index}
+              className='iui-slider-tick'
+            >
```
LostABike, line 451: `iui-slider-ticks` seem to be a wrapper class of `iui-slider-tick` (notice one has "s") so I think they need different props

LostABike: also don't forget to add test for this too

mayank99 (approved):
LGTM. lets merge it without second approval, as Annie is with a different team now

packages/itwinui-react/src/core/Slider/Slider.tsx
marashad001: fixed

PR #1437, opened 2023-07-25
Table DOM Node customization 

xman343:
## Changes

Improving users' ability to customize the DOM nodes of the `Table` component. 
Addresses #1368. 

## Testing

Unit & visual testing will be conducted.

## Docs

TBD


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1060,7 +1094,14 @@ export const Table = <
             </Box>
           )}
           {!isLoading && data.length === 0 && !areFiltersSet && (
-            <Box className='iui-table-empty'>
+            <Box
+              as='div'
+              {...emptyTableContentProps}
+              className={cx(
+                'iui-table-empty',
```
mayank99, line 1101: there are multiple instances of the iui-table-empty div. all of them should be updated.

xman343: Done!

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -4695,3 +4695,46 @@ it('should ignore top-level Header if one is passed', async () => {
...
 });
+
+it('should pass custom props to different parts of Table', () => {
+  const { container } = renderComponent({
+    data: [],
+    headerWrapperProps: {
+      className: 'custom-header-wrapper-class',
+      style: { fontSize: 12 },
+    },
+    headerProps: { className: 'custom-header-class', style: { fontSize: 12 } },
```
LostABike, line 4706: nit: I would test different value for each different class (i.e 12, 13, 14...) just in case a style get inherited and not actually the passed in prop value and we wouldn't be able to tell when the values are the same

xman343: Good call, I'll implement that in the test

mayank99 (approved):
LGTM for first iteration. We can revisit later if more customization is necessary

LostABike (approved):

packages/itwinui-react/src/core/Table/Table.test.tsx
xman343: Added!

.changeset/lovely-trees-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
```
mayank99, line 2: NOT major. Please read https://semver.org/

```suggestion
'@itwin/itwinui-react': minor
```

PR #1439, opened 2023-07-25
Animate Side Navigation submenu with CSS

LostABike:
## Changes

- Removed `WithCSSTransition` from Side Navigation submenu's transition and instead use [CSS](https://css-tricks.com/css-grid-can-do-auto-height-transitions/)
- Not working correctly yet, animation seems off speed for the different layers:
sidenav

## Testing

- Manual testing in Storybook

## Docs

- TBD: changeset

### DIFF
--- a/packages/itwinui-css/src/side-navigation/side-navigation.scss
+++ b/packages/itwinui-css/src/side-navigation/side-navigation.scss
@@ -161,8 +161,27 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   }
 }
 
+//animating technique from https://css-tricks.com/css-grid-can-do-auto-height-transitions/
+.iui-submenu {
+  display: grid;
+  grid-template-columns: 0fr;
+  overflow: hidden;
+  z-index: -1;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: grid-template-columns var(--iui-duration-2), visibility var(--iui-duration-2),
+      transform var(--iui-duration-2);
+    transition-timing-function: ease-out;
+  }
+}
+
 .iui-side-navigation-submenu {
-  min-inline-size: calc(var(--iui-size-3xl) * 2);
+  // min-block-size: 0;
+  min-inline-size: 0;
+  visibility: hidden;
+  @media (prefers-reduced-motion: no-preference) {
+    transition: visibility var(--iui-duration-2) ease-out;
+  }
+  // min-inline-size: calc(var(--iui-size-3xl) * 2);
   max-inline-size: 50vw;
   block-size: 100%;
   overflow-x: hidden;
@@ -171,13 +190,6 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   resize: horizontal;
   background-color: var(--iui-color-background);
   border-inline-end: 1px solid var(--iui-color-border);
-
-  @include mixins.iui-transition-group;
-
-  &.iui-enter-active,
-  &.iui-exit-active {
-    display: flex;
-  }
 
   &-content {
     padding: 0 var(--iui-size-s) var(--iui-size-s);
@@ -214,8 +226,24 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   }
 }
 
+.iui-submenu.iui-expanded {
+  grid-template-columns: 1fr;
+}
+
+.iui-submenu.iui-expanded .iui-side-navigation-submenu {
+  visibility: visible;
+  transform: translateX(0%);
+}
+
+.iui-submenu {
+  &:not(.iui-expanded) {
+    transform: translateX(-100%);
+  }
+}
+
 .iui-side-navigation-wrapper {
   display: flex;
   position: relative;
   block-size: 100%;
+  isolation: isolate;
 }
--- a/packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
+++ b/packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import cx from 'classnames';
-import { WithCSSTransition, SvgChevronRight, Box } from '../utils/index.js';
+import { SvgChevronRight, Box } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { IconButton } from '../Buttons/index.js';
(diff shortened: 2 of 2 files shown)


LostABike:
@mayank99 can you take a look? I feel like parent's div (`iui-submenu`) moves independently from child `iui-side-navigation-submenu` which is why they looks off-sync. I have tried setting the same animat`on for both like adding `transform: translateX` and `transition: visibility var(--iui-duration-2), transform var(--iui-duration-2);` for both but they are still off-sync :c

mayank99:
> @mayank99 can you take a look? I feel like parent's div (`iui-submenu`) moves independently from child `iui-side-navigation-submenu` which is why they looks off-sync.

to keep the child from moving, you'll need to remove `transform: translateX(-100%)` from the parent and add `justify-content: end` + `justify-items: end`.

LostABike:
> > @mayank99 can you take a look? I feel like parent's div (`iui-submenu`) moves independently from child `iui-side-navigation-submenu` which is why they looks off-sync.
> 
> to keep the child from moving, you'll need to remove `transform: translateX(-100%)` from the parent and add `justify-content: end` + `justify-items: end`.

Wow thanks! That works! But I am so confused. How come in expandable block which I feel was exact logic, worked with `transform: translateX(-100%)` but not here and we need the `justify-content` + `items` here?

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -151,14 +151,14 @@ export const SideNavigation = React.forwardRef((props, forwardedRef) => {
         {expanderPlacement === 'bottom' && ExpandButton}
       </Box>
       {submenu && (
-        <WithCSSTransition
-          in={isSubmenuOpen}
-          dimension='width'
-          timeout={200}
-          classNames='iui'
+        <Box
```
mayank99, line 154: any chance we can avoid the extra div? there was already a wrapping div before, so now there are two.

packages/itwinui-css/backstop/tests/side-navigation.html
mayank99, line 1: why changed all formatting?

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -161,8 +161,36 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   }
 }
 
+//animating technique from https://css-tricks.com/css-grid-can-do-auto-height-transitions/
+.iui-side-navigation-submenu-wrapper {
+  display: grid;
+  grid-template-columns: 0fr;
+  overflow: hidden;
+  z-index: -1;
```
mayank99, line 169: is z-index necessary?

packages/itwinui-css/src/side-navigation/side-navigation.scss
```diff
@@ -161,8 +161,36 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
...
+  justify-content: end;
+  justify-items: end;
```
mayank99, line 171: Answering your earlier question in a thread:

> > to keep the child from moving, you'll need to remove `transform: translateX(-100%)` from the parent and add `justify-content: end` + `justify-items: end`.
>
> Wow thanks! That works! But I am so confused. How come in expandable block which I feel was exact logic, worked with `transform: translateX(-100%)` but not here and we need the `justify-content` + `items` here?

This is very similar to `align-self: end` added in #1396.



Maybe `align-content: end` would have also helped on the parent there. Not sure. And the `transform: translate` part was just to work around the padding issue from https://github.com/iTwin/iTwinUI/pull/1396#discussion_r1258323140.

There is also a difference of horizontal vs vertical, so the code is expected to be different.

I'm also realizing that sidenav has `overflow: hidden`, so this whole grid technique might even be unnecessary and could be replaced with a simple `transform: translateX(-100%)`. See https://web.dev/building-a-sidenav-component/#transforms.

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
LostABike: do you mean the wrapper for both the side nav bar and the submenu?

mayank99: no, there are two divs around the submenu content

mayank99: oh, maybe the wrapper is needed for resizing 🤔 
https://github.com/iTwin/iTwinUI/assets/9084735/6a2f3bf6-26fa-4e22-90c4-cb37c5672a34

LostABike: now that you mention this, I notice it seems to be an issue even for the original version (maybe)
Original version, when you resize then retract the submenu, then expand again, it doesn't keep the previously resized size. Is that intended? Probably can't be done differently without JS
sidenavog

and also an issue for the current solution (with the extra div):
sidenavgrid

which I think is due to grid being fixed after resizing?

mayank99: > Original version, when you resize then retract the submenu, then expand again, it doesn't keep the previously resized size.

imo this is ok. the size is temporary

> and also an issue for the current solution (with the extra div)
> which I think is due to grid being fixed after resizing?

this is not ok. it should not occupy any space after closing. for now, maybe you could use some js to remove the `width` style when closing.

LostABike: is there a way to do that without adding an extra prop to pass `isSubmenuOpen` to submenu?

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -150,16 +151,7 @@ export const SideNavigation = React.forwardRef((props, forwardedRef) => {
...
-        <WithCSSTransition
-          in={isSubmenuOpen}
-          dimension='width'
-          timeout={200}
-          classNames='iui'
-        >
-          {submenu}
-        </WithCSSTransition>
-      )}
+      {submenu ?? submenu}
```
adamhock, line 154: This can be simplified to just `submenu`. The `a ?? b` notation means "if `a` is undefined, call `b`", so it's essentially calling `submenu` whether its `undefined` or not

I noticed there is a unit test that is failing and changing this to `isSubmenuOpen && submenu` seemed to fix this test.

adamhock (commented):
I have no additional comments at this time

PR #1443, opened 2023-07-26
v2: allow `styleType='high-visibility'` in `DropdownButton`

### LINKED ISSUE #1394: [`DropdownButton`] Add support for "high-visibility" styleType
The `<DropdownButton>` component works nicely if adding a `styleType="high-visibility"`, but this isn't supported by the `DropdownButtonProps`.

This works in my source code, but doesn't in code sandbox; unsure why the discrepancy.

**(Local)**
image
High vis dropdown

**(Code sandbox https://codesandbox.io/s/high-visibility-dropdown-button-mxvlcd?file=/src/App.tsx)**
Lo vis dropdown

mayank99:
## Changes

Pretty simple change in types, as all the code is already wired up correctly. Closes #1394

## Testing

https://github.com/iTwin/iTwinUI/assets/9084735/b0903553-0846-4ff1-b578-e403b4a2216c

## Docs

Added changeset.

Not adding story or docs example, because we don't want to encourage this pattern everywhere. I believe we originally added the restriction on purpose, but multiple users have asked for this since then. /cc @FlyersPh9

### DIFF
--- a/packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
+++ b/packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
@@ -97,7 +97,6 @@ export const DropdownButton = React.forwardRef(
             )
           }
           ref={refs}
-          aria-label='Dropdown'
           {...rest}
         >
           {children}
--- a/packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
+++ b/packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
@@ -28,7 +28,7 @@ export type DropdownButtonProps = {
    * Use 'borderless' to hide outline.
    * @default 'default'
    */
-  styleType?: 'default' | 'borderless';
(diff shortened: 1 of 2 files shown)


packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
```diff
@@ -28,7 +28,7 @@ export type DropdownButtonProps = {
    * Use 'borderless' to hide outline.
    * @default 'default'
    */
-  styleType?: 'default' | 'borderless';
+  styleType?: 'default' | 'borderless' | 'high-visibility';
```
gretanausedaite, line 31: Should we add CTA button too?

gretanausedaite (approved):

packages/itwinui-react/src/core/Buttons/DropdownButton/DropdownButton.tsx
mayank99: lets wait until someone asks? imo the CTA button should not be used anywhere

FlyersPh9 (approved):

PR #1303, opened 2023-05-23
Accessible soft background colors

FlyersPh9:
## Changes

- Make soft background color palette change with the theme & meet WCAG AA / AAA scores.  Closes #1293.
- Use these colors in avatar & badge.  Both components have a text color semi-transparent border which helps the component stand out against any background color.
- Add tests on `color.html`.

Screenshot 2023-07-06 at 12 20 24 PM
Screenshot 2023-07-06 at 12 20 35 PM

## Testing

Using `color.html` and `badge.html` to check contrast scores via Chrome dev tools.

## Docs

N/A

PR #1446, opened 2023-07-26
refactor Avatar to be a single element

mayank99:
## Changes

css and dom structure refactor.
```diff
 <span
  class="iui-avatar iui-large"
+ data-iui-status="online"
 >
-  <span class="iui-initials">
    RM
-  </span>
   <img src='...' />
-  <span class='iui-stroke'></span>
-  <span class='iui-status iui-online'></span>
 </span>
```

now uses same contrast technique and colors as Badge. supersedes #1303

## Testing

looks fine in all demos in html and storybook. manually tested contrast ratio against all elevent SoftBackgrounds to confirm all of them are meeting >7:1 ratio.

lots of visual test images changed because we use avatar in many different component demos for some reason.

## Docs

changesets added.


packages/itwinui-css/src/avatar/avatar.scss
mayank99, line 1: @FlyersPh9 thoughts on this refactor? it's literally a single div now and passes contrast requirements without touching color variables (thus working around concerns from https://github.com/iTwin/iTwinUI/pull/1303#issuecomment-1624148478).

only issue i ca see so far is that "offline" dot uses themed colors, whereas i'm hardcoding white. it still passes 3:1 contrast ratio (minimum required for icons), so might be ok





you can preview at https://itwin.github.io/iTwinUI/1446/react/?path=/story/core-avatar

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -3,59 +3,39 @@
...
-
-  > .iui-stroke {
-    display: block;
+  font-weight: var(--iui-font-weight-semibold);
+  text-align: center;
+  text-transform: uppercase;
+  color: #{$avatar-text-color};
+  line-height: var(--_iui-avatar-size);
+  font-size: var(--_iui-avatar-font-size);
+
```
FlyersPh9, line 24: Currently the border gets covered up when an image is displayed.
Screenshot 2023-07-28 at 3 24 10 PM

This can be fixed if we add:
```scss
.iui-avatar {
  isolation: isolate;

  &::before {
    z-index: 1;
  }

  &[data-iui-status] {
    &::after {
      z-index: 2;
    }
  }
}
```

Screenshot 2023-07-28 at 3 27 04 PM

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -3,59 +3,39 @@
...
   display: inline-flex;
   align-items: center;
   justify-content: center;
   position: relative;
-  margin: 0;
-  user-select: none;
   inline-size: var(--_iui-avatar-size);
   block-size: var(--_iui-avatar-size);
-  background-color: hsl(72, 51%, 56%);
+  background-color: var(--iui-color-background-seagreen);
```
FlyersPh9, line 16: I don't see `--iui-color-background-seagreen` defined anywhere.

Since all avatars have a color assigned based off the initials, how about we set the default color to `var(--iui-color-background)` and remove:
```scss
.iui-avatar-count {
}
```

https://github.com/iTwin/iTwinUI/blob/648b4ca22976b0de7bfa892ddcf31dd930f309c2/packages/itwinui-css/src/avatar/avatar.scss#L16C8-L16C8

packages/itwinui-css/src/avatar/avatar.scss
```diff
@@ -3,59 +3,39 @@
 @use '../mixins';
 @use './mixins' as *;
 
+$avatar-text-color: hsl(0 0% 0% / 0.6); // same as badge, meets AA contrast against soft backgrounds
```
FlyersPh9, line 6:
```suggestion
$avatar-text-color: hsl(0 0% 0% / 0.77); // same as badge, meets AA contrast against soft backgrounds
```
If we change the opacity from `0.6` to `0.77` we can pass AAA contrast against all soft backgrounds.  If we do this, I would suggest the same change to `badge.scss`.

packages/itwinui-css/src/avatar/avatar.scss
mayank99: to clarify why the offline icon is not changing: it is using the svg inside a fixed `background-image`. we cannot use `mask` because the status `background-color` is also applied on the same element.

packages/itwinui-css/src/avatar/avatar.scss
mayank99: fixed

packages/itwinui-css/src/avatar/avatar.scss
mayank99: updated in avatar for now. looks nice!

we could make this a shared variable in `@itwin/itwinui-variables`? 🤔

packages/itwinui-css/src/avatar/avatar.scss
FlyersPh9: Ahh right, I forgot about hardcoding the text color so using `--iui-color-background` would not work.  Your change sounds good to me.

packages/itwinui-css/src/avatar/avatar.scss
FlyersPh9: > we could make this a shared variable in `@itwin/itwinui-variables`? 🤔

Since this value is being used in both avatar & badge, yeah, I think that would be a good idea.

FlyersPh9: I think the last thing I was doing on Friday before quitting for the weekend was exploring adding a `$avatar-icon-color` that has a little less contrast than the text but still meeting AAA requirement.  I don't think I landed on a value yet tho.

gretanausedaite (approved):
I don't have any comments, refactored code looks good to me.

adamhock (approved):
LGTM. I would wait for all open conversations to be resolved before merging

packages/itwinui-css/src/avatar/avatar.scss
mayank99: lets tackle this in a separate PR

PR #1448, opened 2023-07-27
Add translate style to every resize corner

gretanausedaite:
## Changes

Added translate style to all resizer options.

### DIFF
--- a/packages/itwinui-react/src/core/utils/components/Resizer.tsx
+++ b/packages/itwinui-react/src/core/utils/components/Resizer.tsx
@@ -143,6 +143,7 @@ export const Resizer = (props: ResizerProps) => {
         }
         case 'right': {
           width = elementRef.current.style.width = `${initialWidth - diffX}px`;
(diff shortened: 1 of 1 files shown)


packages/itwinui-react/src/core/utils/components/Resizer.tsx
```diff
@@ -143,19 +143,23 @@ export const Resizer = (props: ResizerProps) => {
         }
         case 'right': {
           width = elementRef.current.style.width = `${initialWidth - diffX}px`;
+          height = elementRef.current.style.height = `${initialHeight}px`;
+          elementRef.current.style.transform = `translate(${translateX}px, ${translateY}px)`;
```
mayank99, line 147: need to update unit tests for these changes?

mayank99: weird that it was passing before then, but i'm guessing it has something to do with the dialog css present before resizing

mayank99 (approved):
LGTM

PR #1456, opened 2023-07-28
Combobox a11y fix 

xman343:
## Changes

#1148 - Added `aria-label` to `ComboBoxInput` so that the component can meet a11y standards. 

## Testing

A11y tests conducted. 

## Docs

N/A


packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -225,6 +225,7 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
         autoCorrect='off'
         style={multiple ? { paddingLeft: tagContainerWidth + 18 } : {}}
         aria-describedby={multiple ? `${id}-selected-live` : undefined}
+        aria-label='Menu'
```
gretanausedaite, line 228: How did you choose that the best label for this input is 'menu'?

.changeset/four-lobsters-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
```
gretanausedaite, line 2: Does it require major release?

xman343: I initially thought that new changes were to be labeled 'major' due to a major release coming up, but after reading through Semver documentation, it seems that a 'patch' label would be more accurate, since the changes I made are most closely described as a bug fix.

packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
mayank99: Generally speaking, labels should not be hardcoded for any component where we do not know how it will be used.

This combobox could be used for selecting fruits or cities or anything else, so the user is the one who should provide the label.

There is no change needed in the component; only the examples need to be updated. You can see existing examples that are already passing:
- https://github.com/iTwin/iTwinUI/blob/dev/examples/ComboBox.label.tsx#L46
- https://github.com/iTwin/iTwinUI/blob/dev/examples/List.combobox.tsx#L14

xman343: Removed the change from the component and implemented a `<VisuallyHidden>` label in the `MultipleSelect` example. I followed a similar pattern to `List.combobox`.

mayank99: update other combobox examples? i'm sure this is not the only one failing

.changeset/four-lobsters-lay.md
mayank99: changeset no longer needed

mayank99 (approved):
LGTM.

Will need to update docs in a separate PR to emphasize that labels should always be used.

adamhock (approved):

Issue #1444, opened 2023-07-26
Ability to add non-breadcrumb links to a header


ben-polinsky:
#### Feature
The iTwinUI Header allows consumers to add links via a `breadcrumbs` prop. This displays the menu items with carets pointing to the next item to the right, making them feel logically connected. While this suits iTwin applications which display iTwins containing iModels containing versions, it limits apps displaying more traditional headers.

#### Examples
The current Connect Header has header items which do not feel connected in a bread-crumby fashion.
![image](https://github.com/iTwin/iTwinUI/assets/78756012/b0a4af62-8905-4cbd-8991-736cf4a79e2a)

mayank99:
This will likely require breaking changes if we update `Header`. Or we could create a new version from scratch (`HeaderNew`) after releasing v3, and deprecate the old `Header`.

/cc @iTwin/itwinui-1

FlyersPh9:
iTwinUI is now in maintenance mode per [the support policy](https://github.com/iTwin/iTwinUI/wiki/Support-policy), with [StrataKit](https://github.com/iTwin/stratakit) replacing it as the primary design system.

PR #1457, opened 2023-07-28
fix(Header): Added `htmlName` prop to `HeaderButton`

### LINKED ISSUE #971: v3: Header button name & description prop renamed
- Name is a native attribute in HTML.

adamhock:
## Changes
Added `htmlName` prop to the `HeaderButton` subcomponent which handles the native `name` attribute in `<button>`.

Closes #971 

## Testing

Tested by setting `htmlName` prop in website example and looking for the `name` attribute in the resulting HTML

## Docs

Changeset added


mayank99 (approved):
LGTM. I would have like to refactor the Header components at some point (#1444) so that we don't even have a `name` prop, but this is good enough for now.

LostABike (approved):

PR #1458, opened 2023-07-28
Side Navigation customization 

xman343:
## Changes

#1368 - Improving customization of `SideNavigation` component. 

## Testing

Will be tested thru unit & visual testing. 

## Docs

TBD


packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -57,6 +57,22 @@ type SideNavigationProps = {
...
+   */
+  sideNavContentProps?: React.ComponentProps<'div'>;
+  /**
+   * Passes props for SideNav top.
+   */
+  sideNavTopProps?: React.ComponentProps<'div'>;
+  /**
+   * Passes props for SideNav bottom.
+   */
+  sideNavBottomProps?: React.ComponentProps<'div'>;
```
mayank99, line 75: `sideNav` prefix is unnecessary from all cases imo

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -57,6 +57,22 @@ type SideNavigationProps = {
    * @default false
    */
   isSubmenuOpen?: boolean;
+  /**
+   * Passes props for SideNav.
+   */
+  sideNavProps?: React.ComponentProps<'div'>;
```
mayank99, line 63: this could be named `innerProps`? it is an inner-wrapper inside the outer-wrapper

mayank99 (approved):
LGTM. dont forget changeset

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -110,14 +130,28 @@ export const SideNavigation = React.forwardRef((props, forwardedRef) => {
       {...rest}
```
adamhock, line 130: I think the `...rest` props should pass into the `iui-side-navigation` element and then create a `wrapperProps` prop for this element, similar to what was done in the [RadioTile PR](https://github.com/iTwin/iTwinUI/pull/1409/files). Thoughts? @mayank99

mayank99: i'm not sure. In the case `RadioTile`, the `<input>` was the main element, but in this case they are both regular divs (neither is more important than the other).

cc @gretanausedaite

xman343: Would that be within the scope of this PR, or should that be covered by a separate one?

gretanausedaite: If we are keeping current code, I'd like wrapperProps more than innerProps.

> maybe we need to break this down into subcomponents: `SideNavigation.Wrapper` and `SideNavigation`

Is there a use case when these two are not together?

mayank99: > If we are keeping current code, I'd like wrapperProps more than innerProps.

ok, lets go with `wrapperProps` then.

---

> > maybe we need to break this down into subcomponents: `SideNavigation.Wrapper` and `SideNavigation`
> 
> Is there a use case when these two are not together?

i think the user might to omit the wrapper if they already have a parent with `position: relative`. it would be similar to info-panel.

xman343: I've never created + implemented a new subcomponent within a component before, so it's hard for me to gauge if such a change exceeds the scope of this PR.

mayank99: maybe you misread the thread. the consensus is to go with simple `wrapperProps`, without any subcomponents.

xman343: Done

mayank99 (approved):
add changeset and LGTM

xman343:
Reverted the changes that were meant for another PR. Should work now

mayank99 (approved):

PR #1459, opened 2023-07-31
Side navigation submenu animation with simpler transition

LostABike:
## Changes

@mayank99 following your suggestion using only `transformX` and based on https://web.dev/building-a-sidenav-component/#transforms. Also made another PR because this change effectively undo 95% of the [other PR](https://github.com/iTwin/iTwinUI/pull/1439) so might as well.

The `z-index: -1` is needed here because otherwise the submenu will slide on top of side navigation bar during transition

So far so good, only one issue: the parent/wrapper doesn't resize (to only the size of side navigation) when submenu is transited out of view.
Things I have tried include:
- add `flex-shrink: 1` to submenu
- add `justify-content: end` and `justify-items: end` to wrapper (also tried `align-self: end` too)
- testing different values for `block-size` and `width` submenu

Do you know what else could work?

### DIFF
--- a/packages/itwinui-css/src/side-navigation/side-navigation.scss
+++ b/packages/itwinui-css/src/side-navigation/side-navigation.scss
@@ -171,12 +171,22 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   resize: horizontal;
   background-color: var(--iui-color-background);
   border-inline-end: 1px solid var(--iui-color-border);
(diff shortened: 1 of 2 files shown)


mayank99:
> only one issue: the parent/wrapper doesn't resize (to only the size of side navigation) when submenu is transited out of view.

ah yeah, it won't work with this technique, so let's abandon this and move forward with the grid technique from #1439

PR #1460, opened 2023-07-31
scope variables within react stylesheet

mayank99:
## Changes

After #1359, the `styles.css` created within react would cause global side effects, so I've created a simple postcss plugin to dynamically scope the variables to only the hashed iui-root class.

The result:
1. `@itwin/itwinui-variables` is still usable without a class, only data attributes.
2. `@itwin/itwinui-react` is self-contained and no longer affects independent usage of `@itwin/itwinui-variables`.

## Testing

Confirmed working by inspecting generated stylesheet.

## Docs

changeset not needed, as this is a small, almost-invisible change in an unreleased feature.

i will work on docs for variables compatibility table later.


gretanausedaite (approved):
🌵

PR #1467, opened 2023-08-02
use latest `onSlideChange` in Carousel

mayank99:
## Changes

Instead of only using the first reference to the `onSlideChange` prop, the latest version will be used any time it changes. This avoids issues where the stale callback was being invoked instead of the current one.

## Testing

Tested this sandbox locally: https://codesandbox.io/s/itwinui-react-example-carousel-forked-j2fksk?file=/src/App.tsx

clicking next/prev buttons after changing the gradient state logs the correct count now ("3" instead of "10")

https://github.com/iTwin/iTwinUI/assets/9084735/f3f4f68c-ce32-4655-838b-c1b0103f95ca

## Docs

Changeset added.

### DIFF
--- a/packages/itwinui-react/src/core/Carousel/Carousel.tsx
+++ b/packages/itwinui-react/src/core/Carousel/Carousel.tsx
@@ -4,7 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import cx from 'classnames';
-import { getRandomValue, useMergedRefs, useTheme } from '../utils/index.js';
+import {
+  getRandomValue,
+  useLatestRef,
+  useMergedRefs,
+  useTheme,
+} from '../utils/index.js';
 import { CarouselContext } from './CarouselContext.js';
 import { CarouselSlider } from './CarouselSlider.js';
 import { CarouselSlide } from './CarouselSlide.js';
@@ -122,10 +127,10 @@ export const Carousel = Object.assign(
       }
     };
 
-    const userOnSlideChange = React.useRef(onSlideChange);
+    const userOnSlideChange = useLatestRef(onSlideChange);
     React.useEffect(() => {
       userOnSlideChange.current?.(currentIndex);
-    }, [currentIndex]);
+    }, [userOnSlideChange, currentIndex]);
 
     return (
       <section
--- a/.changeset/chatty-actors-attend.md
+++ b/.changeset/chatty-actors-attend.md
@@ -1,5 +1,0 @@
----
-'@itwin/itwinui-react': minor
----
-
-`DropdownButton` now supports `styleType='high-visibility'` to make it blue. This should be used sparingly, as a default or borderless dropdown button is the better choice in most cases.
--- a/.changeset/forty-spoons-relate.md
+++ b/.changeset/forty-spoons-relate.md
@@ -1,5 +1,0 @@
----
-'@itwin/itwinui-react': minor
----
-
-Changed `all: revert` to `all: revert-layer` so that only styles from the v1 layer are reverted, thus avoiding issues with inadvertently removing browser default styles.
--- a/.changeset/gentle-carpets-fetch.md
+++ b/.changeset/gentle-carpets-fetch.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue in Carousel where latest `onSlideChange` callback was being ignored.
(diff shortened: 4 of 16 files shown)


packages/itwinui-react/src/core/Carousel/Carousel.tsx
```diff
@@ -122,10 +127,10 @@ export const Carousel = Object.assign(
       }
     };
 
-    const userOnSlideChange = React.useRef(onSlideChange);
+    const userOnSlideChange = useLatestRef(onSlideChange);
     React.useEffect(() => {
       userOnSlideChange.current?.(currentIndex);
-    }, [currentIndex]);
+    }, [userOnSlideChange, currentIndex]);
```
rohan-kadkol, line 133: Just wondering, why is a ref needed here? Like would something like this work?

```tsx
React.useEffect(() => {
  onSlideChange(currentIndex);
}, [currentIndex]);
```

Or am I missing something? (I tried to test it myself but having some repo setup issues. So thought I'd ask here.)

packages/itwinui-react/src/core/Carousel/Carousel.tsx
```diff
@@ -122,10 +127,10 @@ export const Carousel = Object.assign(
       }
     };
 
-    const userOnSlideChange = React.useRef(onSlideChange);
+    const userOnSlideChange = useLatestRef(onSlideChange);
```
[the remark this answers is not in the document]
rohan-kadkol: >  I'm also wondering why this fixed the error of not calling `onSlideChange()`

Oops, sorry, I think I got my answer from the PR description.

But I am still curious about the difference between `useLatestRef` and a normal `useRef`

packages/itwinui-react/src/core/Carousel/Carousel.tsx
mayank99: the latest ref pattern is needed to avoid stale callbacks (the exact bug we are trying to solve). you can read more here: https://overreacted.io/making-setinterval-declarative-with-react-hooks/

what repo issues are you having?

packages/itwinui-react/src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: take a look inside [`useLatestRef`](https://github.com/iTwin/iTwinUI/blob/dev/packages/itwinui-react/src/core/utils/hooks/useLatestRef.ts), see the link in my [other comment](https://github.com/iTwin/iTwinUI/pull/1467#discussion_r1282312145), and let me know if it still doesn't make sense

packages/itwinui-react/src/core/Carousel/Carousel.tsx
rohan-kadkol: Thanks, I am going through that link. I am getting this error that is not allowing me to open storybook:

image

mayank99: ah, try deleting the `styles.js` folder manually before starting dev server

packages/itwinui-react/src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
rohan-kadkol: Ohh, I think I now understood how `useLatestRef` works to always use the latest callback and not the stale one. Thanks.

packages/itwinui-react/src/core/Carousel/Carousel.tsx
rohan-kadkol: Unfortunately, that didn't work. It is still giving the same error even after deleting the `packages/itwinui-react/src/styles.js` folder.

Am I missing something? Do you recommend recloning the repo?

rohan-kadkol:
To confirm, is it alright for the `onSlideChange()` callback to be called after each button click? I don't see it happening in your video. But it is happening for me when I implement something locally that is similar to the codesandbox code.

[1.webm](https://github.com/iTwin/iTwinUI/assets/45748283/1741b7fe-b00a-42e7-a84a-9de605dcb25e)

mayank99:
> To confirm, is it alright for the `onSlideChange()` callback to be called after each button click? I don't see it happening in your video. But it is happening for me when I implement something locally that is similar to the codesandbox code.

hmm, i don't think `onSlideChange` should be called unless a slide is actually changed, so this sounds like a bug. *however*, this is a completely different issue and no one has complained about it yet; we can look into fixing it at some other point (maybe in v3?)

rohan-kadkol (approved):

PR #1469, opened 2023-08-02
Refactor slider

FlyersPh9:
## Changes

**UNFINISHED**

- Refactor slider so that the track, fill, handle can be used without any of the surrounding wrapper.  When used solo there is no unnecessary space above the slider causing visual misalignment.
- Use `aria-orientation` instead of `.iui-slider-vertical` class modifier.
- What was `.iui-slider-rail` is now a `::before` pseudo element.

## Testing

None yet.

## Docs

None yet.

### DIFF
--- a/packages/itwinui-css/backstop/tests/slider.html
+++ b/packages/itwinui-css/backstop/tests/slider.html
@@ -27,25 +27,26 @@
       @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
       section[id^='demo'] {
         width: auto;
+        display: flex;
+        gap: 2rem;
+      }
+
+      #demo-default,
+      #demo-range {
+        flex-direction: column;
       }
 
       .iui-tooltip {
         display: none;
       }
 
-      .iui-slider-container .iui-slider-thumb:focus + .iui-tooltip,
-      .iui-slider-container .iui-slider-thumb:hover + .iui-tooltip,
-      .iui-slider-container .iui-slider-thumb.iui-active + .iui-tooltip {
+      .iui-slider .iui-slider-thumb:focus + .iui-tooltip,
+      .iui-slider .iui-slider-thumb:hover + .iui-tooltip,
+      .iui-slider .iui-slider-thumb.iui-active + .iui-tooltip {
         position: absolute;
         display: block;
         margin-top: -32px;
         transform: translateX(-50%);
-      }
-
-      #demo-default-vertical,
-      #demo-range-vertical {
-        display: flex;
-        gap: 2rem;
       }
     </style>
 
@@ -63,18 +64,131 @@
 
     <h2>Default</h2>
     <section id="demo-default">
+      <div
+        class="iui-slider iui-progress-slider"
+        aria-orientation="horizontal"
+      >
+        <div
+          class="iui-slider-track"
+          style="right: 50%;"
+        ></div>
+        <div
+          style="left: 50%;"
+          class="iui-slider-thumb"
+          role="slider"
+          tabindex="0"
+          aria-label="Volume Setting"
+          aria-valuemin="0"
+          aria-valuenow="50"
+          aria-valuemax="100"
+          id="test-slider-1"
+        ></div>
+        <output
+          class="iui-tooltip"
+          style="left: 50%;"
+        >
+          50
+        </output>
+      </div>
+
+      <div
+        class="iui-slider iui-progress-slider"
+        aria-orientation="horizontal"
+      >
+        <div
+          class="iui-slider-track"
+          style="right: 50%;"
+        ></div>
+        <div
+          style="left: 50%;"
+          class="iui-slider-thumb"
+          role="slider"
+          tabindex="0"
+          aria-label="Volume Setting"
+          aria-valuemin="0"
+          aria-valuenow="50"
+          aria-valuemax="100"
+          id="test-slider-1"
+        ></div>
+        <output
+          class="iui-tooltip"
+          style="left: 50%;"
+        >
+          50
+        </output>
+        <div class="iui-slider-ticks">
+          <span class="iui-slider-tick">0</span>
+          <span class="iui-slider-tick">10</span>
+          <span class="iui-slider-tick">20</span>
+          <span class="iui-slider-tick">30</span>
+          <span class="iui-slider-tick">40</span>
+          <span class="iui-slider-tick">50</span>
+          <span class="iui-slider-tick">60</span>
+          <span class="iui-slider-tick">70</span>
+          <span class="iui-slider-tick">80</span>
+          <span class="iui-slider-tick">90</span>
+          <span class="iui-slider-tick">100</span>
+        </div>
+      </div>
+
+      <div
+        class="iui-slider iui-progress-slider"
+        aria-orientation="horizontal"
+        aria-disabled="true"
+      >
+        <div
+          class="iui-slider-track"
+          style="right: 50%;"
+        ></div>
+        <div
+          style="left: 50%;"
+          class="iui-slider-thumb"
+          role="slider"
+          tabindex="0"
+          aria-label="Volume Setting"
+          aria-valuemin="0"
+          aria-valuenow="50"
+          aria-valuemax="100"
+          id="test-slider-1"
+        ></div>
+        <output
+          class="iui-tooltip"
+          style="left: 50%;"
+        >
+          50
+        </output>
+        <div class="iui-slider-ticks">
+          <span class="iui-slider-tick">0</span>
+          <span class="iui-slider-tick">10</span>
+          <span class="iui-slider-tick">20</span>
+          <span class="iui-slider-tick">30</span>
+          <span class="iui-slider-tick">40</span>
+          <span class="iui-slider-tick">50</span>
+          <span class="iui-slider-tick">60</span>
+          <span class="iui-slider-tick">70</span>
+          <span class="iui-slider-tick">80</span>
+          <span class="iui-slider-tick">90</span>
+          <span class="iui-slider-tick">100</span>
+        </div>
+      </div>
+
       <div class="iui-input-container">
         <div class="iui-label">Volume</div>
 
-        <div class="iui-slider-component-container iui-slider-horizontal iui-input-group">
+        <div
+          class="iui-slider-container iui-input-group"
+          aria-orientation="horizontal"
+        >
           <span class="iui-slider-min">
             <svg-sound-mute
               class="iui-icon"
               aria-hidden="true"
             ></svg-sound-mute>
           </span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
+          <div
+            class="iui-slider iui-progress-slider"
+            aria-orientation="horizontal"
+          >
             <div
               class="iui-slider-track"
               style="right: 50%;"
@@ -107,15 +221,15 @@
 
       </div>
 
-      <br /><br />
-
       <div class="iui-input-container">
         <div class="iui-label">Slider with ticks</div>
 
-        <div class="iui-slider-component-container iui-slider-horizontal iui-input-group">
+        <div class="iui-slider-container iui-slider-horizontal iui-input-group">
           <span class="iui-slider-min">0</span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
+          <div
+            class="iui-slider iui-progress-slider"
+            aria-orientation="horizontal"
+          >
             <div
               class="iui-slider-track"
               style="right: 50%;"
@@ -138,7 +252,6 @@
               50
             </output>
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -157,15 +270,16 @@
 
       </div>
 
-      <br /><br />
-
       <div class="iui-input-container iui-disabled">
         <div class="iui-label">Disabled slider</div>
 
-        <div class="iui-slider-component-container iui-slider-horizontal iui-input-group iui-disabled">
+        <div class="iui-slider-container iui-slider-horizontal iui-input-group iui-disabled">
           <span class="iui-slider-min">0</span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
+          <div
+            class="iui-slider iui-progress-slider"
+            aria-orientation="horizontal"
+            aria-disabled="true"
+          >
             <div
               class="iui-slider-track"
               style="right: 50%;"
@@ -187,7 +301,6 @@
               50
             </output>
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -207,7 +320,7 @@
       </div>
     </section>
 
-    <br />
+    <!-- <br />
 
     <h2>Ranges</h2>
     <section id="demo-range">
@@ -216,8 +329,7 @@
 
         <div class="iui-slider-component-container iui-slider-horizontal iui-input-group">
           <span class="iui-slider-min">0</span>
-          <div class="
-              iui-slider-container iui-range-slider">
+          <div class="iui-slider-container iui-range-slider">
             <div class="iui-slider-rail"></div>
             <div
               class="iui-slider-track"
@@ -312,7 +424,6 @@
               80
             </output>
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -378,7 +489,6 @@
               80
             </output>
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -396,28 +506,99 @@
         </div>
 
       </div>
-    </section>
+    </section> -->
 
     <br />
 
     <h2>Default vertical</h2>
-    <section id="demo-default-vertical">
+    <section
+      id="demo-default-vertical"
+      style="height: 500px;"
+    >
+      <div
+        class="iui-slider"
+        aria-orientation="vertical"
+      >
+        <div
+          class="iui-slider-track"
+          style="bottom: 50%;"
+        ></div>
+        <div
+          style="bottom: 50%;"
+          class="iui-slider-thumb"
+          role="slider"
+          tabindex="0"
+          aria-label="Volume Setting"
+          aria-valuemin="0"
+          aria-valuenow="50"
+          aria-valuemax="100"
+          id="test-slider-1"
+        ></div>
+        <output
+          class="iui-tooltip"
+          style="bottom: 50%;"
+        >
+          50
+        </output>
+      </div>
+
+      <div
+        class="iui-slider"
+        aria-orientation="vertical"
+      >
+        <div
+          class="iui-slider-track"
+          style="bottom: 50%;"
+        ></div>
+        <div
+          style="bottom: 50%;"
+          class="iui-slider-thumb"
+          role="slider"
+          tabindex="0"
+          aria-label="Volume Setting"
+          aria-valuemin="0"
+          aria-valuenow="50"
+          aria-valuemax="100"
+          id="test-slider-1"
+        ></div>
+        <output
+          class="iui-tooltip"
+          style="bottom: 50%;"
+        >
+          50
+        </output>
+        <div class="iui-slider-ticks">
+          <span class="iui-slider-tick">0</span>
+          <span class="iui-slider-tick">10</span>
+          <span class="iui-slider-tick">20</span>
+          <span class="iui-slider-tick">30</span>
+          <span class="iui-slider-tick">40</span>
+          <span class="iui-slider-tick">50</span>
+          <span class="iui-slider-tick">60</span>
+          <span class="iui-slider-tick">70</span>
+          <span class="iui-slider-tick">80</span>
+          <span class="iui-slider-tick">90</span>
+          <span class="iui-slider-tick">100</span>
+        </div>
+      </div>
+
       <div class="iui-input-container">
         <div class="iui-label">Volume</div>
-
-        <div
-          class="iui-slider-component-container iui-slider-vertical iui-input-group"
-          style="height: 500px;"
-        >
+        <div class="iui-slider-container iui-slider-vertical iui-input-group">
           <span class="iui-slider-min">
             <svg-sound-mute
               class="iui-icon"
               aria-hidden="true"
             ></svg-sound-mute>
           </span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
-            <div class="iui-slider-track"></div>
+          <div
+            class="iui-slider"
+            aria-orientation="vertical"
+          >
+            <div
+              class="iui-slider-track"
+              style="bottom: 50%;"
+            ></div>
             <div
               style="bottom: 50%;"
               class="iui-slider-thumb"
@@ -443,23 +624,17 @@
             ></svg-sound-loud>
           </span>
         </div>
-
       </div>
-
-      <br /><br />
 
       <div class="iui-input-container">
         <div class="iui-label">Slider with ticks</div>
-
-        <div
-          class="iui-slider-component-container iui-slider-vertical iui-input-group"
-          style="height: 500px;"
-        >
+        <div class="iui-slider-container iui-slider-vertical iui-input-group">
           <span class="iui-slider-min">0</span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
-            <div class="iui-slider-track">
-            </div>
+          <div
+            class="iui-slider"
+            aria-orientation="vertical"
+          >
+            <div class="iui-slider-track"></div>
             <div
               class="iui-slider-thumb"
               role="slider"
@@ -469,13 +644,11 @@
               aria-valuenow="50"
               aria-valuemax="100"
               id="test-slider-2"
-            >
-            </div>
+            ></div>
             <output class="iui-tooltip">
               50
             </output>
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -491,19 +664,17 @@
           </div>
           <span class="iui-slider-max">100</span>
         </div>
-
       </div>
 
       <div class="iui-input-container iui-disabled">
         <div class="iui-label">Disabled slider</div>
-
-        <div
-          class="iui-slider-component-container iui-slider-vertical iui-input-group iui-disabled"
-          style="height: 500px;"
-        >
+        <div class="iui-slider-container iui-slider-vertical iui-input-group iui-disabled">
           <span class="iui-slider-min">0</span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
+          <div
+            class="iui-slider"
+            aria-orientation="vertical"
+            aria-disabled="true"
+          >
             <div class="iui-slider-track">
             </div>
             <div
@@ -519,10 +690,7 @@
             <output class="iui-tooltip">
               50
             </output>
-
-
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -541,7 +709,7 @@
       </div>
     </section>
 
-    <br />
+    <!-- <br />
 
     <h2>Ranges Vertical</h2>
     <section id="demo-range-vertical">
@@ -635,7 +803,6 @@
             </output>
 
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -695,7 +862,6 @@
             </output>
 
             <div class="iui-slider-ticks">
-              <!-- You could generate the ticks based on your min, max & step values. -->
               <span class="iui-slider-tick">0</span>
               <span class="iui-slider-tick">10</span>
               <span class="iui-slider-tick">20</span>
@@ -713,7 +879,7 @@
         </div>
 
       </div>
-    </section>
+    </section> -->
 
     <script>
 
@@ -863,7 +1029,8 @@
       }
 
       function getIsVerticalSlider(sliderContainer) {
-        return sliderContainer.parentNode.classList.contains("iui-slider-vertical");
+        // return sliderContainer.parentNode.classList.contains("iui-slider-vertical");
+        return sliderContainer.getAttribute('aria-orientation') === 'vertical';
       }
 
       function initializeThumbProcessing(thumb, sliderContainer, mode) {
@@ -881,7 +1048,7 @@
           thumb.classList.add("iui-active");
           sliderContainer.classList.add("iui-grabbing");
           var ownerDoc = thumb.ownerDocument;
-          var track = sliderContainer.querySelector(".iui-slider-rail");
+          var track = sliderContainer.querySelector(".iui-slider::before");
           var handleMouseMove = function (event) {
             moveSlider(event, thumb, track, progress, mode, toolTip, otherThumb, sliderContainer);
           };
@@ -905,7 +1072,7 @@
       }
 
       function initializeAllProgressSliders() {
-        const allProgressSliderContainers = document.querySelectorAll(".iui-slider-container.iui-progress-slider");
+        const allProgressSliderContainers = document.querySelectorAll(".iui-slider.iui-progress-slider");
         allProgressSliderContainers.forEach(progressSliderContainer => {
           const thumb = progressSliderContainer.querySelector(".iui-slider-thumb");
           initializeThumbProcessing(thumb, progressSliderContainer, "progress");
--- a/packages/itwinui-css/src/slider/slider.scss
+++ b/packages/itwinui-css/src/slider/slider.scss
@@ -2,42 +2,25 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @use '../mixins';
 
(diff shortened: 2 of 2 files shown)


mayank99 (commented):
comments from perfunctory review:
- the track and rail are still using absolute positioning, i think we could instead center these with css grid. the ticks could moved out of `<div class="iui-slider">` to avoid messing up the alignment.
- `aria-orientation` needs to go on the slider thumb, it is invalid on the wrapping div. for styling the wrapping div, a data attribute can be used.
  - same with `aria-disabled`.
- the number of css variables could be reduced probably. not everything needs to be a css variable; (it is most useful for properties that are used by descendants).

other than that, looks pretty ok.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -491,19 +664,17 @@ <h2>Default vertical</h2>
...
       </div>
 
       <div class="iui-input-container iui-disabled">
         <div class="iui-label">Disabled slider</div>
-
-        <div
-          class="iui-slider-component-container iui-slider-vertical iui-input-group iui-disabled"
-          style="height: 500px;"
-        >
+        <div class="iui-slider-container iui-slider-vertical iui-input-group iui-disabled">
```
mayank99, line 671: some of these are messed up (presumably you're aware of it)


if we can avoid using `iui-input-group`, that might help. it was never designed for this purpose.

FlyersPh9:
> * the track and rail are still using absolute positioning, i think we could instead center these with css grid. the ticks could moved out of `<div class="iui-slider">` to avoid messing up the alignment.

Removed `position: absolute;` from everything except for the thumb, using `display: grid;` for most everything, and `.iui-slider-ticks` has been moved out of `.iui-slider`.

> * `aria-orientation` needs to go on the slider thumb, it is invalid on the wrapping div. for styling the wrapping div, a data attribute can be used.
>   
>   * same with `aria-disabled`.

`aria-orientation` & `aria-disabled` are now on `.iui-slider-thumb`, using data-attributes on parent.  `.iui-slider-container` is currently using `:has`, I need to think up a fallback.

> * the number of css variables could be reduced probably. not everything needs to be a css variable; (it is most useful for properties that are used by descendants).

CSS variables have been seriously reduced, although I did not check specificity levels yet.

mayank99:
> `aria-orientation` & `aria-disabled` are now on `.iui-slider-thumb`, using data-attributes on parent. `.iui-slider-container` is currently using `:has`, I need to think up a fallback.

`:has` can be avoided if the data attribute is moved up one level

FlyersPh9:
> > `aria-orientation` & `aria-disabled` are now on `.iui-slider-thumb`, using data-attributes on parent. `.iui-slider-container` is currently using `:has`, I need to think up a fallback.
> 
> `:has` can be avoided if the data attribute is moved up one level

If I move the `data-iui-orientation` up 1 level then to use the slider with no extra features (ticks, min/max value) it will always require `.iui-slider-container`.  The way I have it now you can use `.iui-slider` alone or `.iui-slider-container` to include ticks & min / max value.

I suppose this wouldn't be that bad:
```html
<slider-container data-orientation="horizontal">
	<slider>
</slider-container>
```

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,175 @@
...
-$tick-height: var(--iui-size-s); // 12px
+// #region Slider container
+.iui-slider-container {
+  --_iui-slider-container-gap: var(--iui-size-s);
+  --_iui-slider-rail-color: var(--iui-color-border);
+  --_iui-slider-thumb-border-color: var(--iui-color-border-foreground);
+  --_iui-slider-thumb-color: var(--iui-color-background);
+  --_iui-slider-thumb-cursor: grab;
+  --_iui-slider-thumb-size: var(--iui-size-m);
+  --_iui-slider-ticks-placement: 37px; // TODO: Figure out a calc for this
```
FlyersPh9, line 13: There's got to be some math we can do here to achieve this, I haven't figured it out yet tho.

mayank99: why does it need this strange margin value anyway? any way to avoid it?

FlyersPh9: At first I tried `.iui-slider-container` having 2 rows & 3 columns.
Screenshot 2023-09-07 at 1 12 40 PM

For the horizontal, I thought I could add a negative top margin to the ticks to get them closer to the rail.  That might work on horizontal, but on vertical we have no idea how wide the 1st column will be, so we can't safely use a negative margin.
Screenshot 2023-09-07 at 1 14 47 PM

So then I switched `.iui-slider-container` to have 1 row & 3 columns.
Screenshot 2023-09-07 at 1 06 30 PM

And the ticks are placed in the same grid area as the slider itself.
Screenshot 2023-09-07 at 1 17 06 PM

`37px` gets the ticks aligned, but does introduce empty space at the top.
Screenshot 2023-09-07 at 1 18 36 PM
Screenshot 2023-09-07 at 1 19 50 PM

It even works when `--_iui-slider-thumb-size`'s value is changed.
Screenshot 2023-09-07 at 1 20 32 PM

I just can't seem to figure out why that is the magic value that works.

FlyersPh9: Also considered doing a grid within a grid, but the min / max grid areas would not align with the slider, it would align with the slider + ticks as a whole.
Screenshot 2023-09-07 at 1 38 28 PM

mayank99: i think i have an idea for a slight improvement. can you resolve merge conflicts in your branch and then i can take a look?

FlyersPh9: Merge conflicts are resolved.  I'm very curious what your idea is because I have run out of ideas.

mayank99: instead of the margin, you can use `transform`

horizontal:
```
transform: translateY(50%);
```
vertical:
```
transform: translateX(50%);
```

this is exactly the calculation we want, because without the ticks margins, the grid is as big as the slider and we want to place the ticks right in the center.

FlyersPh9: That did it, great suggestion!

packages/itwinui-css/backstop/scenarios/slider.js
```diff
@@ -6,27 +6,25 @@ const { scenario, hover, focus } = require('../scenarioHelper');
 
 module.exports = [
   // Slider types
-  scenario('Type Default', {
-    selectors: ['#demo-default'],
+  scenario('Type Horizontal', {
```
mayank99, line 9: could you do this change in two commits and link to them so it's easy to compare what changed visually?

right now the diff is unhelpful

packages/itwinui-css/backstop/tests/slider.html
mayank99, line 1: all dom changes need to also be made in react

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,174 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @use '../mixins';
 
-$thumb-height: var(--iui-size-m); // 16px
-$track-height: var(--iui-size-2xs); // 4px
-$tick-height: var(--iui-size-s); // 12px
+// #region Slider container
+.iui-slider-container {
+  --_iui-slider-container-gap: var(--iui-size-s);
```
mayank99, line 7: do we need all these variables or can we reduce some of them?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,174 @@
...
+  --_iui-slider-tick-size: var(--iui-size-s);
+  --_iui-slider-tick-thickness: 1px;
+  --_iui-slider-track-color: var(--iui-color-border-accent);
+  --_iui-slider-track-thickness: calc(var(--_iui-slider-thumb-size) * 0.25);
+
+  display: grid;
+
+  &[data-iui-orientation='horizontal'] {
+    grid-template-columns: auto 1fr auto;
+    grid-template-areas: 'min content max';
```
mayank99, line 22: the grid areas make sense here but if i noticed you are using numbers in some places, particularly in pseudo elements. could you explain all parts of the grid?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,174 @@
...
-  align-items: flex-start;
+    .iui-slider-min,
+    .iui-slider-max {
+      align-items: center;
+    }
 
-  &.iui-disabled {
-    cursor: not-allowed;
+    .iui-slider-min {
+      margin-inline-end: var(--_iui-slider-container-gap);
```
mayank99, line 31: is it possible to use grid gap instead of margin here?

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,174 @@
...
 }
 
 .iui-slider-min,
 .iui-slider-max {
   user-select: all;
   display: flex;
-  justify-content: center;
 
   svg {
     @include mixins.iui-icon-style('m');
```
mayank99, line 162: can we use the `iui-svg-icon` class instead of duplicating the icon styles here?

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: I originally was using `gap`, but if the min / max grid-areas are left empty or removed, the gap is still present.

With `gap`:
Screenshot 2023-09-11 at 4 56 39 PM
Screenshot 2023-09-11 at 4 58 45 PM


Without `gap`:
Screenshot 2023-09-11 at 4 57 49 PM

Maybe that's ok and not a big deal.

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,116 +2,174 @@
...
 
+    .iui-slider::before,
+    .iui-slider-track {
+      block-size: var(--_iui-slider-track-thickness);
+    }
+
+    .iui-slider-track {
+      inset-inline-start: 0;
+      inset-block-start: 50%;
+      transform: translateY(-50%);
```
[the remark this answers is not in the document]
FlyersPh9: It is possible to not use `position: absolute` for the typical slider, but when it came to the range I was unable to figure out an alternative solution.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: Yes, I've made this change.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: I've removed the cursor CSS var.  I can replace some CSS vars with SASS vars, unsure if that would be preferred or not.

gretanausedaite (commented):
No comments on css.
You should rerun visual tests, toast image seems to be broken.

packages/itwinui-css/src/slider/slider.scss
mayank99: CSS variables serve two main purposes:
- from inside: allow changing properties inside nested selectors without repeating those selectors.
- from outside: allow changing properties in inline styles. usually for the parts that need to change dynamically, such as the slider thumb position.

I think there is a readability concern with CSS variables, so it's better to avoid them if all you need is a place to store values. Sass variables are a perfectly fine alternative for storing values.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -893,9 +928,9 @@ <h2>Ranges Vertical</h2>
           thumb.classList.add("iui-active");
           sliderContainer.classList.add("iui-grabbing");
           var ownerDoc = thumb.ownerDocument;
-          var track = sliderContainer.querySelector(".iui-slider-rail");
+          var track = sliderContainer.querySelector(".iui-slider::before");
```
rohan-kadkol, line 931: This can be removed since `track` is no longer used in the code.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -61,674 +61,709 @@ <h1>Slider</h1>
...
-      <div class="iui-input-grid">
-        <div
-          class="iui-input-label"
-          data-iui-disabled="true"
-        >Disabled slider</div>
-
-        <div class="iui-slider-component-container iui-slider-horizontal iui-disabled">
-          <span class="iui-slider-min">0</span>
-          <div class="iui-slider-container iui-progress-slider">
-            <div class="iui-slider-rail"></div>
```
rohan-kadkol, line 171: Jon and I spoke about this before, but just wanted to write a written comment too to confirm with other reviewers. Is it okay to change `iui-slider-rail` from an HTML element to a pseudo element? Because:

* In JS, the `.iui-slider::before` selector does not work since `::before` is not a real dom element.
* In React, I don't think we can add ref, props, etc. to the rail since this is not a real dom element.

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -61,674 +61,709 @@ <h1>Slider</h1>
...
-              class="iui-slider-thumb"
-              role="slider"
-              aria-label="Percent Complete"
-              aria-valuemin="0"
-              aria-valuenow="50"
-              aria-valuemax="100"
-              id="test-slider-2"
-            ></div>
-            <output
-              class="iui-tooltip"
```
rohan-kadkol, line 187: Just curious, is there a reason why you removed most of the tooltips?

packages/itwinui-css/backstop/tests/slider.html
gretanausedaite: I don't think users would want or even should to customise rail + it can be customised by passing a class to wrapping element with ::before selector.

packages/itwinui-css/backstop/tests/slider.html
FlyersPh9: The tooltips were bugging out & when they are used in a real app they'd use iTwinUI-React, so seemed unnecessary here aside from just showing where they would display for implementation details in React.

packages/itwinui-css/backstop/tests/slider.html
FlyersPh9: The only rail customization I've seen so far is what's done by us in the color picker.  It can still be customized via `.myCustomClass::before` and I think that should be fine.

packages/itwinui-css/backstop/tests/slider.html
FlyersPh9: Done.

packages/itwinui-css/backstop/scenarios/slider.js
FlyersPh9: The demos I have currently aren't 1-to-1 with what they were previously.  Previously all demos were wrapped with `.iui-input-container`.  Do you want me to try to match the demos with what they looked like previously?

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -61,674 +61,709 @@ <h1>Slider</h1>
...
-            <div
-              style="left: 20%;"
-              class="iui-slider-thumb"
-              role="slider"
-              tabindex="0"
-              aria-label="Start Value"
-              aria-valuemin="0"
-              aria-valuenow="20"
-              aria-valuemax="79"
-              data-range-max="100"
```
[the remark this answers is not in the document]
FlyersPh9: I don't know if it is still required, but I added it back because I did not intend to remove it.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: I replaced a couple more CSS vars with SASS vars.

packages/itwinui-css/backstop/scenarios/slider.js
mayank99: would it be too much effort to change the demo structure in a separate preceding PR? mainly i would like git to show a side-by-side diff rather than separate deleted+add files

packages/itwinui-css/src/slider/slider.scss
mayank99: `--_iui-slider-thumb-size` can also be replaced?

packages/itwinui-css/backstop/scenarios/slider.js
FlyersPh9: I've made all the tests match how they looked previously & discovered a bug for the vertical slider along the way.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: Swapped that CSS var with a SASS var.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: I typed out a much longer response to this, but it never saved. 😓. 

I have replaced the `grid-row: 1 / 3;` to just use `grid-row-start: 1;` for consistency.  Without using these, you get this: 
Screenshot 2023-09-19 at 5 34 58 PM
image

packages/itwinui-css/src/slider/slider.scss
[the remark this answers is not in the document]
mayank99: slight improvement: could we use a CSS variable to determine how long the track should be? (right now we pass the standard `inset-inline-end` in JS)

same improvement can also be made for the thumb.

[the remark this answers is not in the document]
mayank99: as for improving the vertical centering (which is what the [original comment](https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1321884836) was about), i think you can rely on flexbox.

so the final code would look something like this:
```diff
.iui-slider {
  cursor: pointer;
  isolation: isolate;
  grid-area: content;
  position: relative;
- display: grid;
+ display: flex;
+ align-items: center;
}

.iui-slider-track {
- inset-inline-start: 0;
- inset-block-start: 50%;
- transform: translateY(-50%);
+ inset-inline: var(--iui-slider-track-position); // set from JS
}
```

packages/itwinui-css/src/slider/slider.scss
mayank99: if you make `.iui-slider` a flexbox (see [comment](https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1331921337)), then it should work automatically without needing explicit grid line numbers

packages/itwinui-css/src/slider/slider.scss
mayank99: it should be possible to use `gap` with flexbox because we won't need to define the structure upfront (the children would define the structure, so the gap will only be present if children are present).

```css
.iui-slider-container {
  display: flex;
  align-items: center;

  .iui-slider-min {
    order: -1;
  }

  .iui-slider-min {
    order: 1;
  }

  .iui-slider {
    flex: 1;
  }
}
```

is this doable without breaking any other visuals that rely on the grid?

packages/itwinui-css/src/slider/slider.scss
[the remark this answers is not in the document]
FlyersPh9: I've changed `.iui-slider` to use `display: flex` and removed the `grid` stuff.  I also introduced these 4 variables:
```scss
--iui-slider-track-position-start: 0;
--iui-slider-track-position-end: var(--iui-slider-percentage);
--iui-slider-thumb-position: var(--iui-slider-percentage);
```

For a regular slider, you can do the following:
```html
<div
  class="iui-slider-container"
  style="--iui-slider-percentage: 50%;"
>
```

For a range, you would set the 3 other variables on the element & not use `--iui-slider-percentage`.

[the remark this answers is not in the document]
FlyersPh9: This change did introduce one issue with the opacity slider in the color picker.
Screenshot 2023-09-20 at 4 05 24 PM

The checkered background pattern is currently applied to the `::before` & the transparent to yellow gradient is applied on a color picker specific `::after`.  Replacing `grid` with `flex` makes it look like this:
Screenshot 2023-09-20 at 4 07 22 PM

The quick solution would be to add `position: absolute` to the `::after`.  However, I'm trying to see if we can remove the `::after` all together and instead include the yellow gradient in the checkered gradient.  Below is what I was trying but was unsuccessful:
```scss
background-image:
  repeating-linear-gradient(
    45deg,
    #C7CCD1 25%,
    transparent 25%,
    transparent 75%,
    #C7CCD1 75%,
    #C7CCD1
  ),
  repeating-linear-gradient(
    45deg,
    #C7CCD1 25%,
    #EDEFF2 25%,
    #EDEFF2 75%,
    #C7CCD1 75%,
    #C7CCD1
  ),
  linear-gradient(to right, transparent 0%, var(--iui-color-picker-selected-color) 100%);
```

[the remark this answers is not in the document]
mayank99: That's too many CSS variables. We only need two: `--iui-slider-thumb-position` and `--iui-slider-track-position`.

For supporting the simpler slider case, the default value of `--iui-slider-position` can be set as follows:
```css
.iui-slider-track {
  $default-position: 0 var(--iui-slider-thumb-position);
  inset-inline: var(--iui-slider-track-position, #{$default-position});
}
```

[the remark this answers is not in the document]
mayank99: for fixing the opacity slider, i think you'll need to specify the background-size and background-position for each background-image (in correct order).

```css
.iui-opacity-slider .iui-slider::before {
  content: '';
  forced-color-adjust: none;

  $square-size: var(--iui-size-xs);
  background-position:
    0,
    0,
    calc($square-size * 0.5) calc($square-size * 0.5);
  background-size:
    auto,
    $square-size,
    $square-size;
  background-image:
    linear-gradient(to right, transparent 0%, var(--iui-color-picker-selected-color) 100%),
    repeating-linear-gradient(45deg, #C7CCD1 25%, transparent 25%, transparent 75%, #C7CCD1 75%, #C7CCD1),
    repeating-linear-gradient(45deg, #C7CCD1 25%, #EDEFF2 25%, #EDEFF2 75%, #C7CCD1 75%, #C7CCD1);
}
```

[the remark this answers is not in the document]
FlyersPh9: Ok, I have reduced the CSS vars down to those 2.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: `flex` certainly does make the code more minimal, but the only area I can't figure out a solution for is the ticks.
Screenshot 2023-09-20 at 5 45 09 PM

mayank99: oh yeah, then don't worry about it. i forgot this was one of the main reasons we chose grid (it allows overlapping grid areas)

mayank99: the only way around it i can think is if you move `iui-slider-ticks` inside `iui-slider`.

FlyersPh9: Without testing this, I _predict_ if we did do that then the slider + ticks would vertically center with the min / max instead of only the slider vertically aligning with the min / max.  I mocked-up how I think that might display incorrectly:
Screenshot 2023-09-21 at 2 40 06 PM

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: `.iui-slider` is now using `display: flex`. https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1332116719

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -4,18 +4,22 @@
...
-  background-image: repeating-linear-gradient(
-    45deg,
-    #C7CCD1 25%,
-    transparent 25%,
-    transparent 75%,
-    #C7CCD1 75%,
-    #C7CCD1
-  ),
-    repeating-linear-gradient(45deg, #C7CCD1 25%, #EDEFF2 25%, #EDEFF2 75%, #C7CCD1 75%, #C7CCD1);
+@mixin iui-transparent-background($size: var(--iui-size-m), $variant: swatch) {
```
FlyersPh9, line 7: @mayank99, this was my attempt at getting [your opacity slider gradient suggestion](https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1332203652) to work with this mixin.  This did also remove a `::after` from the swatch.  I would not be surprised if there is a cleaner way to achieve this, but this is what I was able to cook up.

mayank99: why complicate this so much? it's not worth it. just duplicate the code. it's getting duplicated in the final css anyway

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -61,674 +58,643 @@ <h1>Slider</h1>
...
           </div>
-          <span class="iui-slider-max">100</span>
         </div>
-
       </div>
     </section>
 
-    <script>
-
+    <!-- <script>
```
FlyersPh9, line 697: @rohan-kadkol, could you update this script once more?  Currently it sets `right` or `left` when a handle is dragged.  For the slider, can it instead set [the variables mentioned here](https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1332161230)?  I think the slider & range will have different implementations.  @mayank99, is it even worth it having the sliders work in the HTML/CSS demo page?  Maybe we just remove all of this JS?

packages/itwinui-css/backstop/tests/slider.html
FlyersPh9: @siddhantrawal, I believe the HTML structure & CSS class names are all pretty finalized at this point, all that remains are some minor details that I don't believe should effect the React implementation.  Can you branch this PR and start the React implementation?

packages/itwinui-css/backstop/tests/slider.html
mayank99: lets update it to use the CSS variables for now, since it's fairly simple.

in the future, i have ideas for bigger changes

packages/itwinui-css/src/color-picker/color-picker.scss
FlyersPh9: I've removed the mixin now.

packages/itwinui-css/backstop/tests/slider.html
rohan-kadkol: > @rohan-kadkol, could you update this script once more?

https://github.com/iTwin/iTwinUI/pull/1469/commits/dd9e702d01973a32385dc46c3fa00594c085f50a Sounds good, I made the change in the script to use `--iui-slider-thumb-position` and `--iui-slider-track-position`.

packages/itwinui-css/backstop/tests/color-picker.html
```diff
@@ -219,11 +219,13 @@ <h2>Advanced</h2>
             ></div>
           </div>
 
-          <div class="iui-slider-component-container iui-slider-horizontal iui-hue-slider">
-            <div class="iui-slider-container iui-progress-slider">
-              <div class="iui-slider-rail"></div>
+          <div
+            class="iui-slider-container iui-hue-slider"
+            data-iui-orientation="horizontal"
+            style="--iui-slider-thumb-position: 16%;"
```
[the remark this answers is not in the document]
mayank99: for single thumb slider, it is sufficient to set `--iui-slider-thumb-position` on the container. no `--iui-slider-track-position` needed

for multi thumb sliders, we need to set `--iui-slider-thumb-position` for each thumb, and `--iui-slider-track-position` for each track.

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,251 +2,255 @@
...
+      inset-inline: var(--iui-slider-track-position, #{$slider-position-horizontal});
+    }
+
+    .iui-slider-thumb {
+      inset-inline-start: var(--iui-slider-thumb-position);
+      transform: translateX(-50%);
+    }
+  }
+
+  &[data-iui-orientation='vertical'] {
```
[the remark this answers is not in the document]
mayank99: if the container has a height, it might be enough?

we could also provide a `min-block-size` of a few pixels just to avoid collapsing like that

packages/itwinui-css/backstop/tests/slider.html
```diff
@@ -61,674 +58,643 @@ <h1>Slider</h1>
...
-              class="iui-slider-track"
-              style="right: 50%;"
-            ></div>
+          <span class="iui-slider-max">
+            <svg-sound-loud
+              class="iui-svg-icon"
+              data-iui-icon-size="m"
+              aria-hidden="true"
+            ></svg-sound-loud>
+          </span>
```
rohan-kadkol, line 83: nit: Since on the React side, we are using the order of min-slider-max (https://github.com/iTwin/iTwinUI/pull/1606#discussion_r1346083116), maybe we can do the same here too.

gretanausedaite (approved):
No comments. LTGM 🌵

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,251 +2,256 @@
...
+      block-size: $iui-slider-tick-thickness;
+
+      &::before {
+        inline-size: $iui-slider-tick-size;
+        block-size: $iui-slider-tick-thickness;
+      }
+    }
+
+    .iui-slider {
+      inline-size: calc($iui-slider-thumb-size + 2px);
```
rohan-kadkol, line 116: Just wondering, what is the `2px` for?

packages/itwinui-css/backstop/tests/slider.html
rohan-kadkol, line 1: I remember that Jon had an example with custom min/max labels. Do we still want to add that? @FlyersPh9

packages/itwinui-css/src/slider/slider.scss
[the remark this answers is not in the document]
mayank99: yeah, just do `min-block-size: var(--iui-size-xl)` or something. 300px is too much

apps/storybook/cypress-visual-screenshots/baseline/Slider.test.ts-Vertical.png
rohan-kadkol, line 1: I think the test images need to be re-approved. They would have changed with the recent code changes.

packages/itwinui-css/src/slider/slider.scss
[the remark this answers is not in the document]
mayank99: no, the purpose of this min height is only to show that the slider *exists*. the actual height still needs to be controlled by the user of the component (either through explicitly setting height/min-height or through the use of a flex/grid container).

you could maybe make it `var(--iui-size-2xl)` but i'm afraid that any larger would get in the way if someone wants a small slider.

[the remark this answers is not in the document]
mayank99: the story should be updated to use a realistic height, similar to before.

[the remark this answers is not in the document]
mayank99: you shouldn't need to reach for `railContainerProps`. you can use `display: grid` on the decorator that already has height (near the top of the file)

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -112,10 +112,11 @@ $iui-slider-tick-thickness: 1px;
       }
     }
 
+    // Added min-block-size to show that slider exists.
```
rohan-kadkol, line 115: nit: Would be better if you move it as an inline comment on L119.

rohan-kadkol (approved):
I tried reviewing `slider.scss` and `color-picker.scss` the best I could. But since those files are restructured significantly, it is likely that I missed something. So better to wait for more approvals for a more thorough review.

Giving a pre-emptive approval since I have no comments other than the ones already opened.

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,24 +492,19 @@ export const Slider = React.forwardRef((props, ref) => {
           {minValueLabel}
         </Box>
       )}
+      {tickMarkArea}
       <Box
-        as='div'
         ref={containerRef}
         {...railContainerProps}
```
mayank99, line 498: since the "rail" element was removed, we should rename `railContainerProps` to `trackContainerProps` or something

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,24 +492,19 @@ export const Slider = React.forwardRef((props, ref) => {
           {minValueLabel}
         </Box>
       )}
+      {tickMarkArea}
```
mayank99, line 495: was this intentionally moved here or can it be restored to the original position?

mayank99: there is a stacking issue where the ticks are appearing over the slider.

can be fixed using:
```scss
.iui-slider-container {
  // ...
  isolation: isolate;
}

.iui-slider-ticks {
  // ...
  z-index: -1;
}
```

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,251 +2,258 @@
...
+}
+
+.iui-slider-max {
+  grid-area: max;
+}
+
+.iui-slider-ticks {
+  position: relative;
+  display: flex;
+  pointer-events: none;
```
mayank99, line 174:
```suggestion
```

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,251 +2,258 @@
...
+  justify-content: space-between;
+  user-select: none;
```
mayank99, line 176:
```suggestion
```

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -2,251 +2,258 @@
...
+  z-index: -1;
+}
 
-  svg {
-    @include mixins.iui-icon-style('m');
-    @include mixins.iui-icon-color;
+.iui-slider-tick {
+  position: relative;
+  display: flex;
+  pointer-events: none;
```
mayank99, line 184:
```suggestion
```

mayank99 (commented):
not sure why we are disabling pointer-events and user-select on ticks. lets remove that

mayank99 (approved):
LGTM

packages/itwinui-react/src/core/Slider/Slider.tsx
rohan-kadkol: nit: I believe now the order of the `Slider`'s children ( `iui-slider-min`, `iui-slider`, `tickMarkArea`, `iui-slider-max` ) is different from the children in the slider html demo ( `iui-slider-min`, `tickMarkArea`, `iui-slider`, `iui-slider-max` ).

But this small mismatch can be ignored.

rohan-kadkol (approved):

packages/itwinui-css/backstop/tests/slider.html
FlyersPh9: Let's merge PR as is & I can submit a new PR with demo test changes.  I would typically change the tests as the last commit but running low on time, I don't want to hold this up.

packages/itwinui-css/src/slider/slider.scss
FlyersPh9: `$iui-slider-thumb-size: var(--iui-size-m);` & size m = 16px.

If I inspect the thumb, I see 18x18:
Screenshot 2023-10-16 at 2 23 19 PM

So I think it must be the size of the thumb + the border thickness.

Maybe it should be this instead?
```scss
$iui-slider-thumb-border-thickness: 1px;

.iui-slider {
  inline-size: calc($iui-slider-thumb-size + ($iui-slider-thumb-border-thickness * 2));
}
```

PR #1470, opened 2023-08-02
ColorPicker a11y fixes 

xman343:
## Changes

#1148  - Editing `ColorPicker` examples so that they meet a11y requirements.

## Testing

a11y testing will be conducted.

## Docs

N/A


packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
```diff
@@ -140,6 +140,8 @@ export const ColorPalette = React.forwardRef((props, ref) => {
     >
       {label && <Box className='iui-color-picker-section-label'>{label}</Box>}
       <Box
+        role='listbox'
```
mayank99, line 143: how did you decide on using `role=listbox` and `role=option`?

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
```diff
@@ -140,6 +140,8 @@ export const ColorPalette = React.forwardRef((props, ref) => {
...
+        aria-label='Color palette'
```
mayank99, line 144: there is already a label on line 141, we should try to use that (through `id` + `aria-labelledby`) and only fall-back to `aria-label` if there is no label passed.

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
```diff
@@ -305,6 +307,7 @@ export const ColorBuilder = React.forwardRef((props, ref) => {
 
       {showAlpha && (
         <Slider
+          aria-label='Opacity slider'
```
mayank99, line 310: `aria-label` does nothing on the Slider. it is getting applied on the outer Slider container, which is a generic div.

packages/itwinui-react/src/core/Slider/Thumb.tsx
```diff
@@ -154,6 +154,7 @@ export const Thumb = (props: ThumbProps) => {
       {...tooltipProps}
     >
       <Box
+        aria-labelledby='Slider'
```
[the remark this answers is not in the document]
xman343: It needed a label to meet a11y requirements, but the user is unable to give it one outside of the component level.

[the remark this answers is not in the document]
rohan-kadkol: Should this accessible name include the purpose? Like "Choose a color", or "Slider for alpha", etc. instead of just what it is (Slider)?

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
rohan-kadkol: I'm a little unsure by what do you mean here. But if you intend to pass an `aria-label` to the slider, you would need to pass it to the thumb props. This is because the thumbs are the actual input tags that need to have an accessible name.

https://github.com/iTwin/iTwinUI/blob/bd12e4917550aa39d4144d41a72ffcd6c3d53242/examples/Slider.tooltipnone.tsx#L30-L34

(from PR #1474)

packages/itwinui-react/src/core/Slider/Thumb.tsx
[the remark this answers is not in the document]
rohan-kadkol: Ohh, this is actually inside the Slider code. I thought this was in the ColorPicker code.

So you can remove this and pass the aria-label to the thumb using thumbProps, similar to https://github.com/iTwin/iTwinUI/pull/1470/files#r1287227854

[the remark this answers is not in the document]
mayank99: there should be no changes inside Slider in this PR. `thumbProps` will let you keep all the changes inside ColorPicker

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
mayank99: That makes sense, but previously there was no role, so I'm asking why/how you decided to add one. Which specific a11y violation is this addressing? Would have been nice if you made the PR description more... descriptive.

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
```diff
@@ -287,6 +287,8 @@ export const ColorBuilder = React.forwardRef((props, ref) => {
       </Box>
 
       <Slider
+        aria-label='Hue slider'
+        id='Slider'
```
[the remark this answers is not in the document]
rohan-kadkol: Do you need to give an `aria-label` to the Slider div? Instead, I believe you can pass an `aria-label` to the thumbs directly (https://github.com/iTwin/iTwinUI/pull/1470/files#r1287238103). Then there is no need for an id.

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: It specifically addresses the `aria-input`fieldname` rule, which was violated in MainExample and AdvancedExample, because the `Sliders` needed an accessible name. 
aria-input-field-name 1
aria-input-field-name 2

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
```diff
@@ -148,8 +150,11 @@ export const ColorPalette = React.forwardRef((props, ref) => {
         {colors &&
           colors.map((_color, index) => {
             const color = getColorValue(_color);
+            const swatchLabel = getColorValue(color).toHslString(true);
             return (
               <ColorSwatch
+                role='option'
+                aria-label={swatchLabel}
```
rohan-kadkol, line 157: Should we have a default `aria-label` that the `ColorSwatch` adds? Like `swatchLabel` could be calculated inside `ColorSwatch`. If the user provided an `aria-label`, we use it. But if not, we use swatchLabel as the `aria-label`.

examples/ColorPicker.advancedPopover.tsx
```diff
@@ -69,7 +69,10 @@ export default () => {
           visible={isOpen}
           placement='bottom-start'
         >
-          <IconButton onClick={() => setIsOpen((open) => !open)}>
+          <IconButton
+            label='Color Picker toggle'
```
rohan-kadkol, line 73: Would a better label be "Show/Hide Color Picker" or "Color Picker Toggle Visibility"? Because just "toggle" could also mean toggle the *color format (e.g. hsla, hex)* and not the *visibility* of the color picker

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
```diff
@@ -148,8 +150,11 @@ export const ColorPalette = React.forwardRef((props, ref) => {
         {colors &&
           colors.map((_color, index) => {
             const color = getColorValue(_color);
+            const swatchLabel = getColorValue(color).toHslString(true);
```
rohan-kadkol, line 153: The [NVDA screen reader](https://www.nvaccess.org/download/) reads the `hsla` of `hsla(0, 100%, 50%, 1)` as one word ("his-lea") instead of as four individual letters ("ache-es-el-ea"). Same is the case for `rgba`. But for hex colors, it works alright.

Capitalizing the "hsla"/"rgba" reads it correctly in the screen reader. So I think you should capitalize the string.

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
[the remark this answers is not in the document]
xman343: I'll follow Rohan's suggestion.

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: I'd like to implement this, although it might be a bit too involved of a refactor for the scope of this PR. I'll see what I can do.

rohan-kadkol: Actually, it is very small change, sorry if I didn't explain it clearly in my comment.

I meant like instead of calculating `swatchLabel` on [L153](https://github.com/iTwin/iTwinUI/blob/8449063c254110e011897ff724bf5854a5507fd6/packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx#L153), you can do it in `ColorSwatch.tsx` instead. Because you are passing the color to `ColorSwatch` on [L159](https://github.com/iTwin/iTwinUI/blob/8449063c254110e011897ff724bf5854a5507fd6/packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx#L159).

Then in `ColorSwatch.tsx`, you can do something like this:
image

rohan-kadkol: Ohh, actually `colorString` already calculates the string representation of the color. You could try reusing that instead of `swatchLabel`.

examples/ColorPicker.advancedPopover.tsx
xman343: Fixed!

mayank99: can we use more natural language?

i.e. "Toggle color picker visibility"

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
mayank99: this looks unrelated, `aria-input-field-name` is only violated for the sliders, whereas i'm asking about the changes in ColorPalette

xman343: That's totally my bad. Lot of threads to keep track of! 

The `listbox` and `option` roles were added to the `ColorPalette` and the `ColorSwatches`, respectively, because without those roles, it violated the 'aria-allowed-attribute' rule. This is because the Color Swatch component originally had the `aria-selected` attribute, despite the component not being given a role. 
a11y-allowed-attr

Other roles that can take `aria-selected ` - Tab, Row, and Grid - either weren't appropriate for `ColorPalette` & `ColorSwatch`, or (in the case of Grid) would have required a more extensive refactoring of the code than I figured would be necessary. (More info [here](https://www.digitala11y.com/aria-selected-state/).)

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: Added!

mayank99: Please avoid hardcoding ids inside components (in this case `'wrapper'`). `id`s are meant to be globally unique. Hardcoding a static id will result in every instance of this component getting the same clashing id.

You can generate a unique id using the `useId` utility (see example in `LabeledInput.tsx`)

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: Now that I think about it, I wonder if I can try to pass the names of each color in `ColorList` as labels for the swatches...

mayank99: `colorString` is the closest thing to the "source" value, so i would be fine with just `colorString.toUpperCase()`. hex codes are ok because this is literally a color-picker, in which technical accuracy is important. 

also the developer always has the ability to override the aria-label

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
xman343: Implemented with [this ](https://github.com/iTwin/iTwinUI/pull/1470/commits/630a2a2f2693c658dd0cbaeda68cc3fe1b5686a5)commit from yesterday.

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: Done!

mayank99: the `useId` import can be combined with the others near line 11

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: Switched from hsl to hex

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
```diff
@@ -45,7 +45,8 @@
  * </ColorPalette>
  */
 export const ColorPalette = React.forwardRef((props, ref) => {
-  const { colors, label, className, children, ...rest } = props;
+  const uid = useId();
```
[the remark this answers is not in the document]
mayank99: might be able to avoid this weird warning by using a different name than `uid`

or even better, remove the variable name altogether

```diff
- const uid = useId();
- const { colors, label, className, children, id = uid, ...rest } = props;
+ const { colors, label, className, children, id = useId(), ...rest } = props;
```

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
mayank99: please read the comment again, the suggestion is not to switch from hsl to hex, but to use `colorString` which uses the source string (can be hex or hsl or whatever string the color was instantiated with)

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
mayank99: and restore the blank line

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
```diff
@@ -287,6 +287,11 @@ export const ColorBuilder = React.forwardRef((props, ref) => {
       </Box>
 
       <Slider
+        thumbProps={() => {
+          return {
+            'aria-label': `Hue slider`,
```
rohan-kadkol, line 292: You can remove the *"slider"* because "slider" is anyways announced by screen readers (at least the NVDA screen reader). So it would read out an extra "slider" ("Hue slider slider")

I also saw [this example](https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-color-viewer/) on w3.org to confirm this (see the `aria-labeledby` on any one of the example color sliders).

```suggestion
            'aria-label': `Hue`,
```

packages/itwinui-react/src/core/ColorPicker/ColorBuilder.tsx
```diff
@@ -305,6 +310,11 @@ export const ColorBuilder = React.forwardRef((props, ref) => {
 
       {showAlpha && (
         <Slider
+          thumbProps={() => {
+            return {
+              'aria-label': `Opacity slider`,
```
rohan-kadkol, line 315: You can remove the "slider" from the `aria-label`

```suggestion
              'aria-label': `Opacity`,
```

examples/ColorPicker.advancedPopover.tsx
```diff
@@ -69,7 +69,10 @@ export default () => {
           visible={isOpen}
           placement='bottom-start'
         >
-          <IconButton onClick={() => setIsOpen((open) => !open)}>
+          <IconButton
```
rohan-kadkol, line 72: nit: On the below `<Button>` on L82, you can add an `aria-label` with the capitalized color string (similar to https://github.com/iTwin/iTwinUI/pull/1470#discussion_r1287280570)

mayank99: that button does not need an aria-label, it already has text content

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: I had tried using `colorString` for the swatch label already, but I couldn't figure out how to use it in `ColorPalette.tsx`. When I try using `colorString` in `ColorSwatches.tsx`, its original scope, `ColorPickerBasicExample` throws an `aria-allowed-attribute` error for the selected color swatch on top.
image
This is because that swatch doesn't have a role, since the roles are distributed in `ColorPalette.tsx`. I could try giving that swatch its own role in the `examples` file, but it's hard for me to tell what role is appropriate. I could also continue to assign the swatch labels the way I have been, even if it is redundant.

mayank99: should be fixable if you use visually-hidden text instead of aria-label

```jsx
<Box
  className={cx('iui-color-swatch', { 'iui-active': isActive }, className)}
  // ...
  {...rest}
>
  <VisuallyHidden>{colorString.toUpperCase()}</VisuallyHidden>
</Box>
```

examples/ColorPicker.advancedPopover.tsx
mayank99: then the displayed value should be fixed, instead of slapping an aria-label on it.

further reading (if interested):
- https://ericwbailey.design/published/aria-label-is-a-code-smell/
- https://benmyers.dev/blog/overriding-screenreader-pronunciations/

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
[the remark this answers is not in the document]
xman343: Done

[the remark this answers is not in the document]
mayank99: actually that breaks rule of hooks, so should be reverted. try just using a different variable name, e.g. `stableId` instead of `uid`.

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: done

examples/ColorPicker.advancedPopover.tsx
rohan-kadkol: Thanks for the articles. Upon reading it, I am inclined towards leaving it as it is.

Because when users hear "his-lea", they can guess that it means "ache-es-el-ea". By adding an `aria-label` to clear this ambiguity, we might be making the braille display harder to read. And also it may lead to errors where we change the text content but forget to update the `aria-label`.

packages/itwinui-react/src/core/ColorPicker/ColorPalette.tsx
xman343: Done. It works!

PR #1474, opened 2023-08-07
Fixed a11y violations for Slider examples

rohan-kadkol:
## Changes

Added `aria-label` prop to all thumbs in all `Slider` examples. But no change was needed in component code.

## Testing

Ensured that the cypress tests for Slider reported no more violations.

## Docs

N/A

### DIFF
--- a/examples/Slider.labels.tsx
+++ b/examples/Slider.labels.tsx
@@ -19,11 +19,7 @@ export default () => {
         Choose a happiness level
       </Text>
       <Slider
-        thumbProps={() => {
-          return {
-            'aria-labelledby': labelId,
-          };
-        }}
+        thumbProps={() => ({ 'aria-labelledby': labelId })}
         values={[50]}
         minLabel={<SvgSmileySadVery />}
         maxLabel={<SvgSmileyHappyVery />}
--- a/examples/Slider.labels.tsx
+++ b/examples/Slider.labels.tsx
@@ -15,7 +15,9 @@ export default () => {
 
   return (
     <div style={{ width: 'min(100%, 300px)' }}>
-      <Label id={labelId}>Choose a happiness level</Label>
+      <Label id={labelId} as='div'>
+        Choose a happiness level
+      </Label>
       <Slider
         thumbProps={() => ({ 'aria-labelledby': labelId })}
         values={[50]}
--- a/examples/Slider.labels.tsx
+++ b/examples/Slider.labels.tsx
@@ -13,6 +13,11 @@ export default () => {
   return (
     <div style={{ width: 'min(100%, 300px)' }}>
       <Slider
+        thumbProps={() => {
+          return {
+            'aria-label': `Choose a value`,
+          };
+        }}
         values={[50]}
         minLabel={<SvgSmileySadVery />}
         maxLabel={<SvgSmileyHappyVery />}
--- a/examples/Slider.labels.tsx
+++ b/examples/Slider.labels.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Slider, Text } from '@itwin/itwinui-react';
+import { Slider, Label } from '@itwin/itwinui-react';
 
 import {
   SvgSmileyHappyVery,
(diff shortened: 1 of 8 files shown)


mayank99 (commented):
can we try to use a visible label in most examples? will likely need `aria-labelledby` instead of `aria-label`

rohan-kadkol:
> can we try to use a visible label in most examples? will likely need `aria-labelledby` instead of `aria-label`

I adding `<Text as="label">...</Text>` to most examples for `<Slider>`. I linked the slider to the labels using `aria-labeledby` for these cases, and `aria-label` when not using the explicit labels. (https://github.com/iTwin/iTwinUI/pull/1474/commits/0b444e0afb722016df7d45f157def5951ffd93e7)

examples/Slider.labels.tsx
```diff
@@ -3,19 +3,25 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Slider } from '@itwin/itwinui-react';
+import { Slider, Text } from '@itwin/itwinui-react';
+import { useId } from '../packages/itwinui-react/src/core/utils/hooks/index.js';
```
mayank99: use `React.useId` instead of this internal hook

examples/Slider.labels.tsx
```diff
@@ -3,19 +3,25 @@
...
   SvgSmileyHappyVery,
   SvgSmileySadVery,
 } from '@itwin/itwinui-icons-react';
 
 export default () => {
+  const labelId = useId();
+
   return (
     <div style={{ width: 'min(100%, 300px)' }}>
+      <Text id={labelId} as='label' variant='leading'>
```
mayank99, line 18: `<Label>` styling should work better than `<Text variant='leading'>`.

also not sure if we should use `as='label'` or `as='div'`.

examples/Slider.labels.tsx
```diff
@@ -3,19 +3,25 @@
...
+        Choose a happiness level
+      </Text>
       <Slider
         thumbProps={() => {
           return {
-            'aria-label': `Choose a rating`,
+            'aria-labelledby': labelId,
```
mayank99, line 24: nit: the `return` can be condensed using parentheses

```
thumbProps={() => ({
  'aria-labelledby': labelId
})}
```

rohan-kadkol: Ohh, right, changed for all Slider examples with 1-2 keys in the `thumbProps` object (https://github.com/iTwin/iTwinUI/pull/1474/commits/65f28ded5083045369be3ac44799fc272593722f)

examples/Slider.tooltipnone.tsx
```diff
@@ -26,7 +28,14 @@ export default () => {
 
   return (
     <div style={{ width: 'min(100%, 300px)' }}>
+      <Text id={labelId} as='label' variant='leading'>
+        Choose a start date
+      </Text>
       <Slider
+        thumbProps={() => ({
+          'aria-labelledby': labelId,
+          'aria-valuetext': dateFormatter.format(currentValue.date),
```
mayank99, line 37: i feel like `aria-valuetext` should not need to be set by users. why was this needed?

rohan-kadkol: This was not needed (did not cause any a11y violation). But since the screen reader would just read out the date number (e.g. `35`), I thought of adding `aria-valuetext` with the formatted date string (e.g. something like `4 Feb`)

examples/Slider.labels.tsx
rohan-kadkol: > `<Label>` styling should work better than `<Text variant='leading'>`.

https://github.com/iTwin/iTwinUI/pull/1474/commits/0eb23657b5e8e0d71dc948c55cc879621677db8e: Ohh, nice. Yes, `<Label>` styling looks great. It also now looks more similar to the label in LabeledSelect.

> also not sure if we should use `as='label'` or `as='div'`.

https://github.com/iTwin/iTwinUI/pull/1474/commits/03ac73f11df68437e7a7cdf8f20b1ee6248596e8: Since the slider is actually a `<div role="slider">` and not an `<input type="range">`, I changed the `as` of `<Label>` to be a `div` and not a `label`. I found [this example on the MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role#examples) to confirm this pattern of:

| Label   | Slider|
|----------|-------------|
| `<label for="sliderId">` |  `<input id="sliderId" type="range">`  |
| **`<div id="labelId">`** | **`<div role="slider">`** (our iTwinUI Slider) |

mayank99 (approved):

PR #1476, opened 2023-08-08
bump `@cypress/request` again

mayank99:
## Changes

fixes https://github.com/advisories/GHSA-p8p7-x288-28g6.

## Testing

n/a

## Docs

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,8 @@
     "yaml": "^2.2.2",
     "semver": "^7.5.2",
     "optionator": "~0.9.3",
-    "typescript": "~5.1.6"
+    "typescript": "~5.1.6",
+    "@cypress/request": "^3.0.0"
   },
   "engines": {
     "npm": ">=7.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,7 +75,8 @@
     "yaml": "^2.2.2",
     "semver": "^7.5.2",
     "optionator": "~0.9.3",
-    "typescript": "~5.1.6"
+    "typescript": "~5.1.6",
+    "@cypress/request": "^3.0.0"
```
rohan-kadkol, line 79: A little confused, the PR title says *bump*, but here you are *adding a new* dep of `@cypress/request`.

If this package was not used before this PR, why was `yarn audit` failing?

rohan-kadkol, line 79: Where is `@cypress/request` needed/used? I could not find it with the VSCode keyword search.

Is it used as a dependency of some other package and not in our code? If so, by explicitly stating a version here, are you forcing to use the latest version? Or should we use yarn's [`resolutions`](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-how-to-use-it)?

mayank99: you need to look in the lockfile, it was previously at 2.88.12

mayank99: you can do a find in the lockfile or by running `npm ls @cypress/request`.

and yes i am forcing the latest version using `resolutions`, because `^2.88.11` is not compatible with `3.0.0`

rohan-kadkol (approved):

PR #1477, opened 2023-08-08
add `portalContainer` prop to ThemeProvider

mayank99:
## Changes

added new `portalContainer` prop to ThemeProvider.
- if passed, Toaster will be portaled into it, and no `<div>` will be rendered at the end of ThemeProvider.
- gets stored in ThemeContext, so children can access it and portal into it.
- it is user's responsibility to set theme, maybe through ThemeProvider.
- when the prop is not passed, old behavior will continue to work (`<div>` is rendered and its ref is stored in ThemeContext)

## Testing

tested with a few different components in playground (Tooltip, Dialog, Toast, Select). Here's a simple test case:

<details>
<summary>App.tsx</summary>

```tsx
import {
  ThemeProvider,
  Dialog,
  Button,
  Tooltip,
  Select,

  const [myPortals, setMyPortals] = React.useState<HTMLElement>();
  const [isDialogOpen, setIsDialogOpen] = React.useState(false);

  return (
    <>
      <ThemeProvider portalContainer={myPortals}>
        <Button onClick={() => setIsDialogOpen(true)}>Open</Button>

        <Dialog portal isOpen={isDialogOpen} isDismissible>
          <Dialog.Backdrop />
          <Dialog.Main>
            <Dialog.TitleBar>My portaled dialog</Dialog.TitleBar>
            <Dialog.Content>
              Lorem ipsum dolor sit amet consectetur adipisicing elit. Modi ex
              ea sapiente, culpa aliquam consequatur perferendis eum odit cum
              harum.
              <Select
                options={[
                  { value: 1, label: 'Option 1' },
                  { value: 2, label: 'Option 2' },
                  { value: 3, label: 'Option 3' },
                ]}
              />
              <Tooltip content='this will close the dialog'>
                <Button onClick={() => setIsDialogOpen(false)}>Close</Button>
              </Tooltip>
            </Dialog.Content>
          </Dialog.Main>
        </Dialog>
      </ThemeProvider>

      {ReactDOM.createPortal(
        <ThemeProvider className='my-portals' ref={setMyPortals}>
          <></>
        </ThemeProvider>,
        document.body,
      )}
    </>
  );
}
```

</details>

everything correctly portals into the `.my-portals` div:

https://github.com/iTwin/iTwinUI/assets/9084735/a6fff557-99d3-4cd8-b3ab-d7144ad8fbdc

also added unit test.

## Docs

added changeset and jsdocs.

### DIFF
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeContext.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeContext.tsx
@@ -9,7 +9,7 @@ export const ThemeContext = React.createContext<
   | {
       theme?: ThemeType;
       themeOptions?: ThemeOptions;
-      portalContainerRef?: React.RefObject<HTMLElement>;
+      portalContainer?: HTMLElement | null;
     }
   | undefined
 >(undefined);
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -7,6 +7,7 @@ import { render, screen } from '@testing-library/react';
 import * as UseMediaQuery from '../utils/hooks/useMediaQuery.js';
 
 import { ThemeProvider } from './ThemeProvider.js';
+import { Dialog } from '../Dialog/index.js';
 
 let useMediaSpy: jest.SpyInstance;
 
@@ -178,4 +179,23 @@ it('should inherit theme from data attribute if no context found', () => {
 it('should default to light theme if no parent theme found', () => {
   render(<ThemeProvider>Test</ThemeProvider>);
   expect(screen.getByText('Test')).toHaveAttribute('data-iui-theme', 'light');
+});
+
+it('should respect the portalContainer prop', async () => {
+  const myPortals = document.createElement('my-portals');
+  document.body.appendChild(myPortals);
+
+  const { getByText } = render(
+    <ThemeProvider portalContainer={myPortals}>
+      <div>hello</div>
+
+      <Dialog portal={{ to: myPortals }} isOpen>
+        a dialog
+      </Dialog>
+    </ThemeProvider>,
+  );
+
+  getByText('hello');
+  expect(document.querySelector('my-portals .iui-toast-wrapper')).toBeTruthy();
+  expect(document.querySelector('my-portals')).toHaveTextContent('a dialog');
 });
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -3,11 +3,12 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import { render, screen } from '@testing-library/react';
 import * as UseMediaQuery from '../utils/hooks/useMediaQuery.js';
 
 import { ThemeProvider } from './ThemeProvider.js';
-import { Dialog } from '../Dialog/index.js';
+import { ThemeContext } from './ThemeContext.js';
 
 let useMediaSpy: jest.SpyInstance;
 
@@ -185,17 +186,22 @@ it('should respect the portalContainer prop', async () => {
   const myPortals = document.createElement('my-portals');
   document.body.appendChild(myPortals);
 
+  const InPortal = () => {
+    const { portalContainer } = React.useContext(ThemeContext) || {};
+    return (
+      portalContainer &&
+      ReactDOM.createPortal(<div>in portal!</div>, portalContainer)
+    );
+  };
+
   const { getByText } = render(
     <ThemeProvider portalContainer={myPortals}>
-      <div>hello</div>
-
-      <Dialog portal={{ to: myPortals }} isOpen>
-        a dialog
-      </Dialog>
+      <div>not in portal</div>
+      <InPortal />
     </ThemeProvider>,
   );
 
-  getByText('hello');
+  getByText('not in portal');
   expect(document.querySelector('my-portals .iui-toast-wrapper')).toBeTruthy();
-  expect(document.querySelector('my-portals')).toHaveTextContent('a dialog');
+  expect(document.querySelector('my-portals')).toHaveTextContent('in portal!');
 });
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import * as UseMediaQuery from '../utils/hooks/useMediaQuery.js';
 
 import { ThemeProvider } from './ThemeProvider.js';
@@ -18,13 +18,11 @@ beforeEach(() => {
 
 afterEach(() => useMediaSpy.mockRestore());
 
-it.each(['default', 'light', 'dark'] as const)(
+it.each(['light', 'dark'] as const)(
   'should render correctly with %s theme',
   (theme) => {
     const { container } = render(
-      <ThemeProvider theme={theme === 'default' ? undefined : theme}>
-        Test
-      </ThemeProvider>,
+      <ThemeProvider theme={theme}>Test</ThemeProvider>,
     );
     const element = container.querySelector('div');
     expect(element).toBeTruthy();
@@ -142,6 +140,7 @@ it('should apply iui-root-background to the topmost ThemeProvider', () => {
 it('should default applyBackground to false when inheriting theme', () => {
   const { container, rerender } = render(
     <ThemeProvider theme='inherit'>Hello</ThemeProvider>,
+    { wrapper: ({ children }) => <div data-iui-theme='light'>{children}</div> },
   );
   const element = container.querySelector('.iui-root');
   expect(element).not.toHaveClass('iui-root-background');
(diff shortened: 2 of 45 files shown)


packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -96,19 +115,23 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
     theme: themeProp = 'inherit',
     children,
     themeOptions,
+    portalContainer: portalContainerProp,
```
mayank99, line 118: i also thought about using the parent portalContainer but i'm not sure becaude it really complicates the default behavior.

gretanausedaite (approved):

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -96,19 +115,23 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
...
     ...rest
   } = props;
 
-  const portalContainerRef = React.useRef<HTMLDivElement>(null);
+  const [portalContainerState, setPortalContainerState] = React.useState(
+    portalContainerProp || null,
+  );
+  const portalContainer = portalContainerProp || portalContainerState;
```
rohan-kadkol, line 125: What is the need for `portalContainer`? I am a bit confused why `portalContainerState` is not enough.

packages/itwinui-react/src/core/ThemeProvider/ThemeContext.tsx
```diff
@@ -9,7 +9,7 @@ export const ThemeContext = React.createContext<
   | {
       theme?: ThemeType;
       themeOptions?: ThemeOptions;
-      portalContainerRef?: React.RefObject<HTMLElement>;
```
rohan-kadkol, line 12: Just wondering, why was `portalContainer` needed when `portalContainerRef` already exists? I guess something along the lines of when the `portalContainerRef` that was passed was changed by the user's code, the `portalContainerRef` was not updated inside our code?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -179,3 +181,27 @@ it('should default to light theme if no parent theme found', () => {
...
+  };
+
+  const { getByText } = render(
+    <ThemeProvider portalContainer={myPortals}>
+      <div>not in portal</div>
+      <InPortal />
+    </ThemeProvider>,
+  );
+
+  getByText('not in portal');
```
rohan-kadkol, line 204: In mulitple `.test.tsx` files, I am noticing this pattern of `getByText(...)`'s return value not being stored in a variable. Is this just a way to make sure that there is a node with textContent of `"not in portal"` but not do anything with it?

Similarly, is L204 the shorthand for:
```tsx
expect(getByText('not in portal').toBeTruthy())
```

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -122,9 +145,14 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
       >
         <ToastProvider>
           {children}
-          <div ref={portalContainerRef}>
-            <Toaster />
-          </div>
+
+          {portalContainerProp ? (
+            ReactDOM.createPortal(<Toaster />, portalContainerProp)
+          ) : (
```
rohan-kadkol, line 151: When portaling a component, do we not also need to portal its Provider?

i.e. something like:

```tsx
{portalContainerProp ? (
  ReactDOM.createPortal(
    <ToastProvider>
      {children}
      <Toaster />
    </ToastProvider>, 
    portalContainerProp)
) : (
  <ToastProvider>
    {children}
    <div ref={setPortalContainerState}>
      <Toaster />
    </div>
  </ToastProvider>
)
}
```

In other words, can Toaster (or any other component for that matter) access its associated Provider if it has been portaled in such a way where the Provider is not an ancestor of that component?

packages/itwinui-react/src/core/utils/components/Popover.tsx
```diff
@@ -64,8 +64,7 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
   const computedProps: Partial<TippyProps> = {
     allowHTML: true,
     animation: false,
-    appendTo: (el) =>
-      themeInfo?.portalContainerRef?.current || el.ownerDocument.body,
+    appendTo: (el) => themeInfo?.portalContainer || el.ownerDocument.body,
```
rohan-kadkol, line 67: I'm just wondering, is using `||` instead of `??` here and in all other places a personal preference or is there some benefit?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: We want the prop to take preference. The state should only be used when the prop is not passed.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
mayank99: yeah pretty much. it will throw an error if the text is not found

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: the provider gets automatically connected. see the second paragraph here: https://react.dev/reference/react-dom/createPortal#reference

packages/itwinui-react/src/core/utils/components/Popover.tsx
mayank99: it shouldn't matter in most cases, except `??` is for when you specifically only want to guard against `undefined` and `null`, whereas `||` will also consider booleans and other falsy values like empty strings

packages/itwinui-react/src/core/ThemeProvider/ThemeContext.tsx
rohan-kadkol: I meant to ask that why was `portalContainerRef` replaced by `portalContainer` ([diff](https://github.com/iTwin/iTwinUI/pull/1477/files#diff-653d15c515af2a52d6613822762b8a6f494de559d9c9b9c0f7833d66f35e972dL12-R12))? Like what were the problems with `portalContainerRef` that prompted this change?

mayank99: did you see where it's used? it's no longer a ref

rohan-kadkol: iiuc, the main change is that instead of using a ref (`portalContainerRef`), you are now using the actual Element (`portalContainer`). So I am wonder, can the same result behavior of this PR be achieved by reusing `portalContainerRef`?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -96,19 +115,23 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
     theme: themeProp = 'inherit',
     children,
     themeOptions,
+    portalContainer: portalContainerProp,
     ...rest
   } = props;
 
-  const portalContainerRef = React.useRef<HTMLDivElement>(null);
+  const [portalContainerState, setPortalContainerState] = React.useState(
+    portalContainerProp || null,
```
rohan-kadkol, line 123: Is this initialization necessary? Is it not okay if it remains null/undefined when a `portalContainerProp` is passed. Because anyways portalContainer is what is really being passed to the context.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: i've created a new `useUncontrolledState` hook to make this less confusing

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: i think you're right in this case, but in other future cases of `useUncontrolledState`, it could be useful. plus it ensures the correct types are used for the return values of `useState`

https://github.com/iTwin/iTwinUI/blob/05af7c3dd536f090aafeb808f5423c0f1a556ea6/packages/itwinui-react/src/core/ExpandableBlock/ExpandableBlock.tsx#L81-L82

packages/itwinui-react/src/core/utils/hooks/useUncontrolledState.ts
```diff
@@ -0,0 +1,16 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+
+/**
+ * Wrapper over `useState` that always gives preference to the
+ * uncontrolled state (which often comes from a prop).
+ */
+export const useUncontrolledState = <T>(uncontrolledState: T) => {
+  const [controlledState, setControlledState] =
+    React.useState(uncontrolledState);
```
rohan-kadkol, line 13: Do we need to initialize `controlledState` on L13? Because:
* If `uncontrolledState` is passed, then we don't use `controlledState` (L15), so no benefit of initializing it.
* If `uncontrolledState` is not passed, we are initializing controlledState with `undefined`

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
rohan-kadkol: Superseded by https://github.com/iTwin/iTwinUI/pull/1477/commits/512da81a36f108e886528ecb3160f6a0a85d400e#r1291543460

packages/itwinui-react/src/core/ThemeProvider/ThemeContext.tsx
rohan-kadkol: Sounds good, then the current implementation using `useState` works well.

packages/itwinui-react/src/core/utils/hooks/useUncontrolledState.ts
mayank99: ok, removed [`3cc41d6`](https://github.com/iTwin/iTwinUI/pull/1477/commits/3cc41d64cf8dbe5eae2f63eb9e651210b052d8eb)

PR #1479, opened 2023-08-08
Fixed a11y violations for Select examples

rohan-kadkol:
## Changes

* Added `<Label>` for most `Select` examples
* Either added `aria-labeledby` when `<Label>` exists, else added `aria-label`
* There were `color-contrast` violations for the placeholder text. We chose to ignore those since:
  * Placeholder text disappears after typing a value. So, the `color-contrast` on the placeholder is not too important.
  * We prefer `<label>`s over placeholders. And the `<label>`s pass the `color-contrast`

Previously reported violations for Select:

image

## Testing

* Confirmed that the NVDA screen reader's callouts for the Select examples were acceptable

## Docs

N/A

### DIFF
--- a/examples/Select.disable.tsx
+++ b/examples/Select.disable.tsx
@@ -3,35 +3,28 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Select, Label } from '@itwin/itwinui-react';
+import { LabeledSelect } from '@itwin/itwinui-react';
 
 export default () => {
-  const labelId = React.useId();
-
   return (
-    <div>
-      <Label id={labelId}>Disabled Select Label</Label>
-      <Select<number>
-        triggerProps={{
-          'aria-labelledby': labelId,
-        }}
-        disabled
-        options={[
-          {
-            value: 1,
-            label: 'Item #1',
-          },
-          {
-            value: 2,
-            label: 'Item #2',
-          },
-          {
-            value: 3,
-            label: 'Item #3',
-          },
-        ]}
-        placeholder={'Placeholder text'}
-      />
-    </div>
+    <LabeledSelect<number>
+      disabled
+      label={'Disabled Select Label'}
+      options={[
+        {
+          value: 1,
+          label: 'Item #1',
+        },
+        {
+          value: 2,
+          label: 'Item #2',
+        },
+        {
+          value: 3,
+          label: 'Item #3',
+        },
+      ]}
+      placeholder={'Placeholder text'}
+    />
   );
 };
--- a/examples/Select.disable.tsx
+++ b/examples/Select.disable.tsx
@@ -3,27 +3,35 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Select } from '@itwin/itwinui-react';
+import { Select, Label } from '@itwin/itwinui-react';
 
 export default () => {
+  const labelId = React.useId();
+
   return (
-    <Select<number>
-      disabled
-      options={[
-        {
-          value: 1,
-          label: 'Item #1',
-        },
-        {
-          value: 2,
-          label: 'Item #2',
-        },
-        {
-          value: 3,
-          label: 'Item #3',
-        },
-      ]}
-      placeholder={'Placeholder text'}
-    />
+    <div>
+      <Label id={labelId}>Disabled Select Label</Label>
+      <Select<number>
+        triggerProps={{
+          'aria-labelledby': labelId,
+        }}
+        disabled
+        options={[
+          {
+            value: 1,
+            label: 'Item #1',
+          },
+          {
+            value: 2,
+            label: 'Item #2',
+          },
+          {
+            value: 3,
+            label: 'Item #3',
+          },
+        ]}
+        placeholder={'Placeholder text'}
+      />
+    </div>
   );
 };
--- a/examples/Select.disable.tsx
+++ b/examples/Select.disable.tsx
@@ -7,7 +7,7 @@ import { LabeledSelect } from '@itwin/itwinui-react';
 
 export default () => {
   return (
-    <LabeledSelect<number>
+    <LabeledSelect
       disabled
       label={'Disabled Select Label'}
       options={[
--- a/testing/a11y/src/Component.cy.tsx
+++ b/testing/a11y/src/Component.cy.tsx
@@ -6,46 +6,33 @@ import * as React from 'react';
 import * as allExamples from 'examples';
 import { ThemeProvider } from '@itwin/itwinui-react';
 
+type commonRulesKeys = 'ignore_color_contrast_for_select_placeholder';
+
+const commonRules: Record<commonRulesKeys, Record<string, any>> = {
+  ignore_color_contrast_for_select_placeholder: {
+    id: 'color-contrast',
+    enabled: true,
+    selector: ':not(._iui3-placeholder > span._iui3-content)',
+  },
+};
+
+const skipRules = {
+  SelectIconExample: {
+    rules: [commonRules.ignore_color_contrast_for_select_placeholder],
+  },
+  SelectMainExample: {
+    rules: [commonRules.ignore_color_contrast_for_select_placeholder],
+  },
+  SelectStatusesExample: {
+    rules: [commonRules.ignore_color_contrast_for_select_placeholder],
+  },
+  SelectSublabelsExample: {
+    rules: [commonRules.ignore_color_contrast_for_select_placeholder],
+  },
+};
+
 describe('Should have no WCAG violations', () => {
   Object.entries(allExamples).forEach(([name, Component]) => {
-    const testsToTestFor = [
-      'SelectDisableExample',
-      'SelectIconExample',
-      'SelectMainExample',
-      'SelectStatusesExample',
-      'SelectSublabelsExample',
-      'SelectTruncateExample',
-    ];
-
-    type commonRulesKeys = 'ignore_color_contrast_for_select_placeholder';
-
-    const commonRules: Record<commonRulesKeys, Record<string, any>> = {
-      ignore_color_contrast_for_select_placeholder: {
-        id: 'color-contrast',
-        enabled: true,
-        selector: ':not(._iui3-placeholder > span._iui3-content)',
-      },
-    };
-
-    const skipRules = {
-      SelectIconExample: {
-        rules: [commonRules.ignore_color_contrast_for_select_placeholder],
-      },
-      SelectMainExample: {
-        rules: [commonRules.ignore_color_contrast_for_select_placeholder],
-      },
-      SelectStatusesExample: {
-        rules: [commonRules.ignore_color_contrast_for_select_placeholder],
-      },
-      SelectSublabelsExample: {
-        rules: [commonRules.ignore_color_contrast_for_select_placeholder],
-      },
-    };
-
-    if (!testsToTestFor.includes(name)) {
-      return;
-    }
-
     it(name, () => {
       cy.mount(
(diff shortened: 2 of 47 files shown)


testing/a11y/src/Component.cy.tsx
```diff
@@ -6,6 +6,34 @@ import * as React from 'react';
...
 
+// Axe config rules docs: https://www.deque.com/axe/core-documentation/api-documentation/#api-name-axeconfigure
+// Axe rules that apply to multiple examples
+type commonRulesKeys = 'ignore_color_contrast_for_select_placeholder';
+const commonRules: Record<commonRulesKeys, Record<string, any>> = {
+  ignore_color_contrast_for_select_placeholder: {
+    id: 'color-contrast',
+    enabled: true,
+    selector: ':not(._iui3-placeholder > span._iui3-content)',
+  },
```
mayank99, line 17: oh this is too much code and too hacky. now i'm thinking we should just fix the contrast.

my only concern is there will not be enough distinction between placeholder text and regular text.

cc @FlyersPh9 for opinions on visuals

rohan-kadkol: > oh this is too much code and too hacky. now i'm thinking we should just fix the contrast.

This PR was WIP. The PR was going to be ready for review in a few minutes, during which I was going to leave a PR comment saying I am open to suggestions to improve this code, if needed. But you were quicker than me! 😅

> my only concern is there will not be enough distinction between placeholder text and regular text.

I intentionally added very selective `selector` to only ignore color contrast only for the placeholder text and not any other text.

mayank99: > > my only concern is there will not be enough distinction between placeholder text and regular text.
> 
> I intentionally added a very selective `selector` to only ignore `color-contrast` only for the placeholder text and not any other text.

no i'm saying if we increase contrast for the placeholder, then it will look very similar to the regular text (for the end user)

testing/a11y/src/Component.cy.tsx
```diff
@@ -6,6 +6,34 @@ import * as React from 'react';
 import * as allExamples from 'examples';
 import { ThemeProvider } from '@itwin/itwinui-react';
 
+// Axe config rules docs: https://www.deque.com/axe/core-documentation/api-documentation/#api-name-axeconfigure
+// Axe rules that apply to multiple examples
+type commonRulesKeys = 'ignore_color_contrast_for_select_placeholder';
```
rohan-kadkol, line 11: I added this union `type` to store all the keys for `commonRules` so that if we do `commonRules.key_does_not_exist`, it gives a TS error during development time, instead of returning `undefined` at runtime. Is there a better way to do this?

testing/a11y/src/Component.cy.tsx
FlyersPh9: > oh this is too much code and too hacky. now i'm thinking we should just fix the contrast.

This is the code to ignore the color contrast so that it can pass our GitHub checks?  It might be a lot and hacky, but I'd value the end result more over the test code being clean.

> my only concern is there will not be enough distinction between placeholder text and regular text.

Definitely a concern.  I tried changing the color from `--iui-color-text-disabled` to `--iui-color-text-muted` which does pass contrast scores with a 7 in light theme & 4.62 in dark theme, but is _very_ easy to mistake for regular text.

**Current implementation:**
Screenshot 2023-08-10 at 2 53 54 PM

**Using muted text color:**
Screenshot 2023-08-10 at 2 52 17 PM

**Actual text:**
Screenshot 2023-08-10 at 2 52 33 PM

We could consider adding a new text color variable specifically for placeholder text, but the dark disabled text color is already really close to the 4.5 minimum requirement.

mayank99: for the placeholder, could we try a color that is the bare minimum (4.5)? and if that's still to close to regular text, can we try bumping the contrast of regular text?

rohan-kadkol: > This is the code to ignore the color contrast so that it can pass our GitHub checks?

Yes, this code is to ignore the reported `color-contrast` violation for just the placeholder texts of only the examples for `Select`

> We could consider adding a new text color variable specifically for placeholder text, but the dark disabled text color is already really close to the 4.5 minimum requirement.

Just including the exact contrast ratio for our reference. The contrast ratio for the placeholder text is `2.93`.

image

FlyersPh9: How about this?  I didn't try high contrast yet.

`light.scss`
```diff
- --iui-color-text-disabled: #{hex($fg-700)};
+ --iui-color-text-disabled: #{hex($fg-600)}; 
```

`dark.scss`
```diff
- --iui-color-text-muted: #{hex($fg-500)};
+ --iui-color-text-muted: #{hex($fg-400)};
- --iui-color-text-disabled: #{hex($fg-700)};
+ --iui-color-text-disabled: #{hex($fg-500)};
```

**Before:**
Screenshot 2023-08-10 at 3 39 08 PM Screenshot 2023-08-10 at 3 28 22 PM

<table>
<tr>
 <td>Variable
 <td>Light
 <td>Dark
<tr>
 <td>Text
 <td>15.52
 <td>8.13
<tr>
 <td>Muted
 <td>7
 <td>4.62
<tr>
 <td>Disabled
 <td><strong>4.11</strong>
 <td><strong>2.93</strong>
</table>

**After:**
Screenshot 2023-08-10 at 3 40 34 PM Screenshot 2023-08-10 at 3 27 28 PM

<table>
<tr>
 <td>Variable
 <td>Light
 <td>Dark
<tr>
 <td>Text
 <td>15.52
 <td>8.13
<tr>
 <td>Muted
 <td>7
 <td>5.68
<tr>
 <td>Disabled
 <td>5.32
 <td>4.62
</table>

mayank99: > ```diff
> - --iui-color-text-muted: #{hex($fg-500)};
> + --iui-color-text-muted: #{hex($fg-400)};
> ```

isn't that bringing it even closer to regular text?

also would need to see in the context of a select. hard to tell from just looking at colors.

> didn't try high contrast yet.

just remember it's ok if placeholder doesn't pass 7:1 because it still meets minimum requirements (which is all that's needed from a legal perspective)

> 4.11 ⚠️ | 2.93 ⚠️

also remember disabled elements are exempt from contrast requirements

rohan-kadkol: > How about this?

@FlyersPh9 Thanks for the detailed info! It's nice that the contrast ratio minimum of 4.5 is met. However, I find very little distinction between (normal & muted), and even lesser between (muted & disabled) with the new colors.

> just remember it's ok if placeholder doesn't pass 7:1 because it still meets minimum requirements (which is all that's needed from a legal perspective)

@mayank99, to confirm, is the legal minimum `4.5` for AA? If so, are we are not clearing the minimum for AA? Because the contrast ratios are `4.11` and `2.93`?

mayank99: yes https://webaim.org/articles/contrast/#sc143

rohan-kadkol: > @mayank99, to confirm, is the legal minimum `4.5` for AA? 

> yes https://webaim.org/articles/contrast/#sc143

Thanks for the link.

>  If so, are we are not clearing the minimum for AA? Because the contrast ratios are `4.11` and `2.93`?

Ohh, I meant to say that you had said that we *are* meeting the minimum requirements (I assume AA (4.5:1) is the legal minimum) for the placeholder. So, it is okay if we don't meet the more ambitious AAA requirement (7:1). 

> just remember it's ok if placeholder doesn't pass 7:1 because it still meets minimum requirements (which is all that's needed from a legal perspective)

But since the old contrast ratios for disabled are `4.11` and `2.93`, the placeholder is *not* meeting the minimum AA (4.5:1) legal requirement.

Or maybe I am misunderstanding something?

FlyersPh9: > > ```diff
> > - --iui-color-text-muted: #{hex($fg-500)};
> > + --iui-color-text-muted: #{hex($fg-400)};
> > ```
> isn't that bringing it even closer to regular text?

Yes, this makes muted closer to regular.  But muted had to change because disabled is now equal to what muted was.

> i am also confused about which color we're using (disabled vs muted) for placeholder text.

Placeholder text is currently using `--iui-color-text-disabled`.

> > to confirm, is the legal minimum `4.5` for AA? If so, are we are not clearing the minimum for AA? Because the contrast ratios are `4.11` and `2.93`?
> yes https://webaim.org/articles/contrast/#sc143

I swear that we looked this up and placeholder text did not need to meet 4.5, but as I'm googling this now [it appears that placeholder text DOES need to meet 4.5](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html#:~:text=including%20placeholder%20text).

mayank99: placeholder needs to meet 4.5, disabled/inactive doesn't.

the 4.5 ratio is the only requirement because level AA is what is used in the laws worldwide. AAA is nice to have, but is not a legal requirement anywhere in the world afaict

rohan-kadkol: In our case, since we are using the disabled colors for the placeholder, looks like we still have to meet this 4.5 contrast ratio requirement. Because the placeholder/Select is still *user interactable*. ([link](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html))

image

So we may need to create that new variable, or modify the existing colors to meet that requirement.

mayank99: yeah, that's why i've been trying to suggest using a completely new color specifically for placeholder.

FlyersPh9: Take a look at #1485.

examples/Select.disable.tsx
```diff
@@ -3,27 +3,35 @@
...
-        },
-        {
-          value: 3,
-          label: 'Item #3',
-        },
-      ]}
-      placeholder={'Placeholder text'}
-    />
+    <div>
+      <Label id={labelId}>Disabled Select Label</Label>
```
mayank99, line 13: can we instead show `LabeledSelect` in the examples? it handles everything (generally, it's better if the user doesn't need to worry about aria)

rohan-kadkol: Ohh, right, changed `<Label>` + `<Select>` examples to use `<LabeledSelect>` https://github.com/iTwin/iTwinUI/pull/1479/commits/44d332b9744bfb97d0ebf36683dd43eaa46ac103

mayank99: while you're changing the examples, could you also remove the unnecessary generics? e.g. `<Select<number>>` -> `<Select>`

testing/a11y/src/Component.cy.tsx
mayank99, line 1: i'd say undo all changes in this file and we can merge this PR.

contrast will be fixed in #1485 or a subsequent PR

examples/Select.disable.tsx
rohan-kadkol: Sounds good, removed (https://github.com/iTwin/iTwinUI/pull/1479/commits/9bb9de411d115b32d0e9ee19922ed35098a98def)

mayank99 (approved):
LGTM

testing/a11y/src/Component.cy.tsx
rohan-kadkol: Sounds good, removed all the changes in this file. (https://github.com/iTwin/iTwinUI/pull/1479/commits/521d4db40a680a8bf69654267c26aa3c96b2d292)

After #1485 adds `--iui-color-text-placeholder`, I will create a new PR to have `Select` use that variable for the placeholder text.

mayank99:
CSS tests have been super flaky lately. https://github.com/iTwin/iTwinUI/actions/runs/5857407299/job/15879220154?pr=1479

Will need to investigate separately, but for now I am merging this.

PR #1480, opened 2023-08-09
fix z-index issues in demos

mayank99:
## Changes

changed `portalContainer` of all examples to `<body>`. fixes some website demos where content was getting clipped because of z-index.

## Testing

before:

image

image

after:

image

image

## Docs

n/a

### DIFF
--- a/examples/index.tsx
+++ b/examples/index.tsx
@@ -6,8 +6,15 @@ import * as React from 'react';
 import { ThemeProvider } from '@itwin/itwinui-react';
 
 const withThemeProvider = (Component: () => React.ReactElement) => () => {
+  const [portalContainer, setPortalContainer] = React.useState<HTMLElement>();
+  React.useEffect(() => void setPortalContainer(document.body), []);
+
   return (
-    <ThemeProvider theme='dark' themeOptions={{ applyBackground: false }}>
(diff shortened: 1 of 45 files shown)


gretanausedaite (approved):

examples/index.tsx
```diff
@@ -6,8 +6,15 @@ import * as React from 'react';
 import { ThemeProvider } from '@itwin/itwinui-react';
 
 const withThemeProvider = (Component: () => React.ReactElement) => () => {
+  const [portalContainer, setPortalContainer] = React.useState<HTMLElement>();
+  React.useEffect(() => void setPortalContainer(document.body), []);
```
rohan-kadkol, line 10: Is there a reason to not initialize the `portalContainer` itself? Like why not pass `document.body` to `useState()` on L9?

Does this have something to do with `document.body` not being available before `React.useEffect()` can be called? Like some SSR related quirk?

rohan-kadkol, line 10: Why is `void` needed?

mayank99: yes, `document` only works on the client

mayank99: the `void` is to ensure that if there is ever a return a value, it will be discarded.

PR #1482, opened 2023-08-10
Fixed a11y violations for InputGroup examples

rohan-kadkol:
## Changes

* Changed the svg labels to the text labels in the `<Radio>`

Previously reported violations for InputGroup:

image

## Testing

Confirmed that the NVDA screen reader's callouts for the failing `InputGroupRadioGroupExample` was acceptable.

## Docs

N/A


examples/InputGroup.radioGroup.tsx
```diff
@@ -9,8 +9,16 @@ export default () => {
   return (
     <>
       <InputGroup label='Radio group' message='Tell me how happy you are'>
-        <Radio name='choice' value='option1' label={<SvgSmileyHappy />} />
-        <Radio name='choice' value='option2' label={<SvgSmileySad />} />
+        <Radio
+          name='choice'
+          value='option1'
+          label={<SvgSmileyHappy aria-label='Happy' />}
```
rohan-kadkol, line 15: Is there an accessibility-related difference between adding the `aria-label` on the `<svg>` and adding it on the `<Radio>`? This is because the screen reader callouts are the same for both.

I chose to add it to the `<svg>`, since the `<svg>` was being used as a label but did not have any text content.

mayank99: you could test out the difference yourself, but honestly imo this example should just be changed to not use a text label instead of icon.

rohan-kadkol: > you could test out the difference yourself

As I mentioned, I already tested it on the screen reader and found no difference. Did you mean testing some other accessibility aspect apart from the screen reader?

> but honestly imo this example should just be changed to not use a text label instead of icon.

Hmm... I'm a bit confused by your wording. Because this example is already not using a text <label> instead of an icon. Or to rephrase, it is using an icon instead of some text content being passed to the `label` prop of the `<Radio>`. Below is the screenshot of this example.

image

rohan-kadkol: > > As I mentioned, I already tested it on the screen reader and found no difference. Did you mean testing some other accessibility aspect apart from the screen reader?
> 
> when testing with screen readers, just stating something is not enough. you should generally post a screenshot or screen recording, along with the screen reader name, browser name, OS

After testing with NVDA + Edge 116 + Windows 11, the callouts for both are the same (as below).

image

* Applying aria-label to the svg: [8.webm](https://github.com/iTwin/iTwinUI/assets/45748283/9c789443-42c1-41ca-90a4-0360b6177902)
* Applying aria-label to the Radio: [9.webm](https://github.com/iTwin/iTwinUI/assets/45748283/df468931-f0c7-4821-a92f-d29f80a344ac)

> > Hmm... I'm a bit confused by your wording. Because this example is already not using a text instead of an icon. Or to rephrase, it is using an icon instead of some text content being passed to the `label` prop of the `<Radio>`. Below is the screenshot of this example.
> 
> sorry i meant "this example should just be changed to ~not~ use a text label instead of icon". icons are generally decorative and not sufficient, and this example is a bit silly

Sounds good, changed the svgs to text labels (https://github.com/iTwin/iTwinUI/pull/1482/commits/ada98c3d9ff7bb272bc3dd6f97af1fbb9306ba71)

mayank99 (approved):

PR #1483, opened 2023-08-10
improve website sidebar

mayank99:
## Changes

two changes to sidebar:
- increased separation between nav groups.
- used scroll snapping to scroll to the currently active link when page loads.

both of these were based on user feedback.

## Testing

before:

https://github.com/iTwin/iTwinUI/assets/9084735/7f0a0191-59c3-4ffe-9a64-76d813f6d756

after:

https://github.com/iTwin/iTwinUI/assets/9084735/b3d4d1b8-def1-4aaf-940f-23e6a1678b37

## Docs

n/a

### DIFF
--- a/apps/website/src/components/LeftSidebarContent.astro
+++ b/apps/website/src/components/LeftSidebarContent.astro
@@ -68,8 +68,6 @@ sections.set(
   </ul>
 </nav>
 
-<script></script>
-
 <style lang='scss'>
   @layer components {
     nav {
--- a/apps/website/src/components/LeftSidebarContent.astro
+++ b/apps/website/src/components/LeftSidebarContent.astro
@@ -68,6 +68,8 @@ sections.set(
   </ul>
 </nav>
 
+<script></script>
+
 <style lang='scss'>
   @layer components {
     nav {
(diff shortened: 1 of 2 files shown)


apps/website/src/components/LeftSidebarContent.astro
```diff
@@ -68,6 +68,8 @@ sections.set(
   </ul>
 </nav>
 
+<script></script>
```
gretanausedaite, line 71: Empty script tag needed?

mayank99: oop

rohan-kadkol (approved):

apps/website/src/components/LeftSidebarContent.astro
```diff
@@ -169,6 +170,19 @@ sections.set(
       color: var(--color-text);
       position: relative;
 
+      // scroll to current link when page loads, without any JS.
+      // based on https://oh-snap.netlify.app/#scroll-start-trick
```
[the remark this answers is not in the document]
mayank99: the JS approach would just query for `[aria-current=page]` and call `scrollIntoView` on it, no need for nanostores.

and yes, nanostores keeps the state in memory so it would be lost anyway

PR #1485, opened 2023-08-11
Add placeholder text color variable

FlyersPh9:
## Changes

- Add `--iui-color-text-placeholder` to pass WCAG AA.
- Change input & select placeholder text styling to use the new variable, also change font weight to light.
- Set input opacity to fix Firefox inconsistency.

<table>
<tr>
 <td>Theme
 <td>Before
 <td>After
<tr>
 <td>Light
 <td>4.11 ⚠️
 <td>4.67
<tr>
 <td>Light high contrast
 <td>4.11 ⚠️
 <td>4.67
<tr>
 <td>Dark
 <td>2.93 ⚠️ 
 <td>4.62
<tr>
 <td>Dark high contrast
 <td>4.93
 <td>4.93
</table>

## Testing

- Add placeholder text on `color.html`.

## Docs

None yet.


packages/itwinui-css/backstop/tests/color.html
```diff
@@ -130,6 +131,7 @@ <h2 class="iui-text-title iui-text-spacing">Background</h2>
           <ul>
             <li class="iui-text-block">Normal text</li>
             <li class="iui-text-block iui-text-muted">Muted text</li>
+            <li class="demo-text-placeholder">Placeholder text</li>
```
rohan-kadkol, line 134: Do we also have to add this line near L186, under "Background backdrop"?

rohan-kadkol (commented):
The new placeholder color looks great!

packages/itwinui-css/backstop/tests/color.html
FlyersPh9: I think placeholder text color would only ever appear above `--iui-color-background` and never above any other background color.

FlyersPh9:
> can we combine this change with increasing the contrast for regular text?

What's the reasoning for increasing contrast for regular text?  To increase the gap between regular, muted, placeholder, & disabled?

Light theme's text contrast get any higher.  `--iui-color-text-hover` is pure black & `--iui-color-text` is 1 step beneath that (7% decrease in lightness).  We could double our `$fg-*` SASS vars so each step is only 3.5% which would give us a little more granular control.

FlyersPh9:
> alternatively, we could try using `--iui-font-weight-light` (300) for placeholder text. this might actually be better, as color will not be the only differentiator.

This is a good idea!  I've pushed a change to `input.scss` & `select.scss` to use the new placeholder color variable as well as changed the font weight to light.  I did not push test images yet because the file count will increase by a lot, I'll wait for feedback before pushing the test images.

packages/itwinui-css/src/input/input.scss
```diff
@@ -27,7 +27,8 @@
   @include iui-input-size;
 
   &::placeholder {
-    --_iui-input-text-color: var(--iui-color-text-disabled);
+    --_iui-input-text-color: var(--iui-color-text-placeholder);
+    font-weight: var(--iui-font-weight-light);
```
mayank99, line 31: lets also set `opacity: 1` to make it consistent in firefox. see https://github.com/iTwin/iTwinUI/discussions/1417

mayank99 (commented):
new color + font-weight looks good to me

packages/itwinui-css/src/input/input.scss
FlyersPh9: Done!

.changeset/curvy-tools-notice.md
```diff
@@ -4,4 +4,5 @@
 ---
 
 - Add new placeholder text variable.
-- Change input & select placeholder to use the new variable, also change font weight to light.
+- Change input & select placeholder to use the new variable & set font weight to light.
```
mayank99, line 7: can we separate the changesets? one for variables (minor), the other for css and react (patch).

```md
---
'@itwin/itwinui-variables': minor
---

Added new `--iui-color-text-placeholder` variable.
```
```md
---
'@itwin/itwinui-css': patch
'@itwin/itwinui-react': patch
---

Improved color contrast for input and select placeholder text.
```

FlyersPh9: Done, but just so I know what is the benefit / reason for doing so?

mayank99: multiple reasons:
- the only new api is the new variable, so only itwinui-variables needs a `minor` bump
- the new variable is an implementation detail of css/react components, so doesn't necessarily need to be mentioned in the css/react changelogs
- the component changes are not relevant for variables changelog

mayank99 (commented):
should we make this change in v2 as well? it would minimize visual differences and also api differences between v2 and v3 variables.

rohan-kadkol:
> should we make this change in v2 as well? it would minimize visual differences and also api differences between v2 and v3 variables.

I have a few questions:

* If an existing component changes only its appearance (CSS styles of the existing CSS classes) but not the DOM (the html tags keep using the same class names), can we push this change as a `MINOR` version bump?
* What DOM changes can we make while still keeping it a `MINOR` version bump?
* What are some example changes that force us to push it as a `MAJOR` version bump?
* By "api difference", do you mean the addition of the new `--iui-color-text-placeholder` variable?

mayank99:
> By "api difference", do you mean the addition of the new `--iui-color-text-placeholder` variable?

yes.

the other questions are unrelated, as i'm only talking about merging this PR into `main` branch (vs `dev` branch).

rohan-kadkol:
> the other questions are unrelated

To summarize my questions: I have a decent idea for what type of semver change should we add for `itwinui-react`.  I just had some confusions regarding the semver change for `itwinui-css`. Could you give some examples each for what constitutes a `MAJOR`, `MINOR`, and `PATCH` change?

mayank99:
> Could you give some examples each for what constitutes a `MAJOR`, `MINOR`, and `PATCH` change for `itwinui-css`?

sure yeah. assuming you've read https://semver.org,
- any small change that fixes something without affecting the public api should go in a patch release. examples: changing border color, padding, z-index, etc
- any new features/apis should go in a minor release. examples: new class, new attribute, new attribute value.
  - in some cases, a change that would technically be a "patch" could be too large in scope so it might be better to release it in a minor bump.
- any change that will break user's code should go in a major release. examples: class name changes, dom structure changes, file name changes, class removals.

(again this discussion is completely off-topic, so if you have further questions, please use a different channel)

packages/itwinui-css/backstop/tests/color.html
FlyersPh9: In the dev documentation I put a brief explanation.  Without me expanding upon what is already written, [does this section make sense to you?](https://dev-itwinui.bentley.com/docs/variables#background)  Because if this isn't clear enough, I suppose it should be updated.

rohan-kadkol: Thanks! Yes, the description was clear for me.

Although not needed, it would help if we have some picture examples for each color. E.g., a screenshot (or codesandbox type example) of a layout that uses most/all colors, along with arrows to show which color is used where.

E.g., for this [layout](https://itwin.github.io/iTwinUI-layouts), we can have arrows to the tiles to say it is `background` and to the background of the page to say it is `background-backdrop`

FlyersPh9:
> should we make this change in v2 as well? it would minimize visual differences and also api differences between v2 and v3 variables.

It seems like a pretty simple change that we can make in v2, I vote lets backport it.

mayank99:
> It seems like a pretty simple change that we can make in v2, I vote lets backport it.

ok, i've updated the base branch to `main` for this to be released in v2.

mayank99 (approved):

PR #1486, opened 2023-08-14
DatePicker a11y fixes

xman343:
## Changes

#1148 - All examples are currently failing. Any rule violations that can be resolved just by modifying the examples will be implemented. If it is impossible to resolve a rule violation at the example level, then a change to the component may be necessary. 

* `Basic`, `DateRange`, `Disabled`, `Localized`, `Main`, `WithCombinedTime`, `WithTime`, and `WithYear` are violating the `aria-input-field-name` rule, due to the date table within the calendar not having an accessible name. ("ARIA input fields must have an accessible name.")
* * Passing fix: [Implemented](https://github.com/iTwin/iTwinUI/pull/1486/commits/23fc9bba486c6813fde46c8cafd1211273381a10) `useId()` and `aria-labeledby` in component so that the date table is labeled by the currently selected month in the month header. 
* `WithCombinedTime` and `WithTime` are violating the `color-contrast` rule, due to the background color of each selectable `TimePicker` list item being the same as the background color of the `TimePicker`. ("Elements must meet minimum color contrast ratio thresholds.")
* `Menu` is violating the `button-name` rule due to the Date Picker Button not having a proper label. ("Buttons must have discernible text.")
* * Passing fix: [Implemented](https://github.com/iTwin/iTwinUI/pull/1486/commits/d63332f0e50afa6b121e11f5b60e7ecb0095aea3) `aria-label` in `DatePickerMenuExample` for `IconButton` so that the screen reader will correctly identify the button's role as the date picker button.

## Testing

## Docs

TBD


xman343:
Another note: since the `color-contrast` rule is being violated by the `TimePicker` component, I imagine issues pertaining to `TimePicker` would be best handled in its own PR.

xman343:
Changeset coming shortly.

xman343:
> Changeset coming shortly.

Added.

packages/itwinui-react/src/core/DatePicker/DatePicker.tsx
```diff
@@ -226,6 +227,8 @@ export const DatePicker = React.forwardRef((props, forwardedRef) => {
     startDate,
     endDate,
     isDateDisabled,
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
mayank99, line 230: rules of hooks shouldn't be disabled. move `useId()` out into a variable

.changeset/pink-shirts-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
```
mayank99, line 2: patch change, not minor

examples/DatePicker.menu.tsx
```diff
@@ -12,7 +12,11 @@ export default () => {
 
   return (
     <>
-      <IconButton onClick={() => setOpened(!opened)} id='picker-button'>
+      <IconButton
+        onClick={() => setOpened(!opened)}
+        id='picker-button'
+        aria-label='date picker'
```
mayank99, line 18: Using `label` prop would work better than `aria-label` as it also shows a tooltip for sighted users. And the contents should be in _Sentence case_.

also (unrelated): the `id` can be removed

packages/itwinui-react/src/core/DatePicker/DatePicker.tsx
xman343: Re-enabled hook and created the variable for useId() [here](https://github.com/iTwin/iTwinUI/pull/1486/commits/9c9f1d00e7f9cc3ec499fccbb5549154a7f2c72f).

.changeset/pink-shirts-trade.md
xman343: Fixed.

examples/DatePicker.menu.tsx
xman343: Changes made [here](https://github.com/iTwin/iTwinUI/pull/1486/commits/4e60dd9c07558fec535d9b0f7180057ca6deff4f). `id` removed, `aria-label` switched to `label`, converted to Sentence case.

mayank99:
> Something I noticed while testing the component with NVDA is that the day-of-the-week columns are read out by the phonetic reading of their abbreviations ("Su, Mon, Tu" are read out as "Soo, Mon, Two"), which could be confusing. On the other hand, this [example date picker](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/) has NVDA read out the letters of the abbreviation ("Su, Mo, Tu" become "S.U., M.O., T.U."), which at least more accurately reflects what is supposed to be communicated.
> 
> I think it would be a good idea to modify the component so that the days of the week are read out like the latter example.

i just tested the APG example myself and this is what i found:
- in VoiceOver+Safari (Mac), navigating through the heading cells announces them phonetically, same as iTwinUI DatePicker. but navigating through the grid announces the full name of the day ("Monday").
  - <details>
	<summary>see screen recording</summary>
	
	https://github.com/iTwin/iTwinUI/assets/9084735/656aa8b1-a519-4aaf-8b2e-df8ec2792115
	
	</details>
- in NVDA+Firefox (Windows), sometimes it announces the abbreviations, while other times it announces them phonetically. (not obvious in screen recording)
  - <details>
	<summary>see screen recording</summary>
	
	https://github.com/iTwin/iTwinUI/assets/9084735/95bca998-48a8-441f-8ab5-14a51c08c15a
	
	</details>

in any case, we cannot control how a screen-reader is going to pronounce something; we can only change the text itself. i think it would be best if it always pronounced the full name of the day. we can achieve this using a combination of `aria-hidden` (on abbreviated names) and `VisuallyHidden` (on full names).

examples/DatePicker.menu.tsx
mayank99: can you also make the label text more useful? it should consist of the action rather than the name of the component. e.g. "Choose date" or "Pick date" instead of "Date picker"

xman343: Renamed to "Choose date"

mayank99 (approved):
LGTM. i'm ok with handling https://github.com/iTwin/iTwinUI/pull/1486#issuecomment-1679488181 in a separate PR

PR #1488, opened 2023-08-14
Fixed Algolia search box styling

rohan-kadkol:
## Changes

The broken Algolia styles were because of some unnecessary astro scoping in the Algolia styles. I fixed this by adding [`is:global`](https://docs.astro.build/en/guides/styling/#global-styles).

|Old|New|
|-|-|
|image|image|

<details>
<summary>The problem was the unnecessary astro scoping</summary>

image

</details>

---

I am not sure why didn't this "missing styles" issue happen in the dev version. I suspect it has something to do with the bug in astro where importing styles into a layer also duplicates those styles outside the layer (as mentioned in https://github.com/iTwin/iTwinUI/pull/1365#discussion_r1262621490).

But what I'm confused about is why when running `yarn dev` before this PR shows this order of styles/stylesheets in the browser inspect pane:

* `body` in no layer
* `:root` in no layer
* `:root` in the `thirdparty.algolia` layer

i.e., what surprises me is that even though `body` has a lower specificity (`(0, 0, 1)`) than `:root` (`(0, 1, 0)`), `body`'s styles get preference over `:root`.

<details>
<summary>screenshots</summary>

image
image

</details>

---

Regardless, making sure the Algolia styles are global solves the styling issue in the `prod` and `dev` builds. So this PR is good as-is.

## Testing

Ran `yarn build` and `yarn preview` in the `website` workspace to confirm that the styles show up correctly in the production build.

## Docs

N/A

### DIFF
--- a/apps/website/src/components/AlgoliaSearch.astro
+++ b/apps/website/src/components/AlgoliaSearch.astro
@@ -24,7 +24,7 @@
   });
 </script>
 
-<style lang='scss'>
+<style lang='scss' is:global>
   @import '@docsearch/css' layer(thirdparty.algolia);
 
   body {


mayank99 (approved):
LGTM. nice work finding the issue!

---

> what surprises me is that even though `body` has a lower specificity (`(0, 0, 1)`) than `:root` (`(0, 1, 0)`), `body`'s styles get preference over `:root`

i've explained this in https://github.com/iTwin/iTwinUI/pull/1365#discussion_r1281858841 - specificity doesn't matter, position in the dom does.

```html
<html> // <- :root
  <body> // lower in dom
    ...
  </body>
</html>
```

rohan-kadkol:
> LGTM. nice work finding the issue!

Thanks!

> > what surprises me is that even though `body` has a lower specificity (`(0, 0, 1)`) than `:root` (`(0, 1, 0)`), `body`'s styles get preference over `:root`
> 
> i've explained this in [#1365 (comment)](https://github.com/iTwin/iTwinUI/pull/1365#discussion_r1281858841) - specificity doesn't matter, position in the dom does.
> 
> ```
> <html> // <- :root
>   <body> // lower in dom
>     ...
>   </body>
> </html>
> ```

Ohh, right, this concept of which CSS variable definition to use often confuses me. I played around with this [codesandbox](https://codesandbox.io/s/sad-bassi-f2fgpg?file=/src/styles.css) to understand this concept. I will keep this in mind for the next time.

PR #1490, opened 2023-08-14
updated footer documentation.

siddhantrawal:
## Changes

Added Footer subcomponents documentation.


apps/website/src/pages/docs/footer.mdx
```diff
@@ -14,14 +14,54 @@ import * as AllExamples from 'examples';
...
 
-<Placeholder componentPageName='core-footer' />
-
 <LiveExample src='Footer.main.tsx' truncate={false}>
   <AllExamples.FooterMainExample client:load />
 </LiveExample>
 
 The footer is fixed to the bottom of a web page, and contains your typical legally required links. The legal footer is only required to be displayed on pages that are considered a “main entry point into your application”. This means that the sign in and home page of any hub / app are good places to display the footer. You do not need to display the footer within an App UI widget, or deeper pages within an app. You can not include the footer on a page with infinite scrolling because you can not reach the bottom of the page.
 
+## Variants
```
mayank99, line 23: This section should be renamed and probably have some text explanation. Currently it feels too robotic - the examples that follow are not "variants", they are just different customizations to the default footer.

apps/website/src/pages/docs/footer.mdx
```diff
@@ -14,14 +14,54 @@ import * as AllExamples from 'examples';
...
+
+### Customized Default And Custom Elements
+
+Add customized default and custom elements by creating `customUrls` and `translatedTitles` objects. Show Footer component by returning new array containing `customizedDefaultElements` and `customElements`.
+
+<LiveExample src='Footer.customizeddefaultandcustomelements.tsx'>
+  <AllExamples.FooterCustomizedDefaultAndCustomElementsExample client:load />
+</LiveExample>
+
+### Bottom Fixed
```
mayank99, line 49: This can be omitted imo. There is nothing special about `position: absolute; bottom: 0;` - it is regular CSS.

apps/website/src/pages/docs/footer.mdx
siddhantrawal: I am kind of confused about what needs to be changed here. Here is my understanding: 
1) Change the name of the header "Variants" to something suitable?
2) Do you want me to add some info underneath "Variants" or change this text: The footer is fixed...  ...bottom of the page?

mayank99: I think you might be confused by github's UI. When looking at review comments in the Conversation tab, only the last line is relevant. Github shows a few lines above it, but those are unrelated to the comment.



So yes, change the heading and add some text under the heading.

siddhantrawal: Makes sense. Thanks

mayank99: generally it's best to keep headings short, so try just "Customizations"

siddhantrawal: Fixed

mayank99 (approved):

examples/Footer.customcontent.tsx
```diff
@@ -0,0 +1,52 @@
...
+                        {element.title}
+                      </a>
+                    ) : (
+                      element.title
+                    )}
+                  </Footer.Item>
+                </React.Fragment>
+              );
+            }),
+          ,
```
rohan-kadkol, line 47: Unnecessary comma

apps/website/src/pages/docs/footer.mdx
```diff
@@ -14,14 +14,48 @@ import * as AllExamples from 'examples';
...
+
+Add custom element by passing `customElements` prop to Footer component, set the `title` property as label for the custom link and `url` property for URL associated with the custom link.
+
+<LiveExample src='Footer.appendedcustomelements.tsx'>
+  <AllExamples.FooterAppendCustomElementsExample client:load />
+</LiveExample>
+
+### Only Custom Elements
+
+Add only custom elements by passing `customElements` prop to Footer component. It expects an array of objects each describing a custom link for the footer. Set the `title` and `url` property for each defined custom element.
```
rohan-kadkol, line 37: > Set the `title` and `url` property for each defined custom element.

But the example below this paragraph contains elements with just the `title` property. So maybe there should be a mention that the `url` is optional; and if not passed, it will render as plain text and not an anchor.

siddhantrawal: I think it kind of makes sense that it is optional because the first element in the array has both title and URL and the next one has only title set. I am not sure if I should add it or not. I can add it if you would like.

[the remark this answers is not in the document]
siddhantrawal: Fixed

examples/Footer.customcontent.tsx
siddhantrawal: Fixed

apps/website/src/pages/docs/footer.mdx
rohan-kadkol: I mean to say that the doc website says, "Set the `title` and `url` property for each defined custom element.". i.e. set both `title` and `url`. But the example below has some elements only with `title` and no `url`. So I was thinking the last sentence on L37 needs to be rephrased.

siddhantrawal: Fixed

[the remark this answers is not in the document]
siddhantrawal: Fixed.

rohan-kadkol (approved):

PR #1492, opened 2023-08-14
 Fixed a11y violations for Tree examples

rohan-kadkol:
## Changes

`aria-owns` expects a *space* separated list of ids ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns#values)). However, we were passing a *comma* separated list of ids. I corrected this mistake and added a short changeset.

## Testing

Confirmed that the cypress-axe `TreeMainExample` test passes.

## Docs

N/A

### DIFF
--- a/.changeset/mighty-fireants-check.md
+++ b/.changeset/mighty-fireants-check.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': minor
 ---
 
-Fixed `aria-owns` syntax for `TreeNode`'s `<ul>` `subNodes`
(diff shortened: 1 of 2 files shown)


.changeset/mighty-fireants-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
```
mayank99, line 2:
```suggestion
'@itwin/itwinui-react': patch
```

mayank99 (approved):

PR #1494, opened 2023-08-15
ensure color dot always has enough contrast

mayank99:
## Changes

uses a css filter to switch between black/white outline depending on which one has more contrast with the current background.

(technique adapted from this article: https://css-tricks.com/methods-contrasting-text-backgrounds/)

## Testing

tested working against various colors in firefox, chrome, safari.

https://github.com/iTwin/iTwinUI/assets/9084735/beefcc8e-78f5-4226-a916-62c40b004f35

## Docs

n/a


packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -228,13 +228,20 @@ $iui-hover-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-border);
...
   transform: translate(calc(0px - var(--iui-size-xs)), calc(0px - var(--iui-size-xs)));
   cursor: crosshair;
-  box-shadow: rgba(0, 0, 0, var(--iui-opacity-1)) 0 0 0 calc(var(--iui-size-3xs) - 1px),
-    inset 0 0 0 calc(var(--iui-size-3xs) - 1px) rgba(0, 0, 0, var(--iui-opacity-6));
   background-color: var(--iui-color-picker-selected-color);
 
-  &:hover {
-    box-shadow: $iui-hover-box-shadow;
+  &:focus-visible {
+    outline-offset: 3px;
```
FlyersPh9, line 234: Can you think of a similar solution for the `:focus` outline?  It can also get lost depending on the background color.
Screenshot 2023-08-15 at 11 02 43 AM

FlyersPh9 (approved):
I love this solution! 😍 Giving an approval despite [this comment I left](https://github.com/iTwin/iTwinUI/pull/1494/files#r1294718736), could always be addressed at a later time.

packages/itwinui-css/src/color-picker/color-picker.scss
mayank99: i don't want to make it too complicated, but one option is to use a two-tone focus ring
https://www.w3.org/WAI/WCAG21/Techniques/css/C40

mayank99: added secondary box-shadow, it should help a bit. can you test it out on your end?

mayank99:
> Giving an approval despite [this comment I left](https://github.com/iTwin/iTwinUI/pull/1494/files#r1294718736), could always be addressed at a later time.

ok, merging this. lets revisit the focus outline in v3 if needed.

PR #1495, opened 2023-08-15
Header a11y fix

xman343:
## Changes

https://github.com/iTwin/iTwinUI/issues/1148 - The only example `Main` is currently failing. This is due to the lack of an accessible label for the `HeaderSplitButton`. 
* Passing solution: [Implemented ](https://github.com/iTwin/iTwinUI/pull/1495/commits/8fd0c97e498ffa3443c973a7a2bfe0755ca4d689) `aria-label` "More" for `HeaderSplitButton`. This helps the screen reader communicate the purpose of the button to the user.
## Testing

Testing will be conducted through the Cypress web a11y script. Any breaking changes will be evaluated through unit testing & visual tests. Example also read through NVDA to assess screen reader compatibility. 

## Docs

TBD


xman343:
Conflicted on whether to implement the accessible name through `VisuallyHidden` or a `title` for the button. I'm going with `title` for now since it serves the extra function of displaying the label in a popover when it's hovered on. I could change it if we decide that it's unnecessary for this component.

packages/itwinui-react/src/core/Header/HeaderSplitButton.tsx
```diff
@@ -52,6 +52,7 @@ export const HeaderSplitButton = React.forwardRef((props, forwardedRef) => {
         onHide={React.useCallback(() => setIsMenuOpen(false), [])}
       >
         <ButtonBase
+          title='Dropdown menu button'
```
mayank99, line 55: please don't use `title`. https://www.tpgi.com/using-the-html-title-attribute-updated/

`aria-label` should work fine.

and "Dropdown menu button" is not good as the label; but at the same time we don't really know what will be inside the dropdown, so the user should be the one providing it.

i'd say lets go with just "More" for now. in the future, we can make it customizable (#1444)

```suggestion
          aria-label='More'
```

.changeset/ninety-geckos-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added accessible tille to dropdown menu button in HeaderSplitButton.
```
mayank99, line 5:
```suggestion
Added accessible name to dropdown menu button in HeaderSplitButton.
```

packages/itwinui-react/src/core/Header/HeaderSplitButton.tsx
xman343: Thanks for sharing the link. I didn't realize that `title` was that problematic.

mayank99 (approved):

siddhantrawal (approved):

PR #1496, opened 2023-08-15
SideNavigation a11y fix

xman343:
## Changes

#1148 - All SideNavigation examples are failing due to `button-name` Axe violations. In all examples it occurs in the `expand` button, and in `SubmenuExample` it also occurs in the `settings` button within the `submenu`. The examples and / or component will be modified to resolve these rule violations. 
* Passing solution: [Button name added](https://github.com/iTwin/iTwinUI/pull/1496/commits/c49810556d22cd0485babc905dd0ddad50fe3646) by implementing label `Expand or collapse side menu` in the expand button in `SideNavigation`. 
* Passing solution: [Button name added](https://github.com/iTwin/iTwinUI/pull/1496/commits/d38c4d9be3cc4b80e458405367edc5ec540e18b7) to `SubMenuExample` `SubMenu Settings Button` through label `Submenu settings`. 

## Testing

## Docs

N/A


xman343:
I'm still conflicted on the label for the `expand` button. "Expand or collapse side menu" describes both of the actions that the user is meant to perform with the element, but I fear that it ironically runs the risk of making the button's purpose confusing.

"Toggle" was another idea I had, but it's kind of vague as to what exactly it's toggling. I could have it as "Toggle side menu," but then the question is... toggle *what* about the side menu, exactly? 

I could just leave it as "Expand side menu," and the user would likely assume that the button would also collapse the side menu if it's already expanded (a common design pattern), but again, I don't want to run the risk of being too vague. 

I have the label as "Expand or collapse side menu" for now because I'd rather run the risk of giving too much information at first than not enough. It's less work to have the full intent established before we whittle it down into something better, than to build something new from the ground up imo.

xman343:
On another note: even when the `expand` button has an accessible label, my other concern now is that it doesn't read out whether the side menu is expanded or collapsed, when I'm pretty sure that widgets like this are supposed to read out their expand/collapse status. Even though the purpose for this particular example is just to reveal the text labels for each button, I feel that it's still important to communicate that context to the user.

packages/itwinui-react/src/core/SideNavigation/SideNavigation.tsx
```diff
@@ -113,6 +113,7 @@ export const SideNavigation = React.forwardRef((props, forwardedRef) => {
 
   const ExpandButton = (
     <IconButton
+      label='Expand or collapse side menu'
```
mayank99, line 116: i'd go with "Toggle icon labels". It's more accurate and concise.

also `aria-expanded` for assistive tech

```suggestion
      label='Toggle icon labels'
      aira-expanded={_isExpanded}
```

examples/SideNavigation.submenu.tsx
```diff
@@ -11,6 +11,7 @@ import {
   Text,
   SidenavSubmenu,
   SidenavSubmenuHeader,
+  VisuallyHidden,
```
mayank99, line 14: not used

.changeset/polite-eels-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added accessible button name to expand button.
```
mayank99, line 5: which expand button?

examples/SideNavigation.submenu.tsx
xman343: removed

.changeset/polite-eels-unite.md
xman343: Just changed it to `Added accessible button name to SideNavigation expand button.`, hopefully that's enough context?

PR #1498, opened 2023-08-15
simplify skip-to-content demo

mayank99:
## Changes

visual tests sometimes start randomly failing because of the complex header content, so hopefully simplifying it reduces the flakiness.

## Testing

n/a

## Docs

n/a


packages/itwinui-css/backstop/tests/skip-to-content.html
```diff
@@ -70,117 +70,6 @@ <h1>Skip to content link</h1>
               aria-hidden="true"
             ></svg-imodel-hollow>
           </div>
-
-          <div class="iui-page-header-divider"></div>
```
rohan-kadkol, line 74: A comment about why the header is empty might help.

rohan-kadkol: > but why? this page has nothing to do with the header.

This is not a requirement. Just a suggestion, in case someone thinks to readd the other parts of the header.

> also it's not empty, it has the logo which is focusable

By empty I meant it just has the logo.

mayank99: ok, done

rohan-kadkol (approved):

PR #1499, opened 2023-08-16
remove special handling for iphone notch

mayank99:
## Changes

removed safe area insets that were added for iphone notch support because they cause issues when used with RTL or vertical languages.

by default, applications do not fill up the full viewport, so they are already safe from getting clipped by the notch.

if a user wants to override the default behavior using `viewport-fit=cover`, they can also manually add the safe-area-insets to header/side-nav.

## Testing

Was testing it together with @FlyersPh9 in https://github.com/iTwin/iTwinUI/pull/1489#discussion_r1293949164 and found two issues in vertical writing modes:
- when using physical padding (current approach) in header, it results in the header becoming too thick:
   
- when using logical properties, it results in extra gap at the bottom:
   

removing special handling of notch should avoid both issues.

## Docs

n/a

### DIFF
--- a/.changeset/serious-mice-repair.md
+++ b/.changeset/serious-mice-repair.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Removed special handling of iPhone notch in Header and SideNavigation.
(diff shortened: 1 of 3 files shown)


siddhantrawal (approved):

FlyersPh9 (approved):
I don't want to remove notch support, but I understand why we need to.

PR #1500, opened 2023-08-16
Surface a11y fix

xman343:
## Changes

#1148 - Currently, two`Surface` examples are failing the automated a11y tests. 

* SurfaceHeaderFooterExample: violating `button-name` rule
* * Passing fix: [Implemented](https://github.com/iTwin/iTwinUI/pull/1500/commits/233061b748377fde948ac712b278375db058e459) accessible `label` `Settings` given to the settings button in `HeaderFooterExample`. 
* SurfaceNoPaddingExample: violating `list` (ul and ol must only directly contain li, script or template elements) and `scrollable-region-focusable` (Scrollable region must have keyboard access) rules
* * Passing fix: [Implemented](https://github.com/iTwin/iTwinUI/pull/1500/commits/e1f12439217123abcad7ac7bf422535408ce444e) `as='li'` for each `<Divider>` in the `<ul>`. This helps `NoPaddingExample` follow the `list` rule.
* * Passing fix: [Implemented](https://github.com/iTwin/iTwinUI/pull/1500/commits/85a414162723c7ef34f91ef9ad2bb16412b857d3) tab accessibility for  `Anchors` within the `<ul>` in `NoPaddingExample` so that the region will be keyboard accessible; all interactive elements (including links) must be keyboard accessible. 

The examples and / or components will be modified to resolve these rule violations.

## Testing

## Docs

N/A


examples/Surface.nopadding.tsx
```diff
@@ -30,27 +30,27 @@ export default () => {
               padding: 0,
             }}
           >
-            <li style={listStyle}>
+            <li style={listStyle} tabIndex={0}>
```
mayank99, line 33: please don't add tabindex to everything

examples/Surface.nopadding.tsx
```diff
@@ -30,27 +30,27 @@ export default () => {
...
               <Anchor>Daily log</Anchor>
             </li>
-            <Divider />
-            <li style={listStyle}>
+            <Divider as='li' />
```
mayank99, line 36: instead of turning divider into dummy list items, it would be better to just turn the `<ul>`/`<li>` elements into `<div>` and avoid the issue altogether

examples/Surface.nopadding.tsx
mayank99: to fix the scrollable violation, add `tabIndex={0}` only to `<Surface.Body>` and also add `as='article'` (so it has a sensible role) and `aria-labelledby` (with an id pointing to the `h2` element)

examples/Surface.nopadding.tsx
xman343: Converted the list structure to`<div>` elements [here](https://github.com/iTwin/iTwinUI/pull/1500/commits/aab4ea1fa72f0a30d754c4bc2490c86ab1886bee).

examples/Surface.nopadding.tsx
xman343: Implemented this suggestion.

mayank99 (approved):

PR #1501, opened 2023-08-16
remove `overflow: overlay` to fix Table misalignment

### LINKED ISSUE #1503: Table: Body scroll is misaligning body cell with header one
This is reproducible in Storybook too (on some machines and browsers)
https://itwin.github.io/iTwinUI/react/?path=/story/core-table--full-2

image

mayank99:
## Changes

now using this code in both table-header-wrapper and table-body:
```css
overflow-y: scroll;
scrollbar-gutter: stable;
```

this will always reserve equal amount of space, regardless of whether the scrollbar is visible, thus ensuring alignment. fixes #1503

note: horizontal overflow still uses `auto`.

### Background

We keep having alignment issues when using `overflow: overlay` because the table header is not part of the scroll container, so it's best to just avoid it. I also tried `overflow: auto` but i was running into [Safari-specific issues](https://github.com/iTwin/iTwinUI/assets/9084735/53328d34-749b-46e6-81ec-3a613166201d) when the body is not scrollable, so `overflow: scroll` works as a fallback for `scrollbar-gutter`.

## Testing

Tested working in all three browsers on mac.

Chrome:

Firefox:

Safari:

## Docs

added changeset.


siddhantrawal (approved):
It's working both on chrome and firefox on windows.
Screenshot (3)
Screenshot (4)

PR #1505, opened 2023-08-17
Table a11y fix

xman343:
## Changes

https://github.com/iTwin/iTwinUI/issues/1148 - `Table` failing due to `subrow expand button` failing `button-name` rule. 

*Passing solution: [Implemented ](https://github.com/iTwin/iTwinUI/pull/1505/commits/41137d64d4ecd864e4d7d8f5d0b0b41edf2040c0)`aria-label` for `subrow expand button`. 
## Testing

## Docs

N/A


xman343:
Switched from using `label` to `aria-label` because the former caused two unit tests (`should render sub-rows and handle expansions` and `should render filtered sub-rows`) to fail. Both solutions pass the Cypress a11y evaluation, and both are screen-reader legible. I'll keep this solution unless there is one that is supposed to be better.

packages/itwinui-react/src/core/Table/SubRowExpander.tsx
```diff
@@ -26,6 +26,7 @@ export const SubRowExpander = <T extends Record<string, unknown>>(
         expanderCell(cellProps)
       ) : (
         <IconButton
+          aria-label='Toggle sub row'
```
mayank99, line 29: could also add `aria-expanded`

xman343: [Added](https://github.com/iTwin/iTwinUI/pull/1505/commits/f6d1b811ebae1a69e99d95d25f7e47e5fb5bb2cd)

mayank99 (approved):

Issue #1301, opened 2023-05-23
[DropdownMenu] Difficult to select submenu items


HaveSpacesuit:
`.iui-menu` is supposed to have `max-height: var(--iui-menu-max-height)`. However, this does not appear to be defined.

![image](https://github.com/iTwin/iTwinUI/assets/15036346/4e8caece-a497-4a06-8071-59911f5d074d)

In our app, we have defined our own max height to create a scrollbar for long menus. When the scrollbar is manually added, it becomes very difficult to select any submenu items, as the pop up menu usually disappears as soon as the mouse is over the scrollbar area.

![Submenu close](https://github.com/iTwin/iTwinUI/assets/15036346/68a08baa-d7b1-4063-b699-33a56def0ddf)

Looking at the code it seems as if I manually add the `.iui-scroll` class to the dropdown menu, it will calculate a max height for me. However, I don't think as a consumer I am supposed to use the iui classes, and the difficult hover issue is still present.

Reproducible at https://codesandbox.io/s/overflow-dropdown-menu-8sfjx1?file=/src/App.tsx.

mayank99:
Related issues:
- #879
- #653

rseydler:
I'm also faced with a scroll issue.. in that the scroll bars do not appear at all. I have a headerbutton that overflows but no scroll bars show. Sample here - https://codesandbox.io/s/pedantic-chandrasekhar-n7xchg?file=/src/App.tsx

mayank99:
This has been fixed in #1576 and released in [2.11.8](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%402.12.8) 🚀

Issue #879, opened 2022-12-12
Ability to override DropdownMenu submenu expand behavior (on click, instead of on hover)


Juliakas:
#### Feature
Additional prop for `DropdownMenu` component (something like `submenuExpandBehavior` with values `onClick`, `onHover` - _default_) which as suggested would allow to override behavior on how submenu items should be expanded, retracted.

#### Examples
![image](https://user-images.githubusercontent.com/36437897/207004949-a321da10-bca4-4e32-b09b-c574a98e6ef9.png)
When setting `submenuExpandBehavior` to `onCLick`, submenu is only shown when user clicks on _Item \# 3_. `onHover` would be same behavior as it is now. 

Can contribute to this feature if needed.

FlyersPh9:
@Juliakas can you explain why you want this feature?  I shared the issue with members of UX and they are curious as to why.

Juliakas:
@mayank99 yes, can contribute.

@FlyersPh9 as for why we need it - in some cases the submenu list of items can become long with narrow entries - user has to be very precise with their mouse movement in order to not accidently hover away from the items. It also forces user to carefuly hover into submenu by moving mouse horizontally:
![image](https://user-images.githubusercontent.com/36437897/232762613-0625fe28-a1d9-4ef6-9c81-919a5288376c.png)
It would be nice to go straight to item 6 from item 3 by moving mouse diagonally.

Now that I think about it, there could even be a short delay before closing submenu (like 0.3 sec?) when hovering away from items which gives user a chance to go back to the list if they accidently unhovered it by a couple of pixels. Either way, need to test if it fixes the problem.

mayank99:
> Now that I think about it, there could even be a short delay before closing submenu (like 0.3 sec?) when hovering away from items which gives user a chance to go back to the list if they accidently unhovered it by a couple of pixels. Either way, need to test if it fixes the problem.

I like this idea, it is much easier to implement than reworking the whole behavior. Do you want to try doing that for an initial version and then get feedback from your users? (Of course this change would only go in 2.x from our side, but you could patch node modules if you're still on 1.x)

Another option that we are currently working on is #653, which seems like a more robust solution with better UX.

Juliakas:
#653 Looks interesting, think this could work, but have to make sure first.

Also, should we be able to use both 1.x and 2.x components with help of npm package aliases?

mayank99:
> Also, should we be able to use both 1.x and 2.x components with help of npm package aliases?

[...]

PR #1506, opened 2023-08-17
New popover

mayank99:
## Changes

_Continued from #1392_

Replaced old tippy-based private `Popover` component with new floatingui-based private `usePopover` hook and public `Popover` component.

`Popover` defaults to `role=dialog` and does some focus management, whereas `usePopover` gives more control and is useful for internal components that bring their own logic for roles and focus management and custom triggers. The public `Popover` is also used in some internal components that do not need such custom logic.

Removed old `popover.scss`; it was mainly used for reverting tippy styles (which don't exist anymore). The new popover does not need any CSS but also does not come with any animations (can be added in the future).

The public `Popover` has an `applyBackground` prop which converts it into a Surface. Since this is a new component, I've decided not expose all floating-ui props to keep API surface small; we can re-evaluate in the future based on user requests.

## Testing

Thoroughly tested all affected components in storybook and vite playground.

Here's a screen recording showing various components in action:

https://github.com/iTwin/iTwinUI/assets/9084735/a186e76b-4b29-4927-a1cc-cb3afb1cca4f

Updated all affected unit tests.

## Docs

Changesets added. Pending documentation for `Popover` (future PR).


mayank99 (commented):
@gretanausedaite @rohan-kadkol I think the code is ready for review at this point. I'm still working on tests, changesets, PR description, etc.

packages/itwinui-react/src/core/utils/functions/dom.ts
```diff
@@ -52,3 +52,43 @@ export const getTranslateValues = (element: HTMLElement | null | undefined) => {
...
+/**
+ * Accepts a handler which gets called when focus leaves the element
+ * as a result of a keypress (rather than mouse clicks).
+ *
+ * Must be passed to `keydown` to ensure it gets called before `blur` event.
+ *
+ * @example
+ * onKeyDown={handleFocusOut(() => {})}
+ */
+export const handleFocusOut =
```
gretanausedaite, line 65: Why do we need this handler?

packages/itwinui-react/src/core/utils/components/Popover.tsx
```diff
@@ -4,162 +4,273 @@
...
+type PopoverInternalProps = {
+  /**
+   * autoUpdate options that recalculates position
+   * to ensure the floating element remains anchored
+   * to its reference element, such as when scrolling
+   * and resizing the screen
+   *
+   * https://floating-ui.com/docs/autoUpdate#options
+   */
+  autoUpdateOptions?: {
```
gretanausedaite, line 79: Decided to keep autoUpdateOptions and middleware internal?

packages/itwinui-react/src/core/utils/components/InputContainer.tsx
```diff
@@ -52,10 +50,13 @@ export const InputContainer = <T extends React.ElementType = 'div'>(
       data-iui-status={status}
       data-iui-label-placement={isLabelInline ? 'inline' : undefined}
       style={style}
+      ref={forwardedRef}
       {...rest}
     >
       {label && (
         <Label
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore -- shut up for the love of god
```
gretanausedaite, line 59: Maybe don't keep this comment.

packages/itwinui-react/src/core/Table/filters/FilterToggle.tsx
```diff
@@ -7,9 +7,9 @@ import cx from 'classnames';
 import type { HeaderGroup } from 'react-table';
 import {
   useGlobals,
-  Popover,
   SvgFilterHollow,
   SvgFilter,
+  Popover,
```
gretanausedaite, line 12: nit: maybe undo position change? It was alphabetical before.

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
```diff
@@ -26,12 +37,16 @@ export type DropdownMenuProps = {
    * Child element to wrap dropdown with.
    */
   children: React.ReactNode;
-} & Omit<PopoverProps, 'content'> &
-  Omit<CommonProps, 'title'>;
+} & Pick<
+  React.ComponentProps<typeof Popover>,
```
gretanausedaite, line 41: Is there a reason why you are using Popover for types here and `Parameters<typeof usePopover>[0]` in Select?

gretanausedaite (commented):
It's a bit strange that we are building new popovers in all our components. Except DatePickerInput which somehow is able to use Popover you are exporting for users.

Was this your main idea? `usePopover` for internal stuff and `<Popover />` for users?

packages/itwinui-react/src/core/utils/components/Popover.tsx
mayank99: yup

packages/itwinui-react/src/core/utils/components/InputContainer.tsx
mayank99: oh wow. lets pretend you didn't see that

packages/itwinui-react/src/core/Table/filters/FilterToggle.tsx
mayank99: done

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
mayank99: hmm, maybe this one could also use `usePopover`.

i'm also wondering if we should decide on which props to expose and create a helper for that

packages/itwinui-react/src/core/utils/functions/dom.ts
mayank99: i added it for closing the popover when tabbing out of it, but it looks like my implementation has changed significantly and this is not working anymore.

will investigate

mayank99: so it looks like `handleFocusOut` was not working correctly and is not needed if [`FloatingFocusManager`](https://floating-ui.com/docs/floatingfocusmanager) is used, because the latter will properly return focus to the trigger and close the popover.

https://github.com/iTwin/iTwinUI/assets/9084735/8f5782f6-76b5-4947-bd70-9cd456789eab

removed in [`29e37fd`](https://github.com/iTwin/iTwinUI/pull/1506/commits/29e37fd89418072739117ffae30bb20b83106c45).

i also think i ended up in some weird in-between state in storybook where `FloatingFocusManager` was being used even after i removed it. all the more reason we need to get rid of storybook asap

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -126,17 +147,27 @@ export const MenuItem = React.forwardRef((props, ref) => {
       case 'ArrowRight': {
         if (subMenuItems.length > 0) {
           setIsSubmenuVisible(true);
+
+          // flush and reset state so we are ready to focus again next time
+          flushSync(() => setFocusOnSubmenu(true));
+          setFocusOnSubmenu(false);
```
mayank99, line 153: after removing `handleFocusOut`, i'm having a bit of difficulty keeping the menus open. when ArrowRight is pressed, focus moves into the submenu and closes the parent menu.

mayank99: can we remove hover/focus behavior and instead use clicks?

we can always explore adding it back in the future if needed? it would also solve #879.

gretanausedaite: I agree with using clicks.
Maybe switching to controlled mode would solve this? Like parent visible:true once you click on a submenu?

mayank99: it's already using controlled mode, but there are many conditions under which we want to set visible:false, one of which is when focus leaves the menu.

i'm looking into changing the dom structure (currently the submenu is a sibling of the parent menu as they get portaled separately).

mayank99: fixed by removing `FloatingFocusManager` and adding a simple `onKeyDown` listener on `Menu`


this works because:
- `close()` will send focus back to the trigger, thus preserving original tab order and avoiding the need for complex focus management.
- events inside react portals bubble to the react parents, so `Menu` will get keydown events from submenus even if they are sublings in the dom.

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -1661,39 +1661,6 @@ it('should not trigger sorting when filter is clicked', async () => {
   expect(onSort).not.toHaveBeenCalled();
 });
 
-it('should render filter dropdown in the correct document', async () => {
```
mayank99, line 1664: removed after fighting with it for a week.


this test is mocking everything (including me) so not super useful

apps/storybook/cypress-visual-screenshots/baseline/ComboBox.test.ts-With Custom Message Icon (Open).png
mayank99, line 1: this one looks strange, i think it's because the input-container is used as the positioning reference.

gretanausedaite (approved):
Looks good!

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -93,10 +89,6 @@ export const ComboBoxMenu = React.forwardRef((props, forwardedRef) => {
                   // set as constant because we don't want it shifting when items are unmounted
                   maxInlineSize: 0,
 
-                  // max-height must be on the outermost element for virtual scroll
-                  maxBlockSize:
-                    'calc((var(--iui-component-height) - 1px) * 8.5)',
-                  overflowY: isOverflowOverlaySupported() ? 'overlay' : 'auto',
```
mayank99: since we are now using `iui-menu` as the outer element, we don't need to set these inline styles for virtualization

while writing this, i also realized that we can remove `iui-scroll` and make it the default behavior. this will fix #1301 as well. i will create a separate PR for it

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
```diff
@@ -26,12 +37,16 @@ export type DropdownMenuProps = {
...
+  'visible' | 'onVisibleChange' | 'placement' | 'matchWidth'
+> &
```
rohan-kadkol, line 43: Are you not exposing all Popover props to avoid making it to complicated for the users?

packages/itwinui-react/src/core/utils/components/Popover.tsx
```diff
@@ -4,162 +4,271 @@
...
+      matchWidth &&
+        size({
+          apply: ({ rects }) => {
+            setReferenceWidth(rects.reference.width);
+          },
+        } as SizeOptions),
+      middleware.autoPlacement && autoPlacement(),
+      middleware.inline && inline(),
+      middleware.hide && hide(),
+    ].filter(Boolean),
```
rohan-kadkol, line 145: What does `.filter(Boolean)` do?

packages/itwinui-react/src/core/utils/components/Popover.tsx
```diff
@@ -4,162 +4,271 @@
...
+    (userProps?: React.HTMLProps<HTMLElement>) =>
+      interactions.getFloatingProps({
+        ...userProps,
+        style: {
+          ...floating.floatingStyles,
+          zIndex: 9999,
+          ...(matchWidth && referenceWidth
+            ? {
+                minInlineSize: `${referenceWidth}px`,
+                maxInlineSize: `min(${referenceWidth * 2}px, 90vw)`,
```
rohan-kadkol, line 172: Is this to make sure the popover never becomes larger than the screen width? What does the `* 2` signify?

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
mayank99: yes and to avoid issues where users start misusing random props and breaking things

packages/itwinui-react/src/core/utils/components/Popover.tsx
mayank99: it will filter out any falsy values

packages/itwinui-react/src/core/utils/components/Popover.tsx
mayank99: yes we want it to never be larger than 2x the size of the trigger and also to never overflow the screen width

it's basically copied from before

https://github.com/iTwin/iTwinUI/blob/566257c666c26e357dc5be606bdc320db359e477/packages/itwinui-react/src/core/Select/Select.tsx#L383-L384

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -352,30 +352,31 @@ const TileMoreOptions = React.forwardRef((props, forwardedRef) => {
...
-            'iui-visible': isMenuVisible,
-          },
-          className,
-        )}
-        ref={forwardedRef}
-        {...rest}
+      <DropdownMenu
+        onVisibleChange={(open) =>
+          open ? setIsMenuVisible(true) : setIsMenuVisible(false)
+        }
```
rohan-kadkol, line 369: nit
```suggestion
        onVisibleChange={setIsMenuVisible}
```

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -352,30 +352,31 @@ const TileMoreOptions = React.forwardRef((props, forwardedRef) => {
...
+      className={cx(
+        'iui-tile-more-options',
+        {
+          'iui-visible': isMenuVisible,
+        },
+        className,
+      )}
+      ref={forwardedRef}
+      {...rest}
     >
```
rohan-kadkol, line 365: Just wondering, why `<Box>` is now a parent of `<DropdownMenu>` instead of how it was earlier, a child?

mayank99: previously it was wrong.

dropdownmenu's children should almost always be a `<button>` instead of a plain `<div>`

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -100,55 +100,47 @@ const DatePickerInput = React.forwardRef((props, forwardedRef) => {
...
+              setFocus
+              isDateDisabled={isDateDisabled}
+              localizedNames={localizedNames}
+            />
+          }
+          placement='bottom-end'
+          visible={isVisible}
+          onVisibleChange={setIsVisible}
+          closeOnOutsideClick
         >
```
rohan-kadkol, line 137: Any reason to move `<Popover>` inside instead of letting it be the top most component? Something similar to https://github.com/iTwin/iTwinUI/pull/1506#discussion_r1321671312?

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -239,63 +242,32 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
-        break;
-      }
-      default:
-        break;
-    }
-  };
+  const hide = React.useCallback(() => {
+    setIsOpen(false);
+    selectRef.current?.focus({ preventScroll: true });
+    popoverProps?.onVisibleChange?.(false);
```
rohan-kadkol, line 269: nit: The old comment was helpful to understand this code. So better to bring it back.

```suggestion
    selectRef.current?.focus({ preventScroll: true }); // move focus back to select button
```

packages/itwinui-react/src/core/Menu/Menu.tsx
```diff
@@ -89,9 +94,9 @@ export const Menu = React.forwardRef((props, ref) => {
       as='div'
       className={cx('iui-menu', className)}
       role='menu'
-      onKeyDown={onKeyDown}
       ref={refs}
       {...rest}
+      onKeyDown={mergeEventHandlers(props.onKeyDown, onKeyDown)}
```
rohan-kadkol, line 99: nit: Reorder to reduce diff

```suggestion
      role='menu'
      onKeyDown={mergeEventHandlers(props.onKeyDown, onKeyDown)}
      ref={refs}
      {...rest}
```

mayank99: no, the order is intentional. `...rest` would override `props.onKeyDown`

packages/itwinui-react/src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
mayank99: same response here as well. `referenceProps` gets applied on the cloned children and it includes aria attributes which should go on the correct element

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
```diff
@@ -177,17 +177,22 @@ it('should render sublabel', () => {
 });
 
 it('should show sub menu on hover', () => {
+  jest.useFakeTimers();
```
rohan-kadkol, line 180: Why do the timers have to be manually handled?

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
```diff
@@ -244,24 +254,32 @@ it('should handle key press with sub menus', async () => {
...
+  const subMenuItem = screen.getByTestId('sub');
+  expect(subMenuItem).toHaveTextContent('Test sub');
+
+  // go right to move focus
   await userEvent.keyboard('{ArrowRight}');
-  const subTippy = container.querySelector('[data-tippy-root]') as HTMLElement;
-  const subMenu = subTippy.querySelector('.iui-list-item') as HTMLElement;
-  expect(subMenu.textContent).toBe('Test sub');
-  expect(container.ownerDocument.activeElement).toEqual(subMenu);
+  expect(subMenuItem).toHaveTextContent('Test sub');
```
rohan-kadkol, line 266: Duplicate of L262?

```suggestion
```

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
mayank99: probably because everything in jest is fake and it doesn't know to wait for the delay before menu appears.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -655,14 +655,13 @@ it('should call onExpand and onCollapse when dropdown is opened and closed', asy
...
-  const list = document.querySelector('.iui-menu') as HTMLElement;
-  expect(list).toBeVisible();
+  expect(document.querySelector('.iui-menu')).toBeVisible();
   expect(onExpand).toHaveBeenCalled();
 
   const menuItem = document.querySelector('.iui-list-item') as HTMLElement;
 
   await userEvent.click(menuItem);
-  expect(list).not.toBeVisible();
+  expect(document.querySelector('.iui-menu')).toBeFalsy();
```
rohan-kadkol, line 664: nit: You used `toBeVisible()` on L658. Did you mean to use `.not.toBeVisible()` here?

rohan-kadkol (approved):

rohan-kadkol:
> @gretanausedaite @rohan-kadkol I think the code is ready for review at this point. I'm still working on tests, changesets, PR description, etc.

Ohh, sorry, I missed this message. I reviewed and approved it. There were a few parts that involved concepts that I don't understand too well. But for the parts that I did understand, I reviewed them better.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
mayank99: i think `not.toBeVisible()` doesn't always work because we remove this element from the dom. so `not.toBeFalsy()` is more reliable

PR #1508, opened 2023-08-21
use swc to generate cjs styles

mayank99:
## Changes

Previously we were generating `cjs/styles.js` by hand. Turns out mixed default+named exports in CommonJS are complicated and need more work than simply using `module.exports=` and `exports.whatever=`. See [playground](https://play.swc.rs/?version=1.3.74&code=H4sIAAAAAAAAA0utKMgvKlFISU1LLM0pUVAvycgsVgCikoxUmKC6NVcqRFVyfl5xiUJafr6CLUJlXmJuaoq6NQDsn8KpSQAAAA%3D%3D&config=H4sIAAAAAAAAA1VPOw7DIAzdOQXy3KHN2Dv0EIg6EYifMJGKoty9QCBtNr%2B%2FvDHOQZOEJ9%2FKWUAQkTCeuDCUXRKfwgBKK0hGFRLchqqpSrMwhI3aDwWSiAumlqLpPj16Aoz3hCPROaucmvP%2FpvQ2RCS6GqtVuMXgdZH1VbD%2BvTax%2F5JyqKi2We80wc85Fs92UPQa8RRXZPsXM%2FmJHxwBAAA%3D).

So now, I'm using `swc` to transform the ESM exports into CJS.

Unrelated: I'm auto-deleting the `styles.js` that gets left behind when switching between `dev`/`main` branches, as it messes up build sometimes.

## Testing

Confirmed that correct output is generated after build. Confirmed that it works in CJS by testing locally in a CRA test app and also testing in a real project with @veekeys.

## Docs

N/A

### DIFF
--- a/.changeset/thirty-beds-mate.md
+++ b/.changeset/thirty-beds-mate.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue with styles not getting included in CommonJS environments.
--- a/.changeset/thirty-beds-mate.md
+++ b/.changeset/thirty-beds-mate.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Fixed an issue with styles not getting included in CommonJS environments.
+Fixed an issue where styles were not included in CommonJS environments.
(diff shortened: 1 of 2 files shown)


rohan-kadkol (approved):
The code LGTM. But I will be able to test it after ~30 mins (because of a meeting).

Since you already tested it with Vykintas, and since this PR is urgent, I thought to pre-emptively give an approval and test it in 30 mins.

mayank99:
released in 2.12.4 and tested again in codesandbox:
- confirmed 2.12.3 is broken. https://codesandbox.io/s/2-12-3-cjs-broken-yhqlrk
- confirmed 2.12.4 works. https://codesandbox.io/s/frosty-voice-wkxvgw

PR #1510, opened 2023-08-22
Customisation effort: toast

gretanausedaite:
## Changes

- add `contentProps`

## Testing

Unit tests added

## Docs

Changeset


.changeset/cool-gifts-dress.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
```
rohan-kadkol, line 2: Since adding functionality is a `minor` change.

```suggestion
'@itwin/itwinui-react': minor
```

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -248,17 +267,34 @@ export const ToastPresentation = (props: ToastPresentationProps) => {
...
   return (
     <Box className={cx(`iui-toast iui-${category}`, className)} {...rest}>
-      <Box className='iui-status-area'>
-        {<StatusIcon className='iui-icon' />}
+      <Box
+        {...statusAreaProps}
+        className={cx('iui-status-area', statusAreaProps?.className)}
+      >
+        {!!startIcon ? (
+          <Icon>{startIcon}</Icon>
```
rohan-kadkol, line 285: Should we apply the default `fill` according to the category even when the `startIcon` is provided? Users can then override it if they would want to.

```tsx
startIcon: (
  <SvgAirplane
    style={{
      fill: 'red',
    }}
  />
),
```

Because if they want to use the status color on a custom icon and we don't apply it by default, they'll have to handle the color for each category themselves, which is a bit difficult.

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -35,6 +36,18 @@ export type ToastProps = {
...
    */
   content: React.ReactNode;
+  /**
+   * Passes props to content
+   */
+  contentProps?: React.ComponentPropsWithoutRef<'div'>;
+  /**
+   * Passes props to status area
+   */
+  statusAreaProps?: React.ComponentPropsWithoutRef<'div'>;
```
gretanausedaite, line 46: Not sure if needed?
I can't imagine a scenario where user want to customise this area.

packages/itwinui-react/src/core/Toast/Toast.tsx
gretanausedaite: 1. Status icons with colours are standard. If user wants something custom, they have to fully customise it themselves.
2. Starting to think to remove <Icon /> wrapper and just simply render what user provided. It would be user responsibility to pass correct Icon.

rohan-kadkol: > 1. Status icons with colours are standard. If user wants something custom, they have to fully customise it themselves.

I guess both ways have benefits. My point was just that if a user just wants a custom icon but wants to keep the same fill colour, then it would be a bit of extra work to find which iui color variable to use for that category and then set it manually. Also, if they have multiple categories, then they need to know the colour for the other categories too.

> 2. Starting to think to remove  wrapper and just simply render what user provided. It would be user responsibility to pass correct Icon.

Sounds good, we can then document that we require users to wrap their icon with `<Icon>` to get some styling and accessibility features.

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -35,6 +36,18 @@ export type ToastProps = {
...
+  /**
+   * Svg icon shown on the left of toast message.
+   */
+  startIcon?: React.ReactNode;
```
mayank99, line 50: has anybody asked for support for adding custom icons?

mayank99: alert was refactored completely, and it started from a user request.

if someone requests it in toast, we can revisit this, but right now the goal is to just be able to add class names to existing dom elements

gretanausedaite: Fine

packages/itwinui-react/src/core/Toast/Toast.tsx
mayank99: i'm ok with leaving this one out for now. Most important parts are the toast and the toast-message

packages/itwinui-react/src/core/Toast/Toast.tsx
rohan-kadkol: Just curious, is there a downside to supporting a custom start icon? Like if there is no downside, should we leave it as it is instead of spending time undoing it?

packages/itwinui-react/src/core/Toast/Toast.tsx
rohan-kadkol: Both options sound good to me. Since it is already added, might be better to leave it.

mayank99: > Since it is already added, might be better to leave it.

It takes 15 seconds to remove it, so this is not a strong argument. It's always worth taking the extra time to think about the public-facing API; we don't want to carelessly add undesired props just because we didn't want to spend a few extra seconds.

rohan-kadkol: > It takes 15 seconds to remove it, so this is not a strong argument. 

I don't mean to save time by doing things hastily. I just meant to say if we are all on the fence about whether to keep it or not, no need to bother removing it. But if we all feel strong to remove it, then let's remove it.

> It's always worth taking the extra time to think about the public-facing API; we don't want to carelessly add undesired props just because we didn't want to spend a few extra seconds.

For sure, I definitely agree. But I think that in general, if exposing one prop would help some user in the future, it may not be wise to be this conservative on our end. But if the cons of including the prop outweigh the added customizability, let's not expose it.

In this case, I am okay with both: keeping or removing `statusAreaProps`.

gretanausedaite: Removed. It's easier to add props in a future than to remove or change them.

packages/itwinui-react/src/core/Toast/Toast.tsx
gretanausedaite: Removed icon prop.

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -223,6 +228,7 @@ export const Toast = (props: ToastProps) => {
             type={type}
             hasCloseButton={hasCloseButton}
             onClose={close}
+            contentProps={contentProps}
```
mayank99, line 231: currently, the content (message) can take a custom classname, but the toast itself cannot.

suggestion: replace `contentProps` with `domProps` which is an object with two fields: `content` and `toast`
```js
toaster.positive('hello', {
  domProps: {
    toast: { className: 'my-toast' },
    content: { className: 'my-toast-content' },
  }
});
```

gretanausedaite: better?

gretanausedaite: done

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -233,7 +243,10 @@ export const Toast = (props: ToastProps) => {
 export type ToastPresentationProps = Omit<
   ToastProps,
   'duration' | 'id' | 'isVisible' | 'onRemove'
-> & { onClose?: () => void } & CommonProps;
+> & {
+  onClose?: () => void;
+  contentProps?: React.ComponentProps<'div'>;
+} & CommonProps;
```
rohan-kadkol, line 249: Since `ToastPresentationProps` includes `ToastProps`, we can use the `ToastProps.domProps` instead of defining `contentProps` as a new prop.

I also realized that `CommonProps` would be redundant, since `domProps.toastProps` already includes the common props. So we can remove `CommonProps` and use `domProps.toastProps` instead. Removal is possible here since `ToastPresentation` is private.

```suggestion
export type ToastPresentationProps = Omit<
  ToastProps,
  'duration' | 'id' | 'isVisible' | 'onRemove'
> & { onClose?: () => void };
```

packages/itwinui-react/src/core/Toast/Toast.tsx
rohan-kadkol: I believe my comment (https://github.com/iTwin/iTwinUI/pull/1510#discussion_r1305591750) is related to this thread.

packages/itwinui-react/src/core/Toast/Toast.tsx
mayank99: what's the benefit of this? top-level `className` makes more sense than `domProps.toastProps.className`

i believe AppUI is using `ToastPresentation` (even though it's private) so they might be relying on top-level props

rohan-kadkol: > what's the benefit of this? top-level `className` makes more sense than `domProps.toastProps.className`

I feel `domProps` should be the single source of truth, despite the long name (`domProps.toastProps.className`).

Moreover, I saw a TODO saying that we should remove `CommonProps` once it is not used any more. So I believe we want to move away from `CommonProps`.

https://github.com/iTwin/iTwinUI/blob/f79ed42611492393bdaca91aefc9fc592386a19d/packages/itwinui-react/src/core/utils/props.ts#L7-L12

> i believe AppUI is using `ToastPresentation` (even though it's private) so they might be relying on top-level props

Since they are using `<ToastPresentation>`, we can consider keeping `CommonProps` and className. But ideally, we should be free to make breaking changes to our private components/APIs.

mayank99: The only reason we have `domProps` is because toaster's API is imperative; we don't want a hundred different props directly under the options passed to `toaster.positive()`. For actual components, top-level props makes sense.

CommonProps is ok to remove. I'm fine with it, we can forward ref, spread top-level rest props and merge `className`.
```tsx
export type ToastPresentationProps = Omit<
  ToastProps,
  'duration' | 'id' | 'isVisible' | 'onRemove' | 'domProps'
> & {
  onClose?: () => void;
  contentProps?: React.ComponentProps<'div'>;
};

export const ToastPresentation = React.forwardRef((props, forwardedRef) => {
  // ...
}) as PolymorphicForwardRefComponent<'div', ToastPresentationProps>;
```

But you're right about our ability to make changes to private apis, so i'll leave it on @gretanausedaite to decide.

gretanausedaite: I'd like to keep top-level props as it makes the most sense for me.

I'd like this PR to be customisation changes only to keep it granular. Agree with switching to poly component in a different PR as it does not affect props I added.

mayank99: could you just change the props type then? i believe `domProps` needs to be omitted

gretanausedaite: yup, updated

mayank99 (approved):

packages/itwinui-react/src/core/Toast/Toast.tsx
rohan-kadkol: The type `ToastPresentationProps` seem to have some type issues. Like when calling `<ToastPresentation>`, is is called with `{...domProps?.toastProps}` of type `React.ComponentProps<'div'>`.

```tsx
<ToastPresentation
  // ...
  {...domProps?.toastProps} // toastProps is of type React.ComponentProps<'div'>;
  contentProps={domProps?.contentProps}
/>
```

But `ToastPresentationProps` does not accept `React.ComponentProps<'div'>`.

mayank99: ah yeah, my suggestion about Polymorphic above should fix this type error

gretanausedaite: #1519

rohan-kadkol: There still seems to be some type error when calling `<ToastPresentation>` with `{...domProps?.toastProps}`. It complains about a missing `as`. The error goes away after removing `{...domProps?.toastProps}` or keeping it but adding `as` to `<ToastPresentation>`

image

mayank99: `as='div'` should fix it

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -248,21 +262,29 @@ export const ToastPresentation = React.forwardRef((props, forwardedRef) => {
...
     className,
+    contentProps,
     ...rest
   } = props;
 
   const StatusIcon = StatusIconMap[category];
 
   return (
     <Box
+      as='div'
```
rohan-kadkol, line 273: `as` will come from `...rest`, so does not need to be hardcoded here.

```suggestion
```

packages/itwinui-react/src/core/Toast/Toast.test.tsx
```diff
@@ -185,3 +185,26 @@ it('should close temporary toast after 7s', () => {
 
   jest.useRealTimers();
 });
+
+it('should pass content props correctly', () => {
```
rohan-kadkol, line 189: nit: Since earlier, we were using `CommonProps` (id, className, style), it is good to test if className and style are applied correctly without `CommonProps`, like you have it now.

But, you could also test if some other props apart from those three are passed correctly. E.g., `onClick`.

This is just an optional suggestion.

packages/itwinui-react/src/core/Toast/Toast.tsx
mayank99: it's needed to work around typescript issues
https://github.com/iTwin/iTwinUI/pull/1510#discussion_r1307512274

rohan-kadkol: The `as='div'` was needed when calling `<ToastPresentation>` (https://github.com/iTwin/iTwinUI/pull/1510/commits/5f48b75c0149125dc1f6f56eb2033c0d054fbc10). But in the code for `ToastPresentation`, the top-most box gets the `as` from the props.

Also, no type errors come when removing L273.

gretanausedaite: Removed

packages/itwinui-react/src/core/Toast/Toast.test.tsx
gretanausedaite: Leaving it as is for now.

PR #1516, opened 2023-08-24
fix: SearchBox placeholder color

gretanausedaite:
## Changes

#1485 introduced new placeholder styling for Inputs. Adding the same styling to SearchBox that is using InputFlexContainer.

## Testing

Visual tests updated.

## Docs

Changeset


packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -104,6 +104,14 @@
     &:is(:not(:first-child)) {
       padding-inline-start: var(--iui-size-2xs);
     }
+
+    &::placeholder {
+      --_iui-input-text-color: var(--iui-color-text-placeholder);
```
mayank99, line 109: it looks like this variable only gets used in the parent, so it doesn't need to be set here. we can directly use `color: var(--iui-color-text-placeholder)`

mayank99: remove line 113

gretanausedaite: Updated?

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -104,6 +104,14 @@
...
+      font-weight: var(--iui-font-weight-light);
+      opacity: 1;
+      user-select: none;
+      color: var(--_iui-input-text-color);
```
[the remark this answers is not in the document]
gretanausedaite: Yes, we can. See conversation above ^

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -104,6 +104,13 @@
     &:is(:not(:first-child)) {
       padding-inline-start: var(--iui-size-2xs);
     }
+
+    &::placeholder {
+      color: var(--iui-color-text-placeholder);
+      font-weight: var(--iui-font-weight-light);
+      opacity: 1;
+      user-select: none;
```
mayank99, line 112: is `user-select: none` needed?

mayank99 (approved):
LGTM. just one question below

packages/itwinui-css/src/input-container/input-container.scss
gretanausedaite: Probably not, but it's added to the inputs so kept it here.

mayank99: looks like it wasn't added in #1485, it was already present before.


if it's not needed, then let's not add it here in searchbox

gretanausedaite: removed from SearchBox

PR #1517, opened 2023-08-25
Corrected the path of `/styles.js` used inside `generateStyles.mjs`

rohan-kadkol:
## Changes

@siddhantrawal and I noticed that the `packages/itwinui-react/src/styles.js` folder was not automatically deleted when we switched from the `dev` branch to the `main` branch.

I realized that is because of a mistake in the path of the `styles.js` folder that is used in `generateStyles.mjs`. This PR:
* corrects that path from (https://github.com/iTwin/iTwinUI/pull/1508)
* adds `recursive: true` to [`fs.promises.rmdir()`](https://nodejs.org/api/fs.html#fspromisesrmdirpath-options) to delete the entire `styles.js` folder 

## Testing

Confirmed that the `packages/itwinui-react/src/styles.js` folder is deleted after running `yarn build --filter=itwinui-react`

## Docs

N/A


gretanausedaite (commented):
Why css tests are failing?

mayank99 (approved):

mayank99:
> Why css tests are failing?

i re-ran and it passed

PR #1461, opened 2023-07-31
make disabled buttons interactable

mayank99:
## Changes

Created new `<ButtonBase>` component used to replace all instances of `<button>`. This component intercepts the `disabled` prop, converts it to `aria-disabled` and manually disables clicks.

Because disabling clicks requires javascript, the `disabled` attribute is still used on first render to make it progressively enhanced.

See https://css-tricks.com/making-disabled-buttons-more-inclusive/

## Testing

✅ focusable
✅ hoverable (tooltip appears)
✅ not clickable
✅ styled like disabled
✅ conveyed as disabled by AT

https://github.com/iTwin/iTwinUI/assets/9084735/ee962612-69ff-4f80-9b32-e0e3f32c5f10

Safari + VO:

https://github.com/iTwin/iTwinUI/assets/9084735/c96beeb2-5d27-4c2c-9e5c-c2504a54ed6f

Firefox + NVDA:

https://github.com/iTwin/iTwinUI/assets/9084735/20b5ad6e-0805-4661-afe7-fcdb575220eb

## Docs

Pending

PR #1523, opened 2023-08-28
add `ariaStrategy` prop to `Tooltip`

mayank99:
## Changes

*This PR is a follow-up to #1311 and precursor to #1506.*

- Added new `ariaStrategy` prop to `Tooltip` to allow switching between `aria-describedby`, `aria-labelledby` and nothing (see https://github.com/iTwin/iTwinUI/pull/1311#pullrequestreview-1505237337).
  - This prop is used in `IconButton` to replace the custom Popover-based tooltip implementation.
  - This prop is also used in `Slider` thumb because (1) the tooltip value keeps updating but is not reflected in screen-readers, and (2) the slider already has its value exposed through `aria-valuenow`.
- Added `aria-hidden` to Tooltip to prevent duplicate announcements (the text should already be exposed through `aria-describedby`/`aria-labelledby`/visuallyhidden text) and removed `role=tooltip` because of [poor support](https://a11ysupport.io/tech/aria/tooltip_role).
- Instead of conditionally rendering the tooltip when it's open, it will now *always* be rendered and its display will be toggled using the `hidden` property.
  - The intention is to make `aria-describedby` available to screen reader users, even when the tooltip is not open (see testing note in https://github.com/iTwin/iTwinUI/pull/1461#issue-1829866396).
  - Only works in VoiceOver (see screen-recording below). I checked that the description was exposed in the accessibility tree in both Chrome and Firefox devtools, so it looks like a bug in NVDA ([more info](https://adrianroselli.com/2022/04/accessible-description-exposure.html)). This behavior is not worse than before and the description is still accessible by tabbing, so I'm fine with it.
- Removed unused `.iui-tooltip-visible`/`.iui-tooltip-container` and added CSS for `[hidden]`. I believe these classes were designed to make animation work, but are not useful in practice; we can explore animation in the future.

## Testing

Tested functionality of all components in storybook.

<details>

<summary>NVDA 2023.1 + Firefox 114 (Windows 11)</summary>

Traversing ButtonGroup using virtual cursor (in both directions):

https://github.com/iTwin/iTwinUI/assets/9084735/05e1e601-e831-4d22-a77b-73094e10eb3b

Traversing buttons (regular and disabled) that have tooltips with default `ariaStrategy`, first using virtual cursor then tabbing:

https://github.com/iTwin/iTwinUI/assets/9084735/7d5bcf4a-9d40-4b51-b1bd-2d927ed18c03

</details>

<details>

<summary>VoiceOver + Safari 16.6 (macOS 13.5)</summary>

Traversing ButtonGroup using virtual cursor (in both directions):

https://github.com/iTwin/iTwinUI/assets/9084735/7ada1244-fa6f-4c9f-ace0-a5c59665369a

Traversing buttons (regular and disabled) that have tooltips with default `ariaStrategy`, using virtual cursor:

https://github.com/iTwin/iTwinUI/assets/9084735/0efafc37-2870-48b9-b130-980a13b4dd0b

</details>

Updated affected unit/visual tests, and added a unit test for `ariaStrategy`.

## Docs

Added changesets. No other doc updates needed I think.

### DIFF
--- a/.changeset/nice-crabs-explode.md
+++ b/.changeset/nice-crabs-explode.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
+---
+
+Removed `iui-tooltip-container` and `iui-tooltip-visible` classes. The display is now toggled using the `hidden` HTML attribute, and the positioning should be managed using custom JS/CSS.
--- a/packages/itwinui-css/backstop/tests/tooltip.html
+++ b/packages/itwinui-css/backstop/tests/tooltip.html
@@ -122,7 +122,7 @@
 
       <div class="demo-tooltip-container">
         Hover here (permanent)
-        <span class="iui-tooltip bottom iui-tooltip-visible">
+        <span class="iui-tooltip bottom">
           This tooltip will not disappear.
         </span>
       </div>
--- a/packages/itwinui-css/backstop/tests/tooltip.html
+++ b/packages/itwinui-css/backstop/tests/tooltip.html
@@ -21,7 +21,10 @@
       @import url("./assets/demo.css") layer(demo);
       @import url('@itwin/itwinui-variables') layer(variables);
       @import url("@itwin/itwinui-css/css/all.css") layer(itwinui);
-      .iui-tooltip-container {
+      .demo-tooltip-container {
+        inline-size: fit-content;
+        block-size: fit-content;
+        position: relative;
         margin-left: 90px;
         margin-bottom: 35px;
       }
@@ -69,29 +72,44 @@
     <br />
 
     <section id="demo-tooltip">
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (top)
-        <span class="iui-tooltip top">Hello, I'm a tooltip!</span>
+        <span
+          class="iui-tooltip top"
+          hidden
+        >Hello, I'm a tooltip!</span>
       </div>
 
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (right)
-        <span class="iui-tooltip right">Hello, I'm a tooltip!</span>
+        <span
+          class="iui-tooltip right"
+          hidden
+        >Hello, I'm a tooltip!</span>
       </div>
 
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (bottom)
-        <span class="iui-tooltip bottom">Hello, I'm a tooltip!</span>
+        <span
+          class="iui-tooltip bottom"
+          hidden
+        >Hello, I'm a tooltip!</span>
       </div>
 
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (left)
-        <span class="iui-tooltip left">Hello, I'm a tooltip!</span>
+        <span
+          class="iui-tooltip left"
+          hidden
+        >Hello, I'm a tooltip!</span>
       </div>
 
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (Lorem ipsum)
-        <span class="iui-tooltip top">
+        <span
+          class="iui-tooltip top"
+          hidden
+        >
           Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
           tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
           veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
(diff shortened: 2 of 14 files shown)


packages/itwinui-css/src/tooltip/tooltip.scss
```diff
@@ -44,4 +22,8 @@
   border: 1px solid rgba(255, 255, 255, var(--iui-opacity-4));
 
   @include mixins.iui-blur($hsl: 0 0% 0%, $opacity: 3);
+
+  &[hidden] {
+    display: none !important;
```
gretanausedaite, line 27: `!important` is needed?

gretanausedaite (approved):
Looks good.

packages/itwinui-css/backstop/tests/tooltip.html
```diff
@@ -102,12 +120,20 @@ <h1>Tooltip</h1>
         </span>
       </div>
 
-      <div class="iui-tooltip-container">
+      <div class="demo-tooltip-container">
         Hover here (permanent)
         <span class="iui-tooltip bottom iui-tooltip-visible">
```
rohan-kadkol, line 125:
```suggestion
        <span class="iui-tooltip bottom">
```

packages/itwinui-react/src/core/Slider/Slider.test.tsx
```diff
@@ -196,7 +196,7 @@ it('should show tooltip when focused', async () => {
   expect(document.activeElement).toEqual(
     container.querySelector('.iui-slider-thumb'),
   );
-  expect(document.querySelector('.iui-tooltip')?.textContent).toBe('50');
+  expect(document.querySelector('.iui-tooltip')).toBeVisible();
```
rohan-kadkol, line 199: We can keep the `textContent` check too (similar to "should show tooltip on thumb hover" and "... focus" tests)

```suggestion
  const tooltip = document.querySelector('.iui-tooltip');
  expect(tooltip).toBeVisible();
  expect(tooltip?.textContent).toBe('50');
```

mayank99: the text is not important, and it's already checked in the other tests you mentioned. the goal here is to test visibility

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -251,15 +263,18 @@ export const Tooltip = React.forwardRef((props, forwardRef) => {
             children as JSX.Element,
             tooltip.getReferenceProps({
               ref: childrenRef,
+              ...(ariaStrategy === 'label' && {
+                'aria-labelledby': tooltip.getFloatingProps().id,
+              }),
+              'aria-describedby':
+                ariaStrategy === 'description'
+                  ? tooltip.getFloatingProps().id
+                  : undefined,
```
rohan-kadkol, line 272: A ternary if-else chain is better suited here

```suggestion
              ...(ariaStrategy === 'label'
                ? {
                    'aria-labelledby': tooltip.getFloatingProps().id,
                  }
                : ariaStrategy === 'description'
                ? {
                    'aria-describedby': tooltip.getFloatingProps().id,
                  }
                : {}),
```

packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
```diff
@@ -39,14 +39,22 @@ it('should be visible', () => {
...
-  expect(queryByText('some text')).toBeNull();
+  expect(queryByText('some text')).not.toBeVisible();
+
+  rerender(
+    <Tooltip content='some text' visible>
+      <div>Not visible!</div>
+    </Tooltip>,
+  );
+
+  expect(queryByText('some text')).toBeVisible();
```
rohan-kadkol, line 57: I guess the `<div>` contents on L45/54 ("Visible" / "Not visible") are swapped

```suggestion
it('should respect visible prop', () => {
  const { queryByText, rerender } = render(
    <Tooltip content='some text' visible={false}>
      <div>Not visible!</div>
    </Tooltip>,
  );

  expect(queryByText('some text')).not.toBeVisible();

  rerender(
    <Tooltip content='some text' visible>
      <div>Visible!</div>
    </Tooltip>,
  );

  expect(queryByText('some text')).toBeVisible();
```

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: i don't think so. This decision is made intentionally - we want to override the `aria-describedby` that comes from floating-ui

rohan-kadkol: Ohh, I see, then probably a comment near this code block would help.

mayank99: added

.changeset/nice-crabs-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
+---
+
+Removed `iui-tooltip-container` and `iui-tooltip-visible` classes. The display is now toggled using the `hidden` HTML attribute, and the positioning should be managed using custom JS/CSS.
```
rohan-kadkol, line 5: Should there be a mention about _"positioning"_? Because I believe even before this PR, positioning had to be handled separately.

Also, might be useful to mention that the tooltip is now always in the DOM.

mayank99: > Should there be a mention about _"positioning"_? Because I believe even before this PR, positioning had to be handled separately.

it was partly managed with `iui-tooltip-container`. theoretically, you could easily calculate the inset values as they would be relative to the container, whereas now we don't provide a container

> Also, might be useful to mention that the tooltip is now always in the DOM.

this part is kinda implied by `hidden`, but also it's not a requirement. it can still be removed from the DOM if desired (this is the css changeset)

PR #1524, opened 2023-08-28
add `FloatingDelayGroup` to `ButtonGroup`

mayank99:
## Changes

improves the experience by sharing the delay, so that there is no delay between buttons within a group.

see https://floating-ui.com/docs/floatingdelaygroup and https://github.com/iTwin/iTwinUI/pull/1311#pullrequestreview-1505237337

## Testing

Before:

https://github.com/iTwin/iTwinUI/assets/9084735/e9922f21-36e8-47ed-8c04-f273dee773ba

After:

https://github.com/iTwin/iTwinUI/assets/9084735/d4aa4770-5736-4498-88f8-921efb36766e

## Docs

not needed i think


mayank99:
> Everything is good but visual tests (css) are failing.

passing now after updating branch. the css tests are also just flaky in CI 🤷

PR #1514, opened 2023-08-22
Improve Table typing

rohan-kadkol:
## Changes

* We no longer depend on [`@types/react-table`](https://yarnpkg.com/package?q=types%2Freact-table&name=%40types%2Freact-table). Instead, I merged the types from `@types/react-table` with our custom types from [`react-table.config.ts`](https://github.com/iTwin/iTwinUI/blob/f97488f629d5e97d6f8d5ff4708f11f1724f86c8/packages/itwinui-react/src/types/react-table-config.ts#L1). The new file is called `react-table.ts`.
  * We now export all react-table types from a new endpoint called `/react-table`.
    * ```diff
      import { Table } from '@itwin/itwinui-react';
      - import { Column, TableState } from 'react-table';
      + import { Column, TableState } from '@itwin/itwinui-react/react-table';
      ```
    * But for our internal files, we make no change. i.e. We still import from `react-table`. This is because of the following reason: 
      * Since we don't depend on `@types/react-table` anymore, I added [`paths`](https://www.typescriptlang.org/tsconfig/#paths) to the `compilerOptions` in `itwinui-react/tsconfig.json` to point to our `react-table-types.d.ts` for the types. This enables intellisense/type checking to work correctly without any import change.
* Exposed endpoint `/react-table` works only for new node versions. Thus, `react-table.ts` is transpiled to `react-table.d.ts` and stored in the root folder so that this file is accessible to users with old node versions who cannot use the `exports` map in `package.json` and have to import by paths. (https://github.com/iTwin/iTwinUI/pull/1514#discussion_r1317310924)
* Instead of each of the row data to extend `object` (like in `@types/react-table`), each row data now extends `Record<string, unknown>` (like in our Table files (e.g. [TextFilter](https://github.com/iTwin/iTwinUI/blob/321e0d894cad911743526a4d0b589c4460625e3b/packages/itwinui-react/src/core/Table/filters/TextFilter/TextFilter.tsx#L18C28-L18C61)))
  * ```diff
    - export type Column<D extends object = {}> = ...
    + export type Column<D extends Record<string, unknown> = {}> = ...
    ```
* Since we are now fixing Table types properly, I undid the parts of #1072 that were meant to be temporary changes.
* Created `tsconfig.react-table.json` to transpile `react-table.ts` to `react-table.d.ts`. This is used in `@itwinui-react`'s `build` and `dev:types` scripts.
[...]

PR #1525, opened 2023-08-28
`.stylelintrc.json` for css and variables packages

rohan-kadkol:
## Changes

* When merging `dev` into my PR #1514, I got this error:
    * image
* Even in the `dev` branch, I then noticed that making some dummy changes to some of the scss files in `packages/itwinui-variables` gives the same error when trying to make a commit.
* So I did the following:
  * Moved the `.stylelintrc.json` from `packakges/itwinui-css` to `internal/configs`
  * `itwinui-css` and `itwinui-variables` each have a `.stylelintrc.json` that extends the `internal/configs/.stylelintrc.json` file.

<details>
<summary>Old PR description</summary>

Before https://github.com/iTwin/iTwinUI/pull/1525#discussion_r1308828422):

* So I did the following:
  * Move `.stylelintrc.json` from `packages/itwinui-css` back to the root folder (undoing the file move in #1468)
  * Unrelated: Removed `"lib/**/*"` from `ignoreFiles` because I didn't find any folder named `lib`

<details>

## Testing

Tested that committing dummy changes in `itwinui-variables` (e.g. `light.scss`) works.

## Docs

N/A


.stylelintrc.json
mayank99, line 1: This was moved because it was linting CSS that it should not affect (e.g. in website). Suggestion: either duplicate the stylelintrc in both itwinui-css and itwinui-variables, or move it to `internal/configs` and find a way to import it in both places

rohan-kadkol: Ohh, I see, sounds good.

https://github.com/iTwin/iTwinUI/pull/1525/commits/a286b430f197bc6168044c1f6ebff48be4aa2f22: I moved the main `.stylelintrc.json` to `internal/configs`. `itwinui-css` and `itwinui-variables` each have a `.stylelintrc.json` that extends the main `.stylelintrc.json` file.

mayank99: nit: could you see if old formatting could be preserved to minimize the diff? if not, that's ok

mayank99: also could you see if you can reference the `configs` package name instead of the path?

```diff
"extends": [
-  "../../internal/configs/.stylelintrc.json"
+  "configs/.stylelintrc.json"
]
``

mayank99: run `yarn` from the root to avoid lockfile changes

mayank99: > I noticed that the `.stylelintrc.json` was included in `node_modules/configs` without adding the file to `files` in `configs/package.json`. I'm not sure why that is happening. Probably because yarn workspaces include internal packages as the complete folder in the form of a symbolic link (not sure).

still could you add it anyway? it's more correct. we don't want to run into unexpected issues when we switch to pnpm

mayank99 (approved):

.stylelintrc.json
rohan-kadkol: https://github.com/iTwin/iTwinUI/pull/1525/commits/d74adae4ad3391368990ac1b21efdcde037b2fd4: Ohh, makes sense, done.

PR #1502, opened 2023-08-16
restore old `ExpandableBlock` api

mayank99:
## Changes

undid the breaking change from #1354 and brought back `title`/`caption`/`endIcon` props while also supporting the new subcomponents added in that PR.

renamed `ExpandableBlock.Header` to `ExpandableBlock.Trigger` as it makes more sense.

these two snippets are now equivalent:
```jsx
<ExpandableBlock title='click me' caption='or not' endIcon={<svg>...</svg>}>
  content in block
</ExpandableBlock>
```
```jsx
<ExpandableBlock.Wrapper>
  <ExpandableBlock.Trigger>
    <ExpandableBlock.ExpandIcon />
    <ExpandableBlock.LabelArea>
      <ExpandableBlock.Title>click me</ExpandableBlock.Title>
      <ExpandableBlock.Caption>or not</ExpandableBlock.Caption>
    </ExpandableBlock.LabelArea>
    <ExpandableBlock.EndIcon><svg>...</svg></ExpandableBlock.EndIcon>
  </ExpandableBlock.Trigger>
  <ExpandableBlock.Content>content in block</ExpandableBlock.Content>
</ExpandableBlock.Wrapper>
```

## Testing

updated one story and added one unit test.

## Docs

updated old changeset. updated documentation page.

will update migration guide too.

PR #1527, opened 2023-08-29
Tile: add back old api

gretanausedaite:
## Changes

Adding back props API. Now Tile works both ways:

```js
<Tile.Wrapper>
  <Tile.Name name='Best tile ever' />
  <Tile.ThumbnailArea>
    <Tile.ThumbnailPicture url='/local-picture' />
  </Tile.ThumbnailArea>
</Tile.Wrapper>
```

```js
<Tile
  name='Second best tile ever'
  thumbnail='/another-local-file'
/>
```

## Testing

Added unit test

## Docs

Changeset


packages/itwinui-react/src/core/Tile/Tile.test.tsx
```diff
@@ -273,45 +273,121 @@ it('should propagate misc props correctly', () => {
...
+      buttons={[
+        <Button key='button-1'>Button 1</Button>,
+        <Button key='button-2'>Button 2</Button>,
+      ]}
+      moreOptions={[
+        <MenuItem key='menuitem-1'>Item 1</MenuItem>,
+        <MenuItem key='menuitem-1'>Item 2</MenuItem>,
+      ]}
+      leftIcon={
+        <Tile.IconButton>
```
mayank99, line 337: why `IconButton` doesn't work here?

gretanausedaite: It works. Users must pass size='small' button for it to be "as designed" 
left is Tile.IconButton, right is IconButton:
image

+ off topic:
I think we should fix dark icons within those buttons. CSS tests have them white.

mayank99: oh wow, the icons are messed up only in light theme so i didn't even notice 😅

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -397,10 +397,187 @@ TileMoreOptions.displayName = 'Tile.MoreOptions';
...
+  /**
+   * Flag whether the tile is disabled.
+   *
+   * Note: This only affects the tile. You need to manually disable
+   * the buttons and other interactive elements inside the tile.
+   *
+   * @default false
+   */
+  isDisabled?: boolean;
+  onClick?: React.MouseEventHandler<HTMLElement>;
```
mayank99, line 503: add jsdoc here? it's not obvious that this onClick is not applied on the wrapper

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -410,23 +587,40 @@ TileButtons.displayName = 'Tile.Buttons';
...
+ *  rightIcon={<IconButton><SvgStar /></IconButton>}
+ *  isSelected={true}
+ *  isNew={false}
+ * />
  */
 export const Tile = Object.assign(TileComponent, {
+  /**
+   *
+   */
+  Wrapper: TileWrapper,
```
mayank99, line 618: fill jsdocs?

packages/itwinui-react/src/core/Tile/Tile.tsx
```diff
@@ -397,10 +397,187 @@ TileMoreOptions.displayName = 'Tile.MoreOptions';
...
+  isSelected?: boolean;
+  /**
+   * Whether the tile is "new". Tile name becomes bold and gets a new status icon.
+   */
+  isNew?: boolean;
+  /**
+   * Default tile variant or the folder layout.
+   * @default 'default'
+   */
+  variant?: 'default' | 'folder';
```
mayank99, line 479: does folder tile work correctly? i think we had some issues supporting both variants with old apis

packages/itwinui-react/src/core/Tile/Tile.test.tsx
mayank99: for now i would say use `IconButton` with manual `size='small'` in this unit test. we don't want to mix and match old and new apis.

the color issue can be fixed in a separate PR

packages/itwinui-react/src/core/Tile/Tile.tsx
mayank99: just tested in the playground and it's working correctly, even with `onClick`.

looks like it was fixed in https://github.com/iTwin/iTwinUI/pull/1255#issuecomment-1553662141 + https://github.com/iTwin/iTwinUI/pull/1255#issuecomment-1561251594

mayank99 (approved):

rohan-kadkol (commented):
Would be good if there are some examples using the old Tile API. (Similar to https://github.com/iTwin/iTwinUI/pull/1502#pullrequestreview-1587639586)

`examples/Tile.main.tsx` and the other Tile examples that are not currently in the PR are giving type errors. I believe `Tile` should be changed to `Tile.Wrapper` for the other Tile examples where we want to demo the new API.

image

mayank99:
i've updated the missed references. i'll leave it on Greta to handle the rest

rohan-kadkol:
> Would be good if there are some examples using the old Tile API. (Similar to https://github.com/iTwin/iTwinUI/pull/1502#pullrequestreview-1587639586)

@gretanausedaite I noticed you added this to the merge queue. Are we not adding examples using the old Tile API to the website?

gretanausedaite:
> @gretanausedaite I noticed you added this to the merge queue. Are we not adding examples using the old Tile API to the website?

Created #1538

PR #1529, opened 2023-08-29
remove location-marker css

mayank99:
## Changes

not used anywhere yet, so it's safe to remove.

## Testing

n/a

## Docs

n/a


gretanausedaite (approved):

rohan-kadkol (approved):
Are we removing these styles because we don't plan to create a React component for this?

PR #1530, opened 2023-08-29
use regular divs for menuitems

mayank99:
## Changes

Using `<div>`s instead of `<ul>`/`<li>` because there is absolutely no benefit to the latter. We override the `role` for both the menu and menuitem, so `<div>` works just fine and causes fewer problems.

This change was needed because in #1506, I'm exploring setting the role on the popover itself. Doing so makes the html invalid when there's `<li>` elements inside.

Also I believe we need this approach anyway if we ever decide to add grouped menu items.

## Testing

updated any failing unit tests that were querying for `li`. no other visual or behavioral changes.

## Docs

n/a


gretanausedaite (commented):
Do we want to update html tests to not use `li`?

mayank99:
> Do we want to update html tests to not use `li`?

i left it because we don't deliver html and it works with both `<div>` and `<li>`

PR #1531, opened 2023-08-29
add checkmark to selected options

mayank99:
## Changes

Added `<SvgCheckmark />` to selected options in both Select and ComboBox. This makes it consistent with the css demos and makes it easier to tell them apart from focused options.

## Testing

updated visual tests.

## Docs

added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/Select/Select.test.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.test.tsx
@@ -38,7 +38,9 @@ function assertMenu(
   expect(menuItems.length).toBe(3);
   menuItems.forEach((item, index) => {
     expect(item.textContent).toContain(`Test${index}`);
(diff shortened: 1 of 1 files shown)


packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -310,6 +311,7 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
         key: `${label}-${index}`,
         isSelected,
         startIcon: startIcon,
+        endIcon: isSelected ? <SvgCheckmark aria-hidden /> : null,
```
gretanausedaite, line 314: Unit tests for this?

mayank99: Ahh idk, comparing svg nodes within jest always feels pointless. We do have visual tests, so I figured it would be enough

rohan-kadkol (approved):

PR #1535, opened 2023-08-29
fix tooltip reference prop

mayank99:
## Changes

noticed that aria props were not being applied when using `reference` because it is not something we render. fixed by manually copying aria props in a useEffect

## Testing

verified in playground

### DIFF
--- a/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
+++ b/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
@@ -14,7 +14,6 @@ import {
   useHover,
   useFocus,
   useDismiss,
-  useRole,
   useInteractions,
   safePolygon,
   size,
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -146,44 +141,86 @@ const useTooltip = (options: TooltipOptions = {}) => {
...
+
+  React.useEffect(() => {
+    if (!reference) {
+      return;
+    }
+
+    const oldValues: Record<string, string | null> = {}; // for cleanup
+    Object.entries(ariaProps).forEach(([key, value]) => {
+      oldValues[key] = reference.getAttribute(key);
+      reference.setAttribute(key, value);
```
mayank99, line 165: @gretanausedaite I was able to fix focus event handlers for `reference` prop using this code. What do you think?

```diff
+ /** e.g. onPointerDown --> pointerdown */
+ const domEventName = (e: string) => e.toLowerCase().substring(2);

  const oldValues: Record<string, string | null> = {};
  Object.entries({ ...ariaProps, ...interactions.getReferenceProps() }).forEach(([key, value]) => {
+   if (typeof value === 'function') {
+     reference.addEventListener(domEventName(key), value);
    } else if (value) {
      oldValues[key] = reference.getAttribute(key);
      reference.setAttribute(key, value);
    }
  });
```

mayank99: [`da3c1f5`](https://github.com/iTwin/iTwinUI/pull/1535/commits/da3c1f5809d8d0c1f77bf12a55070e4bbaeff6a3)

PR #1537, opened 2023-08-30
add `onVisibleChange` prop to Tooltip

mayank99:
## Changes

This prop is useful when the user needs to access the current visible state but doesn't want to reimplement all hover/focus behavior that come with uncontrolled mode.

More importantly, this exact same implementation will be used in #1506 to access current visible state in several places.

## Testing

Updated 

Also tested controlled and uncontrolled modes in vite playground:

<details>
<summary>App.tsx</summary>

```tsx
import { Button, Tooltip } from '@itwin/itwinui-react';

export default () => {
  const [open, setOpen] = React.useState(false);

  return (
    <>
      <Tooltip content='hello'>
        <Button>uncontrolled</Button>
      </Tooltip>

      <Tooltip content='world' visible={open} onVisibleChange={setOpen}>
        <Button>controlled</Button>
      </Tooltip>
    </>
  );
};
```

</details>

## Docs

n/a. I considered adding a changeset, but Tooltip already has had all its props changed from the last version so maybe doesn't make sense to add more notes.


packages/itwinui-react/src/core/utils/hooks/useControlledState.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const useControlledState = <T>(
+  initialValue: T,
+  controlledState: T,
+  setControlledState?: React.Dispatch<React.SetStateAction<T>>,
+) => {
+  const [uncontrolledState, setUncontrolledState] =
+    React.useState<T>(initialValue);
+
+  const state = controlledState ?? uncontrolledState;
```
rohan-kadkol, line 26: Just wondering, what happens if we do `useControlledState(10, null)`? In this case, we want to use the value `null` since `controlledState` takes preference over `uncontrolledState`/`initialValue`.

But `controlledState ?? uncontrolledState` = `null ?? 10` = `10` (and not `null`).

Is this expected behavior? Or should we use `uncontrolledState` instead of `controlledState` only when `controlledState` is `undefined` (not passed by the user, etc.)?

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -178,27 +188,17 @@ const useTooltip = (options: TooltipOptions = {}) => {
 
   const { delay } = useDelayGroupContext();
 
-  const hover = useHover(floating.context, {
-    enabled: controlledOpen == null,
```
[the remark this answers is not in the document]
mayank99: if we don't enable it, controlled mode will not work. `onVisibleChange` will not be called for hover/focus

packages/itwinui-react/src/core/utils/hooks/useControlledState.ts
mayank99: sure, changed.

gretanausedaite (approved):
No questions from my side.

packages/itwinui-react/src/core/utils/hooks/useControlledState.ts
rohan-kadkol: Would be great is this can be clarified in the JSDocs too. Like what does the "preference" stand for ([here](https://github.com/iTwin/iTwinUI/pull/1537/files#diff-c9341964923594e472777ba3dbbaa4073dea52a6608a4456954eb1725f8d1c4bR8-R9)). Because preference could mean non-null, truthy, etc.

rohan-kadkol (approved):
The suggestion in https://github.com/iTwin/iTwinUI/pull/1537#discussion_r1311729168 is optional.

PR #1541, opened 2023-08-31
added label to TransferListMain example.

siddhantrawal:
## Changes

Added area-label to TransferListMain example. Fixed a11y violation.

## Testing

Before:
Screenshot 2023-08-31 102926

After:
Screenshot 2023-08-31 120016

## Docs

N/A


examples/TransferList.main.tsx
```diff
@@ -56,7 +56,7 @@ export default () => {
   return (
     <TransferList style={{ width: 500 }}>
       <TransferList.ListboxWrapper>
-        <TransferList.Listbox>
+        <TransferList.Listbox aria-label='listbox'>
```
mayank99, line 59: don't use `aria-label`. instead, add `TransferList.Label`.

mayank99: can you also change the label names to be more relevant?

this example is very contrived, but it would still help if it said "Column A"/"Column B" instead of just "Listbox".

mayank99 (approved):

rohan-kadkol (commented):
There is no difference between the TransferList.main and TransferList.withlabel. Also, if the label is required, then "With Label" is no longer a *variant*; it is the *default*.

@mayank99, was there a reason why these two nearly identical examples existed? Were we planning to document more variants in the future? If there's no special reason, let's remove this TransferList.withLabel along with the "Variants" headings.

rohan-kadkol (approved):
The issue I mentioned (https://github.com/iTwin/iTwinUI/pull/1541#pullrequestreview-1607130145) can be tackled in another PR

mayank99:
> There is no difference between the TransferList.main and TransferList.withlabel. Also, if the label is required, then "With Label" is no longer a _variant_; it is the _default_.

Agreed, we can remove the "withlabel" example, and document that labels should always be provided.

PR #1543, opened 2023-09-01
Removed `{...args}` from Table's Storybook file

rohan-kadkol:
## Changes

Solves part of #886 . Helper PR for #1514 .

Passing `{...args}` to `<Table>` in Table stories could hide TS errors (as mentioned in #886). Thus:

* Removed `{...args}` and Storybook controls from all Table Storybook stories.

## Testing

Tested that all stories are still working as before.

## Docs

N/A

### DIFF
--- a/apps/storybook/src/Table.stories.tsx
+++ b/apps/storybook/src/Table.stories.tsx
@@ -37,7 +37,7 @@ import {
   ProgressRadial,
   BaseFilter,
 } from '@itwin/itwinui-react';
-import { Story, Meta } from '@storybook/react';
+import { Meta } from '@storybook/react';
 import { useMemo, useState } from '@storybook/addons';
 import { action } from '@storybook/addon-actions';
 import {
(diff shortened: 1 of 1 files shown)


apps/storybook/src/Table.stories.tsx
```diff
@@ -52,55 +52,9 @@ import {
...
-    manualFilters: { table: { disable: true } },
-    paginateExpandedRows: { table: { disable: true } },
-    expandSubRows: { table: { disable: true } },
-    autoResetExpanded: { table: { disable: true } },
-    manualExpandedKey: { table: { disable: true } },
-  },
 } as Meta<TableProps>;
 
-export const Basic: Story<Partial<TableProps>> = (args) => {
+export const Basic: Story<Partial<TableProps>> = () => {
```
mayank99, line 57: The type `Story<Partial<TableProps>>` can also be removed

rohan-kadkol: https://github.com/iTwin/iTwinUI/pull/1543/commits/45ca69b7b9a46a7970122c711495b0f39bedd826: Removed `Story<Partial<TableProps>>`

https://github.com/iTwin/iTwinUI/pull/1543/commits/a5bf4c7dad28171acc9c927d82508fe2f2cce684: I also removed `as Meta<TableProps>` since I believe that has something to do with assigning the correct type to `args`, and thus is no longer needed.

mayank99 (approved):

PR #1548, opened 2023-09-07
Tabs: composition

gretanausedaite:
## Changes

WIP

## Testing

WIP

## Docs

WIP


apps/storybook/src/Tabs.stories.tsx
```diff
@@ -2,198 +2,263 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+/* eslint-disable react/jsx-key */
```
mayank99, line 5: why disable this?

examples/Tabs.main.tsx
```diff
@@ -3,33 +3,44 @@
...
-    </div>
+    <Tabs.Wrapper>
+      <Tabs.TabList>
+        <Tabs.Tab value='item1' label='Item1' />
+        <Tabs.Tab value='item2' label='Item2' />
+        <Tabs.Tab value='item3' label='Item3' />
+      </Tabs.TabList>
+
+      <Tabs.Panel value='item1'>
+        Bentley Systems, Incorporated, is an American-based software development
```
mayank99, line 18: nit: can we use more generic text? i think we were supposed to remove all company-specific content when we went open source, but somehow missed this one

better idea: use fruit names as tabs, with a description of the fruit as panel content

examples/Tabs.overflow.tsx
```diff
@@ -3,57 +3,78 @@
...
-    <Tab key={3} label='Uses' />,
-    <Tab key={4} label='Reinvestment' />,
-    <Tab key={5} label='Location' />,
-    <Tab key={6} label='Revenue' />,
-    <Tab key={7} label='Founders' />,
-    <Tab key={8} label='History' />,
-  ];
+  const [active, setActive] = React.useState('Bentley Overview');
+
+  const [tabData] = React.useState([
```
mayank99, line 11: doesn't need to be state, it's just data. can be defined outside of component.

also the panel content could be defined in the same place too.

examples/Tabs.overflow.tsx
```diff
@@ -3,57 +3,78 @@
...
-      >
-        {getContent()}
-      </Tabs>
+    <div style={{ maxWidth: 425, border: '1px solid lightpink', padding: 8 }}>
+      <Tabs.Wrapper overflowOptions={{ useOverflow: true }}>
+        <Tabs.TabList>
+          {tabData?.map((item, index) => {
+            return (
+              <Tabs.Tab
+                value={`tab${index}`}
```
mayank99, line 29: can we use name instead of index as the value?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+  const { value, className, children, ...rest } = props;
+
+  const { activeValue } = useSafeContext(TabsContext);
+
+  if (value === activeValue) {
+    return (
+      <Box
+        className={cx('iui-tabs-content', className)}
+        role='tabpanel'
+        id={value}
```
mayank99, line 718: we could prefix this value with a unique id (unique and shared throughout the whole Tabs) in case the user uses two sets of tabs with overlapping ids.

also i believe we will need to account for the user passing `id` directly

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+      >
+        {labels.map((label, index) => {
+          const tabIndex = `label-${index}`;
+          return React.isValidElement(label) ? (
+            React.cloneElement(label as JSX.Element, {
+              value: tabIndex,
+              active: activeIndex === index || tabIndex === currentActiveValue,
+              onClick: (args: unknown) => {
+                onTabSelected?.(index);
+                setCurrentActiveValue(tabIndex);
```
mayank99, line 824: i don't think this `onClick` should be needed. it should be automatically handled in Tab. i believe the `onActivated` prop is what's missing

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -451,36 +455,116 @@ export const Tabs = (props: TabsProps) => {
...
+  label?: string | React.ReactNode;
+  /**
+   * Flag whether the tab is active.
+   * @default 'false'
+   */
+  isActive?: boolean;
+  /**
+   * Callback fired when the isActive prop changes.
+   */
+  onActivated?: () => void;
```
mayank99, line 514: bikeshed: `onActivated` vs `onActiveChange`?

for comparison: Popover and Tooltip have `onVisibleChange`

apps/storybook/src/Tabs.stories.tsx
```diff
@@ -2,198 +2,263 @@
...
+        (Interactive Graphics Design System) installations. Their first product
+        was shown to potential users who were polled as to what they would be
+        willing to pay for it. They averaged the answers, arriving at a price of
+        $7,943. A DOS-based version of MicroStation was introduced in 1986.
+      </Tabs.Panel>
+    </Tabs.Wrapper>
   );
 };
+BorderlessTabs.args = {
+  type: 'borderless',
```
mayank99, line 112: `args` doesn't do anything anymore

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+const TabsPanel = React.forwardRef((props, ref) => {
+  const { value, className, children, ...rest } = props;
+
+  const { activeValue } = useSafeContext(TabsContext);
+
+  if (value === activeValue) {
+    return (
+      <Box
+        className={cx('iui-tabs-content', className)}
+        role='tabpanel'
```
mayank99, line 717: should panel also have `aria-labelledby` pointing back to the `id` of the tab that it's associated with?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+   */
+  wrapperClassName?: string;
+  /**
+   * Content inside the tab panel.
+   */
+  children?: React.ReactNode;
+} & TabsOrientationProps &
+  TabsOverflowProps;
+
+const TabsComponent = React.forwardRef((props, forwardedRef) => {
```
mayank99, line 784: should we deprecate old api? i believe it was failing a11y

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+   */
+  value: string;
+};
+
+const TabsPanel = React.forwardRef((props, ref) => {
+  const { value, className, children, ...rest } = props;
+
+  const { activeValue } = useSafeContext(TabsContext);
+
+  if (value === activeValue) {
```
mayank99, line 713: i think it might better to keep all tabs in the DOM. we can hide the non-active ones with `display: none` for example

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -2,21 +2,26 @@
...
-  useIsClient,
   useResizeObserver,
-  Box,
+  ButtonBase,
 } from '../utils/index.js';
-import { Tab } from './Tab.js';
+import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import styles from '../../styles.js';
 
-export type OverflowOptions = {
```
rohan-kadkol, line 19: Any reason we are not exporting `OverflowOptions` anymore?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+    setStripeProperties,
+  } = useSafeContext(TabsContext);
+
+  const items = React.useMemo(
+    () => (Array.isArray(children) ? children : [children]),
+    [children],
+  );
 
-  const tablistRef = React.useRef<HTMLUListElement>(null);
+  const activeValue = activeValueProp || activeValueContext;
```
rohan-kadkol, line 172: Why is `activeValue` needed from both the context and the prop to `TabsList`?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -603,8 +849,233 @@ export const Tabs = (props: TabsProps) => {
...
+   */
+  Panel: TabsPanel,
+});
+
+export const TabsContext = React.createContext<
+  | {
+      /**
+       * Type of the tabs.
+       */
+      type?: string;
```
rohan-kadkol, line 1026: Does the reason why this is a `string` and not actual values have something to do with the `type` values can vary based on the orientation?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+        return;
+      }
+    });
+  }, [items, focusedValue]);
 
   // CSS custom properties to place the active stripe
-  const [stripeProperties, setStripeProperties] = React.useState({});
   useIsomorphicLayoutEffect(() => {
-    if (type !== 'default' && tablistRef.current != undefined) {
+    console.log(activeValue);
```
rohan-kadkol, line 224: nit: Creating a comment so we don't forget to delete it before merging

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -280,7 +291,7 @@ export const Tabs = (props: TabsProps) => {
     ownerDoc.addEventListener('wheel', enableHorizontalScroll);
   }, [overflowOptions?.useOverflow, orientation, enableHorizontalScroll]);
 
-  const isTabHidden = (activeTab: HTMLElement, isVertical: boolean) => {
+  const isTabHidden = (tab: HTMLElement, isVertical: boolean) => {
```
rohan-kadkol, line 294: nit: Might be helpful to add JSDocs for this. E.g. to explain the three return values.

packages/itwinui-css/backstop/tests/information-panel.html
```diff
@@ -1380,44 +1380,38 @@ <h1>Information panel</h1>
               class="iui-tabs-wrapper iui-horizontal"
               id="demo-pill-2"
             >
-              <ul
+              <div
                 class="iui-tabs iui-pill iui-not-animated"
                 role="tablist"
                 aria-label="Tab group name"
               >
-                <li>
```
rohan-kadkol, line 1388: Is there a semantic loss when not using `<ul>`s and `<li>`s?

packages/itwinui-css/backstop/tests/tabs.html
```diff
@@ -178,96 +160,78 @@ <h2>Default</h2>
...
+            role="tab"
+            aria-selected="true"
+            aria-controls="active-tab"
+          >
+            <svg-placeholder
+              class="iui-tab-icon"
+              aria-hidden="true"
+            ></svg-placeholder>
+            <span class="iui-tab-label">Active</span>
+            <span class="iui-tab-description">DESCRIPTION</span>
```
rohan-kadkol, line 179: Should be `Sublabel`?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -35,12 +40,13 @@ type TabsOverflowProps =
       /**
        * Type of the tabs.
        *
-       * If `orientation = 'vertical'`, `pill` is not applicable.
+       * If `type = 'pill' | 'borderless'`, `overflowOptions` is not applicable.
        * @default 'default'
```
rohan-kadkol, line 44: nit: Looks like `@default` appears twice when using the prop because VSCode combines both the JSDocs. Maybe we can remove one `@default` if it's not needed.

image

.changeset/strong-owls-sleep.md
```diff
@@ -0,0 +1,29 @@
+---
+'@itwin/itwinui-css': major
+---
+
+The `iui-tabs` element has changed from `<ul>` to `<div>`.
+`iui-tabs` children are no longer wrapped in an `<li>` element. 
+The `iui-tab-label` wrapper class has been removed. 
+The `iui-tab-label` class now applies to the `<span>` which holds the tab's label. 
+`iui-tab-description` is now a `<span>` element.
```
rohan-kadkol, line 9: nit: A list would be better

packages/itwinui-css/src/tabs/tabs.scss
```diff
@@ -17,10 +17,6 @@
   }
 }
 
-.iui-tabs-actions-wrapper {
```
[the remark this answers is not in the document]
gretanausedaite: Yes, it was intentionally removed.
I like it centered in the area better than having bigger gap to the panel.

I can add it back if we want to keep the old way.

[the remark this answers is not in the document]
mayank99: it still becomes centered when there is enough space available. the bigger gap is only used when space is tight.

it looks even worse in the case of default tabs
| Before | Now |
| --- | --- |
|  |  |

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -73,22 +80,27 @@ $borderless-horizontal-tab-min-height: calc(
     &,
     svg {
       @include mixins.iui-icon-style('m');
```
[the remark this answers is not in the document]
gretanausedaite: Using Icon

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -32,14 +32,21 @@ $borderless-horizontal-tab-min-height: calc(
     @include mixins.iui-reset;
```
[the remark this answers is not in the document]
gretanausedaite: Removed styles

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -228,27 +238,28 @@ export const Tabs = (props: TabsProps) => {
...
-    setHasSublabel(
-      type !== 'pill' && // pill tabs should never have sublabels
-        !!tablistRef.current?.querySelector(
-          `.${styles['iui-tab-description']}`, // check directly for the sublabel class
-        ),
-    );
-  }, [type]);
+  }, [focusedValue, onTabClick, focusActivationMode]);
 
   const enableHorizontalScroll = React.useCallback((e: WheelEvent) => {
```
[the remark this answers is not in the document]
gretanausedaite: Using scrollIntoView inside Tab and deleted most of that complex code. Keeping mouse scroll behaviour takeover.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+const TabDescription = React.forwardRef((props, ref) => {
+  const { className, children, ...rest } = props;
+  const { type } = useSafeContext(TabsContext);
+  const { hasSublabel, setHasSublabel } = useSafeContext(TabListContext);
+
+  useIsomorphicLayoutEffect(() => {
+    type !== 'pill' && !hasSublabel && setHasSublabel && setHasSublabel(true);
+  }, []);
+
+  if (type !== 'pill') {
```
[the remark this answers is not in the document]
gretanausedaite: Updated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
-      )}
+// ----------------------------------------------------------------------------
+// Tabs.Actions component
+
+const TabsActions = React.forwardRef((props, ref) => {
+  const { className, children, ...rest } = props;
+
+  const { type } = useSafeContext(TabsContext);
+
+  if (type !== 'pill') {
```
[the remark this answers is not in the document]
gretanausedaite: Updated

packages/itwinui-css/src/tabs/base.scss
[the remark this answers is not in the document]
mayank99: also html needs to be updated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -526,74 +610,236 @@ export const Tabs = (props: TabsProps) => {
...
+// ----------------------------------------------------------------------------
+// Tabs.TabDescription component
+
+const TabDescription = React.forwardRef((props, ref) => {
+  const { className, children, ...rest } = props;
+  const { type } = useSafeContext(TabsContext);
+  const { hasSublabel, setHasSublabel } = useSafeContext(TabListContext);
+
+  useIsomorphicLayoutEffect(() => {
+    type !== 'pill' && !hasSublabel && setHasSublabel && setHasSublabel(true);
```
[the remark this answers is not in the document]
mayank99: you still need a `useEffect`.

see https://github.com/iTwin/iTwinUI/pull/1236

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
gretanausedaite: Will undo this change then

apps/storybook/src/Tabs.stories.tsx
gretanausedaite: No reason, removed.

examples/Tabs.main.tsx
gretanausedaite: Using generic text

mayank99: `value` can also be changed to `apple`, `orange`, etc

packages/itwinui-css/src/tabs/tabs.scss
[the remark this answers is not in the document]
gretanausedaite: Undid

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
 
-  const [currentActiveIndex, setCurrentActiveIndex] = React.useState(() =>
-    activeIndex != null
-      ? getBoundedValue(activeIndex, 0, labels.length - 1)
-      : 0,
-  );
   useIsomorphicLayoutEffect(() => {
-    if (activeIndex != null && currentActiveIndex !== activeIndex) {
-      setCurrentActiveIndex(getBoundedValue(activeIndex, 0, labels.length - 1));
+    // Should only call on the first render since activeValue cannot be initialized in TabsWrapper
```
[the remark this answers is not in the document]
gretanausedaite: Removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+      }
+    } else {
+      React.isValidElement(items[0]) &&
+        setActiveValue(items[0].props.value as string);
     }
-  }, [activeIndex, currentActiveIndex, labels.length]);
+  }, []);
+
+  // When activeValue change
+  useIsomorphicLayoutEffect(() => {
```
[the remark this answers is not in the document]
gretanausedaite: removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+    items.forEach((item, index) => {
+      if (React.isValidElement(item) && item.props.value === activeValue) {
+        newActiveIndex.current = index;
+        return;
+      }
+    });
+  }, [items, activeValue]);
+
+  // When focusedValue change
+  useIsomorphicLayoutEffect(() => {
```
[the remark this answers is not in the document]
gretanausedaite: removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -228,27 +238,28 @@ export const Tabs = (props: TabsProps) => {
...
+    [items, setActiveValue],
+  );
 
-  const [focusedIndex, setFocusedIndex] = React.useState<number | undefined>();
   React.useEffect(() => {
-    if (tablistRef.current && focusedIndex !== undefined) {
-      const tab = tablistRef.current.querySelectorAll(`.${styles['iui-tab']}`)[
-        focusedIndex
-      ];
+    if (tablistRef.current && newFocusedIndex.current !== undefined) {
```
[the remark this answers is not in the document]
gretanausedaite: removed.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
gretanausedaite: Oopsie, added.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
gretanausedaite: Renamed

apps/storybook/src/Tabs.stories.tsx
gretanausedaite: Removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: I think we are not exporting any types. So not exporting this one too.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated the type

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Should be better

examples/Tabs.main.tsx
gretanausedaite: Updated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Changed to onActiveChanged. Sounds better.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Should be better.

examples/Tabs.overflow.tsx
gretanausedaite: Updated examples

examples/Tabs.overflow.tsx
gretanausedaite: Updated examples

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Added data attribute to add display none. Is this approach okay to you?

mayank99: you could use [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute, and in css add `&[hidden] { display: none !important; }`.

the `!important` will help prevent users from accidentally overriding it.

examples/Tabs.main.tsx
mayank99: why `apple1`, `orange1`?

the uniqueness can be fixed by https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1336392436

packages/itwinui-css/backstop/tests/information-panel.html
mayank99: nope. the list does nothing except get in the way.

we are overriding the `role` anyway, so it doesn't matter what element we use underneath.

examples/Tabs.overflow.tsx
mayank99: why not just `item.name` instead of `tab${item.name}2`? see https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1342689004 as well.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+      if (React.isValidElement(item) && item.props.value === focusedValue) {
+        newFocusedIndex.current = index;
+        return;
+      }
+    });
+  }, [items, focusedValue]);
 
   // CSS custom properties to place the active stripe
-  const [stripeProperties, setStripeProperties] = React.useState({});
   useIsomorphicLayoutEffect(() => {
```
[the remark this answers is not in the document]
mayank99: not done. it's still in a useEffect

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: it's only done in legacy api, missing in `Tabs.Panel`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -451,36 +206,158 @@ export const Tabs = (props: TabsProps) => {
...
+      } else if (delta === -1) {
+        nextTab = nextTab?.previousElementSibling ?? allTabs?.at(-1);
+      }
+
+      if (nextTab instanceof HTMLElement) {
+        nextTab?.focus();
+
+        if (focusActivationMode === 'auto') {
+          ReactDOM.flushSync(() => {
+            onActiveChange?.(); // TODO: fix this (stale state)
```
mayank99, line 357: this can be fixed using `onFocus`.

<details>
<summary>diff</summary>

```diff
 import React from 'react';
-import * as ReactDOM from 'react-dom';
 import cx from 'classnames';
 import {
   useSafeContext,
@@ -350,13 +349,7 @@ const TabHeader = React.forwardRef((props, forwardedRef) => {
       }
 
       if (nextTab instanceof HTMLElement) {
-        nextTab?.focus();
-
-        if (focusActivationMode === 'auto') {
-          ReactDOM.flushSync(() => {
-            onActiveChange?.(); // TODO: fix this (stale state)
-          });
-        }
+        nextTab.focus();
       }
     };
 
@@ -415,6 +408,11 @@ const TabHeader = React.forwardRef((props, forwardedRef) => {
       tabIndex={isActive ? 0 : -1}
       onClick={mergeEventHandlers(onClickProp, () => onActiveChange?.())}
       onKeyDown={onKeyDown}
+      onFocus={() => {
+        if (focusActivationMode === 'auto') {
+          onActiveChange?.();
+        }
+      }}
       aria-selected={isActive}
       aria-controls={value}
       ref={useMergedRefs(tabRef, forwardedRef, setInitialActiveRef)}
```

</details>

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -512,99 +389,506 @@ export const Tabs = (props: TabsProps) => {
...
+    tabsClassName,
+    contentClassName,
+    wrapperClassName,
+    children,
+    ...rest
+  } = props;
+
+  const [currentActiveValue, setCurrentActiveValue] = React.useState<string>();
+
+  // TODO: fix legacy active logic
```
mayank99, line 608: this can be fixed using `onActiveChange` and `useControlledState`

<details>
<summary>diff</summary>

```diff
@@ -15,6 +15,7 @@ import {
   useResizeObserver,
   ButtonBase,
   mergeEventHandlers,
+  useControlledState,
 } from '../utils/index.js';
 import { Icon } from '../Icon/Icon.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
@@ -593,7 +594,7 @@ const LegacyTabsComponent = React.forwardRef((props, forwardedRef) => {
     onTabSelected,
     focusActivationMode,
     color,
-    activeIndex,
+    activeIndex: activeIndexProp,
     tabsClassName,
     contentClassName,
     wrapperClassName,
@@ -601,16 +602,11 @@ const LegacyTabsComponent = React.forwardRef((props, forwardedRef) => {
     ...rest
   } = props;
 
-  const [currentActiveValue, setCurrentActiveValue] = React.useState<string>();
-
-  // TODO: fix legacy active logic
-  const findActiveIndex = () => {
-    return (
-      labels.findIndex(
-        (label, index) => currentActiveValue === `label-${index}`,
-      ) ?? `label-0`
-    );
-  };
+  const [activeIndex, setActiveIndex] = useControlledState(
+    0,
+    activeIndexProp,
+    onTabSelected,
+  );
 
   return (
     <TabsWrapper className={wrapperClassName} {...rest}>
@@ -618,32 +614,23 @@ const LegacyTabsComponent = React.forwardRef((props, forwardedRef) => {
         color={color}
         className={tabsClassName}
         focusActivationMode={focusActivationMode}
-        // activeValue={currentActiveValue}
-        // setActiveValue={(value) => setCurrentActiveValue(value)}
         ref={forwardedRef}
       >
         {labels.map((label, index) => {
-          const tabValue = `label-${index}`;
+          const tabValue = `tab-${index}`;
           return React.isValidElement(label) ? (
             React.cloneElement(label as JSX.Element, {
               value: tabValue,
-              active: activeIndex === index || tabValue === currentActiveValue,
-              onClick: (args: unknown) => {
-                onTabSelected?.(index);
-                setCurrentActiveValue(tabValue);
-                label.props.onClick?.(args);
-              },
+              active: index === activeIndex,
+              onActiveChange: () => setActiveIndex(index),
             })
           ) : (
             <LegacyTab
               key={index}
               value={tabValue}
               label={label}
-              active={activeIndex === index || tabValue === currentActiveValue}
-              onClick={() => {
-                onTabSelected?.(index);
-                setCurrentActiveValue(tabValue);
-              }}
+              active={index === activeIndex}
+              onActiveChange={() => setActiveIndex(index)}
             />
           );
         })}
@@ -652,13 +639,9 @@ const LegacyTabsComponent = React.forwardRef((props, forwardedRef) => {
       {actions && <TabsActions>{actions}</TabsActions>}
 
       {children && (
-        <Box // TODO: use Tabs.Panel
-          className={cx('iui-tabs-content', contentClassName)}
-          role='tabpanel'
-          aria-labelledby={`label-${findActiveIndex()}`}
-        >
+        <TabsPanel value={`tab-${activeIndex}`} className={contentClassName}>
           {children}
-        </Box>
+        </TabsPanel>
       )}
     </TabsWrapper>
   );
@@ -694,6 +677,12 @@ type TabLegacyProps = {
    * This will be automatically set by the parent `Tabs` component.
    */
   active?: boolean;
+  /**
+   * Callback fired when the tab becomes active.
+   *
+   * This will be automatically set by the parent `Tabs` component.
+   */
+  onActiveChange?: () => void;
 };
```

</details>

packages/itwinui-css/backstop/tests/tabs.html
gretanausedaite: Updated to sublabel

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Using your suggestion

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Using your suggestion

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -414,12 +396,21 @@ const TabHeader = React.forwardRef((props, forwardedRef) => {
...
       role='tab'
       tabIndex={isActive ? 0 : -1}
-      onClick={mergeEventHandlers(onClickProp, () => onActiveChange?.())}
-      onKeyDown={onKeyDown}
       aria-selected={isActive}
       aria-controls={value}
       ref={useMergedRefs(tabRef, forwardedRef, setInitialActiveRef)}
       {...rest}
+      onClick={mergeEventHandlers(props.onClick, () => {
+        if (!props.disabled) {
```
mayank99, line 411: i think this check shouldn't be needed, as `ButtonBase` already checks for `disabled` in `onClick`

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -451,160 +206,676 @@ export const Tabs = (props: TabsProps) => {
...
+    type,
+    setStripeProperties,
+  } = useSafeContext(TabsContext);
+  const { focusActivationMode, tabsWidth } = useSafeContext(TabListContext);
+  const tabRef = React.useRef<HTMLButtonElement>();
+
+  const isActive =
+    isActiveProp !== undefined ? isActiveProp : activeValueState === value;
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
mayank99: but now this will be called every time `isActive` is set to true, causing unnecessary state updates

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: It did not. I'll try again and do screen recording if failing.

gretanausedaite: It selects tab with panel even when disabled without those checks


https://github.com/iTwin/iTwinUI/assets/83585998/c387ebf7-be4b-4745-a699-410a477a7b08

mayank99: are you sure? i can't reproduce.

note that i'm only talking about `onClick` (which affects `focusActivationMode='manual'`). it is fine to keep in `onFocus`.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+
+  const isClient = useIsClient();
+  const tablistRef = React.useRef<HTMLDivElement>(null);
   const [tablistSizeRef, tabsWidth] = useContainerWidth(type !== 'default');
-  const refs = useMergedRefs(tablistRef, tablistSizeRef);
+  const refs = useMergedRefs(ref, tablistRef, tablistSizeRef);
+  const newActiveIndex = React.useRef(0);
+
+  const [focusedValue, setFocusedValue] = React.useState<string | undefined>();
+  const newFocusedIndex = React.useRef<number | undefined>();
```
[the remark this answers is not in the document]
gretanausedaite: Not needed. It was removed with some updated.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -177,46 +86,147 @@ export const Tabs = (props: TabsProps) => {
...
+    if (type !== 'default' && tablistRef.current != undefined && activeValue) {
       const activeTab = tablistRef.current.children[
-        currentActiveIndex
+        newActiveIndex.current
       ] as HTMLElement;
       const activeTabRect = activeTab.getBoundingClientRect();
-
       setStripeProperties({
         ...(orientation === 'horizontal' && {
           '--stripe-width': `${activeTabRect.width}px`,
```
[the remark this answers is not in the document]
gretanausedaite: Updated to have prefix

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: I believe this was fixed during pair programming session.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Yup, you are right, removed condition.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+  /**
+   * Content inside the tab panel.
+   */
+  children?: React.ReactNode;
+  /**
+   * @deprecated Tabs can be scrollable by default
+   *
+   * Options that can be specified to deal with tabs overflowing the allotted space.
+   */
+  overflowOptions?: {
```
gretanausedaite, line 503: Keeping this in legacy props as deprecated to not break build if users have this prop in their code.

packages/itwinui-react/src/core/Tabs/Tabs.test.tsx
```diff
@@ -237,79 +282,185 @@ it.each(['horizontal', 'vertical'] as const)(
...
-  fireEvent.keyDown(tablist, { key: 'ArrowRight' });
-  expect(mockOnTabSelected).not.toBeCalled();
+  fireEvent.keyDown(tabs[0], { key: 'ArrowRight' });
+  expect(mockonActivated).not.toBeCalled();
   expect(document.activeElement).toBe(tabs[1]);
 
   // select 1
-  fireEvent.keyDown(tablist, { key: 'Enter' });
-  expect(mockOnTabSelected).toBeCalledWith(1);
+  fireEvent.keyDown(tabs[1], { key: 'Enter' });
```
gretanausedaite, line 347: Would appreciate some help with this test. I don't know why jest is not clicking enter/activating the tab - it works in dev server. Maybe something in tests?

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -138,31 +142,28 @@ $borderless-horizontal-tab-min-height: calc(
       min-block-size: $borderless-horizontal-tab-min-height;
     }
 
-    &[data-iui-overflow='true'] {
```
gretanausedaite, line 141: We want to remove this data attribute but it's used in setting white-space nowrap in overflow scenarios.
image

We are using wrapping with long labels and descriptions:
image

Is it better to remove this nowrap and have tabs like this when overflowing: 
image

Or add it to tabs in general and have long labels not wrapping:
image

Or keep data attribute?

packages/itwinui-react/src/core/Tabs/Tabs.test.tsx
gretanausedaite: Fixed.

packages/itwinui-css/src/tabs/base.scss
rohan-kadkol: I usually prefer the unwrapped version with something like a [`title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#result_2)/tooltip upon hovering. Maybe we keep the unwrapped as the default and offer a prop to customize this?

examples/Tabs.main.tsx
```diff
@@ -3,33 +3,32 @@
...
-          <Tab key={1} label='Item1' />,
-          <Tab key={2} label='Item2' />,
-          <Tab key={3} label='Item3' />,
-        ]}
-        onTabSelected={setIndex}
-      >
-        {getContent()}
-      </Tabs>
-    </div>
+    <Tabs.Wrapper>
```
rohan-kadkol, line 10: Do we strongly recommend the composition API for Tabs? If we equally suggest both APIs, maybe good to have a few examples with the old APIs.

examples/Tabs.overflow.tsx
```diff
@@ -3,57 +3,72 @@
...
-    <Tab key={1} label='Bentley Overview' />,
-    <Tab key={2} label='Markets' />,
-    <Tab key={3} label='Uses' />,
-    <Tab key={4} label='Reinvestment' />,
-    <Tab key={5} label='Location' />,
-    <Tab key={6} label='Revenue' />,
-    <Tab key={7} label='Founders' />,
-    <Tab key={8} label='History' />,
-  ];
+  const [active, setActive] = React.useState('Bentley Overview');
```
rohan-kadkol, line 48: Leftover?

.changeset/strong-owls-sleep.md
gretanausedaite: Updated changesets to be in separate files.

.changeset/big-rice-live.md
rohan-kadkol, line 1: Is there an advantage to having multiple changesets for the same package and update bump severity (patch, minor, major)?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
-  if (
-    props.type !== 'borderless' &&
-    props.type !== 'pill' &&
-    props.overflowOptions
-  ) {
-    overflowOptions = props.overflowOptions;
-    props = { ...props };
-    delete props.overflowOptions;
-  }
+  defaultChecked?: never; // To remove `defaultChecked` from `<div>` props.
```
rohan-kadkol, line 79: Can make this a TODO comment for visibility

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
-              onClick: (args: unknown) => {
-                onClick();
-                label.props.onClick?.(args);
-              },
-            })
-          )}
-        </li>
-      );
+  // use first tab as active if no `isActive` passed.
+  const setInitialActiveRef = React.useCallback(
```
rohan-kadkol, line 304: Can this not be called in a `useEffect`? Is there an advantage with passing it in a `useCallback()` and passing that to `useMergedRefs`?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: It's not a todo item. This prop IS removing defaultChecked from div props. Maybe comment needs to be updated to full sentence.

.changeset/big-rice-live.md
gretanausedaite: Changset files are being added to a list of changes: https://github.com/iTwin/iTwinUI/pull/1562
It's nicer to have separate items than list within a list.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+   * Tab panel subcomponent which contains the tab's content.
+   * @example
+   * <Tabs.Panel value='tab1'>
+   *   Sample Panel
+   * </Tabs.Panel>
+   */
+  Panel: TabsPanel,
+});
+
+export const TabsContext = React.createContext<
```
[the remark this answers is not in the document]
gretanausedaite: No

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: Is there a reason to not [`Omit<>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) it instead of deprecating it but allowing to pass it?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
-  contentClassName?: string;
+  value?: string;
   /**
-   * Custom CSS class name for the tabs wrapper.
+   * Function that gets called when active tab is changed.
+   *
+   * Should be used alongside `activeValue` prop.
    */
-  wrapperClassName?: string;
+  onValueChange?: (value: string) => void;
```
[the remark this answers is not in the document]
gretanausedaite: Users set controlled active value by passing "value" prop to the tabs wrapper. activeValue is internal variable that is used to render correct subcomponents.

examples/Tabs.overflow.tsx
gretanausedaite: Yes, thanks! Changed to cherry.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
rohan-kadkol: I mean to highlight that the public-facing JSDocs for `onValueChange` uses the term `activeValue`, which is not something that is exposed. So I guess you meant `value`.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+     * i.e. when there is not enough space to fit all the tabs.
+     *
+     * Not applicable to types `pill` and `borderless`.
+     */
+    useOverflow?: boolean;
+  };
+
+  // To remove `defaultValue` and `defaultChecked` from `<div>` props.
+  defaultValue?: never;
+  defaultChecked?: never;
```
rohan-kadkol, line 517: Same comment as https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1350373178

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: We did not deprecate other legacy APIs. But maybe we should, to tell users please use composition.

examples/Tabs.main.tsx
gretanausedaite: We do strongly recommend the new API.
We are considering marking old API deprecated: https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1336398920 Feel free to add you opinion.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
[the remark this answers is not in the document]
gretanausedaite: Updated jsdoc comment.

packages/itwinui-css/src/tabs/base.scss
gretanausedaite: Decided to remove this nowrap from tabs.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: We need the first tab. This way is more stable to keep track of it.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: I don't remember. @mayank99 any insights?

mayank99: you can try omitting it, i think it will be added back by `<div>` props (via `PolymorphicForwardRefComponent`

packages/itwinui-css/src/tabs/base.scss
mayank99: i think we can use `flex-shrink: 0` on individual tabs to prevent them from wrapping when overflowing.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
rohan-kadkol: Both ways sound good. Additionally, if we are not deprecating the old API, would be good to have an example of it on the docs site. https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1350187225

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: maybe we can explicitly spell it out that this prop does nothing. and remove old jsdocs.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+
+  const { activeValue, idPrefix } = useSafeContext(TabsContext);
+
+  return (
+    <Box
+      className={cx('iui-tabs-content', className)}
+      aria-labelledby={`${idPrefix}-tab-${value}`}
+      role='tabpanel'
+      hidden={activeValue !== value ? true : undefined}
+      id={`${idPrefix}-panel-${value}`}
```
mayank99, line 438: lets move this `id` after `rest` in case the user accidentally tries to override it.

same comment in Tab too

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+  const tabRef = React.useRef<HTMLButtonElement>();
+
+  const isActive = activeValue === value;
+
+  useIsomorphicLayoutEffect(() => {
+    if (isActive) {
+      tabRef.current?.scrollIntoView({
+        behavior: 'smooth',
+        block: 'center',
+        inline: 'center',
```
mayank99, line 224: i think we'll need to separate this into two different statements for vertical and horizontal, because we don't to scroll vertically in the case of horizontal tabs, and vice versa.

also should probably check for reduced-motion before using `'smooth'`

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -72,31 +75,35 @@ $borderless-horizontal-tab-min-height: calc(
   .iui-tab-icon {
     &,
     svg {
-      @include mixins.iui-icon-style('m');
+      grid-area: icon;
     }
 
     svg {
       fill: var(--iui-color-icon);
```
mayank99, line 82: this can be removed too?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +43,759 @@ type TabsOrientationProps =
...
+  );
+}) as PolymorphicForwardRefComponent<'button', TabLegacyProps>;
+
+// ----------------------------------------------------------------------------
+// exports
+
+export { LegacyTab as Tab };
+
+/**
+ * Tabs organize and allow navigation between groups of content that are related and at the same level of hierarchy.
```
mayank99, line 644: should we explain upfront that `value` is used for connecting tabs with panels?

.changeset/brave-snakes-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+Deprecated `overflowOptions` and `useOveflow` props in Tabs.
```
mayank99, line 5:
```suggestion
---
'@itwin/itwinui-react': major
---

Tabs are now always scrollable. Deprecated `overflowOptions` prop.
```

.changeset/chatty-bags-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-css': major
+'@itwin/itwinui-react': major
```
mayank99, line 3:
```suggestion
```

.changeset/early-weeks-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': major
+---
+
+Removed `data-iui-scroll-placement` and `data-iui-overflow` attributes.
```
mayank99, line 5: react user won't care, so this changeset can be removed

packages/itwinui-css/src/tabs/base.scss
gretanausedaite: Works perfectly for overflow, long labels does get big. Are we ok with this? 
image

.changeset/early-weeks-watch.md
gretanausedaite: Removed

packages/itwinui-css/src/tabs/base.scss
gretanausedaite: Removed fill

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated code to check motion and use different statements

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Moved ids after rest

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Added a short comment.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Not needed. This was changed.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: This was removed

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: Suggest rewording for clarity:
> `Tabs.Tab` and `Tabs.Panel` can be associated with each other by passing them the same `value`.

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: Suggest restructuring for clarity and to avoid bugs with useEffect dependencies:

```tsx
  useIsomorphicLayoutEffect(() => {
    if (isActive) {
      const behavior = isMotionOk() ? 'smooth' : 'auto';

      if (orientation === 'horizontal') {
        tabRef.current?.scrollIntoView({ behavior, inline: 'center' });
      } else {
        tabRef.current?.scrollIntoView({ behavior, block: 'center' });
      }
    }
  }, [isActive]);
```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: the comments could be reworded as "Removing…" instead of "To remove…"

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -336,4 +335,8 @@ $borderless-horizontal-tab-min-height: calc(
 
 @mixin iui-tabs-content {
   grid-area: tabs-content;
+
+  &[hidden='true'] {
```
mayank99, line 339: there is no `='true'` in the markup
```suggestion
  &[hidden] {
```

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated comments

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated using your second code example

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Added this rewording

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -64,547 +48,767 @@ type TabsOrientationProps =
...
-      scrollLeft += 25;
-    } else if (e.deltaY < 0 || e.deltaX < 0) {
-      scrollLeft -= 25;
-    }
-    ownerDoc.scrollLeft = scrollLeft;
-  }, []);
+  return (
+    <Box
+      className={cx('iui-tabs-wrapper', `iui-${orientation}`, className)}
+      style={stripeProperties}
```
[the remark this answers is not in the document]
gretanausedaite: Merged

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Created #1622 with docs changes in website

packages/itwinui-react/src/core/Tabs/Tabs.tsx
gretanausedaite: Updated comments for both

packages/itwinui-css/src/tabs/base.scss
```diff
@@ -186,37 +184,38 @@ $borderless-horizontal-tab-min-height: calc(
...
+    animation-name: scroll-shadow-inset-vertical;
+    animation-timing-function: linear;
+    /* stylelint-disable-next-line property-no-unknown */
+    animation-timeline: scroll(self inline);
 
-      &[data-iui-scroll-placement='start'] {
-        /* stylelint-disable-next-line property-no-vendor-prefix */
-        -webkit-mask-image: linear-gradient(0deg, transparent 0%, rgb(0, 0, 0) 5%, rgb(0, 0, 0) 100%, transparent 100%);
-        mask-image: linear-gradient(0deg, transparent 0%, rgb(0, 0, 0) 5%, rgb(0, 0, 0) 100%, transparent 100%);
+    @keyframes scroll-shadow-inset-vertical {
```
rohan-kadkol, line 204: non-blocking: I am slightly confused about how the two animations work. Could you elaborate?

apps/storybook/src/Tabs.stories.tsx
```diff
@@ -206,68 +259,56 @@ export const HorizontalOverflow = () => {
...
+          <Tabs.Panel key={item.name} value={item.name}>
+            {item.content}
+          </Tabs.Panel>
+        ))}
+      </Tabs.Wrapper>
     </div>
   );
 };
 
 export const VerticalOverflow = () => {
```
rohan-kadkol, line 290: May not be related to this PR, but just noticed some shift in the width of the tabs panel in the "Vertical Overflow" story

[14.webm](https://github.com/iTwin/iTwinUI/assets/45748283/2bb1bd2a-facc-408b-bb07-a4b5e5ffbda0)

apps/storybook/src/Tabs.stories.tsx
```diff
@@ -206,68 +259,56 @@ export const HorizontalOverflow = () => {
...
+    { name: 'Item 1', content: 'Tab Content One', disabled: false },
+    { name: 'Item 2', content: 'Tab Content Two', disabled: false },
+    { name: 'Item 3', content: 'Tab Content Three', disabled: false },
+    { name: 'Item 4', content: 'Tab Content Four', disabled: false },
+    { name: 'Item 5', content: 'Tab Content Five', disabled: false },
+    { name: 'Item 6', content: 'Tab Content Six', disabled: true },
+    { name: 'Item 7', content: 'Tab Content Seven', disabled: false },
+    { name: 'Item 8', content: 'Tab Content Eight', disabled: false },
+    { name: 'Item 9', content: 'Tab Content Nine', disabled: true },
+    { name: 'Item 10', content: 'Tab Content ten', disabled: false },
```
rohan-kadkol, line 303: nit

```suggestion
    { name: 'Item 10', content: 'Tab Content Ten', disabled: false },
```

rohan-kadkol (approved):

apps/storybook/src/Tabs.stories.tsx
mayank99: i cannot reproduce?

rohan-kadkol: I am on a Windows 11 + Edge 117. What I did was:
* Press tab to focus the selected "Item 11`"
* Press up or down

gretanausedaite: Could not find a reason why this is happening.

rohan-kadkol: Maybe we can make another issue for this since it is rare. If it happens again, we can update that issue.

packages/itwinui-css/src/tabs/base.scss
gretanausedaite: Updated with https://github.com/iTwin/iTwinUI/pull/1548/commits/6dbb70fed6756dc73569f7629dd949246974c9cc

PR #1552, opened 2023-09-08
updated File Upload documentation.

siddhantrawal:
## Changes

Added File Upload subcomponents documentation.

## Testing

"N/A"

## Docs

 "N/A"

### DIFF
--- a/apps/website/src/components/PropsTable.astro
+++ b/apps/website/src/components/PropsTable.astro
@@ -82,7 +82,7 @@ const componentDoc = [...docgenResults].find((docs) => docs['displayName'] === c
       width: 100%;
 
       &,
-      * {
+      :global(*) {
         font-size: var(--type--1);
       }
     }
@@ -99,6 +99,7 @@ const componentDoc = [...docgenResults].find((docs) => docs['displayName'] === c
     td {
       padding: var(--space-3) var(--space-2);
       max-width: 35ch;
+      overflow-wrap: break-word;
 
       &:first-child {
         padding-left: 0;
--- a/apps/website/src/pages/docs/buttongroup.mdx
+++ b/apps/website/src/pages/docs/buttongroup.mdx
@@ -61,7 +61,7 @@ You can combine [inputs](input) with buttons in a button group.
 
 ## Props
 
-To be defined
+<PropsTable path={frontmatter.propsPath} />
 
 ## Related components
 
--- a/apps/website/src/pages/docs/slider.mdx
+++ b/apps/website/src/pages/docs/slider.mdx
@@ -93,4 +93,4 @@ Tooltips can also be customized such as changing the placement or the informatio
 
 ## Props
 
-To be defined.
+<PropsTable path={frontmatter.propsPath} />
--- a/examples/Avatar.groupoverflowtooltip.tsx
+++ b/examples/Avatar.groupoverflowtooltip.tsx
@@ -27,7 +27,7 @@ export default () => {
   /**
    * Ref is set on the last avatar for tooltip positioning.
    */
-  const avatarRef = React.useRef<HTMLDivElement>(null);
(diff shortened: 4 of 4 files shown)


apps/website/src/pages/docs/fileupload.mdx
```diff
@@ -14,14 +14,22 @@ import * as AllExamples from 'examples';
...
   <AllExamples.FileUploadMainExample client:load />
 </LiveExample>
 
 File uploading is a common interaction most users do multiple times per day. File uploading may happen in various different ways depending on the context. While the overall design may differ from UI to UI, the basics always stay the same: let the user know they can upload content, display upload status progress, and confirm that the upload’s status upon completion.
 
+## Customizations
+
+### Custom File Upload Card
+
+Change standard file upload card by adding custom label and text.
```
gretanausedaite, line 27: I think it would be useful to list all subcomponents we are exposing and adding an example with default subcomponents.

gretanausedaite (commented):
Chat and Table sections are a bit out of context for me. They are talking about usage/workflows and not the component itself. 
Maybe move those sections under `Usage`?

examples/FileUpload.custom.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <FileUploadCard.InputLabel>Custom Label Text</FileUploadCard.InputLabel>
+        <div>Custom Description Text</div>
+      </FileEmptyCard.Text>
+    </FileEmptyCard>
+  );
+  return (
+    <FileUpload
+      onFileDropped={(files) => {
+        setFiles(files);
+        action(`${files.length} files uploaded`)();
```
rohan-kadkol, line 32: `action`s can be changed to `console.log()` or removed. This is because actions are storybook specific

examples/FileUpload.custom.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <FileEmptyCard.Text>
+        <FileUploadCard.InputLabel>Custom Label Text</FileUploadCard.InputLabel>
+        <div>Custom Description Text</div>
+      </FileEmptyCard.Text>
+    </FileEmptyCard>
+  );
+  return (
+    <FileUpload
+      onFileDropped={(files) => {
+        setFiles(files);
```
rohan-kadkol, line 31: There is some type error here.

image

apps/website/src/pages/docs/fileupload.mdx
gretanausedaite: I have in mind documentation, not code example: 
image

apps/website/src/pages/docs/fileupload.mdx
```diff
@@ -14,23 +14,57 @@ import * as AllExamples from 'examples';
...
+- `FileUploadCard.Input`
+- `FileUploadCard.InputLabel`
+- `FileEmptyCard.Icon`
+- `FileEmptyCard.Text`
+
+## Customizations
+
+### Wrapping Input
+
+Drag and Drop Files with custom message.
```
rohan-kadkol, line 43:
```suggestion
Drag and Drop Files along with an input.
```

rohan-kadkol (approved):

apps/website/src/pages/docs/fileupload.mdx
gretanausedaite: The list of components can be found in react code.
You should move that list here or add another example in the section above.

gretanausedaite: > You should move that list here

Remove Composition API section and just add the list in the customisations section

**OR**

> add another example in the section above

Create an example for composition API with standard subcomponents without customising them.

Section called "Composition API" with no example actually does not explain how to use subcomponents. Simply listing subcomponents does not make user understand it better. Those two sections are not linked together or anything. How would users know that these subcomponents are used in customisation example???

gretanausedaite (approved):

PR #1555, opened 2023-09-08
Button group refactor

FlyersPh9:
## Changes

- Button group now uses `aria-orientation` instead of class modifier.
- Button group is using `<ul>` and `<li>`, but doesn't always have to.

## Testing

- Visual tests change only because the diving line between disabled buttons in a button group has changed.

## Docs

N/A


mayank99 (commented):
i was actually thinking of simplifying the DOM so we wouldn't need any wrapping elements around the buttons. #1556

also `aria-orientation` is invalid on a regular `<ul>`, plus we are trying to move away from lists in general.

PR #1556, opened 2023-09-11
remove extra divs from button-group

mayank99:
## Changes

remvoed the extra divs which were preivously added to work around issues with popovers being injected into the button-group and breaking border-radius.

## Testing

no visual changes.

## Docs

added changeset.


packages/itwinui-css/backstop/tests/button.html
```diff
@@ -1072,338 +1072,278 @@ <h2>Grouped</h2>
...
+          <svg-thumbnails
+            class="iui-button-icon"
+            aria-hidden="true"
+          ></svg-thumbnails>
+        </button>
       </span>
 
       <br />
 
       <span class="iui-button-group-vertical">
```
FlyersPh9, line 1290: Can this be changed to `class="iui-button-group" data-iui-orientation="vertical"`?

mayank99: yeah, totally

mayank99: i was working on this but it feels a bit repetitive and verbose because:
- the attribute is applied on the button-group but most of the styles are applied to nested children
- the border-radius properties are different for vertical and horizontal

gretanausedaite (approved):

PR #1557, opened 2023-09-11
restore old Breadcrumbs api

mayank99:
## Changes

undoes the breaking change from #1369 using `children.type` and `children.props`.

both of these apis are brittle and not really recommended, so i'm logging a warning to encourage using the new api. we can also auto-convert to the new api in our codemod.

## Testing

added unit test for it.

## Docs

since there is no real benefit to using the old api (it's not any less verbose), i'm not documenting it anywhere.


packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -0,0 +1,33 @@
...
+} catch {}
+
+/**
+ * Logs message one time only in dev environments.
+ *
+ * @example
+ * const logWarningInDev = createWarningLogger();
+ * logWarningInDev("please don't use this")
+ */
+const createWarningLogger = !isDev
```
gretanausedaite, line 21: Love this!

gretanausedaite (approved):

.changeset/soft-crabs-flow.md
```diff
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': major
+'@itwin/itwinui-react': minor
 ---
 
-Added new `Breadcrumbs.Item` subcomponent to use for custom `a` and `button` elements within the `Breadcrumbs`
+Added new `Breadcrumbs.Item` subcomponent to use within `Breadcrumbs`. Directly passing `<a>` and `<Button>` as children is still supported but is deprecated.
```
rohan-kadkol, line 5: nit: Should we also mention about using `<span>`?

rohan-kadkol (approved):

.changeset/soft-crabs-flow.md
mayank99: sure.
```suggestion
Added new `Breadcrumbs.Item` subcomponent to use within `Breadcrumbs`. Directly passing `<a>`/`<Button>`/`<span>` as children is still supported but is deprecated.
```

Issue #925, opened 2021-10-01
ComboBox: loading state


mayank99:
![image](https://user-images.githubusercontent.com/9084735/135276322-4dd1e6ba-ed7a-43b6-8c9d-7dddfa0a267e.png)
https://react-select.com/home#async

_Originally posted by @mayank99 in https://github.com/iTwin/iTwinUI-react/pull/338#discussion_r718504790_

SpencerWBarnes:
I see this enhancement is still open. Is the intent to offer both the MenuItemSkeleon and the meatball icon to represent loading? Should there only be one approach for consistency? If not, which method would be recommended?

mayank99:
@SpencerWBarnes It would depend on your use case. The skeleton ui would be more applicable when lazy loading the menu items.

The loading indicator on the right side (once implemented) would probably be more useful because you get to keep showing the currently loaded menu items, and then replace it with the new list when done loading.

https://user-images.githubusercontent.com/9084735/217003006-0d2b2857-4ab4-4d7b-b33c-0888f37724fe.mov

SpencerWBarnes:
@mayank99 Gotcha so the skeleton ui best represents a total change to the list whereas the loading indicator best represents a progressive change. Thanks!

FlyersPh9:
iTwinUI is now in maintenance mode per [the support policy](https://github.com/iTwin/iTwinUI/wiki/Support-policy), with [StrataKit](https://github.com/iTwin/stratakit) replacing it as the primary design system.

PR #1567, opened 2023-09-13
ComboBox: Add customisation props

gretanausedaite:
## Changes

Added endIconProps and statusMessageProps;

Input wrapper can be customised with props;
Input can be customised with inputProps;
Input endIcon > endIconProps;
Dropdown menu > dropdownMenuProps;
Status message > statusMessageProps;

## Testing

Added unit tests;

## Docs

Changeset


packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -560,7 +578,11 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
                 }
               />
             </>
-            <ComboBoxEndIcon disabled={inputProps?.disabled} isOpen={isOpen} />
+            <ComboBoxEndIcon
+              disabled={inputProps?.disabled}
+              isOpen={isOpen}
+              {...endIconProps}
```
rohan-kadkol, line 584: Should `endIconProps` be before `isOpen`?

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -1023,3 +1023,29 @@ it('should update live region when selection changes', async () => {
...
+  const endIcon = container.querySelector(
+    '.iui-svg-icon.iui-end-icon.my-end-icon',
+  );
+  expect(endIcon).toBeTruthy();
+  const messageIcon = container.querySelector(
+    '.iui-status-message > .iui-svg-icon.my-message-icon',
+  );
+  expect(messageIcon).toBeTruthy();
+  const statusContent = getByText('test message');
+  expect(statusContent).toHaveClass('my-message-content');
```
rohan-kadkol, line 1050: nit: For symmetry, you can use the same method for checking the class of `messageIcon` and `statusContent`. Either all using selectors with the class name, or all using `.toHaveClass()`

.changeset/weak-adults-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added `endIconProps` and `statusMessageProps` to ComboBox. This allows users to customise end icon and status message.
```
rohan-kadkol, line 5: Typo

```suggestion
Added `endIconProps` and `statusMessageProps` to ComboBox. This allows users to customize end icon and status message.
```

gretanausedaite: "Customize and customise are both English terms."

I guess it's a preference issue.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
gretanausedaite: I can get text with getByText function. Not sure how to get icons without class selectors.
Keeping as is, because it gets the job done.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -1023,3 +1023,29 @@ it('should update live region when selection changes', async () => {
...
+  const { container, getByText } = render(
+    <ComboBox
+      options={[0, 1, 2].map((value) => ({ value, label: `Item ${value}` }))}
+      message='test message'
+      status='negative'
+      statusMessageProps={{
+        contentProps: { className: 'my-message-content' },
+        iconProps: { className: 'my-message-icon' },
+      }}
+      endIconProps={{ className: 'my-end-icon' }}
```
mayank99, line 1037: maybe also test for other parts? main className, dropdownMenuProps, and anything else

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -102,6 +103,17 @@ export type ComboBoxProps<T> = {
...
    */
   dropdownMenuProps?: React.ComponentProps<'div'>;
+  /**
+   * End icon props.
+   */
+  endIconProps?: React.ComponentProps<typeof ComboBoxEndIcon>;
+  /**
+   * status message props.
+   */
+  statusMessageProps?: Pick<
```
mayank99, line 113: i've just realized that this may not actually be needed, since user can pass the actual `<StatusMessage>` component with all customizations

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -102,6 +103,17 @@ export type ComboBoxProps<T> = {
    * Props to customize dropdown menu behavior.
    */
   dropdownMenuProps?: React.ComponentProps<'div'>;
+  /**
+   * End icon props.
+   */
+  endIconProps?: React.ComponentProps<typeof ComboBoxEndIcon>;
```
mayank99, line 109: should we also allow replacing the icon completely?

gretanausedaite: I would rather not because it's combobox.
Users can do InputWithDeco and new Popover if they want something else there.

mayank99: what if i want to show a loading spinner in the combobox (#925)? Popover is not a suitable replacement for all ComboBox functionality.

doesn't necessarily need to be handled in this PR if it's too complicated, but it's good to think about it now, because it affects the API.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Updated!

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
gretanausedaite: Updated tests

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: wont this rotate the custom icon by 180deg when menu opens?

gretanausedaite: Nope, rotation is done with open prop that we are not passing in this case.
Will add story and tests for this, if the approach is ok.

mayank99: hmm yeah this could work. can you add example using ProgressRadial instead of camera?

in the future we can think about how to support multiple icons too; maybe the user wants both arrow and spinner

gretanausedaite: Removed endIcon from this PR.
It requires some css and/or combobox layout changes for it to work properly. Will handle in different PR or after major release.

.changeset/weak-adults-scream.md
mayank99, line 5:
```suggestion
Added `endIconProps` prop to ComboBox.
```

mayank99 (approved):

PR #1570, opened 2023-09-13
Fix Table types from being hidden

rohan-kadkol:
## Changes

Stemmed from https://github.com/iTwin/iTwinUI/pull/1514#discussion_r1323515670.

* Prevent the table types from bring hidden
  * Removed `paths` from `itwinui-react/tsconfig` since that was causing unnecessary `any` when the type could not be found on the user's side.
  * Updated all imports to import `react-table.ts` by file path
  * Before:
    image
  * After:
    image
* Fixed type errors in `Table.stories.tsx` that were visible after the above fix.
  * One of these errors was that I noticed that when passing no data (`[]`), the type of row data could not be inferred. So the row data type was becoming never. Thus, it was giving errors like:
    * image
    * So I added a generic type to `<Table>`
      * image

## Testing

Ran `yarn test:types` and `yarn build` in `@itwin/itwinui-react`.

## Docs

Added info to the [v3 migration guide](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-react-v3-migration-guide):

image


mayank99 (commented):
i think the non-type imports still need to be done from the `react-table` package; that's where all the actual code comes from.

you'll likely get an error about undefined types, but it's safe to ignore or work around using a [dummy `.d.ts` file](https://stackoverflow.com/a/54174427).

rohan-kadkol:
> i think the non-type imports still need to be done from the `react-table` package; that's where all the actual code comes from.

Ohh, right, changed all non-types imports to import from `react-table` itself.

> you'll likely get an error about undefined types, but it's safe to ignore or work around using a [dummy `.d.ts` file](https://stackoverflow.com/a/54174427).

To solve this error, the suggestion from that StackOverflow answer to use `/// <reference path='../../../react-table.d.ts' />` did work unfortunately.

image

The second idea related to using `files` in tsconfig stopped showing the red underline. But there was no intellisense. And if we hover over the import, it gives the same error.

image

So I decided to revert back to using `paths` in `tsconfig`. This means that there will be no error if we import types from `react-table` in our internal Table files. Thus the unexpected `any` on the users' side that this PR is trying to solve may come back. So we just have to be careful to always import types using the file path.

Since we are moving to the Tanstack v8 table, this might be okay to do, since we may not make many changes to the table files in v3 when working on v4.

But if there is a better way than the above approach, then I can try implementing that instead.

gretanausedaite (approved):
No comments from my side.

apps/storybook/src/Table.stories.tsx
```diff
@@ -92,7 +92,7 @@ export const Basic = () => {
       },
     ],
     [],
-  );
+  ) satisfies Column<{ name: string; description: string }>[];
```
mayank99, line 95: this i dont understand. so everyone will get an error if their columns don't have explicit types set?

rohan-kadkol: Not always, but yes, in most cases users might need to use `satisfies` or `as` to specify a type for props that they pass to `<Table>`. Sometimes, the generic type can be omitted.

```diff
- const columns = [...] satisfies Column<{ name: string, description: string }>[]
+ const columns = [...] satisfies Column[]
```

mayank99: lets show the less verbose version then

mayank99:
```suggestion
  ) satisfies Column[];
```

rohan-kadkol: https://github.com/iTwin/iTwinUI/pull/1570/commits/31623868a5a2b50d0fe40f0545cbc68b448e8188: I implemented the above change and also removed any unnecessary type assertions. I tried to show examples that are as type lenient as possible. I also replaced a few type assertions to use `satisfies` instead.

mayank99 (approved):
LGTM

PR #1572, opened 2023-09-13
adjust border and shadows for dark theme

mayank99:
## Changes

changes to variables in dark theme:
1. increased color contrast for borders (`iui-color-border`, `iui-color-border-subtle`, `iui-color-border-disabled`)
2. increased shadow intensity by reducing opacity in `iui-shadow-color`

*pulled out of #1565 for visual consistency between v2 and v3.*

## Testing

updated visual tests. see image diffs for visual changes.

deploy preview: https://itwin.github.io/iTwinUI/1572

## Docs

n/a.

there will be some docs updates in v3 when more changes are made in components.

Issue #1222, opened 2023-04-20
Menu


gretanausedaite:
Menu:

- [ ] Use ListItem
- [ ] Deprecate MenuItem
- [ ] Use floating-ui (TBD)

MenuItem:

- [ ] Change props: badge > endIcon, icon > startIcon

gretanausedaite:
- TimePicker, ColorPicker, DatePicker embed in page option mockup:
![image](https://user-images.githubusercontent.com/83585998/234831507-02d141b6-7e33-4ba0-b81f-51452eef4634.png)

PR #1573, opened 2023-09-13
allow toggling background in color-picker and date-picker

mayank99:
## Changes

Pulled out of #1572 as independent change. Addresses multiple points in #1222 (see "Embed in page").

Kept default value of `applyBackground` to `true` to avoid breaking change. This part is debatable.

## Testing

added unit test for new prop.

no visual changes (those will happen in #1572).

## Docs

n/a


gretanausedaite (approved):
Any idea why TimePicker images changed?

Also it would be nice to add storybook and website examples with embedded in page components.

mayank99:
> Any idea why TimePicker images changed?

it's because we hide the trigger, which makes floating-ui reposition the popover.

> Also it would be nice to add storybook and website examples with embedded in page components.

i'm not sure about this. i do not fully understand the use case for "embed in page" and it does not make sense in isolation. the background would bleed into the foreground and cause contrast issues. but it can be done in a future PR if it's actually desirable

PR #1576, opened 2023-09-14
make dropdown menus scrollable by default 

mayank99:
## Changes

this PR fixes #1301 in two steps:
- added `iui-scroll` by default to regular `DropdownMenu`s and submenus.
- added 100ms delay to submenu.

in #1577: `iui-scroll` will be removed and the overflow css will be moved into base styling in v3.

## Testing

| Before | After |
| --- | --- |
|  |  |

## Docs

n/a

PR #1577, opened 2023-09-14
make menus scrollable by default

### LINKED ISSUE #1301: [DropdownMenu] Difficult to select submenu items

mayank99:
## Changes

same as #1576 except all menus are scrollable right in the css. fixes #1301.

it's ok to make this change this now mainly because in #1506 the outermost element of virtualized combobox was made to be `<Menu>` again, see https://github.com/iTwin/iTwinUI/pull/1506#discussion_r1321528058 (this was possible as a result of #1530 which gave us more freedom over dom structure)

## Docs

n/a


.changeset/healthy-pigs-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
```
rohan-kadkol, line 2: Since it is a major change, do we have to include it in the migration guide? (I noticed you wrote "n/a" under "Docs")

rohan-kadkol (approved):

.changeset/healthy-pigs-fry.md
mayank99: i think it's not that important because the class will do nothing if it's still present in the markup. but i'll make a brief note in the migration guide.

gretanausedaite (approved):

PR #1581, opened 2023-09-18
Ignoring a11y false positives violations related to `li` elements inside DatePicker with Time Component.

siddhantrawal:
Two of the a11y violations related to DatePicker Component were failing locally but passing correctly on Cypress GUI. The issue was caused due to all the `li` elements inside the DatePickerWithTimeExample and DatePickerWithCombinedTimeExample Component not meeting `color-contrast` criteria locally.

Screenshot 2023-09-18 111002
Screenshot 2023-09-20 104806

Electron Browser ( DatePickerWithCombinedTimeExample): 

FailureSummary: Element has insufficient color contrast of 1.38 (foreground color: #dbdbdb, background color: #ffffff, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1"

image

Electron Browser ( DatePickerWithTimeExample): 

FailureSummary: Element has insufficient color contrast of 1.38 (foreground color: #dbdbdb, background color: #ffffff, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1"
image

Locally (Terminal) : 
Screenshot 2023-09-05 114508

Cypress GUI:
Screenshot 2023-09-18 105202

As discussed during the call, they were false positives. 

## Change

Thanks to Rohan for this. Workaround for this was to ignore these violations as shown in this [config](https://github.com/iTwin/iTwinUI/blob/631f884149881b0136b7c67f4c4a621ab665baf1/testing/a11y/src/Component.cy.tsx).

## Testing

Ran it locally and on Cypress GUI and no longer got the failure result.

Locally (Terminal) : 
Screenshot 2023-09-18 111559

Cypress GUI:
Screenshot 2023-09-18 130712

## Docs

 "N/A".


mayank99 (commented):
suggestions to improve maintainability:
(1) in the PR description, add an explanation with screenshot of exactly what part is triggering the false positive.
(2) in the code, include a comment linking to this PR.

testing/a11y/src/Component.cy.tsx
```diff
@@ -6,6 +6,24 @@ import * as React from 'react';
 import * as allExamples from 'examples';
 import { ThemeProvider } from '@itwin/itwinui-react';
 
+type commonRulesKeys = 'ignore_color_contrast_violation';
+const commonRules: Record<commonRulesKeys, Record<string, any>> = {
+  ignore_color_contrast_violation: {
+    id: 'color-contrast',
+    enabled: true,
+    selector: ':not(.iui-root, [data-iui-theme]) .iui-time > ol > li)',
```
rohan-kadkol, line 14: Should the closing parenthesis of the `:not` be after the `li`?

Because I believe the logic of this L13-14 is to enable the `color-contrast` rule for all tags that match the selector. So because of the `:not( ... li)`, it means the rule is enabled for all tags that are *not* those `li`s.

Or maybe, can you explain your reasoning for that selector?

rohan-kadkol: > I was unsure about the right path to get to the `li` element. So I kind of followed my way as shown in the screenshot.

I see. That CSS selector might also work. But still the `:not()` would likely need to encapsulate the entire selector:

```diff
- selector: ':not(.iui-root, [data-iui-theme]) .iui-time > ol > li)',
+ selector: ':not(:where(.iui-root, [data-iui-theme]) .iui-time > ol > li)', 
```

> Also, I did this because I thought we only wanted to select those `li` elements that are descendants of `ol` and are children of class `iui-time`.

Yes, the goal is to select *only* the `li` in `ol` in `iui-time`. So a more specific selector is ideal here. 

> My logic will exclude the selecting of `li` elements which are descendants of class `iui-root` with data attribute `data-iui-theme`.
> 
> Not sure if this correct.

I think just enabling the rule for `.iui-time > ol > li` would be enough. The `.iui-root, [data-iui-theme]` may not be needed. But I haven't tested this. So you can confirm it.

gretanausedaite (approved):
No comments from my side.
There are couple more contrast failures in a11y test results. They are not false positive?

rohan-kadkol (approved):

mayank99 (approved):
LGTM for now.

Since we discovered this issue was specific to Electron browser (which is what gets used in CI), we might want to consider switching to Chromium in a future PR.

PR #1589, opened 2023-09-21
reorganize file structure

mayank99:
## Changes

two main changes to simplify file structure:
- removed subfolders inside `Buttons` and `Typography`
- moved public utility components out of `utils` because this folder is mainly for internal utilities
   - this change will likely help with bundle size since we are importing utils in most components, and now this folder is smaller.

no visible change from user perspective.

## Testing

n/a

## Docs

n/a


mayank99 (commented):
thoughts on removing most of the `index` files? it provides absolutely no benefit, creates confusion about import paths, and can sometimes cause circular dependency issues.

gretanausedaite (commented):
> thoughts on removing most of the index files (except the top-level src/index.js and the ones in utils)?

I like having it because it's easier to see what component is exported. We have a lot of subcomponents that are not exported and it's rather easy to miss what we are delivering just by looking into large tsx files.

gretanausedaite (approved):

mayank99:
> I like having it because it's easier to see what component is exported. We have a lot of subcomponents that are not exported and it's rather easy to miss what we are delivering just by looking into large tsx files.

The main barrel file (`src/index.ts`) is the correct place to look for what is part of the public API. Something could be incorrectly exported in `src/core/Alert/index.ts` and not exported in the barrel (creating confusion, rather than saving time).

gretanausedaite:
> The main barrel file (`src/index.ts`) is the correct place to look for what is part of the public API. Something could be incorrectly exported in `src/core/Alert/index.ts` and not exported in the barrel (creating confusion, rather than saving time).

Then ok, remove the index files.

PR #1590, opened 2023-09-22
Fix `useScrollToRow` crash on property access of undefined `pageRef.current[scrollToIndex]` object

DaumantasJankauskas:
## Changes
Fixes `useScrollToRow` crash on property access of undefined `pageRef.current[scrollToIndex]` object
## Testing
N/A

## Docs

N/A

### DIFF
--- a/.changeset/bright-onions-draw.md
+++ b/.changeset/bright-onions-draw.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed undefined property access in `useScrollToRow`.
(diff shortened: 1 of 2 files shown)


rohan-kadkol (approved):
Great catch; thanks for this PR!

I think this should go to `main` instead of the `dev` branch since this is not specific to our upcoming v3 release. But we can handle that on our end.

DaumantasJankauskas:
Thanks, that would be nice

mayank99 (approved):

mayank99:
@DaumantasJankauskas this should be released in 2.12.10. thanks for your PR

PR #1597, opened 2023-09-26
Adding animation for Table Expandable

siddhantrawal:
## Changes
Passing an object to classNames instead of a `iui` string prefix.

## Testing

Manual testing on storybook. 

## Docs

Added changeset


mayank99 (changes requested):
what is this?

mayank99 (commented):
i think we need to support padding in expanded content, because the user should be free to control the content however they want.

the task here is only to _make the transition work_. the CSS grid hack was just a suggestion, so if this is too difficult, i think it's also totally ok to keep the transition in JS.

to fix the difference between CSS-modules'd classes and `iui` prefix passed to `CSSTransition`, you can use pass an object (instead of a string) to `classNames` ([docs](https://reactcommunity.org/react-transition-group/css-transition#CSSTransition-prop-classNames)).

and to get access to CSS-module classes, you can import styles from `styles.js` (similar to how we do in `polymorphic.tsx`).

```diff
 import { CSSTransition } from 'react-transition-group';
 import type { CSSTransitionProps } from 'react-transition-group/CSSTransition';
+import styles from '../../../styles.js';
 
 export const WithCSSTransition = (
   props: Partial<CSSTransitionProps> & {
@@ -41,7 +42,12 @@ export const WithCSSTransition = (
         node.style[`min${dimensionCamelCase}`] = 'initial';
         node.style[dimension] = '0px';
       }}
-      classNames='iui'
+      classNames={{
+        enter: styles['iui-enter'],
+        enterActive: styles['iui-enter-active'],
+        exit: styles['iui-exit'],
+        exitActive: styles['iui-exit-active'],
+      }}
       {...rest}
     >
       {React.isValidElement(children) ? (
```

mayank99 (approved):
LGTM now. No other changes needed.

https://github.com/iTwin/iTwinUI/assets/9084735/245a16bf-a482-4ae8-b653-8efcdc5aa5e9



By importing styles, I just meant this object (which you already are using):
```jsx
import styles from '../../../styles.js';
```

.changeset/little-cat-eating.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added table expandable animation.
```
mayank99, line 5: changeset not needed because the animation already existed in v2, so there is no "change" from a user perspective.

gretanausedaite (approved):

PR #1598, opened 2023-09-26
add Popover docs

mayank99:
## Changes

Changes to docs: added new `/popover` page with usage and accessibility details.

While working on docs, I found I had to make changes to Popover component (now moved into #1602).

## Testing

Tested all examples in multiple browsers and with keyboard.

## Docs

n/a. These *are* docs.


examples/Popover.focus.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <Popover
+      applyBackground
+      aria-labelledby={headingId}
+      aria-describedby={descriptionId}
+      visible={isOpen}
+      onVisibleChange={setIsOpen}
+      style={{ maxWidth: '45ch', border: 'none' }}
+      content={
+        <Surface elevation={0}>
```
mayank99, line 24: I had to add `border: none` override because Surface always insists on bringing its own border.

Also, for some reason `Surface.Body` only brings horizontal padding not vertical.


@FlyersPh9 Is this intentional?

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -101,6 +101,10 @@ type PopoverInternalProps = {
    */
   trigger?: Partial<Record<'hover' | 'click' | 'focus', boolean>>;
   role?: 'dialog' | 'menu' | 'listbox';
+  /**
+   * Whether the popover should match the width of the trigger.
+   */
+  matchWidth?: boolean;
```
gretanausedaite, line 107: 1. I liked that we are giving the users this feature. I think it would be useful for them.
2. If we want to make it private, prop imports in components that use Popover needs to be updated as it is using only public props.

mayank99: i can't think of any use case where the user might want `matchWidth`, it's mostly useful for menus (which we already provide multiple components for). do you have another example in mind? if a user asks for it, we can always add it back in the future.

i've fixed the props types in [`d146a85`](https://github.com/iTwin/iTwinUI/pull/1598/commits/d146a859002a0e53538c30395b813a2ef02eee71)

mayank99 (commented):
@rohan-kadkol Reading through these docs will help you make sense of the changes in #1602

gretanausedaite (approved):

examples/Popover.focus.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <Flex flexDirection='column' alignItems='start'>
+              <Text as='p' id={descriptionId}>
+                (TODO: improve text) Lorem ipsum dolor sit amet consectetur
+                adipisicing elit. Voluptatem, voluptates impedit rem incidunt
+                excepturi accusamus facilis doloribus saepe quibusdam
+                reiciendis.
+              </Text>
+
+              <Button
+                ref={(el) => el?.focus()}
```
rohan-kadkol, line 40: nit: It might be helpful if there were an explanation comment here. Something like, "To grab focus when the popover opens".

rohan-kadkol (approved):

examples/Popover.focus.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Surface elevation={0}>
+          <Surface.Header>
+            <Text as='h3' id={headingId} variant='leading'>
+              Terms and conditions
+            </Text>
+          </Surface.Header>
+          <Surface.Body isPadded>
+            <Flex flexDirection='column' alignItems='start'>
+              <Text as='p' id={descriptionId}>
+                (TODO: improve text) Lorem ipsum dolor sit amet consectetur
```
mayank99, line 33: @gretanausedaite @rohan-kadkol @siddhantrawal Can somebody suggest better text or a different example? I dunno how this has 3 approvals with TODO text 😅

rohan-kadkol: > Can somebody suggest better text or a different example?

How about a "see image preview" popover? If we click on the trigger button that says "show image preview", the popover shows the image preview along with an "OK" button and a "Download" button.

> I dunno how this has 3 approvals with TODO text 😅

I thought that TODO was for you so you will handle it yourself 😅 So I gave a preemptive review to not block the PR because everything else LGTM.

examples/Popover.focus.tsx
FlyersPh9: `data-iui-padded="true"` does have vertical padding, it's just very little.
Screenshot 2023-09-28 at 4 28 04 PM
The examples on [the surface demo page](https://itwin.github.io/iTwinUI/css/surface.html) look nice, but the agree button in your screenshot does not look so great.  Was the padding changed when we implemented `scrollbar-gutter`?

Btw, just now found out [`scrollbar-gutter` doesn't work in Safari](https://caniuse.com/mdn-css_properties_scrollbar-gutter). 😓

examples/Popover.focus.tsx
gretanausedaite: > Can somebody suggest better text or a different example?

Some sliders to "tune settings", eg. quality, grain, color, saturation, etc. (like in photoshop) with ok or save / cancel buttons at the bottom.

gretanausedaite: > I dunno how this has 3 approvals with TODO text

We have more than one todo in repo that is supposed to be solved later.

mayank99: updated the example to show a settings dialog

gretanausedaite: It's good. We have this in css fieldset tests if you want to switch inputs but you don't have to. This example is to show focus control, it does not have to be useful or realistic.
image

rohan-kadkol (approved):

PR #1600, opened 2023-09-27
switch from yarn to npm

mayank99:
> [!IMPORTANT]
> Abandoned effort. This is somehow even worse than Yarn.

---

## Changes

So I've just learned that npm supports using `yarn.lock`. It uses the information in `yarn.lock` to generate `package-lock.json` and keep them both in sync. https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson.html

Was able to get it working fine locally on Mac.

There are three main changes to terminal commands:
- **Required**: `yarn`/`yarn install` replaced with `npm install`
- Optional: `yarn <some-script> <args>` replaced with `npm run <some-script> -- <args>`
  - Example: `yarn build --filter=storybook` is now `npm run build -- --filter=storybook`
  - `yarn workspace <some-workspace> <some-script>` replaced with `npm run --workspace=<some-workspace> <some-script>`. Unfortunately this one doesn't support passing args (or at least I can't find a way to do it), so using yarn commands is ok here.
- Optional: `yarn <some-package>` replaced with `npx <some-package>`
  - Example: `yarn serve` is now `npx serve`

## Testing

Tested running dev and build commands on:
- [x] Mac.
- [ ] Linux. (**Github CI fails**)
- [ ] Windows.

## Docs

Updated contributing guide, PR template, etc


veekeys:
I would suggest considering using `pnpm`. I had issues with npm and especially lock file in monorepo setup (it was weirdly not updating as expected), whereas pnpm works really smoothly, always easy to update lock file with simple `pnpm i` command whereas you update version in package.json or just want to solve merge conflicts in lock file... As of now, it brings the best experience.
Another option, wait for `bun` to support windows (if you need it?)

mayank99:
yeah, the plan is to use `pnpm` eventually but currently it causes lots of issues (i've tried it two times in this repo in the last 6 months), so i was hoping `npm` with `yarn.lock` would offer a smoother migration path. `bun` will never take windows seriously, so that's out of the equation.

i think the safest solution would be to blow away lockfile and start fresh and fix all the issues that arise. we will revisit this once we get rid of storybook.

veekeys:
Yeah.
Quickly tried migrating to `pnpm`, but seems you are using lots of non likeable features :D
At the end, got stuck on this one..

image

PR #1602, opened 2023-09-27
fix Popover focus

mayank99:
## Changes

Changes pulled out of #1598 for ease of review.
- fixed focus behavior by changing props in `FloatingFocusManager` and adding `FloatingPortal`. also added fallback `id`/`aria-labelledby` to ensure the dialog is always labelled when it receives focus.
  - correct behavior described in Accessibility section of docs in #1598.
- removed `matchWidth` prop from public api as it's for internal use
- added missing default values and jsdocs to some props

## Testing

Tested focus behavior with and without a button that grabs focus.

<details>
<summary>Code</summary>

```tsx
import { Button, Flex, Popover, Surface, Text } from '@itwin/itwinui-react';

export default () => {
  const idPrefix = React.useId();
  const headingId = `${idPrefix}-label`;
  const descriptionId = `${idPrefix}-description`;

  const [isOpen, setIsOpen] = React.useState(false);

  return (
    <Popover
      applyBackground
      aria-labelledby={headingId}
      aria-describedby={descriptionId}
      visible={isOpen}
      onVisibleChange={setIsOpen}
      style={{ maxWidth: '45ch', border: 'none' }}
      content={
        <Surface elevation={0}>
          <Surface.Header>
            <Text as='h3' id={headingId} variant='leading'>
              Terms and conditions
            </Text>
          </Surface.Header>
          <Surface.Body isPadded>
            <Flex flexDirection='column' alignItems='start'>
              <Text as='p' id={descriptionId}>
                (TODO: improve text) Lorem ipsum dolor sit amet consectetur
                adipisicing elit. Voluptatem, voluptates impedit rem incidunt
                excepturi accusamus facilis doloribus saepe quibusdam
                reiciendis.
              </Text>

              <Button
                // ref={(el) => el?.focus()} // uncomment this to override focus
                onClick={() => setIsOpen(false)}
                styleType='high-visibility'
              >
                Agree
              </Button>
            </Flex>
          </Surface.Body>
        </Surface>
      }
    >
      <Button>Details…</Button>
    </Popover>
  );
};
```

</details>

https://github.com/iTwin/iTwinUI/assets/9084735/b3ecf1a4-fad8-4eaa-8757-32de2a234b68

https://github.com/iTwin/iTwinUI/assets/9084735/9ca4ebb9-1e65-4731-b727-cd6584fa60fd

Screen reader output when popover opens:

<details>
<summary>Firefox 118 + NVDA 2023.2</summary>

</details>

<details>
<summary>Safari 17 + VoiceOver</summary>

</details>

## Docs

n/a. Changeset not needed for component changes since we haven't released v3 yet.


packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol, line 1: Might be good to add an `@example` for Popover. Apart from the common Popover props, we could also highlight the best accessibility practices here.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -239,36 +249,46 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
...
   );
 
+  const triggerId = `${useId()}-trigger`;
+  const hasAriaLabel = !!props['aria-labelledby'] || !!props['aria-label'];
+
   return (
     <>
       {cloneElementWithRef(children, (children) => ({
+        id: children.props.id || triggerId,
         ...popover.getReferenceProps(children.props),
```
rohan-kadkol, line 259: Is having an id on `children` required? If so, why?

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -239,36 +249,46 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
...
-                className={cx(
-                  { 'iui-popover-surface': applyBackground },
-                  className,
-                )}
-                {...popover.getFloatingProps(rest)}
-                ref={popoverRef}
+              <FloatingFocusManager
+                context={popover.context}
+                modal={false}
+                initialFocus={popover.refs.floating}
```
rohan-kadkol, line 272: Would users ever want to override the initial focus? Because in #1406, we are leaning towards asking the users to handle the focus themselves.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -193,8 +197,16 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
 // ----------------------------------------------------------------------------
 
 type PopoverPublicProps = {
+  /**
+   * Content displayed inside the popover.
+   */
   content?: React.ReactNode;
   children?: React.ReactNode;
```
rohan-kadkol, line 204: Maybe we can add JSDocs to say this is the trigger?

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -193,8 +197,16 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
...
+  /**
+   * Whether the popover adds recommended CSS for background-color, box-shadow, etc.
```
rohan-kadkol, line 206: Would be clearer to say *to what* this be added to

rohan-kadkol (commented):
In the screen recordings you posted, you showed one recording where the "Agree" button grads initial focus and one where it does not. To confirm, what were you trying to test? What is the significance of those two cases in particular?

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: Resolved. I learned that like the second screen recording above and the `Popover.focus.tsx` example, we can use refs to override the focus.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: yes, for labelling.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: the `children` jsdocs don't get exposed anywhere

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: added `@example`. the rest of the content goes in the docs page

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: added

gretanausedaite (approved):

gretanausedaite (commented):
Code looks ok to me.

rohan-kadkol (approved):

apps/storybook/cypress-visual-screenshots/baseline/Table.test.ts-Custom Filter.png
mayank99, line 1: I've changed the logic of custom filter to auto focus the radio button, but I'm hesitant to do the same for ColorPicker, because it already needs significant a11y changes.

should i leave focus on the popover for now?

gretanausedaite: It's hard to focus the first colour in the grid?

rohan-kadkol: > should i leave focus on the popover for now?

Yes, that sounds good for now. But if it's hard to focus to the current color, should we bring back `setFocus`?

If we want to focus the current color without any package code changes, would something like this work? I guess this only works when the popover is just opened. But it doesn't work when the popover is already opened and we try tabbing from the trigger to the popover/active color.

<details>
<summary>code</summary>

image

</details>

PR #1605, opened 2023-09-28
add `'use client'` directive

mayank99:
## Changes

added `"use client"` to barrel file, since this is the only place where components can be imported from.

this makes it possible to use our library in Nextjs App Router without hacks.

see https://react.dev/reference/react/use-client

while not all of our components (for example `Text`) necessarily need to be Client Components, we use context in every single one, so they kinda have to be.

## Testing

n/a.

we can test it in migration tool once released.

## Docs

n/a

### DIFF
--- a/.changeset/chatty-flowers-clean.md
+++ b/.changeset/chatty-flowers-clean.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added `'use client'` directive for React Server Components support.
--- a/packages/itwinui-react/src/index.ts
+++ b/packages/itwinui-react/src/index.ts
@@ -2,6 +2,8 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+'use client';
+
 export { Alert } from './core/Alert/Alert.js';
 
 export { Avatar } from './core/Avatar/Avatar.js';


.changeset/chatty-flowers-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
```
rohan-kadkol, line 2: Since this is something new and not a bug fix, should it be `minor`/`major`?

mayank99: doesn't matter since it will go in 3.0 anyway, but i think it can be seen as `patch` since it "fixes RSC support"?

it's certainly not a new feature that we implemented ourselves.

rohan-kadkol (commented):
> while not all of our components (for example `Text`) necessarily need to be Client Components, we use context in every single one, so they kinda have to be.

To confirm, is context used in `Text.tsx`? I could not find the use of any context there. Maybe you're referring to the use of context somewhere in the polymorphic portion that is common to all components?

mayank99 (commented):
yes, every component uses context via `Box` or `polymorphic`.

https://github.com/iTwin/iTwinUI/blob/66f292e41d3fa3924f3a873a21abbb86e248e6e2/packages/itwinui-react/src/core/utils/functions/polymorphic.tsx#L18

https://github.com/iTwin/iTwinUI/blob/66f292e41d3fa3924f3a873a21abbb86e248e6e2/packages/itwinui-react/src/core/utils/hooks/useGlobals.ts#L20-L21

gretanausedaite (approved):

PR #1606, opened 2023-09-28
updating react DOM nodes inside slider

siddhantrawal:
Updating all the react changes associated with this [PR](https://github.com/iTwin/iTwinUI/pull/1469). Refer to this [comment](https://github.com/iTwin/iTwinUI/pull/1469/files#r1321876551).

## Changes

- Fixed Slider with CustomTickNoToolTip by controlling the state.
- Added `data-iui-orientation` and `data-iui-disabled` attributes.
- Removed `iui-slider-rail` and changed the order of DOM nodes as shown in HTML.
- Thumb uses this  style = `--iui-slider-thumb-position: ${lowPercent}%`

## Testing
React visual and unit test cases passing.

## Docs
N/A

### DIFF
--- a/apps/storybook/src/Slider.stories.tsx
+++ b/apps/storybook/src/Slider.stories.tsx
@@ -132,7 +132,7 @@ export const CustomTickNoTooltip = () => {
   }, []);
 
   return (
-    <div style={{ width: '50%' }}>
+    <>
       <Slider
         min={1}
         max={365}
@@ -155,9 +155,17 @@ export const CustomTickNoTooltip = () => {
       >
         {dateFormatter.format(currentDate)}
       </Text>
-    </div>
+    </>
   );
 };
+
+CustomTickNoTooltip.decorators = [
+  (Story) => (
+    <div style={{ width: '50%' }}>
+      <Story />
+    </div>
+  ),
+];
 
 export const DecimalIncrement = () => {
   return <Slider min={0} max={50} step={2.5} values={[25]} />;


packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,23 +497,36 @@ export const Slider = React.forwardRef((props, ref) => {
...
           {
             'iui-grabbing': undefined !== activeThumbIndex,
           },
           railContainerProps?.className,
         )}
         onPointerDown={handlePointerDownOnSlider}
       >
-        <Box
-          as='div'
-          {...railProps}
```
gretanausedaite, line 518: Don't forget to remove this prop from type.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -94,7 +94,7 @@ WithCustomThumb.args = {
         alignItems: 'center',
         backgroundColor: '#999',
         width: '36px',
-        height: '26px',
+        height: '20px',
```
rohan-kadkol: Just curious, was this necessary?

rohan-kadkol: Ohh, I see. But in the second screenshot too, it seems a bit uncentered vertically. Do you know the root cause of the vertical misalignment? Maybe we need to fix that instead?

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -160,10 +160,6 @@ export type SliderProps = {
    * Allows props to be passed for slider-max
    */
   maxProps?: React.ComponentProps<'span'>;
-  /**
-   * Allows props to be passed for slider-rail
-   */
-  railProps?: React.ComponentProps<'div'>;
```
rohan-kadkol, line 166: I noticed that #1436 didn't add a changeset. So should we add a changeset now for the new props added in that PR? @mayank99 @gretanausedaite

mayank99: there should be a CSS changeset in #1436, and a react changeset in this PR.

this PR should also have no css/html changes if possible.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -130,41 +129,30 @@ export const CustomTickNoTooltip = () => {
...
-              marginRight: '6px',
-            }}
-          >
-            {dateFormatter.format(currentDate)}
-          </Text>
-        </div>
-      }
-      tickProps={{ className: 'some-tick' }}
-    />
+    <div style={{ width: '50%' }}>
```
mayank99, line 132: this change should not be needed

packages/itwinui-css/backstop/tests/slider.html
mayank99, line 1: unrelated to react (undo)

packages/itwinui-css/src/slider/slider.scss
```diff
@@ -84,6 +84,7 @@ $iui-slider-tick-thickness: 1px;
       'min';
     justify-items: center;
     inline-size: fit-content;
+    block-size: 100%;
```
mayank99, line 87: unrelated to react (undo or move to #1469 if necessary)

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,23 +493,36 @@ export const Slider = React.forwardRef((props, ref) => {
...
+      )}
+      {tickMarkArea}
       <Box
-        as='div'
         ref={containerRef}
         {...railContainerProps}
         className={cx(
-          'iui-slider-container',
+          'iui-slider',
+          'iui-progress-slider',
```
mayank99, line 511: i don't think `iui-progress-slider` does anything

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,23 +493,36 @@ export const Slider = React.forwardRef((props, ref) => {
           {minValueLabel}
         </Box>
       )}
+      {maxValueLabel && (
```
mayank99, line 496: i think we can keep the original order

packages/itwinui-react/src/core/Slider/Thumb.tsx
```diff
@@ -156,13 +154,12 @@ export const Thumb = (props: ThumbProps) => {
...
         {...rest}
-        data-index={index}
         ref={thumbRef}
         style={{
           ...style,
-          ...(orientation === 'horizontal'
-            ? { insetInlineStart: `${lowPercent}%` }
-            : { insetBlockEnd: `${lowPercent}%` }),
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
mayank99, line 161: remove this ts-ignore

you can do this:

```tsx
style={{
  ...style,
  '--iui-slider-thumb-position': `${lowPercent}%`,
} as React.CSSProperties}
```

packages/itwinui-css/src/slider/slider.scss
rohan-kadkol: Linking the relevant discussion: https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1341785009

packages/itwinui-css/backstop/tests/slider.html
rohan-kadkol: Relevant discussion: https://github.com/iTwin/iTwinUI/pull/1469#discussion_r1341787847

packages/itwinui-css/src/color-picker/color-picker.scss
```diff
@@ -175,8 +181,7 @@ $iui-hover-box-shadow: 0 0 0 var(--iui-size-3xs) var(--iui-color-border);
   forced-color-adjust: none;
   background-position: 0, 0, calc(var(--iui-size-xs) * 0.5) calc(var(--iui-size-xs) * 0.5);
   background-size: auto, var(--iui-size-xs), var(--iui-size-xs);
-  background-image:
-    linear-gradient(to right, transparent 0%, var(--iui-color-picker-selected-color) 100%),
+  background-image: linear-gradient(to right, transparent 0%, var(--iui-color-picker-selected-color) 100%),
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, in that case, it would be alright to leave these changes in. Because they might happen again in future PR when working on this file. So it's alright to commit these changes in this PR.

But if you want to avoid these changes (might be better since you didn't make any changes to this file), you can commit *just this file* with the `--no-verify` flag.

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -500,24 +493,19 @@ export const Slider = React.forwardRef((props, ref) => {
           {minValueLabel}
         </Box>
       )}
+      {tickMarkArea}
```
gretanausedaite, line 496: tickMarkArea needs to be higher in DOM than before?

rohan-kadkol (commented):
The vertical slider example on the docs site and storybook looks broken. I believe the fix for this will be tracked in https://github.com/iTwin/iTwinUI/pull/1606#discussion_r1346081388

gretanausedaite (commented):
react part looks good to me.

gretanausedaite (approved):

.changeset/little-monkey-eating.md
```diff
@@ -0,0 +1,8 @@
+---
+'@itwin/itwinui-react': minor
```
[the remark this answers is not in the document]
mayank99: the user of this component does not care about data attributes or classes. i think we could have a simple changeset that says something like "Updated internal DOM structure in Slider."

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -446,6 +441,7 @@ export const Slider = React.forwardRef((props, ref) => {
           as='div'
           {...ticksProps}
           className={cx('iui-slider-ticks', ticksProps?.className)}
+          aria-hidden='true'
```
mayank99, line 444: this should probably not have `aria-hidden` because the user might have important text inside the ticks

packages/itwinui-react/src/core/Slider/Slider.tsx
```diff
@@ -483,12 +479,9 @@ export const Slider = React.forwardRef((props, ref) => {
...
       ref={ref}
-      className={cx(
-        'iui-slider-component-container',
-        `iui-slider-${orientation}`,
-        { 'iui-disabled': disabled },
-        className,
-      )}
+      className={cx('iui-slider-container', className)}
+      data-iui-orientation={orientation}
+      data-iui-disabled={disabled}
```
mayank99, line 484: generally you want to use `undefined` for the falsy case, to avoid adding unnecessary data attribute

```suggestion
      data-iui-disabled={disabled ? 'true' : undefined}
```

apps/storybook/src/Slider.stories.tsx
mayank99: then you could use an actual story decorator to make the displayed code nicer.

```tsx
CustomTickNoTooltip.decorators = [
  (Story) => <div style={{ width: '50%' }}><Story /></div>
]
```

rohan-kadkol (approved):

mayank99 (approved):

PR #1610, opened 2023-10-03
add `border` prop to Surface

mayank99:
## Changes

- changed `--iui-surface-border-color` to `--iui-surface-border` so that the entire border value can be set.
- added `border` prop to `Surface`; can be set to false to turn off border.
- removed `--iui-surface-border-radius` and `--iui-surface-background-color` (both unused)

originates from https://github.com/iTwin/iTwinUI/pull/1598#discussion_r1337795106.

## Testing

added unit test.

## Docs

added changesets.


gretanausedaite (approved):

FlyersPh9 (approved):

packages/itwinui-css/src/surface/base.scss
```diff
@@ -8,9 +8,12 @@
   $border-radius: var(--iui-border-radius-1),
   $elevation: var(--iui-shadow-1)
 ) {
-  background-color: var(--iui-surface-background-color, #{$background-color});
-  border: var(--iui-surface-border, #{1px solid $border-color});
-  border-radius: var(--iui-surface-border-radius, #{$border-radius});
-  box-shadow: var(--iui-surface-elevation, #{$elevation});
+  --iui-surface-border: 1px solid #{$border-color};
+  --iui-surface-elevation: #{$elevation};
```
mayank99, line 12: i've made some more adjustments:
- re-declared variables to fix inheritance and support nested surfaces (rare but might be valid in case of e.g. popovers)
- removed `--iui-surface-border-radius` and `--iui-surface-background-color` as they are not used anywhere.

[`20c9dba`](https://github.com/iTwin/iTwinUI/pull/1610/commits/20c9dbaa8db45acfaf167d702c106923eff480f2)

PR #1625, opened 2023-10-11
fix styles being adopted from different window

### LINKED ISSUE #1624: Popout window 'adoptedStyleSheets' error with a DropdownMenu component

Similar to #1579. Expanding a `DropdownMenu` from a popout window results in an error:

```
NotAllowedError: Failed to set the 'adoptedStyleSheets' property on 'Document': Sharing constructed stylesheets in multiple documents is not allowed
```

#### Expected behavior

`DropdownMenu` works correctly in a popout.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/itwinui-react-upgrade-and-popout-forked-hgw4nm?file=/src/App.tsx:390-410

##### Steps to reproduce
1. Click the "Click me!" button to open a popout.
2. Click the "DropdownMenu" button in a popout. Observe that all elements are removed from the root node of a popout and the error is logged.

#### Additional information

Introduced with #1415

mayank99:
## Changes

fixes #1624 (v2 bug)

after debugging together with Gerardus, we found that the main problem was `loaded` being stored in a ref that gets initialized for every single component. this means `DropdownMenu` was calling `new CSSStyleSheet()` (from default `window`) and then trying to adopt it into the child window.

so i:
- moved loaded state out of the hook and into `ThemeProvider`, ensuring that it's shared across components.
- used [`document.defaultView`](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView) to get child window and call `CSSStyleSheet()` from there.

slightly unrelated: i optimized ThemeContext value to be correctly memoized. This reduces the number of times the useEffect inside `useStyles` is called (and the number of re-renders in general).

## Testing

tested with this code in vite playground:

<details>
<summary>App.tsx</summary>

```tsx
import {
  Button,
  DropdownMenu,
  MenuItem,
  ThemeProvider,

const DummyComponent = () => {
  return (
    <div>
      <DropdownMenu
        menuItems={(close) => [
          <MenuItem key='1' onClick={close}>
            Item 1
          </MenuItem>,
          <MenuItem key='2' onClick={close}>
            Item 2
          </MenuItem>,
        ]}
      >
        <Button>DropdownMenu</Button>
      </DropdownMenu>
    </div>
  );
};

const childHtml = `<!DOCTYPE html>
<html lang="en-ie" data-child>
<head>
  <meta charset="utf-8" />
</head>
<body>
  <div id="root"></div>
</body>
</html>`;

function renderChildWindow(container: Element | DocumentFragment) {
  ReactDOM.render(
    <ThemeProvider className='child'>
      <DummyComponent />
    </ThemeProvider>,
    container,
  );
}

function popout() {
  const childWindow = window.open(
    '',
    '',
    'menubar=no,resizable=yes,scrollbars=no,status=no,location=no',
  )!;
  childWindow.document.title = 'Popout window!';
  childWindow.document.write(childHtml);

  setTimeout(() => {
    copyStyles(childWindow.document);
    setTimeout(() => {
      const reactConnectionDiv = childWindow.document.getElementById('root')!;
      renderChildWindow(reactConnectionDiv);
    });
  });
}

  return (
    <ThemeProvider>
      <div>
        <DummyComponent />
        <button onClick={() => popout()}>Pop me out</button>
      </div>
    </ThemeProvider>
  );
}

function copyStyles(targetDoc: Document, sourceDoc: Document = document) {
  const stylesheets = Array.from([
    ...sourceDoc.styleSheets,
    ...sourceDoc.adoptedStyleSheets,
  ]);
  stylesheets.forEach((stylesheet) => {
    const css = stylesheet;
    if (stylesheet.href) {
      const newStyleElement = targetDoc.createElement('link');
      newStyleElement.rel = 'stylesheet';
      newStyleElement.href = stylesheet.href;
      targetDoc.head.appendChild(newStyleElement);
    } else {
      if (css && css.cssRules && css.cssRules.length > 0) {
        const newStyleElement = targetDoc.createElement('style');
        Array.from(css.cssRules).forEach((rule) => {
          newStyleElement.appendChild(targetDoc.createTextNode(rule.cssText));
        });
        targetDoc.head.appendChild(newStyleElement);
      }
    }
  });
}
```

</details>

## Docs

n/a

### DIFF
--- a/packages/itwinui-react/src/core/utils/hooks/useStyles.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useStyles.ts
@@ -16,6 +16,9 @@ const _React = React;
 const useIsomorphicInsertionEffect =
   _React.useInsertionEffect ?? useIsomorphicLayoutEffect;
 
+/** This lives outside the hook, because we only want to load styles once, no matter what.  */
+let loaded = false;
+
 /**
  * Dynamically loads the iTwinUI styles as a constructed stylesheet,
  * falling back to a regular `<style>` tag in `<head>`.
@@ -25,13 +28,12 @@ export const useStyles = (options?: {
   document?: () => Document | undefined;
 }) => {
   const context = _React.useContext(ThemeContext);
-  const loaded = _React.useRef(false);
 
   const includeCss = options?.includeCss ??
     context?.includeCss ?? { withLayer: true };
 
   useIsomorphicInsertionEffect(() => {
-    if (loaded.current || !includeCss) {
+    if (loaded || !includeCss) {
       return;
     }
 
@@ -44,7 +46,7 @@ export const useStyles = (options?: {
 
     loadStyles({ withLayer, document });
 
-    loaded.current = true;
+    loaded = true;
   }, [context, options, includeCss]);
 };
 
--- a/packages/itwinui-react/src/core/utils/hooks/useStyles.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useStyles.ts
@@ -16,9 +16,6 @@ const _React = React;
 const useIsomorphicInsertionEffect =
   _React.useInsertionEffect ?? useIsomorphicLayoutEffect;
 
-/** This lives outside the hook, because we only want to load styles once, no matter what.  */
-let loaded = false;
-
 /**
  * Dynamically loads the iTwinUI styles as a constructed stylesheet,
  * falling back to a regular `<style>` tag in `<head>`.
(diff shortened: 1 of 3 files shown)


gretanausedaite (approved):

packages/itwinui-react/src/core/utils/hooks/useStyles.ts
```diff
@@ -25,12 +25,17 @@ export const useStyles = (options?: {
...
-
-  const includeCss = options?.includeCss ??
-    context?.includeCss ?? { withLayer: true };
 
   useIsomorphicInsertionEffect(() => {
+    if (!context) {
+      return;
+    }
+
+    const loaded = context.stylesLoaded;
```
gretanausedaite, line 34: Storybook does not load styles because it does not have context.

Maybe we could define loaded outside isomorphic effect and not do early return when context is not provided?
```js
  const loaded = !!context ? context.stylesLoaded : _React.useRef(false);
```

Or we could fix storybook styling.

mayank99: looks like v3 storybook has context but v2 one doesn't.

in v2 we support using our components without ThemeProvider, so i guess we'll need to support usage without context.

added fallback in [`b1ff2e7`](https://github.com/iTwin/iTwinUI/pull/1625/commits/b1ff2e7e5afd346d5cde2ecaef02bd60f5978e50)

PR #1626, opened 2023-10-11
fix carousel a11y

### LINKED ISSUE #912: Carousel: tab navigation breaks when slides have focusable elements

When Carousel slides have focusable elements inside, then keyboard navigation breaks. Pressing <kbd>Tab</kbd> repeatedly cycles between two slides and makes it impossible to move forward in tab navigation.

#### Expected behavior

Keyboard navigation should behave normally. See bullet points below for expected behavior _(in italics)_ in context of repro.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/carousel-focus-trap-bug-xkfwsf?

##### Steps to reproduce
1. Click anywhere in viewport, then press <kbd>Tab</kbd> until the Carousel as a whole is focused.
2. Press <kbd>Tab</kbd> again to focus button that reads `1`.
3. Press <kbd>Tab</kbd> again to see slide is advanced and the Carousel is focused again. _(This should have instead either focused the next button `2` or should have focused the button that is outside the carousel)._
4. Press <kbd>Tab</kbd> again to see slide is reset back to first one. _(This should have either focused button `3` or the next button outside the carousel)_

#### Additional information

This issue is likely because of this line:
https://github.com/iTwin/iTwinUI-react/blob/628cc26afc044e1b3f1c09b2489b31fe15b34de1/packages/iTwinUI-react/src/core/Carousel/Carousel.tsx#L78

mayank99:
## Changes

Few different a11y improvements to Carousel:
- Carousel now only adds two <kbd>Tab</kbd> stops the dots (tab) and the active slide (tabpanel), essentially following the tabs pattern. 
  - Finally also fixes #912
  - Firefox and Chrome make scrollable regions focusable by default, so had to add `tabindex="-1"` to dots list and slider.
  - Used `inert` attribute to prevent tabbing into non-active slides.
  - Moved keyboard navigation code from Carousel root into DotsList.
- To make the above more intuitive, the DOM order has been changed so the tablist comes before the tabpanel. Visual order is unchanged.
- Extra roles (`nav`/`ol`/`li`) have all been changed to `div`s because they introduce other issues. ([`nav` needs to be labeled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#labeling_landmarks) and `role=tabpanel` is not allowed on `li`).

## Testing

- [x] Mouse usable ([video](https://github.com/iTwin/iTwinUI/assets/9084735/1a770806-948b-4fd1-8960-80fc8492d339))
- [x] Touch usable ([video](https://github.com/iTwin/iTwinUI/assets/9084735/4578320b-674b-4cc0-b28e-3c111a1dcd6c))
- [x] Keyboard usable ([video](
https://github.com/iTwin/iTwinUI/assets/9084735/b910cda8-b52d-4075-9d19-4022bf27f620))
- [x] Screen-reader usable (video: [NVDA](https://github.com/iTwin/iTwinUI/assets/9084735/b68dedd4-00d5-4f01-a490-63000b4fc3eb))
- [x] Axe passes ([screenshot](https://github.com/iTwin/iTwinUI/assets/9084735/b9a851cc-e609-409f-937b-72375562c58f))

## Docs

added changeset and updated jsdoc.


packages/itwinui-react/src/core/Carousel/Carousel.tsx
```diff
@@ -61,52 +61,11 @@ const CarouselComponent = React.forwardRef((props, ref) => {
...
-      return;
-    }
-
-    const key = event.key;
-    if (key === 'ArrowLeft' || key === 'ArrowRight') {
-      setKeysPressed((old) => ({ ...old, [key]: true }));
-    }
-  };
-
-  const handleKeyUp = (event: React.KeyboardEvent) => {
```
gretanausedaite, line 91: Removing this broke switching slides when focus is on the slide.

packages/itwinui-react/src/core/Carousel/Carousel.tsx
```diff
@@ -144,26 +99,33 @@ const CarouselComponent = React.forwardRef((props, ref) => {
...
  *     <Carousel.Slide>...</Carousel.Slide>
  *   </Carousel.Slider>
- *   <Carousel.Navigation />
  * </Carousel>
  */
 export const Carousel = Object.assign(CarouselComponent, {
   Slider: CarouselSlider,
   Slide: CarouselSlide,
+  /**
+   * Contains the dots and previous/next buttons for navigating the slides. Must be present _before_ the slides in DOM.
```
gretanausedaite, line 123: Aren't DOM changes technically breaking changes?

packages/itwinui-react/src/core/Carousel/Carousel.tsx
mayank99: shouldn't be a problem. APG example behaves the same way: https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/

rohan-kadkol: On the html demo page, when focusing the slide instead of the dots, and then pressing the arrow keys, the slide moves, but the dots do not update.

[15.webm](https://github.com/iTwin/iTwinUI/assets/45748283/0ae09af3-01c6-49a6-9ad0-fd1337e83193)

To reproduce:
* Press <kbd>Tab</kbd> to switch focus from the dots to the slide
* Press <kbd>←</kbd>/<kbd>→</kbd>

To confirm, is this only on my end?

mayank99: the HTML page is not meant to show the behavior, only the visuals. you should test the React components in storybook/playgrounds/website.

gretanausedaite: It's really broken now. I would think to disable slide switching when focused on slide if we don't want to fix it.


https://github.com/iTwin/iTwinUI/assets/83585998/097f4952-64b3-4f16-b274-8005bcc7922a

gretanausedaite: What's happening: focus on slide > click right arrow > it breaks > click tab > focuses slide again

packages/itwinui-react/src/core/Carousel/Carousel.tsx
gretanausedaite: Also update changeset?

packages/itwinui-react/src/core/Carousel/Carousel.tsx
mayank99: arrow keys are used for reading characters in NVDA so it makes sense that it's spelling out "A c t i o n s" once you exit the carousel in storybook



https://dequeuniversity.com/screenreaders/nvda-keyboard-shortcuts

packages/itwinui-react/src/core/Carousel/CarouselNavigation.tsx
```diff
@@ -44,20 +43,14 @@ const PreviousButton = React.forwardRef((props, ref) => {
...
-  } = context;
+  const { slideCount, currentIndex, setCurrentIndex, scrollToSlide } = context;
 
   return (
     <IconButton
       styleType='borderless'
       size='small'
       tabIndex={-1}
-      data-pressed={keysPressed['ArrowLeft'] || undefined}
+      aria-label='Previous'
```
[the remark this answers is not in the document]
mayank99: yeah that was my reasoning

packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
```diff
@@ -49,14 +54,21 @@ export const CarouselSlide = React.forwardRef((props, ref) => {
 
   return (
     <Box
-      as='li'
       className={cx('iui-carousel-slider-item', className)}
       role='tabpanel'
       aria-roledescription='slide'
+      tabIndex={index === currentIndex ? 0 : undefined}
```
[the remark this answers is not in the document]
mayank99: i think it should be the same. when the tab is not active, we add `inert` attribute, which makes it non-focusable.

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
```diff
@@ -86,14 +87,13 @@ export const CarouselSlider = React.forwardRef((props, ref) => {
 
   return (
     <Box
-      as='ol'
-      aria-live='polite'
```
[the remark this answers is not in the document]
mayank99: i believe live region is unnecessary here because we do not support (or recommend) auto-rotating carousels. since the slide changes are triggered manually, the user doesn't need to be "notified" (which is the purpose of a live region).

.changeset/hot-turtles-chew.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Improved carousel accessibility and changed the markup so that the dots are present before the slides.
```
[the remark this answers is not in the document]
mayank99: normally i would, but i think this should be enough. our v3 changelog is going to be too long already

rohan-kadkol (approved):

gretanausedaite (approved):

PR #1632, opened 2023-10-12
adjust gap between adjacent icons

### LINKED ISSUE #1520: Icon: fix padding between icons
Imagine `*` is an icon and `.` is a space.
We want a list of icons to be something like this:

`. . * . . * . . * . .` 

To achieve this we need to update icon css code.

mayank99:
## Changes

fixes #1520

three changes:
- removed size from `.iui-svg-icon` (now only applied on `svg` inside) so that it can use padding
- collapsed padding between adjacent icons
- collapsed padding between searchbox icon and search input

## Testing

| Before | After |
| --- | --- |
| image | image |
| image | image |

## Docs

no docs changes needed. changeset added.


gretanausedaite (approved):
Looks good.

Issue #1633, opened 2023-10-13
Select : Incorrect tags being displayed when we have long string as dropdown menu item in Multi Select component


anjaligondsebentley:
#### Describe the bug (actual behavior)
For multi select component if we have long string in dropdown menu option(I set the width as 300px for select component) and we select the same entry then instead of showing tag with "1item(s)", it shows "2items(s)" though no other item is selected.
![chrome_kKvz6Momt3](https://github.com/iTwin/iTwinUI/assets/113327714/1ff72d1d-8623-4d35-9a84-14b53ed8e175)

#### Expected behavior
Select should show "1item(s)" selected instead of showing "2items(s)" if string length exceeds the dropdown width else it should display selected string in tag. 

#### Reproduction
[Link to a minimal repro: ](https://codesandbox.io/s/react-typescript-forked-7zvs6d?file=/src/App.tsx)

##### Steps to reproduce
1. Create a file with below piece of code
 ```
 import "./styles.css";
import {
  Label,
  LabeledSelect,
  MenuItem,
  MiddleTextTruncation,
  Select,
  SelectValueChangeEvent
} from "@itwin/itwinui-react";
import React from "react";
import { useCallback, useEffect, useMemo, useState } from "react";

export default function App() {
  const [selectedOption, setSelectedOption] = useState<string[]>();
  const list = useMemo(
    () => [
      {
        label: "instance long long very long text ABCDEFGHIJKLMNOPQRSTUVWXYZ",
        value: "Instance1Value"
      },
      { label: "Instance2", value: "Instance2Value" },
      { label: "Instance3", value: "Instance3Value" },
      { label: "Instance4", value: "Instance4Value" },
      { label: "Instance5", value: "Instance5Value" },
      { label: "Instance6", value: "Instance6Value" },
      { label: "Instance7", value: "Instance7Value" },
      { label: "Instance8", value: "Instance8Value" },
      { label: "Instance9", value: "Instance9Value" },
      { label: "Instance10", value: "Instance10Value" },
      { label: "Instance11", value: "Instance11Value" },
      { label: "Instance12", value: "Instance12Value" },
      { label: "Instance13", value: "Instance13Value" },
      { label: "Instance14", value: "Instance14Value" }
    ],
    []
  );
  const onChange = useCallback(
    (value: string, event: SelectValueChangeEvent) => {
      let options = selectedOption ? [...selectedOption] : [];
      if (event === "added") {
        options.push(value);
      }
[...]

Issue #1438, opened 2023-07-25
ComboBox: inaccurate selected items count at small widths


mattl-bentley:
#### Describe the bug (actual behavior)

Multiselect combobox displays the wrong count under certain conditions. See repro details.

#### Expected behavior

Visible tags match selected items.

#### Reproduction
See [codesandbox](https://codesandbox.io/s/blue-glitter-mmlzfy?file=/src/App.tsx) for an example. Notice a single element is selected, but the tag says "+2 item(s)."

The order you select the items changes the display behavior. 

1. clear selected items
2. select "Lofts"
3. select "Townhomes".
Notice the inaccurate count.
4. clear selected items
5. select "Townhomes"
6. select "Lofts"
Notice only "Townhomes" is displayed.

gretanausedaite:
Should be fixed in [2.12.13](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%402.12.13)

PR #1634, opened 2023-10-13
useOverflow: fix visibleCount when item is too big for the space

gretanausedaite:
## Changes

Suggesting a fix for #1633.

Our `useOverflow` hook returns `visibleCount: 0` when item is bigger when space available. I think that should be '1'.

## Testing

Using user provided code in vite sandbox.

Added unit test.

## Docs

TBD

### DIFF
--- a/packages/itwinui-react/src/core/utils/hooks/useOverflow.tsx
+++ b/packages/itwinui-react/src/core/utils/hooks/useOverflow.tsx
@@ -78,7 +78,7 @@ export const useOverflow = <T extends HTMLElement>(
     if (availableSize < requiredSize) {
       const avgItemSize = requiredSize / visibleCount;
       const visibleItems = Math.floor(availableSize / avgItemSize);
-      setVisibleCount(visibleItems);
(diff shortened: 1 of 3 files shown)


mayank99 (approved):
seems to (kinda) work for both #1438 and #1633. it's not worse than what we had before, so i'd say let's go with it.

packages/itwinui-react/src/core/utils/hooks/useOverflow.tsx
```diff
@@ -78,7 +78,7 @@ export const useOverflow = <T extends HTMLElement>(
     if (availableSize < requiredSize) {
       const avgItemSize = requiredSize / visibleCount;
       const visibleItems = Math.floor(availableSize / avgItemSize);
-      setVisibleCount(visibleItems);
+      setVisibleCount(visibleItems > 0 ? visibleItems : 1);
```
mayank99, line 81: maybe add a comment above this so we remember this bug when we open this file again in 2 years

PR #1622, opened 2023-10-11
Tabs: documentation page update after API changes

gretanausedaite:
## Changes

Changes to Tabs documentation page in website.

## Testing

N/A

## Docs

N/A

PR #1636, opened 2023-10-16
Tabs: fix scrollIntoView bug in overflow

gretanausedaite:
## Changes

Removed JS scrollIntoView functionality.
Added css scrollIntoView functionality.

## Testing

TBD

## Docs

TBD

PR #1639, opened 2023-10-16
fix Tabs scrolling the whole page

mayank99:
## Changes

fixes https://github.com/iTwin/iTwinUI/pull/1622#discussion_r1357185138, supersedes #1636

this is done in two stages:
- on first mount, we use `scrollTo` to explicitly set the scroll position of the tablist
- on manual tab changes, we use `scrollIntoView` within an event handler (instead of Effect). this takes into account the `scroll-margin` set in CSS, to ensure that there is some room left near the edges.

## Testing

tested in all three browsers that active tab is scrolled into view and page is not scrolled on load.

https://github.com/iTwin/iTwinUI/assets/9084735/1cb4dec7-3af8-4cc6-a328-2ca0364924ff

https://github.com/iTwin/iTwinUI/assets/9084735/4c0156ba-7201-4dee-8745-4166a9ca48e7

note: i increased the width of the container in horizontal overflow story so that the final tab stops wrapping in test images.

## Docs

n/a


gretanausedaite (approved):
Scrolling is much better!

Side note: I think masks with scroll animation are not working.

mayank99 (commented):
> Side note: I think masks with scroll animation are not working.

It's working in Chrome (the only browser where it's currently supported), same as before.

PR #1659, opened 2023-10-31
added misMatchThreshold for table CSS test.

siddhantrawal:
## Changes
 There was a slight difference in the image inside parts of the table test. Logs were showing errors in this specific image :

Screenshot 2023-10-31 114959

Added misMatchThreshold: 0.1, to avoid causing the breakdown of tests (The above example shows only 0.79 difference).

## Testing
Ran it multiple times locally and it passed properly. Passing on GitHub as well.

### DIFF
--- a/packages/itwinui-css/backstop/scenarios/table.js
+++ b/packages/itwinui-css/backstop/scenarios/table.js
@@ -11,6 +11,7 @@ module.exports = [
   }),
   scenario('Type Extras', {
     selectors: ['#demo-extras'],
+    misMatchThreshold: 0.1,
   }),
   scenario('Type Expandable Rows', {
     selectors: ['#demo-expandable-rows'],


rohan-kadkol (approved):
To confirm, was the CI failing before this? Not sure because looks like the [latest CI run](https://github.com/iTwin/iTwinUI/actions/runs/6655261754) was passing. Was there any PR that might have caused this flakiness?

mayank99 (commented):
i don't think adding threshold is the right thing to do here. it seems like that particular test fails due to scroll position, rather than rendering differences

mayank99:
probably not worth looking into, unless these failures happen too often (>50%). manually re-triggering CI would take less time.

mayank99:
> Should we keep the misMatchThreshold then

no, because these things are unrelated.
- the threshold % is for the diff between two screenshots; useful for rendering differences (e.g. sometimes the colors don't match fully)
- this particular scenario is failing because of different content (based on scroll position), rather than any rendering differences

PR #1660, opened 2023-10-31
Table Ctrl+Shift click logic improved

rohan-kadkol:
## Changes

* Ctrl + Shift click behavior matches that of Windows explorer

<details>

Actions taken:
* Click row 0
* Ctrl click row 2
* Ctrl + Shift click row 6
* Ctrl click row 5
* Ctrl + Shift click row 3

Before:

[Before](https://github.com/iTwin/iTwinUI/assets/45748283/9d850e0b-34d0-413c-bbcc-f81e8304fa21)

After:

[After](https://github.com/iTwin/iTwinUI/assets/45748283/7c5781bf-897f-4328-9d98-1d56fdc9855e)

</details>

* Fix mismatch between Table state and the table that is shown (https://github.com/iTwin/iTwinUI/pull/1660#issuecomment-1791567011)

## Testing

Updated unit test

## Docs

Added changeset

### DIFF
--- a/.changeset/curly-peas-peel.md
+++ b/.changeset/curly-peas-peel.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
+Made Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
--- a/.changeset/curly-peas-peel.md
+++ b/.changeset/curly-peas-peel.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': minor
+'@itwin/itwinui-react': patch
 ---
 
 Made Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
--- a/.changeset/curly-peas-peel.md
+++ b/.changeset/curly-peas-peel.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': patch
+'@itwin/itwinui-react': minor
 ---
 
 Made Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
--- a/.changeset/curly-peas-peel.md
+++ b/.changeset/curly-peas-peel.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
--- a/packages/itwinui-react/src/core/Table/actionHandlers/selectHandler.ts
+++ b/packages/itwinui-react/src/core/Table/actionHandlers/selectHandler.ts
@@ -156,7 +156,7 @@ export const onShiftSelectHandler = <T extends Record<string, unknown>>(
   // If ctrl + shift click, do not lose previous selection
   // If shift click, start new selection
   const selectedRowIds: Record<string, boolean> = !!action.ctrlPressed
-    ? state.selectedRowIds
+    ? { ...state.selectedRowIds }
     : {};
 
   // 1. All rows between start and end are assigned the state of the last selected row
--- a/packages/itwinui-react/src/core/Table/actionHandlers/selectHandler.ts
+++ b/packages/itwinui-react/src/core/Table/actionHandlers/selectHandler.ts
@@ -152,7 +152,7 @@ export const onShiftSelectHandler = <T extends Record<string, unknown>>(
   const isLastSelectedRowIdSelected =
     state.lastSelectedRowId != null
       ? !!state.selectedRowIds[state.lastSelectedRowId]
-      : false;
+      : true; // When no row is selected before shift click, start selecting from first row to clicked row
 
   // If ctrl + shift click, do not lose previous selection
   // If shift click, start new selection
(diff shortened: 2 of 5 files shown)


rohan-kadkol (commented):
I could make the change to match the Shift click pattern followed by Windows Explorer (screen recordings in PR description).

However, I could not solve the issue where the Table does not display the correct state. This happens in some cases.

<details>
<summary>One of the case</summary>

* Shift click row 2
* Shift click row 4
* Shift click row 2
* Ctrl + Shift click row 4
  * The state shows that Rows 0-4 are selected (also seen in the paginator row count)
  * But the rows don't have the checkboxes to be selected
  * Upon expanding a row, the checkboxes appear correctly

[Screen recording](https://github.com/iTwin/iTwinUI/assets/45748283/2d4ddd93-1793-4555-888f-2d5dd36bfe77)

</details>

Any ideas why this could be happening? @iTwin/itwinui-1 @iTwin/itwinui-2

packages/itwinui-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -147,21 +149,26 @@ export const onShiftSelectHandler = <T extends Record<string, unknown>>(
     endIndex = temp;
   }
 
+  const isLastSelectedRowIdSelected =
+    state.lastSelectedRowId != null
+      ? !!state.selectedRowIds[state.lastSelectedRowId]
+      : true; // When no row is selected before shift click, start selecting from first row to clicked row
```
mayank99, line 155: can be simplified
```suggestion
  const isLastSelectedRowIdSelected =
    state.lastSelectedRowId == null || // When no row is selected before shift click, start selecting from first row to clicked row
    !!state.selectedRowIds[state.lastSelectedRowId];
```

mayank99 (commented):
> I could not solve the issue where the Table does not display the correct state. This happens in some cases.

could it be that `instance.selectedFlatRows` is not getting correctly updated?

do we rely on this property to be set anywhere else? if not, we could update our paginator props to compute total count from `selectedRowIds` instead.

rohan-kadkol:
> could it be that `instance.selectedFlatRows` is not getting correctly updated?

I don't think so because based on what Ctrl+Shift click I did in the example, the paginator was indeed showing the correct count. 

> do we rely on this property to be set anywhere else? if not, we could update our paginator props to compute total count from `selectedRowIds` instead.

I found only one use of `instance.selectedFlatRows` and it was in the paginator. After using `selectedRowIds` instead, it still didn't work.

image

mayank99 (commented):
I spent some time looking into this and found that this mismatch is happening due to memoization of Table rows. We need to update the [comparison function](https://github.com/iTwin/iTwinUI/blob/9cfbdbd265d0ae28cfcde9db3756d0f604912419/packages/itwinui-react/src/core/Table/TableRowMemoized.tsx#L163) which probably needs some updated checks.

---

Tangent:

While digging into the old code and PRs, I got the impression that we originally didn't intend to support <kbd>Ctrl</kbd> + <kbd>Shift</kbd>. Then later it was bolted onto the old code which wasn't designed to support this case.

Since our plan is to rewrite the Table entirely (using tanstack table v8), we can more intentionally support all the desired cases from the get-go, to avoid such difficult bugs. It might help if we can use a queue-like data structure to keep track of the order of user's operations.

rohan-kadkol:
> I spent some time looking into this and found that this mismatch is happening due to memoization of Table rows. We need to update the [comparison function](https://github.com/iTwin/iTwinUI/blob/9cfbdbd265d0ae28cfcde9db3756d0f604912419/packages/itwinui-react/src/core/Table/TableRowMemoized.tsx#L163) which probably needs some updated checks.

After looking in this direction for a while, I realized that the problem lied with with how the new state was created.
```diff
- const newState = oldState // incorrect
+ const newState = { ...oldState } // correct
```

With the incorrect version, the old selection state was accidentally also being changed along with the new selection state. Thus `React.memo`'s `arePropsEqual()` returned true since the old state equals the new state.

The render issue is now fixed after https://github.com/iTwin/iTwinUI/pull/1660/commits/fe53c5d72cf679ab80979f54d1f9086767234879

.changeset/curly-peas-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Table's Ctrl+Shift click implementation more consistent with Windows Explorer's implementation.
```
mayank99, line 5: nit: missing verb here

e.g. "**Made** Table's Ctrl+Shift click…"

mayank99 (approved):

.changeset/curly-peas-peel.md
```diff
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': patch
+'@itwin/itwinui-react': minor
```
mayank99, line 2: lets stick with patch, we want to avoid minor releases for a little longer

.changeset/curly-peas-peel.md
rohan-kadkol: I thought to leave it out to be concise. But yes, adding that one word makes it more clearer. Added (https://github.com/iTwin/iTwinUI/pull/1660/commits/8aa84ef16b1a01bebde9047332ee0e8b495d8e98).

.changeset/curly-peas-peel.md
rohan-kadkol: Sounds good, changed (https://github.com/iTwin/iTwinUI/pull/1660/commits/084ceed710dc148853e838e2f950eaa5a67901e5)

rohan-kadkol: Actually, since both changeset points are patches, should I combine them into one changeset file?

mayank99: two changesets is totally fine

PR #1481, opened 2023-08-10
upgrade vite to 4.4 and sb to 7.2

mayank99:
## Changes

brings back #1447 with some more changes.

## Testing

pending.

## Docs

n/a

PR #1669, opened 2023-11-02
migrate from storybook to ladle

mayank99:
## Changes

follow-up to #1668

[ladle](https://ladle.dev/) is pretty much a drop-in replacement for storybook (as advertised). there were no big code changes needed (e.g. no changes to stories except to fix unrelated errors). even the visual testing setup is mostly unchanged.

two differences:
- prod build output goes into `build` folder instead of `storybook-static`
- story urls have changed; we might want to take this opportunity to re-organize/flatten some categories even for our docs site (in a separate PR)

i've tried to keep most of the same setup as before, but there are still ~200 file changes (mostly minor).

**advantages** (compared to storybook): reduced build time, reduced client bundle size, faster load times, much fewer node modules, no major bugs. hopefully now, we'll waste less time working around storybook issues (#1481, #1543) and potentially switch package managers too (#1600) 🤞

## Testing

tested the new setup locally, in CI, and on gh-pages. see live preview: https://itwin.github.io/iTwinUI/1669/react

some visual test images changed but still look ok.

## Docs

updated user-facing text in some places. updated placeholder links in docs site.

no changesets needed as there are no package changes.

**side note**: we should slowly phase this public-facing site out completely as we flesh out our real docs in the coming months. ladle will then be used primarily as a local workshop environment for developing/testing components.


apps/storybook/src/Overview.stories.mdx
```diff
@@ -14,65 +10,32 @@ import { Meta } from '@storybook/addon-docs';
...
+- [v1 storybook](https://itwin.github.io/iTwinUI-react)
+- [v2 storybook](https://itwin.github.io/iTwinUI/legacy/v2/react)
 
-```
-npm install @itwin/itwinui-react
-```
-
-```
-yarn add @itwin/itwinui-react
-```
```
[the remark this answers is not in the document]
mayank99: since the plan is to phase out the public-facing aspect, i've shortened the readme, instead encouraging the use of documentation site. plus, installation instructions are the least useful part

apps/storybook/cypress/support/e2e.js
```diff
@@ -33,7 +33,8 @@ Cypress.on('uncaught:exception', (err, runnable) => {
  * @param {string} storyName
```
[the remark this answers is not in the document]
mayank99: done https://github.com/iTwin/iTwinUI/pull/1669/commits/8e2c9a81c6a3c3c89ee6a9480cfaf25aceacfcdc

apps/storybook/src/Slider.stories.tsx
```diff
@@ -168,7 +167,7 @@ CustomTickNoTooltip.decorators = [
       <Story />
     </div>
   ),
-];
+] satisfies StoryDecorator[];
```
rohan-kadkol, line 170: There are other similar uses of `ComponentXYZ.decorators = ...` that likely also need `satisfies StoryDecorator[]` to fix the type errors. Example places: ButtonGroup L194, Carousel L179, etc.

You can do these changes in this PR, or I had planned going through all stories and fixing type errors. So I can do it then too.

apps/storybook/src/Table.test.ts
```diff
@@ -20,7 +20,7 @@ describe('Table', () => {
     'Global Filter',
     'Initial State',
     'Full',
-    'Full 2',
+    'Full2',
```
rohan-kadkol, line 23: This was an intended change? Does Ladle treat these names differently, thus requiring this change?

apps/storybook/package.json
```diff
@@ -4,44 +4,35 @@
...
-    "eslint-plugin-storybook": "^0.6.11",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
-    "react-table": "^7.1.0",
-    "serve": "^14.0.1",
-    "sb": "~7.0.2",
-    "storybook-dark-mode": "^3.0.0",
-    "typescript": "~5.1.6",
-    "vite": "^4.1.5"
+    "typescript": "*",
```
rohan-kadkol, line 23: Is there a benefit for having `*` as the version for multiple packages?

I remember you once said that typescript sometimes has breaking changes even in minor releases. So with `*`, do we automatically use the latest version and thus might need to often make modifications to our code?

rohan-kadkol (commented):
* Does storybook-static need to be in `apps/storybook/.gitignore`?
* The "will open in a new tab message" in SkipToContentLink.stories.tsx can be removed

apps/storybook/src/Slider.stories.tsx
mayank99: fixed all decorators https://github.com/iTwin/iTwinUI/pull/1669/commits/46c2c6b8047f10ecee307d2aacc430db3e7a9655

apps/storybook/src/Table.test.ts
mayank99: yes, storybook automatically adds a space, whereas ladle doesn't.

notice the extra "-" in the storybook url
- storybook: https://itwin.github.io/iTwinUI/react/?path=/story/core-table--full-2
- ladle: https://itwin.github.io/iTwinUI/1669/react/?story=core--table--full2

apps/storybook/package.json
mayank99: `*` basically means "anything is fine" rather than "i want latest". i did it this way to help avoid unnecessary duplicate versions as we use some of the same deps in multiple places in the monorepo

the actual typescript version is controlled in the root package.json, so this should be fine fine

mayank99 (commented):
> * Does storybook-static need to be in `apps/storybook/.gitignore`?

i initially left it out because it would result in unnecessary changes in the working tree when switching branches, but i've added it now since we're close to merging. https://github.com/iTwin/iTwinUI/pull/1669/commits/92fdd4db7467f15cc9b67cfd217d815f6e77b680

> * The "will open in a new tab message" in SkipToContentLink.stories.tsx can be removed

done (https://github.com/iTwin/iTwinUI/pull/1669/commits/28eb01a022162729d3d7437cd677289f36ecd226). i also adjusted the padding because the images were changing anyway

rohan-kadkol (approved):

PR #1656, opened 2023-10-25
updated alert docs.

siddhantrawal:
## Changes
Updated documentation for the Alert component.

## Testing

N/A

## Docs

N/A

PR #1673, opened 2023-11-06
remove scrollbar styles from docs

mayank99:
## Changes

not sure why these were added in the first place; the default scrollbars work fine.

my guess is the site was made before we had scoped styling support (`iui-root`), so iTwinUI scrollbars may have been messing up docs scrollbars, but now we are running into the opposite issue (https://github.com/iTwin/iTwinUI/issues/941#issuecomment-1492479881, https://github.com/iTwin/iTwinUI/pull/1656#discussion_r1382236355), so it's better to just use defaults.

## Testing

tested that default scrollbars are appearing ok

## Docs

n/a

### DIFF
--- a/apps/website/src/pages/_global.css
+++ b/apps/website/src/pages/_global.css
@@ -50,37 +50,6 @@
     list-style: none;
     padding: 0;
   }
-
-  @media (pointer: fine) and (hover: hover) {
-    :root {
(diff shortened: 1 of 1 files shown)


rohan-kadkol (approved):
Confirmed that the default scrollbars look alright on Edge, Chrome, and Firefox 👍

PR #1676, opened 2023-11-06
Add small tree size

FlyersPh9:
## Changes

- Add `data-iui-size=small` option.
- Decrease indentation for default size tree.

## Testing

- Added new CSS test for different sizes.
- Updated CSS and React test images.

## Docs

- Added sizes section to tree documentation page.


packages/itwinui-css/src/tree/tree.scss
```diff
@@ -2,19 +2,54 @@
...
+            (var(--iui-size-m) + var(--_iui-tree-node-padding-inline) * 2)
+          ); // TODO: The last part is added for checkboxes, how does this work without checkboxes? 🤔
+        background-color: var(--_iui-tree-line-color, transparent);
+        transition: background-color var(--iui-duration-1) ease;
+        z-index: 1; // TODO: This gets the lines above the background hover, but not above the buttons when focused
+      }
+    }
+
+    .iui-button {
+      z-index: 2;
```
FlyersPh9, line 50: Really would like to avoid `z-index`.  The 2 instances of `z-index` address:
1. When hovering over a row, the hover background color was covering up the lines.
2. When focusing the expander button, the line was covering up the focus outline.

mayank99: use `z-index: -1` on the pseudo-element?

apps/css-workshop/pages/tree.html
```diff
@@ -1239,6 +1246,1300 @@ <h2>With Icons</h2>
...
+          >
+            <li
+              class="iui-tree-item"
+              role="treeitem"
+              tabindex="0"
+              aria-expanded="true"
+            >
+              <div
+                class="iui-tree-node"
+                style='--level: 1;'
```
mayank99, line 1303: am i understanding correctly that the old markup still works for existing features?

the preferable scenario would be:
1. `--level` on tree-node continues to work for tree without lines
2. `--level` on sub-tree works for all cases, with and without lines

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -2,19 +2,54 @@
...
+
+    .iui-sub-tree {
+      position: relative;
+
+      &::before {
+        content: '';
+        inline-size: 1px;
+        block-size: 100%;
+        position: absolute;
+        inset: 0 auto 0
```
mayank99, line 38: can you explain what math is going on here? a little hard to follow.

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: The 2nd point doesn't seem to be an issue anymore, my guess is reverting back to the small button for all tree sizes makes this not an issue anymore.

I'll look into the 1st point tomorrow, here is a screenshot of the bug when all `z-index`s are removed.
Screenshot 2023-12-05 at 5 17 49 PM

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: Oh gosh, let's see if I can remember. 😅

GitHub isn't showing the code, so here it is for reference:
```sass
::before {
  inset: 0 auto 0 calc(var(--_iui-expander-button-width) * 0.5 * var(--level, 0) + (var(--iui-size-m) + var(--_iui-tree-node-padding-inline) * 2));
}
```

`var(--_iui-tree-node-padding-inline) * 2` is to account for the padding on the tree body as well as the gap between checkbox & expander button.
Screenshot 2023-12-05 at 5 29 57 PM Screenshot 2023-12-05 at 5 30 18 PM

`var(--_iui-expander-button-width) * 0.5` is to account for the expander button, but only half of it, so the line is centered on the button.

`var(--level, 0) + (var(--iui-size-m)` is the indentation level as well as the checkbox which uses M width.

Will need to think up some way to accommodate for tree without checkboxes, because that won't work at the moment.

apps/css-workshop/pages/tree.html
```diff
@@ -1239,6 +1246,1300 @@ <h2>With Icons</h2>
       </ul>
     </section>
 
+    <hr />
+    <h2>Sizes</h2>
```
rohan-kadkol, line 1250: Would it be good to have titles to tell which size is used in each tree demo? Because now it's a little confusing if the upper tree is small or the default

packages/itwinui-css/src/tree/tree.scss
rohan-kadkol: To make the calculation easier to read, would it be better to create SASS variables only in this scope? You could then add comments for each variable to explain what it does.

So something along the lines of:
```sass
&::before {
  $_tree-body-padding-and-gap: calc(var(--_iui-tree-node-padding-inline) * 2) // explanation ...
  $_expander-button-width: var(--_iui-expander-button-width) * 0.5 // explanation ...
  $_indentation: var(--level, 0) + (var(--iui-size-m)) // explanation ...
  // ... other helper variables

  $_left-padding: combination of the helper variables
}
```

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -2,19 +2,54 @@
...
   list-style: none;
   border-radius: var(--iui-border-radius-1);
   background-color: var(--iui-color-background);
+
+  &[data-iui-size='small'] {
+    --_iui-tree-node-padding-inline: var(--iui-size-2xs);
+    --_iui-tree-row-height: calc(var(--iui-size-3xs) * 14);
+  }
+
+  &[data-iui-lines='true'] {
```
mayank99, line 22: just a thought: should we handle lines in a separate PR, to unblock small size? it seems to be the most complicated part of this PR and it's not even part of the request.

.changeset/many-avocados-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': minor
+'@itwin/itwinui-css': minor
+---
+
+Add small tree size option & decrease indentations.
```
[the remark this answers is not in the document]
FlyersPh9: Updated, let me know if this is more clear.

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -2,19 +2,22 @@
...
-// padding of small borderless button + 1px transparent border
-$iui-expander-inline-padding: calc(var(--iui-size-2xs) + 1px);
-
-// width of small borderless button (button is 1:1 ratio)
-$iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
+$iui-active-outline: 1px solid var(--iui-color-border-accent);
 
 .iui-tree {
   @include mixins.iui-reset;
+  --_iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
```
mayank99, line 9: this can be moved back to sass variable as it's not changed anywhere

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -88,10 +88,10 @@ $iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
     &-label {
       flex: 1;
       min-inline-size: 0;
-      padding-inline-start: $iui-expander-inline-padding;
+      // padding-inline-start: var(--_iui-tree-node-padding-inline);
```
mayank99, line 91: this can be removed? already handled by `gap`

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -41,37 +44,34 @@ $iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
...
-    margin-inline-start: calc(#{$iui-expander-button-width} * (var(--level, 0)));
+    min-block-size: var(--_iui-tree-row-height);
+    margin-inline-start: calc(var(--_iui-expander-button-width) * 0.5 * var(--level, 0));
     overflow: hidden;
     padding-inline-start: var(--iui-size-3xs);
     inline-size: 100%;
 
     &-icon {
-      margin-inline: $iui-expander-inline-padding;
+      margin-inline: var(--_iui-tree-node-content-icon-margin-inline);
```
mayank99, line 66: this variable does not exist

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -41,37 +44,34 @@ $iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
...
 
-      &,
       svg {
```
mayank99, line 68: this looks unrelated, should be undone. I think it was done this way because the `-content-icon` class could be applied on the `svg` itself

packages/itwinui-css/src/tree/tree.scss
```diff
@@ -2,19 +2,22 @@
...
   @include mixins.iui-reset;
+  --_iui-expander-button-width: calc(var(--iui-component-height-small) + 2px);
+  --_iui-tree-node-padding-inline: var(--iui-size-xs);
+  --_iui-tree-row-height: calc(var(--iui-size-3xs) * 18);
+
   list-style: none;
   border-radius: var(--iui-border-radius-1);
   background-color: var(--iui-color-background);
+
+  &[data-iui-size='small'] {
```
mayank99, line 17: can you handle the react part too? should be fairly simple to add a new prop which sets this data attribute (see `Button.tsx` as example)

apps/css-workshop/pages/tree.html
FlyersPh9: We don't use `<h2>` to call out the sizes in other demo pages, but we do include the size within the button / select labels for example.  I've pushed a change so the first node of each size tree calls out the size.

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: Will do this in a separate PR addressing lines.  Resolving this comment to declutter this PR for now.

apps/css-workshop/pages/tree.html
rohan-kadkol: Then I guess the second node's text also needs to be changed? Because the second node says "Facility 2", which seems a bit odd since there is no "Facility 1".

Maybe the nodes can be named "Small Tree Node 1" and "Small Tree Node 2"?

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: In the small tree I am using this:
```sass
<svg-placeholder data-iui-icon-size="s"></svg-placeholder>
```

If I remove the `&,` we get this (pink added to help with the visuals):
Screenshot 2023-12-07 at 1 07 21 PM

If I add the `&,` back, we get this alignment issue:
Screenshot 2023-12-07 at 1 07 45 PM

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: Hmm, must have been leftover from something I was trying?  I don't know, removed.

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: Can do, reverted back to SASS var.

packages/itwinui-css/src/tree/tree.scss
rohan-kadkol: I guess then we can exclude that specific case using `:not()`?

```suggestion
      &:not([data-iui-icon-size='s']),
      svg {
```

mayank99: Some more context:

In React, the user only passes the actual svg, and we do not use the `Icon` component (which consists of `iui-svg-icon` and `data-iui-icon-size`). In the next major, we should replace all uses of the `iui-icon-style`/`iui-icon-color` mixins with `iui-svg-icon`, but right now we need to avoid such breaking changes.



I think the simplest way would be to include `iui-icon-style(m)` for default tree and `iui-icon-style(s)` for small tree, and keep the html unchanged.

FlyersPh9: Ahh, right, React handles icon differently.  Dang, I'm getting rusty.

`&,` added back in and some CSS added back at the top to set the icon mixin to S for small tree.  Also cleaned up the HTML by removing `.iui-svg-icon` and the size.

packages/itwinui-react/src/core/Tree/Tree.tsx
```diff
@@ -57,6 +57,10 @@ type FlatNode<T> = {
 export type NodeRenderProps<T> = Omit<NodeData<T>, 'subNodes'>;
 
 export type TreeProps<T> = {
+  /**
+   * Modify size of the tree.
+   */
+  size?: 'small';
```
mayank99, line 63: i think we should also reserve a value for default size

```suggestion
  size?: 'default' | 'small';
```

apps/website/src/content/docs/tree.mdx
```diff
@@ -15,6 +15,14 @@ thumbnail: Tree
 
 A tree can be used to organize data in an application specific way, or it can be used to sort, filter, group, or search data as the user deems appropriate. Each line of data level may begin with an eye icon for toggling its visibility. In this context, this icon is called specialty checkbox.
 
+## Size
+
+There are 2 different sizes available, which can be applied to any tree. The medium size is a default and should always be the first choice, unless there are good reasons to use the small version.
```
mayank99, line 20: I don't believe we use the term "medium" anywhere. There's a "default" size and a "small" size. Also we should mention the `size` prop.

```suggestion
There are 2 different sizes available. The default size should suffice for most cases. When a smaller version of the tree is needed, use `size="small"`.
```

(Rephrase as necessary)

examples/Tree.size.tsx
mayank99, line 1: we only need to show the small size here, because default is already implied and shown in the main example.

suggest removing the default size from this example and changing the file name (and the name in `index.tsx`).

mayank99: also we should add a story for "Small" size in `Tree.stories.tsx` and a visual test in `Tree.test.ts` next to it

packages/itwinui-react/src/core/Tree/Tree.tsx
mayank99: we should probably add a `'default'` string wherever it's missing. it's more explicit that way

a bunch of other components already do it this way
image

apps/website/src/content/docs/tree.mdx
FlyersPh9: Ok, I'll go with your suggestion.  Just a heads up though that I copy / pasted that text from [the buttons page](https://itwinui.bentley.com/docs/button#size), so we should also remove medium from there as well.

examples/Tree.size.tsx
rohan-kadkol: > we only need to show the small size here, because default is already implied and shown in the main example.

Or maybe it's better to show a side-by-side comparison so users can know what's best for them?

mayank99: no it's way too much code, the examples need to be kept simple.

apps/css-workshop/pages/tree.html
```diff
@@ -20,6 +20,13 @@
       import "@itwin/itwinui-icons-elements/chevron-right";
       import "@itwin/itwinui-icons-elements/placeholder";
     </script>
+    <script>
+      customElements.define('svg-chevron-right-small', class extends HTMLElement {
+        connectedCallback() {
+          this.innerHTML = `<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" width="1rem" height="1rem" fill="var(--iui-color-icon-muted, currentColor)"><path d="m5.525 2-1.05 1.05L9.425 8l-4.95 4.95L5.525 14l6-6-6-6Z"/></svg>`;
```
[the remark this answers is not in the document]
FlyersPh9: PR created just now:
https://github.com/iTwin/iTwinUI-icons/pull/76

packages/itwinui-css/src/tree/tree.scss
FlyersPh9: Change made.

examples/Tree.size.tsx
FlyersPh9: I also like having the small next to the default size, similar to the button demo.

However, the button is like 3 lines of code for each button where the tree on the other hand is a lot of lines, so I understand wanting to keep the demo code slim.

Renamed to `Tree.small.tsx`.

mayank99: also `TreeSizeExample` → `TreeSmallExample`

apps/css-workshop/pages/tree.html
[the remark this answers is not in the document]
mayank99: for the purpose of this PR, the icon should also be added internally within itwinui-react and then used within expander of small tree.

this is actually a little more involved, so you might need help. let me know

apps/css-workshop/package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {},
   "devDependencies": {
     "@itwin/itwinui-css": "*",
-    "@itwin/itwinui-icons-elements": "0.2.0",
+    "@itwin/itwinui-icons-elements": "0.20.0",
```
mayank99, line 9: this will likely cause visual tests to fail because of some icon changes. best done in a separate PR

mayank99: #1732 is now merged

apps/css-workshop/pages/tree.html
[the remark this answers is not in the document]
mayank99: (i thought i left this comment before)

we should probably use default ChevronRight for default tree size, and ChevronRight**Small** for small tree size, unless it is _intentional_ to use ChevronRight**Small** for both sizes

examples/Tree.size.tsx
FlyersPh9: Done!

apps/css-workshop/pages/tree.html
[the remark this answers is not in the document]
FlyersPh9: I've done `git merge origin/main` and using the small chevron for small tree, default chevron for default tree.  Checking the changes tab & running test images now.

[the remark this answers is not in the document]
mayank99: i've pushed the same update in react

mayank99 (approved):
LGTM

packages/itwinui-react/src/core/Tree/Tree.tsx
```diff
@@ -57,6 +57,10 @@ type FlatNode<T> = {
 export type NodeRenderProps<T> = Omit<NodeData<T>, 'subNodes'>;
 
 export type TreeProps<T> = {
+  /**
+   * Modify size of the tree.
+   */
+  size?: 'default' | 'small';
```
rohan-kadkol, line 63: Is this needed? Or I guess it's obvious without it too? I saw this included in DropdownButton, but maybe this doesn't apply here.

```suggestion
  /**
   * Modify size of the tree.
   * 
   * @default 'default'
   */
  size?: 'default' | 'small';
```

packages/itwinui-react/src/core/Tree/Tree.tsx
```diff
@@ -334,6 +340,7 @@ export const Tree = <T,>(props: TreeProps<T>) => {
           onKeyDown={handleKeyDown}
           onFocus={handleFocus}
           className={className}
+          data-iui-size={size}
```
rohan-kadkol, line 343: When `size` of `"default"` is passed, `data-iui-size` is set to `default`. But in the HTML demo, `data-iui-size` is not set for the default size.

This doesn't affect any styles. But to confirm, is setting this attribute when `size="default"` alright?

packages/itwinui-react/src/core/Tree/Tree.tsx
```diff
@@ -159,6 +163,7 @@ export const Tree = <T,>(props: TreeProps<T>) => {
     className,
     nodeRenderer,
     getNode,
+    size,
```
rohan-kadkol, line 166: Should `size="default"` be set as the default? It works without it too, but would this be better for consistency?

apps/css-workshop/pages/tree.html
```diff
@@ -1239,6 +1240,334 @@ <h2>With Icons</h2>
...
+                  class="iui-button-icon iui-tree-node-content-expander-icon  iui-tree-node-content-expander-icon-expanded"
+                  aria-hidden="true"
+                ></svg-chevron-right-small>
+              </button>
+              <svg-placeholder
+                class="iui-tree-node-content-icon"
+                aria-hidden="true"
+              ></svg-placeholder>
+              <span class="iui-tree-node-content-label">
+                <div class="iui-tree-node-content-title">Small tree</div>
```
rohan-kadkol, line 1280: nit: Since now there is just a small tree demo, this can be reverted back to Facility 1.

rohan-kadkol (approved):
Just a few small comments. The nit comment can be addressed in other PRs. But everything else LGTM.

packages/itwinui-react/src/core/Tree/Tree.tsx
mayank99: doesn't matter too much, but we can do this:
```suggestion
          data-iui-size={size === "small" ? "small" : undefined}
```

packages/itwinui-react/src/core/Tree/Tree.tsx
mayank99: sure

```suggestion
    size = 'default',
```

apps/css-workshop/pages/tree.html
mayank99: i'll leave this as-is since it's not part of the component code

PR #1661, opened 2023-11-01
fix infinite rerenders in `Select`

mayank99:
## Changes

fixes #1657

in react 17, `popover.refs.setFloating` was being called repeatedly on each render, thus causing more re-renders. so added early return.

~~bumped floating-ui as well, to ensure it's up-to-date.~~

## Testing

tested by recreating react 17 sandbox from linked issue.

## Docs

added changeset.

Issue #1657, opened 2023-10-27
`<Select>` will crash when used with react v17.x.x


mdastous-bentley:
#### Describe the bug (actual behavior)
![image](https://github.com/iTwin/iTwinUI/assets/8309609/8666f192-fae8-44cd-9964-080705ab6411)

#### Reproduction
[Link to a minimal repro](https://codesandbox.io/s/young-shadow-5lvddj?file=/src/App.tsx)

##### Steps to reproduce
Click to expand the `<Select> `component

Only occurs with `@itwin\itwinui-react` 3.0.0

mayank99:
spent some time looking into this; i've traced the problem to this line getting called on every render, thus causing even more re-renders:

https://github.com/iTwin/iTwinUI/blob/7d313a2b21fcbaa45b92e75e3a7e7de7eef4c538/packages/itwinui-react/src/core/Select/Select.tsx#L445

mayank99:
fixed in 3.0.1

PR #1681, opened 2023-11-08
fix Select positioning after first render

### LINKED ISSUE #1679: `Select` menu is offset when opened for 2nd time

`<Select popoverProps={{ placement: 'bottom-end' }} />`
1. Click once, menu opens properly
2. Close menu
3. Click, menu opens with offset

Similar bug is visible in storybook too.
<https://itwin.github.io/iTwinUI/react/?story=input--select--basic>
If you open it for seconds time, big empty space appears below it

#### Expected behavior

Menu opens in same position every time

#### Reproduction
Link to a minimal repro: 
<https://stackblitz.com/edit/stackblitz-starters-ywppax>

##### Steps to reproduce
`<Select popoverProps={{ placement: 'bottom-end' }} />`
1. Click once, menu opens properly
2. Close menu
3. Click, menu opens with offset

#### Additional information

itwinuibug

mayank99:
## Changes

fixes #1679

This PR effectively undoes #1661 which was the cause of the issue. In short: after the menu closes, we were still holding onto the old DOM reference, thus causing issues with calculating new position when menu re-opens (it's a new DOM node).

After investigating, I found that the real cause of #1657 was that we were not memoizing the random value used in `useId` fallback, so it was returning a different value on each render, and this was causing an infinite loop (presumably because we are using the id in a `key`, which resets all state). The simple fix is to memoize the random value.

## Testing

tested that Select stories work correctly

tested `placement: 'bottom-end'` working correctly in vite playground, in both react 17 and 18.

## Docs

added changeset.

### DIFF
--- a/.changeset/sixty-seahorses-count.md
+++ b/.changeset/sixty-seahorses-count.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue where `Select` was being positioned incorrectly after first render.
--- a/.changeset/sixty-seahorses-count.md
+++ b/.changeset/sixty-seahorses-count.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Fixed an issue where `Select` was being positioned incorrectly after first render.
+Fixed an issue where `Select`'s menu was being positioned incorrectly after its first render.
--- a/packages/itwinui-react/src/core/utils/hooks/useId.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useId.ts
@@ -18,4 +18,5 @@ export const useId = () => {
 // This is needed to avoid bundlers trying to import non-existing export.
 // Read more: https://github.com/webpack/webpack/issues/14814
 const _React = React;
-const useUniqueValue = _React.useId ?? (() => getRandomValue(10));
+const useUniqueValue =
+  _React.useId ?? (() => React.useMemo(() => getRandomValue(10), []));
--- a/packages/itwinui-react/src/core/utils/hooks/useId.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useId.ts
@@ -18,5 +18,6 @@ export const useId = () => {
 // This is needed to avoid bundlers trying to import non-existing export.
 // Read more: https://github.com/webpack/webpack/issues/14814
 const _React = React;
+
 const useUniqueValue =
   _React.useId ?? (() => React.useMemo(() => getRandomValue(10), []));
--- a/.changeset/curly-turtles-yawn.md
+++ b/.changeset/curly-turtles-yawn.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Memoize `useToaster`'s return value so that it can be more reliably used in dependency arrays.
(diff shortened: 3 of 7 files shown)


packages/itwinui-react/src/core/utils/hooks/useId.ts
```diff
@@ -18,4 +18,5 @@ export const useId = () => {
 // This is needed to avoid bundlers trying to import non-existing export.
 // Read more: https://github.com/webpack/webpack/issues/14814
 const _React = React;
-const useUniqueValue = _React.useId ?? (() => getRandomValue(10));
+const useUniqueValue =
+  _React.useId ?? (() => React.useMemo(() => getRandomValue(10), []));
```
mayank99, line 22: should we backport this to v2? seems like a major oversight, and most of our users are not using react 18 yet.

.changeset/sixty-seahorses-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed an issue where `Select` was being positioned incorrectly after first render.
```
rohan-kadkol, line 5: To confirm, do we need to say the *dropdownMenu* was positioned incorrectly?

```suggestion
Fixed an issue where `Select`'s dropdownMenu was being positioned incorrectly after its first render.
```

rohan-kadkol (commented):
Confirmed that #1657 and #1679's issues are fixed on React 18👍

This may not be related to this PR, but I noticed some UI jump when clicking on the select for the first time when the Select has `popoverProps={{ placement: 'bottom-end' }}`. The dropdown menu seems to be stuck to the right, and then instantly spreads across all the width. (You might need to manual move the video slider slowly to notice it.)


[35.webm](https://github.com/iTwin/iTwinUI/assets/45748283/c3682eab-1211-4811-9ecf-4f350147bb7b)

.changeset/sixty-seahorses-count.md
mayank99: how about this?
```suggestion
Fixed an issue where `Select` menu was being positioned incorrectly after first render.
```

mayank99 (commented):
>  The dropdown menu seems to be stuck to the right, and then instantly spreads across all the width.

this is probably because it needs an extra render to match the width. not sure if there's a better way as it's done using floating-ui middleware

.changeset/sixty-seahorses-count.md
rohan-kadkol: Sounds good, but I feel with the `'s`, the meaning is a bit clearer. Also, without the `its`, it could be unclear if the "first render" is of the *Select component* or of *its menu*.

```suggestion
Fixed an issue where `Select`'s menu was being positioned incorrectly after its first render.
```

But maybe I'm nitpicking 😅 Your code suggestion is good too.

rohan-kadkol (approved):

rohan-kadkol (commented):
**UPDATE**:
Confirmed the #1657 and #1679's issues are fixed on React 17 too.

Also noticed that Select's menu jump issue that I mentioned earlier does not happen in React 17.

packages/itwinui-react/src/core/utils/hooks/useId.ts
rohan-kadkol: Yes, I feel we should, especially since it's just 1-2 lines. Although #1679 and #1657 are I think v3 only bugs, this backport might fix some bug that we are not aware of.

PR #1682, opened 2023-11-08
memoize toaster object

mayank99:
## Changes

when working on the migration tool, i noticed `toaster` object was used in dependency arrays in multiple arrays, so it makes sense to memoize it.

## Testing

no other code changes, so everything works the same.

## Docs

n/a

### DIFF
--- a/.changeset/curly-turtles-yawn.md
+++ b/.changeset/curly-turtles-yawn.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Memoize `useToaster` return value so it can be more reliably used in dependency arrays.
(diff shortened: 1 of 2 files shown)


.changeset/curly-turtles-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Memoize `useToaster` return value so it can be more reliably used in dependency arrays.
```
rohan-kadkol, line 5: nit (optional): Not sure if this is the correct grammar or just my personal preference.

```suggestion
Memoize `useToaster`'s return value so that it can be more reliably used in dependency arrays.
```

rohan-kadkol (approved):

PR #1683, opened 2023-11-08
fix borderless expandable block transition

mayank99:
## Changes

In #1396, we added `transform: translateY(-100%);` to combat the sudden shift that results from `padding-block`. but this was mainly tested on very small content with default variant. When used with larger content and borderless variant, the content shift is super noticeable as the trigger is transparent.

to improve this, i'm now transitioning a small fixed value (`translateY(-1rem)`) and opacity.

## Testing

Before:

https://github.com/iTwin/iTwinUI/assets/9084735/869ac429-7a01-4aa7-865a-6c1d318c86ef

After:

https://github.com/iTwin/iTwinUI/assets/9084735/96ff2a79-1103-49cc-b2c4-9f3a48a6676a

## Docs

n/a


mayank99 (commented):
@FlyersPh9 I wonder if we should make the hover/focus background of borderless variant opaque to avoid this problem altogether. 🤔

```diff
- --_iui-expandable-block-header-background-color-hover: var(--iui-color-background-transparent-hover);
+ --_iui-expandable-block-header-background-color-hover: var(--iui-color-background-hover);
```

https://github.com/iTwin/iTwinUI/assets/9084735/8b945c20-76a1-4fc5-a70f-2c311ba89f03

FlyersPh9:
> @FlyersPh9 I wonder if we should make the hover/focus background of borderless variant opaque to avoid this problem altogether. 🤔

Look at it above both `background` and `background-backdrop`. If it looks good, I'm fine with this change.  👍

mayank99:
> Look at it above both `background` and `background-backdrop`. If it looks good, I'm fine with this change. 👍

i've pushed this change in [`68368bd`](https://github.com/iTwin/iTwinUI/pull/1683/commits/68368bd1cd7c1a5339266c39c79d64a3d6e7eb1d). see deploy preview:
- [`background`](https://itwin.github.io/iTwinUI/1683/react/?arg-background=var%28--iui-color-background%29&story=core--expandableblock--borderless)
- [`background-backdrop`](https://itwin.github.io/iTwinUI/1683/react/?story=core--expandableblock--borderless)

you can also play around with all 4 themes using the controls at the bottom. let me know what you think

FlyersPh9 (approved):
Looks good to me!

PR #1690, opened 2023-11-13
flatten all stories

mayank99:
## Changes

As discussed today, all story names have been flattened to reduce unnecessary complexity.

The categories were previously also present in the urls, so the story urls have changed as a result. Placeholders in docs pages have been updated to use new story urls.
  - Previously: `"Core/Alert"` (`Empty` story) → `?story=core--alert--empty`
  - Now: `"Alert"` (`Empty` story) → `?story=alert--empty`

As a bonus: Now that we know these urls won't change, I've added a `useEffect` to redirect old (v2) storybook urls to new ones.

## Testing

image

## Docs

N/A


mayank99 (commented):
docs will be updated separately as they're completely unrelated

PR #1696, opened 2023-11-17
flatten docs categories using collections

mayank99:
## Changes

This PR does 3 things:
- [`f98f67f`](https://github.com/iTwin/iTwinUI/pull/1696/commits/f98f67f9eef767dc52f0dac82008e117833eb826): Uses [Astro content collections](https://docs.astro.build/en/guides/content-collections) to make it easier to query pages.
  - Frontmatter `layout` does not need to be set anymore, as it will be manually used in `[...slug].astro`.
  - Frontmatter `group` only needs to be set for "intro" and "utilities". All others are assumed to be "components".
- [`2ad7f8d`](https://github.com/iTwin/iTwinUI/pull/1696/commits/2ad7f8d28b32f7b293a814620a118aefbd609c3b): Fixed LeftSidebarContent to use flattened categories (except "Utilities", as discussed)
- [`02cb06d`](https://github.com/iTwin/iTwinUI/pull/1696/commits/02cb06d5176c72086c0e81e012031a678ac06641): Uses [`htm`](https://github.com/developit/htm) to reduce some of the repetition in LeftSidebarContent.

## Testing

Tested website locally to ensure all categories are showing correctly, all urls are unchanged, and links are working.

## Docs

N/A


mayank99:
@siddhantrawal That delay is unrelated and only happens locally (even before this PR), because it is running the code on demand. In the deployed prod version, the html has already been generated ahead of time so there is no delay.

apps/website/src/content/docs/alert.mdx
```diff
@@ -1,9 +1,7 @@
 ---
 title: Alert
 description: A small box to quickly grab the user's attention and communicate a brief message.
-layout: ./_layout.astro
 propsPath: '@itwin/itwinui-react/esm/core/Alert/Alert.d.ts'
-group: core
 thumbnail: Alert
 ---
```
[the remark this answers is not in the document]
mayank99: was caused by #1656. fixed in [`9bef589`](https://github.com/iTwin/iTwinUI/pull/1696/commits/9bef589b6e8068ff1777bdd833241aa2dbe9af5c)

apps/website/src/pages/docs/components.astro
```diff
@@ -1,16 +1,11 @@
...
-    url,
-    title: frontmatter.title,
-    description: frontmatter.description,
-    thumbnail: frontmatter.thumbnail,
-    group: frontmatter.group,
-  }))
-  .filter((page) => page.group);
+const componentPages = (
+  await getCollection('docs', ({ data: { group } }) => group !== 'intro')
+).map(({ slug, data }) => ({ url: `/docs/${slug}`, ...data }));
```
rohan-kadkol, line 8: I'm a little unfamiliar with what "slug" is. Looking at the types, it seems to be the component names.

image

But looking at [this docs section](https://docs.astro.build/en/core-concepts/routing/#route-priority-order), I thought slug has something to do with a rest parameter for the path.

apps/website/src/components/LeftSidebarContent.astro
```diff
@@ -1,109 +1,86 @@
...
-);
-sections.set(
-  'typography',
-  docsPages.filter((item) => item.group === 'typography'),
-);
-sections.set(
-  'utilities',
-  docsPages.filter((item) => item.group === 'utilities'),
-);
+  return html`<a href="${url}" aria-current=${isCurrent ? 'page' : 'false'} ...${props}>
```
rohan-kadkol, line 20: To confirm, what "repition" was reduced after using `htm`?

apps/website/src/pages/docs/[...slug].astro
```diff
@@ -0,0 +1,21 @@
+---
+import { getCollection } from 'astro:content';
+import Layout from './_layout.astro';
+
+export async function getStaticPaths() {
+  return (await getCollection('docs')).map((entry) => {
+    let slug;
+    if (entry.slug) {
```
Comment (rohan-kadkol) [inline apps/website/src/pages/docs/[...slug].astro:8]: Just curious, how are the types for `entry.slug` generated?

image

rohan-kadkol (approved):

apps/website/src/pages/docs/components.astro
mayank99: slug is part of the url. in this case it simply maps to the file name. see https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs

what you're linking to is the `[...]` syntax, which creates a catch-all dynamic route. so a single `[...slug].astro` file can be used to render all routes under `docs/`

apps/website/src/components/LeftSidebarContent.astro
mayank99: without htm, i would need to repeat all this code three times (each instance of NavLink)
image

basically, htm allows me to define inline components in the frontmatter

apps/website/src/pages/docs/[...slug].astro
Comment (mayank99) [reply apps/website/src/pages/docs/[...slug].astro:8]: it's done automatically by astro, but it's not important (at least i don't find the types useful). there is some info here: https://docs.astro.build/en/guides/content-collections/#the-astro-directory

Issue #1617, opened 2023-10-05
Tooltips not showing in Dropdown's submenu


IvanJK97:
#### Describe the bug (actual behavior)

Tooltips in dropdown's submenu are not showing

#### Expected behavior

Should show tooltip when hovering over a submenu item

#### Reproduction
Link to a minimal repro: 
https://codesandbox.io/s/menu-item-with-tooltip-forked-z86p2h?file=/src/App.tsx
Observe that Item#4 tooltip does not show when hovered.

##### Steps to reproduce
1.  Create a dropdown menu with a menu item that has a submenu
1.  Create some menu items wrapped with tooltip in the submenu
1.  Hovering over those menu items does not show tooltip

#### Additional information

We recently upgraded from 2.11.6 to 2.12.10, which was when we started seeing this issue.

mayank99:
I think the root cause of this issue is the same as #1137, which was fixed in v3.

For v2, it might also get addressed together with #1616, so I'll close this for now.

PR #1698, opened 2023-11-20
append nested submenus out of parent menu

### LINKED ISSUE #1616: Nested submenus in dropdown is not being shown

Dropdown menu component's is not displaying nested submenus. It shows the first layer, but not subsequent layers.

#### Expected behavior

Should show nested submenus regardless of depth

#### Reproduction
Link to a minimal repro: 
Can be seen in Storybook demo: https://itwin.github.io/iTwinUI/react/?path=/story/core-dropdownmenu--submenu
Observe that Item 4's submenu items 7 and 8 are not showing.

##### Steps to reproduce
1.  Create a dropdown menu with a menu item with a submenu
1.  In the submenu, create another menu item with a submenu
1.  Items are not shown in the nested submenu

#### Additional information

Seems to be a fairly recent regression. Eric mentioned storybook was working last week.

mayank99:
## Changes

fixes #1616 and #1617 (v2 bugs, can be reproduced in [v2 storybook](https://itwin.github.io/iTwinUI/legacy/v2/react/?path=/story/core-dropdownmenu--submenu))

the main problem was that the parent menu always has `overflow: auto` (after #1576), and we were appending sub-menus into the parent menu, thus causing clipping.

to fix this, i am now appending into the parent's parent (which will be some div that does not have `overflow: auto`).

| Before | After |
| --- | --- |
| image | image |

the code for this component is very delicate, so i've arrived at this minimal solution after lots of iteration and playing around. i had to replace the `onBlur` code with `onKeyDown` to avoid breaking keyboard focus behavior.

## Testing

verified that nested submenus open correctly in storybook.

## Docs

n/a

### DIFF
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -121,13 +121,11 @@ export const MenuItem = React.forwardRef<HTMLLIElement, MenuItemProps>(
           if (subMenuItems.length > 0) {
             setIsSubmenuVisible(true);
             event.preventDefault();
-            event.stopPropagation();
           }
           break;
         }
         case 'ArrowLeft': {
           parentMenuItemRef?.current?.focus();
-          event.stopPropagation();
           event.preventDefault();
           break;
         }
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -192,16 +190,20 @@ export const MenuItem = React.forwardRef<HTMLLIElement, MenuItemProps>(
...
           delay={100}
           content={
             <div
               onMouseLeave={() => setIsSubmenuVisible(false)}
-              onBlur={(e) => {
-                !!(e.relatedTarget instanceof Node) &&
-                  !subMenuRef.current?.contains(e.relatedTarget as Node) &&
-                  !subMenuRef.current?.isEqualNode(e.relatedTarget as Node) &&
+              onKeyDown={(e) => {
+                if (!e.altKey && e.key === 'ArrowLeft') {
```
rohan-kadkol, line 203: To confirm, is this so that the browser's "go back" keyboard shortcut will take precedence? If so, what will happen if the user uses a browser that has some other shortcut? Or maybe the user has overridden the shortcut somehow to something else?

Since the <kbd>Alt</kbd> equivalent on the Mac is <kbd>Option</kbd>, but the "go back" shortcut is <kbd>Command</kbd>+<kbd>←</kbd>, is the browser "go back" keybaord shortcut given preference on Mac too?

rohan-kadkol (approved):

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: i'm assuming you're asking about `e.altKey`, so . i've kept that part the same as before, so it doesn't work for Mac currently.

if we improve this someday, we should probably do it for all components

PR #1700, opened 2023-11-21
fix Table sort triggering after closing filter

### LINKED ISSUE #1550: Table is sorted on enter key when it should only be filtered
When adding a filter to a column, if I am using enter key instead of "Apply" button, it filters the column **AND** sorts (as column header is focused, so table takes that the enter key means sorting as well)

image
image

Would expect that when filter is active, sorting would not be triggered as well.

itwinui-react version: 2.12.6

mayank99:
## Changes

1. added `stopPropagation` to avoid bubbling keydown event from filter button upwards to the header-cell. this fixes #1550.
2. added `:has(:focus-visible)` (with `:focus-within` as fallback) to avoid showing sort icon after filter dialog closes and focus moves back.

## Testing

both keyboard and mouse behavior can be tested in storybook: https://itwin.github.io/iTwinUI/react/?story=table--full2

Before:

https://github.com/iTwin/iTwinUI/assets/9084735/ac30c04e-6cbe-4f70-ac68-36cc5361580f

After:

https://github.com/iTwin/iTwinUI/assets/9084735/82cd363e-9d02-4eb7-9135-cc68c4b20072

## Docs

n/a


packages/itwinui-css/src/table/base.scss
```diff
@@ -90,16 +92,23 @@
...
+
+    @supports not selector(:has(+ *)) {
+      &:focus-within {
+        background-color: var(--iui-color-background-hover);
+
+        :is(.iui-table-sort, .iui-table-filter-button) {
+          opacity: 1;
+        }
+      }
+    }
```
rohan-kadkol, line 111: I read the second bullet point of the PR description but am still a little confused about why this change was needed. Could you elaborate?

rohan-kadkol (approved):

PR #1702, opened 2023-11-22
fix Select uncontrolled mode

mayank99:
## Changes

I noticed that the basic Select usage shown in jsdocs and website was not updating state, so i've fixed that by adding internal uncontrolled state. Preference is given to user-passed `value` prop.

## Testing

Tested that uncontrolled state works fine in playground and on docs website. Added unit test.

## Docs

Added changeset.

Although this behavior should be mentioned on docs page too, I didn't touch it because the whole thing needs some restructuring/rewriting.

### DIFF
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -15,6 +15,7 @@ import {
   Portal,
   useMergedRefs,
   SvgCheckmark,
+  useLatestRef,
 } from '../utils/index.js';
 import type { CommonProps } from '../utils/index.js';
 import { SelectTag } from './SelectTag.js';
@@ -236,8 +237,8 @@ export const Select = React.forwardRef(
 
     const {
       options,
-      value,
-      onChange,
+      value: valueProp,
+      onChange: onChangeProp,
       placeholder,
       disabled = false,
       size,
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -298,11 +302,22 @@ export const Select = React.forwardRef(
...
+            // update internal value state and also call external onChange
+            if (isSingleOnChange(onChangeRef.current, multiple)) {
+              setUncontrolledValue(option.value);
+              onChangeRef.current?.(option.value);
               hide();
             } else {
-              onChange?.(option.value, isSelected ? 'removed' : 'added');
+              setUncontrolledValue((prev: T[]) =>
+                isSelected
+                  ? prev?.filter((i) => option.value !== i)
```
rohan-kadkol, line 314: Since prev is of type `T[]`, is the `?` needed?


```suggestion
                  ? prev.filter((i) => option.value !== i)
```

rohan-kadkol (approved):

rohan-kadkol (commented):
Should there be at least one story and/or website example that shows the uncontrolled Select?

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: you should never trust typescript. `uncontrolledValue` being initialized with `undefined`, so it may still be possible (can't hurt if not)

mayank99 (commented):
> Should there be at least one story and/or website example that shows the uncontrolled Select?

the main website example is already uncontrolled

packages/itwinui-react/src/core/Select/Select.tsx
rohan-kadkol: Then would adding `undefined` as a possible type on L312 be better?

```diff
- setUncontrolledValue((prev: T[]) =>
+ setUncontrolledValue((prev: T[] | undefined) =>
```

mayank99: it's fine, the runtime JS code is unaffected and user-facing TS types are also unaffected.

rohan-kadkol: nit: That's true, but I mean if we add `undefined` to the possible types of `T`, then typescript would explicitly give an error when we don't add the `?` when accessing properties of T.

But without adding `undefined`, we have to remember that `T` can be `undefined` and safeguard accordingly.

PR #1704, opened 2023-11-23
simplify related components and remove imports

mayank99:
## Changes

1. All component imports within mdx files have been removed, using [`astro-auto-import`](https://github.com/delucis/astro-auto-import).
2. `RelatedComponents` component has been removed, in favor of a `related` frontmatter property, which will automatically insert the markup. Styling has been simplified.

## Testing

N/A

## Docs

N/A


apps/website/astro.config.ts
```diff
@@ -10,13 +10,26 @@ import rehypeSlug from 'rehype-slug';
...
   site: 'https://itwinui.bentley.com',
-  integrations: [react(), mdx(), sitemap()],
+  integrations: [
+    AutoImport({
+      imports: [
+        './src/components/PropsTable.astro',
+        './src/components/LiveExample.astro',
+        './src/components/Placeholder.astro',
+        { examples: 'AllExamples' },
+      ],
```
rohan-kadkol, line 27: To confirm, when the imports are not used in the file, are the imports still added? If so, are there any perceivable performance issues?

apps/website/src/content/docs/combobox.mdx
```diff
@@ -3,13 +3,12 @@ title: ComboBox
 description: A dropdown list that allows users to type a value to filter the options.
 propsPath: '@itwin/itwinui-react/esm/core/ComboBox/ComboBox.d.ts'
 thumbnail: ComboBox
+related:
+  - dropdownmenu
+  - select
+  - popover
```
rohan-kadkol, line 9: Are we not having `Label` in this related components list any more?

apps/website/src/content/docs/dropdownmenu.mdx
```diff
@@ -3,13 +3,12 @@ title: Dropdown menu
 description: The dropdown menu displays a list of selectable choices and options.
 propsPath: '@itwin/itwinui-react/esm/core/DropdownMenu/DropdownMenu.d.ts'
 thumbnail: Menu
+related:
+  - select
+  - combobox
+  - popover
```
rohan-kadkol, line 9: Are we not including Button any more?

apps/website/src/pages/docs/_layout.astro
```diff
@@ -16,7 +17,8 @@ const currentPage = new URL(Astro.request.url).pathname;
     <div class='content-wrapper' id='article'>
       <h1 id='overview'>{content.title}</h1>
       <slot />
-      <div class='bottom-links'>
+      <div class='aftermain'>
+        {content.related ? <RelatedComponents related={content.related} /> : null}
```
rohan-kadkol, line 21: non-blocking: Just a thought. Would it be good to keep the related components section near the top of the page, just after the first code example? Users may find it easy to see all related components at once before deciding which component from that category best suits their needs.

rohan-kadkol (commented):
Should the "Related Components" section title show up in the right side nav?

apps/website/astro.config.ts
mayank99: it should get tree shaked, plus this is all at build time so the code runs ahead of time anyway

apps/website/src/content/docs/combobox.mdx
mayank99: we don't have a label page anymore

apps/website/src/content/docs/dropdownmenu.mdx
mayank99: i'm not sure why we had button here, i changed it to more relevant components

apps/website/src/pages/docs/_layout.astro
mayank99: no, this is "extra" content. the documentation for the current component should come first

for deciding which component, i think we should have text explanation, similar to how we mention **toast** in the intro section of **alert** page

mayank99 (commented):
> Should the "Related Components" section title show up in the right side nav?

the table of contents is generated from the mdx file, which does not contain such a heading, so it will not appear. i don't think it's a big deal

PR #1707, opened 2023-11-27
expose underlying drop event in `FileUpload`

### LINKED ISSUE #1611: FileInput: Support directory upload
#### Feature
File Input should support exposing the file structure inside dropped directories. We'd like to migrate from [react-dropzone](https://github.com/react-dropzone/react-dropzone) to the iTwin implementation, but we cannot until we can support dropping a file structure inside a directory.

#### Examples

Folder upload

mayank99:
## Changes

resolves #1611 by simply adding a second parameter to `onFileDropped` prop.

other than that, the component also correctly merges event handlers passed as props.

## Testing

tested behavior in vite playground using this code (with unchanged helpers from the [linked issue's sandbox](https://github.com/iTwin/iTwinUI/issues/1611#issuecomment-1762170236)).

```tsx
<>
  <FileUpload
    onFileDropped={async (_, event) => {
      const directoryStructure =
        await dataTransferItemListToDirectoryStructure(
          event.dataTransfer.items,
        );
      setDirectoryStructure(directoryStructure);
    }}
  >
    <p>Drop files here</p>
  </FileUpload>

  {directoryStructure && (
    <FileTree directoryStructure={directoryStructure} />
  )}
</>
```

</details> 

updated affected unit test.

## Docs

added changeset. update jsdocs for `onFileDropped` prop. updated `fileupload` docs page to more accurately describe our components.


packages/itwinui-react/src/core/FileUpload/FileUpload.tsx
```diff
@@ -69,10 +62,7 @@ export const FileUpload = React.forwardRef((props, ref) => {
     e.stopPropagation();
 
     // only set inactive if secondary target is outside the component
-    if (
-      isDragActive &&
-      !fileUploadRef.current?.contains(e.relatedTarget as Node)
-    ) {
+    if (isDragActive && !e.currentTarget?.contains(e.relatedTarget as Node)) {
```
rohan-kadkol, line 65: * To confirm, what is the `relatedTarget` in dragLeave events?
* What is the difference between `e.target` and `e.currentTarget`?

rohan-kadkol (approved):

packages/itwinui-react/src/core/FileUpload/FileUpload.tsx
mayank99: > * To confirm, what is the `relatedTarget` in dragLeave events?
> * What is the difference between `e.target` and `e.currentTarget`?

The only thing that changed in this PR is `fileUploadRef.current` → `e.currentTarget`. The [`currentTarget` property](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) is always the element on which the event handler is attached, so it avoids the need to use a ref.

`target` is usually not relevent unless using [event delegation](https://javascript.info/event-delegation) (which we don't). And for `relatedTarget`, you need to look at some [very specific docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget).

> Also, do you recommend any particular docs to better understand the drag-drop API?

MDN has a good AI-free writeup: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop

PR #1709, opened 2023-11-28
change `placement` in vertical `ButtonGroup`

### LINKED ISSUE #1652: `<IconButton>` within a vertical `<ButtonGroup>` tooltip obscures adjacent buttons
#### Feature
When `<IconButton>` is placed within `<ButtonGroup orientation='vertical'>` the tooltip appears above the hovered button covering up the button above it.  Ideally either the system is smart enough to place the tooltip in the correct position or allow the developer to define the tooltip placement location.

#### Examples

The 2nd button is hovered and it's tooltip is covering up the button above it.
Screenshot 2023-10-24 at 10 27 01 AM

mayank99:
## Changes

fixes #1652 

1. `ButtonGroupContext` will be used to store `orientation`
2. `IconButton` will read `ButtonGroupContext` to decide default placement of tooltip
3. `labelProps.placement` can be used to override it if needed

## Testing

updated ButtonGroup stories to use `label`.

updated unit test to check for `labelProps`.

## Docs

added changeset and jsdoc for new prop


packages/itwinui-react/src/core/Buttons/IconButton.tsx
```diff
@@ -71,7 +83,12 @@ export const IconButton = React.forwardRef((props, ref) => {
   );
 
   return label ? (
-    <Tooltip content={label} ariaStrategy='none'>
+    <Tooltip
+      content={label}
+      ariaStrategy='none'
+      placement={buttonGroupOrientation === 'vertical' ? 'right' : 'top'}
+      {...labelProps}
```
rohan-kadkol, line 90: Would it be safer to move the `{...labelProps}` at the top of the props passed to `<Tooltip>` to prevent the user from overriding these props?


```suggestion
      {...labelProps}
      content={label}
      ariaStrategy='none'
      placement={buttonGroupOrientation === 'vertical' ? 'right' : 'top'}
```

.changeset/sixty-chefs-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`IconButton`s inside a vertical `ButtonGroup` will show tooltips on the right side by default, to avoid obscuring adjacent buttons in the group. This can be customized using `labelProps.placement`.
```
rohan-kadkol, line 5: Should this also mention that `labelProps` is a new prop added to `IconButton`?

packages/itwinui-react/src/core/Buttons/IconButton.tsx
```diff
@@ -71,7 +83,12 @@ export const IconButton = React.forwardRef((props, ref) => {
   );
 
   return label ? (
-    <Tooltip content={label} ariaStrategy='none'>
+    <Tooltip
+      content={label}
+      ariaStrategy='none'
+      placement={buttonGroupOrientation === 'vertical' ? 'right' : 'top'}
```
rohan-kadkol, line 89: In the viewer, I think the vertical button group is placed in the bottom right corner (not sure). So would `left` be better when `buttonGroupOrientation === 'vertical'`?

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -8,6 +8,16 @@ import { useOverflow, useMergedRefs, Box } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { FloatingDelayGroup } from '@floating-ui/react';
 
+// ----------------------------------------------------------------------------
+
+/** This context is used for letting descendant IconButtons know the ButtonGroup's orientation. */
+export const ButtonGroupContext = React.createContext<string | undefined>(
```
rohan-kadkol, line 14: Do we want to support allowing users to set the `labelProps` of all `IconButton` descendants by setting it just once in the `ButtonGroup`? Or would that be too specific and thus better to not support and thus not clutter the `ButtonGroup`'s available props?

.changeset/sixty-chefs-chew.md
mayank99: something like this?
```suggestion
`IconButton`s inside a vertical `ButtonGroup` will now show tooltips on the right side by default, to avoid obscuring adjacent buttons in the group. The behavior of these tooltips can also be customized using the new `labelProps` prop, which can contain `labelProps.placement`, etc.
```

packages/itwinui-react/src/core/Buttons/IconButton.tsx
mayank99: But toolbars can also be on the left side, so there is no right or wrong answer. The important part is not obscuring the adjacent buttons in the button-group. For anything else, the user should configure `labelProps`.

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
mayank99: No, this part is not controllable by the user. All we're doing here with the context is changing the default behavior of IconButton.

The user can always create a wrapper component if they want multiple IconButtons to behave similarly.

packages/itwinui-react/src/core/Buttons/IconButton.tsx
mayank99: at best, i can see something like this, since we exclude `content` and `ariaStrategy` from the types:
```suggestion
      content={label}
      ariaStrategy='none'
      {...labelProps}
      placement={buttonGroupOrientation === 'vertical' ? 'right' : 'top'}
```

.changeset/sixty-chefs-chew.md
rohan-kadkol: Yes, sounds good! 👍

packages/itwinui-react/src/core/Buttons/IconButton.tsx
rohan-kadkol: Ohh, right, but to confirm, should we disallow overriding `content` and `ariaStrategy since we exclude it in the type?

```suggestion
      placement={buttonGroupOrientation === 'vertical' ? 'right' : 'top'}
      {...labelProps}
      content={label}
      ariaStrategy='none'
```

mayank99: whoops youre right, my version has it flipped

rohan-kadkol (approved):

PR #1710, opened 2023-11-29
advanced Popover cases

mayank99:
## Changes

WIP


packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -234,6 +244,10 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
     placement = 'bottom-start',
     onVisibleChange,
     closeOnOutsideClick = true,
+    middleware,
+    //
+    // extra props
+    positionReference,
     //
     // dom props
```
rohan-kadkol, line 252: Should this be called "public props" to stay consistent with `PopoverPublicProps`

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -262,6 +277,14 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
   const triggerId = `${useId()}-trigger`;
   const hasAriaLabel = !!props['aria-labelledby'] || !!props['aria-label'];
 
+  useIsomorphicLayoutEffect(() => {
+    if (!positionReference) {
+      return;
+    }
+    popover.refs.setPositionReference(positionReference);
+    return () => void popover.refs.setPositionReference(null);
```
rohan-kadkol, line 285: Why is `void` needed here?

rohan-kadkol (approved):

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: no, those are dom props, as in they are passed to the underlying div

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: `void` is syntactic sugar for wrapping with `{}`. it discards the return value (if any).
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void#non-leaking_arrow_functions

PR #1716, opened 2023-11-30
Truncate first `select-tag` in `select-tag-container`

### LINKED ISSUE #1633: Select : Incorrect tags being displayed when we have long string as dropdown menu item in Multi Select component

rohan-kadkol:
## Changes

Fixes #1633.

* Truncate the `select-tag` when there is only one tag in `select-tag-container`

When just one long list item is selected:

|Before #1634|Current (after #1634)|With this PR|
|:-|:-|:-|
|image|image|image|
|incorrect overflow count|showing just one long cut-off tag = we can't see the overflow count = we don't know how many items are really selected|it still shows only one long tag, but the tag is truncated instead of cutoff = we know there is only one selected tag|

Unrelated:
* Removed an extra space from a test name

## Testing

All tests still pass

## Docs

Added changeset


packages/itwinui-react/src/core/Select/SelectTagContainer.tsx
mayank99, line 1: i'm not sure about this solution, shouldn't we use CSS to truncate the overflowing text instead of showing "+1 items"? would be much more useful that way.

rohan-kadkol: To confirm, so with CSS, are you suggesting something along these lines:

```scss
@mixin iui-select-tag {
  // ...
  
  // Truncate the first tag
  &:first-child {
    max-inline-size: 100%;
  }
}
```

Or another solution would be using a `<MiddleTextTruncation>` for the first tag. But I guess that is not recommended ([docs](https://itwinui.bentley.com/docs/select#truncation))?

image

mayank99: `::only-child` selector would probably work better.

this is how i expect it to look:


Middle text truncation should definitely be avoided except when needed for very specific use cases.

packages/itwinui-css/src/select/select.scss
```diff
@@ -102,11 +102,16 @@
...
+  // Truncate when only one tag is present
+  &:only-child {
+    max-inline-size: 100%;
+  }
 }
 
 @mixin iui-select-tag-label {
   @include tag.iui-tag-label;
-  display: inline-flex;
+  display: block;
```
mayank99, line 114: i think flexbox was needed for centering. do you want to try `display: flex`?

.changeset/brown-turtles-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
```
mayank99, line 2:
```suggestion
'@itwin/itwinui-css': patch
'@itwin/itwinui-react': patch
```

packages/itwinui-css/src/select/select.scss
```diff
@@ -102,11 +102,16 @@
   align-items: center;
   block-size: 80%;
   max-block-size: calc(var(--iui-size-s) * 3);
+
+  // Truncate when only one tag is present
+  &:only-child {
+    max-inline-size: 100%;
```
mayank99, line 108: now i'm wondering if we even need the `:only-child`. shouldn't the max-width always be 100%, regardless of whether it's the only child?

.changeset/brown-turtles-drop.md
rohan-kadkol: Ohh, to confirm, since this is a CSS only change, do we need to add `itwinui-react`?

packages/itwinui-css/src/select/select.scss
rohan-kadkol: I had changed `inline-flex` to `block` to get the ellipses to show up. Because with `inline-flex` or `flex`, the ellipses didn't show up:

|`block`|`flex`/`inline-flex`|
|-|-|
|image|image|

packages/itwinui-css/src/select/select.scss
rohan-kadkol: Ohh, that's true, removed `:only-child` (https://github.com/iTwin/iTwinUI/pull/1716/commits/877b8b0df93811001f5c8a41e680b338bb07ca22)

.changeset/brown-turtles-drop.md
mayank99: yes, because itwinui-react produces its own stylesheet

packages/itwinui-css/src/select/select.scss
mayank99: flexbox children will refuse to collapse beyond their "real" width by default. could you try adding `min-inline-size: 0` to the child element?

rohan-kadkol: As discussed, `select-tag-label` does not have any child tag.

image

So, I went ahead with `display: block` instead of `display: inline-flex`, and so also removed the previous `align-items` attribute on `select-tag-label`. (https://github.com/iTwin/iTwinUI/pull/1716/commits/e1956468dbee883b57c17a253f2dbe6ee66a94db)

mayank99 (approved):

PR #1718, opened 2023-12-01
Add border thickness variable

FlyersPh9:
## Changes

- Add `--iui-border-thickness` variable.
- Use `--iui-border-thickness` variable in all components & demo page elements.
- Update `theme.js` to allow changing `--iui-border-thickness` from default value `1px` to double `2px`.

## Testing

When updating all components I set the border thickness variable temporarily to 2px to confirm everything was indeed changing.  I then set the value back to 1px, ran visual tests, and got no failures.

## Docs

None yet, will require the variable to be added to the variables documentation page.
May also require iTwinUI-Layouts to be updated to make use of the new variable.


FlyersPh9:
Closing for now, may revisit if requested again.

PR #1723, opened 2023-12-04
bump floating-ui

mayank99:
## Changes

bumped `@floating-ui` to latest 0.26.x version. needed to make some changes:
- **Tooltip**: added `.nativeEvent` to event object for `reference` prop, because it was throwing an error (some floating-ui code relies on this property).
- **ComboBoxInput**: added a ref to freeze the previous state of `isOpen` for access inside `onClick`, because floating-ui was already opening the menu (on focus) before the click handler executed, causing the menu to immediately close after opening.

## Testing

everything should work the same way. all tests pass.

## Docs

n/a. didn't add changeset as this change should be "transparent" to the user.

### DIFF
--- a/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
+++ b/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
@@ -196,8 +196,12 @@ const useTooltip = (options: TooltipOptions = {}) => {
       ...interactions.getReferenceProps(),
     }).forEach(([key, value]) => {
       if (typeof value === 'function') {
-        reference.addEventListener(domEventName(key), value);
-        cleanupValues[key] = value;
+        // manually add `.nativeEvent` (react concept) because floating-ui needs it
+        const patchedHandler = (event: Event) => {
+          value({ ...event, nativeEvent: event });
+        };
+        reference.addEventListener(domEventName(key), patchedHandler);
+        cleanupValues[key] = patchedHandler;
       } else if (value) {
         cleanupValues[key] = reference.getAttribute(key);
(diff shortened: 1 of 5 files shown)


packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -196,8 +196,12 @@ const useTooltip = (options: TooltipOptions = {}) => {
       ...interactions.getReferenceProps(),
     }).forEach(([key, value]) => {
       if (typeof value === 'function') {
-        reference.addEventListener(domEventName(key), value);
-        cleanupValues[key] = value;
+        // manually add `.nativeEvent` (react concept) because floating-ui needs it
+        const patchedHandler = (event: Event) => {
```
rohan-kadkol, line 200: I saw that you mentioned in the PR description that without this change, some error was thrown. Just curious, under what circumstances was that error thrown? Was it thrown from some of our stories, examples from the docs site, etc.?

rohan-kadkol (approved):

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: it's thrown by the component when `reference` prop is used and you try to focus the trigger element.

i noticed it when one of the tests started failing and confirmed it in the playground. 

here's the source code in floating-ui which uses this property:
https://github.com/floating-ui/floating-ui/blob/cff301f4cb714ac0e699de5948460f1b8c571110/packages/react/src/hooks/useFocus.ts#L121

PR #1725, opened 2023-12-05
Table body scrollable horizontally even when no rows

### LINKED ISSUE #1204: not able to clear column filter in iTwin table
We're using an itwin react table with a column filter. If we search the data presented in the table, we will get the expected result and a horizontal scroll bar based on the width. If the table data does not include the searched result, the horizontal scroll bar will disappear and we will be unable to clear the filter.

for reference:
https://codesandbox.io/s/itwinui-react-minimal-example-forked-wiirrf?file=/src/TreeComponent.tsx

rohan-kadkol:
## Changes

Fixes #1204

WIP

#### Background context
The table header is intentionally non-scrollable for the user. So, if the user wants to see the last column, they need to scroll the table body which programmatically modifies the `scrollLeft` of the table header to keep both scroll positions in sync.

#### The problem in the issue
No rows in the table = no horizontal overflow = the table body becomes non-scrollable. Consequently, since the table header's scroll is always in sync with the now non-scrollable table body, there is no way for the user to scroll horizontally and see columns that are out of the scroll view.

#### This PR
* Adds a dummy `<div>` to the table body, that serves as a shadow root
  * Inside this div is a another div with the width equal to the width of the table header. This ensures that the table body is always scrollable enough to see all column headers, even when no rows are present.

## Testing

Tested with the below code in the vite playground. This code is just the [Table -> HorizontalScroll story](https://itwin.github.io/iTwinUI/react/?story=table--horizontal-scroll&theme=dark) with a filter added to the second last row, among a few minor changes.

<details>
<summary>Code</summary>

import { Table, tableFilters } from '@itwin/itwinui-react';
import type { CellProps, Column } from '@itwin/itwinui-react/react-table';
import React from 'react';

const App = () => {
  const data = React.useMemo(
    () => [
      {
        product: 'Product 1',
        price: 5,
        quantity: 500,
        rating: '4',
        deliveryTime: 5,
      },
      {
        product: 'Product 2',
        price: 12,
        quantity: 1200,
        rating: '1',
        deliveryTime: 25,
      },
      {
        product: 'Product 3',
        price: 2.99,
        quantity: 1500,
        rating: '3',
        deliveryTime: 7,
      },
      {
        product: 'Product 4',
        price: 20,
        quantity: 50,
        rating: '4',
        deliveryTime: 2,
      },
      {
        product: 'Product 5',
        price: 1.99,
        quantity: 700,
        rating: '5',
        deliveryTime: 1,
      },
      {
        product: 'Product 6',
        price: 499,
        quantity: 30,
        rating: '5',
        deliveryTime: 20,
      },
      {
        product: 'Product 7',
        price: 13.99,
        quantity: 130,
        rating: '1',
        deliveryTime: 30,
      },
      {
        product: 'Product 8',
        price: 5.99,
        quantity: 500,
        rating: '4',
        deliveryTime: 5,
      },
      {
        product: 'Product 9',
        price: 12,
        quantity: 1200,
        rating: '1',
        deliveryTime: 25,
      },
      {
        product: 'Product 10',
        price: 2.99,
        quantity: 200,
        rating: '3',
        deliveryTime: 17,
      },
    ],
    [],
  );

    () => [
      {
        id: 'product',
        Header: 'Product',
        accessor: 'product',
        minWidth: 400,
      },
      {
        id: 'price',
        Header: 'Price',
        accessor: 'price',
        width: 400,
        Cell: (props: CellProps<(typeof data)[0]>) => {
          return <>{`$${props.value}`}</>;
        },
      },
      {
        id: 'quantity',
        Header: 'Quantity',
        accessor: 'quantity',
        width: 400,
      },
      {
        id: 'rating',
        Header: 'Rating',
        accessor: 'rating',
        width: 400,
        Filter: tableFilters.NumberRangeFilter(),
        filter: 'between',
        sortType: 'number',
      },
      {
        id: 'deliveryTime',
        Header: 'Delivery Time',
        accessor: 'deliveryTime',
        width: 400,
        Cell: (props: CellProps<(typeof data)[0]>) => {
          return <>{`${props.value} day(s)`}</>;
        },
      },
    ],
    [],
  ) satisfies Column<(typeof data)[0]>[];

  return (
    <Table
      columns={columns}
      data={data}
      emptyTableContent='No data.'
      style={{ maxWidth: '1300px' }}
      emptyFilteredTableContent='No data matches the filter criteria.'
    />
  );
};

export default App;

</details>

## Docs

Added a changeset

### DIFF
--- a/.changeset/happy-chicken-tie.md
+++ b/.changeset/happy-chicken-tie.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Table is now scrollable horizontally even when no rows are present.
--- a/packages/itwinui-css/src/table/base.scss
+++ b/packages/itwinui-css/src/table/base.scss
@@ -359,6 +359,9 @@
 
   // Empty & loading states
   > .iui-table-empty {
+    position: sticky;
+    inset-inline-start: 50%;
+    transform: translateX(-50%);
     text-align: center;
     padding: var(--iui-size-xl);
     display: flex;
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -873,6 +873,21 @@ export const Table = <
 
   const isHeaderDirectClick = React.useRef(false);
 
+  // If host already hosts a shadow tree, catch the exception and do not re-add the shadow tree
+  try {
+    const host = document.querySelector('#host');
+    const shadow = host?.attachShadow({ mode: 'closed' });
+
+    const div = document.createElement('div');
+    div.style.height = '0';
+    div.style.width = `${headerRef.current?.scrollWidth}px`;
+    div.ariaHidden = 'true';
+
+    shadow?.appendChild(div);
+  } catch (e) {
+    // Do nothing
+  }
+
   return (
     <>
       <Box
@@ -1059,6 +1074,7 @@ export const Table = <
             (isSelectable && selectionMode === 'multi') || undefined
           }
         >
+          <div id='host' />
           {data.length !== 0 && (
             <>
               {enableVirtualization ? (
@@ -1085,7 +1101,14 @@ export const Table = <
             </Box>
           )}
           {isLoading && data.length !== 0 && (
-            <Box className='iui-table-row'>
+            <Box
+              className='iui-table-row'
+              // To center the row even when scrolling horizontally
+              style={{
+                position: 'sticky',
+                left: 0,
+              }}
+            >
               <Box
                 className='iui-table-cell'
                 style={{ justifyContent: 'center' }}
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -1076,10 +1076,7 @@ export const Table = <
             (isSelectable && selectionMode === 'multi') || undefined
           }
         >
-          <div
-            ref={tableBodyWidthSetterRef}
-            className='iui-table-body-width-setter'
-          />
+          <div ref={tableBodyWidthSetterRef} />
           {data.length !== 0 && (
             <>
               {enableVirtualization ? (
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -873,9 +873,11 @@ export const Table = <
 
   const isHeaderDirectClick = React.useRef(false);
 
+  const tableBodyWidthSetterRef = React.useRef<HTMLDivElement>(null);
+
   // If host already hosts a shadow tree, catch the exception and do not re-add the shadow tree
   try {
-    const host = document.querySelector('#host');
+    const host = tableBodyWidthSetterRef.current;
     const shadow = host?.attachShadow({ mode: 'closed' });
 
     const div = document.createElement('div');
@@ -1074,7 +1076,10 @@ export const Table = <
             (isSelectable && selectionMode === 'multi') || undefined
           }
         >
-          <div id='host' />
+          <div
+            ref={tableBodyWidthSetterRef}
+            className='iui-table-body-width-setter'
+          />
           {data.length !== 0 && (
             <>
               {enableVirtualization ? (


packages/itwinui-css/src/table/base.scss
mayank99, line 1: if we're changing the CSS to use sticky positioning anyway, why don't we just set the width on the table-body instead of a hidden div inside?

rohan-kadkol: I had tried adding the width to the table-body but noticed that it broke the UI:

image

https://github.com/iTwin/iTwinUI/assets/45748283/c8176ddd-dc7d-4e98-915c-002782131ea4

I believe this is because setting the width on the table-body would affect the `offsetWidth` but not the `scrollWidth`.

Since the `scrollWidth` is a read-only property, I could not find any workaround to set the `scrollWidth` of the table-body programmatically. Thus, I thought to use the dummy div approach to set the `scrollWidth` of the table-body.

packages/itwinui-css/src/table/base.scss
```diff
@@ -224,7 +224,7 @@
 
     // #region Selection
     &:not([aria-selected='true']) + [aria-selected='true'] .iui-table-cell,
-    &[aria-selected='true']:first-child .iui-table-cell {
+    &[aria-selected='true']:nth-child(2) .iui-table-cell {
```
rohan-kadkol, line 227: Is this change alright? I had to do it since with the new shadow root div inside table-body, the first selectable element in table body is now the second child.

Without this change, these test images were failing: https://github.com/iTwin/iTwinUI/pull/1725/commits/f6ea60fda7923d4da9336addbb74f64087f5c3ba

mayank99: no, this is because you're not using shadow dom correctly.

the shadow-root should be attached to the table-body element, so that these selectors continue to work.

after attaching the shadow tree, you can add two elements to it:
1. your dummy `<div />` for setting the width
2. a [`<slot />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) for showing the rows that are its "real" children

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1085,7 +1103,14 @@ export const Table = <
             </Box>
           )}
           {isLoading && data.length !== 0 && (
-            <Box className='iui-table-row'>
+            <Box
+              className='iui-table-row'
+              // To center the row even when scrolling horizontally
+              style={{
+                position: 'sticky',
```
mayank99, line 1110: what specific case does this handle?

inline styles should generally be avoided, so i'm trying to think of ways we can move this into the CSS.

rohan-kadkol: I did this to handle the case where the table is loading when there is pre-existing data.

I thought of adding these styles inline because I noticed that it was also done on [L1116](https://github.com/iTwin/iTwinUI/pull/1725/files/f1aadc6ad7a4578ac17f3ac95cc9ace86224eb4e#diff-ef4196b6638c028042adefbb785a48b755d08db5caa109a8619fbfde44e40669R1116). But I can try using a CSS selector if that is preferrable.

<details>

Before inline styles:

https://github.com/iTwin/iTwinUI/assets/45748283/8c0ecf58-34e2-40ff-8cc9-cf2922e650b8

After inline styles:

https://github.com/iTwin/iTwinUI/assets/45748283/0dc4df61-22f6-410e-bf94-de6f5dbd0b9c

</details>

mayank99: i think we should make a new selector `.iui-table-row-loading` or `.iui-table-row[data-iui-loading="true"]` that adds handles both sticky positioning and centering

rohan-kadkol: Sounds good, removed the two inlines in favor of `.iui-table-row[data-iui-loading="true"]`. (https://github.com/iTwin/iTwinUI/pull/1725/commits/29d1cb51469c267a866c28b9ecdb999802487409 and https://github.com/iTwin/iTwinUI/pull/1725/commits/b4403c0da9a216bb5b2b2e5218ade7f12ded6440)

packages/itwinui-css/src/table/base.scss
rohan-kadkol: Ohh, I see. I now attached the shadow-root to the table-body and used `<slot />` to make table-body's children visible even when the shadow-root is attached.

As a result, I also reverted the CSS change from this PR comment. (https://github.com/iTwin/iTwinUI/pull/1725/commits/98bec1d236b93b7cc4e2e91bc2b2f142bbe8a695)

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1128,3 +1137,34 @@ export const Table = <
...
 
 export default Table;
+
+// ----------------------------------------------------------------------------
+
+/**
+ * Wrapper around `<template>` element that attaches shadow root to its parent
+ * and moves its children into the shadow root.
+ */
+const ShadowTemplate = ({ children }: { children: React.ReactNode }) => {
```
mayank99, line 1147: i noticed two things:
1. empty `<template>` tag is left behind in the DOM
2. state updates from inside the template are not correctly reflected.

i was able to fix these using `template.remove()` and `createPortal()`

try these two snippets in the playground to see the difference:


<details><summary>Before</summary>

```tsx

  const [count, setCount] = React.useState(0);

  return (
    <>
      <div>
        <ShadowTemplate>
          <button onClick={() => setCount(count + 1)}>{count}</button>
        </ShadowTemplate>
      </div>
    </>
  );
}

const ShadowTemplate = ({ children }: { children: React.ReactNode }) => {
  const [root, setRoot] = React.useState<HTMLDivElement | null>(null);

  const attachShadowRef = React.useCallback(
    (template: HTMLTemplateElement | null) => {
      const parent = template?.parentElement;
      if (!template || !root || parent?.shadowRoot) {
        return;
      }

      const shadowRoot = parent?.attachShadow({ mode: 'open' });
      shadowRoot?.appendChild(root);
    },
    [root],
  );

  return (
    <template ref={attachShadowRef}>
      <div ref={setRoot} style={{ display: 'contents' }}>
        {children}
      </div>
    </template>
  );
};

```

</details> 

<details><summary>After</summary>

```tsx

  const [count, setCount] = React.useState(0);

  return (
    <>
      <div>
        <ShadowTemplate>
          <button onClick={() => setCount(count + 1)}>{count}</button>
        </ShadowTemplate>
      </div>
    </>
  );
}

const ShadowTemplate = ({ children }: { children: React.ReactNode }) => {
  const [root, setRoot] = React.useState<HTMLDivElement | null>(null);

  const attachShadowRef = React.useCallback(
    (template: HTMLTemplateElement | null) => {
      const parent = template?.parentElement;
      if (!template || !root || parent?.shadowRoot) {
        return;
      }

      const shadowRoot = parent?.attachShadow({ mode: 'open' });
      shadowRoot?.appendChild(root);

      template.remove();
    },
    [root],
  );

  return (
    <template ref={attachShadowRef}>
      <div ref={setRoot} style={{ display: 'contents' }}>
        {root && ReactDOM.createPortal(children, root)}
      </div>
    </template>
  );
};

```

</details>

rohan-kadkol: Ohh, I now understand why portal was needed, thanks! Added `template.remove()` & `createPortal` (https://github.com/iTwin/iTwinUI/pull/1725/commits/527ba95058f9228641a3208f636f41d57635b2f4)

rohan-kadkol: Ohh, sounds good, removed (https://github.com/iTwin/iTwinUI/pull/1725/commits/d722e44498297a227345ff2aac49caf76922cd67)

packages/itwinui-css/src/table/base.scss
```diff
@@ -371,10 +371,15 @@
...
   > .iui-table-empty {
+    text-align: center;
     position: sticky;
     inset-inline-start: 50%;
     transform: translateX(-50%);
     padding: var(--iui-size-xl);
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    flex-grow: 1;
```
mayank99, line 382: required for what

rohan-kadkol: `test-align: center` was needed to fix this failing css-workshop test:
image

And the flex related attributes were required to fix this failing css-workshop test:
image

.changeset/happy-chicken-tie.md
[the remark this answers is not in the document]
rohan-kadkol: Sounds good, added (https://github.com/iTwin/iTwinUI/pull/1725/files/035f9b9f65e4693e565eef0179c1b07250756925..3beb6e22f9af304874a3f4b1ada528b465e9bb83)

mayank99 (approved):
LGTM

PR #1731, opened 2023-12-08
dispatch theme changes via custom events

mayank99:
## Changes

**Context** (no pun intended): In #1478, we made `theme="inherit"` more resilient, by reading the `data-iui-theme` attribute when react Context fails (for example, v2 Context cannot be consumed in v3). However this only works on first load. Subsequent theme changes are not reflected.

**This PR**: The ThemeProvider root element will now dispatch [native custom events](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) which can be subscribed to anywhere in the tree regardless of Context or package version. Specifically, `iui:themechange` events will be dispatched every time the `theme` prop changes. Then, for the `theme="inherit"` fallback, we will listen to these events and update the theme each time.

**Subsequent PR**: The same change will need to be made in v2, so that v3 can listen to theme changes from v2.

## Testing

A little difficult to test locally without making a release, but it can be emulated in vite playground by manually dispatching events (pretend these events are dispatched by v2 ThemeProvider).

<details><summary>main.tsx</summary>

```tsx
import { createRoot } from 'react-dom/client';
import styled from '@emotion/styled';
import App from './App.tsx';
import { SvgMoon, SvgSun } from '@itwin/itwinui-icons-react';

const Shell = () => {
  const rootRef = React.useRef<HTMLDivElement>(null);
  const [theme, setTheme] = React.useState<'light' | 'dark'>(() =>
    matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
  );
  const [highContrast, setHighContrast] = React.useState(false);

  return (
    <>
      <div data-iui-theme={theme} ref={rootRef}>
        <Main>
          <ThemeToggle
            aria-label='Toggle theme'
            onClick={() => {
              const newThemeValue = theme === 'dark' ? 'light' : 'dark';
              setTheme(newThemeValue);
              rootRef.current?.dispatchEvent(
                new CustomEvent('iui:themechange', {
                  detail: { theme: newThemeValue, highContrast },
                }),
              );
            }}
          >
            {theme === 'dark' ? <SvgMoon /> : <SvgSun />}
          </ThemeToggle>

          <HighContrastToggle
            onClick={() => {
              const newHcValue = !highContrast;
              setHighContrast(newHcValue);
              rootRef.current?.dispatchEvent(
                new CustomEvent('iui:themechange', {
                  detail: { theme, highContrast: newHcValue },
                }),
              );
            }}
          >
            {highContrast ? 'High Contrast' : 'Regular contrast'}
          </HighContrastToggle>

          <ThemeProvider theme='inherit'>
            <App />
          </ThemeProvider>
        </Main>
      </div>
    </>
  );
};

const Main = styled.main`
  padding: 2rem 1rem;
  height: 100vh;
`;

const ThemeToggle = styled.button`
  all: unset;
  display: inline-grid;
  place-items: center;
  position: fixed;
  font-family: system-ui;
  user-select: none;
  top: 0;
  right: 0;
  min-width: 2.5rem;
  height: 2.5rem;
  cursor: pointer;
  border-radius: 50%;

  &:hover {
    background: hsl(0 0% 0% / 0.2);
  }

  & > * {
    width: 1.5rem;
    height: 1.5rem;
    fill: currentColor;
  }
`;

const HighContrastToggle = styled(ThemeToggle)`
  right: 2.5rem;
`;

createRoot(document.getElementById('root')!).render(
  <React.StrictMode>
    <Shell />
  </React.StrictMode>,
);
```

</details> 

Added unit tests as well.

After this PR is merged, I will backport it to v2, make a release, then test with actual v2 ThemeProvider.

## Docs

N/A


packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -211,29 +232,58 @@ const Root = React.forwardRef((props, forwardedRef) => {
...
+  const parentThemeRef = useLatestRef(parentContext?.theme);
 
   useIsomorphicLayoutEffect(() => {
-    setParentTheme(
-      (old) =>
-        old ||
-        (rootRef.current?.parentElement
-          ?.closest('[data-iui-theme]')
-          ?.getAttribute('data-iui-theme') as ThemeType),
+    // bail if we already have theme from context
```
rohan-kadkol, line 252: nit: clarity

```suggestion
    // bail if we already have theme (& highContrast) from context
```

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -211,29 +232,58 @@ const Root = React.forwardRef((props, forwardedRef) => {
...
+    if (parentThemeRef.current) {
+      return;
+    }
+
+    // find parent theme from closest data-iui-theme attribute
+    const closestRoot = rootElement?.parentElement?.closest('[data-iui-theme]');
+
+    if (!closestRoot) {
```
rohan-kadkol, line 260: nit: Is any comment needed here?

```suggestion
    // If no closestRoot exists, then there is nothing to listen to
    if (!closestRoot) {
```

rohan-kadkol (commented):
I left a few nit comments in `ThemeProvider.tsx`. They were mainly for me to better understand the code. So, feel free to ignore those suggestions if the code is self-explanatory as-is and also maybe also correct me if I misunderstood what the highlighted code does.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -228,3 +228,75 @@ it('should inherit the portalContainer if inheriting theme', async () => {
...
+      ref={(el) => {
+        el?.dispatchEvent(
+          new CustomEvent('iui:themechange', { detail: props }),
+        );
+      }}
+    >
+      <ThemeProvider theme='inherit' data-root>
+        hello
+      </ThemeProvider>
+    </div>
```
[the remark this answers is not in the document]
mayank99: yes

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -159,7 +176,11 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
       <Root
         theme={theme}
         themeOptions={themeOptions}
-        ref={useMergedRefs(forwardedRef, rootRef)}
+        ref={useMergedRefs(
+          forwardedRef,
+          setRootElement,
+          dispatchThemeChangesRef,
+        )}
```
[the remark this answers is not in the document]
mayank99: it will be called when the useCallback dependencies change

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: not necessary, the code is self explanatory

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: this is wrong, we don't care about highContrast and are not even checking for it

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -211,29 +232,58 @@ const Root = React.forwardRef((props, forwardedRef) => {
...
   const [parentThemeState, setParentTheme] = React.useState(
     parentContext?.theme,
   );
+  const [parentHighContrastState, setParentHighContrastState] = React.useState(
+    parentContext?.themeOptions?.highContrast,
+  );
+
+  const parentThemeRef = useLatestRef(parentContext?.theme);
 
   useIsomorphicLayoutEffect(() => {
```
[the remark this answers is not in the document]
rohan-kadkol: nit: Yes, that is also sufficient. But when reading the code from top to bottom, it would be helpful if the useEffect has a comment with its *overarching* purpose. That gives some context to the future readers of the code when they scroll down and see the contents in the useEffect. That context is helpful for me, at least.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
rohan-kadkol: To confirm, is this because if the theme is there from the context, then the highContrast would also be there? So just checking for theme is enough?

rohan-kadkol (approved):

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: yes and no. `theme` is definitely the important part, but `themeOptions.highContrast` can be undefined even when `theme` is present.

raplemie:
@mayank99, since this will require the v2 package to be up to date for the event to be fired, and you need a ref to the element anyway, would `MutationObserver` allow the same thing but would not require the new event ? (By looking at mutation on the ref attributes, which I think contain all the information needed (theme and high contrast))

mayank99:
> @mayank99, since this will require the v2 package to be up to date for the event to be fired, and you need a ref to the element anyway, would `MutationObserver` allow the same thing but would not require the new event ? (By looking at mutation on the ref attributes, which I think contain all the information needed (theme and high contrast))

@raplemie i find mutation observers to be too unwiedly. events feel more natural because of their "dispatch"/"subscribe" model. however, considering how bad applications are at consuming the latest version (many of them are still on 2.11.x), it is the better option overall. i will create another PR, thanks for the suggestion.

PR #1733, opened 2023-12-11
fix tabs overflow layout shift

### LINKED ISSUE #1627: TabsPanel changes width upon first change of the tab
> May not be related to this PR, but just noticed some shift in the width of the tabs panel in the "Vertical Overflow" story
>
> [14.webm](https://github.com/iTwin/iTwinUI/assets/45748283/2bb1bd2a-facc-408b-bb07-a4b5e5ffbda0)

_Originally posted by @rohan-kadkol in https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1355274285_

> I am on a Windows 11 + Edge 117. What I did was:
> * Press tab to focus the selected "Item 11`"
> * Press up or down

_Originally posted by @rohan-kadkol in https://github.com/iTwin/iTwinUI/pull/1548#discussion_r1355294539_

mayank99:
## Changes

fixes #1627

after investigating for a few hours, i was able to fix the layout shift by conditionally applying `scrollbar-gutter` on mount. i also fixed the css to use grid layout (no `white-space: nowrap` hacks).

unrelated: fixed all ids to convert spaces to hyphens (spaces are not allowed in ids). this also required changing tsconfig `lib` because `replaceAll` is a relatively new function.

## Testing

Tested using vite playground in all three browsers.

https://github.com/iTwin/iTwinUI/assets/9084735/a9ac91e0-76f1-4f94-a873-7545df3f0d9f

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Tabs/Tabs.tsx
+++ b/packages/itwinui-react/src/core/Tabs/Tabs.tsx
@@ -150,7 +150,12 @@ const TabList = React.forwardRef((props, ref) => {
   const isClient = useIsClient();
   const tablistRef = React.useRef<HTMLDivElement>(null);
   const [tablistSizeRef, tabsWidth] = useContainerWidth(type !== 'default');
-  const refs = useMergedRefs(ref, tablistRef, tablistSizeRef);
+  const refs = useMergedRefs(
+    ref,
+    tablistRef,
+    tablistSizeRef,
+    useScrollbarGutter(),
+  );
 
   return (
     <Box
@@ -325,10 +330,10 @@ const Tab = React.forwardRef((props, forwardedRef) => {
       role='tab'
       tabIndex={isActive ? 0 : -1}
       aria-selected={isActive}
-      aria-controls={`${idPrefix}-panel-${value}`}
+      aria-controls={`${idPrefix}-panel-${value.replaceAll(' ', '-')}`}
       ref={useMergedRefs(tabRef, forwardedRef, setInitialActiveRef)}
       {...rest}
-      id={`${idPrefix}-tab-${value}`}
+      id={`${idPrefix}-tab-${value.replaceAll(' ', '-')}`}
       onClick={mergeEventHandlers(props.onClick, () => setActiveValue(value))}
       onKeyDown={mergeEventHandlers(props.onKeyDown, onKeyDown)}
       onFocus={mergeEventHandlers(props.onFocus, () => {
@@ -438,12 +443,12 @@ const TabsPanel = React.forwardRef((props, ref) => {
   return (
     <Box
       className={cx('iui-tabs-content', className)}
-      aria-labelledby={`${idPrefix}-tab-${value}`}
+      aria-labelledby={`${idPrefix}-tab-${value.replaceAll(' ', '-')}`}
       role='tabpanel'
       hidden={activeValue !== value ? true : undefined}
       ref={ref}
       {...rest}
-      id={`${idPrefix}-panel-${value}`}
+      id={`${idPrefix}-panel-${value.replaceAll(' ', '-')}`}
     >
       {children}
     </Box>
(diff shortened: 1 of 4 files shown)


rohan-kadkol (commented):
When testing this PR with the Vite playground code from this morning (https://github.com/iTwin/iTwinUI/commit/252c230d07cc6790a58ab8b4edd2e980140bed31), I noticed some change in the height of the tabs when orientation = vertical

|Earlier|Now|
|-|-|
|image|Screenshot 2023-12-11 at 2 50 57 PM|

mayank99:
> When testing this PR with the Vite playground code from this morning ([252c230](https://github.com/iTwin/iTwinUI/commit/252c230d07cc6790a58ab8b4edd2e980140bed31)), I noticed some change in the height of the tabs when orientation = vertical and there's no overflow.

fixed in https://github.com/iTwin/iTwinUI/pull/1733/commits/1545b175abe2ad30378b4aa2b34f790db2e795bc

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -325,10 +330,10 @@ const Tab = React.forwardRef((props, forwardedRef) => {
       role='tab'
       tabIndex={isActive ? 0 : -1}
       aria-selected={isActive}
-      aria-controls={`${idPrefix}-panel-${value}`}
+      aria-controls={`${idPrefix}-panel-${value.replaceAll(' ', '-')}`}
       ref={useMergedRefs(tabRef, forwardedRef, setInitialActiveRef)}
       {...rest}
-      id={`${idPrefix}-tab-${value}`}
+      id={`${idPrefix}-tab-${value.replaceAll(' ', '-')}`}
```
rohan-kadkol, line 336: Was this change necessary? Or was it addressing something else?

Also, might be a contrived example, but what if users have two tab items with values "Item 1" and "Item-1". Will there be two `iui-tab`s with the same id?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
```diff
@@ -797,3 +802,26 @@ const TabListContext = React.createContext<
...
+
+// ----------------------------------------------------------------------------
+
+/**
+ * This conditionally adds `scrollbar-gutter: stable` only if the content overflows.
+ * This is a workaround to prevent layout shift that happens because of scrollbar width.
+ *
+ * @see https://github.com/iTwin/iTwinUI/issues/1627
+ */
+const useScrollbarGutter = () => {
```
rohan-kadkol, line 814: Is it called only once? Or does it listen to height changes of the parent?

If the parent become larger, should we listen for changes and remove the `scrollbar-gutter` when there is no longer an overflow?


https://github.com/iTwin/iTwinUI/assets/45748283/4cb71079-73ca-4926-9d73-87be95d1262e

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: It's unrelated. Spaces are not valid in `id`, and `aria-controls` can accept multiple space-separated ids. I noticed this in our stories when debugging the main issue.

I can't think of a better way to address this, so I think it's fine. We do the same thing in ComboBox iirc

packages/itwinui-react/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "target": "es2020",
     "module": "esnext",
-    "lib": ["es6", "dom"],
+    "lib": ["esnext", "dom"],
```
rohan-kadkol, line 6: Was this a necessary change?

packages/itwinui-react/src/core/Tabs/Tabs.tsx
mayank99: yes it's only called once. I don't think it's worth the extra code for listening to resizes.

the main issue is the layout shift between initial render and subsequent clicks. if the scrollbar-gutter stays after that, it's no big deal

rohan-kadkol (approved):

PR #1734, opened 2023-12-12
automatically load css at runtime as fallback

mayank99:
## Changes

1. In `styles.css`, we will set a variable `--_iui-v3-loaded`. At runtime, we check for this variable and do nothing else if found.
2. If variable not found, that means styles are not loaded. We will first attempt to use a side-effect import to dynamically import `styles.css`. Many bundlers know how to resolve this, so we should be good.
3. If this doesn't work, we'll fall back to loading `styles.css` from a CDN.

All of this only happens when using `theme="inherit"`. The `includeCss` prop can be used to control this behavior for other themes as well.

## Testing

Tested thoroughly across different bundlers/frameworks.

- Vite and Astro works ok.
- Nextjs complains about [global CSS within components](https://nextjs.org/docs/messages/css-global), even though the code is not even called. Can be worked around using [`transpilePackages`](https://nextjs.org/docs/architecture/nextjs-compiler#module-transpilation).
- Released `3.2.0-dev.0` for external testing.
  - Tested in sandboxes:
    - ✅ [vite](https://stackblitz.com/edit/github-76vvyu?file=src%2Fmain.tsx,package.json)
    - ✅ [webpack](https://stackblitz.com/edit/github-76vvyu-xdouxb?file=package.json,src%2Fmain.tsx)
    - ✅ [nextjs](https://stackblitz.com/edit/nextjs-bc8kz9?file=app%2Flayout.tsx,package.json)
    - ✅ [cra](https://stackblitz.com/edit/github-76vvyu-3zdote?file=package.json,src%2Findex.js)
    - ✅ [esbuild](https://stackblitz.com/edit/github-76vvyu-ttphsr?file=package.json,src%2Fmain.jsx,index.html)
    - ❌ [remix](https://stackblitz.com/edit/remix-run-remix-edmvcy?file=app%2Froot.tsx,app%2Froutes%2F_index.tsx)
  - Tested the "v2 consuming v3" scenario externally: https://github.com/iTwin/admin-components-react/pull/113#discussion_r1427324700

## Docs

Added jsdocs and changeset.


packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -84,6 +86,16 @@ type ThemeProviderOwnProps = Pick<RootProps, 'theme'> & {
...
+  /**
+   * This prop will be used to determine if `styles.css` should be automatically imported at runtime.
+   *
+   * When using `theme='inherit'`, styles will be automatically loaded if not already found.
+   * This default behavior is useful for packages that want to support incremental adoption of latest iTwinUI,
+   * without requiring consuming applications (that might still be using an older version) to manually import the CSS.
+   *
+   * If true or false is passed, it will override the default behavior.
+   */
+  includeCss?: boolean;
```
rohan-kadkol, line 98: nit: Clarity

```suggestion
  /**
   * This prop will be used to determine if `styles.css` should be automatically imported at runtime.
   *
   * By default, this is set to `true` if `theme='inherit'`. If `true`, styles will automatically be loaded, if not already found.
   * This default behavior is useful for packages that want to support incremental adoption of latest iTwinUI,
   * without requiring consuming applications (that might still be using an older version) to manually import the CSS.
   *
   * Passing `true` or `false` will override the default behavior.
   * 
   * @default
   * theme='inherit'
   */
  includeCss?: boolean;
```

packages/itwinui-react/src/styles.js/styles.module.css
```diff
@@ -7,3 +7,9 @@
 
 @import '@itwin/itwinui-variables' layer(itwinui.v3);
 @import '@itwin/itwinui-css' layer(itwinui.v3);
+
+@layer itwinui.v3 {
+  .iui-root {
+    --_iui-v3-loaded: yes;
```
rohan-kadkol, line 13: Any advantage of choosing `yes` instead of something like `true`?

mayank99: i deliberately didn't use "true" because this is a string and i don't want to confuse it with the Javascript value `true`

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: How about this?

I don't like specifying `@default` here since that is intended for simple values.

```suggestion
  /**
   * This prop will be used to determine if `styles.css` should be automatically imported at runtime (if not already found).
   *
   * By default, this is enabled when using `theme='inherit'`.
   * This default behavior is useful for packages that want to support incremental adoption of latest iTwinUI,
   * without requiring consuming applications (that might still be using an older version) to manually import the CSS.
   *
   * If true or false is passed, it will override the default behavior.
   */
  includeCss?: boolean;
```

rohan-kadkol: Ohh, yes, your concise version sounds good 👍

mayank99: pushed

rohan-kadkol (approved):

mayank99 (commented):
i've retested this PR and #1744 via `3.2.0-dev.0`, works fine everywhere except remix. going to merge and release this now 🚀

PR #1736, opened 2023-12-13
improve tabs documentation

mayank99:
## Changes

added info about:
- `key` for resetting state
- `focusActivationMode` prop.

## Testing

n/a

## Docs

n/a


apps/website/src/content/docs/tabs.mdx
```diff
@@ -37,9 +37,9 @@ You can build Tabs with subcomponents that are fully customizable.
 </Tabs.Wrapper>
 ```
 
-`Tabs.Tab` and `Tabs.Panel` can be associated with each other by passing them the same `value`.
+`Tabs.Tab` and `Tabs.Panel` can be associated with each other by passing them the same `value`. `Tabs.Panel` will be shown when its `value` matches the `value` of the currently active `Tabs.Tab`.
 
-`Tabs.Tab` has `label` prop for ease of use. You can also build `Tabs.Tab` with subcomponents and add sublabel and icon to your tab.
+`Tabs.Tab` has a `label` prop for ease of use. You can also build `Tabs.Tab` with subcomponents and add sublabel and icon to your tab.
```
rohan-kadkol, line 42: nit: grammar
```suggestion
`Tabs.Tab` has a `label` prop for ease of use. You can also build `Tabs.Tab` with subcomponents and add a sublabel or an icon to your tab.
```

apps/website/src/content/docs/tabs.mdx
```diff
@@ -59,7 +59,7 @@ You can build Tabs with subcomponents that are fully customizable.
 
 Alternatively, you can use Tabs directly with labels, actions, and children as panels. Although this pattern can suffice for simple cases, we strongly recommend using composition API for the best experience.
 
-The following code is equivalent to the [composition example](#composition-api) above:
+The following code is (mostly) equivalent to the [composition example](#composition-api) above:
```
rohan-kadkol, line 62: nit: Would making "(mostly)" italic be better?

```suggestion
The following code is _(mostly)_ equivalent to the [composition example](#composition-api) above:
```

rohan-kadkol (approved):

apps/website/src/content/docs/tabs.mdx
mayank99: italics for emphasis, i don't think we want to emphasize it here

PR #1738, opened 2023-12-14
Added styles.css to sideEffects

### LINKED ISSUE #1737: Missing styles.css in package.json `sideEffects`
I'm using webpack with ESM. When importing styles.css it gets treeshaken-out.
And styles are missing completely in output

Possible webpack bug: https://github.com/webpack/webpack/issues/7094

#### Expected behavior
importing styles.css just works

#### Reproduction
Link to a minimal repro: 
Bug reproduction
https://stackblitz.com/edit/webpack-5-react-starter-zvm3dw?file=src%2Fapp.tsx,package.json

Patched version with `"sideEffects": ["styles.css"]`
https://stackblitz.com/edit/webpack-5-react-starter-m3a2tr?file=src%2Fapp.tsx,package.json

##### Steps to reproduce
1. Use webpack bundler with ESM
2. Import iTwinUI styles.css
3. no css in output

#### Additional information
Possible solution:
Add package.json sideEffects for styles.css file
```json
{
    sideEffects: ["styles.css"]
}
```

Woyken:
This fix is intended for webpack users.
When using ESM, webpack will apply it's treeshaking too harshly sometimes.
This applies to imported css files too.
Possible webpack bug: https://github.com/webpack/webpack/issues/7094

Fixes #1737 

## Changes

Added `styles.css` to `sideEffects` in `package.json`

## Testing

"N/A"

## Docs

"N/A"

### DIFF
--- a/.changeset/plenty-icons-cry.md
+++ b/.changeset/plenty-icons-cry.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added missing `sideEffects` to prevent tree-shaking `styles.css` in webpack.
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -133,5 +133,7 @@
     ]
   },
   "prettier": "configs/prettier-config",
-  "sideEffects": false
+  "sideEffects": [
+    "styles.css"
+  ]
 }


mayank99 (approved):
I've always found `sideEffects` to be confusing. Using this field in the past, we've had problems with some bundlers including *all* .css files even if their components are not actually used. But now that we only have a single stylesheet, that should be fine I guess.

I also thought `"sideEffects": false` would be the correct move, since we do not import it within our components, but apparently not.

My only remaining concern is regarding packages that depend on `itwinui-react`. If they use a bundler, it might decide to include `styles.css` even if it's not imported. But packages really should not be using bundlers in the first place, so maybe it's not a big concern.

Lets go with this and see if anybody complains. 👍

PR #1742, opened 2023-12-15
fix dynamic imports in CJS and jest

mayank99:
## Changes

Extracted out of #1734. The `dynamicImport` wrapper has been removed, as the hacks are no longer necessary:
- dynamic `import()` works fine with Jest in Node 20, so i've updated Node version in CI.
  - this required using the `experimental-vm-modules` flag (see [docs](https://jestjs.io/docs/ecmascript-modules))
- dynamic `import()` is correctly preserved by typescript in CJS when using `moduleResolution: nodenext`.
  - this also required updating some imports for `react-transition-group` and `@testing-library/user-event` (not user facing)

## Testing

Everything works correctly now.

## Docs

N/A


.changeset/small-bottles-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed `ErrorPage` to correctly lazy import from `@itwin/itwinui-illustrations-react`.
```
mayank99, line 5: this one is interesting — we tested it in #1400, but it broke sometime after that (presumably because of #1452).

bundlers generally prefer simple/basic uses of dynamic imports, i.e. the actual string being passed to `import()`, without any variables or wrapper functions, so it is fixed now.

packages/itwinui-react/src/core/utils/components/WithCSSTransition.tsx
```diff
@@ -24,21 +23,21 @@ export const WithCSSTransition = (
...
+      onEntered={(node: HTMLElement) => {
         node.style[`min${dimensionCamelCase}`] = '';
         node.style[dimension] = '';
       }}
-      onExit={(node) => {
+      onExit={(node: HTMLElement) => {
         node.style[dimension] = `${expandedSize.current}px`;
       }}
-      onExiting={(node) => {
+      onExiting={(node: HTMLElement) => {
```
rohan-kadkol, line 40: nit: Types for `node` in `onExit` and `onExisting` does not seem to be needed. But type does not seem to be inferred in the other three callbacks. So, if its needed in three places, might also be alright to define the type in the remaining two places too, like you have it.

packages/itwinui-react/package.json
```diff
@@ -63,8 +63,8 @@
     "clean:coverage": "rimraf coverage",
     "clean": "rimraf .turbo && yarn clean:coverage && yarn clean:build && rimraf node_modules",
     "test": "yarn test:types && yarn test:unit",
-    "test:unit": "jest",
-    "test:unit:watch": "jest --watch",
+    "test:unit": "dotenv -v NODE_OPTIONS=--experimental-vm-modules -- npx jest",
```
rohan-kadkol, line 66: Since `dotenv` is used in `react-workshop` and `itwinui-react`, should its dependency be defined in the root `package.json`?

packages/itwinui-react/src/core/utils/components/WithCSSTransition.tsx
mayank99: yeah lets leave it this way for consistency i guess

rohan-kadkol (approved):

.changeset/small-bottles-cheer.md
rohan-kadkol: To confirm, should we have/reintroduce tests for deprecated components like `ErrorPage` to catch regresses like these?

mayank99: we do have tests, but since we were previously using hacks for jest (which didn't support dynamic `import()`) we couldn't catch them

packages/itwinui-react/src/core/NonIdealState/ErrorPage.tsx
mayank99, line 1: something i noticed as a result of this change is that webpack wants to bundle illustrations even if they're not being used :(

mayank99: i think we'll have to move `@itwin/itwinui-illustrations-react` from peer deps to direct deps because of this. it's really unfortunate

PR #1744, opened 2023-12-18
add direct dependency on illustrations

mayank99:
## Changes

fixes https://github.com/iTwin/iTwinUI/pull/1742#discussion_r1430493256 which happens because `ErrorPage` is part of the barrel file (`index.js`), resulting in bundlers trying to resolve everything even if it's not used (it later gets tree-shaked)

first i tried `catch`ing these errors ([`a6c4039`](https://github.com/iTwin/iTwinUI/pull/1744/commits/a6c40392023184eeb05b10dbca013f5de4c56812)), but vite and nextjs don't respect this `catch`. so i've resorted to moving it from a peer dependency to direct dependency.

this effectively undoes the breaking change, so users have one less thing to worry about. and if `ErrorPage` is not used, this *should not* be part of the final bundle, so no real downside there.

## Testing

no more errors! i've done preliminary testing locally, but I plan to test in sandboxes some more, together with #1734 after this PR is merged.

## Docs

added changeset. will also update migration guide after this merges.

### DIFF
--- a/.changeset/gentle-worms-dance.md
+++ b/.changeset/gentle-worms-dance.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`@itwin/itwinui-illustrations-react` has been made a direct dependency again, to avoid issues with bundlers attempting to bundle it even if `ErrorPage` is not used anywhere.
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -114,14 +114,8 @@
     "vite": "^4.5.1"
   },
   "peerDependencies": {
-    "@itwin/itwinui-illustrations-react": "^2.1.0",
     "react": ">= 17.0.0 < 19.0.0",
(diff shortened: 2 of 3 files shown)


rohan-kadkol (commented):
To confirm, would something like `new Function('specifier', 'return import(specifier)')` and/or `/* webpackIgnore: true */` like we had before ([code](https://github.com/iTwin/iTwinUI/commit/2b8627d2c15dbbe9d88162eaf56be11da2fa34f1#diff-301fb82041d47aad68eb5e6905c179614ff98afd8e896044fd0c4094e9ef9919L13-L15)) not fix this issue?

mayank99 (commented):
no, that's what we had before #1742 and it doesn't work because the illustrations don't get bundled. see https://stackblitz.com/edit/nextjs-ukqiky?file=package.json,app%2Fpage.tsx

webpackignore and viteignore are only to be used for importing from URLs (which the browser knows how to handle).

rohan-kadkol (approved):

PR #1747, opened 2023-12-20
improve `Tooltip` docs

mayank99:
## Changes

added the most important information that was missing from the `/tooltip` page:
- floating-ui
- best practices/guidelines
- accessibility features
- iconbutton `label`

can be improved further in the future to explain more advanced features. missing docs for:
- `reference` prop
- middleware, autoUpdate, etc
- controlled `visible`/`onVisibleChange` props
- disclosure pattern as an alternative

## Testing

N/A

## Docs

N/A


apps/website/src/content/docs/button.mdx
```diff
@@ -93,7 +93,7 @@ Pressing this button will always open a menu. Upon selecting an option from the
 
 <PropsTable path={frontmatter.propsPathDropdownButton} />
 
-## Icon button
+## IconButton
```
mayank99, line 96: i changed this here because it affects the fragment link that i'm using in `/tooltip`.

we should get rid of these spaces everywhere on all pages, and use the actual component name instead.

apps/website/src/content/docs/tooltip.mdx
```diff
@@ -1,36 +1,73 @@
 ---
 title: Tooltip
-description: Tooltips display informative text when users hover over, focus on, or tap an element.
-propsPath: '@itwin/itwinui-react/esm/core/Tooltip/Tooltip.d.ts'
+description: Tooltips display informative text when users hover, focus, or tap an interactive element.
 thumbnail: Tooltip
+related:
+  - popover
+  - expandableblock
```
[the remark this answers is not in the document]
mayank99: it is a disclosure pattern that should be preferred over tooltips in most cases. showing/hiding info on click rather than hover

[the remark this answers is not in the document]
mayank99: i would also like to expose a lower-level disclosure utility in the future to replace expandableblock here, but in the meantime this works 🤷

apps/website/src/content/docs/tooltip.mdx
```diff
@@ -1,36 +1,73 @@
...
+  <AllExamples.TooltipMainExample client:load />
+</LiveExample>
 
-It is not recommended to use tooltips to display critical information, or as a means to navigate an interface. They are necessary only when supplemental information is required. Functionality of the product should NOT be dependent on the user reading tooltips.
+For everything to work correctly, the trigger element must:
 
-Tooltips are enabled by default but can be turned off.
+- be an interactive element, for example, a `<button>`
+- forward its ref
+- delegate (spread) any arbitrary props
```
rohan-kadkol, line 26: Not sure I understand this point.

apps/website/src/content/docs/tooltip.mdx
```diff
@@ -1,36 +1,73 @@
...
 
-### Placement
+## Positioning
 
-Tooltips can also be customized such as changing the placement.
+`Tooltip` handles positioning using an external library called [Floating UI](floating-ui.com/). To control which side the tooltip should be placed relative to its trigger, use the `placement` prop (which defaults to `"top"`). If not enough space is available, then it will automatically flip to the opposite side.
```
rohan-kadkol, line 30: The Floating UI link is broken. Also fluency (nit).

```suggestion
`Tooltip` handles positioning using an external library called [Floating UI](https://floating-ui.com/). To control which side the tooltip should be placed relative to its trigger, use the `placement` prop (which defaults to `"top"`). If enough space is not available, then it will automatically flip to the opposite side.
```

rohan-kadkol (approved):

apps/website/src/content/docs/tooltip.mdx
mayank99: spread:
```
(props) => <div {...props}>
```

rohan-kadkol: Ohh, I guess this can be made clearer by showing it in one of the existing examples?

mayank99: link?
```suggestion
- delegate ([spread](https://react.dev/learn/passing-props-to-a-component#forwarding-props-with-the-jsx-spread-syntax)) any arbitrary props
```

rohan-kadkol: Ohh, I was not referring to the concept of spread. I was just confused as to what to spread and where.

But now re-reading the sentence above it ("For everything to work correctly, the trigger element must:"), I now understand that the props from the `Tooltip` have to be passed down to the trigger, if the trigger is a separate component instead of directly written inline.

So how it's worded currently is good 👍

rohan-kadkol: Ohh, sorry, with the link would be better 😅

PR #1748, opened 2023-12-20
improve codeblocks

mayank99:
## Changes

- updated astro and related deps
- now using the exact same nightOwl theme for both static and live examples (theme changes can happen in a separate PR)
- added [`astro-expressive-code`](https://github.com/expressive-code/expressive-code/blob/main/packages/astro-expressive-code/README.md) for syntax highlighting

this unlocks some more powerful features like emphasizing specific lines that i've demonstrated in themeprovider

## Testing

everything works the same as before, apart from codeblocks. tested in dev server and prod build.

## Docs

n/a


apps/website/src/pages/docs/_layout.astro
```diff
@@ -248,6 +248,10 @@ const currentPage = new URL(Astro.request.url).pathname;
           display: grid;
           gap: var(--space-1);
         }
+
+        :global(.expressive-code) {
+          overflow: auto;
+        }
```
rohan-kadkol, line 254: Is this needed? I don't think I noticed any difference on the ThemeProvider docs page after commenting this out.

So I guess this is something pre-emptive, that will be needed for the future?

apps/website/astro.config.ts
```diff
@@ -27,24 +29,27 @@ export default defineConfig({
       ],
     }),
     react(),
+    astroExpressiveCode({
+      themes: [nightOwl],
+      frames: { showCopyToClipboardButton: false },
```
rohan-kadkol, line 34: The "copy to clipboard" button is shown on all code examples except the ThemeProvider page's examples. Should this also be made true?

apps/website/astro.config.ts
```diff
@@ -27,24 +29,27 @@ export default defineConfig({
...
   },
   server: { port: 1700 },
   scopedStyleStrategy: 'where',
   vite: {
     ssr: {
-      noExternal: [
-        '@fontsource/noto-sans',
-        '@tippyjs/react',
-        '@itwin/itwinui-react',
-        !isDev && '@itwin/itwinui-icons-react',
```
rohan-kadkol, line 45: Just curious, why is this not needed anymore?

apps/website/src/components/LiveExample.astro
```diff
@@ -1,7 +1,7 @@
 ---
 import fs from 'node:fs';
 import OpenInCodesandbox from './OpenInCodesandbox.astro';
-import Code from 'astro/components/Code.astro';
+import { Code } from 'astro/components';
```
rohan-kadkol, line 4: Not sure, but do the code examples from ThemeProvider use a slightly different viewer than other pages? Because I noticed that the `.expressive-code` class was only on the code examples from the ThemeProvider docs page.

apps/website/src/pages/docs/_layout.astro
mayank99: try it on pages that have longer lines of code. resize your browser to be narrower if that helps

apps/website/astro.config.ts
mayank99: i believe this was needed before we fixed ESM support in icons

apps/website/src/components/LiveExample.astro
mayank99: `expressive-code` is used for static examples, `Code` is used for live examples (that have a preview)

apps/website/astro.config.ts
mayank99: this comes from expressive-code and it looks different from the one in our LiveExample, so i've disabled it.

plus, since static examples are intended for shorter snippets and do not show the full code, there is no need to make the whole thing copyable

apps/website/src/components/LiveExample.astro
mayank99: compare the tabs page with the [deployed version](https://dev-itwinui.bentley.com/docs/tabs) to see the difference more clearly

apps/website/astro.config.ts
rohan-kadkol: > this comes from expressive-code and it looks different from the one in our LiveExample, so i've disabled it.

Ohh, I see, sounds good.

> plus, since static examples are intended for shorter snippets and do not show the full code, there is no need to make the whole thing copyable

At least for me, when the code snippets are small, I tend to copy the entire thing. That way I don't have to reconstruct the DOM for components that use the composition API (e.g. Tabs).

For full code, I only copy selective lines that are pertinent to my use case.

mayank99: then i'll flip it so that only static examples are copyable. i want to avoid the styling inconsistency that occurs when both are present

[`ffb3b0b`](https://github.com/iTwin/iTwinUI/pull/1748/commits/ffb3b0b4721b796957759f9e4b2024a68c787549)

rohan-kadkol (approved):

yarn.lock
mayank99, line 1: this is driving me insane. i cannot get the react visual tests to pass, no matter what i do.

this is very likely because yarn not knowing how to correctly resolve multiple versions of the same dependency within a monorepo 😐

mayank99: i've just updated the images 🤷

PR #1355, opened 2023-06-14
Rebuilding input container

gretanausedaite:
### React:

InputFlexContainer:
- Add ‘size’ prop

InputContainer:
- Using InputGrid css
- Using new StatusMessage component
- Removed ‘isIconInline’ prop;

Icon:
- Added ‘padded’ prop (false by default)

TransferList:
- Remove 'iui-transfer-list-listbox-label' class
- Using simple Label

Textarea:
- Added ‘status’ prop

DatePickerInput:
- Using InputGrid, Label, InputWithDecorations

StatusMessage:
- Added wrapper class ‘iui-status-message’
- Added ‘iconProps’ and ‘contentProps’

Select:
- Added ‘status’ prop
- Removed click ripple effect animation from caret down.

LabeledTextarea:
- Added ‘wrapperProps’, ‘labelProps’’, ‘messageContentProps’, ‘iconProps’;
- Using InputGrid and InputWithDecorations;
- Removed ‘textareaClassName’ and ‘textareaStyle’ props, Users can pass custom styling to textarea using classname and style props

LabeledSelect:
- Added ‘wrapperProps’, ‘labelProps’’, ‘messageContentProps’, ‘messageIconProps’;
- Using InputGrid;
- Removed ‘selectClassName’, ‘selectStyle’ props. Users can pass custom styling to select using classname and style props;

LabeledInput:
- Added ‘messageContentProps’, iconProps, labelProps, inputWrapperPRops;
- Removed ‘inputClassName’ and ‘inputStyle’ props;
- Using InputGrid and InputWithDecorations;

Label:
- Added disabled prop;

InputWithDecorations:
- New component
- Subcomponents: Input and Button.

InputGroup:
- Added labelProps, messageProps and innerProps;
- Using InputGrid

InputGrid:
- New component

Input:
- Add status prop;

ComboBox:
- Removed functionality to close menu on caret down click;

ColorInputPanel:
- Simplified DOM structure;

### CSS:

Icon:
- Added padded Icon styling using data-iui-padded attribute

TransferList:
- Removed iui-transfer-list-listbox-label

ColumnFIlter:
- Updated to use input-grid

StatusMessage:
- Added wrapper class iui-status-message

Select:
- Added status colouring

NonIdealState:
- Updated to use input-grid

InputContainer:
- Removed mixins;
- Removed iui-input-container class;
- Added iui-input-grid, iui-input-label, iui-input-group-wrapper, ‘iui-input-group’  classes;
[...]

PR #1753, opened 2023-12-27
`LabeledInput` API and styling improvements

rohan-kadkol:
Since this PR changes `LabeledInput`'s internal working, this is a pre-requisite for the `LabeledInput` related PR #1777.

## Motivation

This PR handles some `LabeledInput` API and styling changes based on the suggestions in https://github.com/iTwin/iTwinUI-migration-tool/pull/96#discussion_r1436581272:

1. When `StatusMessage`'s `svgIcon=null`, no icon is displayed along with the message (https://github.com/iTwin/iTwinUI/pull/1753/commits/fd663c0797512efe66f3069c76ca4802600c5628)
> it appears like this, but i'm thinking we should change the source code to auto-apply the fill too.

2. Apply status fill to `LabeledInput`'s custom icon also (https://github.com/iTwin/iTwinUI/pull/1753/commits/6c387d5055186a182dbf017d0a5b62de30800f71)
> since we only support basic svg icons to be passed into the `svgIcon` prop, we should probably _not_ use `InputWithDecorations` inside `LabeledInput`/`LabeledTextarea` in v3.
`InputWithDecorations` should be reserved for advanced cases (like passing buttons).
3. Replaced `InputWithDecorations` with `InputFlexContainer` in `LabeledInput` and `LabeledTextarea` (https://github.com/iTwin/iTwinUI/pull/1753/commits/b2fe1397528082d0adbd8ece8121e0a4b11eacff)

## Changes

1. Addressed the points in the motivation section.
2. `.iui-input-with-icon` now does not add `padding-inline-end` when no end-icon is present.
3. Removed unnecessary `status` passed to `Select` and `Input` in `LabeledSelect` & `LabeledInput` respectively.
4. Removed accidental styling support for the invalid case of `.iui-input` and `.iui-input-with-icon > .iui-select-button` with `[data-iui-status='informational']`. This is invalid because an `informational` status on our input or select adds little to no meaning.
5. Deprecated `svgIcon` in `LabeledSelect` in favor of passing `<StatusMessage startIcon={svgIcon} />` to the `message` prop.
6. Updated `input.html` and `select.html` to visual test cases for adding `data-iui-status` on `iui-input-grid`, `iui-input`, and `iui-select-button`,
7. Add missing `@include iui-text-selection($status);` for the `.iui-input-flex-container[data-iui-status={…}] .iui-input` case.
8. Removed duplicate status styling applied on for the case where `data-iui-status` is applied on `.iui-input-grid` and the case where it is applied on `.iui-input, .iui-input-with-icon > .iui-select-button`.
9. Added missing `data-iui-status` for all uses of `<InputGrid>` in itwinui-react, since this data attribute exists in itwinui-css.

## Testing

* Motivation point 1: Added unit test in `StatusMessage.test.tsx` to confirm that no icon is in the DOM if `svgIcon=null`
* Motivation point 2: Added new "With Svg Icon" stories to `LabeledInput` and `LabeledTextarea` to test that the status fill is applied to `svgIcon`. Open to suggestions for better ways to test this.
* Motivation point 3: Edited failing unit tests to account for this change.
* Change point 2: The input with type number in the Table → Filter [story](https://github.com/iTwin/iTwinUI/pull/1753/files#diff-368e5b17d4aba72dcf8f6f5c2bbbd2bab587ac49076302931a5ca931167dbd38) tests this.
* Change point 3: No test images failed and the focus outlines appeared correctly with manual testing
* Change point 4: Mayank and I manually verified that adding an informational status on an input or select conveyed little to no extra meaning than without the status.
* Change point 5: Updated `LabeledSelect` story to discontinue using the newly deprecated `svgIcon` prop
* Added react-workshop tests on existing stories to test that the status focus outline is applied correctly.

## Docs

Added changesets

## TODOs after PR:

* [ ] Look into having a consistent padding among `input-flex-container`, `input-with-icon`, `search-box` (https://github.com/iTwin/iTwinUI/pull/1753#discussion_r1449405834)
  * Maybe create a new issue (or add a bullet point to an existing) issue
* [ ] Fix `InputGroup` bug: passing `<StatusMessage startIcon={…} />` to the message of `<InputGroup status="…" />` shows duplicate status icons. (https://github.com/iTwin/iTwinUI/pull/1753#discussion_r1459238732)


packages/itwinui-react/src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -104,24 +104,24 @@ export const LabeledInput = React.forwardRef((props, ref) => {
         </Label>
       )}
 
-      <InputWithDecorations
+      <InputFlexContainer
```
mayank99, line 107: Simply swapping in `InputFlexContainer` does not do anything; `InputWithDecorations` already uses `InputFlexContainer` underneath. The important thing to realize is that these two components are designed for complex/arbitrary custom decorations on either side ([example](https://itwin.github.io/iTwinUI/react/?story=searchbox--basic-with-custom-items&theme=dark)).

In the case of `LabeledInput` and `LabeledTextarea`, we only allow one simple svg to passed into `svgIcon` and we know exactly where to put it, so we can use a simpler approach similar to v2:
1. show the input/textarea as-is.
2. absolutely position the icon on the right end
3. add some padding onto the input/textarea.

This will also fix the strange bug where the textarea resize handle appears offset from the bottom-right corner.

| v2 | v3 |
| --- | --- |
| image | image |

rohan-kadkol: To confirm, is [`d3b16d5` (#1753)](https://github.com/iTwin/iTwinUI/pull/1753/commits/d3b16d516d08191ef413e57bbad6cdbbb2ea2d7f) (WIP) a step in the right direction? In this commit, I tried re-using `iui-input-with-icon`.

This fixed the textarea resize issue you mentioned. But I still need to investigate why the border color is not the same as the status color when a `LabeledInput`/`LabeledTextarea` is focused.

https://github.com/iTwin/iTwinUI/assets/45748283/7375163b-ec8b-4f6e-a7eb-8cca1d7bee55

mayank99: Yeah, that's generally the right direction. Do we already have the `iui-input-with-icon` class? i.e. no css changes needed?

rohan-kadkol: > Do we already have the `iui-input-with-icon` class? i.e. no css changes needed?

Yes, I believe it is used in Select and the ComboBox to show the down arrow icon ([full repo search](https://github.com/search?q=repo%3AiTwin%2FiTwinUI%20iui-input-with-icon&type=code)).

By reusing `iui-input-with-icon`, big CSS changes may not be needed, but I still might need to make a few small changes to fix issues such as the border color issue. Will investigate this further.

packages/itwinui-react/src/core/LabeledTextarea/LabeledTextarea.test.tsx
```diff
@@ -33,18 +33,21 @@ it('should show status', () => {
...
+  expect(textarea).toHaveAttribute('data-iui-status', 'negative');
 });
 
 it('should be disabled', () => {
   const { container } = render(
     <LabeledTextarea label='Label' disabled={true} />,
   );
 
-  const input = container.querySelector('.iui-input-flex-container');
-  expect(input).toHaveAttribute('data-iui-disabled', 'true');
```
rohan-kadkol, line 46: Removed this because `data-iui-disabled` is not applied to `iui-input-with-icon`

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -16,6 +16,10 @@ type InputGridOwnProps = {
    * @default 'default'
    */
   labelPlacement?: 'default' | 'inline';
+  /**
+   * Status of the InputGrid (and its descendants).
+   */
+  status?: 'positive' | 'warning' | 'negative';
```
rohan-kadkol, line 22: Added the `status` prop since in itwinui-css, the input grid needs `data-iui-status` to correctly add status colored outlines.

packages/itwinui-css/src/input/input.scss
```diff
@@ -108,7 +108,11 @@
 
   > :first-child {
     grid-area: 1 / -1;
-    padding-inline-end: var(--iui-component-height); // Width of the end-icon button
+
+    // When there is an end-icon button (i.e. first-child is not the only child)
+    &:not(:only-child) {
+      padding-inline-end: var(--iui-component-height); // Width of the end-icon button
+    }
```
rohan-kadkol, line 115: To avoid adding padding when no end-icon is present in `.iui-input-with-icon`. This fixes the [Table → Filter story](https://itwin.github.io/iTwinUI/react/?story=table--filters&theme=dark) test.

|Without this change|With this change|
|-|-|
|Screenshot 2024-01-11 at 3 41 08 PM|Screenshot 2024-01-11 at 3 38 57 PM|

apps/react-workshop/src/LabeledTextarea.test.ts
```diff
@@ -21,4 +21,19 @@ describe('LabeledTextarea', () => {
       cy.compareSnapshot(testName);
     });
   });
+
+  // To also test that the correct status outline is applied
+  ['Positive', 'Warning', 'Negative'].forEach((testName) => {
```
rohan-kadkol, line 26: There are tests in `css-workshop` for this. Adding this in `react-workshop` too since the [status outline not applied issue](https://github.com/iTwin/iTwinUI/pull/1753#discussion_r1437277459) was only on the react side.

apps/react-workshop/cypress-visual-screenshots/baseline/LabeledInput.test.ts-Inline.png
rohan-kadkol, line 1: The visually perceived padding of the end icon has changed since we're now using `iui-input-with-icon` instead of `iui-input-flex-container`.

|Previously|Now|
|-|-|
|Screenshot 2024-01-11 at 3 24 20 PM|Screenshot 2024-01-11 at 3 24 10 PM|

packages/itwinui-react/src/core/LabeledInput/LabeledInput.test.tsx
```diff
@@ -28,20 +40,22 @@ it('should render disabled component', () => {
...
+        '.iui-input-with-icon > input',
+      ) as HTMLInputElement
+    ).disabled,
+  ).toBe(true);
 });
 
 it('should handle required attribute', () => {
   const { container } = render(<LabeledInput label='some label' required />);
   assertBaseElement(container);
-  expect(container.querySelector('.iui-input-flex-container')).toBeTruthy();
```
rohan-kadkol, line 44: Removed some `expect(inputContainer).toBeTruthy()` test like this since this is already tested in `assertBaseElement`.

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
mayank99: how was it working previously then? we weren't using the CSS that sets `data-iui-status`?

packages/itwinui-css/src/input/input.scss
mayank99: we can make this more robust using `:has`.

```scss
&:has(+ .iui-end-icon) {
  padding-inline-end: var(--iui-component-height);
}

@supports not selector(:has(+ *)) {
  &:not(:only-child) {
    padding-inline-end: var(--iui-component-height);
  }
}
```

it is not good to rely on selectors `:first-child`/`:only-child` etc

apps/react-workshop/cypress-visual-screenshots/baseline/LabeledInput.test.ts-Inline.png
mayank99: Should we align the padding of input-flex-container to match input-with-icon and search-box? (in a different PR)

packages/itwinui-react/src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -104,25 +105,24 @@ export const LabeledInput = React.forwardRef((props, ref) => {
...
           disabled={disabled}
           required={required}
+          status={status}
           id={id}
           ref={ref}
           {...rest}
         />
         {shouldShowIcon && (
-          <Icon fill={!svgIcon ? status : undefined} padded {...iconProps}>
+          <EndIcon status={status} {...iconProps}>
```
mayank99, line 121: How do we support the case where user wants to override fill?

packages/itwinui-css/src/input/input.scss
rohan-kadkol: Ohh, I see, changed (https://github.com/iTwin/iTwinUI/pull/1753/commits/b7f876304305f55a805493cc305d8ac16a9d97b5)

packages/itwinui-react/src/core/LabeledInput/LabeledInput.tsx
mayank99: Ah ok, it can be handled using `iconProps={{ fill: 'default' }}`.

mayank99, line 121: I feel like this component doesn't add anything valuable?

It's the same as:
```tsx
<Icon status={status} {...iconProps} className={cx('iui-end-icon', iconProps?.className)}>
```

rohan-kadkol: Since we are passing down `{...iconProps}` to `<EndIcon />`, and since `fill={status}` is above `{...rest}`, users can pass their custom fill via `iconProps` to override our fill.

```jsx
<LabeledInput
  label='label'
  message='message'
  status='negative'
  svgIcon={<SvgPlaceholder />}
  iconProps={{
    fill: 'hotpink',
  }}
/>
```

However, upon testing this, it doesn't work:

image

When testing how `<Icon>` behaves when `fill` is a custom string color, I realized that the fill is not applied as I expected ([Stackblitz playground](https://stackblitz.com/edit/github-psiuv8?file=src%2FApp.tsx)).

Is there an other way to pass custom fills to `<Icon>`?

mayank99: I'm not sure if we want to support custom fill. I was just asking about overriding the fill from a status color to `"default"`

Anything else could be achieved using CSS

rohan-kadkol: Ohh, yes, `EndIcon` seems unnecessary; removed (https://github.com/iTwin/iTwinUI/pull/1753/commits/94cf518e3aaa19a08636c25376870dc0c1ee3823).

I think I had it that way since I was basing it off of [`ComboBoxEndIcon`](https://github.com/iTwin/iTwinUI/blob/5d8986451e10101085c98a084fcc4a67b0d76d1e/packages/itwinui-react/src/core/ComboBox/ComboBoxEndIcon.tsx#L16-L36) when I first started working on this PR.

packages/itwinui-react/src/core/StatusMessage/StatusMessage.tsx
```diff
@@ -51,14 +53,17 @@ export const StatusMessage = React.forwardRef((props, ref) => {
 
   const icon = userStartIcon ?? (status && StatusIconMap[status]());
 
+  // If user passes null, we don't want to show the icon
+  const shouldShowIcon = userStartIcon !== null && !!icon;
```
mayank99, line 57: This change made me realize something: we can rely on the default icon behavior of `StatusMessage` in more places. 

Can we go through all uses of `StatusMessage` and remove the ones where we are unnecessarily passing a `startIcon` from `StatusIconMap`?

apps/react-workshop/cypress-visual-screenshots/baseline/LabeledInput.test.ts-Inline.png
rohan-kadkol: Yes, I think that will be better. I added a bullet point to the PR description to remember to address this point after this PR.

packages/itwinui-react/src/core/StatusMessage/StatusMessage.tsx
rohan-kadkol: Sounds good, done (https://github.com/iTwin/iTwinUI/pull/1753/commits/19e955ce6cfd45db7040dd7a7c79e32f1ad0ae33).

When doing this for `InputGroup`, I noticed some potentially incorrect behavior of the icon in the main branch: when adding a status (with or without message), does it feel like a third radio button exists and has been checked?
 
Screenshot 2024-01-12 at 9 51 19 AM

I created a [Stackblitz playground](https://stackblitz.com/edit/github-3frwv5?file=src%2FApp.tsx) for this.

mayank99: Can't you just pass `svgIcon` directly to the `StatusMessage`? Instead of creating an `icon()` wrapper

And the `InputGroup` behavior is not incorrect. That's just how the icon looks. https://itwin.github.io/iTwinUI-icons/#icons-status-success

rohan-kadkol: I had kept the `icon()` wrapper to maintain the logic of showing a status icon only when `status && message`.

`LabeledSelect`:
But looks like that is handled by the check `if (typeof message === "string")`:

```jsx
{typeof message === 'string' ? (
  <StatusMessage
    status={status}
    startIcon={displayStyle === 'default' ? icon() : undefined}
  >
    {message}
  </StatusMessage>
) : (
  message
)}
```

So, when message is `undefined`, we fall into the else branch and render `undefined` (`message`) instead of `StatusMessage`. So yes, looks like the `icon()` wrapper can be removed from `LabeledSelect`.

---

`InputGroup`:
Upon re-reading the previous logic in `icon()` and the `return`, looks like the previous logic can still be maintained by removing the `icon()` wrapper.

---

So removed in both places (https://github.com/iTwin/iTwinUI/pull/1753/commits/5f053f94d85f3a83e3a55841958d88660d2115f9)

Also, I noticed that the `displayStyle === 'default' ? icon() : undefined` check in `LabeledSelect` was not really doing anything ([StackBlitz playground](https://stackblitz.com/edit/github-ejxhhk?file=src%2FApp.tsx)). So I changed it to directly be `svgIcon`.

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
rohan-kadkol: The outline was previously coming from the input-flex-container:

image

… and `--iui-color-border-negative` was being by the `data-iui-status` on the input-flex-container

---

Now, it is coming from input-with-icon:

Screenshot 2024-01-12 at 10 24 37 AM

… and `--_iui-input-with-icon-focus-color` was being set by the `data-iui-status` on the input grid

image

https://github.com/iTwin/iTwinUI/blob/5d8986451e10101085c98a084fcc4a67b0d76d1e/packages/itwinui-css/src/input-container/input-container.scss#L66

mayank99: So we weren't using `iui-input-grid[data-iui-status]` anywhere in React before, correct?

Since `InputGrid` is a public component and is exposing a public `status` prop, I would expect it to work with various permutations, including:
- `Input` within `InputGrid`
- `InputWithDecorations` within `InputGrid`

If such combinations require more work, then we should get rid of the `status` prop and manually set `data-iui-status` inside `LabeledInput` (for now)

rohan-kadkol: > So we weren't using `iui-input-grid[data-iui-status]` anywhere in React before, correct?

It is being used by our private `InputContainer` component. 

https://github.com/iTwin/iTwinUI/blob/5d8986451e10101085c98a084fcc4a67b0d76d1e/packages/itwinui-react/src/core/utils/components/InputContainer.tsx#L48-L50

`InputContainer` is used only in `ComboBoxInputContainer`.

---

Actually, after looking at `InputContainer`, I realized that maybe I can re-use `InputContainer` inside `LabeledInput`, similar to how `ComboBoxInputContainer` does it. If so, they we can skip adding the new `status` to `InputGrid`.

mayank99: no, I think we should remove `InputContainer` to make things less confusing

mayank99: see relevant comment: https://github.com/iTwin/iTwinUI/pull/1355#discussion_r1277577290

rohan-kadkol: Then since this PR has already handles many cases and since this is a pre-requisite for #1777, I think it would be better to add a `status` prop to `InputGrid` in a future PR.

So, I will delete the `status` prop and add `data-iui-status` manually. I will also create a checklist item in the "TODOs after PR" section of this PR to consider adding `status` to `InputGrid` and try removing `InputContainer` in a future PR.

rohan-kadkol: Done in [`405c5cf` (#1753)](https://github.com/iTwin/iTwinUI/pull/1753/commits/405c5cff60877e34fa93b118a074a105cba158ba)

.changeset/chilly-wasps-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`LabeledInput` & `LabeledTextarea` now internally use `iui-input-with-icon` instead of `InputWithDecorations`.
```
mayank99, line 5: These are implementation details that should not be part of react changeset. Try to think about how it affects our users.

```suggestion
`LabeledInput` and `LabeledTextarea` have been slightly reworked to fix alignment and consistency issues when used with `svgIcon`.
```

.changeset/unlucky-crabs-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': minor
```
mayank99, line 2: Probably should be patch? It's not a new feature.

```suggestion
'@itwin/itwinui-css': patch
```

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
```diff
@@ -114,11 +114,7 @@ export const LabeledSelect = React.forwardRef(
     const labelId = `${useId()}-label`;
 
     return (
-      <InputGrid
-        labelPlacement={displayStyle}
-        status={status}
-        {...wrapperProps}
-      >
+      <InputGrid labelPlacement={displayStyle} {...wrapperProps}>
```
mayank99, line 117: Are you forgetting to set `data-iui-status` here? Also same in `InputGroup`

rohan-kadkol: I don't think it is needed in `LabeledSelect` and `InputGroup`, since the outline for `LabeledSelect` and general functioning of status for `InputGroup` work correctly before and after this PR.

Is it better to still add `data-iui-status` to these two components for consistency/correctness?

mayank99: Then I guess I'm confused why in `LabeledInput` you need to set **both** `<InputGrid data-iui-status={status}>` and `<Input status={status}>`. Is the latter needed?

mayank99: seems like there's some duplication here

https://github.com/iTwin/iTwinUI/blob/5d8986451e10101085c98a084fcc4a67b0d76d1e/packages/itwinui-css/src/input-container/input-container.scss#L34-L84

https://github.com/iTwin/iTwinUI/blob/5d8986451e10101085c98a084fcc4a67b0d76d1e/packages/itwinui-css/src/input-container/input-container.scss#L272-L302

rohan-kadkol: Yes, looks like passing `status` to `Input` was not required since the focus outline and the bottom status line still appear correctly on the input. Also, all `react-workshop` test images are still passing. So, I stopped passing `status` to the `Input`. (https://github.com/iTwin/iTwinUI/pull/1753/commits/35fe942b2c1fa42c8f51f4da62b5825e0ae7e78c)

mayank99: Three suggestions then:
1. Lets pass `data-iui-status` to all instances of `iui-input-grid` just to be safe
2. Lets try to remove the `status` passed to `Select` (inside `LabeledSelect`) and see what happens
3. Lets try to fix the CSS duplication by using a common selector. e.g.:
   ```scss
   @each $status in informational, positive, warning, negative {
     :is(.iui-input-grid, .iui-input, .iui-input-with-icon > .iui-select-button) {
       &:where([data-iui-status='#{$status}']) {
   ```

mayank99: also, unit tests need to be fixed

rohan-kadkol: > Lets try to fix the CSS duplication by using a common selector

I tried to remove the duplication by creating a new mixin, `iui-input-status`. (https://github.com/iTwin/iTwinUI/pull/1753/commits/12c8b67f9f2ac964747bfb816ddf7f602875fe00).

When testing, I realized that there were 3 cases of applying status styling to an input:
1. `data-iui-status` on `.iui-input`
2. `data-iui-status` on `.iui-input-with-icon > .iui-select-button`
3. `data-iui-status` on `.iui-input-grid .iui-input`

I modified a few existing css-workshop demos in `input.html` and `select.html` to test cases 2 & 3 (https://github.com/iTwin/iTwinUI/pull/1753/commits/1b9e70a1034e9932fc5b4821f0d387ae0d503784). This also replaces some unnecessary `iui-{status}` classes with `data-iui-status={status}` data attributes. As a result, the missing bottom status colored border on Select with status came back.

To test case 1, I created three new simple `.iui-input` without a `.iui-input-grid` parent (https://github.com/iTwin/iTwinUI/pull/1753/commits/18a67da9950ffe2692d6ad093f53722588bf87d0). I had to rearrange the groups of inputs a little to prevent the two sets of status inputs coming one after the other and thus creating a confusing demo. I am open to ideas to better test it without convoluting the demos too much.

mayank99: Creating a mixin does not remove duplication in the final CSS if you are still using the mixin in two different places. This is why I suggested using a common selector in my [comment above](https://github.com/iTwin/iTwinUI/pull/1753#discussion_r1450724284).

rohan-kadkol: > Lets pass `data-iui-status` to all instances of `iui-input-grid` just to be safe

Sounds good, added `data-iui-status` to all `InputGrid`/`iui-input-grid` + `status` (i.e. excluding DatePickerInput). (https://github.com/iTwin/iTwinUI/pull/1753/commits/1003d5b98f8fd6bf37f8d3fee735b34100da43f2)

> Lets try to remove the `status` passed to `Select` (inside `LabeledSelect`) and see what happens

After removing `status` passed to `Select` inside `LabeledSelect`, the focus outlines worked correctly and the failed images didn't have any breaking change. So, I proceeded with removing `style={style}` (https://github.com/iTwin/iTwinUI/pull/1753/commits/dbd9c75bd209d7edb66541d35aa6e0938f2b7db2).

I also fixed the resulting failing unit tests (https://github.com/iTwin/iTwinUI/pull/1753/commits/acc68ad5f1d7c0b693636dcf558a6ef4d5e594ed)

rohan-kadkol: > Creating a mixin does not remove duplication in the final CSS if you are still using the mixin in two different places. This is why I suggested using a common selector in my [comment above](https://github.com/iTwin/iTwinUI/pull/1753#discussion_r1450724284).

When I first tried to use the common selector, I realized that the cases were a bit involved to combine. Because in the `iui-input-grid` case, the selector was `&[data-iui-status={status}] .iui-input`:

image

… whereas in the `.iui-input` & `.iui-input-with-icon > .iui-select-button` cases, the selector was`&[data-iui-status={status}]`:

image
image

Since `data-iui-status` was applied on different elements, it was a bit involved to combine these selectors and use the same `@each` loop.

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -67,17 +47,7 @@
...
-          content: '';
-          inline-size: 100%;
-          position: absolute;
-          inset-block-end: 0;
-          inset-inline-start: 0;
-          block-size: var(--iui-size-3xs);
-          background-color: var(--iui-color-border-#{$status});
-        }
+        border-block-end: var(--iui-size-3xs) solid var(--iui-color-border-#{$status});
       }
```
rohan-kadkol, line 51: I realized that adding `border-block-end` similar to `iui-input` ([code](https://github.com/iTwin/iTwinUI/pull/1753/files#diff-79a618b1da06b9cedff0df3a138d8a8da22a8185e8f0b5a2cf6c8aeb23a49b5cL51)) also gave the same effect as the `&::after {…}`. So I replaced the multi-line `&::after {…}` with a single line `border-block-end`.

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
mayank99: Does something like this work? Basically matching the selector against both itself as well as against descendants. 

```scss
:is(.iui-input-grid, .iui-input, .iui-input-with-icon > .iui-select-button) {
  @each $status in informational, positive, warning, negative {
    &:where([data-iui-status='#{$status}']) {
      :where(.iui-input, .iui-select-button),
      &:where(.iui-input, .iui-select-button) {
        // …
      }

      :where(.iui-input),
      &:where(.iui-input) {
        // …
      }

      :where(.iui-select-button),
      &:where(.iui-select-button) {
        // …
      }
    }
  }
}
```

packages/itwinui-css/src/input-container/input-container.scss
mayank99: I believe the reason for using `::after` was to preserve vertical text alignment. If the bottom border is 2px and the top border is 1px, then the text does not appear centered.

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
rohan-kadkol: Yes, it worked, but surprisingly, it actually produced a larger `input-container.css` file that the existing mixins method.

I guess that is because the `iui-input-status` mixin has multiple selector branches inside it. Thus, `&` is written multiple times for each iteration of the loop. Since the `&` is more complex in the `:is(…)` than in the simpler existing selectors, each generated css selector is much longer when using the combined `:is(…)` selector. 

Thus, even though the combined scss selector may produce fewer css selectors, each selector is so long that the generated css is actually larger than when having more but smaller selectors.

I played around with a few similar combined `:is(…)` selectors and pasted the generated `input-container.css` file in this [Stackblitz playground](https://stackblitz.com/edit/github-cjgv2b?file=src%2Fdemo.scss).

mayank99: Maybe i didn't explain my suggestion fully, but essentially you're using it wrong by writing all the code inside a mixin and still using the mixin multiple times.

The purpose of those different selectors is to allow branching for small parts of the code.

I'm pasting the full code below if you want to try it out. It should be smaller than the original.

<details>
<summary>input-container.scss</summary>

```scss
@use '../mixins' as *;

.iui-input-grid {
  display: grid;
  row-gap: var(--iui-size-2xs);
  grid-template-columns: 1fr;
  grid-template-rows: auto;
  align-items: center;

  &[data-iui-label-placement='inline'] {
    grid-template-rows: 1fr auto;
    grid-template-columns: auto 1fr;

    .iui-input-label {
      margin: 0 var(--iui-size-m) 0 0;

      &.iui-required {
        margin-inline-end: calc(var(--iui-size-2xs) * 1.5 - 1px);
      }
    }

    .iui-input-group {
      flex-direction: row;
      gap: var(--iui-size-m);
    }
  }

  .iui-status-message {
    grid-column-end: -1;
  }
}

:is(.iui-input-grid, .iui-input, .iui-input-with-icon > .iui-select-button) {
  @each $status in positive, warning, negative {
    &:where([data-iui-status='#{$status}']) {
      :where(.iui-input, .iui-select-button),
      &:where(.iui-input, .iui-select-button) {
        padding-block-end: calc(var(--iui-size-2xs) - 1px);

        &[data-iui-size='small'] {
          padding-block-end: 0;
        }

        &[data-iui-size='large'] {
          padding-block-end: calc(var(--iui-size-xs) - 1px);
        }
        // #endregion

        border-block-end: var(--iui-size-3xs) solid var(--iui-color-border-#{$status});

        &:focus {
          border-block-end: var(--iui-size-3xs) solid var(--iui-color-border-#{$status});
          outline: 2px solid var(--iui-color-text-#{$status});
          outline-offset: -2px;
        }

        &:hover {
          --_iui-input-border-color: var(--iui-color-border-#{$status});
          --_iui-input-with-icon-hover-border-color: var(--iui-color-border-#{$status});
        }
      }

      :where(.iui-select-button),
      &:where(.iui-select-button) {
        border-block-end: transparent;

        &::after {
          content: '';
          inline-size: 100%;
          position: absolute;
          inset-block-end: 0;
          inset-inline-start: 0;
          block-size: var(--iui-size-3xs);
          background-color: var(--iui-color-border-#{$status});
        }
      }
    }
  }
}

.iui-input-label {
  color: var(--_iui-input-label-color);

  &.iui-required {
    &::after {
      content: '*';
      margin-inline-start: var(--iui-size-2xs);
      color: var(--iui-color-text-negative);
    }
  }

  &[data-iui-disabled='true'] {
    --_iui-input-label-color: var(--iui-color-text-disabled);
  }

  cursor: default;

  &:is(label) {
    cursor: pointer;

    &[data-iui-disabled='true'] {
      cursor: not-allowed;
    }
  }
  display: block;

  &.iui-inline {
    margin: 0 var(--iui-size-m) 0 0;

    &.iui-required {
      margin-inline-end: calc(var(--iui-size-2xs) * 1.5 - 1px);
    }

    display: inline-flex;
    align-items: center;
  }
}

.iui-input-flex-container {
  --_iui-input-background-color: var(--iui-color-background);
  --_iui-input-border-color: var(--iui-color-border-foreground);
  --_iui-input-accent-border-color: var(--iui-color-border-accent);
  --_iui-input-text-color: var(--iui-color-text);
  --_iui-input-padding-block: var(--iui-size-2xs);
  --_iui-input-min-height: var(--iui-component-height);

  display: inline-flex;
  align-items: center;
  background-color: var(--_iui-input-background-color);
  color: var(--_iui-input-text-color);
  position: relative;
  block-size: var(--_iui-input-min-height);
  inline-size: 100%;
  isolation: isolate;
  min-block-size: var(--_iui-input-min-height);
  block-size: fit-content;

  // Adding before pseudo-element to make inner elements more centered
  &::before {
    content: '';
    position: absolute;
    inset: 0;
    border: 1px solid var(--_iui-input-border-color);
    transition: border-color var(--iui-duration-1) ease-out;
  }

  &:hover {
    --_iui-input-border-color: var(--iui-color-border-foreground-hover);
  }

  &:has(input:focus-visible),
  &:has(textarea:focus-visible) {
    outline: 2px solid var(--_iui-input-accent-border-color);
    outline-offset: -2px;

    // Added to make sure border color is not visible with outline
    &::before {
      border-color: transparent;
    }
  }

  @supports not selector(:has(+ *)) {
    &:where(:focus-within) {
      outline: 2px solid var(--_iui-input-accent-border-color);
      outline-offset: -2px;

      // Added to make sure border color is not visible with outline
      &::before {
        border-color: transparent;
      }
    }
  }

  input,
  textarea {
    box-sizing: border-box;
    background: transparent;
    border: none;
    font: inherit;
    z-index: 2;
    min-inline-size: var(--_iui-input-min-height);
    min-block-size: var(--_iui-input-min-height);
    flex: 999;
    padding-inline: var(--iui-size-s);
    padding-block: var(--_iui-input-padding-block);

    &:where(:not(:first-child)) {
      padding-inline-start: var(--iui-size-2xs);
    }

    &::placeholder {
      color: var(--iui-color-text-placeholder);
      font-weight: var(--iui-font-weight-light);
      opacity: 1;
    }

    &:focus {
      outline: none;
    }
  }

  &[data-iui-disabled='true'] {
    --_iui-input-background-color: var(--iui-color-background-disabled);
    --_iui-input-border-color: var(--iui-color-border-disabled);
    --_iui-input-with-icon-hover-border-color: var(--iui-color-border-disabled);
    cursor: not-allowed;
  }

  &[data-iui-size='small'] {
    --_iui-input-padding-block: 0;
    --_iui-input-min-height: var(--iui-component-height-small);
  }

  &[data-iui-size='large'] {
    --_iui-input-padding-block: var(--iui-size-xs);
    --_iui-input-min-height: var(--iui-component-height-large);
    font-size: var(--iui-font-size-2);
  }

  @each $status in positive, warning, negative {
    &[data-iui-status='#{$status}'] {
      --_iui-input-accent-border-color: var(--iui-color-border-#{$status});

      &::before {
        border-block-end-width: var(--iui-size-3xs);
        border-block-end-color: var(--_iui-input-accent-border-color);
      }

      &:hover {
        --_iui-input-border-color: var(--iui-color-border-#{$status});
      }
    }
  }
}

.iui-input-group-wrapper {
  &[data-iui-label-placement='inline'] {
    display: flex;
    flex-direction: row;

    .iui-status-message {
      margin-inline-start: var(--iui-size-m);
    }
  }
}

.iui-input-group {
  display: flex;
  flex-direction: column;

  .iui-toggle-switch-wrapper {
    padding: calc(var(--iui-size-s) * 0.5) 0;
  }

  .iui-radio-wrapper,
  .iui-checkbox-wrapper {
    min-block-size: var(--iui-size-l);
  }
}

```
</details>

rohan-kadkol: Ohh, I see, I implemented your suggestion (https://github.com/iTwin/iTwinUI/pull/1753/commits/13d3e14daf8c40155bc6059f547c1f0ce48ef35e). I also added the text selection styles (https://github.com/iTwin/iTwinUI/pull/1753/commits/ab812b16bd7c80eb946f1cadd7e66c82e998d067) and some previous region comments (https://github.com/iTwin/iTwinUI/pull/1753/commits/002ea81e4a83c7045404ce9fc9a5499b82817abd)

I noticed that it did reduce the amount of produced css. But one reason for that could also be that your suggestion is not supporting `informational` that was previously supported. Should we continue supporting `informational`?

https://github.com/iTwin/iTwinUI/blob/8c5da6e200ecdcebca499baf28c7afa5660c5408/packages/itwinui-css/src/input-container/input-container.scss#L273-L275

packages/itwinui-css/src/input-container/input-container.scss
mayank99: Actually I'm no longer sure what "select with status" should look like. The old demos looked like this:
image

I'm trying to investigate.

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
mayank99: I'm ok with removing `informational` because we don't support it in React anyway

mayank99: **update**: Slightly refactored the selectors in [`3deac83` (#1753)](https://github.com/iTwin/iTwinUI/pull/1753/commits/3deac837b3a5da4b032da5b446ab3d2ddb09e5e4) and left out `informational`

packages/itwinui-css/src/input-container/input-container.scss
mayank99: **Update**: removed `::after` in [`3deac83` (#1753)](https://github.com/iTwin/iTwinUI/pull/1753/commits/3deac837b3a5da4b032da5b446ab3d2ddb09e5e4) since it was doing nothing

the vertical centering is handled using `padding-block-end`

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
```diff
@@ -30,7 +30,7 @@ export type LabeledSelectProps<T> = {
   /**
    * Custom svg icon. Will override status icon if specified.
    */
-  svgIcon?: JSX.Element;
+  svgIcon?: React.ComponentPropsWithoutRef<typeof StatusMessage>['startIcon'];
```
mayank99, line 33: we should deprecate this prop to make maintenance easier.

if the user wants to customize the icon, they should pass a `<StatusMessage>` (instead of a string) into `message`

mayank99: also add a patch changeset

mayank99 (approved):

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
rohan-kadkol: > we should deprecate this prop to make maintenance easier.
> 
> if the user wants to customize the icon, they should pass a `<StatusMessage>` (instead of a string) into `message`

Sounds good, deprecated `svgIcon` in `LabeledSelect` (https://github.com/iTwin/iTwinUI/pull/1753/commits/a7c897e979e1bcaf34fd270fb7a2bdb699f5e401). Also added a unit test to confirm the replacement we suggested works (https://github.com/iTwin/iTwinUI/pull/1753/commits/223c0b01d79e6c99f61e8c9584dc1d4ba0ec9cb1), and tested the deprecated and replacement code in a ([Stackblitz playground](https://stackblitz.com/edit/github-snplr9?file=src%2FApp.tsx)).

When trying to deprecate `svgIcon` from `InputGroup` also, I noticed that passing `<StatusMessage startIcon={…} />` to the message of `<InputGroup status="…" />` shows duplicate status icons (same playground as above). So, I didn't deprecate `svgIcon` in `InputGroup` and thus kept the `svgIcon` [type correction change](https://github.com/iTwin/iTwinUI/pull/1753/files#diff-eb53afa1cec202f8e446277c6f6ed48b21db0f22f498291fa4cfda68557a4b7dL44-R44) there.

> also add a patch changeset

Since this deprecation is not a bug fix, should it be a minor change? Added a minor changeset for now (https://github.com/iTwin/iTwinUI/pull/1753/commits/bd8470e18b522d78a6ca42eb8d5b92bafae1aab3).

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
rohan-kadkol: > I'm ok with removing `informational` because we don't support it in React anyway.
> 
> edit: It's technically a breaking change in CSS, but it's very minor and I see it more like correcting a "mistake".

Sounds good, added a changeset for leaving out `informational` (https://github.com/iTwin/iTwinUI/pull/1753/commits/3e55bb85bd6d8db658946d46c47d3761fb371696)

packages/itwinui-react/src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -110,8 +109,8 @@ export const LabeledInput = React.forwardRef((props, ref) => {
       )}
 
       <Box
-        className={cx('iui-input-with-icon', inputWrapperProps?.className)}
         {...inputWrapperProps}
+        className={cx('iui-input-with-icon', inputWrapperProps?.className)}
```
mayank99, line 113: nit: This should already be handled in #1777, so maybe revert it to prevent conflicts (plus it's unrelated to this PR).

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
mayank99: > deprecated `svgIcon` in `LabeledSelect` ([a7c897e](https://github.com/iTwin/iTwinUI/commit/a7c897e979e1bcaf34fd270fb7a2bdb699f5e401)). Also added a unit test to confirm the replacement we suggested works ([223c0b0](https://github.com/iTwin/iTwinUI/commit/223c0b01d79e6c99f61e8c9584dc1d4ba0ec9cb1))

The unit test is not needed, because it is testing implementation details of another component. We don't want this test to break if `StatusMessage` changes in the future.

We already tested in [L56](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.test.tsx#L55) that `message` accepts a react node. So anything passed there works unchanged. And we test for `startIcon` in [`StatusMessage.test.tsx`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/StatusMessage/StatusMessage.test.tsx#L40), so this entire scenario is already covered.

> When trying to deprecate `svgIcon` from `InputGroup` also, I noticed that passing `<StatusMessage startIcon={…} />` to the message of `<InputGroup status="…" />` shows duplicate status icons (same playground as above). So, I didn't deprecate `svgIcon` in `InputGroup` and thus kept the `svgIcon` [type correction change](https://github.com/iTwin/iTwinUI/pull/1753/files#diff-eb53afa1cec202f8e446277c6f6ed48b21db0f22f498291fa4cfda68557a4b7dL44-R44) there.

This is a problem. I think we should correctly support passing `string` vs `ReactNode` into `message`. Currently it will put `<StatusMessage>` inside another `<StatusMessage>` if `message={<StatusMessage>}` is used. Let's try to tackle this in a separate PR in the near future. Ideally it would behave similarly to `LabeledSelect`.

> Since this deprecation is not a bug fix, should it be a minor change? Added a minor changeset for now ([bd8470e](https://github.com/iTwin/iTwinUI/commit/bd8470e18b522d78a6ca42eb8d5b92bafae1aab3)).

`minor` is meant for new features, which this is not, so I would go with `patch`. The `svgIcon` prop still works exactly the same as before, so this is not even a significant change.

PR #1751, opened 2023-12-26
refresh site colors

mayank99:
## Changes

Small visual refresh of the website with a focus on making the colors feel more sophisticated.
- Darker, less blue backgrounds, and fewer gradients.
  - The tokens are set up with a `themeVar` Sass mixin, to enable easier creation of light theme in a future PR.
- More readable code snippets with adjusted font size and a more neutral [`github-dark`](https://vscodethemes.com/e/github.github-vscode-theme/github-dark?language=javascript) theme.
- Adjusted fluid typography to work better on smaller widths.
- Isolated demos using [`revert-layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/revert-layer). Website styles will no longer affect component demos.
- Postprocessing using [`lightningcss`](https://lightningcss.dev/). (Previously we weren't adding vendor prefixes, etc.)
- Small minor fixes here and there.

Still lots more to improve (I wanted to play with text colors) but running out of time. Always can be improved in future PRs.

## Testing

See [deploy preview](https://deploy-preview-1751--itwinui-previews.netlify.app/).

## Docs

N/A. No package changes.

Issue #1571, opened 2023-09-13
Maintenance


gretanausedaite:
Ongoing issue with maintenance tasks we want to do in our repository. Feel free to pick any of them any time.

- [ ] Update stylelint.
    - [x] #1754.
    - [x] #1591
- [x] #864
- [x] #1669
- [x] fix jest source maps (or migrate to [vitest](https://vitest.dev/)) #1784
    - [x] Type check unit test files. (#1561) also handled in [#1784](https://github.com/iTwin/iTwinUI/pull/1784).
- [x] #1872
- [x] set up website deploy previews
- [ ] Show dev-ony warnings for common errors (for example, input missing label)
- [x] #1807
- [x] #1933 
- [ ] replace styleguide with custom lint rules
- [x] #1918 
- [ ] Try removing `InputFlexContainer.tsx` in favor of directly using `InputWithDecorations` (https://github.com/iTwin/iTwinUI/pull/1881#discussion_r1506186586)
- [ ] Update `@types/react` to latest version and fix our component types
- [x] #1920
- [ ] Support React 19

PR #1754, opened 2023-12-27
replace stylelint formatting rules with prettier

mayank99:
## Changes

part of #1571

- `prettier` is now used for formatting all scss files, both in the editor and in the pre-commit hook.
	- ran `prettier` on all scss files to avoid random annoyances in future PRs.
- `stylelint` is only used for non-stylistic linting.
- .scss files are now allowed in `website/` (needed for #1751), with unnecessary rules disabled.

## Testing

tested that linting and formatting works in both IDE and precommit hooks. 

## Docs

n/a

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,26 +14,14 @@
   "html.format.wrapLineLength": 120,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.stylelint": true
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.stylelint": "explicit"
   },
   "[html]": {
     "editor.defaultFormatter": "vscode.html-language-features"
   },
-  "[css]": {
-    "editor.formatOnSave": false
-  },
-  "[scss]": {
-    "editor.formatOnSave": false
-  },
   "[markdown]": {
     "editor.defaultFormatter": "vscode.markdown-language-features"
-  },
-  "[astro]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "[javascript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "javascript.format.semicolons": "insert",
   "javascript.preferences.quoteStyle": "single",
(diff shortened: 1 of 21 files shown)


apps/website/.stylelintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["configs/.stylelintrc.json"],
```
rohan-kadkol, line 2: Is this path correct? What are these paths relative to? If it if relative to the current folder, then `apps/website/configs/.stylelintrc.json` doesn't exist?

.vscode/settings.json
```diff
@@ -14,27 +14,15 @@
   "html.format.wrapLineLength": 120,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.fixAll.stylelint": true
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.stylelint": "explicit"
```
rohan-kadkol, line 18: What is the difference between `true` and `"explicit"`? Why is `"explicit"` needed in this case?

internal/configs/.stylelintrc.json
```diff
@@ -1,35 +1,28 @@
 {
-  "extends": ["stylelint-config-sass-guidelines"],
+  "extends": ["stylelint-config-sass-guidelines", "stylelint-config-prettier"],
   "plugins": ["stylelint-use-logical"],
   "rules": {
-    "color-hex-case": "upper",
```
rohan-kadkol, line 5: Just curious, how did you determine/decide which rules are not needed anymore?

apps/website/.stylelintrc.json
mayank99: `configs` is the name of the `internal/configs` workspace.

check other `.stylelinrc.json` files

.vscode/settings.json
mayank99: i actually didn't make this change (it happened automatically), but `"explicit"` seems to be the correct value.

this is the only documentation i could find: https://github.com/microsoft/vscode-eslint#version-204

internal/configs/.stylelintrc.json
mayank99: i removed formatting-related rules (which are now [deprecated](https://stylelint.io/migration-guide/to-15#deprecated-stylistic-rules)), because prettier will take care of formatting.

rohan-kadkol (approved):

PR #1755, opened 2023-12-28
Improve `Keyboard Keys` docs.

siddhantrawal:
## Changes

Updated documentation for the Keyboard Key component.
Renamed the component from `Keyboard Keys` to `Kbd`.

### DIFF
--- a/apps/website/src/content/docs/kbd.mdx
+++ b/apps/website/src/content/docs/kbd.mdx
@@ -20,7 +20,7 @@ You can pass custom key(s) as `children`.
 <Kbd>A</Kbd>
 ```
 
-You can also use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys.
+You can also use the `KbdKeys` object which includes a convenient set of constants for common modifier keys.
 
 ```tsx
 <Kbd>{KbdKeys.Enter}</Kbd>


rohan-kadkol (commented):
> Should I also explain what is `<kbd>`

Maybe just mentioning that the default DOM element (default `as` prop) is `<kbd>` and having that link to the [MDNDocs for <kbd>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) would be good.

> … and mention all `KbdKeys`?

I think just a brief mention of what is `KbdKeys` and some example keys in this is sufficient. Listing all the keys would likely be verbose.

> Also, is it a good idea to add an accessibility section mentioning about "Each command key has a title attribute that describes the action that the key performs."

The [`title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) prop is actually a standard HTML prop and not one we created. So maybe no need to explain this too much.

But I'm a little curious about what screen readers do when there is a special symbol in the keyboard key. E.g. `KbdKeys.Apple` or `KbdKeys.Down`. I tried using VoiceOver to read out the below sentence:

```tsx
<p>
  Use the <Kbd>{KbdKeys.Down}</Kbd> or the <Kbd>{KbdKeys.Apple}</Kbd> keys to access the panel
</p>
```

VoiceOver announced it as: "Use the down arrow or the apple keys to access the panel". To confirm, on your Windows systems, how does NVDA read this sentence?

rohan-kadkol (commented):
I left a few suggestions. But you might need to modify the suggestions a bit or format it differently to better fit in the document.

apps/website/src/content/docs/keyboardkeys.mdx
rohan-kadkol, line 1: Continuing the conversation from https://github.com/iTwin/iTwinUI/pull/1755#issuecomment-1871413426.

> I tried using the NVDA, not sure if i used it correctly but it is just reading the sentence as it is like "slash less kbd greater kbdkeys dot down ..."

I guess you may not have rendered it correctly. So it's reading the "slash" and "dot" instead of the actual values it represents. (i.e. `KbdKeys dot Enter` (wrong) vs. `\u21b5 Enter` (correct))

You can try pasting that `<p>` tag in `Keyboard.main.tsx`, going to the docs page for Keyboard keys, and then using NVDA to readout the contents of the code example.

rohan-kadkol (commented):
> Done (0a3c3d1). Does it make sense to have it under `Usage Section`?

Yes, that works.

---

No more comments as of now apart from https://github.com/iTwin/iTwinUI/pull/1755/files#r1437840473

apps/website/src/content/docs/keyboardkeys.mdx
```diff
@@ -1,18 +1,27 @@
...
 
-<Placeholder componentPageName='keyboard-key' />
-
 <LiveExample src='Keyboard.main.tsx' truncate={false}>
   <AllExamples.KeyboardMainExample client:load />
 </LiveExample>
 
+## Usage
+
+By default, the `<Kbd>` component renders as a [`<kbd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) HTML element. You can either pass custom key(s) as the `children` or use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys such as `KbdKeys.Command`, `KbdKeys.Enter`, and `KbdKeys.Right`.
```
mayank99, line 15: For readability, we should split the `KbdKeys` part into a separate paragraph and possibly list out all values available.

apps/website/src/content/docs/keyboardkeys.mdx
```diff
@@ -1,18 +1,27 @@
 ---
 title: Keyboard keys
```
mayank99, line 2: The grammatically correct name would be "Keyboard key" not "Keyboard keys".

But the component is called `Kbd` so I'm thinking we should rename the file and title to use "Kbd" as well.

apps/website/src/content/docs/Kbd.mdx
rohan-kadkol, line 1: I believe the file name should be all lower-case `kbd.mdx`. This also fixes the order of the component in the left side nav.

image

apps/website/src/content/docs/Kbd.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+<LiveExample src='Kbd.main.tsx' truncate={false}>
+  <AllExamples.KbdMainExample client:load />
+</LiveExample>
+
+## Usage
+
+The `<Kbd>` component is a styled wrapper around the [`<kbd>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).
+
+You can either pass custom key(s) as the `children` or use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys such as:
```
rohan-kadkol, line 17: Since you are listing out all values, "such as" will not be needed.

```suggestion
You can either pass custom key(s) as the `children` or use the `KbdKeys` enum which includes some of the common keyboard keys:
```

apps/website/src/content/docs/keyboardkeys.mdx
mayank99: could you also split the codeblock into two? first showing a basic example and then showing one with `KbdKeys`

mayank99: yes. by "split" i also mean moving them into the proper positions: paragraph, then codeblock which relates to that paragraph, then another paragraph, then another codeblock which relates to that paragraph.

apps/website/src/content/docs/kbd.mdx
```diff
@@ -14,7 +14,23 @@ thumbnail: Keyboard
 
 The `<Kbd>` component is a styled wrapper around the [`<kbd>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).
 
-You can either pass custom key(s) as the `children` or use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys such as:
+### Basic
+
+You can pass custom key(s) as the `children` parameter. Here's an example of how to use it:
```
rohan-kadkol, line 19: nit: Brevity

```suggestion
You can pass custom key(s) as the `children` parameter.
```

apps/website/src/content/docs/kbd.mdx
```diff
@@ -14,7 +14,23 @@ thumbnail: Keyboard
...
+
+```tsx
+<Kbd>A</Kbd>
+```
+
+### Predefined Key
+
+You can also use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys. Here's what usage looks like:
```
rohan-kadkol, line 27: nit: Brevity

```suggestion
You can also use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys.
```

apps/website/src/content/docs/keyboardkeys.mdx
rohan-kadkol: @mayank99 Thoughts? Does this component have any accessibility considerations?

Is something like an `aria-label` needed for keyboard keys like the apple logo that is not announced on NVDA + Windows?

apps/website/src/content/docs/kbd.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+<LiveExample src='Kbd.main.tsx' truncate={false}>
+  <AllExamples.KbdMainExample client:load />
+</LiveExample>
+
+## Usage
+
+The `<Kbd>` component is a styled wrapper around the [`<kbd>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).
+
+### Basic
```
mayank99, line 17: no need to complicate with extra sub-headings imo

apps/website/src/content/docs/kbd.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+You can also use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys.
+
+```tsx
+<Kbd>{KbdKeys.Enter}</Kbd>
+```
+
+List of supported keys:
+
+- `KbdKeys.Command`
```
mayank99, line 35: would be nice to also show the a preview of the characters. e.g.:

> - `KbdKeys.Command` (⌘ Cmd)
> - `KbdKeys.Shift` (⇧ Shift)
> …

apps/website/src/content/docs/kbd.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: Kbd
+description: This component is used to display which key or combination of keys performs a given action.
+thumbnail: Keyboard
```
mayank99, line 4: should we also rename the thumbnail?

examples/Kbd.main.tsx
mayank99, line 1: The main example could be improved to show a realistic use case?

> Press <kbd>⌘ Cmd</kbd> + <kbd>K</kbd> to search.

apps/website/src/content/docs/keyboardkeys.mdx
mayank99: `aria-label` does nothing here. We could add a note that some of the keys are specific to operating system and may not be available universally.

examples/Kbd.main.tsx
mayank99: remove the flex container, and use a `<p>`

and then you also need to add spaces.

```jsx
<p>
  Press <Kbd>{KbdKeys.Command}</Kbd> + <Kbd>K</Kbd> to search.
</p>
```

apps/website/src/content/docs/kbd.mdx
mayank99: no need to touch ladle or anything, it's fine to leave as-is

for the thumbnail, you also need to rename the actual thumbnail inside `website/src/icons/thumbnails`

apps/website/src/content/docs/kbd.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+<LiveExample src='Kbd.main.tsx' truncate={false}>
+  <AllExamples.KbdMainExample client:load />
+</LiveExample>
+
+## Usage
+
+The `<Kbd>` component is a styled wrapper around the [`<kbd>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).
+
+You can pass custom key(s) as the `children` parameter.
```
mayank99, line 17: children is not a parameter

```suggestion
You can pass custom key(s) as `children`.
```

apps/website/src/content/docs/kbd.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```tsx
+<Kbd>A</Kbd>
+```
+
+You can also use the `KbdKeys` enum which includes a convenient set of constants for common keyboard keys.
```
mayank99, line 23: `KbdKeys` is an object, not an enum

```suggestion
You can also use the `KbdKeys` object which includes a convenient set of constants for common modifier keys.
```

mayank99 (approved):

rohan-kadkol (approved):

PR #1757, opened 2023-12-28
Removed occurrences of old `dev` branch from CI/CD

rohan-kadkol:
## Changes

I noticed that the CI and CD workflow files still had a few occurrences of the old `dev` branch. This PR removes those occurrences.

## Testing

Made sure that there are no more occurrences of the dev branch in any `.yml` file.

## Docs

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - dev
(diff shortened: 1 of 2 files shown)


mayank99 (commented):
This is not necessary, i left it there intentionally because it's not getting in the way and would need to be added back for the next major release anyway

PR #1760, opened 2024-01-02
correctly merge ComboBox `inputProps.style`

### LINKED ISSUE #1759: Combobox style input props overrides padding-left in multiselect
#### Description

When I pass a style to input props in combobox component, even if I don't override the padding-left property (it can even be an empty object for the style property) I lose the inline style for padding-left and the text pointer is stuck at the beginning of the input, because of that, when I have selected options I cannot see what I type into the field when trying to search for a value inside my options.

#### Expected behavior

That the styles passed inside styleProps are merged with the styles needed in the control, and not completely overridden. 

#### Reproduction
Link to a minimal repro: [sandbox](https://codesandbox.io/p/sandbox/currying-meadow-ww29c8?file=%2Fsrc%2FApp.tsx%3A34%2C40)

##### Steps to reproduce
1. Click on the combobox
1. Type anything
1. See that the text is behind the selected options

mayank99:
## Changes

fixes #1759

## Testing

tested in the playground with this code:

```tsx
import { ComboBox, Label, InputGrid } from '@itwin/itwinui-react';

export default () => {
  const options = React.useMemo(
    () => [
      { label: 'Apartments', value: 'apartments' },
      { label: 'Houses', value: 'houses' },
      { label: 'Lofts', value: 'lofts' },
    ],
    [],
  );

  const [selectedOptions, setSelectedOptions] = React.useState(['lofts']);

  return (
    <InputGrid>
      <Label htmlFor='housing-input'>Housing type</Label>
      <ComboBox
        options={options}
        inputProps={{
          id: 'housing-input',
          placeholder: 'Housing type',
          style: { backgroundColor: 'rebeccapurple' },
        }}
        multiple
        value={selectedOptions}
        onChange={(selected) => {
          setSelectedOptions(selected);
        }}
      />
    </InputGrid>
  );
};
```

## Docs

added changeset


packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -224,7 +224,10 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
         spellCheck={false}
         autoCapitalize='none'
         autoCorrect='off'
-        style={multiple ? { paddingInlineStart: tagContainerWidth + 18 } : {}}
+        style={{
+          ...(multiple && { paddingInlineStart: tagContainerWidth + 18 }),
```
rohan-kadkol, line 228: Although it works even when `multiple` is `false`, is there any potential issue with deconstructing `false`? i.e. `{ ...false, otherKey: otherValue}`. Would it be safer to do something like:

```suggestion
          ...(multiple ? { paddingInlineStart: tagContainerWidth + 18 } : {}),
```

mayank99: should be fine

image

rohan-kadkol: It works, but I was just confirming if this pattern of deconstruction is alright/standard. When I had tested it on PlayCode, it had shown an error/warning:

Bildschirmfoto 2024-01-02 um 12 27 36 PM

But yes, since it works, I think it is alright 👍

rohan-kadkol (approved):

packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
rohan-kadkol: I think this is a TS error that's coming since I used `false`. But if I use a boolean variable that can be `true`, the TS error is gone.

mayank99: i can still change it if you want, i'm fine either way

rohan-kadkol: After thinking about it, I feel your shorthand is better. I was just a bit surprised destructing `false` worked when I first saw it. But now that we confirmed it works, I feel your shorthand is cleaner.

PR #1762, opened 2024-01-02
add `LinkAction` docs

mayank99:
## Changes

Added new documentation page for `LinkAction`.

This might seem like an insignificant utility at first, but I've found lots of scenarios where it can be helpful for third-party components, so we should advertise it some more.

Updated the existing docs in `Tile` and `ListItem` to link to this new page.

## Testing

N/A

## Docs

These _are_ docs!


examples/LinkAction.main.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { LinkAction, LinkBox, Surface } from '@itwin/itwinui-react';
+
+export default () => {
+  return (
+    <LinkBox>
+      <Surface>
+        <Surface.Header>
+          <LinkAction href='https://www.example.com'>Stadium</LinkAction>
```
mayank99, line 13: i don't like this example, it does not even look clickable.

accepting suggestions for better examples

apps/website/src/content/docs/linkaction.mdx
```diff
@@ -0,0 +1,42 @@
...
+</LinkBox>
+```
+
+Similar to the [`Anchor`](anchor) component, `LinkAction` supports the polymorphic `as` prop, which can be used to render a different link component (e.g. from a client-side router) or a `"button"` (e.g. when you want to trigger an action within the same page, rather than a page navigation).
+
+## Technical explanation
+
+The implementation of this component is based on [the pseudo-content trick](https://inclusive-components.design/cards/#thepseudocontenttrick), which makes the `LinkAction`'s `::after` pseudo-element expand to fill up all the available space of the `LinkBox` component. A side-effect of this is that the text inside the `LinkBox` will not be selectable, but this is usually not a problem.
+
+If your container includes other elements that should be clickable, you can set `position: relative` on them or give them a higher z-index.
```
mayank99, line 38: i think we should try to automatically handle this; it would be nice if the user didn't have to think about z-index

examples/LinkAction.main.tsx
mayank99: to clarify, i don't like that the surface does not look like something you would click on (it looks like static information). simply changing the title/description would not help.

apps/website/src/content/docs/linkaction.mdx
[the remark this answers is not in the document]
mayank99: i was going to add an example, but i'm thinking we should first tackle https://github.com/iTwin/iTwinUI/pull/1762#discussion_r1440691252

examples/LinkAction.main.tsx
rohan-kadkol: Maybe adding a label above asking the user to select one of the two surfaces would help. I feel the surface UI would look clickable if the title/description looks like an option to select.

mayank99: Yeah that's an option but it's too much code. I was hoping there could be a simple, self-contained example

apps/website/src/content/docs/linkaction.mdx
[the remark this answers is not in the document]
mayank99: i was trying to implement it, but having trouble handling it out-of-the-box. we might need to expose another component to handle it.

for now, i've added a static example: [`3e459f1`](https://github.com/iTwin/iTwinUI/pull/1762/commits/3e459f12547c16b6d9b8cafea6d0440edd94506c)

when/if we add the new component, we can show a live example

examples/LinkAction.main.tsx
mayank99: Let's punt on this. To increase affordance, I will use `<LinkAction as={Anchor}>` after we fix the ordering issue.

mayank99: LinkAction also does not make sense by itself. The whole point of this component is to increase the tap target size, which usually means it should be used within a container that has additional content (such as a card).

apps/website/src/content/docs/tile.mdx
```diff
@@ -54,20 +54,24 @@ A list of all available props can be found [at the end](#props) of this page.
...
 </LiveExample>
 
-Tile can also link to another page. Add `href` prop to the Tile.Action to make whole tile an anchor link. This is because manually wrapping the Tile with an `<a href="...">` is not recommended for accessibility reasons.
+Actionable tiles can also link to another page by passing the `href` prop to `Tile.Action`.
 
 <LiveExample src='Tile.anchorlink.tsx'>
   <AllExamples.TileAnchorLinkExample client:load style={{ padding: 50 }} />
 </LiveExample>
 
+**Important**: Adding `onClick` directly to the `Tile` or manually wrapping the Tile with an `<a href="...">` is not recommended for accessibility reasons.
```
rohan-kadkol, line 71: Might be better to add the [link](https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html) that you added in the linkaction docs. That way, the "accessibility reasons" are explained

apps/website/src/content/docs/linkaction.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+This component is useful in a wide range of situations when you need to increase the clickable area of an element (e.g. a card, a list item, etc.) without compromising semantics. This means you can keep the link text concise and still make the entire area clickable, even if it includes other elements or extra padding.
+
+`LinkAction` is used internally in [`Tile`](tile#actionable) and [`ListItem`](list#with-linkaction) components, and is also available for use in external components.
+
+## Usage
+
+This component consists of two parts:
+
+1. `LinkBox`: A wrapper component that makes the entire area clickable.
```
rohan-kadkol, line 22: Can `LinkAction` be used without a `LinkBox`? If so, should that be documented?

mayank99: technically yes if the container uses `position: relative`, but i would feel more comfortable if users added `LinkBox` anyway

the only exception is within our other components like `Tile` and `ListItem`, but i'm thinking we should have wrappers for all of those (e.g. `ListItem.Action`, similar to `Tile.Action`)

rohan-kadkol (approved):

apps/website/src/content/docs/linkaction.mdx
mayank99: Technically it can if the container has `position: relative`, but I would feel more comfortable if `LinkBox` was always used. Gives us the ability to add more logic to it in the future

PR #1763, opened 2024-01-03
use a proxy for styles object to reduce bundle size

mayank99:
## Changes

Replaced the [`styles` object](https://unpkg.com/browse/@itwin/itwinui-react@3.2.1/esm/styles.js) with a proxy. The giant object is not needed, since we are not using a proper hash in our css modules scoping. The proxy simply transforms the class names at runtime.

In the future if we ever use proper hashing, we can bring back the object.

Note: I had to enable minification because we are now relying on a side-effect import to include all the styles. Without enabling minification, Vite will not tree-shake the giant object even if it's not exported or used anywhere.

## Testing

Everything works the same. All tests pass.

Verified that the bundle size is actually reduced. The difference is ~20KB (or ~4KB compressed).

| Before | After |
| --- | --- |
| 71KB | 67KB |

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/styles.js/classes.mjs
+++ b/packages/itwinui-react/src/styles.js/classes.mjs
@@ -2,6 +2,18 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import styles from './styles.module.css';
+import './styles.module.css';
 
-export default styles;
+export default new Proxy(
+  {},
+  {
+    get(_, prop) {
+      if (typeof prop === 'string' && prop.startsWith('iui-')) {
+        return prop.replace('iui-', '_iui3-');
+      }
+    },
+    has(_, prop) {
+      return typeof prop === 'string' && prop.startsWith('iui-');
+    },
+  },
+);
--- a/packages/itwinui-react/src/styles.js/classes.mjs
+++ b/packages/itwinui-react/src/styles.js/classes.mjs
@@ -2,8 +2,12 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
+// Side-effect import to ensure CSS file is generated even though the classes are not used.
 import './styles.module.css';
 
+// This proxy converts `iui-` to `_iui3-` in class names with very little code.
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/styles.js/classes.mjs
```diff
@@ -2,6 +2,22 @@
...
+export default new Proxy(
+  {},
+  {
+    get(_, prop) {
+      if (typeof prop === 'string' && prop.startsWith('iui-')) {
+        return prop.replace('iui-', '_iui3-');
+      }
+    },
+    has(_, prop) {
+      return typeof prop === 'string' && prop.startsWith('iui-');
```
rohan-kadkol, line 20: Will `iui-foobar` also return `true` since it startsWith `iui-`?

packages/itwinui-react/src/styles.js/classes.mjs
```diff
@@ -2,6 +2,22 @@
...
-import styles from './styles.module.css';
 
-export default styles;
+// Side-effect import to ensure CSS file is generated even though the classes are not used.
+import './styles.module.css';
+
+// This proxy converts `iui-` to `_iui3-` in class names with very little code.
+// This is more efficient than exporting the entire mapping of original-to-transformed classes.
+export default new Proxy(
+  {},
```
rohan-kadkol, line 12: I'm a bit confused. How is this object populated if it starts off as empty?

packages/itwinui-react/src/styles.js/classes.mjs
mayank99: yes

packages/itwinui-react/src/styles.js/classes.mjs
mayank99: the object is never populated. when you try to run `styles['whatever']` it simply runs the `get` function

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

packages/itwinui-react/src/styles.js/classes.mjs
rohan-kadkol: Is this alright? If so, why?

mayank99: it's fine. the `iui-` prefix belongs to us, we will never call it with unexpected strings

the only way this gets used is through the `polymorphic` utility (which is used in every component). we transform anything passed into `className`; this could include `iui-foorbar` passed by user, but they should not be doing that.

the only way to "fix" this would be to revert to the approach before this PR, which negates any bundle size benefits.

rohan-kadkol (approved):

PR #1764, opened 2024-01-03
v2: fix submenus in popout windows

mayank99:
## Changes

in v2, items within a submenu were not reachable when used in a popout window: https://stackblitz.com/edit/github-1uc5ov-knsff9?file=src%2FApp.tsx

https://github.com/iTwin/iTwinUI/assets/9084735/ce1a8bd1-a672-493c-a453-661788953bea

this was because of the `instanceof Node` check, which uses `Node` from the original window.

we don't really care about this check (the `contains()` check is the more important part) so i've replaced it with `try`/`catch`, which ensures that we don't throw if `contains()` is ever called on `null` or a non-`Node` value.

## Testing

tested by recreating the sandbox in vite playground:

https://github.com/iTwin/iTwinUI/assets/9084735/3a923e88-f612-410f-adb5-a70bc6157a28

## Docs

added changeset


rohan-kadkol (approved):
The vite playground's popup showed just a white screen for me. But looking at your screen recordings, looks like it works on your end. So LGTM.

rohan-kadkol (approved):
> I was having the same problem. Try opening it in another window inside the playground (there should be an option). It will work.

Thanks 👍 Actually, sorry for the confusion, because by vite playground, I meant to say the Stackblitz playground was not working for me. Maybe popups don't work in Stackblitz.

But after trying it in the local vite playground, I could reproduce the before and after.

Issue #1767, opened 2024-01-04
`Select` does not allow to specify a generic argument


GerardasB:
#### Describe the bug (actual behavior)

Rendering a select with a generic argument `<Select<string> />` results in `Expected 0 type arguments, but got 1.ts(2558)` error.
Additionally, the implicit value type is `unknown`, resulting in `(_: unknown) => void | undefined` type for `onChange` prop.

#### Expected behavior
Allow specifying a generic argument for the `Select` component.

#### Additional information
[Version 3.2.1](https://www.npmjs.com/package/@itwin/itwinui-react/v/3.2.1)

```
// Select.d.ts
export declare const Select: React.ForwardRefExoticComponent<SelectProps<unknown> & React.RefAttributes<HTMLElement>>;
```

mayank99:
Repro: https://stackblitz.com/edit/github-nubvvd?file=src%2FApp.tsx

This might have been caused by https://github.com/iTwin/iTwinUI/pull/1615. Need to investigate further.

mayank99:
Yes, we need to fix the typescript issue while still forwarding the ref.

mayank99:
fixed in 3.2.4

PR #1774, opened 2024-01-08
Typecasting `forwardRef` inside Select

siddhantrawal:
## Changes

Fixes: #1767 

Added type for forwardRef inside `Select` and `LabeledSelect`  component (Refer to option 1 inside this [link](https://fettblog.eu/typescript-react-generic-forward-refs/)).

Added Type checking test case for both components.

 Added unit test that forwards the ref Inside the components.

For reviewing, turn off whitespace diff.

## Testing

Tested locally in the Vite Playground. No longer getting the error mentioned in #1767.

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.types-test.tsx
+++ b/packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.types-test.tsx
@@ -85,22 +85,3 @@ import LabeledSelect from './LabeledSelect.js';
     </>
   );
 };
-
-() => {
-  return (
-    <>
-      <LabeledSelect<string>
-        label='Select Label'
-        options={[
-          // There should be error: TS 2322
-          // @ts-expect-error (TS 2322): Type 'number' is not assignable to type 'string'.
-          { value: 1, label: 'Item #1' },
-        ]}
-        onChange={(value) => {
-          const returnValue: string = value;
-          return returnValue;
-        }}
-      />
-    </>
-  );
-};


packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -228,249 +228,249 @@ export type SelectProps<T> = {
...
-            if (isSelected && !multiple) {
-              el?.scrollIntoView({ block: 'nearest' });
-            }
-          },
-          role: 'option',
-          ...restOption,
-          ...menuItem.props,
-        });
+
+const SelectInnerFn = <T,>(
```
mayank99, line 232: i think you shouldn't need this inner function, you can revert the whole thing and add the type assertion at the end of the original component

```jsx
export const Select = React.forwardRef(
  <T,>(props, forwardedRef) => {
    // …
  }
) as <T>(
  props: SelectProps<T> & { ref?: React.ForwardedRef<HTMLElement> },
) => JSX.Element;
```

packages/itwinui-react/src/core/Select/Select.tsx
mayank99, line 1: LabeledSelect also needs to be fixed

rohan-kadkol (commented):
No additional comments

packages/itwinui-react/src/core/Select/Select.tsx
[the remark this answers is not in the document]
mayank99: can you improve the test?
- add a test for a different type, e.g. `number`
- add a test for implicit types, i.e. when `<string>` or `<number>` is not passed as a generic
- test that the type is actually correct

also in one of the unit tests, ensure that `ref` is forwarded

mayank99 (commented):
can we also make `ComboBox` behave similarly?

packages/itwinui-react/src/core/Select/Select.types-test.tsx
```diff
@@ -7,24 +7,58 @@
...
-          { value: '1', label: '1' },
-          { value: '2', label: '2' },
-          { value: '3', label: '3' },
+          { value: 1, label: 'Option 1' },
+          { value: 2, label: 'Option 2' },
+          { value: 3, label: 'Option 3' },
         ]}
         onChange={(value) => {
-          console.log(value);
+          return value;
```
rohan-kadkol, line 24: To confirm whether value indeed has the correct type, you can use the `as` keyword.

E.g. if we want value to be a `number` in this case, you can have something like:

```jsx
const returnValue: number = value;
return returnValue;
```

That way, if value is not a number, an error would be given.

packages/itwinui-react/src/core/Select/Select.tsx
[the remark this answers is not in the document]
mayank99: missing these two parts:
> - test that the type of `onChange` is actually correct (its `value` must have the same type as the generic)
> 
> also in one of the unit tests, ensure that `ref` is forwarded

packages/itwinui-react/src/core/Select/Select.types-test.tsx
mayank99: is this the right way?

i was thinking something like https://vitest.dev/guide/testing-types.html

mayank99 (commented):
> > can we also make `ComboBox` behave similarly?
> 
> Is it okay to have a new PR for it or should I have it included in this one?

upto you, i'm fine with either

packages/itwinui-react/src/core/Select/Select.types-test.tsx
rohan-kadkol: I think `.types-test.tsx` files are treated as regular tsx files that users would write. And the type checks are happening when running `tsc` and not `jest`.

So, I don't think using any `jest` function would work. Maybe as `assert typeof value === "number"`? Maybe some of the things @siddhantrawal can try.

mayank99: it's not a jest function (that link is from vitest), we would have to implement our own wrappers and then we can use them anywhere

or maybe it's time to migrate to vitest (part of #1571)

rohan-kadkol: Ohh, since we haven't migrated to vitest, by that vitest link, I thought you meant a similar function from jest.

If we want to test types with `vitest` and not using `tsc`, then I think we also need to make changes to the `test` script in `package.json` to run vitest on this file. If so, this may be a good starting point to migrate to vitest.

mayank99: so two tasks for @siddhantrawal:
1. in this PR, use this:
   ```suggestion
             const returnValue: number = value;
             return returnValue;
   ```
2. in a subsequent PR, migrate from `jest` and types tests to `vitest`

packages/itwinui-react/src/core/Select/Select.tsx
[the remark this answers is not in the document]
mayank99: no, this is just testing the types, i'm asking for a unit test that actually forwards the ref and does something with it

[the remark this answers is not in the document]
mayank99: you don't need `toBeDefined` or `tagName` checks (not reliable).

instead you can check for some custom attribute.
```js
const selectRef = React.createRef<HTMLElement>();
render(
  <Select
    options={[{ value: 1, label: 'Option 1' }]}
    ref={selectRef}
    data-select
  />,
);

expect(selectRef?.current).toHaveAttribute('data-select);
```

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.types-test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+() => {
+  return (
+    <>
+      <LabeledSelect<string>
+        label='Select Label'
+        options={[
+          { value: '1', label: 'Item #1' },
+          { value: '2', label: 'Item #2' },
+          { value: '3', label: 'Item #3' },
```
rohan-kadkol, line 59: I think you can also test that once a generic value is passed, the values are forced to be of that type.

E.g. you can provide the generic type of `string` but pass values of `number`. Then, you can add the comment with `@ts-expect-error` above the lines that cause the error. Example:

https://github.com/iTwin/iTwinUI/blob/0ccd7b7b1a8933becdd5c320756c13d773207d94/packages/itwinui-react/src/react-table/react-table.types-test.tsx#L87-L94

mayank99 (approved):
LGTM, just missing changeset

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.types-test.tsx
```diff
@@ -66,3 +66,41 @@ import LabeledSelect from './LabeledSelect.js';
...
+          { value: 1, label: 'Item #1' },
+        ]}
+        onChange={(value) => {
+          const returnValue: string = value;
+          return returnValue;
+        }}
+      />
+    </>
+  );
+};
```
rohan-kadkol, line 106: Is this accidentally the duplicate of the above test?

.changeset/lovely-pears-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Improved generic typing for the options of `Select` and `LabeledSelect`.
```
mayank99, line 5:
```suggestion
Fixed `Select` and `LabeledSelect` to correctly handle generic types.
```

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.types-test.tsx
rohan-kadkol: I mean these [two tests](https://github.com/iTwin/iTwinUI/commit/9c481a7eba447414d1133d44195fffe78f3a88a3#diff-e1f949a6c475a77c78a59a84858d06b14cb56c02d3cd0e358f6edf5e39cc7a4dR70-R106) seem to have identical code.

rohan-kadkol (approved):

mayank99 (commented):
Action items after this PR merge:

> can we also make `ComboBox` behave similarly?

> migrate all unit tests and types tests to `vitest`

PR #1775, opened 2024-01-08
add banner on homepage

mayank99:
## Changes

Added new `HomepageBanner` component for announcing new updates. It's a simple link within an `aside`.

image

For now it just links to the migration guide, but I assume we will want to look to a proper page within the docs in the future.

Modified `RightArrowIcon` to support individually modifying the line and the caret, for hover effect (try hovering yourself, i won't spoil the effect).

## Testing

See deploy preview: https://deploy-preview-1775--itwinui-previews.netlify.app/

Tested on narrow widths as well.

## Docs

N/A

### DIFF
--- a/apps/website/src/components/HomepageBanner.astro
+++ b/apps/website/src/components/HomepageBanner.astro
@@ -1,0 +1,63 @@
+---
+import RightArrowIcon from '~/icons/RightArrowIcon.astro';
+
+type Props = {
+  href: string;
+  [key: string]: unknown;
+};
+
+const { href, ...props } = Astro.props;
+---
+
+<aside {...props}>
+  <a href={href}>
+    <slot />
+  </a>
+
+  <RightArrowIcon class='arrow' lineProps={{ class: 'line' }} />
+</aside>
+
+<style lang='scss'>
+  aside {
+    border: 1px solid var(--color-line-2);
+    background-color: var(--color-background-1);
+    border-radius: 9e9px;
+    display: inline-flex;
+    align-items: center;
+    color: var(--color-text);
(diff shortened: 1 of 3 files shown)


apps/website/src/components/HomepageBanner.astro
```diff
@@ -0,0 +1,65 @@
...
+
+<aside {...props}>
+  <a href={href}>
+    <slot />
+  </a>
+
+  <RightArrowIcon class='arrow' lineProps={{ class: 'line' }} />
+</aside>
+
+<style lang='scss'>
```
rohan-kadkol, line 20: There seems to be no focus outline when focusing on the banner link

mayank99: yikes! good catch, fixed.
image

note to self: `all: unset` is a terrible hammer

apps/website/src/components/HomepageBanner.astro
```diff
@@ -0,0 +1,70 @@
...
+
+      :global(.line) {
+        opacity: 0;
+        transition: opacity 0.2s ease-in-out;
+      }
+    }
+
+    a {
+      all: unset;
+    }
```
rohan-kadkol, line 43: Is having global styles with a generic name like `line` alright?

rohan-kadkol (approved):

apps/website/src/components/HomepageBanner.astro
mayank99: it's not truly global, it's nested under `aside` which is scoped.

this is because astro cannot scope indirectly passed classes (in this case using `lineProps`)

PR #1777, opened 2024-01-10
automatically associate inputs, labels and messages

mayank99:
## Changes

If a user forgets to connect `Label`/`Input`/`StatusMessage` inside `InputGrid`, we will now handle it for them by cloning `children`.

Since `InputGrid` is also used internally in `LabeledInput`/`LabeledSelect`, I was able to remove some of the explicit id handling in those components.

Lastly, the `StatusMessage` being connected via `aria-describedby` is brand new functionality that we weren't handling before at all, so this is a net improvement for all uses of the `message` prop.

Implementation notes:
- Looping/cloning children is generally [discouraged](https://react.dev/reference/react/Children) and indeed can be quite brittle (breaks if there are wrapper divs anywhere), but it is still a very useful fallback and just a convenient thing.
- It helps ease the v2→v3 migration specifically for the more complex `LabeledInput`→`InputWithDecorations` scenarios (discovered [here](https://github.com/iTwin/iTwinUI-migration-tool/pull/96#discussion_r1446653151)) and especially for React 17 users (who don't have access to `useId`).
- I created an internal `InputWithIcon` component to help with narrowing down children. It's starting to get a little confusing now along with the existing `InputGrid`, `InputContainer`, `InputFlexContainer`, `InputWithDecorations`, `InputGroup` components. We should go through all of them (in a future PR) and try to remove some of them.

## Testing

Made sure existing unit and axe tests pass. Added new unit tests. Also tested a few different permutations of the following code in Vite playground:

```jsx
<InputGrid>
  <Label>Label</Label>
  <Input />
  <StatusMessage>Hint message</StatusMessage>
</InputGrid>
```

Tested the components in screen-readers and the browser's accessibility tree to make sure labels and message are correctly associated and announced.

Works fine in NVDA, but VoiceOver has a bug where it [doesn't automatically announce descriptions on text fields](https://bugs.webkit.org/show_bug.cgi?id=262895#c6).

<details><summary>NVDA 2023.3.1 + Firefox 121</summary>

image

image

</details> 

<details><summary>VoiceOver + Safari 17.0</summary>
<p>

image

image

image

</p>
</details> 

## Docs

Added changesets.

Updated `InputGrid` documentation and jsdocs.

### DIFF
--- a/packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
+++ b/packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
@@ -8,6 +8,11 @@ import * as React from 'react';
 import { InputGrid } from './InputGrid.js';
 import { StatusMessage } from '../StatusMessage/StatusMessage.js';
 import { Label } from '../Label/Label.js';
+import { Input } from '../Input/Input.js';
+import { ComboBox } from '../ComboBox/ComboBox.js';
+import { InputWithDecorations } from '../InputWithDecorations/InputWithDecorations.js';
+import { Textarea } from '../Textarea/Textarea.js';
+import { Select } from '../Select/Select.js';
 
 const assertBaseElement = (container: HTMLElement) => {
   const inputContainer = container.querySelector('.iui-input-grid');
@@ -53,4 +58,58 @@ it('should take class and style', () => {
   ) as HTMLElement;
   expect(inputContainer).toHaveClass('iui-input-grid my-class');
   expect(inputContainer.style.width).toBe('50px');
+});
+
+describe.each([false, true])('should associate label and message', (hasIds) => {
+  it.each`
+    Component                     | name
+    ${Input}                      | ${'Input'}
+    ${Textarea}                   | ${'Textarea'}
+    ${InputWithDecorations.Input} | ${'InputWithDecorations.Input'}
+    ${'input'}                    | ${'input'}
+    ${'textarea'}                 | ${'textarea'}
+    ${'select'}                   | ${'select'}
+  `(`with $name ${hasIds ? '(has user ids)' : ''}`, ({ Component }) => {
+    const { container } = render(
+      <InputGrid>
+        <Label id={hasIds ? 'id1' : undefined}>The label</Label>
+        {Component === InputWithDecorations.Input ? (
+          <InputWithDecorations>
+            <Component id={hasIds ? 'id2' : undefined} />
+          </InputWithDecorations>
+        ) : (
+          <Component id={hasIds ? 'id2' : undefined} />
+        )}
+        <StatusMessage id={hasIds ? 'id3' : undefined}>
+          The message
+        </StatusMessage>
+      </InputGrid>,
+    );
+
+    const input = container.querySelector(
+      'input, textarea, select',
+    ) as HTMLElement;
+    expect(input).toHaveAccessibleName('The label');
+    expect(input).toHaveAccessibleDescription('The message');
+  });
+
+  it.each`
+    Component   | name
+    ${ComboBox} | ${'ComboBox'}
+    ${Select}   | ${'Select'}
+  `(`with $name ${hasIds ? '(has user ids)' : ''}`, ({ Component }) => {
+    const { container } = render(
+      <InputGrid>
+        <Label id={hasIds ? 'id1' : undefined}>The label</Label>
+        <Component id={hasIds ? 'id2' : undefined} options={[]} />
+        <StatusMessage id={hasIds ? 'id3' : undefined}>
+          The message
+        </StatusMessage>
+      </InputGrid>,
+    );
+
+    const combobox = container.querySelector('[role=combobox]') as HTMLElement;
+    expect(combobox).toHaveAccessibleName('The label');
+    expect(combobox).toHaveAccessibleDescription('The message');
+  });
 });
--- a/packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
+++ b/packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
@@ -3,9 +3,21 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Box } from '../utils/index.js';
+import {
+  Box,
+  InputWithIcon,
+  cloneElementWithRef,
+  useId,
+} from '../utils/index.js';
 import cx from 'classnames';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
+import { Label } from '../Label/Label.js';
+import { Input } from '../Input/Input.js';
+import { Textarea } from '../Textarea/Textarea.js';
+import { StatusMessage } from '../StatusMessage/StatusMessage.js';
+import { InputWithDecorations } from '../InputWithDecorations/InputWithDecorations.js';
+import { ComboBox } from '../ComboBox/ComboBox.js';
+import { Select } from '../Select/Select.js';
 
 type InputGridOwnProps = {
   /**
@@ -21,19 +33,28 @@ type InputGridOwnProps = {
 //-------------------------------------------------------------------------------
 
 /**
- * InputGrid component is used to display inputs (input, textarea, select)
+ * InputGrid component is used to display form fields (input, textarea, select)
  * with label and/or status message
  *
- * @usage
+ * Form fields are automatically associated with the label and status message for
+ * better accessibility.
  *
+ * @example
  * <InputGrid>
- *   <Label htmlFor='input-1'>This is label</Label>
- *   <Input id='input-1'/>
- *   <StatusMessage>This is message</StatusMessage>
+ *   <Label>This is a label</Label>
+ *   <Input />
+ *   <StatusMessage>This is a message</StatusMessage>
  * </InputGrid>
  */
 export const InputGrid = React.forwardRef((props, ref) => {
-  const { children, className, labelPlacement = undefined, ...rest } = props;
+  const {
+    children: childrenProp,
+    className,
+    labelPlacement = undefined,
+    ...rest
+  } = props;
+
+  const children = useChildrenWithIds(childrenProp);
 
   return (
     <Box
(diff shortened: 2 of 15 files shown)


packages/itwinui-react/src/core/utils/components/InputWithIcon.tsx
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { polymorphic } from '../index.js';
+
+export const InputWithIcon = polymorphic.div('iui-input-with-icon');
+InputWithIcon.displayName = 'InputWithIcon';
```
rohan-kadkol, line 8: Is it necessary to provide a `displayName` when the name is the same as the component name?

packages/itwinui-react/src/core/utils/components/InputWithIcon.tsx
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { polymorphic } from '../index.js';
+
+export const InputWithIcon = polymorphic.div('iui-input-with-icon');
```
rohan-kadkol, line 7: Should this be marked as `@private`?

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -21,19 +33,28 @@ type InputGridOwnProps = {
 //-------------------------------------------------------------------------------
 
 /**
- * InputGrid component is used to display inputs (input, textarea, select)
+ * InputGrid component is used to display form fields (input, textarea, select)
  * with label and/or status message
  *
- * @usage
+ * Form fields are automatically associated with the label and status message for
```
rohan-kadkol, line 39: Should this mention that if the user passes manual ids, that won't be overriden? Or maybe that can be implied?

mayank99: It's mentioned in the actual documentation

packages/itwinui-react/src/core/utils/components/InputWithIcon.tsx
mayank99: should this comment be marked as "nit"? 😅

packages/itwinui-react/src/core/utils/components/InputWithIcon.tsx
mayank99: Yes, component name is usually not enough especially when using arrow functions and/or `forwardRef`


React devtools will show "ForwardRefComponent" for the following snippet:
```js
const Stuff = React.forwardRef((props, ref) => {});
```

packages/itwinui-react/src/core/utils/components/InputWithIcon.tsx
rohan-kadkol: Since I don't see any changeset saying that `InputWithIcon` is publically exposed, I think this was meant to be a private component. If so, then I think `@private` is important.

mayank99: It's not important because it is not exposed in the barrel file. The user cannot import it.

The whole `@private` thing used to only matter in previous major versions when we were exposing internal sub-paths

rohan-kadkol (commented):
InputGrid related stories can be updated to remove explicit ids. Examples:
* `ComboBox → With Label`
* `InputGrid` stories
* `Label → Inline`

mayank99 (commented):
> InputGrid related stories can be updated to remove explicit ids.

There is no benefit to this, but removed ids from some of them. I kept it in some more complex cases.

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -49,4 +70,207 @@ export const InputGrid = React.forwardRef((props, ref) => {
...
+        return child.props.inputProps?.id || `${idPrefix}--input`;
+      } else if (child.type !== Select) {
+        return child.props.id || `${idPrefix}--input`;
+      }
+    };
+
+    if (child.type === InputWithDecorations || child.type === InputWithIcon) {
+      React.Children.forEach(child.props.children, (child) => {
+        if (isInput(child)) {
+          inputId = findInputId(child);
```
rohan-kadkol, line 172: To be safe, should we stop the `forEach()` after we find the first input?

packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
```diff
@@ -54,3 +59,51 @@ it('should take class and style', () => {
   expect(inputContainer).toHaveClass('iui-input-grid my-class');
   expect(inputContainer.style.width).toBe('50px');
 });
+
+it.each`
```
rohan-kadkol, line 63: Should we also test cases like:
* When user id is given, we should use that instead
* When `<StatusMessage>` is not provided, there should be no description
* `labelId` should also be there when `hasSelect`
* etc.

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -49,4 +70,207 @@ export const InputGrid = React.forwardRef((props, ref) => {
...
+  };
+
+  const cloneInput = (child: React.ReactElement) => {
+    if (child.type === ComboBox) {
+      return cloneElementWithRef(child, (child) => ({
+        inputProps: inputProps(child.props.inputProps),
+      }));
+    }
+
+    if (child.type === Select) {
```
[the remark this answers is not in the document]
mayank99: no, those are just regular form fields so they work with the general input case

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -49,4 +70,207 @@ export const InputGrid = React.forwardRef((props, ref) => {
...
+    if (child.type === Label || child.type === 'label') {
+      hasLabel = true;
+      labelId = child.props.id || `${idPrefix}--label`;
+    }
+
+    if (child.type === StatusMessage) {
+      messageId = child.props.id || `${idPrefix}--message`;
+    }
+
+    const findInputId = (child: React.ReactNode) => {
```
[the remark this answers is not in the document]
mayank99: good catch, moved it out of the loop

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
mayank99: i'm not sure about this; would there ever be more than one input?

there isn't a way to early exit from a `forEach` so i would need to use a `for..of` which requires converting the children to an array (i assume `Children.forEach` has some optimizations that i would be throwing away)

mayank99: actually I realized something:
since i'm no longer using state, it gets reset to `undefined` on every render anyway, so i can just use `||=` to only set it once on every render.

change made: [`0338ecb`](https://github.com/iTwin/iTwinUI/pull/1777/commits/0338ecbcfe395b6d27db731264d9b0cc92a47584)

packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
mayank99: >     * When user id is given, we should use that instead

Added tests for this.  [`b8952de`](https://github.com/iTwin/iTwinUI/pull/1777/commits/b8952dee31b8de3a751dbf9063bef54af20b1bbe)

I also changed the code to handle explicitly-passed `undefined`.

```jsx
<InputGrid>
  <Label id={undefined}>Label</Label>
  <Input id={undefined} />
  <StatusMessage id={undefined}>Hint message</StatusMessage>
</InputGrid>
```

>     * When `<StatusMessage>` is not provided, there should be no description

I don't see why this is necessary. We don't care that there is no description.


>     * `labelId` should also be there when `hasSelect` 
>     * etc.

These are implementation details that should not be tested. We only care about the final result, i.e. accName and accDesc are present.

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
rohan-kadkol: > i'm not sure about this; would there ever be more than one input?

I guess the officially supported case is with only one input, as demonstrated on our "Input with Decorations" docs page. So if more than one input is included for some reason, it's good that we're only considering the first one (after [`0338ecb` (#1777)](https://github.com/iTwin/iTwinUI/pull/1777/commits/0338ecbcfe395b6d27db731264d9b0cc92a47584) & https://github.com/iTwin/iTwinUI/pull/1777/commits/ed62b3e1ee8d4f98316c5f59ab2bf10bc479db3b)

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
```diff
@@ -49,4 +70,207 @@ export const InputGrid = React.forwardRef((props, ref) => {
...
+  React.Children.forEach(children, (child) => {
+    if (!React.isValidElement(child)) {
+      return;
+    }
+
+    if (child.type === Label || child.type === 'label') {
+      hasLabel = true;
+      labelId ||= child.props.id || `${idPrefix}--label`;
+    }
+
```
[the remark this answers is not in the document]
mayank99: what does this have to do with Select?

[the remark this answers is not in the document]
mayank99: that's also unrelated, you might be confusing two different parts:

- L185: We only want `labelId` for `Select`, because it gets referenced in `aria-labelledby` (other inputs use `<Label htmlFor`>)
- L150: We exclude `Select` because its `id` gets set on the wrapping `<div>`. To get its real id, we would have to grab it from `triggerProps`, similar to how we do for ComboBox (L149), but it doesn't matter because we don't even return `inputId` for Select (L186).

packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
rohan-kadkol: > > `labelId` should also be there when `hasSelect`
> > etc.
> 
> These are implementation details that should not be tested. We only care about the final result, i.e. accName and accDesc are present.

I mean a test to confirm that id is not added to `<Label>` unnecessarily. Instead, it is only added when there is a `<Select>`.

mayank99: It still doesn't matter, it's an implementation detail. We do not care if the Label has an id, and the developer using this component certainly does not care either.

I recommend reading this article: https://kentcdodds.com/blog/testing-implementation-details

packages/itwinui-react/src/core/InputGrid/InputGrid.tsx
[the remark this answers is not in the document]
mayank99: added some comments [`ab5c823`](https://github.com/iTwin/iTwinUI/pull/1777/commits/ab5c823f3e3e477b3eef6e0ad6618cf827c6f607)

packages/itwinui-react/src/core/InputGrid/InputGrid.test.tsx
rohan-kadkol: I read that article, but I am not suggesting testing implementation details, like the value of `hasSelect`. I am suggesting having a test to make sure we don't add props when they are not needed.

Even though developers may not care whether an id is added or not, I feel adding it unnecessarily is still incorrect, and thus should be tested that it is not happening.

Basically, since you have written logic to prevent this `id` from being added to `Label` when no `Select` is present, I feel this case/logic should be tested (independent of the implementation details needed to implement this case/logic).

rohan-kadkol: As discussed via chat, we can skip this test for the reasons you mentioned in this thread 👍

rohan-kadkol (approved):
LGTM 🎉

I am about to complete the remaining comments from #1753 (base branch for this PR). So, I think it would be better to merge your PR after mine.

If you feel the same, you might prefer to put your PR back to draft after you get the needed PR approvals to prevent accidentally merging your PR before mine.

rohan-kadkol:
> Some unit test cases are failing.

That would likely be from my base PR #1753. After I fix it in my PR, this will likely be fixed in this PR too.

rohan-kadkol:
I recently merged the PR of the base branch of this PR. So, this PR will now be unblocked 👍

PR #1780, opened 2024-01-11
allow setting ref inside `ComboBox`

siddhantrawal:
## Changes

Forwarded the ref in `ComboBox`. Action Item from this PR #1774.

Added Type and Unit test.

For reviewing, turn off whitespace diff.

### DIFF
--- a/.changeset/smooth-scissors-enjoy.md
+++ b/.changeset/smooth-scissors-enjoy.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': patch
+'@itwin/itwinui-react': minor
 ---
 
-Fixed `ComboBox` to allow setting ref and correctly handling generic types.
+`ComboBox` now forwards its `ref`


packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -165,420 +165,433 @@ const getOptionId = (option: SelectOption<unknown>, idPrefix: string) => {
...
+              popover,
+              show,
+              hide,
+            }}
+          >
+            <ComboBoxInputContainer disabled={inputProps?.disabled} {...rest}>
+              <>
+                <ComboBoxInput
+                  value={inputValue}
+                  ref={forwardedRef}
```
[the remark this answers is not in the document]
rohan-kadkol: nit: you can move this as the first prop (i.e. above `value` on L556)

packages/itwinui-react/src/core/ComboBox/ComboBox.test-types.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *--------------------------------------------------------------------------------------------*/
+/**
+ * This test file tests some ComboBox type related cases that have not been tested in other files
+ * (e.g. react-workshop, unit tests, etc.)
+ */
+
+import React, { useRef } from 'react';
+import ComboBox from './ComboBox.js';
+
+() => {
```
rohan-kadkol, line 13: Do we not need three of these tests, similar to `Select` & `LabeledSelect`?

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: It should be set on the root element, i.e. `ComboBoxInputContainer`. Why are you setting it on `ComboBoxInput`? `ref` should go in the same place as the rest of the top-level props.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -1063,3 +1063,12 @@ it('should be customizable', async () => {
   const statusContent = getByText('My message');
   expect(statusContent).toHaveClass('custom-message-class');
 });
+
+it('should allow passing ref to ComboBox', () => {
+  const selectRef = React.createRef<HTMLElement>();
+  render(
+    <ComboBox options={[{ label: 'Item 1', value: 1 }]} ref={selectRef} />,
+  );
+  console.log(selectRef.current);
```
mayank99, line 1072: please clean up your PR before marking it ready for review
```suggestion
  const comboboxRef = React.createRef<HTMLElement>();
  render(
    <ComboBox options={[{ label: 'Item 1', value: 1 }]} ref={comboboxRef} />,
  );
```

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -1063,3 +1063,12 @@ it('should be customizable', async () => {
...
+  expect(selectRef?.current).toHaveAttribute('role');
```
mayank99, line 1073: use a custom attribute because
1. checking for `role` is not sufficient
2. implementation details should not be tested

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -1063,3 +1063,12 @@ it('should be customizable', async () => {
   const statusContent = getByText('My message');
   expect(statusContent).toHaveClass('custom-message-class');
 });
+
+it('should allow passing ref to ComboBox', () => {
+  const selectRef = React.createRef<HTMLElement>();
+  render(
+    <ComboBox options={[{ label: 'Item 1', value: 1 }]} ref={selectRef} />,
```
mayank99, line 1070: i think it would be useful to also test for `inputProps.ref`

```jsx
<ComboBox
  options={[{ label: 'Item 1', value: 1 }]}
  ref={comboboxRef}
  inputProps={{ ref: inputRef }}
/>,
```

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99, line 1: add changeset before marking PRs ready for review

packages/itwinui-react/tsconfig.test.json
```diff
@@ -5,6 +5,7 @@
     "src/core/Table/**/*",
     "src/react-table/*",
     "src/core/Select/**/*",
-    "src/core/LabeledSelect/**/*"
+    "src/core/LabeledSelect/**/*",
+    "src/core/ComboBox/**/*"
```
mayank99, line 9: nit: you can use a pattern like

```suggestion
    "src/**/*.test-types.tsx"
```

rohan-kadkol: Actually, with the old pattern, the intention was to test all files in that component path, and not just the type tests.

mayank99: so do we plan to keep adding components one-by-one to this list?

rohan-kadkol: Yes, but only for now.

I had actually created a TODO in #1514 (3rd bullet point) as a task to test types in all components. I can create an issue to track that and work on that after all higher priority tasks are complete. After that PR, this include list can just be something like `src/core/**/*`.

mayank99: If you think about it, this PR (and #1774) only adds the `test-types` file, specifically for the purpose of testing the generic types.

If we weren't testing `ComboBox.test.tsx` and `Select.test.tsx` before, why should we do it now? It's equivalent to suddenly testing `Alert.test.tsx`.

Once we migrate to vitest, we can test all our `.test.tsx` files, but for now only `test-types.tsx` files are relevant, so I stand by my original suggestion.

rohan-kadkol: > If we weren't testing `ComboBox.test.tsx` and `Select.test.tsx` before, why should we do it now? It's equivalent to suddenly testing `Alert.test.tsx`.

Existing setup of the repo does not catch type errors in tsx files. So, if one complex usage of the `ComboBox` in a unit test in `ComboBox.test.tsx` triggers a type error but the component works correctly, the `jest` test still will pass. So, we will not realize that the types of `ComboBox` might need to be investigated/fixed.

> Once we migrate to vitest, we can test all our `.test.tsx` files, but for now only `test-types.tsx` files are relevant, so I stand by my original suggestion.

Sure, we can defer type testing all files to when we migrate to `vitest`, but I don't see why `vitest` is a pre-requisite for type testing all files. To clarify, my suggestion is using `tsc` to spot type errors in existing files and creating new type test files that use `jest` or `vitest` helper functions to test the types.

> If you think about it, this PR (and #1774) only adds the `test-types` file, specifically for the purpose of testing the generic types.

Sure, we can test only the `types-test.tsx` files for now. The reason I am suggesting including component paths directly is so we can incrementally proceed to our goal of eventually including all component paths in this tsc test. The other option is to do it all at once later and then fix the type errors that occur.

mayank99: These two points are directly connected:

> we can defer type testing all files to when we migrate to vitest, but I don't see why vitest is a pre-requisite for type testing all files.

> The other option is to do it all at once later and then fix the type errors that occur.

We will need to make changes in lots of test files anyway, so fixing type errors can be handled as part of the same migration.

And just to clarify, I'm suggesting we start doing this migration ***now***. There is no need for a "temporary" or "incremental" stage. It is time to do it the right way.

rohan-kadkol: If we are migrating to `vitest` now, then yes, fixing type errors and then updating the `tsconfig.test.json`'s `include` can come in that PR.

For this PR, we could add one more component to `include` list since anyways, the test fixes are going to happen soon, and thus that list will be shortened. Or we could go ahead with your [suggestion](https://github.com/iTwin/iTwinUI/pull/1780#discussion_r1449235703). I'm okay either way 👍

.changeset/smooth-scissors-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed `ComboBox` to allow setting ref and correctly handling generic types.
```
mayank99, line 5: Should be a minor release

```suggestion
'@itwin/itwinui-react': minor
---

`ComboBox` now forwards its `ref`
```

mayank99 (approved):

rohan-kadkol (approved):

PR #1783, opened 2024-01-15
fix menu item focus outline in firefox

mayank99:
## Changes

Firefox 121 shipped with `:has` support but has a [nasty bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1874042) where `:has` breaks the other selectors used inside `:is` or `:where`. This bug specifically affected `MenuItem` which uses `:has` inside `:where` for focus outlines.

This PR moves the `:has` part out of `:where` which avoids the Firefox bug.

## Testing

Verified that focus outlines appear correctly.

| Before | After |
| --- | --- |
| <p>https://github.com/iTwin/iTwinUI/assets/9084735/1a9ec520-a4fb-415e-9ead-72ed8b203a47</p> | <p>https://github.com/iTwin/iTwinUI/assets/9084735/1a5b63fc-08a1-4132-a2a7-f8fdbfae9974</p> |

Bug can repro'd in live storybook using keyboard ([ComboBox](https://itwin.github.io/iTwinUI/react/?story=combobox--basic))

## Docs

Added changeset.

### DIFF
--- a/.changeset/healthy-windows-whisper.md
+++ b/.changeset/healthy-windows-whisper.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Fixed a Firefox-specific bug where focus outlines were not appearing correctly around menu items (e.g. inside `ComboBox` and `Select`).
--- a/.changeset/healthy-windows-whisper.md
+++ b/.changeset/healthy-windows-whisper.md
@@ -3,4 +3,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Fixed a Firefox-specific bug where focus outlines were not appearing correctly around menu items (e.g. inside `ComboBox` and `Select`).
+Fixed a Firefox-specific bug where focus outlines were not appearing correctly around menu items inside `ComboBox`.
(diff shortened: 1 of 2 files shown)


.changeset/healthy-windows-whisper.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Fixed a Firefox-specific bug where focus outlines were not appearing correctly around menu items (e.g. inside `ComboBox` and `Select`).
```
rohan-kadkol, line 6: Was this issue also in `Select`? When I tested it on Mac + Firefox 121.0, the outlines seemed to show correctly:

https://github.com/iTwin/iTwinUI/assets/45748283/9c882699-5a07-4b94-ba68-36a443a60321

mayank99: I also can't reproduce it now, so I'll change the changelog

```suggestion
Fixed a Firefox-specific bug where focus outlines were not appearing correctly around menu items inside `ComboBox`.
```

rohan-kadkol (approved):

PR #1784, opened 2024-01-16
Migrating all unit tests and types tests to vitest.

siddhantrawal:
Part of #1571 

## Changes

Followed this guide: https://vitest.dev/guide/migration.html#migrating-from-jest

## Testing

All test cases are passing.

## Docs

N/A

## TODOs after PR Merge:

- [x] Handle isJest/isDev issues (Linking to this [convo](https://github.com/iTwin/iTwinUI/pull/1784#discussion_r1469925579)). #1807 
- [ ] Removing `test:types` and writing `type tests` more robustly(Linking to this [convo](https://github.com/iTwin/iTwinUI/pull/1784#discussion_r1469861977)).
- [x] Update `jsdom` to latest version (See [convo](https://github.com/iTwin/iTwinUI/pull/1784#discussion_r1466978438)) #1806

### DIFF
--- a/packages/itwinui-react/src/core/Table/Table.test.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.test.tsx
@@ -2681,9 +2681,9 @@ it('should handle resize with touch', () => {
 it.each(['px', '%', 'rem'])(
   'should handle resize when widths are string units. (E.g. %s)',
   (unit) => {
-    jest
-      .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
-      .mockReturnValue({ width: 100 } as DOMRect);
+    vi.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockReturnValue({
+      width: 100,
+    } as DOMRect);
 
     const columns: Column<TestDataType>[] = [
       {


rohan-kadkol:
> Need some suggestions on what else to change/add? All the test cases are passing, have removed jest and its related dependencies.

I think you can also handle https://github.com/iTwin/iTwinUI/pull/1780#discussion_r1449287359. But since this is already a big PR, you can also handle this in a separate PR.

yarn.lock
rohan-kadkol, line 1: I think the latest `yarn.lock` needs to be pushed. Because when I ran yarn locally, `yarn.lock` had changes.

packages/itwinui-react/package.json
```diff
@@ -99,20 +100,18 @@
     "@types/react-transition-group": "^4.4.10",
     "@typescript-eslint/eslint-plugin": "^5.60.0",
     "@typescript-eslint/parser": "^5.60.0",
+    "@vitest/coverage-v8": "^1.2.1",
     "concurrently": "^5.3.0",
     "eslint": "^8.43.0",
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-require-extensions": "^0.1.3",
-    "jest": "^29.0.0",
```
rohan-kadkol, line 106: After doing a global search for "jest", I still found some occurances. Those will need to be handled too.

.github/workflows/build.yml
```diff
@@ -269,6 +269,8 @@ jobs:
             playgrounds/*/node_modules
           key: modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
 
+      - run: yarn build --filter=@itwin/itwinui-react
```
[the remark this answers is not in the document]
mayank99: Build should not be a prerequisite for tests. The tests run on the source files and are not dependent on the build output. The `styles.css` file is not relevant for tests.

.github/workflows/build.yml
```diff
@@ -378,7 +380,7 @@ jobs:
             playgrounds/*/node_modules
           key: modules-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
 
-      - run: yarn build --filter=itwinui-react
+      - run: yarn build --filter=@itwin/itwinui-react
```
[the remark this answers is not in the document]
mayank99: It's unrelated and unnecessary, undo it.

packages/itwinui-react/package.json
```diff
@@ -77,9 +77,10 @@
     "dev:styles": "yarn build:styles --watch"
   },
   "dependencies": {
-    "@itwin/itwinui-illustrations-react": "^2.1.0",
     "@floating-ui/react": "^0.26.3",
+    "@itwin/itwinui-illustrations-react": "^2.1.0",
     "classnames": "^2.3.2",
+    "jest-environment-jsdom": "^29.7.0",
```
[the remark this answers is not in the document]
mayank99: `jest-environment-jsdom` has nothing to do with `vitest`.

Try using an older version of `jsdom`, maybe `21.0.0`. In a future PR, update it to the latest version and fix any failing tests.

package.json
```diff
@@ -75,7 +75,8 @@
     "semver": "^7.5.2",
     "optionator": "~0.9.3",
     "typescript": "~5.1.6",
-    "@babel/traverse": "^7.23.2"
+    "@babel/traverse": "^7.23.2",
+    "vite": "^5.0.12"
```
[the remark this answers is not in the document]
mayank99: It shouldn't fail, you might have accidentally ended up with a wrong version (maybe bad merge?).

`"resolutions"` is only necessary for forcing a version that would normally be out-of-range (which is not the case for `vite`), so undo this and run `yarn` again.

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+    environment: 'jsdom',
+    setupFiles: 'setupTests.ts',
+    coverage: {
+      provider: 'v8',
+    },
+    exclude: ['/node_modules/', '/esm/', '/cjs/'],
+    reporters: ['junit', 'default'],
+    outputFile: 'coverage/junit.xml',
+    sequence: {
+      hooks: 'list',
```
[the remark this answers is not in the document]
mayank99: Why was it necessary? Some specific test failing?

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,13 +2,13 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-const isJest = typeof jest !== 'undefined';
+const isVitest = typeof vitest !== 'undefined';
```
mayank99, line 5: Undo this. This is not related to our tests, it's part of our runtime code where we actually want to check for `jest`.

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+  },
```
mayank99, line 21: vitest has a [`typecheck`](https://vitest.dev/config/#typecheck) option that we should enable and see how many tests fail. If it's only a few let's fix them right away; if there's more than a few, let's fix them in a separate PR

Also we should remove `test:types` in a future PR

packages/itwinui-react/package.json
```diff
@@ -99,20 +100,18 @@
...
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-require-extensions": "^0.1.3",
-    "jest": "^29.0.0",
-    "jest-cli": "^29.0.0",
-    "jest-environment-jsdom": "^29.0.0",
-    "jest-junit": "^13.2.0",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "ts-jest": "^29.0.0",
     "ts-node": "^10.0.0",
```
mayank99, line 109: Some more dependencies can be removed: `ts-jest`, `ts-node`, `@types/jest`

Also you might want to use the latest version of [`@testing-library/jest-dom`](https://github.com/testing-library/jest-dom), which has support for `vitest`.

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: 'setupTests.ts',
+    coverage: {
+      provider: 'v8',
+    },
```
rohan-kadkol, line 14: Are some coverage options that used to be in `jest.config.cjs` missing here?

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+    exclude: ['/node_modules/', '/esm/', '/cjs/'],
+    reporters: ['junit', 'default'],
+    outputFile: 'coverage/junit.xml',
```
mayank99, line 17: Any reason you chose this specific set of coverage options? Can't we use default `coverage` options?

And have you verified that coverage is working as expected? There are two places we used to show coverage:
- Table displayed when running in the CLI
- A Github check displayed when running in CI

While you're at it, you can also bump the github action `EnricoMi/publish-unit-test-result-action` to the latest version.

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
...
+      {
+        find: /^(.*)\/styles.js$/,
+        replacement: `${__dirname}/src/__mocks__/stylesMock.cjs`,
+      },
+      {
+        find: /^(.*)\/useGlobals.js$/,
+        replacement: `${__dirname}/src/__mocks__/useGlobalsMock.cjs`,
+      },
+    ],
+  },
```
mayank99, line 33: This is quite a roundabout way of adding a mock.

A simpler approach would be to directly use `vi.mock` inside `setupTests.ts`, then remove the `__mocks__` folder.

For example:
```ts
vi.mock('./src/styles.js', () => {
  return {
    default: new Proxy(
      {},
      {
        get: (target, prop) => {
          // instead of returning scoped css modules, we will preserve the original classes
          if (typeof prop === 'string' && prop.startsWith('iui')) {
            return prop;
          }
          return Reflect.get(target, prop);
        },
      },
    ),
  };
});
```

packages/itwinui-react/package.json
```diff
@@ -64,7 +64,7 @@
     "clean:coverage": "rimraf coverage",
     "clean": "rimraf .turbo && yarn clean:coverage && yarn clean:build && rimraf node_modules",
     "test": "yarn test:types && yarn test:unit",
-    "test:unit": "dotenv -v NODE_OPTIONS=--experimental-vm-modules -- npx jest",
+    "test:unit": "dotenv -v NODE_OPTIONS=--experimental-vm-modules -- npx vitest",
```
mayank99, line 67: The experimental vm flag was for jest and can be removed. Also, `vitest` by default starts in watch mode, so `vitest run` should be used instead.

```suggestion
    "test:unit": "vitest run",
```

.github/workflows/build.yml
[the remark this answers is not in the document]
mayank99: Build should not be a prerequisite for tests. The tests run on the source files and are not dependent on the build output. The `styles.css` file is not relevant for tests.

.github/workflows/build.yml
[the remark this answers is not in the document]
mayank99: It's unrelated and unnecessary, undo it.

packages/itwinui-react/package.json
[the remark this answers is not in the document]
mayank99: `jest-environment-jsdom` has nothing to do with `vitest`.

Try using an older version of `jsdom`, maybe `21.0.0`. In a future PR, update it to the latest version and fix any failing tests.

packages/itwinui-react/vitest.config.ts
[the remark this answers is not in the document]
mayank99: Why was it necessary? Some specific test failing?

.github/workflows/build.yml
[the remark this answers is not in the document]
mayank99: Can you try setting [`css`](https://vitest.dev/config/#css) to `false`?

packages/itwinui-react/vitest.config.ts
rohan-kadkol: When I run `yarn workspace @itwin/itwinui-react test:unit`, I don't get the coverage report that we used to get before. Is `coverage: { enabled: true }` required?

image

packages/itwinui-react/tsconfig.json
```diff
@@ -21,6 +21,7 @@
     "skipLibCheck": true,
     "isolatedModules": true,
     "verbatimModuleSyntax": true,
+    "types": ["vitest/globals", "@testing-library/jest-dom/vitest"],
```
rohan-kadkol, line 24: nit: When I saved this file, it also reformated it. To keep the diff small, you can choose to not reformat it. But at least `types` would have been on multiple lines:

```suggestion
    "types": [
      "vitest/globals",
      "@testing-library/jest-dom/vitest"
    ],
```

packages/itwinui-react/vitest.config.ts
rohan-kadkol: The [CI](https://github.com/iTwin/iTwinUI/actions/runs/7670132012/job/20905879601?pr=1784) is failing:

image

packages/itwinui-react/src/core/utils/functions/dev.ts
rohan-kadkol: > Done ([399a837](https://github.com/iTwin/iTwinUI/commit/399a83713910318bbe0086b3ab0e91faf6690861)).

It seems incorrect that `isJest` is checking whether we're in `vitest`. If we want to specifically check for `jest` in our runtime code, we can maybe have both `isJest` and `isVitest`.

> Since I upgraded the [ testing-library/jest-dom](https://github.com/testing-library/jest-dom#with-vitest) to latest version
> and added this import statement `import '@testing-library/jest-dom/vitest'; I am getting the error in the above file as shown in ss.

When I ran `yarn test` in `@itwin/itwinui-react` on the latest commit in this branch (https://github.com/iTwin/iTwinUI/pull/1784/commits/3a51948e0adf4b35cadba7e6f4eeff597a167f25), I didn't get this error.

mayank99: We should have a common reliable check that detects jest, vitest, and mocha

For now, you should be able to work around the error by chaging `typeof `jest` to `typeof globalThis.jest`

packages/itwinui-react/vitest.config.ts
mayank99: Maybe we can change the `test:unit` command from `vitest run` to `vitest run --coverage`.

packages/itwinui-react/vitest.config.ts
mayank99: The path for `useGlobals` is wrong

```diff
- vi.mock('./src/useGlobals.js', () => {
+ vi.mock('./src/core/utils/hooks/useGlobals.js', () => {
```

packages/itwinui-react/vitest.config.ts
```diff
@@ -2,15 +2,16 @@
...
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: 'setupTests.ts',
+    exclude: ['/node_modules/', '/esm/', '/cjs/'],
+    reporters: ['junit', 'default'],
+    outputFile: 'coverage/junit.xml',
+    css: false,
   },
```
mayank99, line 16: Continuing conversation from https://github.com/iTwin/iTwinUI/pull/1784#discussion_r1467022585

Lets try using a vite plugin:

```suggestion
  },
  plugins: [
    {
      name: 'mock-css',
      resolveId(id) {
        if (id.endsWith('styles.css')) {
          return id;
        }
        return null;
      },
      load(id) {
        if (id.endsWith('styles.css')) {
          return '';
        }
        return null;
      },
    },
  ],
```

.github/workflows/build.yml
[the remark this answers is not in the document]
mayank99: Continuing conversation here https://github.com/iTwin/iTwinUI/pull/1784/files#r1467933396

packages/itwinui-react/vitest.config.ts
mayank99: > Added coverage flag as ([36c16d7](https://github.com/iTwin/iTwinUI/commit/36c16d797e98f5f9ba8b5ebebf76ff9c9f2b2f91)) but in the [CI](https://github.com/iTwin/iTwinUI/actions/runs/7672013541/job/20911611358?pr=1784), it was getting added twice like this `vitest run --coverage --coverage`. So I added `coverage.enabled=true` and it works the same, so keeping this.

Why not remove `--coverage` from the github workflow then? Using the `--coverage` feels more correct than the `coverage.enabled` config option, because we are opting into it rather than always having it enabled.

Also still waiting for this:
 
> While you're at it, you can also bump the github action `EnricoMi/publish-unit-test-result-action` to the latest version.

packages/itwinui-react/vitest.config.ts
mayank99: Yeah in that case, I'd say bring back what you had with aliases, except use [`test.alias`](https://vitest.dev/config/#alias) instead of `resolve`

Also, I think it's only needed for `styles.js` (meaning `useGlobals` might be fine as just `vi.mock`).

Lastly, the `__mocks__` folder would make more sense outside of `src` and let's use the name `styles.js` (instead of `stylesMock.cjs`) so that it's ESM and matches the name of the thing we are mocking.

Putting it altogether:
```js
// vitest.config.ts
import { defineConfig } from 'vitest/config';
import * as path from 'node:path';

export default defineConfig({
  test: {
    // ...
    alias: [
      {
        find: /^(.*)\/styles.js$/,
        replacement: path.resolve('./__mocks__/styles.js'),
      },
    ],
  },
  // ...
});
```

```js
// __mocks__/styles.js
export default new Proxy(
  {},
  {
    get: (target, prop) => {
      // instead of returning scoped css modules, we will preserve the original classes
      if (typeof prop === 'string' && prop.startsWith('iui')) {
        return prop;
      }
      return Reflect.get(target, prop);
    },
  },
);
```

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99: Yeah, ts-ignore is fine here

packages/itwinui-react/setupTests.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import '@testing-library/jest-dom';
+import '@testing-library/jest-dom/vitest';
 import { TextEncoder } from 'util';
```
mayank99, line 6: `TextEncoder` might not be necessary to mock anymore. I believe it was added to work around some jest-specific issue

packages/itwinui-react/vitest.config.ts
mayank99: The `css: false` flag might not be working. Try adding this to the `mock-css` vite plugin:

```js
load(id) {
  if (id.endsWith('styles.css')) {
    return '';
  }
  return null;
},
```

packages/itwinui-react/vitest.config.ts
mayank99: Yes. Basically all `test-types` files should be eventually removed. Type tests can be written more robustly using vitest. https://vitest.dev/guide/testing-types.html

rohan-kadkol: > Added ([5e40a96](https://github.com/iTwin/iTwinUI/commit/5e40a96fc6d92b25980c439d3bc43863ffa41a5c)). So no type tests fails as of right now.

I'm not sure. When I manually added a type error in `Table.test.tsx`, the tests still passed.

mayank99: Looks like it only checks `.test-d.tsx` files by default. We need to configure it using [`typecheck.include`](https://vitest.dev/config/#typecheck-include)

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99: Lets handle `isJest`/`isDev` issues in a separate follow-up PR, since it affects runtime code

packages/itwinui-react/vitest.config.ts
rohan-kadkol: > Done ([04958b5](https://github.com/iTwin/iTwinUI/commit/04958b5d40b2521c3a5f9e85957b4089bd7b7e17))

I feel coverage should be opt-in in `package.json`'s `test:unit` script too. For e.g., if I am trying to pass just one failing unit test, the coverage report is unnecessary and obstructive.

packages/itwinui-react/src/core/utils/functions/dev.ts
rohan-kadkol: Sounds good. @siddhantrawal, maybe you can add a bullet point TODO in the PR description so that we can remember to address this after the PR merge.

packages/itwinui-react/vitest.config.ts
mayank99: Only `test.tsx`, because `test-types.tsx` is completely unrelated and being handled by `tsc` explicitly.

```js
include: ['**/*.test.?(c|m)[jt]s?(x)']
```

packages/itwinui-react/vitest.config.ts
mayank99: Yes

mayank99 (approved):
LGTM 🚀 It runs faster and feels much nicer overall compared to `jest`.

rohan-kadkol (approved):
LGTM 🎉

packages/itwinui-react/vitest.config.ts
```diff
@@ -0,0 +1,44 @@
...
+    outputFile: 'coverage/junit.xml',
+    alias: [
+      {
+        find: /^(.*)\/styles.js$/,
+        replacement: path.resolve('./__mocks__/styles.js'),
+      },
+    ],
+    typecheck: {
+      enabled: true,
+      include: ['**/*.test.?(c|m)[jt]s?(x)'],
```
mayank99, line 24: I think this actually broke our tests. The tests are no longer being run; they are only being type-checked.

PR #1787, opened 2024-01-18
All icons use icon fill color

FlyersPh9:
## Changes

UX was questioning why icons in side navigation were not using the icon color variables.

- **Button** icon uses icon instead of text var.
- **Radial progress indicator** icon uses icon instead of text var.
- **Select** icon uses icon instead of `currentColor` which would inherit the text color.

One exception was **avatar** because it uses the soft background colors and the icon fill always needs to be dark.

## Testing

- Searched repo's SCSS for `fill:` and verified all are using the icon var.
- Ran visual tests.

## Docs

n/a


packages/itwinui-css/src/select/select.scss
```diff
@@ -39,7 +39,7 @@
 
   .iui-icon svg {
     @include menu.iui-menu-icon;
-    fill: currentColor;
+    fill: var(--iui-color-icon);
```
FlyersPh9, line 42: I changed this to pink and it did not show up in any tests. Maybe not necessary anymore?

packages/itwinui-css/src/button/base.scss
```diff
@@ -50,7 +50,7 @@
   --_iui-button-active-stripe-inset-inline: initial;
   --_iui-button-active-stripe-background: var(--iui-color-border-accent);
   --_iui-button-text-color: var(--iui-color-text-accent);
-  --_iui-button-icon-fill: var(--iui-color-text-accent);
+  --_iui-button-icon-fill: var(--iui-color-icon-accent);
```
mayank99, line 53: i feel like we had a specific reason we used text color here 🤔 or maybe I'm misremembering

packages/itwinui-css/src/button/base.scss
mayank99, line 1: user-facing changes should have changesets

packages/itwinui-css/src/select/select.scss
mayank99: confirm it in react too, then remove it

packages/itwinui-css/src/button/base.scss
FlyersPh9: I think I remember that too.  I think we thought the icon next to the text looked off.

**With this change:**
Screenshot 2024-01-18 at 6 16 12 PM

I think we should use the icon color now to avoid having to explain why we're using text for icon fill.

packages/itwinui-css/src/button/base.scss
FlyersPh9: Damn I'm rusty, added changesets.

mayank99 (approved):

packages/itwinui-css/src/select/select.scss
FlyersPh9: Checked in our CSS demos, React demos, and test images: no change.

Dev tools shows it being overrode:
Screenshot 2024-01-18 at 6 22 34 PM

Removing it.

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,7 +41,7 @@
   svg {
     @include mixins.iui-icon-style('m');
     transition: fill var(--iui-duration-1) ease-out;
-    fill: var(--_iui-button-icon-fill, currentColor);
+    fill: var(--_iui-button-icon-fill, var(--iui-color-icon));
```
rohan-kadkol, line 44: Wouldn't it be safer that the icon is the same color as the text? Also, although the difference between `currentColor` is not apparent in the default case, it is apparent in the cta, high-visibility, and disabled cases.

https://github.com/iTwin/iTwinUI/assets/45748283/9e8e3c2d-8a81-4d74-8e2d-33eeac4f7a52

FlyersPh9: Reverted back to `currentColor` because it is a safer fallback and if we did our jobs right it should never be applied anyways.

rohan-kadkol (approved):

PR #1788, opened 2024-01-19
decrease `LinkAction` specificity

mayank99:
## Changes

Follow-up to #1762 where I ran into the issue of `LinkAction` overriding other styles when used with the `as` prop.

This PR is a "quick fix" that addresses the problem today by nullifying the specificity (thus giving priority to anything else).

A "proper" solution would likely involve using nested cascade layers to full control the order of styles, but that can happen in the future.

## Testing

Tested that this code preserves `Anchor` styling:

```jsx
<LinkBox>
  <LinkAction as={Anchor} href='#'>Hello</LinkAction>
</LinkBox>
```

| Before | After |
| --- | --- |
| image | image |

## Docs

N/A. Added changesets.

### DIFF
--- a/.changeset/plenty-garlics-laugh.md
+++ b/.changeset/plenty-garlics-laugh.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+`LinkAction` will no longer override styles from other components (such as `Anchor`).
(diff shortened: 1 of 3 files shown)


rohan-kadkol (commented):
Would it be better to add a `react-workshop` story for `LinkAction` and test this case to prevent regressions and also demo `LinkAction`?

mayank99 (commented):
Not needed rn imo. The purpose of this PR is to be "quick fix" as a stopgap for more involved PRs later.

rohan-kadkol (approved):

PR #1791, opened 2024-01-22
expose `ListItem.Action` to wrap `LinkAction`

mayank99:
## Changes

follow-up to https://github.com/iTwin/iTwinUI/pull/1762#discussion_r1443202152

## Testing

N/A. The new component is a direct reference to the old one, so no tests needed.

## Docs

Updated docs and added changeset.

### DIFF
--- a/apps/website/src/content/docs/list.mdx
+++ b/apps/website/src/content/docs/list.mdx
@@ -32,7 +32,7 @@ There is also a `size` prop which can be set to `'large'` to ensure that all ite
 
 `ListItem` has an `actionable` prop which can be set to true to provide hover styling.
 
-Please note that this prop only provides the hover styling and does nothing else. Usually you want to combine this prop either with `LinkAction` (which lets you use an anchor or a button) or with a more advanced component, such as [`Select`](select), which will do many additional things like set `role=option`, manage roving tabindex, and handle arrow-key navigation.
+Please note that this prop only provides the hover styling and does nothing else. Usually you want to combine this prop either with `ListItem.Action` (which lets you use an anchor or a button) or with a more advanced component, such as [`Select`](select), which will do many additional things like set `role=option`, manage roving tabindex, and handle arrow-key navigation.
 
 ### With `ListItem.Action`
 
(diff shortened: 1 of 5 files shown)


examples/List.links.tsx
```diff
@@ -3,25 +3,25 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { List, ListItem, LinkAction } from '@itwin/itwinui-react';
+import { List, ListItem } from '@itwin/itwinui-react';
 
 export default () => {
   return (
     <List>
       <ListItem actionable>
```
rohan-kadkol, line 11: Should all these `ListItem`s have `actionable`? Since the hover styling also come from the `LinkAction`:

https://github.com/iTwin/iTwinUI/blob/b0d4ea174b2d644ee85147ff52d61555804e0f6f/packages/itwinui-css/src/menu/mixins.scss#L52-L53

packages/itwinui-react/src/core/List/ListItem.tsx
```diff
@@ -121,4 +127,14 @@ export const ListItem = Object.assign(ListItemComponent, {
    * </ListItem>
    */
   Description: ListItemDescription,
+  /**
+   * Wrapper over [LinkAction](https://dev-itwinui.bentley.com/docs/linkaction) which allows rendering a link inside a ListItem.
```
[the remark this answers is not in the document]
mayank99: yup

examples/List.links.tsx
mayank99: I think lets leave it, because it can't hurt, will also work for older browsers, and will help in case one of the list items is supposed to be actionable but doesn't include a `ListItem.Action`.

apps/website/src/content/docs/list.mdx
```diff
@@ -34,9 +34,9 @@ There is also a `size` prop which can be set to `'large'` to ensure that all ite
 
 Please note that this prop only provides the hover styling and does nothing else. Usually you want to combine this prop either with `LinkAction` (which lets you use an anchor or a button) or with a more advanced component, such as [`Select`](select), which will do many additional things like set `role=option`, manage roving tabindex, and handle arrow-key navigation.
```
rohan-kadkol, line 35: There seem to be a few leftover references to the old `LinkAction` on this page. E.g.:

```suggestion
Please note that this prop only provides the hover styling and does nothing else. Usually you want to combine this prop either with `ListItem.Action` (which lets you use an anchor or a button) or with a more advanced component, such as [`Select`](select), which will do many additional things like set `role=option`, manage roving tabindex, and handle arrow-key navigation.
```

mayank99: fixed. [`9562e9c`](https://github.com/iTwin/iTwinUI/pull/1791/commits/9562e9ce6a7aed317e1ba3c3a985f2a77c39925c)

rohan-kadkol (approved):

PR #1792, opened 2024-01-22
fix Table select-all state to account for sub-rows

mayank99:
## Changes

The conditions in `SelectionColumn` were bugged, resulting in the "select-all" checkbox (in the Table header) being disabled or checked when it should not be (such as when top-level rows are disabled but their sub-rows are not disabled). I've updated it to use `preFilteredFlatRows` (instead of `rows` or `initialRows`), which should be more reliable because it looks at _all_ rows (including sub-rows, even when a filter is active).

## Testing

Verified in vite playground and unit test that the "select-all" checkbox is not disabled and unchecked.

| Before | After | 
| --- | --- |
| image | image |

```tsx
import { SelectionColumn, Table } from '@itwin/itwinui-react';
import type { Column } from '@itwin/itwinui-react/react-table';

    () => [
      SelectionColumn({
        isDisabled: (rowData: Record<string, any>) => !!rowData.subRows?.length,
      }),
      {
        id: 'name',
        Header: 'Name',
        accessor: 'name',
      },
    ],
    [],
  ) as Column[];

  const data = [
    {
      name: 'Row 1',
      subRows: [
        { name: 'Row 1.1' },
        {
          name: 'Row 1.2',
          subRows: [
            { name: 'Row 1.2.1' },
            { name: 'Row 1.2.2' },
            { ame: 'Row 1.2.3' },
            { name: 'Row 1.2.4' },
          ],
        },
        { name: 'Row 1.3' },
        { name: 'Row 1.4' },
      ],
    },
    {
      name: 'Row 2',
      subRows: [{ name: 'Row 2.1' }, { name: 'Row 2.2' }, { name: 'Row 2.3' }],
    },
  ];

  return (
    <>
      <Table
        emptyTableContent='No data.'
        isSelectable
        isSortable
        data={data}
        columns={columns}
      />
    </>
  );
}

```

</details> 

## Docs

Added changeset.


packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -2368,6 +2368,18 @@ it('should render sub-rows with custom expander', async () => {
   expect(rows.length).toBe(10);
 });
 
+it('should not disable select-all checkbox when top-level rows are disabled but subrows are not', () => {
```
rohan-kadkol, line 2371: nit: clarity

```suggestion
it('should not disable select-all checkbox when all top-level rows are disabled but all subrows are not', () => {
```

rohan-kadkol (approved):

PR #1794, opened 2024-01-22
use `LinkAction` inside `ExpandableBlock`

mayank99:
## Changes

Instead of the entire Trigger area being used as a `<button>`, only the main label will now be wrapped in a `<button>`. The clickable area is still unchanged, through the use of `LinkAction`/`LinkBox`. The `caption` is now associated with the trigger button using `aria-describedby`.

The main benefit is that it allows users to nest secondary actions. This is actually something a user asked for privately, and I had to dissuade them because it would have resulted in `<button>`s being used inside `<button>`s, whereas now the `<button>`s are no longer nested.

image

## Testing

Updated unit tests.

Also tested that the behavior is correct using mouse, keyboard, and screen readers.

image

image

## Docs

Updated docs to show an example of secondary action. Probably could use more detailed docs about the general approach.


examples/ExpandableBlock.actions.tsx
rohan-kadkol, line 1: Apart from this example, I feel there should be a react-workshop visual test for this case too. But that can come after this PR too.

rohan-kadkol (approved):

examples/ExpandableBlock.actions.tsx
rohan-kadkol: I felt it was needed to make sure that passing secondary actions doesn't regress.

But yes, since end icons are already tested in ExpandableBlock → StatusIcon story and also in css-workshops tests, it may be alright to not add a new visual test for this.

PR #1795, opened 2024-01-24
remove `overflow` and inline `transform` from Dialog

### LINKED ISSUE #1551: Modal's "transform: translate" to center it makes text blurry
We have a modal that has a few tabs with different data shown in a table.
However, text in a table becomes very blurry when adding a table height (so that it would not overflow and we would have a scrollbar to look through all entries in the table).
image

Even without height, it does not look clear enough.
image

I tried adding `backface-visibility: hidden;`, `-webkit-font-smoothing: subpixel-antialiased;`, overwritting transform with `transform: translate(calc(-50% + 0.5px), calc(-50% + 0.5px))` and other things but nothing worked. I think this happens because with a lot of elements put in each other, those "transform translate" decimal sizes become worse than if they were on top level (for example, tabs' text looks fine).

The solution for now I came up with for now is removing transform and changing it to `display: flex` and then centering with `justify-content` and `align-items` in the modal's wrapper. The only problem I noticed with that is that if a modal has buttons with navigation when closing the current and loading another modal, there is a noticeable blinking.

Could anyone check this?

### LINKED ISSUE #1789: Dialog becomes off-center when browser window is resized
Using v2's Dialog component (not full-page), when I resize the browser window, the dialog becomes off-center and sometimes partially moves off the top or left side of the screen.  It's especially noticeable with horizontal resizing, but also subtly occurs with vertical resizing.

Normal appearance:
image

Appearance after resizing window:
image

#### Expected behavior
The dialog should stay centered, with about the same proportion of space on each of the 4 sides as it had when it was initially displayed.

#### Reproduction
Link to a minimal repro: [Example](https://stackblitz.com/edit/github-xmpyql?file=package.json,src%2Fmain.tsx)

Note: You'll need to click "Open in new tab" and then "Link to project" to get the preview into its own window so you can better see the effects of resizing.

image

##### Steps to reproduce
1. Create a `Dialog` component with some content
2. View the page in the browser
3. Press the Restore/Maximize button in the browser, and/or drag the window resizing handles horizontally and/or vertically to see the dialog box become off-centered

#### Additional information
An inline `transform` style with specific pixel translations automatically gets added to the component, in addition to the `transform` style (with percent translations) already set for its class in CSS.  If I disable the inline style and keep the style in CSS, the dialog stays centered after a resize, as expected:  (Here, the "good" style is circled in green, and the "bad" style is circled in red.)

image

mayank99:
## Changes

This PR removes a bunch of code added in #1424 and #1693.

- Removed `overflow: hidden`. It causes blurring and is not needed because `Dialog.Content` already has overflow (and is required after #1269).
  - Fixes #1551
  - The `:has`/`data-iui-flex` stuff is also redundant because dialog is already using flex by default.
  - This required explicitly adding border-radius to the title bar.
- Removed the inline rounded `transform` style. It breaks window resizing and is unnecessary because removing `overflow: hidden` is enough to avoid blurring.
  - Fixes #1789

## Testing

Existing tests pass. Verified in Chrome Windows that there is no blurring and that the dialog repositions when the window is resized.

https://github.com/iTwin/iTwinUI/assets/9084735/9eb2089d-a07d-42a4-9c22-1f9dfeaf19c9

## Docs

Added changeset.


.changeset/little-clocks-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Fixed Dialog blurring and resizing issues by removing `overflow: hidden` and inline `transform` styles.
```
rohan-kadkol, line 6: Should we mention `Modal` also?

Also, is `overflow: hidden` and inline `transform` styles internal information that is impertinent for users, and thus should not be in the changeset?

mayank99: Mentioned Modal [`8a16169`](https://github.com/iTwin/iTwinUI/pull/1795/commits/8a161697259b493a725af650a00e9e1ead5e1923)

These styles are internal details yes, but as you can see in the linked issues, our users are already messing with such things, so it's important to mention. If they were manually adding `overflow: visible`, they can remove it after this release.

rohan-kadkol (approved):

Issue #1750, opened 2023-12-21
SvgChevronRight icon in SubRowExpander component rotates 180 degrees instead of 90 on expansion


DenisMicha:
**SvgChevronRight** icon in _SubRowExpander_ component rotates 180 degrees instead of 90 on expansion.

#### Actual behavior
Chevron points to the left side when sub rows expanded (rotates 180deg).

#### Expected behavior
Chevron points down when sub rows expanded (rotates 90deg).

This behavior happens because both **SvgChevronRight** component and its wrapping **IconButton** component include 90deg rotation on expansion, therefore total rotation becomes 180deg. The "transform" property which causes rotation should be removed from **SvgChevronRight** component styles.

mayank99:
Could you provide a repro by forking [this template](https://stackblitz.com/github/iTwin/iTwinUI/tree/main/minimal-sandbox?file=src/App.tsx)? It works fine here: https://itwin.github.io/iTwinUI/react/?story=table--expandable-subrows&theme=dark

DenisMicha:
Hi @mayank99,

Here is the repro code: https://stackblitz.com/edit/github-cph1jx?file=src%2FApp.tsx.

Apparently adding the "subComponent" property to the Table component adds the additional rotation. It is a bit confusing that adding/removing this property impacts chevron rotation, so I think SvgChevronRight style is good but the rotation as a side effect of "subComponent" property should be removed.

rohan-kadkol:
I believe the `subComponent` is not intended to be used when any of the rows have `subRows`.

But yes, the side effect you mentioned in https://github.com/iTwin/iTwinUI/discussions/1749 seems to be duplicate because there already is CSS that applies a transform on `.iui-button-icon`:

https://github.com/iTwin/iTwinUI/blob/5aba6b48820e452adf6f53fcb5dffa5fe8df8242/packages/itwinui-css/src/table/base.scss#L218-L220

mayank99:
@DenisMicha Could you share why you're passing `subComponent` while also using `subRows`? These are somewhat mutually-exclusive features that are not designed to work together.

In your sandbox, `subComponent={() => undefined}` does more than just combine the rotations (from `subComponent` and `subRows`). It also changes some alignment and applies border around the entire row and its subcomponent (which is `undefined` in this case).
| without `subComponent` | with `subComponent` |
| --- | --- |
[...]

PR #1796, opened 2024-01-24
align column header with expandable rows

mayank99:
## Changes

Extracted out the sub-row padding logic into a utility function and used it to add the same padding to column headers.

Fixes https://github.com/iTwin/iTwinUI/issues/1750#issuecomment-1870444146

## Testing

| Before | After |
| --- | --- |
| image | image |

## Docs

Added changeset.


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -930,6 +930,14 @@ export const Table = <
                   {headerGroup.headers.map((column, index) => {
                     const { onClick, ...restSortProps } =
                       column.getSortByToggleProps();
+
+                    const columnHasExpanders =
+                      hasAnySubRows &&
+                      index ===
+                        headerGroup.headers.findIndex(
+                          (c) => c.id !== SELECTION_CELL_ID,
+                        );
```
rohan-kadkol, line 939: Is this selecting the first non-selection column's header? Maybe separating this to a new variable with a description variable name (e.g. `isFirstNonSelectionColumn`) or at least a comment explanation here would make it more understandable.

mayank99: Added comment [`5027473`](https://github.com/iTwin/iTwinUI/pull/1796/commits/5027473e80360d34ce509ed1adb1ef1adeb39aa3)

rohan-kadkol (approved):
LGTM.

Since you wrote "fixes # 1750 (comment)", GitHub will close that entire issue as closed. So, you may have to un-link that issue with this PR or re-open that issue after merging this PR.

mayank99 (commented):
It's intentional. That issue is invalid (using `subRows` and `subcomponent` together), so that specific comment is the only part that needs fixing.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -930,6 +930,14 @@ export const Table = <
                   {headerGroup.headers.map((column, index) => {
                     const { onClick, ...restSortProps } =
                       column.getSortByToggleProps();
+
+                    const columnHasExpanders =
+                      hasAnySubRows &&
+                      index ===
+                        headerGroup.headers.findIndex(
+                          (c) => c.id !== SELECTION_CELL_ID, // first non-selectable column is the expander column
```
rohan-kadkol, line 938: nit: Since columns are not selectable. The column's name is selection column.
```suggestion
                          (c) => c.id !== SELECTION_CELL_ID, // first non-selection column is the expander column
```

rohan-kadkol:
> It's intentional. That issue is invalid (using `subRows` and `subcomponent` together), so that specific comment is the only part that needs fixing.

Then maybe that issue should be closed with a new PR that mentions somewhere that using `subRows` and `subComponents` together is invalid. Possibly in the JSDocs, docs website, etc. Else, future users may still accidentally do that.

PR #1799, opened 2024-01-25
Fix `Table` columns un-resizable when string widths bug

### LINKED ISSUE #897: Table column cannot be resized when the width is set as a percentage
Table columns cannot be resized when the width is set as a percentage and the whole table is not resized. For example, if you shrink or expand the table, then columns will be resizable.

#### Expected behavior
Table columns should be resizable without shrinking or expanding the whole table.

#### Reproduction
[Link to a minimal repro: ](https://codesandbox.io/s/itwinui-react-example-forked-de2ejq?file=/src/App.tsx)

##### Steps to reproduce
`const columns = useMemo(
    () => [
      {
        Header: "Header",
        columns: [
          {
            id: "id1",
            accessor: "id1",
            Header: "id1",
            width: "20%"
          },
          {
            id: "id2",
            accessor: "id2",
            Header: "id2",
            width: "20%"
          },
          {
            id: "id3",
            accessor: "id3",
            Header: "id3",
            width: "60%"
          }
        ]
      }
    ],
    []
  );`

#### Additional information
`"@itwin/itwinui-react": "1.46.0"`

rohan-kadkol:
## Changes

Fixes #897

<details>

Before:

https://github.com/iTwin/iTwinUI/assets/45748283/b7b837b9-83f1-4622-ac87-8948f35df42a

After:

https://github.com/iTwin/iTwinUI/assets/45748283/863d41b6-6aab-4684-a14a-b5237e5e6b89

</details>

The table columns become un-resizable when they are given a width that is a string. (e.g. `"200px"`, `50%`, etc.). This is because when doing the resize, the below code is executed:

https://github.com/iTwin/iTwinUI/blob/607593ef67fc59408284e0fa68a6d5e9c3c8c7d9/packages/itwinui-react/src/core/Table/hooks/useResizeColumns.tsx#L500

… and when `header.width` = `"200px"` for example, `Number("200px")` = `nan`, and when the new values are deemed invalid by:

https://github.com/iTwin/iTwinUI/blob/607593ef67fc59408284e0fa68a6d5e9c3c8c7d9/packages/itwinui-react/src/core/Table/hooks/useResizeColumns.tsx#L298-L307

… we early return and do no resize.

This PR makes sure to also handle `width`s passed to the column as strings.

## Testing

Tested with the below code in the vite playground:

<details>
<summary>App.tsx</summary>

```jsx
import { Table } from '@itwin/itwinui-react';
import { useMemo } from 'react';

const App = () => {
  const columns = useMemo(
    () => [
      {
        Header: 'Header',
        columns: [
          {
            id: 'id1',
            accessor: 'id1',
            Header: 'id1',
            width: '20%',
          },
          {
            id: 'id2',
            accessor: 'id2',
            Header: 'id2',
            width: '20%',
          },
          {
            id: 'id3',
            accessor: 'id3',
            Header: 'id3',
            width: '60%',
          },
        ],
      },
    ],
    [],
  );

  const data = [
    {
      id1: 'id1',
      id2: 'id2',
      id3: 'id3',
    },
    {
      id1: 'id1',
      id2: 'id2',
      id3: 'id3',
    },
    {
      id1: 'id1',
      id2: 'id2',
      id3: 'id3',
    },
  ];

  return (
    <>
      <Table
        isResizable
        columns={columns}
        data={data}
        emptyTableContent='empty'
      />
    </>
  );
};

export default App;
```

</details>

---

* All existing unit tests pass
* Added a unit test to confirm that resizing columns works with string units too.

## Docs

Added changeset.


rohan-kadkol:
> It would be a good idea to maybe have a quick video demo of before and after. I was trying to reproduce, code sandbox link is not working.

Sounds good. Updated the PR description with the "Before/After" screen recordings and also the `App.tsx` code that I used for the demos.

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -2679,6 +2679,79 @@ it('should handle resize with touch', () => {
   expect(headerCells[2].style.width).toBe('100px');
 });
 
+it.each(['px', '%', 'em', 'rem', 'vw', 'vh', 'ch', 'ex'])(
```
mayank99, line 2682: no need for all of these imo. Just two or three should be enough: `px`, `%`, `rem`

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -2679,6 +2679,79 @@ it('should handle resize with touch', () => {
...
+
+    const rows = container.querySelectorAll('.iui-table-body .iui-table-row');
+    expect(rows.length).toBe(3);
+
+    const resizers = container.querySelectorAll(
+      '.iui-table-resizer',
+    ) as NodeListOf<HTMLDivElement>;
+
+    expect(resizers).toBeTruthy();
+    resizers.forEach((resizer) => expect(resizer).toBeTruthy());
```
mayank99, line 2723: nit: these can be removed
```suggestion
```

rohan-kadkol: Since I am testing resizing using the first and the second resizer, `resizers` is needed.

I had initially kept L2715-2716 since they were there in many of the tests. But removed it since it's not needed. (https://github.com/iTwin/iTwinUI/pull/1799/commits/0e6f208a9ae7cfc7b4105353e79e8930282ec7b7)

packages/itwinui-react/src/core/Table/Table.test.tsx
rohan-kadkol: Done (https://github.com/iTwin/iTwinUI/pull/1799/commits/2f9f9613f8a6b39b4bfe76bb2c683c0579941a6e)

packages/itwinui-react/src/core/Table/Table.test.tsx
mayank99: Ah yeah. I think the `toBeTruthy()` checks for `resizers` could still be removed, but it's ok

mayank99 (approved):
LGTM

PR #1803, opened 2024-01-29
Adding typecheck manually for test.tsx files.

siddhantrawal:
Fixes the issue related to this [convo](https://github.com/iTwin/iTwinUI/pull/1784#discussion_r1470249106).

## Changes

Removed `typecheck` from vitest.config file.
Manually added `typecheck` for all `test.tsx` files.

## Testing

All test cases running properly.

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/tsconfig.test.json
+++ b/packages/itwinui-react/tsconfig.test.json
@@ -1,12 +1,5 @@
 {
   "extends": "./tsconfig.json",
   // TODO: Test all the files, and not just the Table files or just the build files.
-  "include": [
-    "src/core/Table/**/*",
-    "src/react-table/*",
-    "src/core/Select/**/*",
-    "src/core/LabeledSelect/**/*",
-    "src/core/ComboBox/**/*",
-    "**/*.test.?(c|m)[jt]s?(x)"
-  ]
+  "include": ["src/**/*.test.*", "src/**/*.types-test.*"]
 }


packages/itwinui-react/vitest.config.ts
mayank99, line 1: Missing part of my comment:

> Manually use `tsc` to type check all `.test.tsx` files, similar to how we type-check selective files right now (using the `test:types` command).

packages/itwinui-react/tsconfig.test.json
```diff
@@ -6,6 +6,7 @@
     "src/react-table/*",
     "src/core/Select/**/*",
     "src/core/LabeledSelect/**/*",
-    "src/core/ComboBox/**/*"
+    "src/core/ComboBox/**/*",
+    "**/*.test.?(c|m)[jt]s?(x)"
```
mayank99, line 10: Can't we simplify this entire thing as follows?

```js
"include": [
   "src/**/*.test.*",
   "src/**/*.types-test.*"
]
```

cc @rohan-kadkol

rohan-kadkol: Yes, sounds good.

The only difference is that we will now not be type testing non-test files in the components from the current `include` (E.g. type errors in `src/core/Table/actionHandlers/*.ts`, etc.). But I feel this can be handled later.

mayank99 (approved):

rohan-kadkol (approved):

packages/itwinui-react/tsconfig.test.json
rohan-kadkol: We can add it to the PR description as a TODO item after this PR.

mayank99: Personally I feel it's unnecessary. Any non-test files should already be type checked when they are converted to `.js` during `build` step.

PR #1806, opened 2024-01-30
Updated jsdom version.

siddhantrawal:
Part of #1784 

## Changes

Bumped `jsdom` version from `21.0.0` to `24.0.0`.

## Testing

All test cases passed.

## Docs

N/A


mayank99 (approved):
LGTM

Personally, I don't think this is the right way (One is checking for the `style` property in JS, while the other is checking for the `style` attribute in HTML), but I do not care too much because everything in JSDOM is fake anyway.

rohan-kadkol (approved):

PR #1807, opened 2024-01-30
Added common reliable check to detect jest,vitest and mocha testing framework.

siddhantrawal:
Part of #1784, fixes isJest/IsDev issues.

## Changes

Added new checks to detect test frameworks other than `jest`:
-  `vitest`  (from [console-fail-test](https://github.com/JoshuaKGoldberg/console-fail-test/blob/4738d5f3e506735c2c16f52588a9cfade58b9569/src/environments/vitest.ts#L11), see [thread](https://github.com/iTwin/iTwinUI/pull/1807#discussion_r1471821627))
- `mocha` (from [console-fail-test](https://github.com/JoshuaKGoldberg/console-fail-test/blob/4738d5f3e506735c2c16f52588a9cfade58b9569/src/environments/mocha.ts#L19-L28), see [thread](https://github.com/iTwin/iTwinUI/pull/1807#discussion_r1471858302))

## Testing

Dev errors not showing. (vitest)

mocha check tested in isolation (see [thread](https://github.com/iTwin/iTwinUI/pull/1807#discussion_r1471832297))

## Docs

N/A

### DIFF
--- a/.changeset/twelve-rules-glow.md
+++ b/.changeset/twelve-rules-glow.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added check to detect `jest`,`vitest` and `mocha` testing framework.
--- a/.changeset/twelve-rules-glow.md
+++ b/.changeset/twelve-rules-glow.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-react': minor
+'@itwin/itwinui-react': patch
 ---
 
-Added check to detect `jest`,`vitest` and `mocha` testing framework.
+iTwinUI will now also check for `vitest` and `mocha` (in addition to `jest`) before running any code that would cause problems in these test runners.
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -13,9 +13,7 @@ import {
   useControlledState,
   useLatestRef,
   importCss,
-  isJest,
-  isMocha,
-  isVitest,
+  isUnitTest,
 } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { ThemeContext } from './ThemeContext.js';
@@ -303,8 +301,8 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
       return;
     }
 
-    // bail if isJest/isMocha/isVitest because unit tests don't care about CSS 🤷
-    if (isJest || isMocha || isVitest) {
+    // bail if isUnitTest because unit tests don't care about CSS 🤷
+    if (isUnitTest) {
       return;
     }
 
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -13,7 +13,7 @@ import {
   useControlledState,
   useLatestRef,
   importCss,
-  isJest,
+  isTestingFramework,
 } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { ThemeContext } from './ThemeContext.js';
@@ -301,8 +301,8 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
       return;
     }
 
-    // bail if jest because it doesn't care about CSS 🤷
-    if (isJest) {
+    // bail if TestingFramework because it doesn't care about CSS 🤷
+    if (isTestingFramework()) {
       return;
     }
 
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -301,7 +301,7 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
       return;
     }
 
-    // bail if TestingFramework because it doesn't care about CSS 🤷
+    // bail if isTestingFramework because unit tests don't care about CSS 🤷
     if (isTestingFramework()) {
       return;
     }
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -302,7 +302,7 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
     }
 
     // bail if isTestingFramework because unit tests don't care about CSS 🤷
-    if (isTestingFramework()) {
+    if (isTestingFramework) {
       return;
     }
 
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -13,7 +13,9 @@ import {
   useControlledState,
   useLatestRef,
   importCss,
-  isTestingFramework,
+  isJest,
+  isMocha,
+  isVitest,
 } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
 import { ThemeContext } from './ThemeContext.js';
@@ -301,8 +303,8 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
       return;
     }
 
-    // bail if isTestingFramework because unit tests don't care about CSS 🤷
-    if (isTestingFramework) {
+    // bail if isJest/isMocha/isVitest because unit tests don't care about CSS 🤷
+    if (isJest || isMocha || isVitest) {
       return;
     }
 
--- a/packages/itwinui-react/src/core/utils/functions/dev.ts
+++ b/packages/itwinui-react/src/core/utils/functions/dev.ts
@@ -5,10 +5,14 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 const isJest = typeof (globalThis as any).jest !== 'undefined';
+
+// See https://github.com/JoshuaKGoldberg/console-fail-test/blob/4738d5f3e506735c2c16f52588a9cfade58b9569/src/environments/mocha.ts#L19-L28
 const isMocha =
   typeof (globalThis as any).beforeEach !== 'undefined' &&
   `${(globalThis as any).beforeEach}`.replace(/\s/g, '') ===
     'function(name,fn){suites[0].beforeEach(name,fn);}';
+
+// See https://github.com/iTwin/iTwinUI/blob/bb1c2260568ce58b4676f2b18081ec7b825812cd/packages/itwinui-react/src/core/utils/functions/dev.ts#L13
 const isVitest = (typeof globalThis as any).__vitest_index__ !== 'undefined';
 
 const isUnitTest = isJest || isVitest || isMocha;
--- a/packages/itwinui-react/src/core/utils/functions/dev.ts
+++ b/packages/itwinui-react/src/core/utils/functions/dev.ts
@@ -2,18 +2,26 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-const isTestingFramework =
-  typeof vitest !== 'undefined' ||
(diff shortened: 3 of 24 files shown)


packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,14 +2,21 @@
...
  *--------------------------------------------------------------------------------------------*/
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const isJest = typeof (globalThis as any).jest !== 'undefined';
+const isTestingFramework = () => {
+  return (
+    typeof vitest !== 'undefined' ||
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    typeof (globalThis as any).jest !== `undefined` ||
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    typeof (globalThis as any).mocha !== 'undefined'
```
mayank99, line 11: Have you actually tested or researched that this is the correct way of detecting mocha?

We can't just check `globalThis.whatever`. We know from experience that it works for `jest` and `vitest` because both of them inject globals, but `mocha` might not do the same.

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,14 +2,21 @@
...
+  );
+};
 
 let isDev = false;
 
 // wrapping in try-catch because process might be undefined
 try {
-  isDev = process.env.NODE_ENV !== 'production' && !isJest;
+  isDev = process.env.NODE_ENV !== 'production' && !isTestingFramework;
```
rohan-kadkol, line 19: `isTestingFramework` is a function, so you would need to call it.

```suggestion
  isDev = process.env.NODE_ENV !== 'production' && !isTestingFramework();
```

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,14 +2,21 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const isJest = typeof (globalThis as any).jest !== 'undefined';
+const isTestingFramework = () => {
```
rohan-kadkol, line 5: Any reason this is a function and not a variable?

packages/itwinui-react/src/core/utils/functions/dev.ts
rohan-kadkol: Can the check for mocha come later when we're using Mocha?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -301,8 +301,8 @@ const FallbackStyles = ({ root }: { root: HTMLElement }) => {
       return;
     }
 
-    // bail if jest because it doesn't care about CSS 🤷
-    if (isJest) {
+    // bail if TestingFramework because it doesn't care about CSS 🤷
```
rohan-kadkol, line 304: nit: Clarity

```suggestion
    // bail if isTestingFramework because unit tests don't care about CSS 🤷
```

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,14 +2,21 @@
...
+  return (
+    typeof vitest !== 'undefined' ||
```
mayank99, line 7: I think it will also fail for vitest if [`globals`](https://vitest.dev/config/#globals) is not enabled.

packages/itwinui-react/src/core/utils/functions/dev.ts
rohan-kadkol: In that case, instead of testing whether we're in a testing framework, can we test if we're in the browser? If so, that will be more reliable since users may use non-jest/vitest/mocha testing frameworks also.

mayank99: Detecting the browser serves a different purpose, and our components can be SSR'd (using one of [React's server apis](https://react.dev/reference/react-dom/server)).

At best we could try to detect if the code is running in `jsdom`, but then that excludes alternative DOM implementations like `happy-dom`.

Plus, both vitest and mocha can run in the browser too. We might want to still bail if it's running in the browser?

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99: No, there is no difference between `vitest` and `globalThis.vitest`; they are one and the same. With `globals` disabled, both will be empty (you should verify this just to be sure) because the usage requires an explicit import ([docs](https://vitest.dev/api/vi.html)).

We might need a different way of detecting vitest.

veekeys: `process.env.VITEST` did not do what you wanted?
image


https://vitest.dev/config/#configuration

mayank99: `process.env` is not available in the browser, but we could use it with `try`-`catch`. e.g.:
```js
let isVitest = false;
try {
  isVitest = process.env.VITEST !== undefined;
} catch {}
```

I'm also thinking we could additionally check a custom environment variable that we can ask users to manually set in our documentation. e.g:
> By default, iTwinUI tries to detect `jest` and `vitest` to bail out of running code that's irrelevant for such test environments. If you are using a different test runner, such as `mocha`, please manually set the environment variable `IS_UNIT_TEST=TRUE`.

packages/itwinui-react/src/core/utils/functions/dev.ts
rohan-kadkol: Upon thinking about it, I just realized, I think our goal should be to highlight *all* incorrect uses of our components. The only exception is when in production.

So, if a user's test case uses our components incorrectly, should we show the console warning when running the unit tests too? I feel that is the safest, and also easier on our end.

```js
// Show dev errors in all developer-side environments. i.e. dev server/browser, testing environments, etc.
isDev = process.env.NODE_ENV !== 'production';
```

mayank99: > Upon thinking about it, I just realized, I think our goal should be to highlight all incorrect uses of our components. The only exception is when in production.

It's not just about highlighting incorrect usages. We straight up don't want to run certain code in test environments. See current usage in `ThemeProvider`.

Also I think it would be kinda annoying to surface dev-only warnings during tests. The terminal could get flooded with lots of warnings. Although, maybe that's fine? It would encourage users to fix those warnings. Regardless, we need separate checks for reasons mentioned above.

> Both the assert statement returns false in both cases when jest/mocha is installed or not. Am I testing something wrong? Need suggestions.

I don't understand what you're testing. The task here is to research what is the correct way to detect mocha, and then test that it works by running that code inside a mocha test. The jest check is already fine and doesn't need to be touched.

mayank99: From my perfunctory research, this is the best I've found: https://github.com/JoshuaKGoldberg/console-fail-test/blob/main/src/environments/mocha.ts#L19-L28

via https://redirect.github.com/mochajs/mocha/issues/5084

Maybe it would be enough to check `typeof beforeEach !== "undefined"` or `typeof describe !== "undefined"`, provided it doesn't trigger for Cypress or Playwright.

mayank99: Lets go with this for now. Can be improved later.

```js
const isMocha =
  typeof globalThis.beforeEach !== "undefined" &&
    `${globalThis.beforeEach}`.replace(/\s/g, "") ===
      "function(name,fn){suites[0].beforeEach(name,fn);}"
```

mayank99: I just followed their [getting started guide](https://mochajs.org/#getting-started) in a new project: https://stackblitz.com/edit/stackblitz-starters-93pyor?file=test%2Ftest.test.js

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99: Lets go with this for now:

```js
const isVitest = typeof globalThis.__vitest_index__ !== "undefined";
```

via [`console-fail-test/src/environments/vitest.ts`](https://github.com/JoshuaKGoldberg/console-fail-test/blob/main/src/environments/vitest.ts) (by one of mocha maintainers)

It's better than what we have right now (nothing). We can revisit later if it ever causes problems.

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -31,4 +43,4 @@ const createWarningLogger = !isDev
       };
     };
 
-export { isJest, isDev, createWarningLogger };
+export { isJest, isMocha, isVitest, isDev, createWarningLogger };
```
mayank99, line 46: Lets create a shared variable and export that instead

```js
const isUnitTest = isJest || isVitest || isMocha;
```

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -5,11 +5,23 @@
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const isJest = typeof (globalThis as any).jest !== 'undefined';
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mayank99, line 8: You can eslint disable for the whole file instead of every individual line

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99, line 1: Add a changeset since this affects user-facing code

packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -2,14 +2,22 @@
...
  *--------------------------------------------------------------------------------------------*/
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
 const isJest = typeof (globalThis as any).jest !== 'undefined';
+const isMocha =
+  typeof (globalThis as any).beforeEach !== 'undefined' &&
+  `${(globalThis as any).beforeEach}`.replace(/\s/g, '') ===
+    'function(name,fn){suites[0].beforeEach(name,fn);}';
+const isVitest = (typeof globalThis as any).__vitest_index__ !== 'undefined';
```
rohan-kadkol, line 12: nit: Maybe good to include the link of where you took this idea from. (https://github.com/JoshuaKGoldberg/console-fail-test/blob/main/src/environments/vitest.ts?rgh-link-date=2024-01-30T19%3A47%3A38Z).

E.g.: `// See: <link>`

It would also be better if you use a permalink instead of the regular link.

.changeset/twelve-rules-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added check to detect `jest`,`vitest` and `mocha` testing framework.
```
mayank99, line 5:
```suggestion
---
'@itwin/itwinui-react': patch
---

iTwinUI will now also check for `vitest` and `mocha` (in addition to `jest`) before running any code that would cause problems in these test runners.
```

mayank99 (approved):

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99: The threads in this PR do include the link, so it might be enough to link to this PR

rohan-kadkol: It would be hard to search for the repo/file link in the collapsed comments of this PR. So if you're including the PR link, then you can add the link to the repo/file in the PR description.

mayank99: I agree with updating PR description

mayank99: Unrelated note: these comments actually appear in out build output; it would be nice to strip them after build

rohan-kadkol (approved):

packages/itwinui-react/src/core/utils/functions/dev.ts
mayank99, line 12: typo?
```suggestion
const isVitest = typeof (globalThis as any).__vitest_index__ !== 'undefined';
```

PR #1809, opened 2024-01-30
Fix incorrect horizontal scroll in Table body

rohan-kadkol:
## Changes

After #1725, I noticed an unnecessary scrollbar in the Table when the window is resized (E.g. [StackBlitz playground](https://stackblitz.com/edit/github-fdpndz?file=src%2FApp.tsx) of the [Table → Basic story](https://itwin.github.io/iTwinUI/react/?story=table--basic)).

<details>

### [Example 1](https://stackblitz.com/edit/github-fdpndz?file=src%2FApp.tsx) (Table → Basic story)

Before:
* Always shows dummy div.

https://github.com/iTwin/iTwinUI/assets/45748283/2a99bc5d-409e-4e11-af0c-7a33f6ac2ffc

After:
* Only shows dummy div when needed.

image

### [Example 2](https://stackblitz.com/edit/github-bqgljj?file=src%2FApp.tsx) to test behavior when resizing columns.

Before:
* dummy div width is never updated.

https://github.com/iTwin/iTwinUI/assets/45748283/a04b9fc0-1857-4ca2-a700-d39f5567dd7f

After:
* dummy div width is updated upon table/column resize.

https://github.com/iTwin/iTwinUI/assets/45748283/e49373ad-e86a-4ffa-82c0-5c790ba101ea

* if header does not overflow, don't show dummy div

https://github.com/iTwin/iTwinUI/assets/45748283/e1a585a3-4ca9-4291-9497-8a8102e8369c

### Table 

</details>

This was happening because:
* The dummy div was added even when rows were there in the table body, and thus no fake overflow in the table-body is needed.
* The dummy div's width was not changing based on the table resizes.

To fix this:
* The dummy div in the shadow dom in the table-body is only added when:
  * There are no rows in the table
  * The header is overflowing, and thus the body also need a fake overflow to scroll the header (since header is never scrollable, even when it overflows)
* The dummy div's width for the fake overflow is recalculated whenever the table resizes

## Testing

* Existing tests pass
* Added a unit test to make sure the dummy div is:
  * added only when needed
  * added/removed and/or resized whenever the Table is resized.
* Also tested with two Table examples (in the `Before/After` section above)

## Docs

Added changeset


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -748,11 +748,34 @@ export const Table = <
...
+            ? headerRef.current?.scrollWidth > headerRef.current.clientWidth
+            : false),
+      );
+
+      // ------------------------------------------------------------------------------------------------
+      // Handle table properties, but only when table is resizable
+
+      if (!isResizable) {
+        return;
+      }
```
mayank99, line 777: Instead of changing the state inside `onTableResize`, would it be sufficient to just do something like this?

```tsx
<ShadowTemplate>
    <slot />
    {data.length === 0 && headerRef.current (
      <div
        aria-hidden
        style={{
          width:
            headerRef.current.scrollWidth > headerRef.current.clientWidth
              ? headerRef.current.scrollWidth
              : 0,
          height: 0.1
        }}
    )}
</ShadowTemplate>
```

rohan-kadkol: I made a few changes to fix some errors:

```diff
<ShadowTemplate>
    <slot />
-     {data.length === 0 && headerRef.current (
+     {data.length === 0 && headerRef.current && (
      <div
        aria-hidden
        style={{
          width:
-             headerRef.current.scrollWidth > headerRef.current.clientWidth
+            headerRef.current != null && headerRef.current.scrollWidth > headerRef.current.clientWidth
              ? headerRef.current.scrollWidth
              : 0,
          height: 0.1
        }}
      />
    )}
</ShadowTemplate>
```

---

I had tried something similar to this approach. But the problem with that is it does not re-run the logic when the `scrollWidth` and `clientWidth` changes.

So, when the table/column width changes, the dummy div's width does not update accordingly.

rohan-kadkol: Sounds good, I stopped using `ref.current` in renders. So now I created two state variables (`headerScrollWidth` & `headerClientWidth`) that are updated in `onTableResize` (https://github.com/iTwin/iTwinUI/pull/1809/commits/b6ac9efac5e54720442cc30882b4fb053f0bd228).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1071,15 +1098,17 @@ export const Table = <
...
-            <div
-              aria-hidden
-              style={{
-                // This ensures that the table-body is always the same width as the table-header,
-                // even if the table has no rows. See https://github.com/iTwin/iTwinUI/pull/1725
-                width: headerRef.current?.scrollWidth,
-                height: 0.1,
-              }}
-            />
+            {data.length === 0 && headerScrollWidth > headerClientWidth && (
```
mayank99, line 1101: Is checking for `data` enough? What if the data is filtered?

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -774,10 +790,26 @@ export const Table = <
...
+      flatHeaders,
+      instance,
+      isResizable,
+    ],
   );
   const [resizeRef] = useResizeObserver(onTableResize);
 
+  // Needed to make Table body horizontally scrollable when there are no rows
+  // See: https://github.com/iTwin/iTwinUI/issues/1204
+  // See: https://github.com/iTwin/iTwinUI/pull/1725
```
mayank99, line 805: Comments unnecessary, these are just state variables

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -748,11 +748,27 @@ export const Table = <
...
   const previousTableWidth = React.useRef(0);
   const onTableResize = React.useCallback(
     ({ width }: DOMRectReadOnly) => {
+      // ------------------------------------------------------------------------------------------------
+      // Handle header properties
+
+      setHeaderScrollWidth(headerRef.current?.scrollWidth ?? 0);
+      setHeaderClientWidth(headerRef.current?.clientWidth ?? 0);
+
+      // ------------------------------------------------------------------------------------------------
```
mayank99, line 765: nit: we typically use this divider comment for separating different components in the same file; it's not necessary here

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: Removed comments (https://github.com/iTwin/iTwinUI/pull/1809/commits/b48fa002d618c93460050604980f299616a77ec0).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -774,10 +790,23 @@ export const Table = <
...
+    ],
   );
   const [resizeRef] = useResizeObserver(onTableResize);
 
+  const [headerScrollWidth, setHeaderScrollWidth] = React.useState(
+    headerRef.current?.scrollWidth ?? 0,
+  );
+  const [headerClientWidth, setHeaderClientWidth] = React.useState(
+    headerRef.current?.clientWidth ?? 0,
+  );
```
mayank99, line 808: These should be initialized to 0. The refs will always be empty on first render, so the code never runs

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: I prefer to keep them since I feel the two sections of the function need to be clearly divided. They are important as in the future, we should not accidentally add code to the wrong section. E.g. some column resizable logic added before the `if (!isResizable) { return; }`.

If not divided using these divider comments, I am open to other ways of making the division of the sections of the function clear.

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: Initialized to 0 (https://github.com/iTwin/iTwinUI/pull/1809/commits/c3c1c1a282e058842a98f227fccee9dffabf6c18).

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: The actual words in the comments are fine. I'm talking only about the "----" comments, which are used for separating top-level components and functions (see `Alert.tsx` and `ListItem.tsx`). These separators do not help with maintainability; they are only there for improving readability.

If you want an actual way of improving maintainability, you could make a generic hook which returns these widths. There is already a `useContainerWidth` hook which you could modify.

Personally I'm ok with the code the way it is right now, other than the "----" comments. That's why this is marked as a nit.

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: Ohh, right, changed (https://github.com/iTwin/iTwinUI/pull/1809/commits/f9d06f80b61bb0f2bc63adfb1b26bfc4e17a0ef6).

mayank99 (approved):

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: After deleting the "----" comments, I feel the sections are not clear. Because the word comment lines look just like standard comments and not section divider type comments.

So I reduced the length of the "----" comments (https://github.com/iTwin/iTwinUI/pull/1809/commits/ffd79209365db846d260216b379ef9f6e6bd8a3b).

But if you really don't want those "----" comments, we can go with [`e060df3` (#1809)](https://github.com/iTwin/iTwinUI/pull/1809/commits/e060df3b6badd640936572b568d860bb8b3799da) which removes those "----" comments and makes the text comments clearer.

mayank99: > After deleting the "----" comments, I feel the sections are not clear. Because the word comment lines look just like standard comments and not section divider type comments.

That's okay. Good, even. Like I said, these comments are best suited at the top-level scope.

When we refactor Table, we can write it in a way that avoids such long functions. This way we won't even need to have these discussions about comments.

PR #1810, opened 2024-01-31
fix `Surface` alignment by not relying on `scrollbar-gutter`

### LINKED ISSUE #1801: `scrollbar-gutter` not working as a substitute for `padding-inline`

- `scrollbar-gutter: stable both-edges;` does not work as a substitute for `padding-inline` on Chrome Canary v123.0.6265.0.
Screenshot 2024-01-29 at 11 31 37 AM

#### Expected behavior

- Inline padding should always be present despite which scrollbar is used (`::-webkit-scrollbar` vs `scrollbar`).

#### Reproduction

- Open [the surface demo page](https://itwin.github.io/iTwinUI/css/surface.html) in Chrome v121 or higher since [the scrollbar styling change](https://developer.chrome.com/docs/css-ui/scrollbar-styling) was made in browsers.

#### Additional information

- Might be worthwhile adding a new helper variable like `--iui-padding-inline: var(--iui-size-s);` so it is consistent across all UI components and could be changed by our dev users.

mayank99:
## Changes

Fixes #1801 

The main change is removing `both-edges` from the value of `scrollbar-gutter`. It's fine on the right side, but we cannot rely on it for alignment purposes on the left side.

I've added a code comment with a more detailed rationale. Also see my comment on the linked issue.

## Testing

Verified that the surface body and header stay aligned regardless of scrollbar visibility.

https://github.com/iTwin/iTwinUI/assets/9084735/8e838d05-6a81-4a01-b3df-fff164124ca7

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-css/src/surface/surface.scss
+++ b/packages/itwinui-css/src/surface/surface.scss
@@ -34,12 +34,13 @@ $surface-padding-with-gutter: var(--iui-size-2xs);
   overflow-y: auto;
 
   &[data-iui-padded='true'] {
-    scrollbar-gutter: stable both-edges;
-    padding-inline: $surface-padding-with-gutter;
-    padding-block: $surface-padding;
+    padding: $surface-padding;
 
-    @supports not (scrollbar-gutter: stable) {
-      padding-inline: $surface-padding;
-    }
+    // Scrollbars might not always occupy space even with a stable gutter, so this can still
+    // result in (tolerable) differences on the right side, depending on scrollbar visibility.
+    // More importantly, the left side will remain aligned, because we are not using a gutter there.
+    // See: https://github.com/iTwin/iTwinUI/issues/1801
+    scrollbar-gutter: stable;
+    padding-inline-end: $surface-padding-with-gutter;
   }
 }
--- a/.changeset/rude-rockets-occur.md
+++ b/.changeset/rude-rockets-occur.md
@@ -1,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Fixed `Surface.Body` not being visually aligned with `Surface.Header` depending on scrollbar visibility.


packages/itwinui-css/src/surface/surface.scss
```diff
@@ -34,12 +34,13 @@ $surface-padding-with-gutter: var(--iui-size-2xs);
...
-      padding-inline: $surface-padding;
-    }
+    padding: $surface-padding;
+
+    // Scrollbars might not always occupy space even with a stable gutter, so this can still
+    // result in (tolerable) differences on the right side, depending on scrollbar visibility.
+    // More importantly, the left side will remain aligned, because we are not using a gutter there.
+    // See: https://github.com/iTwin/iTwinUI/issues/1801
+    scrollbar-gutter: stable;
+    padding-inline-end: $surface-padding-with-gutter;
```
mayank99, line 44: @FlyersPh9 Can you verify that this is what we want? I used a smaller padding on the right side, but we could even keep the larger one.

The smaller padding can potentially be too small for overlay scrollbars, while the larger padding can potentially be too large for non-overlay scrollbars (more common on Chrome Windows, which has the largest share of our users).

https://github.com/iTwin/iTwinUI/assets/9084735/13bf0151-0634-4071-9f20-3ae043f4ec3c

rohan-kadkol: Should we have some JS solution to detect whether space is given to the scrollbar or not?

e.g. when space is given to a scrollbar + `scrollbar-gutter: stable`, I think `clientWidth < offsetWidth`. And when no space is given, `clientWidth === offsetWidth`.

FlyersPh9 (approved):

packages/itwinui-css/src/surface/surface.scss
mayank99: JS is completely unnecessary here. Neither of the two issues I brought up are bugs or dealbreakers; I'm just asking for opinions which one we want to optimize for.

rohan-kadkol: I mean to say that we can optimize both with a JS solution.

i.e. use smaller `inline-end` padding when there the scrollbar takes up space and use larger/regular padding when there the scrollbar takes up no space. That way, the end padding always stays consistent regardless of whether the scrollbar is given space.

mayank99: Can we please not overcomplicate it with unnecessary JS? We can't always optimize for every situation, and I've mentioned multiple times that this inconsistency is tolerable. Simplicity is a vastly more important goal than consistency.

rohan-kadkol: Yes, I know this JS is extra weight and overcomplicates the small issue. My comment was just a suggestion, not a requirement.

Out of the two padding options, I prefer the option you used currently. Having smaller than usual inline-end padding when the scrollbar has no space (overlay) is not too much of a deal breaker, since like you mentioned, most of our users have scrollbars that have space.

rohan-kadkol (approved):

packages/itwinui-css/src/surface/surface.scss
FlyersPh9: I do prefer to keep this CSS only at least for this fix now.

Screenshots of https://itwin.github.io/iTwinUI/1810/css/surface.html

**Chrome Canary 123:**
Screenshot 2024-01-31 at 3 00 09 PM

**Safari:**
Screenshot 2024-01-31 at 3 02 20 PM

And I assume you are trying to avoid simply `padding: $surface-padding;` because if there is no scrollbar & then a scrollbar appears because of resizing or something, all of the body content will shift inwards, right?

mayank99: Yes, exactly.

mayank99: All of the content will *not* shift inwards as a result of resizing/content change because we are still using `scrollbar-gutter: stable` on the right edge. But yes I'm avoiding `padding: $surface-padding` because it does shift the content inwards too much when space is reserved for the scrollbar, especially when the content is not overflowing.

FlyersPh9: Ahh, so trying to avoid this situation when there is no scrollbar present.
Screenshot 2024-01-31 at 3 09 38 PM

FlyersPh9: Tried playing around with giving the scrollbar gutter a color (border would have been nicer), but when the content can't scroll it reserves the space, but hides the gutter color.  So this is not an option (also doesn't look good).

Screenshot 2024-01-31 at 3 13 05 PM
Screenshot 2024-01-31 at 3 13 15 PM

mayank99: Sounds good, leaving the code unchanged and updated PR description to mention this thread.

Issue #1814, opened 2024-02-01
Scrollbar regression on Chrome and Edge


HaveSpacesuit:
The styling added to scrollbars seems to have reverted with newer versions of Chrome and Edge.

https://itwin.github.io/iTwinUI/react/?story=table--basic

**Chrome**
![image](https://github.com/iTwin/iTwinUI/assets/15036346/ab3484d7-0753-4673-ae17-74efcd24ff8f)

**Edge**
![image](https://github.com/iTwin/iTwinUI/assets/15036346/e0ba8203-b3d4-475c-a9f4-aea0e9d58730)

**Firefox**
![image](https://github.com/iTwin/iTwinUI/assets/15036346/d0bc1a36-1916-47d3-b84a-94bda3d66aba)

Disabling the `scrollbar-color` and `scrollbar-width` added by iTwin fixes the issue.

HaveSpacesuit:
Looks like this is a [known regression in Chrome 121](https://syntackle.com/blog/changes-to-scrollbar-styling-in-chrome-121/). (Or maybe not a regression, but a change which causes us problems.)

mayank99:
Not a regression, this is just how scrollbars will look from Chrome 121 onwards.

mayank99:
Actually there is one part we should address: the table header should not display the track buttons, because there will never be a scrollbar there.
![image](https://github.com/iTwin/iTwinUI/assets/9084735/a4ab28e6-aede-43a1-88d3-c0378423e36e)

mayank99:
@HaveSpacesuit I cannot reproduce this issue on Windows. Where/how are you seeing scrollbar tracks in the table header? It looks fine for me:

![image](https://github.com/iTwin/iTwinUI/assets/9084735/648d2cc6-4d38-410d-abb1-20def66a529c)![image](https://github.com/iTwin/iTwinUI/assets/9084735/ba5d8207-7ca3-4b00-bfc1-2695f80d8527)

mayank99:
Now fixed in [3.3.4](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.3.4) 🎉 Thanks for reporting this.


| Before | After |
| --- | --- |
[...]

PR #1816, opened 2024-02-01
fix table header showing dummy scrollbar

mayank99:
## Changes

Fixes https://github.com/iTwin/iTwinUI/issues/1814#issuecomment-1921876655

After some experimenting, I found that I was able to remove the scrollbar track and spin buttons (while preserving alignment) by setting `scrollbar-color` to transparent.

## Testing

Tested in Chrome (and Edge) 121 on Windows. Other platforms unaffected.

| Before | After |
| --- | --- |
| image | image |

## Docs

Added changeset.


.changeset/hip-wasps-accept.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-react': patch
+'@itwin/itwinui-css': patch
+---
+
+Fixed an issue where `Table` was showing a dummy scrollbar track inside the table header in Chrome 121.
```
rohan-kadkol, line 6: Should we mention Chrome 121 *+ Windows*?

```suggestion
Fixed an issue where `Table` was showing a dummy vertical scrollbar track inside the table header in Chrome 121 + Windows.
```

rohan-kadkol: Actually, this also happened to me on Edge 121 + Windows.

rohan-kadkol (commented):
> Tested in Chrome 121 on Windows. Other platforms unaffected.

I was able to reproduce this on Chrome 121 + Windows and Edge 121 + Windows.

.changeset/hip-wasps-accept.md
mayank99: Edge is just Chromium under the hood, so not necessary, but mentioning Windows makes sense 👍  [`13f29d6`](https://github.com/iTwin/iTwinUI/pull/1816/commits/13f29d6fcbbd9407f06aba6ef3f28122f51c94fe)

rohan-kadkol (approved):

mayank99:
> I was able to reproduce this on Chrome 121 + Windows and Edge 121 + Windows.

Yes, Edge also uses Chromium, which added support for standard scrollbar properties in v121. I verified it's fixed there too.

| Before | After |
| --- | --- |
| image | image |

PR #1817, opened 2024-02-01
migrate demos from codesandbox to codepen

mayank99:
## Changes

Replaced the `OpenInCodesandbox` component with `OpenInCodePen`, which uses the [codepen prefill api](https://blog.codepen.io/documentation/prefill/).

The embedded demos are unchanged; this only affects the "external" viewing experience.

A couple important notes:

1. I'm modifying the JSX source to replace bare-module imports with url imports (using https://esm.sh/). I tried using [import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) but couldn't get it to work (CodePen likely doesn't support it).

2. I'm using "TypeScript" as the preprocessor. It works fine but has syntax highlighting issues with JSX syntax.
   image

   In subsequent PRs, we can convert all our examples to plain JSX to get around this issue (using "Babel" as the preprocessor instead). As a bonus, this will also make the code more readable.

3. I have very little CSS going on currently. In subsequent PRs, we can remove inline styles and write real CSS, to be shown in the CodePen CSS panel. This will further improve the readability of our code examples.

## Testing

Try it out: https://deploy-preview-1817--itwinui-previews.netlify.app/docs/

## Docs

N/A


apps/website/src/components/OpenInCodePen.astro
```diff
@@ -0,0 +1,120 @@
...
+    </ThemeProvider>
+  );
+`
+  .replaceAll(`@itwin/itwinui-react`, `https://esm.sh/@itwin/itwinui-react@3`)
+  .replaceAll(`@itwin/itwinui-icons-react`, `https://esm.sh/@itwin/itwinui-icons-react@3`)
+  .replaceAll(`'react'`, `'https://esm.sh/react@18'`)
+  .replaceAll(`react-dom`, `https://esm.sh/react-dom@18`);
+---
+
+<form action='https://codepen.io/pen/define' method='POST' target='_blank' {...rest}>
```
rohan-kadkol, line 45: Do we need `rel="noopener noreferrer"`?

apps/website/src/components/OpenInCodePen.astro
```diff
@@ -0,0 +1,120 @@
...
+  createRoot(document.body).render(
+    <ThemeProvider className='app-wrapper'>
+      <App />
+    </ThemeProvider>
+  );
+`
+  .replaceAll(`@itwin/itwinui-react`, `https://esm.sh/@itwin/itwinui-react@3`)
+  .replaceAll(`@itwin/itwinui-icons-react`, `https://esm.sh/@itwin/itwinui-icons-react@3`)
+  .replaceAll(`'react'`, `'https://esm.sh/react@18'`)
+  .replaceAll(`react-dom`, `https://esm.sh/react-dom@18`);
```
rohan-kadkol, line 42: nit: For consistency and robustness, you can use wrap `react-dom` in quotes, similar to how you did for `react`.


```suggestion
  .replaceAll(`'react-dom'`, `'https://esm.sh/react-dom@18'`);
```

apps/website/src/components/OpenInCodePen.astro
```diff
@@ -0,0 +1,120 @@
...
+  import { ThemeProvider } from '@itwin/itwinui-react';
+
+  createRoot(document.body).render(
+    <ThemeProvider className='app-wrapper'>
+      <App />
+    </ThemeProvider>
+  );
+`
+  .replaceAll(`@itwin/itwinui-react`, `https://esm.sh/@itwin/itwinui-react@3`)
+  .replaceAll(`@itwin/itwinui-icons-react`, `https://esm.sh/@itwin/itwinui-icons-react@3`)
```
rohan-kadkol, line 40: The `Non ideal state` demos are not working since `@itwin/itwinui-illustrations-react` is not handled.

rohan-kadkol: `itwinui-icons-react`'s latest version is 2.8.0

```suggestion
  .replaceAll(`@itwin/itwinui-icons-react`, `https://esm.sh/@itwin/itwinui-icons-react@2`)
```

~But even with this, the demos that import from `https://esm.sh/@itwin/itwinui-icons-react@2` show a blank screen. E.g. `Tree`, `Tile`, `TransferList`, etc.. So I think this need to be investigated further.~ (This was an unrelated issue on my end.)

apps/website/src/components/OpenInCodePen.astro
mayank99: I did it this way on purpose, to account for `react-dom/client`.

I also used `'react'` (vs `react`) on purpose, because `react` is a more common word that could be used elsewhere.

rohan-kadkol (approved):

PR #1818, opened 2024-02-02
Imports useIsomorphicLayoutEffect as useLayoutEffect

Ben-Pusey-Bentley:
## Changes

Imports useIsomorphicLayoutEffect as useLayoutEffect. This change allows the react hooks linter to correctly identify the dependencies  that the effect should have passed in. This PR also updates the style guide to reflect this change.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -168,11 +168,13 @@ window.clearTimeout(1);
 document.createElement('div');
 ```
 
-### Use `useIsomorphicLayoutEffect` instead of `useLayoutEffect`
+### Import `useIsomorphicLayoutEffect as useLayoutEffect` instead of using `React.useLayoutEffect`
 
 ```jsx
 // Good
-useIsomorphicLayoutEffect(() => {});
+import {useIsomorphicLayoutEffect as useLayoutEffect} from '../utils/index.js'
+
+useLayoutEffect(() => {});
 ```
 
 ```jsx
--- a/packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
+++ b/packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
@@ -8,7 +8,7 @@ import { CarouselContext } from './CarouselContext.js';
 import {
   getWindow,
   useMergedRefs,
-  useIsomorphicLayoutEffect,
+  useIsomorphicLayoutEffect as useLayoutEffect,
   Box,
 } from '../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../utils/index.js';
@@ -41,7 +41,7 @@ export const CarouselSlider = React.forwardRef((props, ref) => {
     [children, idPrefix],
   ) as React.ReactNode[];
 
-  useIsomorphicLayoutEffect(() => {
+  useLayoutEffect(() => {
     setSlideCount(items.length);
   }, [items.length, setSlideCount]);
 
--- a/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
@@ -11,7 +11,7 @@ import type { Input } from '../Input/Input.js';
 import {
   mergeRefs,
   useLatestRef,
-  useIsomorphicLayoutEffect,
+  useIsomorphicLayoutEffect as useLayoutEffect,
   AutoclearingHiddenLiveRegion,
   useId,
 } from '../utils/index.js';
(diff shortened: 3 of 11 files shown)


STYLEGUIDE.md
```diff
@@ -168,11 +168,13 @@ window.clearTimeout(1);
...
 
 ```jsx
 // Good
-useIsomorphicLayoutEffect(() => {});
+import {useIsomorphicLayoutEffect as useLayoutEffect} from '../utils/index.js'
+
+useLayoutEffect(() => {});
 ```
 
 ```jsx
```
rohan-kadkol, line 180: The bad examples could also include an example of using `useIsomorphicLayoutEffect` directly.

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
```diff
@@ -8,7 +8,7 @@ import { CarouselContext } from './CarouselContext.js';
 import {
   getWindow,
   useMergedRefs,
-  useIsomorphicLayoutEffect,
+  useIsomorphicLayoutEffect as useLayoutEffect,
```
rohan-kadkol, line 11: Just thinking out loud: What is better, importing as an alias every time; or naming the export itself as `useLayoutEffect` and then no import aliases are required?

rohan-kadkol (commented):
Nice first PR!

Since you updated `STYLEGUIDE.md`, you can mention that under the "Docs" section of the PR description.

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
mayank99: It introduces potential confusion about `React.useLayoutEffect` vs `useLayoutEffect`. Ideally we'd have a custom lint rule that that flags usages of `React.useLayoutEffect`. 

For now, I'm ok with renaming the export to `useLayoutEffect`. 👍

Make sure to update the STYLEGUIDE after that.

Ben-Pusey-Bentley: Latest commit renames the export to `useLayoutEffect`.

STYLEGUIDE.md
Ben-Pusey-Bentley: The name of the export has changed to `useLayoutEffect`, so using `useIsomorphicLayoutEffect` directly shouldn't be possible anymore.

packages/itwinui-react/src/core/utils/hooks/useIsomorphicLayoutEffect.tsx
```diff
@@ -9,5 +9,5 @@ import * as React from 'react';
  *
  * @see https://fb.me/react-uselayouteffect-ssr
  */
-export const useIsomorphicLayoutEffect =
+export const useLayoutEffect =
```
mayank99, line 12: this might be a nit, but i think it would be more clear to keep the original name and rename it when exporting

```jsx
const useIsomorphicLayoutEffect =
  typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;

export { useIsomorphicLayoutEffect as useLayoutEffect };
```

rohan-kadkol: Additionally, you could also add a short comment to explain why we are exporting it with a different name.

Ben-Pusey-Bentley: Changed. Comment was added to the larger comment about the function.

STYLEGUIDE.md
```diff
@@ -168,11 +168,13 @@ window.clearTimeout(1);
 document.createElement('div');
 ```
 
-### Use `useIsomorphicLayoutEffect` instead of `useLayoutEffect`
+### Import `useLayoutEffect` from `utils` instead of using `React.useLayoutEffect`
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Added explanation.

packages/itwinui-react/src/core/utils/hooks/useIsomorphicLayoutEffect.tsx
mayank99: Looks good.

Just a note about PR etiquette: please refrain from self-resolving PR comments. Let the reviewer verify the change and wait for them to resolve it.

mayank99 (approved):
LGTM

STYLEGUIDE.md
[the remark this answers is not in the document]
rohan-kadkol: nit: Maybe something shorter than my previous suggestion would be better?

```diff
- ### Import `useLayoutEffect` from `utils` instead of using `React.useLayoutEffect`, since `useLayoutEffect` from `utils` is SSR safe
+ ### Import `useLayoutEffect` (SSR-safe) from `utils` instead of using `React.useLayoutEffect` (client only)
```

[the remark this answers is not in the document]
Ben-Pusey-Bentley: Changed in latest commit.

rohan-kadkol (approved):

PR #1821, opened 2024-02-05
add new props for customizing internal `Modal` elements

mayank99:
## Changes

Addresses second point of https://github.com/iTwin/iTwinUI/issues/1368#issuecomment-1911812509

## Testing

Added unit tests for new props.

## Docs

Added changesets.

### DIFF
--- a/packages/itwinui-react/src/core/Modal/Modal.test.tsx
+++ b/packages/itwinui-react/src/core/Modal/Modal.test.tsx
@@ -240,3 +240,25 @@ it('should reset body overflow on closing and unmounting', () => {
   unmount();
   expect(document.body.style.overflow).not.toEqual('hidden');
 });
+
+it('should allow passing arbitrary props to internal elements', () => {
+  const { container } = render(
+    <Modal
+      isOpen={true}
+      title='The title'
+      className='modal'
+      wrapperProps={{ className: 'wrapper' }}
+      backdropProps={{ className: 'backdrop' }}
+      titleBarProps={{ className: 'title-bar' }}
+    >
+      …
+    </Modal>,
+  );
+
+  const wrapper = container.querySelector('.wrapper') as HTMLElement;
+  const backdrop = wrapper.querySelector('.backdrop') as HTMLElement;
+  expect(backdrop).toBeTruthy();
+  const modal = wrapper.querySelector('.modal') as HTMLElement;
+  const titleBar = modal.querySelector('.title-bar') as HTMLElement;
+  expect(titleBar).toHaveTextContent('The title');
+});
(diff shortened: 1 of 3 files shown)


packages/itwinui-react/src/core/Modal/Modal.tsx
```diff
@@ -99,10 +114,11 @@ export const Modal = React.forwardRef((props, forwardedRef) => {
       setFocus
       ref={forwardedRef}
```
mayank99, line 115: This part is very annoying: `ref` and `...rest` props are applied on different elements. I thought we fixed it in v3, but evidently not.

packages/itwinui-react/src/core/Modal/Modal.test.tsx
```diff
@@ -240,3 +240,25 @@ it('should reset body overflow on closing and unmounting', () => {
...
+    >
+      …
+    </Modal>,
+  );
+
+  const wrapper = document.querySelector('.wrapper') as HTMLElement;
+  const backdrop = wrapper.querySelector('.backdrop') as HTMLElement;
+  expect(backdrop).toBeTruthy();
+  const modal = wrapper.querySelector('.modal') as HTMLElement;
+  const titleBar = modal.querySelector('.title-bar') as HTMLElement;
```
rohan-kadkol, line 262: nit: Is there any advantage of spliting into `modal` and `titleBar` instead of directly using `titleBar`?

```
const titleBar = wrapper.querySelector('.modal > .title-bar') as HTMLElement;
```

packages/itwinui-react/src/core/Modal/Modal.test.tsx
```diff
@@ -240,3 +240,25 @@ it('should reset body overflow on closing and unmounting', () => {
...
+      className='modal'
+      wrapperProps={{ className: 'wrapper' }}
+      backdropProps={{ className: 'backdrop' }}
+      titleBarProps={{ className: 'title-bar' }}
+    >
+      …
+    </Modal>,
+  );
+
+  const wrapper = document.querySelector('.wrapper') as HTMLElement;
```
rohan-kadkol, line 258: nit: Should we check if `wrapper` is truthy?

rohan-kadkol (approved):

packages/itwinui-react/src/core/Modal/Modal.test.tsx
mayank99: I'm using `wrapper.querySelector` below, which would fail if it was not truthy.

packages/itwinui-react/src/core/Modal/Modal.test.tsx
rohan-kadkol: Sounds good, I was just curious. Then how it is is good.

PR #1824, opened 2024-02-06
use `LinkAction` inside `Tag`

mayank99:
## Changes

When working on demos for interactive `Tag`s (#1815), I realized that the type error when simultaneously using `onClick` and `onRemove` is not enough. The component still renders (no runtime error) with nested buttons. This has a risk of the developer suppressing the type error without giving it further thought.

This PR remedies that by conditionally using `LinkAction` only when both `onClick` and `onRemove` are passed. `LinkAction` will avoid invalid markup while preserving the same clickable area, i.e. clicking anywhere on the tag invokes `onClick` and clicking on the ❌ button invokes `onRemove`. I probably still wouldn't recommend this pattern, but it's better than before.

Also `LinkAction` itself has been updated to automatically render `<button>` (rather than `<a>`) by default when `href` is not passed.

## Testing

Removed type test + added new unit test (runtime) ensuring proper markup.

Tested in playground as well. Everything works correctly, including hover and focus states.

image

## Docs

Updated/added changesets and jsdocs.

Updated docs page to mention accessibility considerations.

### DIFF
--- a/apps/website/src/content/docs/linkaction.mdx
+++ b/apps/website/src/content/docs/linkaction.mdx
@@ -13,7 +13,7 @@ group: utilities
 
 This component is useful in a wide range of situations when you need to increase the clickable area of an element (e.g. a card, a list item, etc.) without compromising semantics. This means you can keep the link text concise and still make the entire area clickable, even if it includes other elements or extra padding.
 
-`LinkAction` is used internally in [`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), and [`ExpandableBlock`](expandableblock) components, and is also available for use in external components.
+`LinkAction` is used internally in other components ([`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), [`ExpandableBlock`](expandableblock), etc), and is also available for use in external components.
 
 ## Usage
 
--- a/apps/website/src/content/docs/linkaction.mdx
+++ b/apps/website/src/content/docs/linkaction.mdx
@@ -13,7 +13,7 @@ group: utilities
 
 This component is useful in a wide range of situations when you need to increase the clickable area of an element (e.g. a card, a list item, etc.) without compromising semantics. This means you can keep the link text concise and still make the entire area clickable, even if it includes other elements or extra padding.
 
-`LinkAction` is used internally in other components ([`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), [`ExpandableBlock`](expandableblock), etc), and is also available for use in external components.
+`LinkAction` is used internally in other components ([`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), [`ExpandableBlock`](expandableblock), etc.), and is also available for use in external components.
 
 ## Usage
 
--- a/apps/website/src/content/docs/tag.mdx
+++ b/apps/website/src/content/docs/tag.mdx
@@ -13,17 +13,13 @@ thumbnail: Tag
 
 Tags are user-generated keywords used to label, categorize, or organize items and files. A group of multiple tags is a standard way of labeling an item to make it easily discoverable by peers who may not know the exact title of said item. A single tag may be mapped to more than one project at once if these projects fit in similar categories. Tags may also be used as a way to filter out content; in that context, the tags are pre-existing, and users select tags that fit their search to narrow down the results.
 
-## Variants
-
-### Default
-
-Default tags can be added and removed by a user. Tags adapt their size to the length of the label inside, making it easier to tell them apart due to their different sizes.
+## Usage
 
 <LiveExample src='Tag.default.tsx'>
   <AllExamples.TagDefaultExample client:load />
 </LiveExample>
 
-### Basic
+### Basic tags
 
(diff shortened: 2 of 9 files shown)


apps/website/src/content/docs/tag.mdx
```diff
@@ -13,28 +13,37 @@ thumbnail: Tag
```
rohan-kadkol, line 13: `TagMainExample` can probably show a tag with `onClose` too, instead of two with `onRemove`.

apps/website/src/content/docs/tag.mdx
```diff
@@ -13,28 +13,37 @@ thumbnail: Tag
...
 
-## Props
+### Props
 
 <PropsTable path={frontmatter.propsPath} />
 
+### Usage guidelines
+
+- Tags are meant to be used for display purposes. While interactivity is possible, it is not the primary use case.
+- Use [basic](#basic) tags when you need to make tags read-only.
```
rohan-kadkol, line 37: Since the heading name changed, the link needs to be updated.

```suggestion
- Use [basic](#basic-tags) tags when you need to make tags read-only.
```

apps/website/src/content/docs/tag.mdx
```diff
@@ -13,28 +13,37 @@ thumbnail: Tag
...
+
+## Accessibility
+
+When the `onClick` prop is passed to `Tag`, it is automatically rendered as a `<button>` element. This gives the tag proper semantics and makes it keyboard interactable.
+
+When the `onRemove` prop is passed to `Tag`, a remove button is rendered inside the tag. This button has an `aria-label` set to `"Delete tag"`, which can be customized using the `removeButtonProps` prop.
+
+If both `onClick` and `onRemove` props are passed, then the tag label (rather than the tag itself) will be rendered as a button, to avoid invalid markup (button nested inside another button). This is achieved using the [`LinkAction`](linkaction) component.
```
rohan-kadkol, line 45: Since `Tag` now uses `LinkAction` internally, we can mention this on the `LinkAction` docs too, in the sentence, "LinkAction is used internally in …".

Or maybe that sentence can be rephrased with something like "etc." to say that more components than the ones listed use `LinkAction` internally.

apps/website/src/content/docs/tag.mdx
```diff
@@ -71,9 +80,8 @@ The scrollable container is much like the truncating container, except it allows
 
 <PropsTable path={'@itwin/itwinui-react/esm/core/Tag/TagContainer.d.ts'} />
 
-## Usage
+### Usage guidelines
```
rohan-kadkol, line 83: Is it alright to have two headings with the same name?

packages/itwinui-react/src/core/Tag/Tag.tsx
```diff
@@ -7,12 +7,31 @@ import * as React from 'react';
...
+   *
+   * When this prop is passed, the tag will be rendered as a button.
+   */
+  onClick?: React.MouseEventHandler;
+  /**
+   * Callback function that handles click on the remove ("❌") button.
+   * If not passed, the remove button will not be shown.
+   *
+   * If both `onClick` and `onRemove` are passed, then the tag label (rather than the tag itself)
+   * will be rendered as a button, to avoid invalid markup (nested buttons).
```
rohan-kadkol, line 28: Should the `onClick`'s JSDocs also mention this?

apps/website/src/content/docs/tag.mdx
mayank99: I don't think so. Tags' main purpose is not to be used with `onClick`.

apps/website/src/content/docs/tag.mdx
mayank99: Yeah it's fine. They are sub-headings

apps/website/src/content/docs/tag.mdx
mayank99: Updated. [`8aa3079`](https://github.com/iTwin/iTwinUI/pull/1824/commits/8aa307984b8dc9e2419f7ddc69d6b0e6b87256f2)

apps/website/src/content/docs/linkaction.mdx
```diff
@@ -13,7 +13,7 @@ group: utilities
 
 This component is useful in a wide range of situations when you need to increase the clickable area of an element (e.g. a card, a list item, etc.) without compromising semantics. This means you can keep the link text concise and still make the entire area clickable, even if it includes other elements or extra padding.
 
-`LinkAction` is used internally in [`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), and [`ExpandableBlock`](expandableblock) components, and is also available for use in external components.
+`LinkAction` is used internally in other components ([`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), [`ExpandableBlock`](expandableblock), etc), and is also available for use in external components.
```
rohan-kadkol, line 16: nit

```suggestion
`LinkAction` is used internally in other components ([`Tile`](tile#actionable), [`ListItem`](list#with-listitemaction), [`ExpandableBlock`](expandableblock), etc.), and is also available for use in external components.
```

rohan-kadkol (approved):

PR #1826, opened 2024-02-07
update deps

mayank99:
## Changes

Updated a few (development-only) dependencies. Too many to list all, but here are some notable ones:
- latest version of `turbo`, includes support for auto-loading `env` files
- removed all overrides that are no longer necessary
- removed some unnecessary dependencies like `@emotion/styled`

## Testing

Ensured that build/dev/test works locally, formatting/linting works in IDE, and CI passes (including audit).

## Docs

n/a


.env.development
mayank99, line 1: open to suggestions on how to best handle env variables/files.

the approach i went with: `.env.development` gets checked in and `.env.local` stays gitignored (it will contain things like algolia id and any other api keys in the future)

package.json
```diff
@@ -42,40 +42,28 @@
...
+    "lint-staged": "^15.2.2",
+    "patch-package": "~8.0.0",
     "prettier": "~3.1.0",
-    "rimraf": "^3.0.2",
+    "rimraf": "^5.0.5",
     "serve": "^14.2.1",
-    "turbo": "~1.6.3"
+    "turbo": "~1.12.3"
   },
   "resolutions": {
```
rohan-kadkol, line 56: Just curious, how did you know/decide which resolutions were removable?

rohan-kadkol (approved):

rohan-kadkol (commented):
Running `yarn lint` fails in `react-workshop` because of a dist file:

image

So maybe now or in a future PR, we need to add `'dist/**',` to `.eslintrc.js`'s `ignorePatterns`

package.json
mayank99: I went through the lockfile and also ran `npm ls` for each dep. Removed the ones that were no longer necessary. Typically this means the dep is no longer being used, or the requested ranges are already satisfactory.

PR #1833, opened 2024-02-07
remove local dependency on changesets

mayank99:
## Changes

#1831 didn't work, so I've reverted it.

WIP

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,11 +38,11 @@
     "approve:css": "yarn workspace css-workshop approve",
     "approve:react": "yarn workspace react-workshop approve",
     "clean": "turbo run clean && rimraf node_modules",
+    "changeset": "npx -y @changesets/cli",
     "release": "yarn build --filter={./packages/*}... && changeset publish"
   },
   "dependencies": {},
   "devDependencies": {
-    "@changesets/cli": "^2.27.1",
     "chokidar-cli": "^3.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,11 +38,11 @@
     "approve:css": "yarn workspace css-workshop approve",
     "approve:react": "yarn workspace react-workshop approve",
     "clean": "turbo run clean && rimraf node_modules",
-    "release": "yarn build --filter={./packages/*}... && changeset publish"
+    "changeset": "npx -y @changesets/cli",
+    "release": "yarn build --filter={./packages/*}... && yarn changeset publish"
```
rohan-kadkol, line 42: non-blocking: Is there any reason for using both `yarn` and `npx`? Would it be better to just use one?

rohan-kadkol (approved):

package.json
mayank99: `yarn changeset` is a shorthand for what was previously `changeset`.

These commands worked before and continue to work now:
```
yarn changeset
yarn changeset version
yarn changeset publish
```

The `npx -y` is an implementation detail of the `changeset` script (L43).

mayank99 (commented):
Third time's the charm! https://github.com/iTwin/iTwinUI/actions/runs/7819317824/job/21331508305

PR #1836, opened 2024-02-08
move dialog resizers into shadow dom

mayank99:
## Changes

This is a slight refactor of the `Resizer` component used in `Dialog`.
- It is now rendered inside shadow DOM, and its styles have been updated to use CSS grid for positioning. As an added benefit of shadow DOM, I was able to use a plain `<style>` here (vs inline styles) without causing performance issues.
- User-facing change: It now protrudes outside the edge of the dialog, as originally intended (before it was reversed in #1424 due to `overflow`, which was recently removed in #1795).
    - Screenshot below visualizes the resizers (corners in hotpink, perpendiculars in wheat); the corner resizers appear on top using a higher z-index.
      image

Internal change: Extracted out `ShadowTemplate` into a shared component and renamed to `ShadowRoot`. Updated the logic inside to correctly handle isomorphic rendering (via [declarative syntax](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#declaratively_with_html)). Also added a `flushSync` to immediately force a re-render (needed by `createPortal`); this ensures that any layout measurements are immediately reflected correctly. This is important because the Dialog resizing logic immediately measures the DOM inside a `useLayoutEffect` and was reporting incorrect measurements if the shadow root content was not present on first mount.

## Testing

Updated affected unit tests. Since `attachShadow` is called inside `queueMicrotask`, I had to use fake timers.

Manually tested that resizing works correctly. Inspect the [Resizable Dialog story](https://itwin.github.io/iTwinUI/1836/react/?story=dialog--draggable-and-resizable) and play with it in devtools to understand the new changes. 

## Docs

Added changeset for the one-user facing change.

### DIFF
--- a/packages/itwinui-react/src/core/Table/Table.test.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.test.tsx
@@ -3452,46 +3452,9 @@ it('should sync body horizontal scroll with header scroll', () => {
   expect(body.scrollLeft).toBe(0);
 });
 
-it('should add a shadow tree to table-body.', () => {
-  const columnWidths = [400, 600, 200];
-  const columnWidthsSum = columnWidths.reduce((a, b) => a + b, 0);
-
-  // Mocking the scrollWidth of the table header
-  vi.spyOn(HTMLDivElement.prototype, 'scrollWidth', 'get').mockReturnValue(
-    columnWidthsSum,
-  );
-
-  const { container } = renderComponent({
-    columns: [
-      {
-        Header: 'Name',
-        accessor: 'name',
-        id: 'name',
-        width: columnWidths[0],
-      },
-      {
-        Header: 'Description',
-        accessor: 'description',
-        id: 'description',
-        width: columnWidths[1],
-      },
-      {
-        Header: 'View',
-        Cell: () => <>View</>,
-        id: 'view',
-        width: columnWidths[2],
-      },
-    ],
-  });
-
-  // body serves as the shadow host
-  const host = container.querySelector('.iui-table-body') as HTMLDivElement;
-
-  const slot = host?.shadowRoot?.querySelector('slot');
-  expect(slot).toBeTruthy();
-});
-
 it('should have a shadow tree in table-body that has a dummy div only when needed', () => {
+  vi.useFakeTimers({ toFake: ['queueMicrotask'] });
+
   const columnWidths = [400, 600, 200];
   const columnWidthsSum = columnWidths.reduce((a, b) => a + b, 0);
 
@@ -3528,6 +3491,7 @@ it('should have a shadow tree in table-body that has a dummy div only when neede
     isResizable: true,
     columnResizeMode: 'expand',
   });
+  act(() => vi.runAllTicks());
 
   // Initial render
   triggerResize({ width: columnWidthsSum } as DOMRectReadOnly);
--- a/packages/itwinui-react/src/core/Table/Table.test.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.test.tsx
@@ -3500,6 +3500,9 @@ it('should have a shadow tree in table-body that has a dummy div only when neede
   let host = container.querySelector('.iui-table-body') as HTMLDivElement;
   expect(host).toBeTruthy();
 
+  const slot = host?.shadowRoot?.querySelector('slot');
+  expect(slot).toBeTruthy();
+
   // When clientWidth >= scrollWidth, the dummy div should not be rendered
   const htmlScrollWidthMock = vi
     .spyOn(HTMLDivElement.prototype, 'scrollWidth', 'get')
--- a/packages/itwinui-react/src/core/utils/components/Resizer.tsx
+++ b/packages/itwinui-react/src/core/utils/components/Resizer.tsx
@@ -33,6 +33,8 @@ export type ResizerProps = {
 /**
  * Component that allows to resize parent element.
  * Parent must have `position: relative`.
+ *
+ * Ideally should be used within a shadow root.
  * @private
  * @example
  * const ref = React.useRef<HTMLDivElement>(null);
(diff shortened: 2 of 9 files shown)


FlyersPh9:
I don't think I'm the best person to review the `.tsx` changes, but I will say the resize interaction does feel much nicer. 👍

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -0,0 +1,64 @@
...
+import * as ReactDOM from 'react-dom';
+
+const isBrowser = typeof document !== 'undefined';
+const supportsDSD =
+  isBrowser && 'shadowRootMode' in HTMLTemplateElement.prototype;
+
+/**
+ * Wrapper around `<template>` element that attaches shadow root to its parent
+ * and moves its children into the shadow root.
+ */
```
rohan-kadkol, line 15: nit: Since this is a private util component, maybe we can mark it with `@private`.

```suggestion
 *
 * @private
 */
```

packages/itwinui-react/src/core/utils/components/Resizer.tsx
```diff
@@ -216,99 +234,80 @@ export const Resizer = (props: ResizerProps) => {
...
         onPointerDown={onResizePointerDown}
-        style={{
-          position: 'absolute',
-          top: 8,
-          left: 0,
-          bottom: 8,
-          width: 8,
-          cursor: 'w-resize',
-        }}
+        style={{ cursor: 'w-resize' }}
```
rohan-kadkol, line 272: nit: Is there any advantage with defining this also in `resizerStyles`?

packages/itwinui-react/src/core/utils/components/Resizer.tsx
```diff
@@ -216,99 +234,80 @@ export const Resizer = (props: ResizerProps) => {
...
+  grid-area: 1 / 1 / -1 / -1;
+  width: 12px;
+  height: 12px;
+  z-index: 1;
+}
+[data-iui-resizer='top'],
+[data-iui-resizer='bottom'] {
+  height: 8px;
+  width: auto;
+  z-index: 0;
```
rohan-kadkol, line 294: Why do the edge resizers get a z-index of 0, while the corners get a z-index of 1?

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -0,0 +1,64 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+
+const isBrowser = typeof document !== 'undefined';
```
rohan-kadkol, line 8: nit: Maybe this can be reused in other places. E.g. `useIsomorphicLayout`

packages/itwinui-react/src/core/utils/components/Resizer.tsx
```diff
@@ -216,99 +234,80 @@ export const Resizer = (props: ResizerProps) => {
...
+[data-iui-resizer] {
+  pointer-events: auto;
+  grid-area: 1 / 1 / -1 / -1;
+  width: 12px;
+  height: 12px;
+  z-index: 1;
+}
+[data-iui-resizer='top'],
+[data-iui-resizer='bottom'] {
+  height: 8px;
```
rohan-kadkol, line 292: Since the edges and corners are now protruding out, would it be better to increase the size of the resizers so that they bleed into the dialog a little bit too? E.g. the resizer can be `x` pixels out and `x` pixels in the Dialog.

With this increased size of the resizers, would the resizing be easier for users?

packages/itwinui-react/src/core/utils/components/Resizer.tsx
mayank99: I used plain CSS to get access to selectors. Inline `style` is ok here, because the `cursor` is different for every resizer.

packages/itwinui-react/src/core/utils/components/Resizer.tsx
mayank99: Yes, see PR description

image

packages/itwinui-react/src/core/utils/components/Resizer.tsx
mayank99: It's a tradeoff. We don't want it to protrude too much, because it could cover up other adjacent elements.

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: I doubt it. `useIsClient` already exists and is better for 99% of the cases.

Utilities/helpers should generally be colocated where they are used, until they are needed in multiple places.

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -3452,46 +3452,9 @@ it('should sync body horizontal scroll with header scroll', () => {
...
-        width: columnWidths[2],
-      },
-    ],
-  });
-
-  // body serves as the shadow host
-  const host = container.querySelector('.iui-table-body') as HTMLDivElement;
-
-  const slot = host?.shadowRoot?.querySelector('slot');
-  expect(slot).toBeTruthy();
```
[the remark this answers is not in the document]
mayank99: Done https://github.com/iTwin/iTwinUI/pull/1836/commits/5bbfdcddb0202bace5e6441af724950ca9f9d4c6

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
rohan-kadkol: I mean, useIsomorphicLayout uses `typeof window !== 'undefined'`, which I think is similar to `typeof document !== 'undefined'`.

https://github.com/iTwin/iTwinUI/blob/ba8714532461d29ba42829321f4098a1fc4fc41a/packages/itwinui-react/src/core/utils/hooks/useIsomorphicLayoutEffect.tsx#L15

packages/itwinui-react/src/core/utils/components/Resizer.tsx
rohan-kadkol: non-blocking: I mean, earlier, the resizers were placed inside the dialog. But now they're only placed outside the dialog. So, I was thinking would it be better to also keep a reasonable amount inside the dialog. That reasonable amount could be how much was kept inside before this PR.

But how it is is okay too. Just sharing my thoughts. We can go with what you have currently too.

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
rohan-kadkol: To reiterate the `nit` part, this is non-blocking. How it is is also alright 👍. This is me just leaving nit-pick type comments, but how it is is also alright.

packages/itwinui-react/src/core/utils/components/Resizer.tsx
mayank99: Oh I misunderstood. Yes, we could make it go deeper inside the dialog, but I deliberately chose the current positioning, specifically because it would otherwise cover up the scrollbar (which is already super thin) and potentially the close button too.

image

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: Going to leave it as-is for now. It's a simple boolean and requires the same (or more) number of characters to import it vs redeclare it.

packages/itwinui-react/src/core/utils/components/Resizer.tsx
rohan-kadkol: Ohh, I see, the current approach is safer too. So, keeping the current approach sounds good 👍.

rohan-kadkol (approved):

PR #1843, opened 2024-02-12
Configure Developpement container for Codespaces

AdamMeza-Bentley:
## Changes

[GitHub CodeSpaces](https://docs.github.com/en/codespaces/overview) are cloud-based development environments which offer numerous benefits including preconfiguration of dependencies, access to recourses such as computing power and memory, and aiding paired programming. 

The motivation for this PR was to enhancing our development experience by allowing visual testing, adding VScode extensions, and run all relevant commands automatically during the set up process.

### `devcontainer.json` 

Dev containers are Docker containers that are configured to provide a fully featured development environment. A `devcontainer.json` is used by VS code, and other IDEs, as a way of setting up the environment to the right specifications. Further information about Dev containers can be [found here](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers). 

The base image is the [Linux Universal image](https://github.com/devcontainers/images/tree/main/src/universal), which is standard for GitHub Codespaces. It supports a wide range of languages and frameworks, including all the ones we use in iTwinUI. It also supports a variety of containers: macOS, Window, and Linux. Given that the standard image suited all of our needs I felt no reason to reach for anything else. 

This file also adds the recommended VS Code extensions via [identifiers](https://code.visualstudio.com/docs/editor/extension-marketplace#_install-an-extension): Astro, ESlint, Prettier, Stylelint, and Code Spell Checker. All of these extensions are recommended by the iTwinUI team. Lastly it installs Yarn and runs the build which sets up the user to begin developing. 

**NOTE ABOUT PR**

In order to reduce build time we've elected to not use `Dockerfile` for this build. While this does cut down the time from about 10 minutes, the process is still a bit slow, averaging about 5 minutes. The absence of a Dockerfile also causes issues with testing dependencies for `a11y` testing. 
  
## Testing

While there are no changes to tests, the efficacy of this PR can be tested by building a codespace on this branch and running `yarn dev`, `yarn test` or any more specific test to see that the development environment is properly set up. 

## Docs

A changset has been added summarizing this PR


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+    xsel \
+    libgtk2.0-0 \
+    libgtk-3-0 \
+    libgbm-dev \
+    libnotify-dev \
+    libnss3 \
+    libxss1 \
+    libasound2 \
+    libxtst6 \
+    xauth
```
mayank99, line 17: I'm not sure that this reasoning is correct (or maybe I'm misunderstanding):

> This Dockerfile also uses the [Advanced Package Tool](https://en.wikipedia.org/wiki/APT_(software)) program apt-get to add the prerequisites Cypress requires to run in a Linux Eviorment, [which are outlined here](https://docs.cypress.io/guides/getting-started/installing-cypress#Linux-Prerequisites). We do this at this level as several dependencies/prerequisites, including xsel and xvfb, can't be handled by yarn.

Cypress already uses its own docker image, so it shouldn't be necessary here.

Do we even need a custom Dockerfile? It makes the use of Codespaces very impractical because of how long it takes to build. For comparison, if you open a Codespace in the `main` branch, it only takes a few seconds.

I think you might be able to replace the Dockerfile with a [predefined template](https://docs.github.com/en/codespaces/developing-in-a-codespace/creating-a-codespace-from-a-template), which starts up much faster.

AdamMeza-Bentley: That's a good point! I'll look into that and see if that's possible

AdamMeza-Bentley: 1. Creating a template from an existing repository [requires access to the settings](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository), which as a non-admin of iTwinUI I don't have access to. Similarly I looked into [prebuilds](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds) also for time reasons and ran into the same issue.

mayank99: I meant using an existing template, rather than creating a new one. Before that though, can you try simply using the linux universal image as-is (instead of custom dockerfile)?

```diff
- "build": {
-   "dockerfile": "Dockerfile"
- },
+ "image": "mcr.microsoft.com/devcontainers/universal:2",
```

AdamMeza-Bentley: When using the universal image, the dependency issues that lead me to use a Dockerfile come up: 
image

I can look into existing templates

mayank99: That error is coming from `a11y` tests, where we do not use a docker container. If you try running the react visual tests for example, it should work fine.

To fix this issue, you can change the `a11y` test configuration to also run in the `cypress:included` container (same as `react-workshop` test configuration).

AdamMeza-Bentley: looking at the `cypress.config` for both the `a11y` tests and the `react-workshop` tests, I don't recognize anything relating to containers. Would these configurations be found somehere else?

mayank99: See here: https://github.com/iTwin/iTwinUI/blob/main/apps/react-workshop/scripts/run-tests.js#L8

AdamMeza-Bentley: sweet thank you! I had found `run-tests.js` and have been working through understanding the customizations actually. I appreciate the tips

AdamMeza-Bentley: I've come to this possible (but not working solution) and I was hoping to get some feedback on whether or not I'm on the right path or not. 

So in the `package.json` of `testing/a11y` I've changed the `test` property to be: `"docker run --rm --entrypoint /bin/bash -v %cd%:/testing/a11y -w /e2e 'cypress/included:13.2.0' run --component "`

This is based on several docs I've read, and the script being run in `run-tests.js`. While there's still some errors I'm working through I just wanted to see if this approach in generally the on the righ track

mayank99: > So in the `package.json` of `testing/a11y` I've changed the `test` property to be: `"docker run --rm --entrypoint /bin/bash -v %cd%:/testing/a11y -w /e2e 'cypress/included:13.2.0' run --component "`

Yeah, that's generally the right approach

mayank99: Resolving this thread based on internal discussion.

> I found some success using this command:
>   ```
>     "test": "docker run --rm --entrypoint /bin/bash -v $PWD:/iui -w /iui  cypress/included:13.2.0 -c \"yarn workspace a11y test:run\"",
>     "test:run": "cypress run --component",
>   ```
> 
> However, there is another issue caused by some optional platform-specific dependencies missing when the host system (responsible for installing deps) is on a different platform. This is a [bug in npm](https://redirect.github.com/npm/cli/issues/4828) itself.
> 
> For the purpose of Codespaces though, we can proceed without a11y testing working. It can be fixed in a future PR.

.changeset/slimy-moles-rescue.md
mayank99, line 1: Changeset not neeed. This is not a user-facing change in our packages.

AdamMeza-Bentley: removed it

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+      "extensions": [
+        "stylelint.vscode-stylelint",
+        "esbenp.prettier-vscode",
+        "dbaeumer.vscode-eslint",
+        "streetsidesoftware.code-spell-checker",
+        "astro-build.astro-vscode"
+      ]
+    }
+  },
+  "postCreateCommand": "yarn install && yarn build"
```
mayank99, line 14: We don't need `yarn build` on creation, and it takes too long. `yarn install` should be enough.

```suggestion
  "postCreateCommand": "yarn install"
```

mayank99: Actually I was testing it out and found that `postCreateCommand` runs too late, after everything in the Codespace is initialized. [`onCreateCommand`](https://containers.dev/implementors/features/#lifecycle-hooks) should work better.

```
"onCreateCommand": "yarn install"
```

AdamMeza-Bentley: made that change

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
```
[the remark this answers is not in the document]
AdamMeza-Bentley: changed that, much faster now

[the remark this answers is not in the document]
mayank99: Are you not getting this error when creating a new codespace? Because I am.
image

I wonder if we need to specify all the things:

```
"hostRequirements": {
  "cpus": 4,
  "memory": "16gb",
  "storage": "16gb"
},
```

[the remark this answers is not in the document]
AdamMeza-Bentley: I'm not getting that error. I can try that change you suggested tho

.devcontainer/devcontainer.json
mayank99: Okay, this is even worse, because it is now waiting for `yarn install` to complete before showing anything at all.

image

The only one left worth trying is "updateContentCommand".

```
"updateContentCommand": "yarn install"
```

AdamMeza-Bentley: tried that and its a lot faster

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mayank99: Still getting the same error. I guess lets remove this. We can always manually increase the specs on-demand using "New with options"
image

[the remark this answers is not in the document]
AdamMeza-Bentley: alright yeah i'll remove them

.devcontainer/devcontainer.json
mayank99: I'm not seeing any change. Are you running a new codespace? Simply rebuilding is not going to be enough to test it.
image


One more thing we can try is changing the ["waitFor"](https://containers.dev/implementors/json_reference/#lifecycle-scripts) so that updateContentCommand runs in parallel with the UI initialization.

```
"updateContentCommand": "yarn install".
"waitFor": "onCreateCommand"
```

AdamMeza-Bentley: Adding the updateContentCommand, and the WaitFor made it a lot faster. still have to wait for yarn install but gets to the UI sooner

mayank99: `waitFor` still isn't doing the right thing for me. The UI starts faster, but after that it waits 10-30 seconds before starting `yarn install`, which then takes 2-4 minutes to complete. I'd say lets remove `waitFor`.

AdamMeza-Bentley: removed it

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+      "extensions": [
+        "stylelint.vscode-stylelint",
+        "esbenp.prettier-vscode",
+        "dbaeumer.vscode-eslint",
+        "streetsidesoftware.code-spell-checker",
+        "astro-build.astro-vscode"
+      ]
+    }
+  },
+  "updateContentCommand": "yarn install"
```
mayank99, line 14: More improvements: can we add [port attributes](https://containers.dev/implementors/json_reference/#port-attributes)?

I'm thinking it would be useful to have labels for all forwarded ports. e.g. 5173 is "css-workshop", 6006 is "react-workshop", etc

Also, running `yarn dev` opens the portal (port 3000) in a whole new window, which is (1) jarring, (2) useless because clicking on any of them does not open the correct thing. I believe `onAutoForward` can be set to `silent` to prevent it opening a new tab?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "customizations": {
```
mayank99, line 3: More improvements: by default, it opens the README, but doesn't do so immediately. So if you start a Codespace and are looking at some file, it will abruptly open README.md.

I believe you can set [`openFiles`](https://containers.dev/supporting#github-codespaces) like this:
```suggestion
  "customizations": {
		"codespaces": {
		  "openFiles": []
		},
```

Empty array would be ideal, but if it doesn't work then you can open `playgrounds/vite/src/App.tsx`.

.devcontainer/devcontainer.json
AdamMeza-Bentley: so after comparing them all, "postCreateCommand" seems to be the best option. they all take longer than we'd like but `"onCreateCommand"` delays the UI, and `"updateCreateCommand"` has a long delay between the UI showing up and `yarn install` runnning

.devcontainer/devcontainer.json
AdamMeza-Bentley: added those

.devcontainer/devcontainer.json
mayank99: Can you update your branch and change it to `pnpm install` and try again?

AdamMeza-Bentley: `install` , `dev` and other basic worflow commands worked

mayank99: I tried it just now and it's still not a good experience. It takes too long to start `pnpm install`, as seen in the recording below


https://github.com/iTwin/iTwinUI/assets/9084735/64abb0c8-0657-4f6d-b7fc-e6646d5f78e4


Recommend changing it back.
```diff
- "postCreateCommand": "pnpm install",
- "waitFor": "onCreateCommand"
+ "onCreateCommand": "pnpm install"  
```

AdamMeza-Bentley: made that change

mayank99: Much better! See full flow in recording below. Everything is ready in just over 1 minute.

https://github.com/iTwin/iTwinUI/assets/9084735/60cfec49-1807-4a97-8e06-6a006c4f6649

.devcontainer/devcontainer.json
mayank99: 3000 is still opening in a new tab. Can you try `"ignore"` instead of `"silent"`?

mayank99 (commented):
One more improvement: can we mention codespaces in `CONTRIBUTING.md`? We can even use their badge and include a clickable link.

Add this heading under "How to setup"

> ### Using GitHub Codespaces (cloud IDE)
> 
> To get started without having to install anything locally, you can create a codespace for this repository by clicking this link:
> 
> [Open in GitHub Codespaces](https://codespaces.new/iTwin/iTwinUI)

<details>
<summary>Raw markdown</summary>

```md
### Using GitHub Codespaces (cloud IDE)

To get started without having to install anything locally, you can create a codespace for this repository by clicking this link:

[![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://codespaces.new/iTwin/iTwinUI)
```

</details>

.devcontainer/devcontainer.json
AdamMeza-Bentley: yeah ill try that

CONTRIBUTING.md
```diff
@@ -48,6 +48,14 @@ _Before running this command, make sure Docker is running. See [Visual testing](
 
 ---
 
+### Using GitHub Codespaces (cloud IDE)
+
+To get started without having to install anything locally, you can create a [codespace](https://docs.github.com/en/codespaces/overview) for this repository by clicking this link:
+
+[![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://codespaces.new/iTwin/iTwinUI)
+
+---
```
mayank99, line 57: This is in the wrong place. It should be directly under "## How to setup" (~L17)

Other suggested changes:
- Add a heading "### Local setup" directly below it.
- Remove the "VSCode users" heading, turn it into static bold text.
- Rename "## Run locally" to "### Commands", and change all sub-headings inside it to be h4 (####)

image

.devcontainer/devcontainer.json
mayank99: It's still not working as expected using `"silent"`, but let's look into it at a different point in the future.

CONTRIBUTING.md
AdamMeza-Bentley: changes made

mayank99 (approved):
LGTM after https://github.com/iTwin/iTwinUI/pull/1843#discussion_r1501009302 is addressed

CONTRIBUTING.md
AdamMeza-Bentley: sorry made those changes before the comment edit. I'll make those now

PR #1845, opened 2024-02-12
remove focus-visible mixin

mayank99:
## Changes

The focus-visible mixin was added in pre-Safari 15 days, where we didn't have widespread support for `:focus-visible`. Two years later, we can remove it. Default focus styles will still apply in older browsers, which imo is good enough.

## Testing

N/A

## Docs

N/A


.changeset/chilled-hornets-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+'@itwin/itwinui-css': patch
+'@itwin/itwinui-react': patch
+---
+
+Removed `:focus-visible` fallback styles for older browsers.
```
rohan-kadkol, line 6: Is this change only an internal change since it relates to mixins? And so, is a changeset not needed?

rohan-kadkol (approved):

.changeset/chilled-hornets-refuse.md
mayank99: By replacing the mixin with `:focus-visible`, we are also removing the fallback that was being included in the mixin, which affects users in older browsers.

PR #1846, opened 2024-02-13
create `LineClamp` utility component

mayank99:
## Changes

Internal `LineClamp` component, created entirely in React and not part of barrel file.

Used in #1863.

(Unrelated: Removed redundant `text-overflow: ellipsis` from the line-clamp mixin used in some CSS classes.)

## Testing

Tested manually. Not sure if it needs unit tests, since it's just a box with some styles. Will be tested more thoroughly in #1863.

## Docs

N/A


packages/itwinui-react/src/core/utils/components/LineClamp.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <ShadowRoot css={css}>
+        <slot />
+      </ShadowRoot>
+      {children}
+    </Box>
+  );
+}) as PolymorphicForwardRefComponent<'div', { lines?: number }>;
+
+const css = /* css */ `
+  :host {
```
rohan-kadkol, line 30: Was `text-overflow: ellipsis;` from the `iui-line-clamp` mixin intentionally left out?

rohan-kadkol: I read the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) for `-webkit-line-clamp` and learned that the ellipses are automatically added. In that case, can we remove `text-overflow: ellipsis;` from the `iui-line-clamp` mixin?

rohan-kadkol (approved):

PR #1854, opened 2024-02-13
Add CSS files for doc site examples

Ben-Pusey-Bentley:
## Changes

This PR moves the inline styles in the live examples for the doc site into CSS files. This change is being done to coincide with the move to Code Pen. The CSS from the new files is add to the CSS Code Pen file. Note that inline styles where the value for the property is based on a JSX value have been left as is.

## Testing

Navigated the doc site with `yarn dev` and ensured that the live example components have the same styling. Opened the Code Pen sandboxes and ensured that the CSS appears in the CSS file.

## Docs

Adds external CSS files for the live examples styling. As a result, the example code has had it's inline styling removed.


examples/Alert.sticky.jsx
```diff
@@ -4,10 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import { Alert, Text } from '@itwin/itwinui-react';
+import './Alert.sticky.css';
```
mayank99, line 7: We should not be importing CSS files into each component like this. This CSS is considered "global" (unscoped) and will cause unnecessary conflicts with other global CSS while forcing us to use long, unwieldy class names.

examples/InformationPanel.horizontal.css
```diff
@@ -0,0 +1,3 @@
+.horizontal-information-panel-table {
+  min-width: 450px !important;
```
mayank99, line 2: You shouldn't need to use `!important` in any of these demos.

examples/SearchBox.basic.jsx
```diff
@@ -4,10 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import { SearchBox, Flex } from '@itwin/itwinui-react';
+import './SearchBox.basic.css';
 
 export default () => {
   return (
-    <Flex style={{ width: '70%' }}>
+    <Flex className='basic-search-box-flex'>
```
mayank99, line 11: A large portion of this task is to also remove the extra containers like `Flex` and handle it all in CSS, so that the JSX code is less verbose to read.

mayank99: The same comment also applies to any wrapping `<div>`s. Ideally, those should not be in the JSX too.

Thinking a bit more about this, we might need some kind of convention for "wrapper" styles because it is so common to have those in our demos. This wrapper could even include a default max-width.

apps/website/src/components/OpenInCodePen.astro
```diff
@@ -13,6 +13,8 @@ const css = outdent`
     padding: 1rem;
   }
 
+  ${exampleStyle}
```
mayank99, line 16: I'm thinking it would be nice if the entirety of the "CSS" pane only contained the demo CSS. 🤔

mayank99 (commented):
This is a good start, but it needs some tweaks to improve both the maintenance and readability. I've left some concerns below to start discussions, but no concrete suggestions yet.

examples/InformationPanel.horizontal.css
Ben-Pusey-Bentley: That `!important` is being used because the `minWidth` style for the table component is being overwritten in the component itself.
image
As a result, it can only be overwritten with `!important`. Could remove this override, but I am not quite sure if it is safe to do so.

apps/website/src/components/OpenInCodePen.astro
mayank99: We can remove the box-sizing style, and move the rest into inline styles. The body styles can be moved into HTML and the app-wrapper styles can be moved into JSX.

image

examples/InformationPanel.horizontal.css
mayank99: You can set the `min-width` on a wrapping element instead of on the Table itself. 

Additionally, since this example is meant to demonstrate InfoPanel, you can move the Table (and the new wrapping element) into a separate component at the bottom of the file. This will improve readability.

examples/Alert.sticky.jsx
Ben-Pusey-Bentley: Okay, I've approved the PR set the base to the shadow-demos branch, and merged the changes from that branch in.

examples/InformationPanel.horizontal.css
Ben-Pusey-Bentley: Alright, sounds good.

apps/website/src/components/OpenInCodePen.astro
Ben-Pusey-Bentley: Will do.

examples/Alert.sticky.jsx
Ben-Pusey-Bentley: Alright, will do.

Ben-Pusey-Bentley: Most recent commit has simplified the class names and added the `demo-` prefix to them.

examples/InformationPanel.horizontal.css
Ben-Pusey-Bentley: Most recent commit has removed `!important` and moved the Table into a separate component.

examples/SearchBox.basic.jsx
Ben-Pusey-Bentley: I noticed while working on this that popover elements don't correctly take styles from the CSS files. This seems to be because they get added to the root DOM and not the shadow DOM. Might be best to just add inline styles to those specific elements, since there aren't that many popover based elements in our examples overall that use non standard styling. Wanted to get your opinion on this though, do you think it would worth the effort to try and add these elements back into the shadow DOM?

mayank99: > do you think it would worth the effort to try and add these elements back into the shadow DOM?

It's at least worth looking into. We explicitly set `<body>` as the portal target in a couple places:

https://github.com/iTwin/iTwinUI/blob/97d9a75dd63449c8122669307209f2732c14925e/examples/index.tsx#L10

https://github.com/iTwin/iTwinUI/blob/97d9a75dd63449c8122669307209f2732c14925e/examples/ColorPicker.advancedPopover.jsx#L66

This was done in order to avoid stacking context issues (see #1480), but it might not be necessary for all popovers.

Ben-Pusey-Bentley: Okay, I'll look into it then.

Ben-Pusey-Bentley: `Flex`s have been removed and replaced, moving on to getting the popover elements to appear in the shadow DOM.

mayank99: If it takes too long, you can leave the inline styles on popovers. We can always look into it in the future.

Ben-Pusey-Bentley: Tested a few different ideas with the portal property for a little bit, and couldn't make the popover portal off of an element in the shadow DOM. Decided to just add the inline styles to the elements for now.

apps/website/src/components/OpenInCodePen.astro
Ben-Pusey-Bentley: Most recent commit resolves this.

examples/SearchBox.basic.jsx
mayank99: Then could you bring back `Flex` in `Popover.focus.jsx`? When inline styles wrap into multiple lines, it really hurts readability

apps/website/src/components/OpenInCodePen.astro
mayank99: [`outdent`](https://www.npmjs.com/package/outdent) is not necessary if you don't have any indented code. You can use simple strings

```js
const css = code.css.trim();
const html = `<body style='margin: 0'></body>`
```

examples/SearchBox.basic.jsx
Ben-Pusey-Bentley: Most recent commit brings the `Flex` back.

apps/website/src/components/OpenInCodePen.astro
Ben-Pusey-Bentley: `outdent` has been removed from these css and html values.

AdamMeza-Bentley (approved):

mayank99 (approved):
LGTM! We can look into adding a [convention for demo containers](https://github.com/iTwin/iTwinUI/pull/1854#discussion_r1488625777) and fixing [portal issues](https://github.com/iTwin/iTwinUI/pull/1854#discussion_r1491579233) at some point in the future.

PR #1856, opened 2024-02-13
improve handling of external `Anchor` links

mayank99:
## Changes

- Added new `isExternal` prop to `Anchor`. When true, displays an icon at the end of link text.
  - This is simply using the `iui-anchor-external` class that was already present in the CSS.
- `Anchor` links that open in a new tab will now add a visually-hidden warning for screen-reader users.
  - This is decoupled from the `isExternal` prop, and only dependent on the `target` prop.
  - This can still be combined with the `isExternal` prop, which adds a visual indication for sighted users.

Further reading: [1](https://www.nngroup.com/articles/new-browser-windows-and-tabs/), [2](https://adrianroselli.com/2020/02/link-targets-and-3-2-5.html), [3](https://theadminbar.com/accessibility-weekly/opening-links-in-new-tabs-or-not/)

## Testing

Updated unit tests and visual tests.

Tested the presence of visually-hidden text in VoiceOver.

image

## Docs

Added changesets and jsdocs.

Updated some of existing documentation text for `Anchor`, and added a new section for "External links".


apps/website/src/content/docs/anchor.mdx
```diff
@@ -31,6 +42,26 @@ The anchor component supports the polymorphic `as` prop to render the anchor as
 
 The `as` prop can also be used with `Link` components from different routers such as [Next.js](https://nextjs.org/docs/api-reference/next/link) and [React Router](https://reactrouter.com/en/main/components/link)
 
+### External links
+
+When linking to an external site, you can use the `isExternal` prop to indicate that the link will take the user to a different domain. This will add a standard icon to the end of the link text.
+
+<LiveExample src='Anchor.external.jsx' truncate={false}>
+  <AllExamples.AnchorExternalExample client:load />
+</LiveExample>
```
mayank99, line 51: I'm not fully sure about this documentation. I feel like the icon is only useful when links are opening in a new tab. Is there any benefit to showing the icon when not specifying `target=_blank`?

At the same time, I do not want to encourage users to blanket add `isExternal target='_blank'` everywhere. That's why I kept `isExternal` and `target=_blank` separate from each other, but can't find a nice way to document it.

packages/itwinui-react/src/core/Typography/Anchor.tsx
```diff
@@ -18,4 +32,24 @@ import { polymorphic } from '../utils/index.js';
...
+      {...rest}
+      ref={forwardedRef}
+      className={cx(
+        'iui-anchor',
+        { 'iui-anchor-external': isExternal },
+        props.className,
+      )}
+    >
+      {children}
+      {props.target === '_blank' && (
```
rohan-kadkol, line 50: non-blocking: Maybe later we can handle more values of the target prop ([docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/target#usage_notes))

apps/website/src/content/docs/anchor.mdx
```diff
@@ -17,7 +16,19 @@ The anchor's `href` attribute is the destination URL while the text within the a
...
 ## Usage
 
-Consider using a named anchor when you want the user to land not only on a specific page, but also to arrive at a content area that may be below the fold so the user does not have to scroll to discover it.
+The `Anchor` component is a styled wrapper over the native `<a>` element, and it can be used in the same way as the native element.
+
+```jsx
+<Anchor href='https://example.com'>Example</Anchor>
+```
+
+Consider using a named anchor when you want the user to land not only on a specific page, but also to arrive at a content area that may be below the fold so the user does not have to scroll to discover it. Named anchors are created by adding an `id` attribute to an element, and then linking to that element using the `href` attribute.
```
rohan-kadkol, line 25: Is this needed? Not sure if this is iTwinUI specific.

apps/website/src/content/docs/anchor.mdx
rohan-kadkol: > I'm not fully sure about this documentation. I feel like the icon is only useful when links are opening in a new tab. Is there any benefit to showing the icon when not specifying `target=_blank`?

I too can't think of any case where the icon without the `target=_blank` is beneficial.

If there really isn't any benefit, should we consider showing the icon whenever we get `target=_blank`? Or maybe letting the users decide will be better to avoid unexpected visual changes for the user. 

> At the same time, I do not want to encourage users to blanket add `isExternal target='_blank'` everywhere. That's why I kept `isExternal` and `target=_blank` separate from each other, but can't find a nice way to document it.

I feel this documentation adequately separates `isExternal` and `target='_blank'`. So as-is is good imo.

rohan-kadkol (approved):

packages/itwinui-react/src/core/Typography/Anchor.tsx
mayank99: The other values of `target` are rarely used, and also do not require any special handling that I can think of.

apps/website/src/content/docs/anchor.mdx
mayank99: I was expanding on the text that was already there. I've shortened it since it's just regular anchor behavior. [`58d121f` (#1856)](https://github.com/iTwin/iTwinUI/pull/1856/commits/58d121f294999ce3dd0172f35a6bf9192373a680)

apps/react-workshop/src/Anchor.stories.tsx
```diff
@@ -20,3 +20,11 @@ export const AsButton = () => {
     </Anchor>
   );
 };
+
+export const External = () => {
+  return (
+    <Anchor href='https://www.example.com/' isExternal>
```
rohan-kadkol, line 26: nit: Since you also added `target="_blank"` in the docs example, maybe you can add it here also.

rohan-kadkol (approved):

apps/react-workshop/src/Anchor.stories.tsx
mayank99: Done [`ea66af0`](https://github.com/iTwin/iTwinUI/pull/1856/commits/ea66af0239cb1390f896235dbe461d384ac057c6)

PR #1858, opened 2024-02-14
`ToggleSwitch`: show checkmark icon by default

mayank99:
## Changes

React: Added `<SvgCheckmark />` by default if no custom `icon` is passed; this is even shown for small sizes. Also added `null` handling in case someone wants to remove the default icon.

CSS: Removed `display: none` to allow showing the checkmark in small toggle switches.

https://github.com/iTwin/iTwinUI/assets/9084735/81bf21e1-a327-4a2a-80aa-9e30a75fca0d

Motivation: This improves the baseline accessibility of our toggles, preventing potential confusion around "on" vs "off" state ([read more](https://axesslab.com/toggles-suck/)). It also reduces the amount of work application developers have to do; there isn't a good reason to make this feature "opt-in" (vs opt-out).

Note: Ideally there would be a text indication of "on" vs "off" (as seen in the [APG example](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/)), but it's outside our control and also unrealistic in dense UIs, so this is the next best solution.

## Testing

Updated visual tests in both HTML and React to cover the various permutations.

image  image

Changed the icon used in the custom icon story, to verify that it indeed overrides the default:
image

Also added a unit test for `null` handling. The existing unit tests are already sufficient for covering all the other scenarios.

## Docs

Added changesets, updated jsdocs.

Removed the "With icon" section from docs. The main use case for `icon` is to add a checkmark; we shouldn't encourage adding a custom icon.

## TODOs for future

- [ ] Look for potentially valid uses of custom `icon` (not checkmark), and deprecate `icon` prop if none found.


apps/css-workshop/pages/toggle-switch.html
```diff
@@ -196,6 +225,10 @@ <h2>Small</h2>
               role="switch"
               disabled
             />
+            <svg-checkmark
+              class="iui-toggle-switch-icon"
+              aria-hidden="true"
+            ></svg-checkmark>
```
rohan-kadkol, line 231: Was this checkmark intentionally not added to all toggle switches on the page? E.g. "Toggle switch on left".

rohan-kadkol (approved):

examples/index.tsx
```diff
@@ -1217,10 +1217,6 @@ const ToggleSwitchLabelsExample = withThemeProvider(
 );
 export { ToggleSwitchLabelsExample };
 
-import { default as ToggleSwitchIconExampleRaw } from './ToggleSwitch.icon';
-const ToggleSwitchIconExample = withThemeProvider(ToggleSwitchIconExampleRaw);
```
Ben-Pusey-Bentley, line 1221: (nit): I think that it might be better to replace this example with one that uses a custom icon instead of completely removing it. I know that we are going to look into deprecating that functionality, but we still have it available currently.

Ben-Pusey-Bentley (approved):

apps/react-workshop/cypress-visual-screenshots/baseline/ToggleSwitch.test.ts-Disabled Checked.png
AdamMeza-Bentley, line 1: Would it be possible to make the check mark slightly thicker or darker here? It's a bit low contrast and hard to tell what the checkmark is initially

apps/react-workshop/cypress-visual-screenshots/baseline/ToggleSwitch.test.ts-Label Right.png
AdamMeza-Bentley, line 1: although here the thickness/contrast is good so maybe it's not worth changing it on the grey.

examples/index.tsx
mayank99: The reasoning is mentioned in the PR description:

> Removed the "With icon" section from docs. The main use case for `icon` is to add a checkmark; **we shouldn't encourage adding a custom icon**.

apps/react-workshop/cypress-visual-screenshots/baseline/ToggleSwitch.test.ts-Disabled Checked.png
mayank99: This PR is not changing the actual visuals of the icon, it's only showing the icon _by default_. Disabled elements are always tricky to code contrast for; the best practice is to avoid using disabled elements altogether.

examples/index.tsx
Ben-Pusey-Bentley: Adding a custom icon can be discouraged without completely removing the example. There could be a highlighted text box warning users about the issues that could be caused by using a custom icon. Not a big deal either way.

PR #1859, opened 2024-02-15
add visually hidden "Loading" text to progress indicators

mayank99:
## Changes

`ProgressRadial` and `ProgressLinear` are empty divs that do not hold semantics. This is fine for sighted users, since a visual loading indicator is a pretty well-understood pattern. However, non-sighted users do not get any indication at all.

This PR tries to remedy that by using a visually-hidden "Loading" message. Screen-readers will read this out just like any other regular text.

This visually hidden text is located inside shadow DOM. To make this work I had to change the `VisuallyHidden` component to include styles (global styles do not affect elements inside shadow DOM). This in turn required some changes in the `ShadowRoot` component.

**Note:** A much better solution would be to use the [`progressbar` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role), however it [requires an accessible name](https://dequeuniversity.com/rules/axe/4.7/aria-progressbar-name), which the application developers might not set. Also, our progress indicators are supposed to allow passing any arbitrary DOM nodes as `children` (for example, buttons), which might cause problems with the `progressbar` role.

## Testing

Added unit tests to check the presence of "Loading" message. And of course, tested manually to ensure that shadow DOM is wired up correctly, and contains the visually-hidden message with all styles.

And verified that the text is indeed announced by screen-readers.

It's near-impossible to write unit tests for shadow DOM CSS, but since the visual tests are passing, the logic is indirectly being tested. (If the CSS was not working, the visual tests would fail).

## Docs

Added changesets for user-facing parts.

Expanded the progressindicator documentation to explain the new functionality, along with an additional recommendation on what to do when loading completes.

## TODOs for future

- [ ] Allow customizing (or disabling) the loading message.
- [ ] Look into incorporating the `progressbar` role, especially for determinate progress indicators.

### DIFF
--- a/.changeset/sour-bears-applaud.md
+++ b/.changeset/sour-bears-applaud.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`ProgressRadial` and `ProgressLinear` will now include a visually hidden "Loading" text alternative for non-sighted users.
--- a/apps/website/src/content/docs/progressindicator.mdx
+++ b/apps/website/src/content/docs/progressindicator.mdx
@@ -26,6 +26,12 @@ thumbnail: ProgressIndicator
   <AllExamples.ProgressRadialMainExample client:load />
 </LiveExample>
 
+## Accessibility
+
+To make the progress indicator accessible to non-sighted users, a visually-hidden "Loading" message will be automatically included when the `value` is not set to 100%.
+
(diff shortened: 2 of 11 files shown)


AdamMeza-Bentley (approved):
This is super interesting. I'm gonna look up more about the shadow DOM. definitely a good fix for accessibility purposes

rohan-kadkol (approved):

PR #1862, opened 2024-02-15
`Select`: fix `null` value handling

mayank99:
## Changes

`Select` had a small (but significant) bug where it wouldn't respect `value={null}` and instead enter uncontrolled mode. This is annoying when trying to reset the value.

This PR fixes it by updating the logic to match the [`useControlledState`](https://github.com/iTwin/iTwinUI/blob/mayank/select-null/packages/itwinui-react/src/core/utils/hooks/useControlledState.ts) hook.

## Testing

Tested in playground and added a unit test.

```tsx
import { Button, Select } from '@itwin/itwinui-react';

  const [value, setValue] = React.useState<number | null>(2);

  return (
    <>
      <Select
        value={value}
        onChange={setValue}
        options={[
          { value: 1, label: 'Option 1' },
          { value: 2, label: 'Option 2' },
          { value: 3, label: 'Option 3' },
        ]}
      />
      <Button onClick={() => setValue(null)}>Reset</Button>
    </>
  );
}
```

</details> 

## Docs

Added changeset.

While this is a bug fix and doesn't really need docs, I've still updated the jsdocs for the `value` prop.

## TODOs for future

- [ ] Investigate [JSDocs not showing up in IDE](https://github.com/iTwin/iTwinUI/pull/1862#discussion_r1492672122).

### DIFF
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -111,7 +111,7 @@ export type SelectMultipleTypeProps<T> =
        *
        * Pass `null` to reset the value.
        */
-      value?: T;
+      value?: T | null;
       /**
        * Callback function handling change event on select.
        */
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -70,6 +70,8 @@ export type SelectOption<T> = {
   size?: 'default' | 'large';
   /**
    * Value of the item.
+   *
+   * Pass `null` to reset the value.
    */
   value: T;
   /**
@@ -258,7 +260,7 @@ export const Select = React.forwardRef(
     const [liveRegionSelection, setLiveRegionSelection] = React.useState('');
 
     const [uncontrolledValue, setUncontrolledValue] = React.useState<T | T[]>();
-    const value = valueProp ?? uncontrolledValue;
+    const value = valueProp !== undefined ? valueProp : uncontrolledValue;
 
     const onChangeRef = useLatestRef(onChangeProp);
     const selectRef = React.useRef<HTMLDivElement>(null);
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -70,8 +70,6 @@ export type SelectOption<T> = {
   size?: 'default' | 'large';
   /**
    * Value of the item.
-   *
-   * Pass `null` to reset the value.
    */
   value: T;
   /**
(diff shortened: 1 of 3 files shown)


rohan-kadkol (approved):

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -70,6 +70,8 @@ export type SelectOption<T> = {
   size?: 'default' | 'large';
   /**
    * Value of the item.
+   *
+   * Pass `null` to reset the value.
```
rohan-kadkol, line 74: Should this line be for the `value` in `SelectMultipleTypeProps` instead?

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -70,6 +70,8 @@ export type SelectOption<T> = {
...
    */
   value: T;
```
rohan-kadkol, line 76: non-blocking: Since we're also accepting `null`, are we expecting users to define `T` to also include `null`?

```diff
- <Select<number>
+ <Select<number | null>
  value={value}
  onChange={(value) => setValue(value)}
  placeholder='Select an option'
  options={[
    { value: 1, label: 'Option 1' },
    { value: 2, label: 'Option 2' },
    { value: 3, label: 'Option 3' },
  ]}
/>
```

Because with this type modification, `options.value` can also be `null`.

Or should we update the type of `value` to also include `null` so that users don't need to change their types?

If this comment involves too many changes, this can come in a later PR too.

AdamMeza-Bentley (approved):

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: I've changed it ([`c8d4b16`](https://github.com/iTwin/iTwinUI/pull/1862/commits/c8d4b161f3706659c729225bdac81da06f5680d3)), but honestly it doesn't matter. The ideal use case doesn't even involve users explicitly passing generics to `Select`, it should be inferred automatically.

The one place it does matter is when defining the type for `useState`, where the user needs to include `null` anyway.

```jsx
const [value, setValue] = React.useState<number | null>(2);
```

packages/itwinui-react/src/core/Select/Select.tsx
rohan-kadkol: > though, strangely, i cannot get the jsdocs to show up

Ohh, same for me, maybe something we can look into a future PR. If so, maybe we can add a TODO for this in the PR description.

rohan-kadkol (approved):

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: Done

PR #1863, opened 2024-02-16
make Table cells responsive on narrow widths

### LINKED ISSUE #908: When decrease table header width the height of header looks wired
When we try to decrease the width of the table header, when it reaches the state where the width is less than the width of the text, the height of the header column starts increasing and it affects the height of the table header.

#### Expected behavior

There should be minimum width set for each header based on the width of the header text

#### Reproduction
Link to a minimal repro: https://itwin.github.io/iTwinUI-react/?path=/story/core-table--full-2

##### Steps to reproduce
1. Open this example https://itwin.github.io/iTwinUI-react/?path=/story/core-table--full-2
1. decrease the width of any column e.g. price, quantity 

#### Additional information
chrome_k368CHOF0c

mayank99:
## Changes

Several changes to improve responsive design and UX of Tables.
- All columns have a non-zero min-width defined to prevent them from getting too small.
  - This will be overridden by user-specified min-width.
- Cells will clamp automatically if the value is a string (can be explicitly enabled/disabled)
  - This is implemented using shadow DOM and can be controlled by the user using `<DefaultCell clamp>`.
- Filter/sort icons will wrap into next line on narrow widths.
- Filter icon is now always visible (previously only on hover).
- Sort icon occupies less space.

Fixes #908 (or at least tries).

## Testing

https://github.com/iTwin/iTwinUI/assets/9084735/9ac52557-4f83-49bb-a19f-290dbe7d5401

Removed failing unit tests (where everything is mocked) and added e2e tests for real interactions (column resizing and sorting).

## Docs

Added changesets.

Not sure about docs, since there is no API change.


apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Full2.png
mayank99, line 1: @rohan-kadkol @FlyersPh9 While I'm fighting unit tests, do you want to look at this PR and let me know what you think of the changes in general?

The deploy preview is here: https://itwin.github.io/iTwinUI/1863/react/?story=table--full2

mayank99: >     * For the default `minWidth`, one idea was to base it in part on the number of characters in the header string. But the columns may look asymmetrical in width if the columns have different length texts.

The intent here is to just increase the default from 0, to avoid the one-character-per-line situation. The user should be responsible for specifying a more sensible default `minWidth` based on their situation.
 
>     * When there is no filter icon, having the sort icon on the right seems a little off to me:

I actually think this looks better than left aligning. Obviously lot of subjectiveness here.

We can also rework the sort UI in the near future.

>     * Would it be good to have a prop to customize the number of lines that the header is clamped to?
>       
>       * For example, if the user was building the [Table → Sticky Columns](https://itwin.github.io/iTwinUI/1863/react/?story=table--sticky-columns&theme=dark) story, they might want it to wrap on just one line:

Maybe. I'm hesitant to add it because our users are notoriously bad at wrapping things. If given the option, they might truncate everything to one line, which is pretty bad imo.

>     * Would it be beneficial to have middle truncation?

This can be achieved in user-land by passing `MiddleTextTruncation` into a custom cell.

mayank99 (commented):
I think this PR is now ready for proper review. The tests are finally good to go!

I just need to figure out this weird issue with undefined/NaN widths, and write some docs.

Ben-Pusey-Bentley (approved):
Looks good to me.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -957,6 +957,11 @@ export const Table = <
                           (c) => c.id !== SELECTION_CELL_ID, // first non-selection column is the expander column
                         );
 
+                    // expander column should be wider to accommodate the expander icon
+                    if (columnHasExpanders && column.minWidth === 72) {
+                      column.minWidth = 108;
+                    }
```
rohan-kadkol, line 963: Would it be better to have something like an object with the possible widths instead of directly using the hardcoded numbers in code? E.g.

```jsx
const defaultColumnWidths = { 
  default: 72,
  widthExpander: 108
};

…

// expander column should be wider to accommodate the expander icon
if (columnHasExpanders && column.minWidth === defaultColumnWidths.default) {
  column.minWidth = defaultColumnWidths.withExpander;
}
```

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1005,12 +1012,26 @@ export const Table = <
...
+                            <LineClamp>
+                              <slot />
+                            </LineClamp>
+                          ) : (
+                            <slot />
+                          )}
+                          <slot name='actions' />
+                          <slot name='resizers' />
+                          <slot name='shadows' />
+                        </ShadowRoot>
```
rohan-kadkol, line 1026: Is there an advantage to using a `ShadowRoot` here?

testing/e2e/app/routes/Table/spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    await resizer.hover({
+      position: { x: resizerBox.width / 2, y: resizerBox.height / 2 },
+    });
+    await page.mouse.down();
+    await page.mouse.move(
+      Math.max(1, resizerBox.x + delta + resizerBox.width / 2),
+      resizerBox.y + resizerBox.height / 2,
+      { steps: step ? Math.abs(delta) : 5 },
+    );
+    await page.mouse.up();
```
rohan-kadkol, line 177: Should we also test for touch events, similar to the old "shoulf handle resize with touch" test?

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -3287,81 +2600,25 @@ it('should resize last and closest column on the left when table width would dec
...
-  );
-  expect(headerCells).toHaveLength(3);
-
-  // Headers widths sum are not lower than table width
-  expect(headerCells[0].style.width).toBe('100px');
-  expect(headerCells[1].style.width).toBe('150px');
-  expect(headerCells[2].style.width).toBe('50px');
-});
-
-it('should not show resizer when column has disabled resizing when resize mode is expand', () => {
```
rohan-kadkol, line 3324: If we remove the `getBoundingClientRect` mock, can this test be brought back?

packages/itwinui-react/src/core/Table/Table.test.tsx
```diff
@@ -4761,18 +4005,6 @@ it('should render row with loading status', () => {
   expect(rows[1]).not.toHaveClass(`iui-loading`);
 });
 
-it('should navigate through table sorting with the keyboard', async () => {
```
rohan-kadkol, line 4764: There don't seem to be any mocks here. So, curious why this was removed. Is it because sorting is tested in `e2e`?

rohan-kadkol (commented):
> I just need to figure out this weird issue with undefined/NaN widths

Could you elaborate on how to repro this? I'm not sure on which element is this happening and under what conditions.

packages/itwinui-react/src/core/Table/Table.test.tsx
mayank99: It was failing for some reason, so I moved it to e2e. 🤷

packages/itwinui-react/src/core/Table/Table.test.tsx
mayank99: I think it's already covered once? The columnResizeMode has no effect on it.

image

mayank99:
> > I just need to figure out this weird issue with undefined/NaN widths
> 
> Could you elaborate on how to repro this? I'm not sure on which element is this happening and under what conditions.

It can be seen on the column header cells if you pass `minWidth: undefined` to a column.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: Yes, it hides the LineClamp wrapper div and thereby avoids changing the DOM shape (which the user might be relying on, especially for table-body cells)

testing/e2e/app/routes/Table/spec.ts
mayank99: I'll have to see if it's possible to do a meaningful test for touch events in playwright. TBD

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: Good suggestion, done: [`f44a13b`](https://github.com/iTwin/iTwinUI/pull/1863/commits/f44a13b0f2ccb3e668ea18699231622b7c32f1ec)

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -421,8 +422,7 @@ export const Table = <
 
   const defaultColumn = React.useMemo(
     () => ({
-      maxWidth: 0,
-      minWidth: 0,
+      minWidth: 72,
```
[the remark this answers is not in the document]
mayank99: Done with https://github.com/iTwin/iTwinUI/pull/1863#discussion_r1526576473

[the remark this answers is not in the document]
rohan-kadkol: I was curious about how 72 was the decided min-width. Adding a comment for that might help us in the future when reading this code.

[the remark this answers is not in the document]
mayank99: I honestly don't know why I chose 72. I think I was iterating and thought 72 was something that's wide enough but not too wide 😅 There is no special logic behind it. 

This is also why i asked for feedback about visuals in https://github.com/iTwin/iTwinUI/pull/1863#discussion_r1515242978

But moving it into a constant object avoids the possibility of bugs at least.

rohan-kadkol (commented):
When testing, I noticed a few potential issues:

* The table header height might jump in certain cases. But this may just be unavoidable.
  
https://github.com/iTwin/iTwinUI/assets/45748283/72b3ff07-dc71-4b34-adc0-b041cf62c5c7

* Clicking on the table cell's content does not select the row (Left = this PR's Full2 story; Right = hosted version's [Full2 story](https://itwin.github.io/iTwinUI/react/?mode=preview&story=table--full2&theme=dark))

https://github.com/iTwin/iTwinUI/assets/45748283/6f49f767-ae42-4823-b208-d5b12161fc89

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Full2.png
mayank99: > Unrelated, but I noticed the cells with a status applied are showing up in the scrollbar gutter

This appears to be a Safari-only thing. When they add support for standard `scrollbar-*` properties, hopefully it should work fine.

mayank99 (commented):
[`758b1c7`](https://github.com/iTwin/iTwinUI/pull/1863/commits/758b1c7a586a569709108e040df10b9dc05064b6) fixes the `min-width: undefinedpx`. I've also added a fix for the very large `max-width` issue. There is still an issue with `width: NaNpx` seen sometimes, but I couldn't figure out why, and didn't notice any real issues caused by it (we would have set it to `undefined` anyway), so left it as-is.

image

---

> The table header height might jump in certain cases. But this may just be unavoidable

Yes, this is unavoidable. I guess we could make the filter button smaller to occupy less space in some situations.

---

> Clicking on the table cell's content does not select the row 

This is probably caused by shadow DOM, but I actually think this behavior is better than before. I always found it annoying that selecting text also selects the row.

In the future maybe we could improve it to only prevent selection when the user is actively selecting text. But not important imo.

testing/e2e/app/routes/Table/spec.ts
mayank99: Yeah I couldn't write anything meaningful. I tried manually constructing touch events (which already kinda defeats the purpose of testing), but even that didn't work.

```tsx
test('should work with touch events', async ({ page }) => {
  await page.goto('/Table');

  const initialWidths = await getColumnWidths(page);

  await page.evaluate(() => {
    const resizer = document.querySelector('[role="separator"]')!;
    const resizerBox = resizer.getBoundingClientRect();

    resizer.dispatchEvent(
      new TouchEvent('touchstart', {
        touches: [
          new Touch({
            identifier: Date.now(),
            target: resizer,
            clientX: resizerBox.x,
            clientY: resizerBox.y,
          } as any),
        ],
      }),
    );
  });

  await page.evaluate(() => {
    const resizer = document.querySelector('[role="separator"]')!;
    const resizerBox = resizer.getBoundingClientRect();
    resizer.dispatchEvent(
      new TouchEvent('touchmove', {
        touches: [
          new Touch({
            identifier: Date.now(),
            target: resizer,
            clientX: resizerBox.x + 100,
            clientY: resizerBox.y ,
          } as any),
        ],
      }),
    );
  });

  await page.evaluate(() => {
    const resizer = document.querySelector('[role="separator"]')!;
    resizer.dispatchEvent(new TouchEvent('touchend'));
  });

  const newWidths = await getColumnWidths(page);
  expect(newWidths[0]).toBe(initialWidths[0] + 100);
});
```

</p>
</details>

fwiw, I think it's fine to leave it untested. The logic for mouse vs touch isn't all that different, plus I verified that it works on my phone, and I haven't touched (no pun intended) the code in this PR.

.changeset/fair-lamps-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": minor
+---
+
+The filter button inside a Table will now alaways be visible, instead of only being shown on hover/focus.
```
rohan-kadkol, line 5: Typo

```suggestion
The filter button inside a Table will now always be visible, instead of only being shown on hover/focus.
```

.changeset/six-pans-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+The filter button inside a Table will now alaways be visible, instead of only being shown on hover/focus.
```
rohan-kadkol, line 5: Typo

```suggestion
The filter button inside a Table will now always be visible, instead of only being shown on hover/focus.
```

rohan-kadkol (commented):
I noticed a potential issue. Commenting out the `width: 200` from the product column of the `Full2` story gives the following:

image

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -943,6 +950,19 @@ export const Table = <
...
+                      column.minWidth = COLUMN_MIN_WIDTHS.default;
+                    }
+
+                    // expander column should be wider to accommodate the expander icon
+                    if (
+                      columnHasExpanders &&
+                      column.minWidth === COLUMN_MIN_WIDTHS.default
+                    ) {
+                      column.minWidth = COLUMN_MIN_WIDTHS.withExpander;
+                    }
```
rohan-kadkol, line 964: Can/Should these two if statements be combined?

rohan-kadkol (commented):
Passing a very large `Header` string in the Filters story causes things to break?

<details>
<summary>Exact string</summary>

`IDs (enter one of the IDs in the filter) foobar foobar foobar foobar foobar foobar foobar foobar foobar foobar 123 foobarfoobarfoobarfoobarfoobar 1234567890 aihuwdohauwjd diajwhbvdghijaowhbjdvhbnalw djaiwbhdajiwd`

</details>

https://github.com/iTwin/iTwinUI/assets/45748283/63e64b83-2ff5-42ed-9495-c27f919ce29c

mayank99 (commented):
> Passing a very large `Header` string in the Filters story causes things to break?

Good catch. Fixed in [`f26eb85`](https://github.com/iTwin/iTwinUI/pull/1863/commits/f26eb85d9eebd71d9eef72beaa5c93bacae51ae2) by reverting to the same values as `main` branch. We only care about changing `minWidth` in this PR.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: Why/how? They are doing completely different things based on different conditions and should be run sequentially

rohan-kadkol: Why: Testing `column.minWidth === COLUMN_MIN_WIDTHS.default` in the second if statement felt unnatural to me because if the `column.minWidth` is set to `72` by the user, I guess we should not override their value.

How: I was thinking of something like this:

```tsx
if (column.minWidth === undefined) {
  if (columnHasExpanders) {
    column.minWidth = COLUMN_MIN_WIDTHS.withExpander;
  } else {
    column.minWidth = COLUMN_MIN_WIDTHS.default;
  }
}
```

mayank99: That's not the same thing. I am explicitly increasing the value of `minWidth` from 72 (set at the top of this component) to 108 for the expander column. But I see what you're saying. Change made in [`f4a99d1`](https://github.com/iTwin/iTwinUI/pull/1863/commits/f4a99d1ec880fe899d7a9e9722c23973dd790a23)

rohan-kadkol: > 72 (set at the top of this component)

Not sure I understand where you are setting it to `72`/`COLUMN_MIN_WIDTHS.default`.

But yes, the updated code is cleaner.

mayank99: image

rohan-kadkol (approved):
LGTM! 🚀

I was unable to repro the `width: NaNpx` issue that you mentioned. But as discussed through DMs, since there error is not too important, I left trying to repro it.

packages/itwinui-react/src/core/Table/Table.tsx
rohan-kadkol: So, if you initialized it at the top, was the first if statement (`if (column.minWidth === undefined) {`) not needed?

mayank99: The user could pass `undefined` explicitly, in which case we also want to override it.

PR #1864, opened 2024-02-16
allow anchors to be underlined by default

mayank99:
## Changes

This PR adds a new `underline` prop to `Anchor` (and a corresponding `data-iui-underline` attribute to CSS) to make anchors underlined in their resting state. Defaults to `false` to avoid a breaking change.

Motivation: Until now, we've only been underlining anchors on hover in regular themes and by default in high-contrast themes. This is done in the name of "clean" design, but as a result we end up relying only on color to distinguish links, leading to potential accessibility issues. Underlines are the widely understood convention for hyperlinks in the web, and we shouldn't deviate from it. ([Further reading](https://adrianroselli.com/2016/06/on-link-underlines.html#recommendation))

In a future major release, we should flip the default and advise applications to include a toggle if users actually want "clean design" ([similar to GitHub's toggle](https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/))

## Testing

Added unit and visual tests.

## Docs

Added changesets, jsdocs and updated documentation page.


apps/website/src/content/docs/anchor.mdx
```diff
@@ -57,6 +57,14 @@ If you do ultimately decide based on your research that it is beneficial to open
 
 **Note**: Not all external links should open in a new tab. The `isExternal` prop is not necessary for external links that open in the same tab.
 
+### Underlines
+
+By default, the anchor is only underlined when hovered, except when using a high-contrast theme. To explicitly make the anchor underlined in all themes, use the `underline` prop.
```
rohan-kadkol, line 62: non-blocking: Maybe this part can also mention the accessibility benefits of having underlines on anchors.

rohan-kadkol (approved):

apps/website/src/content/docs/anchor.mdx
mayank99: [`bd36389`](https://github.com/iTwin/iTwinUI/pull/1864/commits/bd3638915f2fd2e610d8f0c53e6f371a0d379392) added link for now.

PR #1865, opened 2024-02-20
Remove `Dialog`/`Modal` wrapper from DOM when `isOpen={false}`

rohan-kadkol:
## Changes

A use-case reported by one user was that they didn't want to render anything related to a Dialog/Model when `isOpen=false`. As of now, we used to always render the wrapper regardless of `isOpen`. But this PR now removes the wrapper is also removed when `isOpen=false` (after a 600ms timeout for the exit anim).

The backdrop used to rely on CSS transitions. So, mounting the backdrop with `iui-backdrop-visible` would not trigger the CSS transition. Thus, this PR also adds an `iui-fade-in` CSS animation to `iui-backdrop`. Since this `iui-fade-in` is used in multiple places, this PR moves it to `global.scss`.

## Testing

* Tested Dialog and Modal in a playground
  * <details>
    <summary>App.tsx (vite-playground)</summary>

	```tsx
	import {
	  Dialog,
	  Button,
	  Modal,
	  ModalButtonBar,
	  ModalContent,
	
	export default () => {
	  const [isOpen1, setIsOpen1] = React.useState(false);
	  const [isOpen2, setIsOpen2] = React.useState(false);
	
	  return (
	    <>
	      <Button styleType='high-visibility' onClick={() => setIsOpen1(true)}>
	        Open dialog
	      </Button>
	      <Dialog
	        isOpen={isOpen1}
	        onClose={() => setIsOpen1(false)}
	        closeOnEsc
	        closeOnExternalClick
	        preventDocumentScroll
	        trapFocus
	        setFocus
	        isDismissible
	        portal
	        renderWrapperWhenClosed={false}
	      >
	        <Dialog.Main>
	          <Dialog.TitleBar titleText='Dialog' />
	          <Dialog.Content>
	            A dialog informs users about a task and can contain critical
	            information, require decisions, or involve multiple tasks. Dialogs
	            appear in front of app content to provide critical information or
	            ask for a decision.
	          </Dialog.Content>
	          <Dialog.ButtonBar>
	            <Button
	              styleType='high-visibility'
	              onClick={() => setIsOpen1(false)}
	            >
	              Primary
	            </Button>
	            <Button onClick={() => setIsOpen1(false)}>Secondary</Button>
	          </Dialog.ButtonBar>
	        </Dialog.Main>
	      </Dialog>
	
	      <Button styleType='high-visibility' onClick={() => setIsOpen2(true)}>
	        Open modal
	      </Button>
	      <Modal
	        isOpen={isOpen2}
	        title='My modal'
	        onClose={() => setIsOpen2(false)}
	        renderWrapperWhenClosed={true}
	      >
	        <ModalContent>Here is my modal content</ModalContent>
	        <ModalButtonBar>
	          <Button styleType='high-visibility'>Primary button</Button>
	          <Button>Secondary button</Button>
	        </ModalButtonBar>
	      </Modal>
	    </>
	  );
	};
	```

    </details>
* Confirmed that all Dialog/Modal react demos still worked
* Added a unit test

## Docs

- Updated JSDocs for `isOpen` to recommend users to not render Dialog conditionally.
- Added changesets.


packages/itwinui-react/src/core/Dialog/Dialog.tsx
```diff
@@ -37,10 +46,34 @@ const DialogComponent = React.forwardRef((props, ref) => {
     placement,
     className,
     portal = false,
+    renderWrapperWhenClosed = true,
     ...rest
   } = props;
 
+  // When user can/should see the dialog wrapper's children
+  // (e.g. when opacity!=0 and not necessarily when isOpen=false (since dialog is visible when animating out))
+  const [isDialogWrapperChildrenVisible, setIsDialogWrapperChildrenVisible] =
```
rohan-kadkol, line 55: Open to suggestions for variable names or even different approaches to achieving the same functionality.

packages/itwinui-react/src/core/Dialog/Dialog.tsx
```diff
@@ -13,20 +13,29 @@ import { DialogButtonBar } from './DialogButtonBar.js';
...
    */
   children: React.ReactNode;
+  /**
+   * By default, the dialog/modal wrapper is always rendered regardless of `isOpen`.
+   *
+   * Pass `false` to avoid rendering the dialog/modal wrapper when `isOpen=false`.
+   *
+   * @default true
+   */
+  renderWrapperWhenClosed?: boolean;
```
mayank99, line 30: Do we need this prop? If there is no downside to always removing it from the DOM (e.g. all animations, etc work the same), then we don't need to make it configurable.

rohan-kadkol: Yes, all animations and the general behavior is unchanged. I thought of having a prop for this since this removal from the DOM behavior may be a breaking change.

But if we deem it not to be a breaking change, I can remove this prop and always remove the Dialog/Modal from the DOM when `isOpen=false`.

mayank99: Yeah, lets remove it then. It's not a breaking change, more of a bug fix. It never should have been kept in the DOM in the first place (it was only for animation reasons).

rohan-kadkol: Sounds good, removed (https://github.com/iTwin/iTwinUI/pull/1865/commits/b23fe980bd2dc29808d1276361fcde062a76740d).

packages/itwinui-react/src/core/Dialog/Dialog.tsx
rohan-kadkol: I had played around with it and `mountOnEnter` before but couldn't get it working. Upon retrying it now, I was able to get it working after a few more modifications.

So now the logic is slightly simplified (https://github.com/iTwin/iTwinUI/pull/1865/commits/3ccbc9bb341ce4957eb339b307ab7e255519c4c8).

mayank99: So you weren't able to remove `isDialogWrapperChildrenVisible` and the internal `isOpen` state?

Triggering `useEffect` based on props is generally unnecessary and [explicitly discouraged by React](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state).

rohan-kadkol: > Triggering `useEffect` based on props is generally unnecessary and [explicitly discouraged by React](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state).

I believe their recommendation is to avoid a stale pass with an incorrect value. But in our case, we intentionally want the first pass to happen with `isOpen=false` and the next passes with `isOpen=true`. This is for proper CSS transitions. So, I don't think this React recommendation applies in our case.

I still tried to remove it (https://github.com/iTwin/iTwinUI/pull/1865/commits/00d7a38ccc8105f6c05bdce666133a53329886b9). But I needed `useIsFirstRender()` from `ShadowRoot.tsx`. I also needed to add `queueMicrotask` to make it work in Dialog.

I'm not sure if `queueMicrotask` would regress something in ShadowRoot. Also, not sure if `useIsFirstRender` is an apt name when using `queueMicrotask` that may not run immediately after the first render and before the second.

> So you weren't able to remove `isDialogWrapperChildrenVisible` and the internal `isOpen` state?

I wasn't able to remove `isDialogWrapperChildrenVisible` since I need a way to set `isOpen` in the context to `true` only after the wrapper has been rendered/mounted.

If we really want to avoid a new state variable, I can look into creating a new component for the wrapper. I can then use `isFirstRender` in that new component to set `isOpen=false` when `isFirstRender` and `isOpen=true` when `!isFirstRender`. This is just an idea, not sure if it's feasible or even better than the current approach.

mayank99: If context is the only problem, then it could be moved down underneath `Transition`.

```jsx
<Portal …>
  <Transition …>
    <Box …>
      <DialogContext.Provider …>
        {children}
      </DialogContext.Provider>
    </Box>
  </Transition>
</Portal>
```

A new component (within the same file) also makes sense if this doesn't work.

Basically, updating state inside `useEffect` should be the last resort and should have a very clear and strong reason behind it (which I'm not seeing in this case). The whole `isFirstRender` thing is a hack for server-side rendering declarative shadow dom, so it shouldn't be used here.

AdamMeza-Bentley (approved):

packages/itwinui-react/src/core/Dialog/Dialog.tsx
rohan-kadkol: I am not sure of how to achieve the same logic without a useEffect + state variable and also without `useIsFirstRender`.

I created an inner helper function that has a `useEffect` + stateVariable (https://github.com/iTwin/iTwinUI/pull/1865/commits/d80f2f13e0e3308426d5a8ca77e8a6de3528b83b). Although it has a useEffect that updates a state variable, at least this time the dependency array is empty. So, it runs only once. I think updating states in such `useEffect`s that only run once are alright.

mayank99: During internal discussion/pairing session, we were able to get rid of the `useEffect` by using a CSS animation. 👍

mayank99 (approved):
LGTM, after some more cleanup (undoing `App.tsx`, adding/updating changesets, changing PR title/description)

.changeset/stupid-toes-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": minor
+---
+
+Moved `iui-fade-in` animation from just being scoped in Dialog-related classes to now be global/unscoped.
```
mayank99, line 5: This changeset could be refocused to highlight what matters to the user. (`iui-fade-in` is just an implementation detail)

```suggestion
`iui-backdrop` will now also fade in when `iui-backdrop-visible` is present during mount.
```

rohan-kadkol:
> > LGTM, after some more cleanup (undoing `App.tsx`, adding/updating changesets, changing PR title/description)
> 
> Undid `App.tsx`, cleaned up `Dialog.tsx`, added a changeset for the itwinui-css changes, and updated the PR description.

Additionally, updated the PR title, fixed the regression of `setFocus=true` not focusing the Dialog/Modal on open (https://github.com/iTwin/iTwinUI/pull/1865/commits/1012e90c68d28adc652846cb58b7c42db03e51ca), and fixed the failing unit tests.

mayank99 (approved):

PR #1866, opened 2024-02-20
Adds VS Code snippets to iTwinUI.

Ben-Pusey-Bentley:
## Changes

Adds 4 VS code snippets to iTwin UI. Information on VS Code snippets can be found [here.](https://code.visualstudio.com/docs/editor/userdefinedsnippets) They are as follows:

1.  A separator comment. prefix: `//-`. scope: All.
2.  Import statements for a iTwinUI component. prefix: `impforwardRef`. scope: JSX and TSX
3.  Boilerplate that more closely matches how iTwinUI components are declared, with `React.forwardRef` and types. prefix: `forwardRef`. scope: TSX
4. Import statement for cx. prefix: `impcx`. scope: TSX

These snippets should cut down on the time spent typing out repetitive boilerplate. Additional snippet recommendations are appreciated.

## Testing

Made sure that the snippets appear in the appropriate scopes.

## Docs

N/A


AdamMeza-Bentley (approved):
These are so cool! definitely helpful to have shorthand ways to add code.

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,63 @@
...
+    "prefix": "importcx",
+    "body": ["import cx from 'classnames';"],
+    "description": "Imports the cx function from classnames"
+  },
+  "CX": {
+    "scope": "typescriptreact",
+    "prefix": "cx",
+    "body": ["className={cx(${1:classNames})}"],
+    "description": "Conditionaly joins class names"
+  }
```
mayank99, line 62: Is it possible to combine these two? Ideally, typing `cx` and pressing <kbd>Tab</kbd> anywhere in the file should automatically add `import cx from 'classnames';` at the top.

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,63 @@
...
+    "prefix": "component",
+    "body": [
+      "function ${1:Name}({ ${2:props} }) {",
+      "	return (",
+      "		${3:jsx}",
+      "	);",
+      "}"
+    ],
+    "description": "React Component boilerplate"
+  },
```
mayank99, line 37: This one is not helpful/useful. We never write components like this.

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,63 @@
...
+  "iTwinUIComponent": {
+    "scope": "typescriptreact",
+    "prefix": "iuicomponent",
```
mayank99, line 40: Not sure about this prefix, maybe something like "forwardRef" would be better.

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,63 @@
...
+    "body": [
+      "const ${1:Name} = React.forwardRef((props, forwardedRef) => {",
```
mayank99, line 42: This is missing imports for `React`, `PolymorphicForwardRefComponent` and `Box`. Is it possible to automate those?

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,63 @@
...
+  // Example:
+  // "Print to console": {
+  // 	"scope": "javascript,typescript",
+  // 	"prefix": "log",
+  // 	"body": [
+  // 		"console.log('$1');",
+  // 		"$2"
+  // 	],
+  // 	"description": "Log output to console"
+  // }
```
mayank99, line 17: All these placeholder comments can be removed

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: I'll look into it.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Do we want an `export default` component instead, since the examples use that a lot, or do we just want to remove this entirely?

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: I'll look into it as well. Should be possible if it is possible to combine the two `cx` snippets.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Will do.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: I found this [stackoverlow question](https://stackoverflow.com/questions/52747900/visual-studio-code-add-corresponding-import-statements-for-snippets) on the topic. Unfortunately, only option 3 really seems to work for this case, since we are aliasing the `classnames` import and `cx` doesn't get called in the global scope. Option 3 seems to be what is already in the PR, just taking advantage of some VS Code shortcuts to make going to top of the file and returning to where you were easier. Option 2 does look like it will work in the `forwardRef` case, since  that gets called in the global scope. Will test that out a little and see how it feels to use.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Changed in latest commit.

.vscode/iTwinUISnippets.code-snippets
mayank99: I think we can remove it, it's not a lot to type anyway

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Decided to add a new `importForwardRef` snippet instead for handling all of these imports, as the second option on the previously link stack overflow page didn't work cleanly. There were two major issues, one being that the optimize imports command does not work on aliased imports like the `React` import, and the other being that imports were not actually moved to the top of the file by the command. I believe this issue was the result of the type declarations that often appear between our import block and where we declare our components.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Will do.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Removed in latest commit.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Removed in latest commit.

.vscode/iTwinUISnippets.code-snippets
mayank99: Lets remove the `cx` code snippet then. It does nothing useful. The import is the annoying part, so we can keep it. Also suggest changing the prefix to `icx`.

.vscode/iTwinUISnippets.code-snippets
mayank99: Similar to https://github.com/iTwin/iTwinUI/pull/1866#discussion_r1499641221, I would prefer a shorter prefix for the import. `iforwardRef` would be too weird though. Maybe `impforwardRef`?

You can play around with different names and see which one feels best.

Ben-Pusey-Bentley: Alright, will do.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: Okay, will do.

Ben-Pusey-Bentley: Done in latest commit.

.vscode/iTwinUISnippets.code-snippets
Ben-Pusey-Bentley: I do like `impforwardRef` more overall, so I have changed the prefix to that. My only real concern overall is that we have two different import snippets, one that starts with `imp` and the other starts with just `i`. Would it be a good idea to rename the cx import prefix to `impcx`, just to create a consistent start to the import prefixes? Or should we just leave it as is for now?

mayank99: Sure, `impcx` is good with me

Ben-Pusey-Bentley: Alright, changed in latest commit.

.vscode/iTwinUISnippets.code-snippets
```diff
@@ -0,0 +1,39 @@
...
+    "scope": "typescriptreact",
+    "prefix": "forwardRef",
+    "body": [
+      "const ${1:Name} = React.forwardRef((props, forwardedRef) => {",
+      " const {${2:props}} = props",
+      "",
+      " return (",
+      "   ${3:jsx}",
+      " );",
+      "}) as PolymorphicForwardRefComponent<'${4:element}', ${5:PropsType}>;"
```
mayank99, line 30: One last comment: This snippet has too many variables. Can we reduce it to just 1:Name? We can use sensible defaults for the rest (or just omit them):

```jxx
const <CursorGoesHere> = React.forwardRef((props, forwardedRef) => {
 const { ...rest } = props;

 return (
   <Box ref={forwardedRef} {...rest} />
 );
}) as PolymorphicForwardRefComponent<'div'>;
```

Ben-Pusey-Bentley: Okay, will do.

Ben-Pusey-Bentley: Latest commit removes all of the variables but name.

mayank99 (approved):
LGTM

PR #1872, opened 2024-02-22
migrate from `yarn` to `pnpm`

mayank99:
## Changes

Changes the package manager to [Pnpm](https://pnpm.io/) v8. Pnpm is faster, more reliable, causes fewer issues, and has lots of useful built-in features.

This is our third attempt at moving away from Yarn 1 (unmaintained for many years now). This time it was successful (probably because of reducing node_modules weight in #1669, #1826, #1833).

I only ran into one [Typescript build issue](https://redirect.github.com/microsoft/TypeScript/issues/42873) (previously seen in https://github.com/iTwin/iTwinUI/pull/1600#issuecomment-1738642465). I was able to work around this using [`node-linker=hoisted`](https://pnpm.io/npmrc#node-linker) and [`shamefully-hoist`](https://pnpm.io/npmrc#shamefully-hoist).

Since pnpm is stricter about dependency management, I had to declare some missing deps where they were previously being used implicitly. I also had to add the [`workspace:*`](https://pnpm.io/workspaces#workspace-protocol-workspace) protocol when using `configs` and `examples` workspaces, because Pnpm otherwise resolves them to npm packages of the same name. This also required unignoring all workspaces from changesets versioning ([see thread](https://github.com/iTwin/iTwinUI/pull/1872#discussion_r1499991480)).

For the most part, `pnpm` is a drop-in replacement for `yarn` in CLI commands, except for [Pnpm's `--filter` syntax](https://pnpm.io/filtering) which can conflict with [Turbo's `--filter` syntax](https://turbo.build/repo/docs/core-concepts/monorepos/filtering). To work around this, `pnpm run` can be used explicitly to disambiguate. (In the future we might be able to get rid of Turbo entirely, if Pnpm can be used for creating dependency chains).

```diff
- pnpm test --filter=react-workshop
+ pnpm run test --filter=react-workshop
```

Lastly, I had to adjust the github actions workflows to explicitly bring in `pnpm/actions-setup` (I will make a similar change in Azure Pipelines separately). I also had to get rid of some pnpm commands running inside Docker where pnpm is not available by default.

**Unrelated changes**:
- Added missing `include` in `vitest.config`, to prevent it from running tests on compiled JS files.
- Updated versions for some outdated github actions.

## Testing

Tested that all common workflows (dev, build, test, lint) work locally, and that CI passes too. Further testing from reviewers (especially on different platforms) is always appreciated.

- [x] macOS
- [x] Linux (CI)
- [x] Windows

## Docs

N/A. Non user-facing change. Updated contributing guide though.


Ben-Pusey-Bentley (approved):
Can confirm that dev, build, test, and lint all work on windows. LGTM.

rohan-kadkol (approved):
`build`, `dev`, `clean`, and `test` worked for me on my Mac laptop.

.changeset/config.json
```diff
@@ -6,6 +6,5 @@
   "linked": [],
   "access": "public",
   "baseBranch": "main",
-  "updateInternalDependencies": "patch",
-  "ignore": ["!@itwin/*"]
```
mayank99, line 10: Had to remove this because changesets was complaining.

```sh
$ pnpm changeset version

> itwinui-monorepo@0.0.0 changeset /Users/mayank/Desktop/Code/iTwinUI
> npx -y @changesets/cli "version"

🦋  error The package "@itwin/itwinui-css" depends on the ignored package "configs", but "@itwin/itwinui-css" is not being ignored. Please pass "@itwin/itwinui-css" to the `--ignore` flag.
🦋  error The package "@itwin/itwinui-react" depends on the ignored package "configs", but "@itwin/itwinui-react" is not being ignored. Please pass "@itwin/itwinui-react" to the `--ignore` flag.

ELIFECYCLE  Command failed with exit code 1.
```

PR #1873, opened 2024-02-23
Fix jumping horizontal scrollbar in Pill Tabs

rohan-kadkol:
## Changes

WIP

### DIFF
--- a/apps/react-workshop/src/Tabs.stories.tsx
+++ b/apps/react-workshop/src/Tabs.stories.tsx
@@ -287,21 +287,21 @@ export const HorizontalOverflow = () => {
 };
 
 export const VerticalOverflow = () => {
-  const [active, setActive] = React.useState('Item 11');
+  const [active, setActive] = React.useState('Item 1');
 
   const tabData = [
     { name: 'Item 1', content: 'Tab Content One', disabled: false },
     { name: 'Item 2', content: 'Tab Content Two', disabled: false },
     { name: 'Item 3', content: 'Tab Content Three', disabled: false },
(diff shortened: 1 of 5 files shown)


mayank99 (commented):
> It does not resolve the other issue that I was experiencing with the tabs growing taller, so that seems to be a separate issue.

I believe this is happening because of the `useScrollbarGutter` hook added in #1733. I think we only need that hook for vertical tabs? Disabling it for horizontal tabs might fix this issue.

.changeset/itchy-cycles-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed `Tab` stripe width calculation that sometimes used to caused unnecessary overflow and thus caused a horizontal scrollbar in `TabList`.
```
rohan-kadkol, line 5:
```suggestion
Fixed `Tab` stripe width and position calculation that sometimes used to cause unnecessary overflow and thus cause a horizontal scrollbar in `TabList`.
```

.changeset/itchy-cycles-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed `Tab` stripe width and position calculation that sometimes used to cause unnecessary overflow and thus cause a horizontal scrollbar in `TabList`.
```
rohan-kadkol, line 5:
```suggestion
Fixed `Tab` stripe size and position calculation that sometimes used to cause unnecessary overflow and thus cause a horizontal scrollbar in `TabList`.
```

PR #1874, opened 2024-02-23
Fix unable to commit changes to example files

rohan-kadkol:
## Changes

When I tried to commit my changes in `Tabs.pill.tsx`, I received this error:

image

So, I added `configs` to the root `devDependencies`.

## Testing

Confirmed that committing changes in `Tabs.pill.tsx` was possible after the changes from this PR.

## Docs

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "prettier": "~3.1.0",
     "rimraf": "^5.0.5",
     "serve": "^14.2.1",
-    "turbo": "~1.12.3"
+    "turbo": "~1.12.3",
+    "configs": "workspace:*"
   },
   "resolutions": {
     "normalize-package-data": "^6.0.0",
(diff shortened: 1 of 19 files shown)


AdamMeza-Bentley (approved):

package.json
```diff
@@ -49,7 +49,8 @@
     "prettier": "~3.1.0",
     "rimraf": "^5.0.5",
     "serve": "^14.2.1",
-    "turbo": "~1.12.3"
+    "turbo": "~1.12.3",
+    "configs": "workspace:*"
```
mayank99, line 53: Does this mean you can remove the individual `"configs"` from deps in all workspaces?

rohan-kadkol: Yes, I believe so. `install`, `dev`, and `build` still work.

I also made dummy changes in all workspaces. When `configs` was not declared as a dep in any workspace, I got this error:

image

But after adding `configs` to the devDeps of only the root package.json, I was able to make the commit.

So removed the `configs` devDep from other workspaces (https://github.com/iTwin/iTwinUI/pull/1874/commits/3dd4144af7d9e051fdfaf268249cebb8c9ffdd7b).

mayank99 (approved):

PR #1875, opened 2024-02-23
Adds Popover elements to DatePicker demos

Ben-Pusey-Bentley:
## Changes

Adds `Popover` elements to 3 `DatePicker` demos. These demos are the 2 Date + Time demos and the Usage demo. The usage demo was updated to show `esc` key functionality. Because the example did not previously use the popover element, the picker could not be closed with the `esc` key. This change fixes that. It also fixes an issue where the `DatePicker` page would scroll to the Date + Time section on load. This was being caused by the `TimePicker` component using the `scrollIntoView`(https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) method to make sure that the default values of the picker are displayed when the picker is rendered. This has the unintended effect of  scrolling the page if the picker is rendered on page load. Wrapping the picker in a popover button solves this issue as well.

## Testing

Made sure that the `DatePicker` documentation page does not scroll to the Date + Time section on reload, and made sure that the `esc` key closes the `Popover` wrapped pickers.

## Docs

The only change that outside users should notice is the Date + Time examples now being a button that user has to activate to see the picker.

### DIFF
--- a/examples/DatePicker.menu.jsx
+++ b/examples/DatePicker.menu.jsx
@@ -3,30 +3,33 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { DatePicker, IconButton } from '@itwin/itwinui-react';
+import { DatePicker, IconButton, Popover } from '@itwin/itwinui-react';
 import { SvgCalendar } from '@itwin/itwinui-icons-react';
 
 export default () => {
   const [currentDate, setCurrentDate] = React.useState(new Date());
-  const [opened, setOpened] = React.useState(false);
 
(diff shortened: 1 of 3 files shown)


AdamMeza-Bentley (approved):
looks good! I just used `Popover` in a similar way in `AppUI`. Seems like a really good way to delete duplicate functionality

rohan-kadkol (approved):

mayank99 (commented):
This PR is incomplete.
- All datepicker examples need to be migrated, not just select few.
- Focus should be move into the datepicker itself when opened. Pressing arrow keys immediately after opening a datepicker should cycle focus between different dates.
- Selecting a date should also close the popover.

PR #1879, opened 2024-02-26
Add native `<select>` variant

FlyersPh9:
## Changes

- Add styling for native `<select>` element.
  - Move styling that is not needed for `<select>` inside `:is(:not(select))`.
- Added `native` prop to `<Select>`/`<LabeledSelect>` react components.
  - Defaults to `false`, in which case simply renders the old custom Select.
  - When `true`, renders a new, simplified native `<select>` with `<option>`s derived from the `options` prop. `placeholder` prop is used to add an empty option.
  - Needed some fiddling with Typescript but I was able to retrofit it into old types, while correctly diverging supported/unsupported props based on the value of `native`.

## Testing

- Tested in HTML select demo page in Chrome/Firefox/Safari on macOS and Windows.
  - Tried all 3 sizes.
  - Tried disabled.
  - Tried with status indicator.
  - Tried with placeholder and disabled options.
- Added unit tests for `native` prop.
- Added visual tests (both CSS and React).
- Added stories.

## Docs

Added jsdocs and changesets.

Added proper docs in `select.mdx`.

## TODOs for future

- [ ] Fix [duplicate jsdocs](https://github.com/iTwin/iTwinUI/pull/1879#discussion_r1508170157), make them appear on hover, and simplify types.


apps/css-workshop/pages/select.html
```diff
@@ -333,6 +384,91 @@ <h2>Default</h2>
         </div>
       </label>
 
+      <label
+        class="iui-input-grid"
+        data-iui-status="positive"
+      >
+        <div class="iui-input-label">Select label</div>
+        <div class="iui-input-with-icon">
+          <select class="iui-select-button">
```
rohan-kadkol, line 393: Since you're using "Select native label` in other places, you can use it in the remaining places too.

FlyersPh9: Oops, bad copy / paste, my mistake.  Fixed!

rohan-kadkol (approved):
LGTM. Approving this since I think you intend the docs and the React implementation to come in separate PRs that will likely merge in this branch.

mayank99 (changes requested):
Blocking accidental merge until React portion is done

packages/itwinui-css/src/select/select.scss
mayank99, line 1: Should we include placeholder styling? Can be achieved using this selector:

```css
select:has(option[value=""]:checked)
```

mayank99, line 1: Should we style the `option` list on Windows to include `background-color` and `color`?

packages/itwinui-css/src/select/select.scss
```diff
@@ -17,23 +17,32 @@
...
-  align-items: center;
   user-select: none;
-  position: relative;
-  overflow: hidden;
   color: var(--_iui-input-text-color);
   background-color: var(--_iui-input-background-color);
   border: 1px solid var(--_iui-input-border-color);
   line-height: var(--iui-size-l);
+  font-size: var(--iui-font-size-1);
+  font-family: var(--iui-font-sans);
```
mayank99, line 27: These could be combined as `font: inherit`

packages/itwinui-css/src/select/select.scss
```diff
@@ -17,23 +17,32 @@
...
+  appearance: none;
```
mayank99, line 28: I think `appearance: none` belongs inside `:is(select)`

packages/itwinui-css/src/select/select.scss
```diff
@@ -17,23 +17,32 @@
...
+  cursor: pointer;
   @include iui-select-size;
 
-  &.iui-disabled {
+  &:is(:not(select)) {
```
mayank99, line 32: `:is` is redundant

```suggestion
  &:not(select) {
```

packages/itwinui-css/src/select/select.scss
```diff
@@ -17,23 +17,32 @@
...
+    display: flex;
+    align-items: center;
+    gap: var(--iui-size-xs);
+    position: relative;
+    overflow: hidden;
+  }
+
+  &.iui-disabled,
+  &:disabled {
```
mayank99, line 41: `:is` can be used to combine these

```suggestion
  &:is(.iui-disabled, :disabled) {
```

packages/itwinui-css/src/select/select.scss
FlyersPh9: Ooh, this is new to me!  Added!

FlyersPh9: Ah, I was unaware Windows allowed this styling.  I've added `color` and `font-weight` in case the `<option>` inherits both from `:has(option[value=""]:checked)`.

packages/itwinui-css/src/select/select.scss
FlyersPh9: Done!

packages/itwinui-css/src/select/select.scss
FlyersPh9: Moved into `:is(select)`.

packages/itwinui-css/src/select/select.scss
mayank99: `--iui-font-weight` does not exist. Maybe you meant `--iui-font-weight-normal`.

Also, I don't think the placeholder option should be `disabled` in HTML. It needs to be selectable because it is an actual option that is initially selected. Otherwise it looks like this inside the menu:

image

vs without `disabled`:
image

Also these `option` styles should be moved underneath `&:is(select)`

mayank99: Can you use a [forgiving selector list](https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list#forgiving_selector_list) to avoid breaking `iui-placeholder` in older browsers that don't support `:has`?

```diff
  &.iui-placeholder,
- &:has(option[value='']:checked)
+ &:is(select:has(option[value='']:checked))
```

mayank99: Actually I have mixed feelings on the `disabled` attribute. It looks weird initially (notice it's invisible and then becomes visible when I hover any other option), but it works fine after an option is selected. Whereas without the `disabled` attribute, I can always select it, which is probably worse?

https://github.com/iTwin/iTwinUI/assets/9084735/f33d40ea-7dc2-4f95-affa-4f028a6fcb8b

FlyersPh9: > `--iui-font-weight` does not exist. Maybe you meant `--iui-font-weight-normal`.

Oops, yes, fixed.

mayank99: Seems to work for me?

image

image

Maybe some typo somewhere?

FlyersPh9: > Also these `option` styles should be moved inside `&:is(select) {}`

Can do, but that makes me think `.iui-icon`, `.iui-content`, `.iui-placeholder` should be within `&:not(select) {}` then.

mayank99: > Can do, but that makes me think `.iui-icon`, `.iui-content`, `.iui-placeholder` should be within `&:not(select) {}` then.

No, it's fine. Those are "legacy" styles that aren't necessarily getting in the way of new `<select>` functionality. I'm just trying to make the new code more maintainable.

FlyersPh9: Regarding placeholder option being `disabled`, I prefer disabled if the label is `Placeholder…`, but if the label was `None` then it should be enabled.  But should `None` be a placeholder at that point? 🤔 Probably not I would think.

This is what it looks like on macOS:
Screenshot 2024-02-28 at 11 40 05 AM

mayank99: macOS looks fine, but on Windows my main concern with `disabled` was the confusing disappearance from the list as shown above. If we're ok with that, lets keep the `disabled` attribute. The user will still have the choice to do things differently if they want.

mayank99: We use Sass, so all kinds of nesting is valid. I bet there is a bug in Sass or LightningCSS somewhere.

However, it looks like the build process already separates these rules, so I'm ok with leaving as-is for now. We can investigate the invalid selector issue int he future.

image

FlyersPh9: > > Can do, but that makes me think `.iui-icon`, `.iui-content`, `.iui-placeholder` should be within `&:not(select) {}` then.
> 
> No, it's fine. Those are "legacy" styles that aren't necessarily getting in the way of new `<select>` functionality. I'm just trying to make the new code more maintainable.

This is done.



> macOS looks fine, but on Windows my main concern with `disabled` was the confusing disappearance from the list as shown [above](https://github.com/iTwin/iTwinUI/pull/1879/commits/083e504ee49a958c1082e88dcdd0f10b84f86a99#r1506216611). If we're ok with that, lets keep the `disabled` attribute. The user will still have the choice to do things differently if they want.

I'm ok with it.

mayank99: Should we still specify `color` for `option[disabled]`? Currently it looks the same as other options when first opened.

image

image

FlyersPh9: I added styling for disabled options, you'll have to confirm it looks ok in Windows because macOS uses system styling so I can't verify my change.

mayank99: Disabled styling on **Chrome** Windows looks ok now.
image image

Disabled styling on **Firefox** Windows also looks ok, but the selected state is a bit hard to tell. Not sure we can do anything about it, as it's coming from the browser itself.

image image

mayank99 (commented):
I've implemented the React portion using a simple prop called `native`, with most of the old API intact. Give it a shot!

```diff
  <Select
    placeholder='Pick an option, any option'
    options={[
      { value: '1', label: 'Option 1', disabled: false },
      { value: '2', label: 'Option 2', disabled: true },
      { value: '3', label: 'Option 3' },
    ]}
+   native
  />
```

Still working on tests, etc.

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -182,58 +240,11 @@ export type SelectProps<T> = {
...
- *   )}
- *   selectedItemRenderer={(option) => (
- *     <span style={{ backgroundColor: option.value }}>{option.label}</span>
- *   )}
- * />
- */
-export const Select = React.forwardRef(
+// ----------------------------------------------------------------------------
+
+const CustomSelect = React.forwardRef(
```
rohan-kadkol, line 245: The order of the components and types in the file can be made more consistent to make it more easy to read. E.g. `Component` and then `ComponentProps` (or vice versa), helper components like `SelectButton`/`SelectEndIcon` at the bottom of the file, etc.

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -536,3 +540,197 @@ const MultipleSelectButton = <T,>({
...
+
+// This is a smaller, stricter subset of CustomSelectProps, because native
+// <select> only supports string values, and does not support icons or sublabels.
+type NativeSelectProps = SelectCommonProps & {
+  /**
+   * Selected option value.
+   *
+   * Must be a string, because it is passed as an attribute to the native <select>.
+   */
+  value?: string | null;
```
rohan-kadkol, line 610: It can be mentioned that passing `null` resets the value, similar to the value in `SelectMultipleTypeProps`.

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: Sure, I can reorder the thing towards the end, once most of the review is done. I'll leave it as-is for now to minimize diff.

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -536,3 +540,197 @@ const MultipleSelectButton = <T,>({
...
+        status={status}
+        disabled={disabled}
+        defaultValue={defaultValueProp}
+        value={valueProp === null ? '' : valueProp}
+        {...triggerProps}
+        onKeyDown={mergeEventHandlers(triggerProps?.onKeyDown, (event) => {
+          // Firefox does not open the menu on Enter, so we need to do it manually.
+          if (event.key === 'Enter') {
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
rohan-kadkol, line 575: Are these needed? When I removed them, I didn't get any error.

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
```diff
@@ -49,7 +49,10 @@ export type LabeledSelectProps<T> = {
    */
   svgIcon?: JSX.Element;
   /**
-   * If true, shows a red asterisk but does not prevent form submission.
+   * If true, shows a red asterisk.
+   *
+   * Form submission is only disabled if the `native` select is used.
```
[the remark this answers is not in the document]
mayank99: Do you have a better suggestion? I thought it was phrased in a pretty direct way, with no implications or recommendations.

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: Maybe restart your TS server? I can't get rid of this error.

image

image

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -17,13 +17,66 @@ import {
...
+  }
+>;
+
+// ----------------------------------------------------------------------------
+
+const SelectEndIcon = React.forwardRef((props, forwardedRef) => {
+  const { disabled, isOpen, ...rest } = props;
+  return (
+    <Icon
+      as='span'
```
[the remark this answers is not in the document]
mayank99: removed [`7dec131`](https://github.com/iTwin/iTwinUI/pull/1879/commits/7dec1316ded4c47f5cf14698c529eee6346cb26d)

packages/itwinui-react/src/core/LabeledSelect/LabeledSelect.tsx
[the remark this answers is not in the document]
rohan-kadkol: Maybe being explicit might be better here:

```suggestion
   * Form submission is only disabled if the `native` select (`<LabeledSelect native />`) is used.
```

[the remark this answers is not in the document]
mayank99: done [`4e77dba`](https://github.com/iTwin/iTwinUI/pull/1879/commits/4e77dbac6d270a69498bc9e15c764f97b6693df5)

packages/itwinui-react/src/core/Select/Select.tsx
rohan-kadkol: For me, it says `showPicker` exists on `currentTarget`:

image

mayank99: I tried removing it ([`c80fd95`](https://github.com/iTwin/iTwinUI/pull/1879/commits/c80fd952316eff36c1a66f99b53da3c63c86e8b1)) and [CI started failing](https://github.com/iTwin/iTwinUI/actions/runs/8097940476/job/22130189674?pr=1879). Your IDE might be using a newer version of TS maybe.

FlyersPh9:
Read over the docs, lgtm

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: Sone [`4a2fc3d`](https://github.com/iTwin/iTwinUI/pull/1879/commits/4a2fc3d88d51a1631865c250e3c2da4a07b67b1f). The most important things are now at the top, with types for it immediately below, and so on

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -17,79 +17,508 @@ import {
...
-   * Defaults to 'large' if `sublabel` provided, otherwise 'default'.
+   * Status of select.
    */
-  size?: 'default' | 'large';
+  status?: 'positive' | 'warning' | 'negative';
+};
+
+// ----------------------------------------------------------------------------
+
+const CustomSelect = React.forwardRef((props, forwardedRef) => {
```
rohan-kadkol, line 252: Does `CustomSelect` intentionally no longer have the generic type `<T>`?

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -17,79 +17,508 @@ import {
...
+   *
+   * Must be a string, because it is passed as an attribute to the native <select>.
+   *
+   * Alternatively, pass `null` to reset the value.
    */
-  close: () => void;
+  value?: string | null;
   /**
-   * Indicates whether an item is selected.
+   * Callback invoked when the selected value changes.
```
rohan-kadkol, line 200: When `native` is not passed, looks like both the JSDocs sentences are being shown.

image

image

So maybe the part where `SelectProps` changes based `native` needs to be modified a bit.

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: Yes. It is a private component so doesn't need the extra TS noise. The "public" part of it works the same as before.

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: I think JSDocs are just broken for `CustomSelectProps`. If you hover a prop, it doesn't even appear (same as https://github.com/iTwin/iTwinUI/pull/1862#discussion_r1492672122). 

We should try to simplify some of the types and figure out what is going on (in the future). I don't want to make big changes to the types in this PR.

rohan-kadkol: Sounds good, then maybe we can add this point to the after-PR TODOs.

rohan-kadkol (approved):

PR #1880, opened 2024-02-26
Adds Popovers to all DatePicker examples and adjusts Popover behavior.

Ben-Pusey-Bentley:
## Changes

This PR makes further changes to the `DatePicker` examples based on [this comment](https://github.com/iTwin/iTwinUI/pull/1875#pullrequestreview-1900983849). This PR:

1. Adds `Popover` components to all remaining `DatePicker` examples.
2. Sets the focus to be inside of the calendar itself when the `DatePicker` is focused using the `setFocus` prop.
3. Changes behavior of the `Popover` component so that selecting a date in the calendar causes the `Popover` to be dismissed. ~~This change required adding some other events for dismissing the `Popover`, in order to preserve it's default functionality (i.e. being dismissed by pressing the `esc` key).~~

## Testing

Made sure that all of the examples appear as `Popover` wrapped `DatePicker`s, and made sure that the new functionality worked for all examples, as well as making sure that the `esc` correctly dismissed the `Popover`.

## Docs

All `DatePicker` examples will now appear wrapped in `Popover`s, but the core of the examples should otherwise remain identical.

### DIFF
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -15,15 +15,17 @@ related:
 
 When a user needs to schedule deadlines, timeframes, or for age verification purposes, we provide them with a date picker. It behaves much like a select menu. We currently have three different types of date pickers.
 
-## Variants
-
-### Basic
+## Usage
 
 The basic date picker allows the user to pick a date from the calendar and skip a month at a time.
+
+This example and the others under this heading use the Popover utility component to show a real world use case.
 
 <LiveExample src='DatePicker.basic.jsx' truncate={true}>
   <AllExamples.DatePickerBasicExample client:load />
 </LiveExample>
+
+If you would rather see an example of the date picker being used on its own, look at the <a href="#standalone-usage">standalone usage</a> example.
 
 This allows the user to visually select the date rather than entering the date with a keyboard. This has the advantage of also showing weekdays, which may be important in a given context. The date picker is commonly used within column filters in [Table](table).
 
@@ -73,7 +75,7 @@ To improve user experience, date pickers support year skipping buttons. Clicking
   <AllExamples.DatePickerWithYearExample client:load />
 </LiveExample>
 
-## Usage
+## Standalone usage
 
 A date picker can be featured on your page in multiple ways. One can be added directly to your page so that it is visible as soon as the page loads. Additionally, a date picker can be added to a menu button on your page so that the user can toggle its visibility.
 
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -25,7 +25,7 @@ This component is recommended to be used inside the [`Popover`](popover) compone
   <AllExamples.DatePickerBasicExample client:load />
 </LiveExample>
 
-If you would rather see an example of the date picker being used on its own, look at the <a href="#standalone-usage">standalone usage</a> example.
+If you would rather see an example of the date picker being used on its own, look at the [standalone usage](#standalone-usage).
 
 This allows the user to visually select the date rather than entering the date with a keyboard. This has the advantage of also showing weekdays, which may be important in a given context. The date picker is commonly used within column filters in [Table](table).
 
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -17,9 +17,9 @@ When a user needs to schedule deadlines, timeframes, or for age verification pur
 
 ## Usage
 
-The basic date picker allows the user to pick a date from the calendar and skip a month at a time.
+The `DatePicker` component allows the user to pick a date from the calendar and skip a month at a time.
 
-This example and the others under this heading use the Popover utility component to show a real world use case.
+This component is recommended to be used inside the [`Popover`](popover) component, which takes care for placement, collision handling, focus management, outside clicks, and other interaction details.
 
 <LiveExample src='DatePicker.basic.jsx' truncate={true}>
   <AllExamples.DatePickerBasicExample client:load />
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -27,7 +27,6 @@ This component is recommended to be used inside the [`Popover`](popover) compone
 
 If you would rather see an example of the date picker being used on its own, look at the [standalone usage](#standalone-usage).
 
-This allows the user to visually select the date rather than entering the date with a keyboard. This has the advantage of also showing weekdays, which may be important in a given context. The date picker is commonly used within column filters in [Table](table).
 
 ### Localized
 
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -75,7 +75,7 @@ To improve user experience, date pickers support year skipping buttons. Clicking
 
 ### Standalone usage
 
-A date picker can be featured on your page in multiple ways. One can be added directly to your page so that it is visible as soon as the page loads. Additionally, a date picker can be added to a menu button on your page so that the user can toggle its visibility.
+The `DatePicker` does not always need to be used within its own popover. You can instead display it as part of a larger interface containing interconnected controls.
 
 <LiveExample src='DatePicker.menu.jsx' truncate={true}>
   <AllExamples.DatePickerMenuExample client:load />
--- a/apps/website/src/content/docs/datepicker.mdx
+++ b/apps/website/src/content/docs/datepicker.mdx
@@ -77,6 +77,8 @@ To improve user experience, date pickers support year skipping buttons. Clicking
 
 The `DatePicker` does not always need to be used within its own popover. You can instead display it as part of a larger interface containing interconnected controls.
 
+You can set `applyBackground={false}` to make the `DatePicker` blend into its surroundings. This is useful when the surrounding UI already brings its own background-color and border.
+
 <LiveExample src='DatePicker.menu.jsx' truncate={true}>
   <AllExamples.DatePickerMenuExample client:load />
 </LiveExample>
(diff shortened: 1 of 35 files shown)


examples/DatePicker.basic.jsx
```diff
@@ -3,16 +3,45 @@
...
+          onBlur={(event) => {
+            if (event.relatedTarget === null) {
+              setVisible(false);
+            }
+          }}
+          onKeyDown={(event) => {
+            if (event.key == 'Escape') {
+              setVisible(false);
+            }
+          }}
```
mayank99, line 31: You don't need these extra event handlers, you just need to pass `onVisibleChange={setVisible}` to `Popover` and it should take care of outside clicks and <kbd>Esc</kbd> keypresses

examples/DatePicker.basic.jsx
```diff
@@ -3,16 +3,45 @@
...
   return (
-    <DatePicker
-      date={currentDate}
-      onChange={(date) => {
-        setCurrentDate(date);
-      }}
-    />
+    <Popover
+      content={
+        <DatePicker
```
AdamMeza-Bentley, line 15: is adding the content inline like this preferred over setting it to a variable? given that its being pass through props, I wonder if it'd be more readable to separate it via variable.  genuine syntactic question

AdamMeza-Bentley (approved):

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: Okay, latest commit changes over to that.

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: I am kind of going back and forth on this. On the one hand, it makes sense to me for all of the JSX to exist in the return statement, so that people don't have to dig through variables to find the JSX that they are looking for. On the other hand, these examples are supposed to be about the `DatePicker` components specifically, so isolating them in a variable could help to distinguish them from the `Popover` boilerplate.

examples/DatePicker.basic.jsx
```diff
@@ -3,16 +3,36 @@
...
+          date={currentDate}
+          onChange={(date) => {
+            setCurrentDate(date);
+            setVisible(false);
+          }}
+          setFocus
+        />
+      }
+      placement='bottom'
```
mayank99, line 24: Can we change all these to "bottom-start"? Feels a bit weird when it's center aligned

examples/DatePicker.basic.jsx
```diff
@@ -3,16 +3,36 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { DatePicker } from '@itwin/itwinui-react';
+import { DatePicker, IconButton, Popover } from '@itwin/itwinui-react';
+import { SvgCalendar } from '@itwin/itwinui-icons-react';
 
 export default () => {
   const [currentDate, setCurrentDate] = React.useState(new Date());
```
mayank99, line 10: Should we also display the selected date next to the button?

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: Will do.

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: We could, but then the Usage example will feel redundant, since the only thing currently differentiating it is the date display that it controls.

mayank99: The "Usage" section is done all wrong anyway. I think the point was to establish a difference between `DatePicker` used inside a `Popover` vs being embedded directly in a larger UI. It has nothing to do with selected dates.

Here's what I would suggest to improve the documentation:
1. Rename the "Usage" heading to "Standalone usage"
2. Change the `.usage.jsx` example underneath to show a DatePicker without `Popover` (and without `setFocus`).
3. Rename the top "Variants" heading to "Usage". Remove the "Basic" heading.
4. Add some text explaining the two different kinds of usages.
   - Show the `.basic.jsx` example directly underneath the text explaining `Popover` usage
   - Below this example, briefly explain "Standalone" usage and link to that heading which contains more details.

Ben-Pusey-Bentley: Okay, will display dates for all examples and make the other suggested changes to the "Usage" section.

Ben-Pusey-Bentley: Latest commit has the changes with regards to the "Usage" section of the documentation.

apps/website/src/content/docs/datepicker.mdx
```diff
@@ -15,16 +15,18 @@ related:
...
 When a user needs to schedule deadlines, timeframes, or for age verification purposes, we provide them with a date picker. It behaves much like a select menu. We currently have three different types of date pickers.
 
-## Variants
-
-### Basic
+## Usage
 
 The basic date picker allows the user to pick a date from the calendar and skip a month at a time.
 
+This example and the others under this heading use the Popover utility component to show a real world use case.
```
mayank99, line 22: Avoid terms like "this heading", which are too literal.

```suggestion
The `DatePicker` component allows the user to pick a date from the calendar and skip a month at a time.

This component is recommended to be used inside the [`Popover`](popover) component, which takes care for placement, collision handling, focus management, outside clicks, and other interaction details.
```

apps/website/src/content/docs/datepicker.mdx
```diff
@@ -15,16 +15,18 @@ related:
...
+
 <LiveExample src='DatePicker.basic.jsx' truncate={true}>
   <AllExamples.DatePickerBasicExample client:load />
 </LiveExample>
 
+If you would rather see an example of the date picker being used on its own, look at the <a href="#standalone-usage">standalone usage</a> example.
```
mayank99, line 28: Prefer markdown syntax over HTML

```suggestion
If you would rather see an example of the date picker being used on its own, look at the [standalone usage](#standalone-usage).
```

apps/website/src/content/docs/datepicker.mdx
```diff
@@ -15,16 +15,18 @@ related:
...
+
 This allows the user to visually select the date rather than entering the date with a keyboard. This has the advantage of also showing weekdays, which may be important in a given context. The date picker is commonly used within column filters in [Table](table).
```
mayank99, line 30: This one feels out-of-place with the new text added above. Maybe it's ok to remove, as it's a lot of fluff

```suggestion
```

apps/website/src/content/docs/datepicker.mdx
```diff
@@ -73,7 +75,7 @@ To improve user experience, date pickers support year skipping buttons. Clicking
   <AllExamples.DatePickerWithYearExample client:load />
 </LiveExample>
 
-## Usage
+## Standalone usage
```
mayank99, line 78: Should be a sub-heading (h3) under "Usage" (h2)

```suggestion
### Standalone usage
```

apps/website/src/content/docs/datepicker.mdx
```diff
@@ -73,7 +75,7 @@ To improve user experience, date pickers support year skipping buttons. Clicking
...
 
 A date picker can be featured on your page in multiple ways. One can be added directly to your page so that it is visible as soon as the page loads. Additionally, a date picker can be added to a menu button on your page so that the user can toggle its visibility.
```
mayank99, line 80: This text is out-of-date with the example. It should be focused on explaining the standalone usage, and specifically mention the `applyBackground` prop as well.

```suggestion
The `DatePicker` does not always need to be used within its own popover. You can instead display it as part of a larger interface containing interconnected controls.

You can set `applyBackground={false}` to make the `DatePicker` blend into its surroundings. This is useful when the surrounding UI already brings its own background-color and border.
```

examples/DatePicker.menu.jsx
```diff
@@ -3,32 +3,22 @@
...
-            />
-          </div>
-        }
-        placement='bottom'
-      >
-        <IconButton label='Choose date'>
-          <SvgCalendar />
-        </IconButton>
-      </Popover>
+      <DatePicker
```
mayank99, line 14: This example should set `applyBackground={false}`.

Also I think it should be used within a `<Wrapper>` component (which you can create a the bottom of this file). This Wrapper component could be built using `Surface`, for applying background, border, padding, etc

examples/DatePicker.basic.jsx
```diff
@@ -3,16 +3,39 @@
...
+        <IconButton
+          label='Choose date'
+          onClick={() => {
+            setVisible(!visible);
+          }}
+        >
+          <SvgCalendar />
+        </IconButton>
+      </Popover>
+      <span>{currentDate.toString()}</span>
```
mayank99, line 38: Prefer `toDateString()` over `toString()` to make it less verbose, at least for the examples where we do not enable time selection.

Also, there should probably be a wrapping `<div>`, because now the empty space looks weird.

Current:
image

Expected:
image

apps/website/src/content/docs/datepicker.mdx
Ben-Pusey-Bentley: Added these in the latest push.

examples/DatePicker.menu.jsx
Ben-Pusey-Bentley: Will do.

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: Will do.

apps/website/src/content/docs/datepicker.mdx
mayank99: Missing second paragraph

examples/DatePicker.basic.jsx
Ben-Pusey-Bentley: Done in latest commit.

mayank99: Prefer flex `gap` over margin, and prefer variables over magic numbers.

Ben-Pusey-Bentley: Changed in latest commit.

mayank99 (approved):
LGTM

PR #1881, opened 2024-02-26
Collapse padding between icon and input

rohan-kadkol:
## Changes

After-PR TODO from #1828.

WIP

### DIFF
--- a/apps/css-workshop/pages/input.html
+++ b/apps/css-workshop/pages/input.html
@@ -23,6 +23,7 @@
       import "@itwin/itwinui-icons-elements/chevron-up";
       import "@itwin/itwinui-icons-elements/chevron-down";
       import "@itwin/itwinui-icons-elements/search";
+      import "@itwin/itwinui-icons-elements/caret-down-small";
     </script>
     <title>Input | iTwinUI</title>
     <style>
@@ -59,6 +60,237 @@
     <theme-button></theme-button>
     <h1>Input</h1>
     <hr />
+
+    <!-- <h2>Testing Only</h2> -->
+    <section id="demo-testing">
+
+      <h2>iui-input-flex-container (InputWithDecorations)</h2>
+      <h3>Old React LabeledInput</h3>
+      <div
+        class="iui-input-flex-container"
+        data-iui-status="positive"
+      >
+        <input placeholder="Placeholder…" />
+        <svg-status-success
+          class="iui-svg-icon"
+          data-iui-icon-color="positive"
+          data-iui-padded="true"
+          data-iui-icon-size="m"
+          aria-hidden="true"
+        ></svg-status-success>
+      </div>
+
+      <h3>SearchBox</h3>
+
+      <div
+        class="iui-input-flex-container"
+        data-iui-size="small"
+      >
+        <span
+          class="iui-svg-icon iui-search-icon"
+          data-iui-icon-size="s"
+        >
+          <svg-search
+            class="iui-button-icon"
+            aria-hidden="true"
+          /></span>
+        <input
+          placeholder="Placeholder…"
+          type="text"
+          id="search-2"
+          value="test search input"
+        />
+        <button
+          class="iui-button"
+          data-iui-variant="borderless"
+          data-iui-size="small"
+        >
+          <svg-close-small
+            class="iui-button-icon"
+            aria-hidden="true"
+          />
+        </button>
+      </div>
+
+      <div
+        class="iui-input-flex-container iui-searchbox"
+        id="demo-search-hover"
+      >
+        <span class="iui-svg-icon iui-search-icon">
+          <svg-search
+            class="iui-button-icon"
+            aria-hidden="true"
+          /></span>
+        <input
+          placeholder="Placeholder…"
+          type="text"
+        />
+        <button
+          class="iui-button"
+          data-iui-variant="borderless"
+        >
+          <svg-close-small
+            class="iui-button-icon"
+            aria-hidden="true"
+          />
+        </button>
+      </div>
+      <div class="iui-input-flex-container">
+        <span class="iui-svg-icon iui-search-icon">
+          <svg-search
+            class="iui-button-icon"
+            aria-hidden="true"
+          /></span>
+        <input
+          placeholder="Placeholder…"
+          type="text"
+          id="search-4"
+        />
+        <p
+          class="iui-text-block iui-text-muted"
+          style="padding-right: var(--iui-size-s);"
+        >
+          0/3
+        </p>
+        <hr
+          class="iui-divider"
+          aria-orientation='vertical'
+        />
+
+        <button
+          class="iui-button"
+          data-iui-variant="borderless"
+        >
+          <svg-chevron-down
+            class="iui-button-icon"
+            aria-hidden="true"
+          />
+        </button>
+        <button
+          class="iui-button"
+          data-iui-variant="borderless"
+        >
+          <svg-chevron-up
+            class="iui-button-icon"
+            aria-hidden="true"
+          />
+        </button>
+        <button
+          class="iui-button"
+          data-iui-variant="borderless"
+        >
+          <svg-close-small
+            class="iui-button-icon"
+            aria-hidden="true"
+          />
+        </button>
+      </div>
+
+      <h2>iui-input-with-icon (InputWithIcon)</h2>
+      <h3>New React LabeledInput</h3>
+
+      <div
+        class="iui-input-grid"
+        data-iui-status="positive"
+      >
+        <div class="iui-input-with-icon">
+          <input
+            class="iui-input"
+            placeholder="Placeholder…"
+          />
+          <svg-status-success
+            class="iui-svg-icon iui-end-icon"
+            data-iui-icon-color="positive"
+            data-iui-icon-size="m"
+            aria-hidden="true"
+          ></svg-status-success>
+        </div>
+      </div>
+
+      <div
+        class="iui-input-grid"
+        data-iui-status="positive"
+      >
+        <div class="iui-input-with-icon">
+          <input
+            class="iui-input"
+            placeholder="Placeholder…"
+            data-iui-size="small"
+          />
+          <svg-status-success
+            class="iui-svg-icon iui-end-icon"
+            data-iui-icon-color="positive"
+            data-iui-icon-size="m"
+            aria-hidden="true"
+          ></svg-status-success>
+        </div>
+      </div>
+
+      <div
+        class="iui-input-grid"
+        data-iui-status="positive"
+      >
+        <div class="iui-input-with-icon">
+          <input
+            class="iui-input"
+            placeholder="Placeholder…"
+            data-iui-size="large"
+          />
+          <svg-status-success
+            class="iui-svg-icon iui-end-icon"
+            data-iui-icon-color="positive"
+            data-iui-icon-size="m"
+            aria-hidden="true"
+          ></svg-status-success>
+        </div>
+      </div>
+
+      <h3>Select</h3>
+      <div class="iui-input-with-icon">
+        <div
+          tabindex="0"
+          id="test-select-1"
+          class="iui-select-button"
+        >
+          <span class="iui-content">Starts with</span>
+        </div>
+        <svg-caret-down-small
+          class="iui-end-icon iui-svg-icon"
+          aria-hidden="true"
+        ></svg-caret-down-small>
+      </div>
+
+      <div class="iui-input-with-icon">
+        <div
+          tabindex="0"
+          id="test-select-1"
+          class="iui-select-button"
+          data-iui-size="small"
+        >
+          <span class="iui-content">Starts with</span>
+        </div>
+        <svg-caret-down-small
+          class="iui-end-icon iui-svg-icon"
+          aria-hidden="true"
+        ></svg-caret-down-small>
+      </div>
+
+      <div class="iui-input-with-icon">
+        <div
+          tabindex="0"
+          id="test-select-1"
+          class="iui-select-button"
+          data-iui-size="large"
+        >
+          <span class="iui-content">Starts with</span>
+        </div>
+        <svg-caret-down-small
+          class="iui-end-icon iui-svg-icon"
+          aria-hidden="true"
+        ></svg-caret-down-small>
+      </div>
+
+    </section>
 
     <h2>Default</h2>
     <section id="demo-default">


AdamMeza-Bentley (approved):
looks good! the new styling is cleaner. Also really impressive use of css selectors

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -192,6 +188,20 @@
...
+  > :is(input, textarea):has(+ .iui-button[data-iui-variant='borderless'], + .iui-svg-icon[data-iui-padded='true']) {
+    padding-inline-end: 0;
+  }
+
+  > :is(.iui-svg-icon[data-iui-padded='true'], .iui-button[data-iui-variant='borderless']) + :is(input, textarea) {
+    padding-inline-start: 0;
+  }
+
+  // input/textarea is not the first child and does not directly follow a borderless button or padded icon
+  > :not(.iui-svg-icon[data-iui-padded='true']):not(.iui-button[data-iui-variant='borderless']) + :is(input, textarea) {
```
mayank99, line 201: These selectors are brittle and will be really hard to maintain. If we change anything in any of those other components, all of this will break.

Should we introduce a new class and component called `InputWithDecorations.Icon`? It can wrap `<Icon padded>` and its only purpose would be to collapse this padding.

rohan-kadkol: Ohh, I see. Are you suggesting something like https://github.com/iTwin/iTwinUI/pull/1881/files/8b9ccc1c6c41a5328d8580e62b9e3e188d936602..52047f92d2e3fcb9f3edafd880b42ee1447be93c?

mayank99: Close. The `.iui-button[data-iui-variant='borderless']` selector also needs to go. It can use the same class I think, since we already have a place to put it (`InputWithDecorations.Button`).

rohan-kadkol: Sounds good, done in https://github.com/iTwin/iTwinUI/pull/1881/files/52047f92d2e3fcb9f3edafd880b42ee1447be93c..1b2f2a9bedf0fec8b1161861f3d3a117a4dd7f48.

I reused `.iui-input-flex-container-icon` in a new private component `InputFlexContainerButton`. This also simplified the selectors.

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -29,10 +29,3 @@
     padding-inline: 0; // To center iui-searchbox-open-button inline in collapsed state
   }
 }
-
-.iui-search-icon {
```
mayank99, line 33: Is this a breaking change?

It looks like we only use `iui-search-icon` and `iui-search-input` for padding adjustments, so I'm ok with pushing it in a minor release.

All instances of these classes should still be removed though.

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
```diff
@@ -60,33 +65,39 @@ const InputWithDecorationsButton = React.forwardRef((props, ref) => {
...
   'span',
-  React.ComponentProps<typeof IconButton>
+  React.ComponentProps<typeof InputFlexContainerButton>
+>;
+
+const InputWithDecorationsIcon = React.forwardRef((props, ref) => {
+  return <InputFlexContainerIcon ref={ref} {...props} />;
+}) as PolymorphicForwardRefComponent<
+  'span',
+  React.ComponentProps<typeof InputFlexContainerIcon>
```
mayank99, line 86: Why not just reassign it?

```jsx
const InputWithDecorationsIcon = InputFlexContainerIcon;
```

mayank99: Also add separators and `displayName`

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
```diff
@@ -6,6 +6,8 @@ import React from 'react';
 import cx from 'classnames';
 import type { PolymorphicForwardRefComponent } from '../props.js';
 import { Box } from './Box.js';
+import { Icon, type IconProps } from '../../Icon/Icon.js';
+import { IconButton, type IconButtonProps } from '../../Buttons/IconButton.js';
```
mayank99, line 10: we should use `import type` for better compatibility (https://github.com/iTwin/iTwinUI-react/pull/996)

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -192,6 +188,20 @@
...
+    padding-inline-end: 0;
+  }
+
+  > :is(.iui-input-flex-container-icon) + :is(input, textarea) {
+    padding-inline-start: 0;
+  }
+
+  // input/textarea is not the first child and does not directly follow .iui-input-flex-container-icon (Icon or IconButton)
+  > :not(.iui-input-flex-container-icon) + :is(input, textarea) {
+    padding-inline-start: var(--iui-size-2xs);
```
mayank99, line 202: What specific case is this handling? Isn't it the general case, where we should set `padding-inline` on all `:is(input,textarea)`?

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
```diff
@@ -35,3 +37,35 @@ export const InputFlexContainer = React.forwardRef((props, ref) => {
     </Box>
   );
 }) as PolymorphicForwardRefComponent<'div', InputFlexContainerProps>;
+
+/**
+ * Button inside InputFlexContainer that collapses the padding between the button and the input/textarea.
+ * @private
+ */
+export const InputFlexContainerButton = React.forwardRef((props, ref) => {
```
mayank99, line 45: I feel like these extra private components are not really necessary. They can live inside `InputWithDecorations.tsx`.

But I understand that `SearchBox` is currently using `InputFlexContainer`, not `InputWithDecorations`. So I'm fine with keeping it for now, but in the future we should try to get rid of `InputFlexContainer.tsx` altogether.

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
rohan-kadkol: Done (https://github.com/iTwin/iTwinUI/pull/1881/commits/bdbd7db0202214c08d014ff5722a96e8e1cff4fe).

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
mayank99: Why did the length of other separators change?

image

Maybe the vscode snippet is the wrong length and should be fixed?

packages/itwinui-css/src/input-container/input-container.scss
rohan-kadkol: This `padding-inline-start` was initially in the general case (`input, textarea`). I moved it in a separate selector to prevent conflicts between the two selectors that set `padding-inline-start`.

The specificity difference between the [general selector](https://github.com/iTwin/iTwinUI/pull/1881/files#diff-79a618b1da06b9cedff0df3a138d8a8da22a8185e8f0b5a2cf6c8aeb23a49b5cR167-R168) and the [specialized selector](https://github.com/iTwin/iTwinUI/pull/1881/files#diff-79a618b1da06b9cedff0df3a138d8a8da22a8185e8f0b5a2cf6c8aeb23a49b5cR196) resolves the conflict correctly, but I felt being more explicit in the selector might be safer here.

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
rohan-kadkol: I assumed the VSCode snippets to be the correct lengths. That's why I changed the other existing separator comments to match the VSCode snippet's length.

Upon doing a global search, looks like the VSCode separator length is longer than the separators we're using in our repo. I added a bullet point to the after PR TODOs to fix this.

For now, I manually updated the separators to use the existing separator length instead of the length from the VSCode snippet (https://github.com/iTwin/iTwinUI/pull/1881/commits/1d5897fff1208b0706d36ef9e6a1ddd5b5068465).

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
rohan-kadkol: Sounds good, so I will leave it as-is for now. I added a point to the maintenance issue ([#1571](https://redirect.github.com/iTwin/iTwinUI/issues/1571)) to address this in the future.

packages/itwinui-css/src/input-container/input-container.scss
mayank99: > but I felt being more explicit in the selector might be safer here.

It's not safer, it introduces extra unnecessary complexity. We should utilize the power of the CSS cascade where we can. The most specific rules should be at the bottom.

Here's how I think of it, from top to bottom:

1. Most general case where input is used by itself (same as `main` and this PR):

https://github.com/iTwin/iTwinUI/blob/00cef97bad6e912c74f22328826d85e384c7149c/packages/itwinui-css/src/input-container/input-container.scss#L177

2. Less specific case where input has things before or after it:

(same as `main` branch, missing in this PR)

https://github.com/iTwin/iTwinUI/blob/00cef97bad6e912c74f22328826d85e384c7149c/packages/itwinui-css/src/input-container/input-container.scss#L180-L182

(missing in `main` and this PR)

```css
&:where(:not(:last-child)) {
  padding-inline-end: var(--iui-size-2xs);
}
```

3. Most specific case (this PR)

https://github.com/iTwin/iTwinUI/blob/1d5897fff1208b0706d36ef9e6a1ddd5b5068465/packages/itwinui-css/src/input-container/input-container.scss#L191-L198

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
mayank99: Feel free to also update the separate length in this PR. It's too minor to warrant a new PR 😅

packages/itwinui-css/src/input-container/input-container.scss
rohan-kadkol: I had changed to the more specific selector since I felt relying on specificities to resolve selector conflicts is brittle and could lead to unexpected issues down the line.

But I agree with preferring simplicity part. I don't feel strongly in any direction, so I moved the specific selector into the general case (https://github.com/iTwin/iTwinUI/pull/1881/commits/860771ce726674f54b521bcebbbd2ac1b29090f1). I also added the last child selector that was missing from before this PR.

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
rohan-kadkol: Done 🙂 ([`30c62de` (#1881)](https://github.com/iTwin/iTwinUI/pull/1881/commits/30c62de52fec366b97eaecb0ee2bec2d4b409fed))

packages/itwinui-css/src/searchbox/searchbox.scss
rohan-kadkol: Although `iui-search-icon` and `iui-search-input` are no longer used for any styling at the moment, would it be alright to still keep it in our itwinui-css API? That way if we need these classes in the future for some custom styling, we can re-use them.

If by iTwinUI v4 we don't have any use of these classes, then we can remove them in v4 as a breaking change.

mayank99: Sure, but lets maybe add empty rules with a comment?

```scss
// noop
.iui-search-icon {}
.iui-search-input {}
```

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
mayank99: Missed one https://github.com/iTwin/iTwinUI/blob/30c62de52fec366b97eaecb0ee2bec2d4b409fed/packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx#L14

packages/itwinui-css/src/searchbox/searchbox.scss
rohan-kadkol: Sounds good, done (https://github.com/iTwin/iTwinUI/pull/1881/commits/623180c5a8f51e8542f75ba5dc814473ec891e63).

packages/itwinui-react/src/core/utils/components/InputFlexContainer.tsx
rohan-kadkol: Ohh, fixed (https://github.com/iTwin/iTwinUI/pull/1881/commits/d9466bae103ebb3f28cda5506122d29886730a92). Also double checked for other such changes in this PR and found none.

.changeset/cuddly-doors-pump.md
mayank99, line 1: This changeset is now outdated. A new `minor` changeset mentioning the new class would be ideal

mayank99: also the new class is missing in HTML demos currently?

rohan-kadkol: Yes, I was first resolving the comments to stabilize the changes in the PR. I was then planning to update the demos, tests, changesets, etc.

Since all the comments are resolved now, I will be working on this now.

rohan-kadkol: > This changeset is now outdated. A new `minor` changeset mentioning the new class would be ideal

Updated/Added changesets based on the latest state of this PR (https://github.com/iTwin/iTwinUI/pull/1881/commits/52e1050ed2a01058c5c48ca3682010424e29b96a). I can also try reducing the number of changesets if we feel any changeset is not important.

> also the new class is missing in HTML demos currently?

Updated the HTML demos to include `.iui-input-flex-container-icon` (https://github.com/iTwin/iTwinUI/pull/1881/commits/cbbdbc7eaa02ec9852d4c54cd7adebba97715adf). This also includes some unrelated changes like adding missing `.iui-searchbox` and some minor formatting changes like removing unnecessary new lines.

mayank99: > Updated/Added changesets based on the latest state of this PR ([52e1050](https://github.com/iTwin/iTwinUI/commit/52e1050ed2a01058c5c48ca3682010424e29b96a)). I can also try reducing the number of changesets if we feel any changeset is not important.

Yeah it's too much. These two are not necessary:
image

And the other four can probably be combined into two. And there were also some previous changesets (from #1828) that could maybe be removed/combined with this PR?

---

Also since this is an easy-to-migrate change, maybe we should try to handle it automatically in the migration tool?

packages/itwinui-react/src/core/InputWithDecorations/InputWithDecorations.tsx
```diff
@@ -60,33 +66,40 @@ const InputWithDecorationsButton = React.forwardRef((props, ref) => {
...
-      styleType='borderless'
       disabled={localDisabled ?? isDisabled}
       {...rest}
     >
       {children}
-    </IconButton>
+    </InputFlexContainerButton>
   );
 }) as PolymorphicForwardRefComponent<
   'span',
```
mayank99, line 79: Unrelated to this PR: I think this should be "button" rather than "span"?

rohan-kadkol: Yes, fixed ([`7e93013` (#1881)](https://github.com/iTwin/iTwinUI/pull/1881/commits/7e93013d2504437a80d2a687bb5b274e721a1306)).

.changeset/cuddly-doors-pump.md
rohan-kadkol: > Yeah it's too much. These two are not necessary:

Sounds good, removed (https://github.com/iTwin/iTwinUI/pull/1881/commits/8942e216cde2fc449a0ca28bb3d0d706ee67c27c).

> And the other four can probably be combined into two.

Done (https://github.com/iTwin/iTwinUI/pull/1881/commits/74993b7cfb83c30c2a38ed15e553b37adec62bfd).

> And there were also some previous changesets (from #1828) that could maybe be removed/combined with this PR?

I think only the `.iui-search-icon` one could probably be combined with the `.iui-search-icon` changeset from this PR. But since we're not removing that class from the itwinui-css API, I think that changeset from the earlier PR is not outdated. And thus I feel having separate changesets for them is better.

> Also since this is an easy-to-migrate change, maybe we should try to handle it automatically in the migration tool?

Do you mean create a codemod for iTwinUI v3 users to use `InputWithDecorations.Icon` instead of `Icon` in `InputWithDecorations`?

mayank99: > > And there were also some previous changesets (from #1828) that could maybe be removed/combined with this PR?
> 
> I think only the `.iui-search-icon` one could probably be combined with the `.iui-search-icon` changeset from this PR. But since we're not removing that class from the itwinui-css API, I think that changeset from the earlier PR is not outdated. And thus I feel having separate changesets for them is better.

Can't it be removed? If the class literally is a noop now, then it doesn't make sense to have a changeset saying that width/padding has changed.

> > Also since this is an easy-to-migrate change, maybe we should try to handle it automatically in the migration tool?
> 
> Do you mean create a codemod for iTwinUI v3 users to use `InputWithDecorations.Icon` instead of `Icon` in `InputWithDecorations`?

Ah nevermind. I was under the impression that `InputWithDecorations` also exists in v2, but it looks like it only exists in v3 (unlike `SearchBox`, where no API change is required).

rohan-kadkol: > Can't it be removed? If the class literally is a noop now, then it doesn't make sense to have a changeset saying that width/padding has changed.

Initially, I wasn't sure how to say the "search icon" without using its classname `.iui-search-icon` or react component name (`SearchBox.Icon`).

But now, I did the following change (https://github.com/iTwin/iTwinUI/pull/1881/files/74993b7cfb83c30c2a38ed15e553b37adec62bfd..f0ac7f40b9181573f54ad8f8a372543b4160eb46):

```diff
- `.iui-search-icon` …
+ The `.iui-svg-icon` inside the `.iui-searchbox` …
```

mayank99 (approved):

PR #1882, opened 2024-02-26
add template for example requests

mayank99:
## Changes

As discussed previously, developers should have the ability to request example code to better understand how to use our components.

This PR adds a simple link to the ["Examples" category of Discussions](https://github.com/iTwin/iTwinUI/discussions/categories/examples). This link will show up when trying to [create a new issue](https://github.com/iTwin/iTwinUI/issues/new/choose), similar to the "Questions" link.

I thought about using a proper issue template, but decided Discussions is a better place for it.

## Testing

N/a

## Docs

N/a


rohan-kadkol (approved):

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -2,3 +2,6 @@ contact_links:
   - name: Questions 🤔
     url: https://github.com/iTwin/iTwinUI/discussions/categories/q-a
     about: Ask questions and get help from contributors through GitHub Discussions.
+  - name: Request an example 👀
+    url: https://github.com/iTwin/iTwinUI/discussions/categories/examples
```
mayank99, line 6: This link opens the "Examples" page showing existing discussions, but we also have the ability to pre-fill the form using a link like this: https://github.com/iTwin/iTwinUI/discussions/new?category=examples&title=blah&body=pls%20show%20me%20an%20example%20of

Not sure how useful that would be, so I went with the simpler link.

PR #1884, opened 2024-02-28
Adds `contain: inline-size` to horizontal tabs wrappers

Ben-Pusey-Bentley:
## Changes

Adds `contain: inline-size` to horizontal tabs wrappers. This change fixes various bugs stemming from the tabs wrapper resizing when a tab of a different width is selected. One involved the tabs changing heights when clicking from the third tab to any other.

Before:

https://github.com/iTwin/iTwinUI/assets/141063311/5241fd75-c7d6-4bec-8e34-3169283c1136

After:

https://github.com/iTwin/iTwinUI/assets/141063311/a0639571-f169-4e12-9065-5586f82b62e2

Another caused the rest of the page to shift when a different tab was selected. Watch the right side navigation closely.

Before:

https://github.com/iTwin/iTwinUI/assets/141063311/4c000a4c-058b-4175-be1d-9a55ba2dd725

After:

Uploading Borderless tabs - Tabs _ Ladle - Google Chrome 2024-02-28 12-35-42.mp4…

## Testing

Changed the tabs css workshop page to use `width: auto` for all demos. This change was made because having `width: fit-content` and `contain: inline-size` made the width of the tabs 0. The visual tests otherwise show no differences for this change.

## Docs

Added patch changeset for this change.

### DIFF
--- a/.changeset/polite-apples-promise.md
+++ b/.changeset/polite-apples-promise.md
@@ -1,6 +1,5 @@
 ---
 "@itwin/itwinui-css": patch
-"@itwin/itwinui-react": patch
 ---
 
-Fixes bug that caused the tabs `Wrapper` to change size dependent on which `Tab` was active.
+Fixes bug that caused the `.iui-tabs-wrapper` to change size dependent on which `.iui-tab` was active.
--- a/.changeset/polite-apples-promise.md
+++ b/.changeset/polite-apples-promise.md
@@ -1,5 +1,6 @@
 ---
 "@itwin/itwinui-css": patch
+"@itwin/itwinui-react": patch
 ---
 
 Fixes bug that caused the tabs `Wrapper` to change size dependent on which `Tab` was active.
--- a/.changeset/polite-apples-promise.md
+++ b/.changeset/polite-apples-promise.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
+---
+
+Fixes bug that caused the tabs `Wrapper` to change size dependent on which `Tab` was active.
--- a/.changeset/curly-ads-march.md
+++ b/.changeset/curly-ads-march.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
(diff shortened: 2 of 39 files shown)


.changeset/polite-apples-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
```
rohan-kadkol, line 2: You would also need to include `@itwin/itwinui-react` in the changeset since although these are only css changes, `itwinui-react` is also affected. Another reason is that `itwinui-react` does not have a dependency on `itwinui-css`. So from a user's perspective, only `itwinui-react` will seem changed, and thus we need a `itwinui-react` changeset.

```suggestion
"@itwin/itwinui-css": patch
"@itwin/itwinui-react": patch
```

Ben-Pusey-Bentley: Done in latest commit.  ([63ea24c](https://github.com/iTwin/iTwinUI/pull/1884/commits/63ea24c32ddebf81286072efc760a2f02f6bdfc6))

.changeset/polite-apples-promise.md
```diff
@@ -0,0 +1,6 @@
...
+"@itwin/itwinui-react": patch
+---
+
+Fixes bug that caused the tabs `Wrapper` to change size dependent on which `Tab` was active.
```
rohan-kadkol, line 6: nit: Actually, you could consider splitting the changeset into two changesets with mostly the same content. The only difference would be using classnames in css changesets and exact component names in react changesets.

E.g. [react changeset](https://github.com/iTwin/iTwinUI/pull/1828/files#diff-1ae21bfe3b9b7e4ed39fbc59f54c851680020727241decbbc62dce30c2f3c88c), [css changeset](https://github.com/iTwin/iTwinUI/pull/1828/files#diff-65af2d9ac4e82ffa6b9332a961afe3bba98a8bf9de40e46e745cb84ce67a36bb).

rohan-kadkol (approved):

.changeset/polite-apples-promise.md
Ben-Pusey-Bentley: Done in latest commit. [(58ae4fd)](https://github.com/iTwin/iTwinUI/pull/1884/commits/58ae4fd2a674385dac3449c5e84510bb26e64c5d)

AdamMeza-Bentley (approved):

PR #1887, opened 2024-02-29
Adds Remix Playground

Ben-Pusey-Bentley:
## Changes

Adds Remix playground to the iTwinUI playgrounds. This is a playground that runs remix through the remix vite plugin, for testing the UI components with remix. The playground runs on `localhost:1704`. Also updates the concurrency value that the `dev` command is run with to 12, since we now have 11 dev processes that should be running simultaneously.

## Testing

Created the remix playground and made sure that it appears in the portal and runs on `dev`.

## Docs

N/A


playgrounds/remix/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "remix-playground",
+  "private": true,
+  "sideEffects": false,
+  "version": "0.1.0",
+  "type": "module",
+  "scripts": {
+    "build": "remix vite:build",
+    "dev": "remix vite:dev",
+    "lint": "",
```
mayank99, line 10: If we don't want linting, you can remove this script. Also remove `.eslintrc.cjs` as well as all `eslint` and `@typescript-eslint` related dependencies.

playgrounds/remix/app/entry.server.tsx
```diff
@@ -0,0 +1,140 @@
...
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext,
+  // This is ignored so we can keep it in the template for visibility.  Feel
+  // free to delete this parameter in your app if you're not using it!
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  loadContext: AppLoadContext,
+) {
+  return isbot(request.headers.get('user-agent') || '')
```
mayank99, line 27: Since this is a simple playground where we don't care about bots, we can simplify this file to remove `handleBotRequest`. Basically `handleBrowserRequest` can be the default export.

Also `isbot` can be removed from dependencies in package.json (make sure to run `pnpm install` after that).

playgrounds/remix/app/routes/_index.tsx
```diff
@@ -0,0 +1,17 @@
+import type { MetaFunction } from '@remix-run/node';
+import { Button } from '@itwin/itwinui-react';
+
+export const meta: MetaFunction = () => {
+  return [
+    { title: 'New Remix App' },
+    { name: 'description', content: 'Welcome to Remix!' },
```
mayank99, line 7: nit: You can change this meta information if you want.

playgrounds/remix/public/favicon.ico
mayank99, line 1: nit: The favicon can be removed

playgrounds/remix/package.json
Ben-Pusey-Bentley: Will do.

playgrounds/remix/app/entry.server.tsx
Ben-Pusey-Bentley: Alright, will do.

playgrounds/remix/app/routes/_index.tsx
Ben-Pusey-Bentley: Okay, will do.

playgrounds/remix/public/favicon.ico
Ben-Pusey-Bentley: Will do.

playgrounds/remix/package.json
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1887/commits/55c74b1f708bb9d8566d982957f8a55be5973ace)

playgrounds/remix/app/entry.server.tsx
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1887/commits/d42b971de83f06b9700741ce183afc2e74e6e82e)

playgrounds/remix/public/favicon.ico
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1887/commits/2f9f66e45e1939e1a6e5af7f01066a202423b9f6)

package.json
```diff
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "build": "turbo run build",
-    "dev": "turbo run dev",
+    "dev": "turbo run dev --concurrency=12",
```
mayank99, line 30: Does this actually work? I'm having a bit of trouble getting the dev server to open consistently when running `pnpm dev` from the root.

Sometimes it doesn't open at all, while other times it shows an error

image

Ben-Pusey-Bentley: That seems to be a different issue. Adding to the concurrency is meant to fix the issue where not all of the dev commands would run concurrently, because there are 11 and the default max is 10. I thought the issue you just posted a screenshot for had been resolved by fixing the pnpm-lock in one of my earlier commits, but I guess not. Let me see what I can figure out.

AdamMeza-Bentley: im also working on running tests concurrently. I haven't gotten anything to work yet but was looking into [GNU parallel](https://www.gnu.org/software/parallel/). Would something like this work or are you solving a different issue?

mayank99: What? This has nothing to do with running tests or GNU parallel.

The error is happening because of a race condition between the dev servers of `@itwin/itwinui-react` and `remix-playground`. Remix seems to start up before itwinui-react is ready and doesn't seem to be able to recover later.

Ben-Pusey-Bentley: Yeah, the race condition is hard to solve in this case, since we want both of the dev servers to be running simultaneously. I have found a hacky solution by having the `remix-playground dev` command depend on the `@itwin/itwinui-react build` command, which delays the server startup long enough to prevent the race condition. This is obviously not an ideal solution, since it delays the start of the dev server. I can also create a similar issue with the `@itwin/itwinui-css` and `portal` servers, but the portal is able to recover from the error. It seems that Astro recovers better, but I don't yet know why. Similarly, I can most consistently get these issues to happen when running `dev` immediately after `install`, running `build` first seems to prevent them. Again I don't know why, since the `dev` commands run clean on their startup.

mayank99: This might be an opportunity to improve our `dev` command if you want to look into that.

The goal is simple:
> Running `pnpm dev` should start up the dev servers as fast as possible (ideally under five seconds).

Introducing a dependency on `build` definitely makes it slow, but we don't need everything from `build`. (`tsc` is the slowest part of the process, and we definitely don't need that).

playgrounds/remix/app/entry.server.tsx
mayank99: When looking into https://github.com/iTwin/iTwinUI/pull/1887#discussion_r1511517993, I found that you can actually get rid of `entry.client` and `entry.server` files completely. Remix will automatically use the defaults. (This does mean we need to bring back `isbot`, but that's ok)

Ben-Pusey-Bentley: Will do.

package.json
Ben-Pusey-Bentley: I'll take a look. If there is a minimal build that helps, it could fix all issues of this type.

mayank99: Ideally, here's what would happen when running `pnpm dev`:
1. Build `@itwin/itwinui-variables/index.css` and `@itwin/itwinui-css/css/all.css` only if stale version doesn't exist.
   - This should take 0 seconds if file already exists. Maybe a few seconds if fresh build.
2. Transpile all TS(X) files inside `@itwin/itwinui-react` into plain JS.
   - This should take a fraction of a second, because it does not need to run `tsc` to produce `.d.ts` files.
3. Run CSS modules transpilation for producing `styles.css` (see [`styles.js` folder](https://github.com/iTwin/iTwinUI/tree/main/packages/itwinui-react/src/styles.js)).
   - This should take <1 second.
4. Start all dev servers, and also start "watching" SCSS/TS changes inside all `packages`. For every change, basically repeat the above 3 steps on-demand, rather than building the whole thing.
   - This time we will lazily produce `.d.ts` files (note that it was skipped during step 2).

That's pretty close to what already happens today, except steps 1-3 are currently non-blocking, and there are some inconsistencies and imperfections in various places.

---

The "dream" scenario would be to not even need steps 1-3 _at all_. [Turborepo docs](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages) describe it best:

> _Instead of bundling our package ourselves, we're going to make the app which imports the package bundle it for us._

The main hurdle to this right now is that our package.json `exports` explicitly map to `esm`/`cjs` folders, whereas our source code lives in the `src` folder. If we can change the `exports` field only during local development, that might solve all the issues.

https://github.com/iTwin/iTwinUI/blob/ef62ab84978f12540a9d9b3d75606cd6d2039131/packages/itwinui-react/package.json#L10-L15

---

If you're interesting in looking into this, it can be a follow-up task. I'm ok with merging this PR even if the remix `dev` server is in a broken state.

playgrounds/remix/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@itwin/itwinui-react": "*",
+    "@remix-run/node": "^2.7.2",
+    "@remix-run/react": "^2.7.2",
+    "@remix-run/serve": "^2.7.2",
+    "isbot": "^4.1.0",
+    "react": "^18",
+    "react-dom": "^18"
+  },
+  "devDependencies": {
+    "@remix-run/dev": "^2.7.2",
```
mayank99, line 23: You can also bump all of these from `2.7.2` to `2.8.0`

package.json
Ben-Pusey-Bentley: Yeah, that definitely sounds interesting. Certainly worth following up on.

playgrounds/remix/README.md
mayank99, line 1: You can remove README

playgrounds/remix/package.json
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1887/commits/0506f39f64bdfa47817228642cc4731086e163ed)

mayank99 (approved):
LGTM, with https://github.com/iTwin/iTwinUI/pull/1887/files#r1511765741 to be tackled in a future PR

PR #1889, opened 2024-02-29
allow visually adjusting borderless buttons

mayank99:
## Changes

**Background**: Borderless buttons occupy extra space "physically" via padding. This helps them feel more usable with a larger hit target. However, due to the lack of borders, this extra space is "invisible" and makes things look unaligned unless the button is hovered (which changes the background-color slightly).

This PR:
- Adds `data-iui-shift` attribute selectors to borderless buttons for visual alignment.
- Updates close button alignment in `Dialog` using this new attribute.
- is a prerequisite for #1863, where the filter button wraps into its own line and needs to be visually aligned with other text in the table header cell.

**Note**: There is no equivalent React prop for the data attribute, because I'm not sure if this functionality should be exposed to the user. It can always be added in the future if needed.

## Testing

Updated visual tests.

The dialog close button is now visually aligned to the right edge, same as the submit/cancel buttons at the bottom.

| Before | After |
| --- | --- |
| image | image |

## Docs

Added changesets.


rohan-kadkol (approved):

apps/react-workshop/cypress-visual-screenshots/baseline/Dialog.test.ts-Basic.png
rohan-kadkol, line 1: I think you also need to approve the similar test image failures in `Modal`.

mayank99: whoops, yes https://github.com/iTwin/iTwinUI/pull/1889/commits/9433b0ac195d30d6793c26ef179e356973cf43e1

FlyersPh9 (approved):
This is cool!  This issue was something that has always bothered me but I never thought of as something we could fix.  I dig it! 👍

packages/itwinui-css/src/button/borderless.scss
```diff
@@ -26,6 +26,16 @@
     --_iui-button-border-color: transparent;
     --_iui-button-icon-fill: var(--iui-color-icon-disabled);
   }
+
+  // #region Visual adjustments to help align based on the "visible" edge of the button content.
+  &[data-iui-adjust='left'] {
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: I personally prefer "shift", but I think either will be fine so long as it is consistent.

Ben-Pusey-Bentley (approved):

packages/itwinui-css/src/button/borderless.scss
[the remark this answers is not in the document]
rohan-kadkol: I thought of something like "stick-position" or "stick-edge", but that might be too long. Among "adjust" and "shift", I am leaning towards "shift". But "adjust" is also alright.

packages/itwinui-css/src/button/borderless.scss
```diff
@@ -29,11 +29,11 @@
 
   // #region Visual adjustments to help align based on the "visible" edge of the button content.
   &[data-iui-shift='left'] {
-    translate: calc(var(--_iui-button-padding-inline) * -1);
+    margin-inline-start: calc(-1 * var(--_iui-button-padding-inline));
```
mayank99, line 32: I changed the implementation to use negative `margin`, so that its "original position" also shifts. With `transform`, it only seems to move visually while occupying a false original position.

The benefit of this can be seen when there are other items next to it. For example, other custom actions in information-panel can shift along with the close button, and the container retains its size (not committed in this PR but can be done in the future).

| with `transform` | with `margin` |
| --- | --- |
| image | image |

PR #1892, opened 2024-03-01
Fixed DOM in textarea css demos

rohan-kadkol:
## Changes

After-PR TODO from #1828.

This PR fixes the DOM in the textarea css demos to match the `TextArea`/`LabeledTextArea` DOM in React.

This leads to some fixes like the resizer handle appearing to the left of the status icon instead of to the right of it and the missing padding/spacing on the status icon.

|Before|After|
|-|-|
|image|image|

Additionally, this PR removed the `.iui-input-decorator-icon` class from the DOM since I could not find any reference for it when I did a global search for it. I also searched just `decorator` in case we used `&` in scss and so a perfect global search match will not be found. But I still could not find any reference to that class.

## Testing

Confirmed that the DOM looks correct.

## Docs

N/A

## After PR TODOs:

* [ ] (low-priority) Try removing `.iui-input-with-icon` wrapper from `LabeledInput`/`LabeledTextarea` when no there's no icon (https://github.com/iTwin/iTwinUI/pull/1892#discussion_r1511221582)


apps/css-workshop/pages/textarea.html
```diff
@@ -47,27 +47,31 @@ <h2>Default</h2>
...
       >
         <div class="iui-input-label iui-required">Textarea label</div>
-        <textarea
-          placeholder="Placeholder…"
-          class="iui-input"
-          rows="3"
-          id="test-default-singular"
-          required
-        ></textarea>
+        <div class="iui-input-with-icon">
```
mayank99, line 50: Why are we adding this wrapper to the cases where we don't have a status icon?

AdamMeza-Bentley (approved):

mayank99 (approved):

apps/css-workshop/pages/textarea.html
rohan-kadkol: Since #1753, we always wrap the `LabeledInput`'s/`LabeledTextarea`'s input/textarea in an `.iui-input-with-icon`. E.g. [`LabeledTextarea → Basic` story](https://itwin.github.io/iTwinUI/react/?story=labeledtextarea--basic&theme=dark)

<details>
<summary>Screenshot</summary>

image

</details>

I think this was possible because we modified the styles to handle the case where the `.iui-input-with-icon` had no icon ([example](https://github.com/iTwin/iTwinUI/pull/1753/files?file-filters%5B%5D=.scss&show-viewed-files=true#diff-4623585aa5d420e8e06bbb618dd09a22a22d7946948cd311def55eaa7a64e9a0L111-R121)).

But yes, since the class name include "with icon", I guess having an icon always is better. So, I removed the `.iui-input-with-icon` from the textarea demos when there was no end icon.

I added a TODO in the PR description to try updating the react code for `LabeledInput`/`LabeledTextarea` too.

mayank99: I'm also ok with leaving `LabeledInput`/`LabeledTextarea` unchanged for a while. There are more important things to tackle in the meantime.

PR #1897, opened 2024-03-05
update `SvgStatusError` to use the new octagon shape

mayank99:
## Changes

Leftover from https://github.com/iTwin/iTwinUI-icons/pull/57 / https://github.com/iTwin/iTwinUI/pull/1732

## Testing

Force updated any affected visual test images.

## Docs

Added changesets.


rohan-kadkol (approved):

.changeset/lemon-seahorses-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+All instances of `<SvgStatusError>` have been updated to reflect the new octagon shape from the latest version of iTwinUI-icons. Some components affected: `Alert`, `Toast`, `StatusMessage`.
```
rohan-kadkol, line 5: Is `<SvgStatusError>` an itwinui-react package implementation detail? If so, maybe the changeset can use some other name to refer to the error icon.

mayank99: Technically yes, but it's publicly exposed in `@itwin/itwinui-icons-react` with the same name, so this is probably the best way to refer to it.

PR #1898, opened 2024-03-05
Reconfigure Test Command

AdamMeza-Bentley:
## Changes

Running each test for any one given component took longer than desired and required four long commands, two of which needed to be run in particular folders. This PR solves that problem by reconfiguring the `test ` command. Now `test` will serve three use cases:

- `test` still runs all tests
- `test --filter=[filter]`  runs specific tests when a `--filter` is found in the arguments 
- `test [compName]` e.g. `pnpm test Alert` will run all tests for the Alert Component
 
### Quirks/Issues

Due to our use of `concurrently` all processes must be aborted individually. This is an [issue](https://github.com/open-cli-tools/concurrently/issues/452)  that was [recently addressed](https://github.com/open-cli-tools/concurrently/pull/460). Not sure when/how we can use this update. I opted not to update any dependencies to keep the scope of this PR relatively small.

## Testing

I've tested these changes by running `test` on a number of components. I made sure to test working and broken tests to ensure that errors are caught when thrown. 

## Docs

I've added a section on this command inside of `CONTRIBUTING.md` to demonstrate use.


packages/itwinui-react/vitest.config.ts
mayank99, line 1: No changes should be necessary in vitest config.

Before this PR you could already run `pnpm --filter=@itwin/itwinui-react run test Alert` from the root.

package.json
```diff
@@ -40,17 +41,16 @@
...
-  "dependencies": {},
   "devDependencies": {
     "chokidar-cli": "^3.0.0",
     "concurrently": "^8.2.2",
-    "husky": "^9.0.10",
+    "configs": "workspace:*",
+    "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
-    "prettier": "~3.1.0",
+    "prettier": "~3.1.1",
```
mayank99, line 50: No dependency changes should be necessary for your PR. Updating `prettier` in particular is problematic as it can result in unrelated formatting changes throughout the repo.

package.json
```diff
@@ -29,6 +29,7 @@
     "build": "turbo run build",
     "dev": "turbo run dev --concurrency=12",
     "test": "turbo run test",
+    "test:specific": "node scripts/testSpecific.js",
```
mayank99, line 32: `test:specific` feels too long to type. The goal of this task was to extend the `test` command itself.

Ideal usage:

```sh
# runs all tests
pnpm test

# runs all tests in css-workshop
pnpm run test --filter=css-workshop

# runs Alert tests in all workspaces
pnpm test Alert
```

testing/a11y/src/SpecificComponent.cy.tsx
mayank99, line 1: You shouldn't need to create a whole different file here.

Can you elaborate on your reasoning here?

> it was the best solution for this issue as it was cleaner and avoided messy conditionals inside the cypress configuration file

Why can't this filtering happen happen in the existing test file?

scripts/testSpecific.js
mayank99, line 1: Prefer writing this script in [ESM syntax](https://nodejs.org/api/esm.html). See [example](https://github.com/iTwin/iTwinUI/blob/mayank/table-overflow/packages/itwinui-react/scripts/postBuild.mjs).

scripts/testSpecific.js
```diff
@@ -0,0 +1,68 @@
...
+    const [reactResult, cssResult, unitResult, a11yResult] = results;
+
+    console.log('-------------------------------');
+    console.log('      Test      |     Result   ');
+    console.log('-------------------------------');
+    console.log(`React Workshop  |    ${reactResult}`);
+    console.log(`CSS Workshop    |    ${cssResult}`);
+    console.log(`Unit            |    ${unitResult}`);
+    console.log(`A11y            |    ${a11yResult}`);
+    console.log('-------------------------------');
```
mayank99, line 64: Instead of producing all these extra logs, we can forward the console output untampered. (see examples: [1](https://github.com/iTwin/iTwinUI/blob/mayank/table-overflow/packages/itwinui-css/scripts/watchScss.mjs#L34-L35) and [2](https://github.com/iTwin/iTwinUI/blob/mayank/table-overflow/apps/react-workshop/scripts/run-tests.js#L20-L21)).

Also, instead of using naked promises here, we can utilize something like [`concurrently`](https://github.com/open-cli-tools/concurrently#readme) to run multiple scripts together and have their CLI outputs be organized by name and color.

We already use `concurrently` in a basic manner in some places:
https://github.com/iTwin/iTwinUI/blob/9e6c8658cba530789fa44d13168aca9b6cc86472/packages/itwinui-react/package.json#L73

scripts/testSpecific.js
```diff
@@ -0,0 +1,68 @@
...
+      type: 'CSS Visual',
+      command: `pnpm run --filter css-workshop test --filter=${componentName}`,
+    },
+    {
+      type: 'Unit',
+      command: `cd ./packages/itwinui-react/src/core && pnpm test:unit --mode ${componentName} && cd ../../../../..`,
+    },
+    {
+      type: 'A11y',
+      command: `cd ./testing/a11y && pnpm cypress run --component --spec "src/SpecificComponent.cy.tsx" --env componentName=${componentName}`,
```
mayank99, line 30: `cd`ing into directories is generally unnecessary because you run these kinds of commands from the root (using [Pnpm's filter syntax](https://pnpm.io/filtering)).

package.json
AdamMeza-Bentley: So i've been able to work through every issue with the PR besides this one. One approach I've tried has been to replace instances of `pnpm run test` with the direct command `pnpm turbo run test` using `spawn()` in my new script file. This works for most of the use cases, except for the visual tests for css and react which require multiple filters, [which should be supported by Turbo](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#multiple-filters). When investing into `turbo.json` I found that filtering for the specific tests had been configured but can't locate where configuration for specific components is happening. My approach is going to be looking for into the `.config` files.

scripts/testSpecific.js
AdamMeza-Bentley: latest commits use this syntax. I haven't found anything in the documentation about formatting final outputs or configuring a single output. I don't know if this is a priority or not.

scripts/testSpecific.js
AdamMeza-Bentley: fixed this using the filter syntax you linked to

scripts/testSpecific.js
AdamMeza-Bentley: fixed this

testing/a11y/src/SpecificComponent.cy.tsx
AdamMeza-Bentley: I was mistaken, I made changes to the existing tests to support both use cases

package.json
AdamMeza-Bentley: this was unintentional. reverted dependencies back to main

packages/itwinui-react/vitest.config.ts
AdamMeza-Bentley: using this now

AdamMeza-Bentley (commented):
**One Issue** is that using `ctrl + C` to terminate the test process after running `pnpm test Alert`, for example, no longer works. Tests continue to run til completion. I believe this is related to how I'm using `spawn()`. I'm looking into it

package.json
AdamMeza-Bentley: Current solution is working for these use cases. There are some user experiences issues I've highlighted in another comment

AdamMeza-Bentley:
> **One Issue** is that using `ctrl + C` to terminate the test process after running `pnpm test Alert`, for example, no longer works. Tests continue to run til completion. I believe this is related to how I'm using `spawn()`. I'm looking into it

Looking more closely into this I found it's an issue with `concurrently` that's being addressed in an [open pr.](https://github.com/open-cli-tools/concurrently/pull/460) Not sure if we want to merge this as is or wait until they merge that.

scripts/runTests.js
mayank99, line 1: I can't even get it to run :/

```
pnpm run test Alert
```

image

mayank99, line 1: We now have e2e (another kind of test) that also needs to be added to this script. See #1918

scripts/testSpecific.js
mayank99: The `-n` flag (described in the [documentation](https://github.com/open-cli-tools/concurrently?tab=readme-ov-file#usage)) can be used to organize the final output by name

mayank99:
> > **One Issue** is that using `ctrl + C` to terminate the test process after running `pnpm test Alert`, for example, no longer works. Tests continue to run til completion. I believe this is related to how I'm using `spawn()`. I'm looking into it
> 
> Looking more closely into this I found it's an issue with `concurrently` that's being addressed in an [open pr.](https://github.com/open-cli-tools/concurrently/pull/460) Not sure if we want to merge this as is or wait until they merge that.

That PR looks unrelated (it's specific to `max-processes). Have you confirmed whether the problem is with `concurrently` or with `spawn`? You can try running `concurrently` directly in your terminal to see if it has the same issue.

scripts/runTests.js
AdamMeza-Bentley: this is odd. is your package.json updated?

mayank99: Not sure what you mean. I'm running the latest version of your branch, so it has whatever changes you made in `package.json`.

scripts/testSpecific.js
AdamMeza-Bentley: added some prefix formatting including colors in my last commit. I was hoping to get a single chart like output but this is a good solution as well

scripts/runTests.js
AdamMeza-Bentley: Not sure what's going on here. My assumption is that it might be related to the difference in turbo within a mac environment. I found this similar issue on git hub. https://github.com/vercel/turbo/issues/3918. Would you be able to provide more information about the error?

mayank99: This is a problem with your script, not a bug in Turbo. "Alert" should not even passed directly to Turbo (which is what's happening currently). `turbo run test Alert` is meaningless.

You likely need to fix the script to correctly handle what should/shouldn't be passed to Turbo. My guess is your `getComponentNames` function (which I'm still failing to understand what it does exactly) only works for Windows.

In any case, you need to forward only these two specific cases to Turbo. Everything else should be handled by your script.
1. If no arguments provided.
2. If arguments contain `--filter`.

You can try testing using [WSL](https://learn.microsoft.com/en-us/windows/wsl/) and/or by temporarily changing the [CI workflow](https://github.com/iTwin/iTwinUI/blob/main/.github/workflows/build.yml) to run the specific command.

Until it is working properly, maybe move your PR to draft.

AdamMeza-Bentley: So the goal of `getComponentNames` was to return an array of component names by reading files in the repo that end in `test.ts`. Given the way testing works now an invalid component name like `Alertt` would still have passing tests. The goal was to make sure that only valid component names get passed along to the custom script. However if simply checking from the presence of `--filter` is more universally functional than that works as well. 

I'll convert it to a draft until I can it's working properly for all cases

AdamMeza-Bentley:
All functionality is working on my machine, as well as in the dev container.

scripts/runTests.mjs
mayank99, line 1: i still can't get it to run 🫤


```console
$ pnpm run test Alert

> itwinui-monorepo@0.0.0 test
> node scripts/runTests.mjs "Alert"

node:assert:125
  throw new AssertionError(obj);
  ^

AssertionError [ERR_ASSERTION]: [concurrently] no commands provided
    at concurrently (/node_modules/concurrently/dist/src/concurrently.js:36:22)

```

what commands did you use for testing it?

AdamMeza-Bentley: dang! Both those commands are passing for me on my machine and also in the dev environment. I had removed the `componentNames` function as that wouldn't work across environments. I think it might be related to how I'm parsing the CL arguments with `process.argv`. You're on a mac right?

mayank99: yeah, i'm on mac.

it looks like it works fine if i change the code to use quotes around the colors.

```
-c "#8BCC2F,#D8598C,#539AD0,#D1B52A,white"
```

AdamMeza-Bentley: found this thread explaining it. It's a node thing. I'll look into how I would write a script that can handle both use cases. 
https://github.com/nodejs/node/issues/13968#issuecomment-311785964

mayank99: it's not an npm thing. that thread is unrelated.

you're just missing quotes, like i mentioned in my previous comment. it's probably that the macOS shell requires it to disambiguate between args passed to an executable

AdamMeza-Bentley: ahh yeah that makes more sense. Thanks i'll add those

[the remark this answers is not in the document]
AdamMeza-Bentley: A quick google says yes there is. I'll look into that more though

scripts/runTests.mjs
```diff
@@ -0,0 +1,31 @@
...
+  const argument = process.argv[2];
+
+  if (argument?.includes('filter')) {
+    spawn(`turbo run test ${argument}`, {
+      stdio: 'inherit',
+      shell: true,
+    });
+  } else if (argument) {
+    spawn(
+      `conc -g -n "CSS-workshop,Unit,A11y,React-workshop,e2e" -c "#8BCC2F,#D8598C,#539AD0,#D1B52A,white" "pnpm --filter=css-workshop run test --filter=${argument.toLowerCase()}" "pnpm --filter=@itwin/itwinui-react run test:unit ${argument}" "pnpm --filter "./testing/a11y" run test  --env componentName=${argument}" "pnpm --filter=react-workshop run test --spec="**/${argument}.*"" "pnpm --filter=e2e run test -- ${argument}"`,
```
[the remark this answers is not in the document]
AdamMeza-Bentley: Would the best solution involve not bypassing the turbo pipeline or is running the necessary commands enough?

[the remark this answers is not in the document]
mayank99: I don't think we can *not* bypass the turbo pipeline if we want to pass args to bespoke test commands. But we can run the build ourselves in this script and still use turbo to run dependent build commands and cache outputs.

`css-workshop` needs `pnpm run build --filter=css-workshop`
`unit` doesn't need build
`a11y` needs `pnpm run build --filter=a11y`
`react-workshop` needs `pnpm run build --filter=react-workshop`
`e2e` needs `pnpm run build --filter=e2e`

It might be worth adding these as top-level scripts in `package.json`. For example:

```
"test:css": "pnpm run build --filter=css-workshop && pnpm --filter=css-workshop run test",
… etc
```

and then we can simply run `pnpm run test:css --filter=${argument.toLowerCase()}` inside this JS script.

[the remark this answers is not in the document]
AdamMeza-Bentley: Alright sweet! i'll make that change

scripts/runTests.mjs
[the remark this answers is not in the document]
AdamMeza-Bentley: they support a `--pass-with-no-tests` option which i added to the `package.json` in `e2e` in my last push.

CONTRIBUTING.md
```diff
@@ -324,8 +328,8 @@ We reuse our stories for visual tests by taking screenshots of the story iframes
 #### Running visual tests
 
 1. Make sure you have [Docker](https://www.docker.com/get-started) installed and running.
-2. From the monorepo root, run `pnpm run test --filter=react-workshop`. This will build react-workshop and run all cypress tests in docker.
-   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm --filter=react-workshop run test --spec="**/Alert.*"`. But don't forget to build react-workshop first (pnpm run build --filter=react-workshop).
+2. From the monorepo root, run `pnpm run test --filter=react-workshop` or `pnpm test:react`. This will build react-workshop and run all cypress tests in docker.
+   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm:react --spec="**/Alert.*"`. But don't forget to build react-workshop first (pnpm run build --filter=react-workshop).
```
mayank99, line 332: small typo

```suggestion
   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm test:react --spec="**/Alert.*"`. But don't forget to build react-workshop first (pnpm run build --filter=react-workshop).
```

CONTRIBUTING.md
```diff
@@ -388,13 +392,11 @@ For more information on the Axe rule IDs and their meanings, visit [Deque Univer
 
 The `testing/e2e` workspace facilitates testing of complex scenarios in a real browser. This is achieved by running [Playwright](https://playwright.dev/) tests against a [Remix](https://remix.run/) app.
 
-- **To run tests**, use this command: `pnpm run test --filter=e2e`.
+- **To run tests**, use `pnpm run test --filter=e2e` or `pnpm test:e2e`.
 - **To write tests**, add a new [Remix route](https://remix.run/docs/en/main/discussion/routes#conventional-route-folders) and a `.spec.ts` next to it.
 
 For more details on how to write a test, see [Playwright docs](https://playwright.dev/docs/writing-tests).
```
mayank99, line 399: nit: bring back accidentally(?) deleted separator
```suggestion

---

```

mayank99 (approved):
LGTM! I would wait for another approval though

Ben-Pusey-Bentley (approved):
LGTM.

CONTRIBUTING.md
```diff
@@ -324,8 +328,8 @@ We reuse our stories for visual tests by taking screenshots of the story iframes
 #### Running visual tests
 
 1. Make sure you have [Docker](https://www.docker.com/get-started) installed and running.
-2. From the monorepo root, run `pnpm run test --filter=react-workshop`. This will build react-workshop and run all cypress tests in docker.
-   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm --filter=react-workshop run test --spec="**/Alert.*"`. But don't forget to build react-workshop first (pnpm run build --filter=react-workshop).
+2. From the monorepo root, run `pnpm run test --filter=react-workshop` or `pnpm test:react`. This will build react-workshop and run all cypress tests in docker.
+   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm test:react --spec="**/Alert.*"`. But don't forget to build react-workshop first (pnpm run build --filter=react-workshop).
```
rohan-kadkol, line 332: Since `pnpm test:react` already builds the react-workshop, the last sentence seems to not be needed anymore.

```suggestion
   -  If you only need to run tests for a specific component, you can do so by passing the `--spec` argument to cypress. e.g. for testing `Alert`, you can run `pnpm test:react --spec="**/Alert.*"`.
```

CONTRIBUTING.md
```diff
@@ -48,6 +48,10 @@ To clone and build iTwinUI, you'll need [Git](https://git-scm.com), [Node 18+](h
 
 _Before running this command, make sure Docker is running. See [Visual testing](#visual-testing-css) (CSS and React) sections below for more details._
 
+### To run all tests for a specific component
+
+`pnpm test [component-name]` e.g. `pnpm test Alert`
```
rohan-kadkol, line 53: Is there any restriction of the case of the component-name? E.g. "alert" vs "Alert"? If so, it would be better to mention it.

AdamMeza-Bentley: Yes it needs to be properly capitalized. I'll add a reminder

CONTRIBUTING.md
AdamMeza-Bentley: True!

package.json
```diff
@@ -28,7 +28,12 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev --concurrency=15",
-    "test": "turbo run test",
+    "test": "node scripts/runTests.mjs",
+    "test:css": "pnpm run build --filter=css-workshop && pnpm --filter=css-workshop run test",
+    "test:react": "pnpm run build --filter=react-workshop && pnpm --filter=react-workshop run test ",
+    "test:a11y": "pnpm run build --filter=a11y && pnpm --filter './testing/a11y' run test",
```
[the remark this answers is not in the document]
AdamMeza-Bentley: made that change!

PR #1900, opened 2024-03-06
Edits `dev` pipeline to prevent race condition with playgrounds.

Ben-Pusey-Bentley:
## Changes

This PR edits the `dev` command's pipeline to prevent a module not found error caused by a race condition. This work is based on [this comment from a previous PR.](https://github.com/iTwin/iTwinUI/pull/1887#discussion_r1511765741) This is achieved through two major changes. The first is that the playground packages now use the [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) api to import the source code directly from the react package. Note that this change was not made to the `next-playground`, as that playground fails to build with this api. `next-playground` takes so long to build, however, that it has never finished before the react package in my testing. This method also failed for the `remix-playground`, so the `remix-playground` exclusively is dependent on some additional build steps from the react package. The other change is that the `dev` command is now dependent on the `build:styles` command of the react package. The transformations made in this build command to the class names of the selectors is vital to the styles working with the react components. Therefore, it must complete before we can start booting the dev servers, at least for all of the servers that display react components. Since the `portal` and `css-workshop` packages are not dependent on the react package, they start booting immediately after the `all.css` is built in the css package instead.

## Testing

Ran `pnpm install` immediately followed by `pnpm dev` and made sure none of the playgrounds had breaking issues with the changes or missing modules.

## Docs

N/A


mayank99 (changes requested):
I don't understand this PR. It seems like you misinterpreted [my comment](https://github.com/iTwin/iTwinUI/pull/1887#discussion_r1511765741) and came up with a completely different solution that introduces more hacks and difficulties, without meaningfully improving the dev server experience. I wish you'd discussed this and asked for clarification/feedback before implementing these changes.

- Import maps are a no-go. They require changing the authoring experience, have issues with TypeScript, and do not even work for Remix (which was the main task!)
- A lot of the pipeline changes are unnecessary (for example, `itwinui-react#build:styles` already depends on `itwinui-css#build:css`, and most of the playground dev servers do not need to be blocked because they are self-recovering).

I'd suggest closing this PR for now.

Issue #1144, opened 2023-03-17
Migrate react-table to v8


FlyersPh9:
react-table v7 is no longer maintained and is replaced by tanstack-table v8: https://tanstack.com/table/v8

This will be a big rewrite of our current Table and we need to try to maintain as much of the current feature set as possible.

There are also new features we should consider adding later, such as ["smart" column resizing based on browser window size](https://github.com/iTwin/iTwinUI/issues/895).

piatrashkakanstantinas:
Since iTwinUI 3.17 update, react-table now causes react peer dependency resolution warnings when using react 19.
```shell
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: react-table@7.8.0
npm warn Found: react@19.0.0
npm warn node_modules/react
npm warn   react@"^19.0.0" from the root project
npm warn   7 more (@floating-ui/react, @floating-ui/react-dom, ...)
npm warn
npm warn Could not resolve dependency:
npm warn peer react@"^16.8.3 || ^17.0.0-0 || ^18.0.0" from react-table@7.8.0
npm warn node_modules/@itwin/itwinui-react/node_modules/react-table
npm warn   react-table@"^7.8.0" from @itwin/itwinui-react@3.17.0
npm warn   node_modules/@itwin/itwinui-react
npm warn
npm warn Conflicting peer dependency: react@18.3.1
npm warn node_modules/react
npm warn   peer react@"^16.8.3 || ^17.0.0-0 || ^18.0.0" from react-table@7.8.0
npm warn   node_modules/@itwin/itwinui-react/node_modules/react-table
npm warn     react-table@"^7.8.0" from @itwin/itwinui-react@3.17.0
npm warn     node_modules/@itwin/itwinui-react
```

PR #1901, opened 2024-03-06
fix Table `columnOrder` resetting incorrectly

### LINKED ISSUE #1899: Table initial state not setting column order properly.

When setting up the columnOrder prop in table's initial state it's being ignored, and table is rendering columns in the order of the columns array.

### Expected Behavior

Table render columns in the order provided to initial state.

### Link to minimal repro

https://stackblitz.com/edit/github-y5zcvc?file=src%2FDraggableColumns.tsx

### Steps To Reproduce

_No response_

### Anything else?

```"@itwin/itwinui-react": "2.12.24"```

### PR DESCRIPTION
**Pull request by mayank99**
## Changes

Fixes #1899.

The root cause was that the initial value of `lastPassedColumns` is `[]` so it was always resetting the columns. `[]` is never valid so I've added a simple check for it.

## Testing

Tested in playground (using [linked sandbox](https://stackblitz.com/edit/github-y5zcvc?file=src%2FDraggableColumns.tsx)) and added unit test.

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -717,7 +717,10 @@ export const Table = <
   // This is to avoid the old columnOrder from affecting the new columns' columnOrder
   React.useEffect(() => {
     // Check if columns have changed (by value)
-    if (JSON.stringify(lastPassedColumns.current) !== JSON.stringify(columns)) {
+    if (
+      lastPassedColumns.current.length > 0 &&
+      JSON.stringify(lastPassedColumns.current) !== JSON.stringify(columns)
+    ) {
       instance.setColumnOrder([]);
     }
     lastPassedColumns.current = columns;
--- a/.changeset/sharp-gorillas-search.md
+++ b/.changeset/sharp-gorillas-search.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by rohan-kadkol — APPROVED** [review]

File: packages/itwinui-react/src/core/Table/Table.tsx [hunk 1]
```diff
@@ -717,7 +717,10 @@ export const Table = <
...
   React.useEffect(() => {
     // Check if columns have changed (by value)
-    if (JSON.stringify(lastPassedColumns.current) !== JSON.stringify(columns)) {
+    if (
+      lastPassedColumns.current.length > 0 &&
+      JSON.stringify(lastPassedColumns.current) !== JSON.stringify(columns)
+    ) {
       instance.setColumnOrder([]);
     }
     lastPassedColumns.current = columns;
```
mayank99, line 726: Some comments for future:

1. We should try to avoid stringifying columns like this. It can be expensive and error-prone. In our _new_ Table we could require always passing an id, or find a way to compare the internal ids.
 
2. We should maybe create a hook for this "previous state" pattern.
   ```jsx
   function usePreviousState<T>(state: T, onChange: () => void): React.Ref<T>;
   ```

rohan-kadkol: Sounds good, linked this comment in https://github.com/iTwin/iTwinUI/issues/1144.

PR #1877, opened 2024-02-23
Fix `InputGroup.message=<StatusMessage>` rendering two icons

rohan-kadkol:
## Changes

After-PR TODO from #1753 

Earlier, `message` was always wrapped in `<StatusMessage>`. This caused nested `<StatusMessage>`s when `message={<StatusMessage>}`. Nested StatusMessages could lead to double svgIcons:

<details>
<summary>Example code</summary>

```jsx
{/* ❌ Passing <StatusMessage startIcon={…} /> to the message of <InputGroup status="…" /> shows duplicate status icons */}
<InputGroup
  label='Checkbox group'
  status='positive'
  message={
    <StatusMessage startIcon={<SvgStar />} status='positive'>
      Help message
    </StatusMessage>
  }
>
  <Checkbox label='Select all' />
  <Checkbox label='Football' />
  <Checkbox label='Hockey' />
</InputGroup>
```

</details>

image

This PR now wraps `message` in `<StatusMessage>` only when `typeof message !== "string"`. This fixes the duplicate svgIcons issue.

image

As a result of this change, if users were passing `ReactNode`s that are not `<StatusMessage>` to `message`, they might need to manually wrap their `ReactNode` with `<StatusMessage>` for proper styling of `message`.

<details>
<summary>Example code</summary>

```jsx
<InputGroup
  label='some label'
  message={<div className='my-message'>Message</div>}
>
  <Radio />
  <Radio />
  <Radio />
</InputGroup>
```
</details>

|Previously|Now|
|-|-|
|image|image|

## Testing

* Updated unit test to confirm that `message` is not wrapped in `<StatusMessage>` when `typeof message !== 'string'`.
* Fixed failing tests

## Docs

Updated JSDocs of `message` to tell that `message` is automatically wrapped with `<StatusMessage>` when `typeof message === 'string'`. It also tells that users may need to wrap their non-string & non-StatusMessage `message` with `<StatusMessage>` for proper styling.

PR #1903, opened 2024-03-07
No more incorrect rest props in `InputGroup`'s DOM

rohan-kadkol:
## Changes

This PR fixes an issue that @mayank99 noticed: some variables were added to `InputGroup`'s DOM as attribute-value pairs.

This was happening because in #1877, I was no longer deconstructing the variables passed to the new helper component, `BottomMessage`. Thus, the non-deconstructed variables were included in `rest` and added to the DOM.

This PR fixes that by properly deconstructing the props that shouldn't be in the DOM and then passing them down to the helper component (`BottomMessage`).

## Testing

Confirmed manually that `BottomMessage` props are no longer showing up in the DOM.

I thought against having unit tests since this might be too specific. But I can add it if anyone feels differently.

<details>
<summary>App.tsx (vite-playground)</summary>

```jsx
import { SvgStar } from '@itwin/itwinui-icons-react';
import { Checkbox, InputGroup } from '@itwin/itwinui-react';

const App = () => {
  return (
    <>
      <InputGroup
        label='Checkbox group'
        message='Foobar'
        status='positive'
        svgIcon={<SvgStar />}
        displayStyle='default'
        messageProps={{
          contentProps: {
            color: 'hotpink',
          },
        }}
      >
        <Checkbox />
        <Checkbox />
      </InputGroup>
    </>
  );
};

export default App;

```

</details>

Result of the code:

|Old|New|
|-|-|
|image|image|

## Docs

No changeset since #1877 is not released yet. (https://github.com/iTwin/iTwinUI/pull/1903#discussion_r1516398299)

### DIFF
--- a/.changeset/hungry-hats-think.md
+++ b/.changeset/hungry-hats-think.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed an `InputGroup` bug where some variables were unnecessarily added as attribute-value pairs to the DOM.
--- a/.changeset/hungry-hats-think.md
+++ b/.changeset/hungry-hats-think.md
@@ -1,5 +1,0 @@
----
-"@itwin/itwinui-react": patch
----
-
-Fixed an `InputGroup` bug where some variables were unnecessarily added as attribute-value pairs to the DOM.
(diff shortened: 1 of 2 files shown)


.changeset/hungry-hats-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed an `InputGroup` bug where some variables were unnecessarily added as attribute-value pairs to the DOM.
```
mayank99, line 5: No need for changeset, #1877 isn't released yet

mayank99 (approved):

PR #1910, opened 2024-03-11
setup partial premajor releases for css

mayank99:
## Changes

After lots of iteration and experimentation, this PR makes the following changes:
- Ignores `@itwin/itwinui-css` from `pnpm changeset version`, so that its version does not get bumped.
- Makes it `private` to prevent accidental publish (until major release is ready).
- Removes it from dependencies of all workspaces, and adjusts turbo pipeline to manually depend on it.
  - This is necessary because `pnpm changeset` does not allow adding changesets to packages that depend on ignored workspaces.

As a result, the `main` branch can be used to add `major` changesets to `@itwin/itwinui-css`. When releasing, it will be ignored without blocking the normal release of other packages.

Only difference to regular workflow is that itwinui-css changesets now always need to be standalone, rather than being grouped with itwinui-react.

## Testing

Tested creating changesets for `@itwin/itwinui-css` and `@itwin/itwinui-react`, then running `pnpm changeset version --ignore @itwin/itwinui-css` to ensure that only react package gets versioned. The css changesets do not get deleted.

## Docs

N/A

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -11,16 +11,20 @@
       "outputs": ["css/**"]
     },
     "@itwin/itwinui-react#build": {
-      "dependsOn": [
-        "^build",
-        "@itwin/itwinui-css#build",
-        "@itwin/itwinui-variables#build"
-      ],
+      "dependsOn": ["^build", "@itwin/itwinui-css#build"],
       "outputs": ["esm/**", "cjs/**"]
+    },
+    "css-workshop#build": {
+      "dependsOn": ["@itwin/itwinui-css#build"],
+      "outputs": ["dist/**", "build/**"]
     },
     "react-workshop#build": {
       "dependsOn": ["^build"],
       "outputs": ["build/**"]
+    },
+    "portal": {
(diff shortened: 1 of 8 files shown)


mayank99 (commented):
This seems to be failing CI. Investigating.

rohan-kadkol (approved):

turbo.json
```diff
@@ -11,21 +11,28 @@
...
     },
+    "css-workshop#build": {
+      "dependsOn": ["@itwin/itwinui-css#build"],
+      "outputs": ["dist/**", "build/**"]
+    },
     "react-workshop#build": {
       "dependsOn": ["^build"],
       "outputs": ["build/**"]
     },
+    "portal": {
```
Ben-Pusey-Bentley, line 25: Is this supposed to be `portal#build`? Because I have occasionally had some build failures after this PR was merged, and I think this might be the culprit.

mayank99: Whoops yes. For now you can manually run dependent build commands first.

Also would be useful if you review the PR next time 😅

PR #1911, opened 2024-03-12
Fix Broken Tabs Demo Styling

Ben-Pusey-Bentley:
## Changes

Fixes broken styling for the tabs demos on the documentation site. Does this by defining width for `Tabs.Wrapper` component with css.

## Testing

Tested the documentation site with local host and found that this style change fixes the issue.

## Docs

Updated demos with fixed styling.

### DIFF
--- a/examples/Tabs.actions.css
+++ b/examples/Tabs.actions.css
@@ -1,0 +1,3 @@
+.demo-container {
+  width: 100%;
+}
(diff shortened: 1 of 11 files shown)


mayank99 (approved):
While I'm worried that #1884 may have also broken tabs for some of our users, lets go with this for now. If we get any complaints, we can think about adding `inline-size: 100%` to our component CSS.

PR #1915, opened 2024-03-12
Remove the `iui-anchor` mixin.

Ben-Pusey-Bentley:
## Changes

Removes the following `iui-anchor` mixin from the css package.

The `iui-anchor` mixin has instead had it's styles all added to the `iui-anchor` class in the `itwinui-css` package. This is breaking change for all of the components that previously used this mixin in their styles, as they now need to have the `iui-anchor` class added onto them to be styled correctly. The `css-workshop` has been updated to reflect this change. Similarly, the `itwinui-react` package has been updated to use this new convention. There is still a breaking change to the custom elements for the `Footer` component, however, since those links need to have a class defined inside of them to have the correct styling. The `react-workshop` has also been updated to reflect this.

## After-PR TODO(s)
- [ ] Add support for `<button>` and `<a>` elements to both the `AlertAction` and `ToastPresentation` components, as both components should be able to have either functionality. (see [thread](https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1528845986))

## Testing

Ran tests with both the `css-workshop` and `react-workshop` to make sure that no unintended visual changes had been made.

## Docs

Added changeset for both `itwinui-css` and `itwinui-react`.


mayank99 (commented):
Can you split this up into two separate PRs for anchor and globals?

Ben-Pusey-Bentley:
> Can you split this up into two separate PRs for anchor and globals?

Okay, I'll make this one the PR for anchor then.

apps/react-workshop/src/Footer.stories.tsx
```diff
@@ -107,7 +107,12 @@ export const CustomContent = () => {
                   <Footer.Separator />
                   <Footer.Item>
                     {element.url ? (
-                      <a href={element.url} target='_blank' rel='noreferrer'>
+                      <a
+                        href={element.url}
+                        target='_blank'
+                        rel='noreferrer'
+                        className='_iui3-anchor'
```
mayank99, line 114: This is not acceptable. The class names are an implementation detail that should not be exposed to the user. (The "_" prefix indicates that it's considered private)

packages/itwinui-css/src/anchor/mixins.scss
```diff
@@ -2,30 +2,6 @@
...
-    @include iui-anchor-underline('on-initial');
-  }
-
-  &[data-iui-underline='true'],
-  :is([data-iui-contrast='high'] &) {
-    @include iui-anchor-underline('on-initial');
-  }
-}
-
 @mixin iui-anchor-status($status: accent) {
```
mayank99, line 5: These other mixins should also be removed. Everything should be controllable using plain CSS/HTML.

packages/itwinui-react/src/core/Footer/Footer.tsx
```diff
@@ -146,7 +147,15 @@ export const Footer = Object.assign(
                   {index > 0 && <FooterSeparator />}
                   <FooterItem>
                     {element.url ? (
-                      <a href={element.url} target='_blank' rel='noreferrer'>
+                      <a
+                        href={element.url}
+                        target='_blank'
+                        rel='noreferrer'
+                        className={'iui-anchor'
```
mayank99, line 154: You don't need to go through such great lengths to add the class here. Using the `Anchor` component will suffice.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -233,11 +233,12 @@ const BreadcrumbsItem = React.forwardRef((props, forwardedRef) => {
   const defaultAs = !!props.href ? 'a' : !!props.onClick ? 'button' : 'span';
   const children =
     defaultAs === 'button' ? <span>{childrenProp}</span> : childrenProp;
+  const anchorClassName = defaultAs === 'a' ? 'iui-anchor' : '';
```
mayank99, line 236: `Anchor` component could be used here as well

.changeset/green-cheetahs-rush.md
mayank99, line 1: Ideally this change should be transparent to react users.

apps/react-workshop/src/Footer.stories.tsx
Ben-Pusey-Bentley: Okay, then I'll replace the `<a>` here with `Anchor`.

packages/itwinui-react/src/core/Footer/Footer.tsx
Ben-Pusey-Bentley: Will do.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
Ben-Pusey-Bentley: Will do.

.changeset/green-cheetahs-rush.md
Ben-Pusey-Bentley: Ideally, but anyone who is using an `<a>` in a `Footer` custom element will have to replace that with an `Anchor` with how the changes are currently implemented. Could add the styles from the `iui-anchor` class to the `> a` selector for the `iui-legal-footer-item` class, but reusing styles like that feels like it slightly defeats the purpose of removing the mixin.

.changeset/green-cheetahs-rush.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/itwinui-react": minor
+"@itwin/itwinui-css": minor
```
rohan-kadkol, line 3: After #1910, `itwinui-css` would need a separate changeset.

packages/itwinui-css/src/alert/alert.scss
```diff
@@ -54,11 +54,13 @@
...
-  @include anchor.iui-anchor;
-  @include anchor.iui-anchor-underline('on-initial');
   border-radius: var(--iui-border-radius-1);
   cursor: pointer;
   margin-inline-start: var(--iui-size-xs);
   user-select: none;
   white-space: nowrap;
 }
+
+.iui-anchor.iui-alert-link {
```
rohan-kadkol, line 64: Is `.iui-anchor` needed in combination with `.iui-alert-link`? Or is `.iui-alert-link` sufficient?

packages/itwinui-react/src/core/Alert/Alert.tsx
```diff
@@ -136,7 +136,7 @@ const AlertAction = React.forwardRef((props, ref) => {
   return (
     <ButtonBase
       as={(!!props.href ? 'a' : 'button') as 'a'}
-      className={cx('iui-alert-link', className)}
+      className={cx('iui-anchor iui-alert-link', className)}
```
rohan-kadkol, line 139: Should `.iui-anchor` only be applied when the `as` is an `<a>`?

packages/itwinui-css/src/alert/alert.scss
Ben-Pusey-Bentley: It is needed in combination here because using just `.iui-alert-link` causes the style to be overwritten by the style in `.iui-anchor`, so a higher specificity is required.

.changeset/green-cheetahs-rush.md
mayank99: Well we can't suddenly ask our users to use `<Anchor>` instead of `<a>`. That's a breaking change.

What we might be able to do though, is clone the node and modify it? It's not ideal, but backwards compatibility is important.

Ben-Pusey-Bentley: Okay, I'll take a look at cloning the node then.

packages/itwinui-react/src/core/Alert/Alert.tsx
Ben-Pusey-Bentley: The type on that `forwardRef` indicates that the `AlertAction` will always be an `<a>`, which is a little confusing with the `as` potentially being a `<button>`. I will investigate this and see if there are any issues with only applying the `.iui-anchor` class when `as` is `<a>`.

.changeset/green-cheetahs-rush.md
Ben-Pusey-Bentley: Added some code to the `<Footer>` that replaces the `<a>` added to custom content with `<Anchor>` in the latest [commit](https://github.com/iTwin/iTwinUI/pull/1915/commits/45c6861c0477f194b1f2e3955493ea8bdf78321b). Seems like it might be fragile in its current state.

apps/react-workshop/src/Footer.stories.tsx
Ben-Pusey-Bentley: Removed that `className` in the latest [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/45c6861c0477f194b1f2e3955493ea8bdf78321b)

.changeset/green-cheetahs-rush.md
Ben-Pusey-Bentley: Removed the `itwinui-react` from the changeset in this [commit](https://github.com/iTwin/iTwinUI/pull/1915/commits/45c6861c0477f194b1f2e3955493ea8bdf78321b), since the code has been updated to remove breaking changes for react users. [Related Comment](https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1523450673)

.changeset/green-cheetahs-rush.md
mayank99: Yeah it looks pretty bad. I'd say lets just add some CSS for `.iui-legal-footer-item > a`. It should only be a few lines (we don't need everything from `iui-anchor`).

Ben-Pusey-Bentley: Alright, will do.

packages/itwinui-css/src/alert/alert.scss
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/a6144abc9c3d1d69d0640a285709d1d72f47bd5b)

packages/itwinui-css/src/anchor/mixins.scss
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/63c95e30856b494b4444148e0b90542656c6ddb9)

.changeset/green-cheetahs-rush.md
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/1915/commits/45c6861c0477f194b1f2e3955493ea8bdf78321b) and high contrast fixed in [latest.](https://github.com/iTwin/iTwinUI/pull/1915/commits/63c95e30856b494b4444148e0b90542656c6ddb9)

packages/itwinui-css/src/alert/alert.scss
Ben-Pusey-Bentley: Sorry, ignore the last comment here. Accidently put that comment on the wrong conversation.

rohan-kadkol: To confirm, is that commit link correct? Because that commit has no changes for `alert.scss`.

Ben-Pusey-Bentley: Again, please ignore that comment. It was meant for this [conversation.](https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1523450673)

rohan-kadkol: Oops, I missed reading https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1526290640 😅

Ben-Pusey-Bentley: All good. It was my mistake to begin with.

packages/itwinui-css/src/anchor/mixins.scss
mayank99: For `iui-alert-link` and `iui-toast-anchor`, you can use the `underline` prop in React and the `data-iui-underline` attribute in HTML, instead of repeating the underline CSS.

Also you can add new data attribute selector to `.iui-anchor` to reduce the repetition around status colors. The HTML would then become something like `<a class="iui-anchor iui-alert-link" data-iui-status="positive">`.

Ben-Pusey-Bentley: Okay, will do.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/ecdd7cf46b7f29bc1f86b77f40b7f23364723ce5)

mayank99: You don't need to branch on the return value. I'm just suggesting this one line change:

```diff
- const defaultAs = !!props.href ? 'a' : !!props.onClick ? 'button' : 'span';
+ const defaultAs = !!props.href ? Anchor : !!props.onClick ? 'button' : 'span';
```

packages/itwinui-react/src/core/Footer/Footer.test.tsx
```diff
@@ -35,7 +35,9 @@ const assertFooterItems = (
     }
 
     const dataIndex = Math.floor(i / 2);
-    expect(element.textContent).toBe(data[dataIndex].title);
+    expect(element.textContent?.replace(' (opens in new tab)', '')).toBe(
+      data[dataIndex].title,
+    );
```
mayank99, line 40: This is testing implementation details and is therefore brittle. You can instead do a partial match using `toHaveTextContent`:

```js
expect(element).toHaveTextContent(data[dataIndex].title);
```

Ben-Pusey-Bentley: Alright, will do.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
Ben-Pusey-Bentley: I tried that as a string and it didn't work for me. Guess I didn't realize that it had to be a type if it is a React component. Will fix.

AdamMeza-Bentley (approved):

packages/itwinui-css/src/anchor/mixins.scss
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/882a3b0dc56a27a8668d6b79dff78bdedeb4b9fd) Changed the `<ButtonBase>` to `<Anchor>` in the `Alert` component to use the `underline` prop, but couldn't do so with `Toast`, since the `link` object that gets spread across the button there takes in `HTMLButtonProps`. Instead, passed in `data-iui-underline` to the button. Could also add an `underline` button to prop, but I decided against it since we are trying to limit changes to react users.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/4f086a8b25654918d9a640e46f5cc1ee1ba9735f)

packages/itwinui-css/src/alert/alert.scss
Ben-Pusey-Bentley: The style ended up being completely removed in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/882a3b0dc56a27a8668d6b79dff78bdedeb4b9fd#diff-f4d2f744678b91bc8388e5cfed9d9326fa6f7061ae50ae71d5b542a8b09c3389) It was removed because of this [conversation.](https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1526311053)

packages/itwinui-react/src/core/Alert/Alert.tsx
Ben-Pusey-Bentley: I can confirm that the visual tests expect the styles that we provide with the `.iui-anchor` class, even when the element is a `<button>`. It therefore makes sense to apply the class to the `AlertAction`, regardless of the `as` parameter.

packages/itwinui-css/src/anchor/anchor.scss
```diff
@@ -1,9 +1,61 @@
...
+      &:hover {
+        color: var(--iui-color-text-#{$statusColor}-hover);
+      }
+    }
+  }
+
+  border-radius: var(--iui-border-radius-1);
+  cursor: pointer;
+
+  text-decoration: none;
```
mayank99, line 37: Nit: could you reorder this file so that these loose declarations (`color`, `cursor`, etc) are at the top of the class, and the `:hover` declarations are grouped, and the status colors are at the bottom?

Ben-Pusey-Bentley: Sure, will do.

packages/itwinui-react/src/core/Alert/Alert.tsx
```diff
@@ -132,16 +132,19 @@ AlertMessage.displayName = 'Alert.Message';
...
 const AlertAction = React.forwardRef((props, ref) => {
   const { children, className, ...rest } = props;
+  const { type } = useSafeContext(AlertContext);
 
   return (
-    <ButtonBase
+    <Anchor
       as={(!!props.href ? 'a' : 'button') as 'a'}
-      className={cx('iui-alert-link', className)}
+      className={cx('iui-button-base iui-alert-link', className)}
```
mayank99, line 140: nit:

```suggestion
      className={cx('iui-button-base', 'iui-alert-link', className)}
```

packages/itwinui-react/src/core/Alert/Alert.tsx
```diff
@@ -132,16 +132,19 @@ AlertMessage.displayName = 'Alert.Message';
 
 const AlertAction = React.forwardRef((props, ref) => {
   const { children, className, ...rest } = props;
+  const { type } = useSafeContext(AlertContext);
 
   return (
-    <ButtonBase
+    <Anchor
```
mayank99, line 138: By making this change we lose out on the useful behaviors provided by `ButtonBase`.

I'm thinking we should use `Buttonbase` inside `Anchor` too. Or maybe we need a generic component called `Action` (or `ButtonOrLink`) 🤔

(Not sure of the best solution yet, and doesn't need to be resolved in this PR)

packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -285,7 +285,13 @@ export const ToastPresentation = React.forwardRef((props, forwardedRef) => {
         {content}
       </Box>
       {link && (
-        <ButtonBase className='iui-toast-anchor' {...link} title={undefined}>
+        <ButtonBase
+          className='iui-anchor iui-toast-anchor'
```
mayank99, line 289: This should use `cx` for merging class names.

Also it should take into account `link.className`?

```jsx
<ButtonBase
  {...link}
  className={cx('iui-anchor', 'iui-toast-anchor', link.className)}
```

packages/itwinui-react/src/core/Alert/Alert.tsx
Ben-Pusey-Bentley: I can test out using `ButtonBase` inside of the `Anchor`, and see if there are any issues with that.

packages/itwinui-react/src/core/Toast/Toast.tsx
Ben-Pusey-Bentley: Alright, will do.

packages/itwinui-react/src/core/Footer/Footer.test.tsx
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/34d0dad5d00a597dd17d66d6706c4f98dd71c73f)

packages/itwinui-css/src/anchor/anchor.scss
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/aae66014d51e43dc6228a8bb087f946daf705f44)

packages/itwinui-react/src/core/Toast/Toast.tsx
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/bf9f560bd5cdba19aa3422624f98b7086412affa)

.changeset/green-cheetahs-rush.md
mayank99, line 1: Changeset can be reworded and split into two:
- One changeset (major) should clear mention the breaking HTML change (needing to pass `iui-anchor` explicitly)
- The other changeset (minor) can talk about the new status attribute support in `iui-anchor`

Ben-Pusey-Bentley: Will do.

Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/b2d784d01ab0a3cb57ccc6b0c9e89dc95c2e3f52) They may need some more rewording, as they both feel a little long in their current state.

mayank99: If you feel they need rewording, shouldn't you try to improve them before pushing? 

For a good example you can look at this: https://github.com/iTwin/iTwinUI/blob/rohan/table-header-scrollable-on-drag/.changeset/curvy-pugs-agree.md

Long isn't necessarily a problem, but it needs to be clear and easy to parse, without being overly wordy.

Ben-Pusey-Bentley: Made updates to both changesets in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/584dbdd0c7392603b8496ae89486eda4bf3add1f)

mayank99: It still feels overly wordy and unclear.

Here, I rewrote them to be make them more clear:

major:
> `iui-anchor` must now be explicitly added to `<a>` elements, in addition to the previous `iui-alert-link`, `iui-breadcrumbs-content`, `iui-tag-basic`, and `iui-toast-anchor` classes.
> 
> ```diff
> - <a class="iui-alert-link">…</a>
> + <a class="iui-alert-link iui-anchor">…</a>
> ```
> ```diff
> - <a class="iui-breadcrumbs-content">…</a>
> + <a class="iui-breadcrumbs-content iui-anchor">…</a>
> ```
> ```diff
> - <a class="iui-tag-basic">…</a>
> + <a class="iui-tag-basic iui-anchor">…</a>
> ```
> ```diff
> - <a class="iui-toast-anchor">…</a>
> + <a class="iui-toast-anchor iui-anchor">…</a>
> ```
> 
> Additionally, all `<a>` elements within `iui-legal-footer` elements will need the `iui-anchor` class.
> 
> ```diff
>   <li class="iui-legal-footer-item">
> -  <a>…</a>
> +  <a class="iui-anchor">…</a>
>   </li>
> ```

minor:

> `iui-anchor` can now be styled with status colors using a new `data-iui-status` attribute.
> 
> ```html
> <a class="iui-anchor" data-iui-status="negative">…</a>
> ```
> 
> Possible values include: `"positive"`, `"negative"` and `"informational".

Ben-Pusey-Bentley: Okay, thank you. I will update the changesets.

Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/91e9232447652bc3aff673e4ba80d8977bf45bc6) Also added the `warning` value to the possible values for `data-iui-status`.

mayank99 (approved):
LGTM. I would also wait for approval from @rohan-kadkol

.changeset/green-cheetahs-rush.md
```diff
@@ -0,0 +1,11 @@
+---
+"@itwin/itwinui-css": minor
+---
+
+`iui-anchor` can now be styled with status colors using a new `data-iui-status` attribute.
```
rohan-kadkol, line 5: To confirm, are we intending to use the status only when the anchor is used inside another component? Or are we intending it to be used on standalone anchors too?

If we don't intend this to be used on standalone anchors, maybe we should mention this in the changeset?

Also, in general, should we mention something regarding the intended use-cases of this to prevent overruse of anchors with statuses?

packages/itwinui-css/src/footer/footer.scss
```diff
@@ -32,12 +31,27 @@
...
 
   > a {
-    @include anchor.iui-anchor;
-
     &,
     &:hover {
       color: var(--iui-color-text-muted);
     }
+
+    &:not(:hover) {
```
rohan-kadkol, line 39: Is this the same as `&`? Maybe you did this for some specificity conflict?

apps/css-workshop/pages/tag.html
```diff
@@ -733,38 +733,38 @@ <h2>Basic tag containers</h2>
       <div class="iui-tag-container">
         <a
           href="#"
-          class="iui-tag-basic"
+          class="iui-anchor iui-tag-basic"
```
rohan-kadkol, line 736: nit: Looks like there's some extra leading and trailing space.

image

image

I'm wondering how these weren't pointed out by the visual tests?

packages/itwinui-css/src/tag/tag.scss
```diff
@@ -21,7 +20,6 @@
 }
 
 a.iui-tag-basic {
-  @include anchor.iui-anchor;
   @include mixins.iui-focus($offset: 1px);
```
rohan-kadkol, line 23: Is this needed? Or just letting it (`@include mixins.iui-focus(…)`) come from the `.iui-anchor` is sufficient?

packages/itwinui-react/src/core/Alert/Alert.tsx
rohan-kadkol: Not sure why did you change the `ButtonBase` to `Anchor` here. Would just adding `.iui-anchor` along with the necessary data attributes like how you did for `Toast` not be sufficient?

image

apps/css-workshop/pages/tag.html
rohan-kadkol: Actually, this can be be ignored since it's likely an issue on my side. Because [this PR's deployed version's `tag.html`](https://itwin.github.io/iTwinUI/1915/css/tag.html) shows up correctly.

packages/itwinui-css/src/tag/tag.scss
Ben-Pusey-Bentley: It may not be. Let me check.

packages/itwinui-react/src/core/Alert/Alert.tsx
mayank99: `Anchor` is the correct way to do this. `Toast` should do the same

Ben-Pusey-Bentley: The primary reason for making the change was to give access to the `underline` prop, which currently only exists on `Anchor`. The reason that `Toast` doesn't do the same is because the `Toast` anchor takes in a `link` prop, which has `HTMLButtonProps` as a part of it's type. Changing that would introduce a breaking change for the react package.

packages/itwinui-css/src/tag/tag.scss
Ben-Pusey-Bentley: Doesn't seem like it was. Removed in this [commit.](https://github.com/iTwin/iTwinUI/pull/1915/commits/73f14a143597afbe40d9972bdab2586317536b48)

packages/itwinui-css/src/footer/footer.scss
Ben-Pusey-Bentley: `&` and `&:not(:hover)` are not the same. `&` is a rule that is always applied, while `&:not(:hover)` only applies it's styles when the element is not being hovered over. That change would create a conflict that does resolve correctly in this case, since the `:hover` style from `iui-anchor` has a higher specificity. However, I do not think it is wise to rely on the `:hover` rule winning the specificity battle for the footer links to be styled correctly.

.changeset/green-cheetahs-rush.md
Ben-Pusey-Bentley: The primary use case for this attribute is inside of the `Alert` and `Toast` components, but I don't think that means that developers should be cautioned from using this attribute with standalone anchors, if they feel like it. The attribute only changes the anchor's color, so I don't think overuse is much of a concern either. Mentioning the primary use case might make it more clear why we are adding this attribute, but I don't want to accidently make the changeset unclear as a result.

packages/itwinui-css/src/footer/footer.scss
rohan-kadkol: > `&` and `&:not(:hover)` are not the same.

True, but I meant more like selecting something when it's not hovered (`&:not(:hover)`) would likely be the same as selecting that something directly without any pseudo classes (`&`), unless we *really* want to select it only when it's not hovered.

Upon rethinking it, are all those `text-decoration` styles required here ([code](https://github.com/iTwin/iTwinUI/pull/1915/files#diff-1cd601586d11db817d96d51d666331f1ea862f34ce736612b6a16204ed128583R39-R54))? Because I think it they all also from the anchor?

.changeset/green-cheetahs-rush.md
mayank99: This changeset is for the CSS package, which is really only used by us. We are not adding this as a prop in React, so there is no chance of misuse.

You can compare the download numbers of the recent versions:
- css: https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions
- react: https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions

packages/itwinui-css/src/footer/footer.scss
mayank99: > > `&` and `&:not(:hover)` are not the same.
> 
> True, but I meant more like selecting something when it's not hovered (`&:not(:hover)`) would likely be the same as selecting that something directly without any pseudo classes (`&`), unless we _really_ want to select it only when it's not hovered.


`&:not(:hover)` was suggested by me. It is overriding the default user-agent styles of `<a>` element (which comes with an underline)

> Upon rethinking it, are all those `text-decoration` styles required here ([code](https://github.com/iTwin/iTwinUI/pull/1915/files#diff-1cd601586d11db817d96d51d666331f1ea862f34ce736612b6a16204ed128583R39-R54))? Because I think it they all also from the anchor?

`Anchor` may not always be used, like in the Custom Content Footer story

packages/itwinui-react/src/core/Alert/Alert.tsx
mayank99: Yeah lets handle this in a separate PR then. I think both Alert and Toast should correctly support both `<button>` and `<a>` because there are valid uses for both.

You can add an after-PR TODO in the PR description

packages/itwinui-css/src/footer/footer.scss
Ben-Pusey-Bentley: These styles are specifically being included because the `Footer` component allows for custom content. Removing these styles would break the styling for any custom content that does not use the `iui-anchor` class, which would introduce a breaking change to the react components for any users using custom content. As a result, it was decided to keep the styles here. Relevant conversation: [here](https://github.com/iTwin/iTwinUI/pull/1915#discussion_r1525481818)

packages/itwinui-react/src/core/Alert/Alert.tsx
Ben-Pusey-Bentley: Added a TODO to the PR description.

packages/itwinui-css/src/footer/footer.scss
rohan-kadkol: Ohh, I see, I now understand why we need those styles.

In high contrast, the underline even shows up on hover. Is this expected? ([demo](https://itwin.github.io/iTwinUI/1915/react/?arg-high-contrast=true&story=footer--custom-content&theme=dark))

https://github.com/iTwin/iTwinUI/assets/45748283/adeda38d-5ae2-438f-9127-42ccd174b39d

mayank99: I think that's ok. In a future PR, I would like to make all anchors have underlines on hover, even in high-contrast theme.

rohan-kadkol (approved):

PR #1917, opened 2024-03-13
Removes unnecessary global mixins.

Ben-Pusey-Bentley:
## Changes

Removes four sass mixins from the global mixins file in the css package. These mixins are:

- `iui-ripple`
- `iui-scroll-snapping`
- `pseudo-content`
- `iui-scrollbar`

In the cases where these mixins were used, their styles were placed instead. Therefore, there should be no changes to end users from this PR.

## Testing

Ran css and react visual tests to make sure there were no visual changes from the removal of these mixins.

## Docs

N/A

### DIFF
--- a/packages/itwinui-css/src/global.scss
+++ b/packages/itwinui-css/src/global.scss
@@ -23,7 +23,27 @@
 // box-sizing, text selection and scrollbar styling on iui-root and all elements under it
 :where(.iui-root, .iui-root *) {
   @include mixins.iui-text-selection;
-  @include mixins.iui-scrollbar;
+  scrollbar-color: hsl(var(--iui-color-foreground-hsl) / var(--iui-opacity-4)) transparent;
+  scrollbar-width: thin;
+
+  &::-webkit-scrollbar {
+    max-inline-size: var(--iui-size-xs);
+    max-block-size: var(--iui-size-xs);
+  }
+
+  &::-webkit-scrollbar-thumb {
+    background-color: hsl(var(--iui-color-foreground-hsl) / var(--iui-opacity-4));
+    border-radius: var(--iui-border-radius-round);
+
+    &:hover {
+      background-color: hsl(var(--iui-color-foreground-hsl) / var(--iui-opacity-3));
+    }
+  }
+
+  &::-webkit-scrollbar-track,
+  &::-webkit-scrollbar-corner {
+    background-color: transparent;
+  }
 
   &,
   &::before,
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/src/global.scss
```diff
@@ -23,7 +23,27 @@
 // box-sizing, text selection and scrollbar styling on iui-root and all elements under it
 :where(.iui-root, .iui-root *) {
   @include mixins.iui-text-selection;
-  @include mixins.iui-scrollbar;
+  scrollbar-color: hsl(var(--iui-color-foreground-hsl) / var(--iui-opacity-4)) transparent;
```
mayank99, line 26: I'd say keep the scrollbar stuff in a local mixin at the bottom at the file. It should not be used anywhere else in the repo, but it will help with readability.

packages/itwinui-css/src/table/base.scss
```diff
@@ -354,7 +354,13 @@
   flex-grow: 1;
   // Allows columns with static width to take more space than the table width
   align-items: flex-start;
-  @include mixins.iui-scroll-snapping('.iui-table-row');
+  #{if(&, '&.iui-scroll-snapping', '.iui-scroll-snapping')} {
```
mayank99, line 357: This doesn't need branching, since it's pretty clear `&` will always exist

packages/itwinui-css/src/global.scss
Ben-Pusey-Bentley: Okay, will do.

packages/itwinui-css/src/table/base.scss
Ben-Pusey-Bentley: Alright, I'll remove the branching then.

packages/itwinui-css/src/global.scss
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1917/commits/621dd3433bcf61c53e8150c0d131abe3522c334d)

packages/itwinui-css/src/table/base.scss
Ben-Pusey-Bentley: Done in latest [commit.](https://github.com/iTwin/iTwinUI/pull/1917/commits/7702d98d06e76435764cec075b67cf88ca70dd0c)

mayank99 (approved):

PR #1891, opened 2024-03-01
Table header scrollable upon dragging column

rohan-kadkol:
## Changes

Fixes #993

In the implementation of the table thus far, only the table body was made scrollable.  That's why the table header would not automatically scroll horizontally by the browser when hovering a dragged column at the left or the right edge. Additionally, if the user tried scrolling horizontally while their cursor was over the header, nothing would happen. Partially showing the header's overflow content but not allowing to scroll to it could be confusing for the user.

This PR solves these problems by moving the scroll container to the entire table instead of just the body.

|Old|New|
|-|-|
|image|image|

This approach helped simplify or remove some scroll hacks:
* The body's and header's scroll no longer need to be kept in sync using JS (this started with https://github.com/iTwin/iTwinUI/pull/417#issue-1081151472, https://github.com/iTwin/iTwinUI-react/pull/495)
* No need of the dummy div in the shadow dom to make the body have the same width as the header when the body had no rows (https://github.com/iTwin/iTwinUI/pull/1725)
  * So, this PR undid changes from #1725 and #1809

The last "loading row" in [Table → LazyLoading](https://itwin.github.io/iTwinUI/react/?story=table--lazy-loading&theme=dark) was no longer sticky since it was no longer a direct child of the scroll container (`.iui-table-body` is no longer the scroll container). Thus, I added a new class called `.iui-table-body-extra` that places this loading row outside the `.iui-table-body`, and thus a direct child of `.iui-table` which is the new scroll container.

image

<details>
<summary>With/Without `.iui-table-body-extra`</summary>

Without `.iui-table-body-extra`, the loading row is still in `.iui-table-body`:

https://github.com/iTwin/iTwinUI/assets/45748283/8c181c4c-4c70-4556-8a93-e837ea0d790f

With `.iui-table-body-extra`, the loading row is not in `.iui-table-body`, but is instead a sibling after `.iui-table-body`:

https://github.com/iTwin/iTwinUI/assets/45748283/130bc007-e0af-4539-8bdd-9a712996b572

</details>

---

<details>
<summary>TODOs for self</summary>

- [x] Delete all test image changes, run visual tests, approve only the correct ones
- [x] Fix case when no rows in a table that has horizontal scroll
- ~[ ]~ Add a comment about "Localized" story test failure (outdated)
- [x] Update the `<script>` part of table.html
- [x] Verify all cases (stories and css demos)
- [x] Confirm if there are any unnecessary CSS styles

</details>

## Testing

* Added a unit test
[...]

PR #1919, opened 2024-03-13
Add `.iui-field` base class for inputs

FlyersPh9:
## Changes

- Add `.iui-field` class which gives base styling (`background-color`, `color`, `border-color`, `data-iui-variant`, etc) for many inputs (`<button>`, `<input>`, `<textarea>`, `<select>`).  This reduces duplicate CSS and helps keep styling consistent.
- `<input>`, `<textarea>`, `<select>` now have matching hover state with `<button>`.
- `<input>`, `<textarea>`, `<select>` text styling is different:
  - Disabled text is grayed out rather than not.
  - Disabled placeholder text no longer appears, otherwise it matches too closely to disabled text.

## Testing

- CSS test images.
- React test images.

## Docs

- Changesets

## After PR TODOs

- [ ] When an input is disabled, don't show placeholder text because it is too similar to disabled value text color.


FlyersPh9:
I continued working on this yesterday and I'm up to 22 files changed, starting to be concerned that the scope has creeped too far:
- Button mixins have been removed so all files other than `button.scss` have been deleted.
- The removal of these mixins caused having to make changes in breadcrumbs, pagination, etc. because they used these mixins.

mayank99:
I think it makes sense to keep the PRs separate, to make the review easier.
1. This PR changes the class names. (CSS output)
2. One PR removes the mixins from buttons. (internal Sass refactor)

You can even chain one PR to the other one, by creating a branch on top of it and changing the base. (Read more: [1](https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/), [2](https://graphite.dev/blog/stacked-prs#stacking-prs-with-github))

apps/css-workshop/src/pages/breadcrumbs.astro
```diff
@@ -167,19 +196,21 @@ import Layout from './_layout.astro';
...
       <label class='iui-input-grid iui-inline-label iui-inline-icon'>
-        <input
-          value='C:/Root/Lorem/Ipsum/Nested/Double nested folder with truncation/You are here/'
-          class='iui-input'
-          spellcheck='false'
-        />
-        <button class='iui-button' data-iui-variant='borderless' aria-label='Close'>
-          <svg-close-small class='iui-button-icon' aria-hidden='true'></svg-close-small>
-        </button>
+        <div class='iui-input-flex-container'>
```
FlyersPh9, line 204: Unrelated to this PR, but fixes a broken test.

apps/css-workshop/src/pages/file-upload.astro
```diff
@@ -135,9 +135,10 @@ import Layout from './_layout.astro';
                 Modified by
                 <div class='iui-table-header-actions-container'>
                   <button
-                    class='iui-button iui-table-filter-button iui-active'
+                    class='iui-button iui-field iui-table-filter-button'
                     data-iui-variant='borderless'
                     data-iui-size='small'
+                    data-iui-active='true'
```
FlyersPh9, line 141: Unrelated to this PR, but fixes a button issue.

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9, line 1: The plan was to have 1 centralized place for `.iui-field` styling… but the way `input-container.scss` is setup, we may need to have it duplicated here.  Perhaps we could make the CSS vars a mixin and reference them here?  I don't know...

FlyersPh9:
Feeling a little uncertain about this PR in some places.  There were large gaps between when I was working on things, so I was forgetting where I left off.  I a good chunk of yesterday reviewing my own PR and removing commented out code and TODO notes.  If I left any in there, please let me know and I apologize in advance.

packages/itwinui-css/src/input/input.scss
```diff
@@ -3,51 +3,32 @@
...
-    --_iui-input-background-color: var(--iui-color-background-accent-muted);
-    --_iui-input-border-color: var(--iui-color-border-accent);
-    --_iui-input-text-color: var(--iui-color-text-accent);
-  }
-
-  &:hover {
-    --_iui-input-border-color: var(--iui-color-border-foreground-hover);
+  &:where([disabled], :disabled, [aria-disabled='true']) {
+    &::placeholder {
+      --_iui-field-color-text-placeholder: transparent;
```
FlyersPh9, line 24: I think will warrant a discussion.

Before:
- Disabled with value = text color
- Disabled with placeholder = placeholder text color

After:
- Disabled with value = disabled text color
- Disabled with placeholder = transparent, not shown

This was done because it differed from the disabled styling on buttons.  If we were to revert this change, we should maybe consider making changes to `.iui-field` to not handle disabled state or at least handle less.

packages/itwinui-css/src/button/button.scss
```diff
@@ -33,42 +33,35 @@
...
-    @include variant.iui-button-variant(high-visibility);
-  }
+  @include variant.iui-button-variant;
 
-  &[data-iui-variant='cta'] {
-    @include variant.iui-button-variant(cta);
-  }
-
-  &[data-iui-variant='idea'] {
+  &[data-iui-type='feedback'] {
```
Ben-Pusey-Bentley, line 46: I am wondering if this would make more sense to keep as a value on `data-iui-variant` instead. After all, the PR says that the ideas button is going to need to have a specific variant value anyways in order to be colored correctly. If both `high-visibility` and `cta` ideas buttons are meant to be seen as correctly colored, then that would make this change make sense. However, our `css-workshop` page for buttons only shows `high-visibility` coloring. Adding a new example with alternate coloring could make that intention more clear.

FlyersPh9: If we keep the ideas button as a `data-iui-variant` that would mean the CSS for the coloring would need to be duplicated.

If we go this new route, the CSS is less but there is more HTML to define both type and variant.  I don't have anyone requesting a green ideas button, but this new method does allow it if the need arrises.  I think I would lean towards the new method due to less duplicate CSS and more flexibility, but would like to hear what others think as well.

packages/itwinui-css/src/global.scss
```diff
@@ -61,3 +61,79 @@
 }
 
 @include mixins.animations;
+
+:where(.iui-field) {
```
FlyersPh9, line 65: Not convinced this should be within `global.scss`.

packages/itwinui-css/src/button/button.scss
Ben-Pusey-Bentley: Okay, I think that I agree that reducing the duplicate CSS should be the main focus after thinking about it a bit more.

packages/itwinui-css/src/global.scss
Ben-Pusey-Bentley: I'm not sure where else it could go, since it is not related to a specific component. Maybe it would make sense to put it in `input.scss`?

packages/itwinui-css/src/global.scss
```diff
@@ -61,3 +61,79 @@
...
+  --_iui-field-color-background: var(--iui-color-background);
+  --_iui-field-color-background-hover: var(--iui-color-background-hover);
+  --_iui-field-color-background-disabled: var(--iui-color-background-disabled);
+  --_iui-field-color-border: var(--iui-color-border-foreground);
+  --_iui-field-color-border-hover: var(--iui-color-border-foreground-hover);
+  --_iui-field-color-border-disabled: var(--iui-color-border-disabled);
+  --_iui-field-color-focus-outline: var(--iui-color-border-accent);
+  --_iui-field-height: var(--iui-component-height);
+  --_iui-field-focus-outline-offset: -2px;
+  --_iui-field-focus-outline-width: 2px;
```
rohan-kadkol, line 78: Has the outline increased from 1px to 2px?

packages/itwinui-css/src/button/base.scss
```diff
@@ -3,36 +3,25 @@
...
-  transition:
-    background-color var(--iui-duration-1) ease-out,
-    border-color var(--iui-duration-1) ease-out;
-  -webkit-tap-highlight-color: transparent;
   text-decoration: none;
   @include iui-button-size;
-  @include mixins.iui-focus;
+
+  &:where([disabled], :disabled, [aria-disabled='true'], [data-iui-disabled='true'], .iui-disabled) {
+    cursor: not-allowed;
```
rohan-kadkol, line 23: I think this is already handled in `iui-field`?

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,7 +30,17 @@
   svg {
     @include mixins.iui-icon-style('m');
     transition: fill var(--iui-duration-1) ease-out;
-    fill: var(--_iui-button-icon-fill, currentColor);
+    fill: var(--_iui-button-icon-fill, var(--iui-color-icon));
```
rohan-kadkol, line 33: I believe we had decided that it would be safer to have the button text and the button icon match in color, and so we had preferred `currentColor` (https://github.com/iTwin/iTwinUI/pull/1787/files/33c0c084ad43990e3038576bb0002246cd70b17a#r1459075467).

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,7 +30,17 @@
...
+
+    button:hover & {
```
rohan-kadkol, line 35: Should this also be covering `:active`?

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -126,6 +118,10 @@
   block-size: fit-content;
   line-height: var(--iui-size-l);
   padding-inline: 1px; // To account for the left and right edges of the 1px border
+  transition:
+    background-color var(--iui-duration-1) ease-out,
+    border-color var(--iui-duration-1) ease-out,
+    color var(--iui-duration-1) ease-out;
```
rohan-kadkol, line 124: Does the transition for `border-color` already come from the `::before`?

packages/itwinui-css/src/input-container/input-container.scss
rohan-kadkol, line 1: Just wondering, is there any reason to not use `.iui-input-flex-container` in combination with `.iui-field`?

packages/itwinui-css/src/button/button.scss
```diff
@@ -33,42 +33,35 @@
   }
 
   &[data-iui-active='true'] {
-    @include default.iui-button-default-active;
+    @include active.iui-button-active;
+    @include active.iui-button-default-active;
 
-    &[data-iui-variant='borderless'] {
-      @include borderless.iui-button-borderless-active;
+    &:is([data-iui-variant='borderless']) {
```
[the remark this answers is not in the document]
FlyersPh9: I think the only benefit is it reduces the specificity, but tbh I did not check that it actually did so. 😅

packages/itwinui-css/src/global.scss
FlyersPh9: The default button has a `outline-width: 2px` currently.
Screenshot 2024-03-28 at 10 45 27 AM

| Variant | Outline width | Outline offset |
| Default | `2px` | `-2px` |
| Borderless | `1px` | `-1px` |
| High visibility | `1px` | `-3px` |

packages/itwinui-css/src/button/base.scss
FlyersPh9: Ahh right, I forgot about this.

This line overrides the cursor setting in `field.scss`:
https://github.com/iTwin/iTwinUI/blob/9ab9e8fbbae492bbbc621ee8f9848b5b47f55b5b/packages/itwinui-css/src/button/base.scss#L14

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,7 +30,17 @@
...
+      fill: var(--_iui-button-icon-fill-hover, var(--iui-color-icon-hover));
+    }
+
+    button[disabled] &,
+    button:disabled &,
+    button[aria-disabled='true'] {
```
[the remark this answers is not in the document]
FlyersPh9: I don't think we have any buttons that set disabled via `data-iui-disabled`, I think that's exclusive to the `<select>`?  I'll double check.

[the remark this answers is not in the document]
FlyersPh9: You are correct, fixed.

packages/itwinui-css/src/button/base.scss
FlyersPh9: Good catch, fixed!

.changeset/ninety-numbers-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
```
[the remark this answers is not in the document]
FlyersPh9: Updated to `major`.  I don't know what I was thinking when I put it as a patch. 😂

apps/css-workshop/src/pages/breadcrumbs.astro
```diff
@@ -167,19 +196,21 @@ import Layout from './_layout.astro';
   <h2>After clicking "you are here"</h2>
   <section id='demo-not-tested'>
     <nav class='iui-breadcrumbs' aria-label='Breadcrumbs'>
-      <button class='iui-button iui-button-dropdown' data-iui-variant='borderless'>
+      <button class='iui-button iui-field' data-iui-type='dropdown' data-iui-variant='borderless'>
```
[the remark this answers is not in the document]
FlyersPh9: Mentioned `data-iui-type` and the values in the changeset.

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,7 +30,17 @@
...
+      fill: var(--iui-color-icon-disabled);
```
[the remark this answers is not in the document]
FlyersPh9: You are correct, fixed.

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: I did want to use `.iui-field` because the goal of this PR was to centralize the styling in one place.  But `.iui-input-flex-container` sets the border using a pseudo element, so if I did set `.iui-field`, I would then need to unset the border, and re-set it on the pseudo.  Maybe that's ok?

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: You are correct, removed.

packages/itwinui-css/src/button/base.scss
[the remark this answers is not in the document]
FlyersPh9: I can't find any `<button>` with `data-iui-disabled` set.  Can you check if React supports that anywhere?  If no, I found some areas I can remove from the CSS.

apps/css-workshop/src/pages/breadcrumbs.astro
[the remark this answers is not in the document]
rohan-kadkol: Great, but not sure if it needs more explanation. I think having small diffs would be better too, similar to https://github.com/iTwin/iTwinUI/pull/1891#discussion_r1523704695, https://github.com/iTwin/iTwinUI/pull/1891/files#diff-b33a7bd0052338f2c9080c683825f0b756ad27ae745364ef08b2a1a0686677ec.

Also, noticed a small typo in the changeset: ~`iui-data-type`~ → `data-iui-type`.

packages/itwinui-css/src/button/base.scss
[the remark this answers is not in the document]
FlyersPh9: Ahh no wait, I remember why I did this now… I think it's 2 things:
1. A specificity issue.  After switching to the variable, I'm getting incorrect fill color: 
2. If using the CSS var, I need to set both `--_iui-button-icon-fill` and `--_iui-button-icon-fill-hover`.

apps/css-workshop/src/pages/breadcrumbs.astro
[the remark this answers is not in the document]
FlyersPh9: Oh jeez, fixed the typo.

Added this diff:

```diff
<button
  class="
    iui-button
+   iui-field
-   iui-button-dropdown
  "
+ data-iui-type="dropdown"
>
 ```

 ```diff
<button
  class="
    iui-button
+   iui-field
  "
- data-iui-variant="idea"
+ data-iui-variant="high-visibility"
+ data-iui-type="feedback"
>
 ```

File: packages/itwinui-css/src/button/button.scss (hunk 11 above)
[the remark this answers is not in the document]
Comment (FlyersPh9) [reply packages/itwinui-css/src/button/button.scss]: Is not reducing specificity, removed.

File: packages/itwinui-css/src/button/base.scss (hunk 12 above)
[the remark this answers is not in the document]
Comment (rohan-kadkol) [reply packages/itwinui-css/src/button/base.scss]: I believe it's used in `ButtonBase.tsx` in react:

https://github.com/iTwin/iTwinUI/blob/a01444b6a7409c9cf2c8c9a588b12acac02fdd42/packages/itwinui-react/src/core/utils/components/ButtonBase.tsx#L42

File: .changeset/ninety-numbers-vanish.md [hunk 16]
```diff
@@ -0,0 +1,30 @@
...
+<button
+  class="
+    iui-button
++   iui-field
+  "
+- data-iui-variant="idea"
++ data-iui-variant="high-visibility"
++ data-iui-type="feedback"
+>
+ ```
```
rohan-kadkol, line 30: nit

```suggestion
```

```diff
<button
  class="
    iui-button
+   iui-field
  "
- data-iui-variant="idea"
+ data-iui-variant="high-visibility"
+ data-iui-type="feedback"
>
```
```

File: packages/itwinui-css/src/input-container/input-container.scss
Comment (rohan-kadkol) [reply packages/itwinui-css/src/input-container/input-container.scss:1]: Ohh, I see. Yes, I think it would be great to add `.iui-field` here too despite the border conflict issue. Hopefully unsetting the border on `.iui-field` when `.iui-input-flex-container` is also present does not have any regressions. If there are no regressions and it is relatively easy to implement, then I think it's worth exploring imo.

File: packages/itwinui-css/src/button/base.scss (hunk 15 above)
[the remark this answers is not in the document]
Comment (rohan-kadkol) [reply packages/itwinui-css/src/button/base.scss:43]: Ohh, I see, then either way you suggested seems good: either directly setting `fill` or setting both `--_iui-button-icon-fill` and `--_iui-button-icon-fill-hover`.

So, the current way of directly setting `fill` also seems good.

File: packages/itwinui-css/src/button/base.scss (hunk 8 above)
Comment (rohan-kadkol) [reply packages/itwinui-css/src/button/base.scss:33]: I still feel `currentColor` would be safer.

Just curious, is there a strong reason/benefit to use `--iui-color-icon`? If so, and if we're confident that we're setting the `--_iui-button-icon-fill` variable correctly, then `--iui-color-icon` may be good too.

File: packages/itwinui-css/src/button/active.scss [hunk 17]
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+@mixin iui-button-active {
```
rohan-kadkol, line 4: This mixin seems to also be there in `base.scss`. One of them is a duplicate and can be removed?

packages/itwinui-css/src/button/active.scss
```diff
@@ -0,0 +1,36 @@
...
+  --_iui-button-active-stripe-inset-block: initial;
+  --_iui-button-active-stripe-inset-inline: initial;
+  --_iui-button-active-stripe-background: var(--iui-color-border-accent);
+  --_iui-button-icon-fill: var(--iui-color-icon-accent);
+  --_iui-button-icon-fill-hover: var(--iui-color-icon-accent);
+  --_iui-field-color-background-disabled: var(--iui-color-background-disabled);
```
rohan-kadkol, line 10: Is this variable needed to be set here? Because it seems to be set to the default value from `iui-field`.

packages/itwinui-css/src/button/base.scss
FlyersPh9: We discussed this in a call and have decided to stick with `--iui-color-icon` rather than switch it to `currentColor`.

packages/itwinui-css/src/button/base.scss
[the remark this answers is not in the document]
FlyersPh9: Ok.  I've added a sass variable `$disabledButtonSelectors` to target all of the disabled selectors.

.changeset/ninety-numbers-vanish.md
FlyersPh9: Good catch, I did not notice those extra spaces.  Removed.

packages/itwinui-css/src/button/active.scss
FlyersPh9: You are correct.  After our call earlier today where we discussed `base.scss` being set to default button variant values, I made sure the `@mixin iui-button-active` within `base.scss` is set to default values, and then moved the borderless active mixin to `borderless.scss`.

packages/itwinui-css/src/button/active.scss
FlyersPh9: It does appear to be necessary.

With that line included:
Screenshot 2024-04-02 at 5 39 34 PM

After removing that line:
Screenshot 2024-04-02 at 5 40 14 PM

I think it must have to do with removing the background from the disabled borderless button.

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: I've added `.iui-field` to `.iui-input-flex-container`.  I was able to remove a bunch of CSS variables specific to this component, but I left `--_iui-input-border-color` in because it made it easier to apply `--_iui-field-color-border` on the pseudo element.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -345,8 +343,6 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
 
 /// @arg $variant Can be either: type-indicator, quick-action
 @mixin iui-tile-thumbnail-button($variant) {
```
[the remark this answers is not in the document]
FlyersPh9: We're way off topic here, but I'd like to see the thumbnail area support all 4 corners and allow the user to put whatever they want in there.  If you want badges top right and buttons bottom left, go for it.

apps/css-workshop/backstop/results/bitmaps_reference/iTwinUI_table_Type_Extras_dark_0_demo-extras_0_desktop.png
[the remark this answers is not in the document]
FlyersPh9: It was intentional, but I don't know why we had such a weird width on the small pagination buttons prior.  Maybe we had a good reason that I am forgetting?

apps/css-workshop/src/pages/button.astro
```diff
@@ -414,17 +436,17 @@ import Layout from './_layout.astro';
         </span>
         <span>Notification</span>
       </button>
-      <button class='iui-button'>
+      <button class='iui-button iui-field'>
         <span
           class='iui-button-icon iui-notification-marker'
-          data-iui-variant='informational'
+          data-iui-variant='high-visibility'
```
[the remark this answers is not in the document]
FlyersPh9: Removed `data-iui-variant='high-visibility'` and `data-iui-variant='primary'` from `.iui-button-icon`.

apps/css-workshop/backstop/results/bitmaps_reference/iTwinUI_table_Type_Extras_dark_0_demo-extras_0_desktop.png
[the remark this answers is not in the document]
rohan-kadkol: Ohh, I see. I'm not sure if the previous width was intentional. But the reason looks like `borderless` was absent as the second parameter when using the `iui-button-size()` mixin (L71).

https://github.com/iTwin/iTwinUI/blob/376135738bbded28d4a460fb0e263a2bdab9eac5/packages/itwinui-css/src/table/paginator.scss#L66-L72

packages/itwinui-css/src/input/input.scss
rohan-kadkol: I'm not sure if this will lead to any accessibility issues. Because looks like the [color contrast criterion](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) applies also to placeholder text too?

image

I also tried looking at the HTML spec for [placeholder](https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute) and [disabled](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls:-the-disabled-attribute) but couldn't find any relevant information.

Since the native disabled `<input>`s show the placeholder ([playground](https://codepen.io/rohan-kadkol/pen/MWRGLaB?editors=1000)), I guess we should show it too?

/cc @iTwin/itwinui-1 @iTwin/itwinui-2

---

Also, if we are indeed going ahead with making the placeholder transparent on disabled inputs, I believe we would need a changeset for this.

packages/itwinui-css/src/global.scss
rohan-kadkol: > Not convinced this should be within `global.scss`.

If not in `global.scss`, maybe we can create a new file called `field.scss` in `src/utils`?

.changeset/ninety-numbers-vanish.md
```diff
@@ -0,0 +1,30 @@
+---
+"@itwin/itwinui-css": major
+---
+
+Add `.iui-field` class to unify & centralize basic styling for `<button>`, `<input>`, `<textarea>`, and `<select>`.
```
rohan-kadkol, line 5: Should we mention class names (e.g. `.iui-input`) instead of HTML components (e.g. `<input>`)?

Because `<label class=".iui-input-flex-container">` or `<div class=".iui-input-flex-container">` would also require `.iui-field`, and not just `<input>`. Also, we apply `.iui-field` on `<div class="iui-select-button">` too and not exclusively on `<select>`.

Also, overall, it would be great if you could confirm if all user-facing changes have been adequately mentioned in changesets (either in this one or in separate changesets).

rohan-kadkol (commented):
No more comments than the ones already opened. Excited for this change! 🚀

Since this is a big PR, it might be good to get at least one more thorough review and approval from someone else.

packages/itwinui-css/src/input/input.scss
mayank99: Why are we setting transparent? That's a terrible idea and I'm genuinely shocked to see it 😵‍💫

Lets keep the placeholder text color. CSS should not be used this way. The correct way to handle this would be using a combination of TS/React and documentation.

packages/itwinui-css/src/global.scss
mayank99: This should be in its own file, which can be imported in `all.scss`.

.changeset/ninety-numbers-vanish.md
FlyersPh9: Changeset updated to specify classes instead of elements.  Also went through each `.astro` file and made a point to mention other changes.

packages/itwinui-css/src/table/paginator.scss
```diff
@@ -64,18 +61,8 @@
 }
 
 @mixin iui-table-paginator-page-button {
-  @include base.iui-button;
-  @include borderless.iui-button-borderless;
-
-  &-small {
```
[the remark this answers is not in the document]
FlyersPh9: Added this and others to changeset.

packages/itwinui-css/src/global.scss
FlyersPh9: Moved out of `global.scss` to `/src/utils/field.scss` and imported in `all.scss`.

packages/itwinui-css/src/input/input.scss
FlyersPh9: Removed `transparent` styling from disabled placeholder text.  If we're going to remove the disabled placeholder text via TS/React, then I assume I should remove the CSS demos that show that?

mayank99: I'm fine with keeping the demos (they're only used for visual tests) or with removing them (since we don't want to support this scenario).

packages/itwinui-css/src/global.scss
mayank99: The `utils/` folder already creates a `utils.css` file (using [`utils.scss`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/src/utils/utils.scss)), so this one should live outside of `utils/` because we want it to be near the top of `all.scss`.

packages/itwinui-css/src/input/input.scss
FlyersPh9: Ok.  Styling has been reverted, CSS demos have been removed, and changeset has been updated.

packages/itwinui-css/src/global.scss
mayank99: Yeah `src/field.scss` is fine for now. We can think about organizing folders in the future when we introduce internal cascade layers.

.changeset/ninety-numbers-vanish.md
rohan-kadkol: Great, but not sure if the points are organized correctly. Because putting the points on separate lines in markdown would render as just one long paragraph.

image

So, maybe we need separate changeset files? Or maybe at least a bullet point list?

packages/itwinui-css/src/global.scss
FlyersPh9: Moved to `src/field.scss` and updated `all.scss` to reflect that.

.changeset/ninety-numbers-vanish.md
FlyersPh9: Good catch. Formatting should be better now.

packages/itwinui-css/src/button/button.scss
```diff
@@ -33,42 +32,34 @@
...
-  &[data-iui-variant='idea'] {
+  &[data-iui-type='feedback'] {
     @include idea.iui-button-idea;
   }
-}
 
-.iui-button-icon {
-  @include base.iui-button-icon;
-}
+  &[data-iui-type='dropdown'] {
```
mayank99, line 48: I'm thinking this one should stay as a class? `iui-button-dropdown`

Similarly, the ideas button could also be a class: `iui-button-idea`

Multiple reasons:
- The `data-iui-type` attribute feels a little strange; the name is too vague.
- "dropdown" and "idea" have nothing to do with each other. They affect totally different CSS properties.
- In React, `DropdownButton` and `IdeasButton` are their own components.

FlyersPh9: The thought was using a data attribute makes it a binary choice.  You can't combine a dropdown and an ideas button.  If split button wasn't actually a combination of 2 buttons but instead was 1 button, it would fall under this as well.

mayank99: That's not a strong enough argument, especially when you look at the other reasons I listed. You also can't combine a Table and a Tile.

apps/css-workshop/src/pages/table.astro
```diff
@@ -2261,41 +2261,77 @@ import Checkbox_ from '../components/Checkbox.astro';
 
           <span class='iui-table-paginator-pages-group'>
             <button
```
rohan-kadkol, line 2263: Should these also be `Button_`?

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -12,10 +12,15 @@
   }
 }
 
-// noop https://github.com/iTwin/iTwinUI/pull/1881#discussion_r1506166537
-/* stylelint-disable-next-line block-no-empty */
 .iui-search-icon {
+  --_iui-button-icon-fill: var(--iui-color-icon-muted);
+
+  [data-iui-size='small'] & svg {
+    inline-size: 0.75rem;
```
rohan-kadkol, line 19: Should this be `--iui-size-s` instead?

```suggestion
    inline-size: var(--iui-size-s);
```

packages/itwinui-css/src/searchbox/searchbox.scss
```diff
@@ -12,10 +12,15 @@
   }
 }
 
-// noop https://github.com/iTwin/iTwinUI/pull/1881#discussion_r1506166537
-/* stylelint-disable-next-line block-no-empty */
 .iui-search-icon {
+  --_iui-button-icon-fill: var(--iui-color-icon-muted);
```
FlyersPh9, line 16: Ok, this is a weird one, let me explain…after https://github.com/iTwin/iTwinUI/pull/1919/commits/c07bf5572e7e0f2a28a8eb37806dc78ce1c29817 the search input's search icon started failing CSS test images.

image

I found this odd because that commit didn't change any CSS, so I thought previous commits would have failed test images.

The reason it's failing has to do with a specificity or CSS order within `all.scss`.

| localhost | production |
| --- | --- |
| image | image |

One solution was to undo the noop of the `.iui-search-icon` class.  Another option is to undo the re-order of `utils/utils` within `all.scss`, but that caused specificity issues with `iui-field`.

.changeset/ninety-numbers-vanish.md
```diff
@@ -0,0 +1,15 @@
...
+"@itwin/itwinui-css": major
+---
+
+Add `.iui-field` class to unify & centralize basic styling for `.iui-button`, `.iui-input`, `.iui-input-flex-container`, and `.iui-select-button`.
+
+- Paginator buttons now use the `.iui-field` data attribute for sizing rather than the `.iui-table-paginator-page-button-small` modifier class.
+- `.iui-select-tag-button` now requires `.iui-button .iui-field` and `data-iui-variant="borderless"`.
+- `.iui-breadcrumbs-content` now requires `.iui-button .iui-field`.
+- `.iui-sidenav-button .iui-expand` now requires `.iui-button .iui-field` and `data-iui-size="small"`.
+- Feedback button type now requires `data-iui-variant`.
```
rohan-kadkol, line 11: Should this be the ideas button?

```suggestion
- Ideas button type now requires `data-iui-variant`.
```

Also, not sure if this point is outdated. Since we now need to add `.iui-button-idea` + `data-iui-variant="high-visibility"` instead of `data-iui-variant="idea"`?

packages/itwinui-css/src/all.scss
```diff
@@ -4,6 +4,8 @@
 
 @layer itwinui.v3 {
   @include meta.load-css('global');
+  @include meta.load-css('utils/utils');
```
FlyersPh9, line 7: `utils/utils` was moved further up the list because `.iui-button-base` was causing specificity / order issues with `.iui-field`.  This issue was only present in Ladle as the CSS demos do not have the `.iui-button-base` class.  Perhaps they should for a more 1:1 match with production? 🤔

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -661,7 +661,7 @@ const SelectButton = React.forwardRef((props, forwardedRef) => {
       data-iui-variant={styleType !== 'default' ? styleType : undefined}
       {...rest}
       ref={forwardedRef}
-      className={cx('iui-select-button', props.className)}
+      className={cx('iui-select-button iui-field', props.className)}
```
rohan-kadkol, line 664: This can be written as a list for clarity. Similar to .

Same comment for the other places too.

packages/itwinui-react/src/core/Buttons/IconButton.tsx
```diff
@@ -63,7 +63,7 @@ export const IconButton = React.forwardRef((props, ref) => {
   const button = (
     <ButtonBase
       ref={ref}
-      className={cx('iui-button', className)}
+      className={cx('iui-button iui-field', className)}
```
rohan-kadkol, line 66: This can be written as a list, similar to AvatarGroup.

https://github.com/iTwin/iTwinUI/blob/e2e6ba7abbb46ac99ed225775b56942b88d8da04/packages/itwinui-react/src/core/AvatarGroup/AvatarGroup.tsx#L112-L116

Same comment in other places too.

FlyersPh9: Done.  I caught 1 in toast that looks like it came from someone else's PR, ignored that 1.

packages/itwinui-react/src/core/Select/Select.tsx
FlyersPh9: Done.  I caught 1 in toast that looks like it came from someone else's PR, ignored that 1.

.changeset/ninety-numbers-vanish.md
FlyersPh9: Corrected, thanks!

apps/css-workshop/src/pages/table.astro
FlyersPh9: Done in `table.astro` and `breadcrumbs.astro`.

apps/css-workshop/src/pages/table.astro
```diff
@@ -2750,84 +2734,77 @@ import Checkbox_ from '../components/Checkbox.astro';
...
-              data-iui-variant='borderless'
-              data-iui-size='small'
-              data-iui-active='true'
+            <Button_
+              class='iui-table-paginator-page-button'
+              variant='borderless'
+              size='small'
               aria-label='Page 15'
-              type='button'
+              data-iui-active='true'
```
rohan-kadkol, line 2773: You can use the `active` prop instead.

```suggestion
              active={true}
```

Same comment for the similar places.

.changeset/ninety-numbers-vanish.md
mayank99, line 1: I think the changesets should be split up into multiple and they should clearly indicate the diffs for user-facing breaking changes.

Also the visual changes should probably be mentioned in separate changesets, and also be duplicated for React.

packages/itwinui-css/src/button/base.scss
```diff
@@ -2,37 +2,28 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 @use '../mixins';
 
+$disabledButtonSelectors: '[disabled], :disabled, [aria-disabled="true"], [data-iui-disabled="true"], .iui-disabled';
```
mayank99, line 5: I don't think we need `iui-disabled`. Is it used anywhere?

apps/css-workshop/src/components/Button.astro
```diff
@@ -11,8 +11,13 @@ const { class: className, variant, size, active, shift, iuiType, ...props } = As
 ---
 
 <button
-  class:list={['iui-button', { 'iui-button-dropdown': iuiType === 'dropdown' }, className]}
-  data-iui-variant={iuiType === 'feedback' ? 'idea' : variant}
+  class:list={[
+    'iui-field iui-button',
```
mayank99, line 15: We also need `iui-button-base` within this component, to match React implementation. Maybe can be handled in #2012

packages/itwinui-css/src/button/base.scss
```diff
@@ -41,30 +32,40 @@
   svg {
     @include mixins.iui-icon-style('m');
     transition: fill var(--iui-duration-1) ease-out;
-    fill: var(--_iui-button-icon-fill, currentColor);
+    fill: var(--_iui-button-icon-fill, var(--iui-color-icon));
+
+    button:hover &,
+    button:active & {
```
mayank99, line 38: This whole block looks a bit hacky, and it doesn't work for advanced cases like `<Button as="a">`. Maybe it should be handled in `iui-field` instead?

packages/itwinui-css/src/field.scss
```diff
@@ -0,0 +1,77 @@
...
+    border-color var(--iui-duration-1) ease-out,
+    color var(--iui-duration-1) ease-out;
+
+  &:where(:hover) {
+    background-color: var(--_iui-field-color-background-hover);
+    border-color: var(--_iui-field-color-border-hover);
+    color: var(--_iui-field-color-text-hover);
+  }
+
+  &:where(:active) {
```
mayank99, line 40: `:hover` and `:active` can be combined?

packages/itwinui-css/src/field.scss
```diff
@@ -0,0 +1,77 @@
...
+    border-color: var(--_iui-field-color-border-hover);
+    color: var(--_iui-field-color-text-hover);
+  }
+
+  &:where(:focus-visible) {
+    outline: var(--_iui-field-focus-outline-width) solid var(--_iui-field-color-focus-outline);
+    outline-offset: var(--_iui-field-focus-outline-offset);
+  }
+
+  &:where([disabled], :disabled, [aria-disabled='true'], [data-iui-disabled='true'], .iui-disabled) {
```
mayank99, line 51: Same question about `iui-disabled`

packages/itwinui-css/src/field.scss
```diff
@@ -0,0 +1,77 @@
...
+    --_iui-field-color-background-hover: var(--iui-color-background-transparent-hover);
+    --_iui-field-color-background-disabled: transparent;
+    --_iui-field-color-border: transparent;
+    --_iui-field-color-border-hover: transparent;
+    --_iui-field-color-border-disabled: transparent;
+    --_iui-field-focus-outline-offset: -1px;
+    --_iui-field-focus-outline-width: 1px;
+  }
+
+  @each $fieldSize in small, large {
```
mayank99, line 69: nit: this would be easier to read without a loop (and the same amount of code)

packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -106,39 +98,34 @@
...
-  border-radius: var(--iui-border-radius-1);
-  color: var(--_iui-input-text-color);
   position: relative;
-  block-size: var(--_iui-input-min-height);
   inline-size: 100%;
   isolation: isolate;
-  min-block-size: var(--_iui-input-min-height);
   block-size: fit-content;
   line-height: var(--iui-size-l);
+  border: none;
```
mayank99, line 111: This looks a bit random. What's the intent here? I'm honestly pretty confused looking at this file, and there is no useful explanation in the PR description/prior comments.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -230,14 +230,25 @@ const Separator = ({ separator }: Pick<BreadcrumbsProps, 'separator'>) => (
...
 
-  const defaultAs = !!props.href ? Anchor : !!props.onClick ? 'button' : 'span';
+  const resolvedAs = !!props.href ? 'a' : !!props.onClick ? 'button' : 'span';
+
+  const resolvedAsComponentsMap: Record<
+    typeof resolvedAs,
+    PolymorphicForwardRefComponent<'a' | 'button'> | 'span'
+  > = {
+    a: Anchor,
+    button: Button,
```
mayank99, line 240: This implementation looks overcomplicated. Why not just use `<Button styleType="borderless">` (instead of `<Box>`) and keep everything else the same as before?

packages/itwinui-react/src/core/Buttons/IdeasButton.test.tsx
```diff
@@ -7,16 +7,24 @@ import * as React from 'react';
 
 import { IdeasButton } from './IdeasButton.js';
 
+const assertBaseElement = (container: HTMLElement) => {
```
mayank99, line 10: nit: no need for unnecessary abstractions inside tests.

packages/itwinui-react/src/core/Input/Input.tsx
```diff
@@ -37,7 +37,7 @@ export const Input = React.forwardRef((props, ref) => {
   return (
     <Box
       as='input'
-      className={cx('iui-input', className)}
+      className={cx('iui-input iui-field', className)}
```
mayank99, line 40: nit: these should be handled using commas instead of spaces

packages/itwinui-react/src/core/Buttons/IdeasButton.tsx
```diff
@@ -29,7 +29,8 @@ export const IdeasButton = React.forwardRef((props, ref) => {
   return (
     <Button
       ref={ref}
-      data-iui-variant='idea'
+      className='iui-button-idea'
```
mayank99, line 32: `className` needs to be properly merged here

packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -175,9 +175,13 @@ export const TablePaginator = (props: TablePaginatorProps) => {
     (index: number, tabIndex = index === focusedIndex ? 0 : -1) => (
       <ButtonBase
         key={index}
-        className={cx('iui-table-paginator-page-button', {
-          'iui-table-paginator-page-button-small': buttonSize === 'small',
-        })}
+        className={cx(
+          'iui-button',
```
mayank99, line 179: This one should use `<Button>` directly. The `iui-button` class should not be used outside of `Button`.

packages/itwinui-css/src/all.scss
mayank99: I don't think utils should be moved all the way up here. The specificity battle you're running into could be solved by changing `:where(.iui-field)` to `.iui-field`.

apps/react-workshop/cypress-visual-screenshots/baseline/Tile.test.ts-All Props.png
mayank99, line 1: Not sure if it's intentional, but I think these buttons might have the wrong icon fill. The previous fill has more contrast against the blurred background.

packages/itwinui-css/src/searchbox/searchbox.scss
mayank99: See https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1576805293

apps/css-workshop/src/components/Button.astro
mayank99: cc @Ben-Pusey-Bentley

Ben-Pusey-Bentley: Okay, will do.

packages/itwinui-css/src/all.scss
FlyersPh9: I thought we wanted `.iui-field` with as low specificity as possible?  Regardless, I reverted `utils/utils` placement & removed `:where` from `.iui-field`.  Ran tests and made sure everything was passing as expected.  

Also makes https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1572542650 no longer needed.

packages/itwinui-css/src/searchbox/searchbox.scss
FlyersPh9: All reverted and working now. https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1578231027

packages/itwinui-css/src/button/base.scss
FlyersPh9: @rohan-kadkol, do I recall correctly that you found somewhere `.iui-disabled` was being used by `.iui-button`?

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: Ah, good question.

`.iui-field` defines the CSS variables and sets those variables on their respective properties such as `border`.
I added `.iui-field` to `.iui-input-container` so that we could use those CSS variables on the nested `<input>` and `<textarea>` (note that they do NOT use `.iui-input`).
However, the border is applied to those nested elements, NOT the `.iui-input-container` wrapper.  So this line removes the border from the wrapper.

Without this change, we have a double border:
Screenshot 2024-04-24 at 1 19 13 PM

packages/itwinui-react/src/core/Input/Input.tsx
FlyersPh9: Thought I fixed all of these, must have missed this one.  Fixed.

packages/itwinui-css/src/field.scss
FlyersPh9: Jeeeeez, how did I miss this one. 🤦🏻‍♂️ Cleaned up.

mayank99: You can combine it even further using `:where(:hover, :active)`

apps/react-workshop/cypress-visual-screenshots/baseline/Tile.test.ts-All Props.png
FlyersPh9: This was intentional, we found a visual bug in [this Ladle demo](https://itwin.github.io/iTwinUI/react/?story=tile--all-props) where the icons are black instead of white.  It seems like the tests remove the `` which makes the test look good but the end result with the `` not so great.

Screenshot for reference of the Ladle demo bug:
Screenshot 2024-04-24 at 1 40 23 PM

packages/itwinui-css/src/field.scss
FlyersPh9: Changed to the old fashioned way. 🥃

packages/itwinui-css/src/field.scss
FlyersPh9: Simplified even further.

packages/itwinui-css/src/button/base.scss
rohan-kadkol: Our previous discussion was actually about `data-iui-disabled` and not `.iui-disabled` (https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1548079112).

Searching for `.iui-disabled` in the react code, I found that it is used in `SplitButton`:

https://github.com/iTwin/iTwinUI/blob/e2e6ba7abbb46ac99ed225775b56942b88d8da04/packages/itwinui-react/src/core/Buttons/SplitButton.tsx#L114-L116

But I believe that can be removed from react since `.iui-disabled` is not in the disabled `SplitButton`'s HTML demo page.

packages/itwinui-react/src/core/Buttons/IdeasButton.test.tsx
rohan-kadkol: Since I found identical code, I thought to abstract them into `assertBaseElement()`, similar to other unit tests. So, not sure of the advantage of not abstracting common code in this case.

But I don't mind removing it too since the unit test file is anyways small. Reverted the abstraction (https://github.com/iTwin/iTwinUI/pull/1919/commits/a8e0b684479d9e89eb34e14cac0ef602fb88e3d0).

packages/itwinui-react/src/core/Table/TablePaginator.tsx
rohan-kadkol: Ohh, yes, changed to `Button` (https://github.com/iTwin/iTwinUI/pull/1919/commits/be3b0d361fc694bb867fa9dda17a85328176bba7).

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
rohan-kadkol: The reason was that with using a `Button`, we get the `Button`'s hover background when hovering over breadcrumb items that are supposed to be anchors.

<details>
<summary>Screen-recordings</summary>

HTML Demo Page:

https://github.com/iTwin/iTwinUI/assets/45748283/a5fbec2b-2933-4348-8daa-7def7db109e9

`Breadcrumbs → Links` story when using `Button styleType="borderless">`:

https://github.com/iTwin/iTwinUI/assets/45748283/44808e1e-efea-4c2b-bd5d-4e0fe082806f

</details>

<details>
<summary>Diff</summary>

```diff
diff --git a/packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx b/packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
index 4c39eecdd..fe7caa668 100644
--- a/packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
+++ b/packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
@@ -230,30 +230,20 @@ const Separator = ({ separator }: Pick<BreadcrumbsProps, 'separator'>) => (
 const BreadcrumbsItem = React.forwardRef((props, forwardedRef) => {
   const { children: childrenProp, className, ...rest } = props;
 
-  const resolvedAs = !!props.href ? 'a' : !!props.onClick ? 'button' : 'span';
-
-  const resolvedAsComponentsMap: Record<
-    typeof resolvedAs,
-    PolymorphicForwardRefComponent<'a' | 'button'> | 'span'
-  > = {
-    a: Anchor,
-    button: Button,
-    span: 'span',
-  };
-
+  const defaultAs = !!props.href ? Anchor : !!props.onClick ? 'button' : 'span';
   const children =
-    resolvedAs === 'button' ? <span>{childrenProp}</span> : childrenProp;
+    defaultAs === 'button' ? <span>{childrenProp}</span> : childrenProp;
 
   return (
-    <Box
-      as={resolvedAsComponentsMap[resolvedAs] as 'a'}
+    <Button
+      as={defaultAs as 'a'}
       className={cx('iui-breadcrumbs-content', className)}
-      data-iui-variant={resolvedAs === 'button' ? 'borderless' : undefined}
+      styleType='borderless'
       ref={forwardedRef}
       {...rest}
     >
       {children}
-    </Box>
+    </Button>
   );
 }) as PolymorphicForwardRefComponent<'a'>;
 BreadcrumbsItem.displayName = 'Breadcrumbs.Item';
```

</details>

packages/itwinui-css/src/field.scss
FlyersPh9: Looks like `select` is using `.iui-disabled` so needs to stay in `.iui-field` unless we were to update `select`.

packages/itwinui-css/src/button/base.scss
FlyersPh9: Removed from the button SCSS files (not split button).

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: > we get the Button's hover background when hovering over breadcrumb items that are supposed to be anchors.

Is this an actual problem? The breadcrumbs are still underlined to indicate that they're anchors rather than buttons, and the additional hover background is a more accurate indicator of the true size of the breadcrumb.

@FlyersPh9

packages/itwinui-css/src/field.scss
mayank99: Can you we change that to also use `data-iui-disabled`?

FlyersPh9: Done in https://github.com/iTwin/iTwinUI/pull/1919/commits/fb546c8c636fe0c67e3a0b217f8fed07f3f29adb, @rohan-kadkol could you take care of the React side for this please?

packages/itwinui-react/src/core/Buttons/IdeasButton.tsx
FlyersPh9: I think I did this correctly in https://github.com/iTwin/iTwinUI/pull/1919/commits/ad7e0ae9af16b5cc045b62eea13f2de0ca9ec73d but double check for me please.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
FlyersPh9: I just tried this in the CSS demo page like so:
```diff
<a
-  class='iui-anchor iui-breadcrumbs-content'
+  class='iui-anchor iui-button-base iui-button iui-field iui-breadcrumbs-content'
+  data-iui-variant='borderless'
  href='#'>
  Root
</a>
```

And am getting this result:
Screenshot 2024-04-25 at 2 18 32 PM

mayank99: I think you need a `<span>` around the text to make it truncate.

If you use `Button`, this should be handled automatically.

packages/itwinui-css/src/field.scss
rohan-kadkol: Absolutely, done (https://github.com/iTwin/iTwinUI/pull/1919/commits/88233c00a5b21b27ed7b7c237456c4750a36b9d3).

packages/itwinui-css/src/button/base.scss
FlyersPh9: Did this 👆🏻

packages/itwinui-react/src/core/Buttons/IdeasButton.test.tsx
mayank99: @rohan-kadkol The unit tests are failing.

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: Reverted `.iui-field` being required for `.iui-input-container`.

mayank99: I'm still seeing some `iui-field` references in there. https://github.com/iTwin/iTwinUI/blob/55a8896a10abf201f1eae790b8552b845aa168a6/packages/itwinui-css/src/input-container/input-container.scss#L39-L40

And there are some visual differences too; did you verify those?

packages/itwinui-react/src/core/Buttons/IdeasButton.test.tsx
rohan-kadkol: Ohh, fixed (https://github.com/iTwin/iTwinUI/pull/1919/files/55a8896a10abf201f1eae790b8552b845aa168a6..436c5404f92a06216cef4188801c3cf86fcf1cdf)

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
rohan-kadkol: Jon and I tried using a `Button`. Although it added a span, that was not sufficient to solve the issue.

I was able to fix the issue by adding some custom styles to the span.

<details>
<summary>Screenshot</summary>

image

</details>

We can try investigating in which selector to add those styles or if there is a better way to fix the issue. But since making the BreadcrumbsItem look like a Button is not directly related to this iui-field PR, I think it can be done later. If so, we can add an after PR TODO.

For now, I simplified the code a bit more (https://github.com/iTwin/iTwinUI/pull/1919/commits/1ddac82299ab427d48ef97abbea0a175fd57ccf4).

mayank99: I think this diffs should be enough to fix the truncation?

<details><summary>breadcrumbs.scss</summary>

```diff
@@ -35,6 +35,13 @@
 .iui-breadcrumbs-content {
   padding-inline: var(--iui-size-xs);
   overflow: hidden;
   text-overflow: ellipsis;
   line-height: var(--iui-component-height);
 
+  &,
+  > * {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
   &:where(a) {
     &:focus {
       outline-offset: -1px;
@@ -54,12 +61,6 @@
   &:where(button) {
     @include base.iui-button;
@@ -61,5 +61,5 @@
     &[aria-current] {
       color: var(--iui-color-text);
     }
-
-    > * {
-      overflow: hidden;
-      white-space: nowrap;
-      text-overflow: ellipsis;
-    }
```

</details>

rohan-kadkol: Yes, Jon and I had tried that (except we didn't add the `&` in the selector) before and it had worked. But I didn't confirm if anything else breaks as a result. Since it wasn't related to this PR, we considered having it as an after PR TODO.

@FlyersPh9 If you feel that nothing breaks on the css side with this change, I can make this change on the react side following your css/html demo changes.

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: Those lines appear to be within `.iui-input-grid` which nests `.iui-field .iui-input` within it.
`.iui-input-container` however does not use `.iui-field .iui-input` and instead targets the elements `<input>` & `<textarea>` directly.
So those lines I believe are expected to stay.

FlyersPh9: The visual changes are bug fixes.  `.iui-input-container` was not setting the correct text colors but now is.  Credit to @rohan-kadkol for the screenshot.
image

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
FlyersPh9: Got it working in https://github.com/iTwin/iTwinUI/pull/1919/commits/3c56408b945a2c7faf5e923453c5ab6797fc58cf.

rohan-kadkol: Great, made the change in React too (https://github.com/iTwin/iTwinUI/pull/1919/commits/d2d6e369b94ca470760d89da6d59bce2c54f7a24).

apps/css-workshop/src/pages/breadcrumbs.astro
```diff
@@ -38,7 +38,13 @@ import IconButton_ from '../components/IconButton.astro';
       </Button_>
       <ol class='iui-breadcrumbs-list'>
         <li class='iui-breadcrumbs-item'>
-          <a class='iui-anchor iui-breadcrumbs-content' href='#'> Root</a>
+          <a
```
mayank99, line 41: These should all use the `Button.astro` component.

Before:
```html
<a class='iui-anchor iui-breadcrumbs-content' href='#'> Root</a>
```

Current:
```html
<a
  class='iui-anchor iui-button-base iui-button iui-field iui-breadcrumbs-content'
  data-iui-variant='borderless'
  href='#'
>
  <span>Root</span>
</a>
```

Expected:
```html
<Button_
  as='a'
  class='iui-anchor iui-breadcrumbs-content'
  variant='borderless'
  href='#'
>Root</Button_>
```

packages/itwinui-css/src/button/variant.scss
```diff
@@ -1,44 +1,27 @@
...
+  );
+
+  @each $newVariant, $originalVariant in $variant-mappings {
+    &:where([data-iui-variant='#{$newVariant}']) {
+      --_iui-field-color-background: var(--iui-color-background-#{$originalVariant});
+      --_iui-field-color-background-hover: var(--iui-color-background-#{$originalVariant}-hover);
+      --_iui-field-color-border: var(--iui-color-background-#{$originalVariant});
+      --_iui-field-color-border-hover: var(--iui-color-background-#{$originalVariant}-hover);
     }
   }
```
mayank99, line 26: 1. prefer duplicating rather than looping for clarity
2. prefer `accent` over `informational` for high-visbility button

```suggestion
  &:where([data-iui-variant='high-visibility']) {
    --_iui-field-color-background: var(--iui-color-background-accent);
    --_iui-field-color-background-hover: var(--iui-color-background-accent-hover);
    --_iui-field-color-border: var(--iui-color-background-accent);
    --_iui-field-color-border-hover: var(--iui-color-background-accent-hover);
  }
  
  &:where([data-iui-variant='cta']) {
    --_iui-field-color-background: var(--iui-color-background-positive);
    --_iui-field-color-background-hover: var(--iui-color-background-positive-hover);
    --_iui-field-color-border: var(--iui-color-background-positive);
    --_iui-field-color-border-hover: var(--iui-color-background-positive-hover);
  }
```

packages/itwinui-css/src/input/input.scss
```diff
@@ -3,51 +3,26 @@
 @use '../mixins';
 
 @mixin iui-input {
-  --_iui-input-background-color: var(--iui-color-background);
-  --_iui-input-border-color: var(--iui-color-border-foreground);
-  --_iui-input-text-color: var(--iui-color-text);
+  --_iui-field-color-text-placeholder: var(--iui-color-text-placeholder);
```
mayank99, line 6: Since `--_iui-field-color-text-placeholder` does not exist in `iui-field`, this should be renamed or removed (in favor of using `color: var(--iui-color-text-placeholder)` directly).

packages/itwinui-css/src/select/select.scss
```diff
@@ -3,26 +3,14 @@
...
 @use '../mixins';
 @use '../tag/base' as tag;
-@use '../button/base';
-@use '../button/borderless';
 
 @mixin iui-select-button {
-  --_iui-input-background-color: var(--iui-color-background);
-  --_iui-input-border-color: var(--iui-color-border-foreground);
-  --_iui-input-text-color: var(--iui-color-text);
+  --_iui-field-color-text-placeholder: var(--iui-color-text-placeholder);
```
mayank99, line 8: Same comment about `--_iui-field-color-text-placeholder`.

packages/itwinui-css/src/field.scss
```diff
@@ -0,0 +1,71 @@
...
+  --_iui-field-height: var(--iui-component-height);
+  --_iui-field-focus-outline-offset: -2px;
+  --_iui-field-focus-outline-width: 2px;
+  --_iui-field-font-size: var(--iui-font-size-1);
+
+  background-color: var(--_iui-field-color-background);
+  border: 1px solid var(--_iui-field-color-border);
+  border-radius: var(--iui-border-radius-1);
+  color: var(--_iui-field-color-text);
+  font-family: inherit;
```
mayank99, line 23: maybe nit: `font: inherit` might be safer, to ensure [all font properties](https://developer.mozilla.org/en-US/docs/Web/CSS/font#constituent_properties) are inherited

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
```diff
@@ -38,28 +38,27 @@
...
+  &,
+  > * {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
+  &:where(a):focus {
+    outline-offset: -1px;
   }
```
rohan-kadkol, line 52: To confirm, is this needed? I think when looking at this together, we felt that it is already handled by `iui-field`?

apps/css-workshop/src/pages/breadcrumbs.astro
FlyersPh9: Looks like `Button.astro` doesn't support `as='a'` right now.  @Ben-Pusey-Bentley, can you help with this?

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
mayank99: I think this whole thing could be simplified?

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
```diff
@@ -38,28 +38,27 @@
   white-space: nowrap;
   text-overflow: ellipsis;
   line-height: var(--iui-component-height);
+  border: none;
```
mayank99, line 41: Not sure why the `border: none` was added, but I think it helps in forced-colors mode, even if it's not visible in regular themes.

Also see https://github.com/iTwin/iTwinUI/pull/1919#r1586559325

apps/css-workshop/src/pages/breadcrumbs.astro
mayank99: This might be all you need to support `as='a'`:

<details><summary>Button.astro</summary>

```diff
@@ -5,12 +5,13 @@ type Props = {
   isActive?: boolean;
   dropdown?: boolean;
   idea?: boolean;
+  as?: string;
 } & astroHTML.JSX.ButtonHTMLAttributes;
 
-const { class: className, variant, size, isActive, dropdown, idea, ...props } = Astro.props;
+const { as: Element = 'button', class: className, variant, size, isActive, dropdown, idea, ...props } = Astro.props;
 ---
 
-<button
+<Element
   class:list={[
```

</details>

packages/itwinui-css/src/input/input.scss
FlyersPh9: Using `color: var(--iui-color-text-placeholder)` directly.

packages/itwinui-css/src/select/select.scss
FlyersPh9: Using `color: var(--iui-color-text-placeholder)` directly.

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
FlyersPh9: Did it so test images didn't change, but I removed it and approved the new images. https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1586629845

apps/css-workshop/src/pages/breadcrumbs.astro
FlyersPh9: Astro component updated and being used for breadcrumb anchors. https://github.com/iTwin/iTwinUI/pull/1919/commits/84b7e39dcf12ca4a3e16f902897d9f20e785d847

mayank99 (approved):
LGTM!

I would maybe wait for another approval though, just to make sure nothing was missed.

cc @Ben-Pusey-Bentley

.changeset/lazy-seahorses-share.md
```diff
@@ -0,0 +1,13 @@
+---
+"@itwin/itwinui-react": major
```
Ben-Pusey-Bentley, line 2: This changeset should be on the css package, not react.

mayank99:
```suggestion
"@itwin/itwinui-css": major
```

.changeset/perfect-oranges-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
```
Ben-Pusey-Bentley, line 2: This changeset should be on the react package.
```suggestion
"@itwin/itwinui-react": patch
```

apps/css-workshop/src/components/Button.astro
```diff
@@ -5,20 +5,33 @@ type Props = {
...
-    { 'iui-button-idea': iuiType === 'feedback' },
+    'iui-button',
+    'iui-button-base',
+    'iui-field',
+    { 'iui-button-dropdown': dropdown },
+    { 'iui-button-idea': idea },
     className,
   ]}
-  data-iui-variant={variant}
+  type='button'
```
rohan-kadkol, line 32: Should this be there only when `Element === "button"`

.changeset/lazy-seahorses-share.md
```diff
@@ -0,0 +1,13 @@
+---
+"@itwin/itwinui-css": major
+---
+
+`.iui-sidenav-button.iui-expand` now requires `.iui-button .iui-field` and `data-iui-size="small"`.
+
+```diff
+  <button
+-   class="iui-button iui-button-base iui-sidenav-button iui-expand"
++   class="iui-button iui-button-base iui-field iui-sidenav-button iui-expand"
```
rohan-kadkol, line 10: nit/optional: It would be clearer if `iui-sidenav-button` is moved to the same position as in the old line.

```suggestion
-   class="iui-button iui-button-base iui-sidenav-button iui-expand"
+   class="iui-button iui-button-base iui-sidenav-button iui-field iui-expand"
```

.changeset/cyan-candles-walk.md
```diff
@@ -0,0 +1,16 @@
+---
+"@itwin/itwinui-css": major
+---
+
+`.iui-table-paginator-page-button` now requires the same markup as borderless buttons. The `.iui-table-paginator-page-button-small` modifier class has been removed.
+
+```diff
+  <button
+-   class="iui-table-paginator-page-button iui-table-paginator-page-button-small"
++   class="iui-button iui-button-base iui-field iui-table-paginator-page-button"
```
rohan-kadkol, line 10: nit/optional: Reordering the class names makes the diff slightly clearer.

```suggestion
-   class="iui-table-paginator-page-button iui-table-paginator-page-button-small"
+   class="iui-table-paginator-page-button iui-button iui-button-base iui-field"
```

.changeset/chilled-dolls-promise.md
rohan-kadkol, line 1: Should this also mention that `.iui-select-tag-button-icon` now needs `.iui-button-icon`?

rohan-kadkol (approved):
LGTM! 🚀

Just a few more comments of which some are nit/optional.

.changeset/chilled-dolls-promise.md
mayank99: Feel free to suggest a better complete diff.

apps/css-workshop/src/components/Button.astro
mayank99: We could do something like this:

```suggestion
  type={Element === 'button' ? 'button' : undefined}
```

But it's complicated. It wouldn't work for more advanced cases `as={SomeButton}`.

I'd say lets leave it as-is, since this isn't a public component.

.changeset/chilled-dolls-promise.md
mayank99: Fixed in [`7c495fb`](https://github.com/iTwin/iTwinUI/pull/1919/commits/7c495fb7909382fccbb16193e609f81f9bb678ef)

PR #1834, opened 2024-02-07
Make vite server workspaces' dev script persistent in turbo

rohan-kadkol:
## Changes

After #1826, upon running `yarn dev` from the root, I noticed that the vite servers (`vite-playground` and `css-workshop`) were not started. Upon starting only those servers using [`--filter`](https://turbo.build/repo/docs/core-concepts/monorepos/filtering), I realized that the turbo command was just ending even though there were no errors.

So, this PR just adds [`persistent: true`](https://turbo.build/repo/docs/reference/configuration#persistent) to `vite-playground#dev` & `css-workshop#dev`.

|Before|With this PR|
|-|-|
|image<br/>Ending by itself|image<br />Ending only after <kbd>Ctrl</kbd> + <kbd>C</kbd>|

Related: https://github.com/vercel/turbo/issues/6272

## Testing

* Confirmed that the vite-playground and the css-workshop now open upon running `yarn dev` from the root.

## Docs

N/A

PR #1933, opened 2024-03-19
migrate css-workshop to astro

mayank99:
## Changes

Migrated `css-workshop` from regular HTML pages over to Astro. This will open up the path to add Astro components in future PRs to prevent duplication of markup. (I also tried 11ty but it wasn't sparking joy).

The migration started out pretty smooth: I used the existing `.html` pages in Astro. But it quickly devolved into this +30k LOC PR because:
- Astro doesn't resolve npm imports (e.g. our CSS packages) inside html components.
- Astro doesn't allow adding layouts around html components. The layout is used for common `<head>` stuff.

Other than that, the biggest change is that many pages needed to be reformatted to remove whitespace, in order to make existing visual tests pass. This is because Astro's minifier is broken and doesn't collapse whitespace like it should. I've opened a [feature request](https://github.com/withastro/roadmap/discussions/868) in their repo.

Also, the port has been changed to `3050` for both dev and build (previously only used for build).

Lastly, all the `.js` files needed to be renamed to `.cjs` because the workspace now uses `type: module`.

## Testing

Most tests are unchanged, with a few exceptions where I think the current images look better with the extra whitespace. I got the number of changed images down from ~300 to ~30, so hopefully it's easier to review.

## Docs

N/A. Not user-facing.


AdamMeza-Bentley (approved):

rohan-kadkol (approved):
LGTM 🚀

I noticed that the portal dev server was not starting up for me. Maybe something like #1834 is needed. But that can come later.

mayank99 (commented):
> I noticed that the portal dev server was not starting up for me. Maybe something like #1834 is needed. But that can come later.

Hmm yeah, it's not starting for me either. I tried increasing the concurrency, to no effect. Maybe we need to decrease the amount of dev servers we start by default.

Edit: actually `persistent` seems to work? Pushed in [`19d5cc6`](https://github.com/iTwin/iTwinUI/pull/1933/commits/19d5cc62c93e91316c2b22da361da65156c23429)

PR #1934, opened 2024-03-20
Added `includesSome` to `DefaultFilterTypes` in Table

rohan-kadkol:
## Changes

A user pointed out that we have an `includesSome` function in `defaultFilterFunctions`:

https://github.com/iTwin/iTwinUI/blob/84007fe670a9ca85f82f230c3458c0a73fd1b622/packages/itwinui-react/src/core/Table/filters/defaultFilterFunctions.ts#L119-L134

… but don't have `includeSome` in `DefaultFilterTypes`:

https://github.com/iTwin/iTwinUI/blob/84007fe670a9ca85f82f230c3458c0a73fd1b622/packages/itwinui-react/src/react-table/react-table.ts#L747-L755

This PR adds `includesSome` to `DefaultFilterTypes` so that users can get `"includesSome"`  as a type suggestion for the `filter` prop and udpates the list of possible values in the JSDocs accordingly.

Unrelated: Fixed the broken GitHub file link in the `filter` JSDocs.

## Testing

Tested that passing `filter: "includesSome"` to a column even before this PR worked.

This PR only fixes the type. Regarding testing that type, since the storybook and vite playground did not show the TS autocomplete correctly (not sure why), I was unable to properly test it.

image

Open to ideas to test this, if testing it is needed.

**EDIT**: After adding `AnyString` (https://github.com/iTwin/iTwinUI/pull/1934#issuecomment-2012636918), we now get auto completions.

image

## Docs

Added changesets.

### DIFF
--- a/.changeset/itchy-swans-jog.md
+++ b/.changeset/itchy-swans-jog.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Fixed the Table column `filter`'s type to include `"includesSome"`, which is a default filter function that we already supported.
+Fixed the types for `Table` column `filter` to allow `"includesSome"`, which is an already-supported filter function.
--- a/.changeset/itchy-swans-jog.md
+++ b/.changeset/itchy-swans-jog.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Fixed the types for `Table` column `filter` to allow `"includesSome"`, which is an already-supported filter function.
+Fixed the types for `Table` column `filter` to allow `"includesSome"`, which is an already-supported filter function. Also improved the types for the `filter` prop to improve TS autocompletion.
--- a/.changeset/itchy-swans-jog.md
+++ b/.changeset/itchy-swans-jog.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed the Table column `filter`'s type to include `"includesSome"`, which is a default filter function that we already supported.
--- a/packages/itwinui-react/src/react-table/react-table.ts
+++ b/packages/itwinui-react/src/react-table/react-table.ts
@@ -196,7 +196,7 @@ export interface ColumnInterface<D extends Record<string, unknown> = {}>
   /**
    * String value or custom function to use for filtering.
    * Possible string values: `text`, `exactText`, `exactTextCase`, `includes`, `includesAll`, `exact`, `equals`, `between`.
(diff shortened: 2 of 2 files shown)


.changeset/itchy-swans-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed the Table column `filter`'s type to include `"includesSome"`, which is a default filter function that we already supported.
```
mayank99, line 5: Slight rephrase (emphasizing `Table` component, rewording "include", and removing "we").

```suggestion
Fixed types for `Table` column `filter` to allow `"includesSome"`, which is an already-supported filter function.
```

mayank99 (approved):

Ben-Pusey-Bentley (approved):
I think autocomplete might not be working because of how the type for filter includes `| string`, making any string valid input type wise. You can import `DefaultFilterTypes` instead to make sure that `includesSome` is included in that type. LGTM.

rohan-kadkol:
> I think autocomplete might not be working because of how the type for filter includes `| string`, making any string valid input type wise. You can import `DefaultFilterTypes` instead to make sure that `includesSome` is included in that type. LGTM.

Ohh, yes, looks like the reason for no autocomplete suggestions was the `| string` and not any other type issues (TS types still don't work in the Table react story, but they work in the vite playground, which is where I retested it now). Thanks for pointing that out.

So, looks like some users may not get type suggestions for the `filter` prop. But even if not, we at least mentioned it in the JSDocs for that prop, and so I think that would be sufficient.

mayank99 (commented):
You can also use [`AnyString`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/utils/types.ts#L9) if `string` is an issue.

mayank99 (approved):

rohan-kadkol:
> You can also use [`AnyString`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/utils/types.ts?rgh-link-date=2024-03-21T13%3A58%3A09Z#L9) if `string` is an issue.

I tried using it but when testing, I encountered errors in finding `@itwin/itwinui-react/react-table` when importing it from `App.tsx`, even though it exists. Not sure why that's happening.

As discussed via chat, I went ahead with duplicating `AnyString` since importing it from the util types file gave issues (https://github.com/iTwin/iTwinUI/pull/1934/commits/8c830c2069c13c164debd50df59cda73b536ac3b).

PR #1935, opened 2024-03-20
Removes `iui-avatar-size` mixin and replaces size classes with `data-iui-size` attribute.

Ben-Pusey-Bentley:
## Changes

This PR has two changes regarding the `iui-avatar` class in the css package. First, it removes the `iui-avatar-size` mixin, simplifying the css a little. This mixin was also removed from `iui-tile`, as well as the syling that made the default size `x-large` for all `iui-avatar` elements inside the thumbnail. This is a minor change, as the size can be easily set on the avatar itself. The second change was replacing the size classes with the `data-iui-size` attribute. I decided to make this change because the size being controlled by classes allows for some confusion in cases where multiple size classes are defined for one avatar. This change also brings avatar in line with how components like button and input handle size. The `css-workshop`, react package, and react unit tests have all also been updated to account for these changes.

## Testing

Made sure that react unit tests and css and react visual tests all pass with these changes.

## Docs

Added both minor and major changesets for the css package.


AdamMeza-Bentley (approved):

FlyersPh9:
Nice, I love seeing the use of `data-iui-size="small"` over a class modifier like `.iui-small`. 👍🏻

packages/itwinui-react/src/core/Avatar/Avatar.tsx
```diff
@@ -99,11 +99,8 @@ export const Avatar = React.forwardRef((props, ref) => {
   return (
     <Box
       as='span'
-      className={cx(
-        'iui-avatar',
-        { [`iui-${size}`]: size !== 'medium' },
-        className,
-      )}
+      className={cx('iui-avatar', className)}
+      data-iui-size={size}
```
rohan-kadkol, line 103: Based on the CSS, `data-iui-size` doesn't seem to be needed when `size === "medium"`. So we should likely add the data attribute only when `size !== "medium"`.

Same for `AvatarGroup`.

Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1935/commits/9bf284d63c2559846c30a87f20c757e1d6647d67) Also updated unit tests.

packages/itwinui-react/src/core/AvatarGroup/AvatarGroup.test.tsx
```diff
@@ -157,11 +157,17 @@ it.each(['small', 'medium', 'large', 'x-large'] as Array<
...
+    expect(
+      container.querySelectorAll(
+        `.iui-avatar-list > .iui-avatar[data-iui-size="${size}"]`,
+      ).length,
+    ).toBe(6);
+  } else {
+    expect(
+      container.querySelectorAll(`.iui-avatar-list > .iui-avatar`).length,
+    ).toBe(6);
+  }
```
rohan-kadkol, line 170: nit: You can add the if-else/turnery check for the querySelector string instead of for the expect.

```jsx
expect(container.querySelectorAll( size !== medium ? "…" : "…" ))
```

.changeset/silly-bananas-peel.md
```diff
@@ -0,0 +1,12 @@
+---
+"@itwin/itwinui-css": major
+---
+
+Replaces the size classes for `iui-avatar` with a new `data-iui-size` attribute.
+
+```diff
+-<div class="iui-avatar iui-small">
++<div class="iui-avatar" data-iui-size="small">
```
rohan-kadkol, line 9: nit: Spaces makes it easier to read.


```suggestion
- <div class="iui-avatar iui-small">
+ <div class="iui-avatar" data-iui-size="small">
```

Same for the other changeset.

apps/css-workshop/pages/assets/avatar.js
```diff
@@ -35,9 +35,10 @@ class Avatar extends HTMLElement {
 
     const innerHtml = `
       <span
-        class="iui-avatar iui-${size}"
+        class="iui-avatar"
         title="${person.title}"
         style="background-color: ${person.color};"
+        ${size ? 'data-iui-size=' + size : ''}
```
rohan-kadkol, line 41: Should `size` be in double quotes? It seems to work without it also, but having the double quotes would be safer.

```suggestion
        ${size ? `data-iui-size="${size}"` : ''}
```

Same for the `status` on the next line.

rohan-kadkol (approved):

packages/itwinui-react/src/core/AvatarGroup/AvatarGroup.test.tsx
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1935/commits/2a7e3590d20c4637003dc979ee24f0fde12f2314)

.changeset/silly-bananas-peel.md
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1935/commits/82d4feb21e6ac25c68efaf2cbeaa8b59d1ba02a5)

apps/css-workshop/pages/assets/avatar.js
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1935/commits/62c831689c5e557abaa65b7cbd500041c534045f)

rohan-kadkol (approved):

PR #1936, opened 2024-03-20
Fix showing `VisuallyHidden` text inside `ProgressRadial` initially

### LINKED ISSUE #1930: ProgressRadial spinner has "Loading." text visible for short time

image
When spinner appears it sometimes shows "Loading." text for a split second. This easily reproduces on mobile devices (even more visible on slower ones) and looks like a glitch.

https://github.com/iTwin/iTwinUI/assets/88542754/59410e07-abea-43ea-b03b-a170cb800935

### Expected Behavior

No "Loading." text when spinner is shown.

### Link to minimal repro

https://stackblitz.com/edit/github-ccf7j7?file=src%2FApp.tsx

### Steps To Reproduce

1. open minimal reproduction example on mobile device
2. refresh page to render spinner
3. "Loading." text is visible for a split second 
If device is slower it will reproduce almost every time, also reproduces with faster devices but less often.

### Anything else?

Seems like issue appeared after these changes https://github.com/iTwin/iTwinUI/pull/1859

rohan-kadkol:
## Changes

Fixes #1930.

The issue was that the visually hidden "Loading." was rendered until the Shadom DOM was attached.

|Before the Shadow DOM is attached|After the Shadow DOM is attached|
|-|-|
|image|Capture d’écran 2024-03-20 à 10 58 09 AM|

Since `ShadowRoot` attaches the Shadow DOM only after the first frame (since it exits early on the first frame):

https://github.com/iTwin/iTwinUI/blob/84007fe670a9ca85f82f230c3458c0a73fd1b622/packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx#L65-L69

… the `childrenProp` was being rendered as plain text outside the Shadow DOM on the first frame. This PR makes sure to mount the `childrenProp` only after the first frame, after the Shadow DOM has been attached.

<details>
<summary>Old solution</summary>

In `VisuallyHidden.tsx`:

https://github.com/iTwin/iTwinUI/blob/84007fe670a9ca85f82f230c3458c0a73fd1b622/packages/itwinui-react/src/core/VisuallyHidden/VisuallyHidden.tsx#L44-L47

The `childrenProp` was placed outside the `<ShadowRoot>` to gain access to outer css too. But in this case, since the `childrenProp` is meant to be visually *hidden*, trying to get outside css access is not needed.

This PR moves `childrenProp` in the `<ShadowRoot>`. This makes the "Loading." be inside a `<template>` and thus not rendered, even before being attached to the shadow root.

|Before being attached to the shadow root|After being attached to the shadow root|
|-|-|
|image|image|

After PR TODOs:

- [ ] @mayank99 suggested two solutions, one of which is the one I used since it was similar to what I was doing myself. But a more in-depth solution would be to look into `ShadowRoot.tsx` as part of Mayank's second solution that involves moving the css part before the `queueMicrotask` and/or also avoiding rendering `null` for the first client render when ssr is not enabled. But this can come in a later PR.

</details>

After PR TODOs:

- [ ] @mayank99 suggested two solutions, one of which is the one I used in this PR. But a more in-depth solution would be to look into `ShadowRoot.tsx` as part of Mayank's second solution that involves moving the css part before the `queueMicrotask` and/or also avoiding rendering `null` for the first client render when ssr is not enabled. But this can come in a later PR.

## Testing

Confirmed that "Loading." is no longer displayed initially for a short while.

## Docs

Added changeset.

### DIFF
--- a/.changeset/wet-phones-knock.md
+++ b/.changeset/wet-phones-knock.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Fixed the issue where the visually hidden "Loading." text was sometimes incorrectly displayed in the `ProgressRadial` component for a short while in the beginning.
+Fixed an issue in `ProgressRadial` where the visually hidden "Loading." text was sometimes incorrectly displayed for a short while in the beginning.
--- a/.changeset/wet-phones-knock.md
+++ b/.changeset/wet-phones-knock.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed the issue where the visually hidden "Loading." text was sometimes incorrectly displayed in the `ProgressRadial` component for a short while in the beginning.
--- a/packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
@@ -1014,8 +1014,7 @@ it('should update live region when selection changes', async () => {
     />,
   );
 
-  const liveRegion = container.querySelector('[aria-live="polite"]')
-    ?.shadowRoot;
+  const liveRegion = container.querySelector('[aria-live="polite"]');
   expect(liveRegion).toHaveTextContent('');
 
   await userEvent.tab();
--- a/packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
@@ -1014,7 +1014,8 @@ it('should update live region when selection changes', async () => {
     />,
   );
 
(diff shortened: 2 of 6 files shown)


.changeset/wet-phones-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed the issue where the visually hidden "Loading." text was sometimes incorrectly displayed in the `ProgressRadial` component for a short while in the beginning.
```
mayank99, line 5: Frontload the component name

```suggestion
Fixed an issue in `ProgressRadial` where the visually hidden "Loading." text was sometimes incorrectly displayed for a short while in the beginning.
```

mayank99 (approved):

Ben-Pusey-Bentley (approved):

PR #1938, opened 2024-03-21
Removes Breadcrumbs Mixins

Ben-Pusey-Bentley:
## Changes

Removes breadcrumbs mixins from the css package. This change effects the `iui-header-breadcrumbs` and `iui-header-breadcrumbs-list` classes, which now both need the `iui-breadcrumbs` and `iui-breadcrumbs-list` classes, respectively, to be styled correctly. The react package and `css-workshop` have both been updated to reflect this change.

## Testing

Ran visual tests for both css and react to make sure that there were no unintended changes to styles in this change.

## Docs

Added major changeset to the css package for the newly needed classes on the `iui-header-breadcrumbs` elements.


packages/itwinui-css/src/header/header.scss
```diff
@@ -73,7 +72,7 @@
 }
 
 .iui-header-breadcrumbs {
-  @include breadcrumbs.iui-breadcrumbs($breadcrumbs-height: 100%);
+  --_iui-breadcrumbs-height: 100%;
```
rohan-kadkol, line 75: Since `.iui-header-breadcrumbs` and `.iui-breadcrumbs` both have the same specificity and both set `--_iui-breadcrumbs-height`, I think it's brittle.

Should we have a `:where(.iui-breadcrumbs)` only for setting the value of `--_iui-breadcrumbs-height` (where has 0 specificity)? cc: @iTwin/itwinui-1 

`breadcrumbs.scss`:
```diff
.iui-breadcrumbs {
-  --_iui-breadcrumbs-height: var(--iui-component-height);

  @include mixins.iui-reset;
  display: flex;
  align-items: center;
  block-size: var(--_iui-breadcrumbs-height);
}

+ :where(.iui-breadcrumbs) {
+   --_iui-breadcrumbs-height: var(--iui-component-height);
+ }
```

.changeset/witty-cooks-tap.md
```diff
@@ -0,0 +1,12 @@
...
+"@itwin/itwinui-css": major
+---
+
+`iui-breadcrumbs` will need to be added to the classes of all `iui-header-breadcrumbs` elements. `iui-breadcrumbs-list` will also need to be added to the classes of all `iui-header-breadcrumbs-list` elements.
+
+```diff
+-<nav class="iui-header-breadcrumbs">
+-<ol class="iui-header-breadcrumbs-list">
++<nav class="iui-breadcrumbs iui-header-breadcrumbs">
++<ol class="iui-breadcrumbs-list iui-header-breadcrumbs-list">
```
rohan-kadkol, line 11: nit: Spaces make it easier to read.

```suggestion
- <nav class="iui-header-breadcrumbs">
- <ol class="iui-header-breadcrumbs-list">
+ <nav class="iui-breadcrumbs iui-header-breadcrumbs">
+ <ol class="iui-breadcrumbs-list iui-header-breadcrumbs-list">
```

packages/itwinui-css/src/header/header.scss
Ben-Pusey-Bentley: That makes sense to me. I'll work on implementing it after I fix the merge conflict.

.changeset/witty-cooks-tap.md
Ben-Pusey-Bentley: Will do.

AdamMeza-Bentley (approved):

packages/itwinui-css/src/header/header.scss
mayank99: I think it's fine as-is. We control the order of rules in `all.css`, so `iui-breadcrumbs` will always come first.

The proper fix for this would be to use internal cascade layers to set priority order. We can look into that after #1919

rohan-kadkol (approved):

PR #1939, opened 2024-03-21
ButtonGroup: conditionally run overflow logic

### LINKED ISSUE #1923: `ButtonGroup` overflow logic is running w/o `overflowButton` prop

`ButtonGroup` will re-render multiple times based on visible nodes returned by the `useOverflow` even when `overflowButton` prop is not provided. This complicates a use-case where a custom overflow logic is needed.

### Expected Behavior

`ButtonGroup` should always render provided children as is when `overflowButton` is not specified.

### Link to minimal repro

https://stackblitz.com/edit/github-fywssb?file=src%2FApp.tsx

### Steps To Reproduce

1. Open console of the repro
2. Click `Calculate overflow` (simulates conditional rendering of overflow button)
3. Observe that the overflow button is toggled correctly (render is fine)
4. Observe the console - when the overflow button is rendered it logs HTMLCollection w/ 2 elements.

### Anything else?

Final render is fine due to later re-renders.

mayank99:
## Changes

When `overflowButton` is not passed to `ButtonGroup`:

1. Early return `children`
2. Don't attach `overflowRef` 

Fixes #1923

## Testing

Recreated the sandbox from linked issue and verified that `children.length` is 4 (not 2) after clicking "Calculate overflow".

No tested added, since this is more of an optimization (the final result is correct even before this PR, because it eventually re-renders into correct state).

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
+++ b/packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
@@ -93,7 +93,7 @@ export const ButtonGroup = React.forwardRef((props, ref) => {
     !overflowButton,
     orientation,
   );
-  const refs = useMergedRefs(overflowRef, ref);
+  const refs = useMergedRefs(!!overflowButton ? overflowRef : null, ref);
 
   return (
     <FloatingDelayGroup delay={{ open: 50, close: 250 }}>
@@ -114,6 +114,10 @@ export const ButtonGroup = React.forwardRef((props, ref) => {
           {...rest}
         >
           {(() => {
+            if (!overflowButton) {
+              return children;
+            }
+
             if (!(visibleCount < items.length)) {
               return items;
             }
--- a/.changeset/tricky-vans-jam.md
+++ b/.changeset/tricky-vans-jam.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed an issue in `ButtonGroup` where the overflow logic was running even when `overflowButton` prop was not passed.


packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -114,6 +114,10 @@ export const ButtonGroup = React.forwardRef((props, ref) => {
           {...rest}
         >
           {(() => {
+            if (!overflowButton) {
+              return children;
```
rohan-kadkol, line 118: Should this be `return items;`?

mayank99: `children` is more direct and I think React has optimizations for directly passing through `children`. `items` is just an arrayed version of `children`.

Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
rohan-kadkol: So then, should the second return ([code](https://github.com/iTwin/iTwinUI/pull/1939/files#diff-ced50949d49dd8fda606519a8fa8e193e7e7254479750d9b5c3dd56bbb354730R122)) be returning `children`?

mayank99: That code is not related to this PR. Can we leave it be?

I think of it as two main cases:
- non-overflow mode: always return the children untampered
- - overflow mode: always return an array (even if it may sometimes be _all_ children elements)

rohan-kadkol (approved):

Issue #1925, opened 2024-03-18
Disabled `MenuItem` should not display `subMenuItems`?


GerardasB:
### Describe the bug (current behavior)

Currently a `disabled` menu item will show it's `subMenuItems`. Does it make sense to change this behavior?
I.e. a disabled `MenuItem` will not call it's `onClick` handler.

### Expected Behavior

A disabled menu item should not show `subMenuItems` when hovered.

### Link to minimal repro

https://stackblitz.com/edit/github-cpdu8r?file=src%2FApp.tsx

### Steps To Reproduce

- Click `Menu`
- Hover `Item 1_2` (note: this item is disabled)
- Observe that the submenu is opened.

### Anything else?

This can be handled conditionally at the consumer level, i.e.:
`<MenuItem disabled={disabled} subMenuItems={disabled ? undefined : subMenuItems} />`

mayank99:
I think it kinda makes sense. Click and hover/focus are different actions. Disabled only disables clicks by default.

If we ever support #879, I think disabling `subMenuItems` makes sense.

Since this issue has a pretty simple workaround, I will close this issue.

mayank99:
This is fixed / made consistent in [3.10.0](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.10.0) 🚀

Issue #653, opened 2022-05-24
Layered Drop Down Menu


jjbeckman13:
#### Feature
Layered Drop Down Menu

- context menu that can navigate to additional sub context menus
- each context menu could have an optional header that is fixed at the top of the menu that would allow navigating back to previous menu
- each context menu should be scrollable
- the main context menu should be able to display the selected choice of a sub-context menu
- context menu could have (see images below)
    - one or many settings that could be a toggle
    - selection
    -  count of sub menu settings enabled 
- [Miro Example of Navigation Paths of Existing Component](https://miro.com/app/board/uXjVOyOhU9c=/)

#### Examples

![image](https://user-images.githubusercontent.com/47185728/170059958-0ba021dd-5f03-45bd-a19e-a79b912d652a.png)

![image](https://user-images.githubusercontent.com/47185728/170058652-c8aef3bb-a9af-457b-9266-aee2fa5dd22c.png)

![image](https://user-images.githubusercontent.com/47185728/170058682-36486698-8b91-4b15-9f28-f4cd2bb2837b.png)

![image](https://user-images.githubusercontent.com/47185728/170060072-7daea872-3df0-4376-aa08-63e51506fdba.png)

FlyersPh9:
Another user today requested this swiping forward / backward option but in the information panel.  Could also be used in side navigation submenu.

jjbeckman13:
@elephantcatdog, Is there an ETA on when this component will be available?

elephantcatdog:
I'm going to say sometime in February. There's not much time to work on it until next year.

jjbeckman13:
https://bentleycs.visualstudio.com/beconnect/_git/OpenConstructionPlanner?path=%2Fsynchro-mono%2Fmodules%2Fsynchro-schedule%2Fsrc%2Fcomponents%2Ftoolbar%2FScheduleSettingsMenu.tsx&_a=contents&version=GBmaster 

![image](https://github.com/iTwin/iTwinUI/assets/47185728/80638be3-58ce-48ba-bf37-8bb4ea8bab71)


Updated Implementation @FlyersPh9

PR #1942, opened 2024-03-22
Nested `DropdownMenu` stays open at deeper levels

### LINKED ISSUE #1924: `DropdownMenu` close issue

Close handler of the `menuItems` doesn't seem to close all the dropdown menu levels in _some_ cases.

### Expected Behavior

All dropdown menu popovers should close if `menuItems` function `close` is called.

### Link to minimal repro

https://stackblitz.com/edit/github-gaoklu?file=src%2FApp.tsx

### Steps To Reproduce

1. Click `Menu`
2. Hover `Item 1_3`
3. Hover `Item 2_3`
4. Hover `Item 3_3`
5. Click `Item 4_1`

Observe that the popover that contains `Item 1_3` is still open.

### Anything else?

- Repeating the steps for the second time (1 - 5 and then 2 - 5) will close the popovers correctly.
- Uncommenting `Item 1_2` will also make popovers close correctly.
- Less sub menu levels also make popovers close correctly.

rohan-kadkol:
## Changes

Fixes #1924 

WIP

### DIFF
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -15,14 +15,29 @@ import { ListItem } from '../List/ListItem.js';
 import type { ListItemOwnProps } from '../List/ListItem.js';
 import { flushSync } from 'react-dom';
 import { usePopover } from '../Popover/Popover.js';
+import {
+  FloatingFocusManager,
+  FloatingNode,
+  FloatingTree,
+  useFloatingNodeId,
+  useFloatingParentNodeId,
+  useFloatingTree,
+} from '@floating-ui/react';
 
 /**
  * Context used to provide menu item ref to sub-menu items.
  */
 const MenuItemContext = React.createContext<{
   ref: React.RefObject<HTMLElement> | undefined;
+  isNestedSubmenuVisible: boolean | undefined;
   setIsNestedSubmenuVisible: React.Dispatch<React.SetStateAction<boolean>>;
-}>({ ref: undefined, setIsNestedSubmenuVisible: () => {} });
+  value: string | undefined;
+}>({
+  ref: undefined,
+  isNestedSubmenuVisible: undefined,
+  setIsNestedSubmenuVisible: () => {},
+  value: undefined,
+});
 
 export type MenuItemProps = {
   /**
@@ -116,15 +131,44 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   const parent = React.useContext(MenuItemContext);
 
   const onVisibleChange = (open: boolean) => {
-    setIsSubmenuVisible(open);
+    if (!open) {
+      console.log(
+        `${children}, onVisibleChange`,
+        open,
+        isNestedSubmenuVisible,
+        // parent.value,
+        // parent.isNestedSubmenuVisible,
+      );
+    }
+
+    setIsSubmenuVisible(open || isNestedSubmenuVisible);
+
+    if (children === 'Item 2_3' && !open) {
+      return;
+    }
 
     // we don't want parent to close when mouse goes into a nested submenu,
     // so we need to let the parent know whether the submenu is still open.
     parent.setIsNestedSubmenuVisible(open);
+
+    // await new Promise((resolve) => setTimeout(resolve, 1000));
   };
 
+  // if (["Item #10", "Item #11"].includes(children as string)) {
+  //   console.log(children, open, isNestedSubmenuVisible);
+
+  //   let currentNode = parent;
+  //   while (currentNode != null) {
+  //     console.log(currentNode.value, currentNode.isNestedSubmenuVisible);
+  //     currentNode = currentNode.parent;
+  //   }
+  // }
+
   const popover = usePopover({
+    // nodeId,
+    // visible: isSubmenuVisible,
     visible: isSubmenuVisible || isNestedSubmenuVisible,
+    // || children === 'Item #3',
     onVisibleChange,
     placement: 'right-start',
     trigger: { hover: true, focus: true },
@@ -146,6 +190,7 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
       case 'ArrowRight': {
         if (subMenuItems.length > 0) {
           setIsSubmenuVisible(true);
+          // setFocusOnSubmenu(true);
 
           // flush and reset state so we are ready to focus again next time
           flushSync(() => setFocusOnSubmenu(true));
@@ -178,9 +223,59 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   const handlers = {
     onClick: () => !disabled && onClick?.(value),
     onKeyDown,
+    // onClick: (event: React.MouseEvent<HTMLDivElement>) => {
+    //   // props.onClick?.(event);
+    //   !disabled && onClick?.(value);
+    //   tree?.events.emit('click');
+    // },
+    // onFocus: (event: React.FocusEvent<HTMLDivElement>) => {
+    //   // console.log('OnFocus', children, event);
+
+    //   props.onFocus?.(event);
+    //   setIsSubmenuVisible(false);
+    //   parent.setIsNestedSubmenuVisible(true);
+    //   // menu.setHasFocusInside(true);
+    // },
   };
 
+  // const item = useListItem({ label: disabled ? null : label });
+  // const tree = useFloatingTree();
+
+  // // Event emitter allows you to communicate across tree components.
+  // // This effect closes all menus when an item gets clicked anywhere
+  // // in the tree.
+  // React.useEffect(() => {
+  //   if (!tree) return;
+
+  //   function handleTreeClick() {
+  //     setIsSubmenuVisible(false);
+  //   }
+
+  //   function onSubMenuOpen(event: { nodeId: string; parentId: string }) {
+  //     if (event.nodeId !== nodeId && event.parentId === parentId) {
+  //       setIsSubmenuVisible(false);
+  //     }
+  //   }
+
+  //   tree.events.on('click', handleTreeClick);
+  //   tree.events.on('menuopen', onSubMenuOpen);
+
+  //   return () => {
+  //     tree.events.off('click', handleTreeClick);
+  //     tree.events.off('menuopen', onSubMenuOpen);
+  //   };
+  // }, [tree, nodeId, parentId]);
+
+  // React.useEffect(() => {
+  //   if (isSubmenuVisible && tree) {
+  //     tree.events.emit('menuopen', { parentId, nodeId });
+  //   }
+  // }, [tree, isSubmenuVisible, nodeId, parentId]);
+
+  // console.log('nodeid', nodeId);
+
   return (
+    // <FloatingNode id={nodeId}>
     <ListItem
       as='div'
       actionable
@@ -202,6 +297,19 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
       {...(subMenuItems.length === 0
         ? { ...handlers, ...rest }
         : popover.getReferenceProps({ ...handlers, ...rest }))}
+      // onFocus={(e) => {
+      //   console.log('onFocus', children, e);
+      //   e.bubbles = false;
+      //   // e.preventDefault();
+      // }}
+      // onMouseDown={(e) => {
+      //   // prevent focus on mousedown
+      //   e.preventDefault();
+      // }}
+      // onFocusCapture={(e) => {
+      //   // prevent focus on mousedown
+      //   e.preventDefault();
+      // }}
     >
       {startIcon && (
         <ListItem.Icon as='span' aria-hidden>
@@ -225,26 +333,42 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
 
       {subMenuItems.length > 0 && popover.open && (
         <Portal>
+          {/* <FloatingFocusManager
+              context={popover.context}
+              modal={false}
+              // initialFocus={popover.refs.floating}
+              // initialFocus={isNested ? -1 : 0}
+              // returnFocus={!isNested}
+            > */}
           <MenuItemContext.Provider
-            value={{ ref: menuItemRef, setIsNestedSubmenuVisible }}
+            value={{
+              ref: menuItemRef,
+              isNestedSubmenuVisible,
+              setIsNestedSubmenuVisible,
+              value: `${children}`,
+            }}
           >
-            <Menu
-              setFocus={focusOnSubmenu}
-              ref={popover.refs.setFloating}
-              {...popover.getFloatingProps({
-                id: submenuId,
-                onPointerMove: () => {
-                  // pointer might move into a nested submenu and set isSubmenuVisible to false,
-                  // so we need to flip it back to true when pointer re-enters this submenu.
-                  setIsSubmenuVisible(true);
-                },
-              })}
-            >
-              {subMenuItems}
-            </Menu>
+            <FloatingTree>
+              <Menu
+                setFocus={focusOnSubmenu}
+                ref={popover.refs.setFloating}
+                {...popover.getFloatingProps({
+                  id: submenuId,
+                  onPointerMove: () => {
+                    // pointer might move into a nested submenu and set isSubmenuVisible to false,
+                    // so we need to flip it back to true when pointer re-enters this submenu.
+                    setIsSubmenuVisible(true);
+                  },
+                })}
+              >
+                {subMenuItems}
+              </Menu>
+            </FloatingTree>
           </MenuItemContext.Provider>
+          {/* </FloatingFocusManager> */}
         </Portal>
       )}
     </ListItem>
+    // </FloatingNode>
   );
 }) as PolymorphicForwardRefComponent<'div', MenuItemProps>;
--- a/apps/react-workshop/src/DropdownMenu.stories.tsx
+++ b/apps/react-workshop/src/DropdownMenu.stories.tsx
@@ -200,14 +200,14 @@ export const Submenu = () => {
       subMenuItems={[
         <MenuItem
           key={4}
-          subMenuItems={[
-            <MenuItem key={7} onClick={onClick(7, close)}>
-              Item #7
-            </MenuItem>,
-            <MenuItem key={8} onClick={onClick(8, close)}>
-              Item #8
-            </MenuItem>,
-          ]}
+          // subMenuItems={[
+          //   <MenuItem key={7} onClick={onClick(7, close)}>
+          //     Item #7
+          //   </MenuItem>,
+          //   <MenuItem key={8} onClick={onClick(8, close)}>
+          //     Item #8
+          //   </MenuItem>,
+          // ]}
         >
           Item #4
         </MenuItem>,
@@ -217,10 +217,42 @@ export const Submenu = () => {
         <MenuItem
           key={6}
           subMenuItems={[
-            <MenuItem key={9} onClick={onClick(9, close)}>
+            <MenuItem
+              key={9}
+              onClick={onClick(9, close)}
+              // subMenuItems={[
+              //   <MenuItem
+              //     key={10}
+              //     onClick={onClick(10, close)}
+              //     subMenuItems={[
+              //       <MenuItem key={12} onClick={onClick(12, close)}>
+              //         Item #12
+              //       </MenuItem>,
+              //     ]}
+              //   >
+              //     Item #11
+              //   </MenuItem>,
+              // ]}
+            >
               Item #9
             </MenuItem>,
-            <MenuItem key={10} onClick={onClick(10, close)}>
+            <MenuItem
+              key={10}
+              onClick={onClick(10, close)}
+              subMenuItems={[
+                <MenuItem
+                  key={10}
+                  onClick={onClick(10, close)}
+                  subMenuItems={[
+                    <MenuItem key={12} onClick={onClick(12, close)}>
+                      Item #12
+                    </MenuItem>,
+                  ]}
+                >
+                  Item #11
+                </MenuItem>,
+              ]}
+            >
               Item #10
             </MenuItem>,
           ]}
--- a/packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
+++ b/packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
@@ -17,6 +17,7 @@ import type {
 } from '../utils/index.js';
 import { Menu } from '../Menu/Menu.js';
 import { usePopover } from '../Popover/Popover.js';
+// import { FloatingTree } from '@floating-ui/react';
 
 export type DropdownMenuProps = {
   /**
@@ -84,6 +85,8 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
   const triggerRef = React.useRef<HTMLElement>(null);
 
   const close = React.useCallback(() => {
+    console.log('close() called');
+
     setVisible(false);
     triggerRef.current?.focus({ preventScroll: true });
   }, [setVisible]);
@@ -97,7 +100,10 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
 
   const popover = usePopover({
     visible,
-    onVisibleChange: (open) => (open ? setVisible(true) : close()),
+    onVisibleChange: (open) => {
+      // console.log('onVisibleChange called', open);
+      open ? setVisible(true) : close();
+    },
     placement,
     matchWidth,
   });
@@ -128,7 +134,9 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
             })}
             ref={popoverRef}
           >
+            {/* <FloatingTree> */}
             {menuContent}
+            {/* </FloatingTree> */}
           </Menu>
         </Portal>
       )}
--- a/packages/itwinui-react/src/core/List/ListItem.tsx
+++ b/packages/itwinui-react/src/core/List/ListItem.tsx
@@ -30,6 +30,9 @@ const ListItemComponent = React.forwardRef((props, ref) => {
       data-iui-focused={focused ? 'true' : undefined}
       ref={ref}
       {...rest}
+      // style={{
+      //   width: '200px',
+      // }}
     />
   );
 }) as PolymorphicForwardRefComponent<'li', ListItemOwnProps>;
--- a/playgrounds/vite/src/App.tsx
+++ b/playgrounds/vite/src/App.tsx
@@ -1,9 +1,83 @@
-import { Button } from '@itwin/itwinui-react';
+import { Button, DropdownMenu, MenuItem } from '@itwin/itwinui-react';
 
 const App = () => {
   return (
     <>
-      <Button>Hello world</Button>
+      <DropdownMenu
+        menuItems={(close) => [
+          // <MenuItem onClick={close}>Item 1_2</MenuItem>,
+          <MenuItem
+            key='Item 1_3'
+            subMenuItems={[
+              <MenuItem key='Item 2_1' onClick={close}>
+                Item 2_1
+              </MenuItem>,
+              <MenuItem key='Item 2_2' onClick={close}>
+                Item 2_2
+              </MenuItem>,
+              <MenuItem
+                key='Item 2_3'
+                subMenuItems={[
+                  <MenuItem key='Item 3_1' onClick={close}>
+                    Item 3_1
+                  </MenuItem>,
+                  <MenuItem key='Item 3_2' onClick={close}>
+                    Item 3_2
+                  </MenuItem>,
+                  <MenuItem
+                    key='Item 3_3'
+                    subMenuItems={[
+                      <MenuItem key='Item 4_1' onClick={close}>
+                        Item 4_1
+                      </MenuItem>,
+                      <MenuItem
+                        key='Item 4_2'
+                        onClick={close}
+                        // subMenuItems={[
+                        //   <MenuItem key='Item 5_1' onClick={close}>
+                        //     Item 5_1
+                        //   </MenuItem>,
+                        //   <MenuItem key='Item 5_2' onClick={close}>
+                        //     Item 5_2
+                        //   </MenuItem>,
+                        //   <MenuItem
+                        //     key='Item 5_3'
+                        //     // onClick={() => {
+                        //     //   console.log('Item 5_3 clicked');
+                        //     //   close();
+                        //     // }}
+                        //     subMenuItems={[
+                        //       <MenuItem key='Item 6_1' onClick={close}>
+                        //         Item 6_1
+                        //       </MenuItem>,
+                        //       <MenuItem key='Item 6_2' onClick={close}>
+                        //         Item 6_2
+                        //       </MenuItem>,
+                        //       <MenuItem key='Item 6_3'>Item 6_3</MenuItem>,
+                        //     ]}
+                        //   >
+                        //     Item 5_3
+                        //   </MenuItem>,
+                        // ]}
+                      >
+                        Item 4_2
+                      </MenuItem>,
+                    ]}
+                  >
+                    Item 3_3
+                  </MenuItem>,
+                ]}
+              >
+                Item 2_3
+              </MenuItem>,
+            ]}
+          >
+            Item 1_3
+          </MenuItem>,
+        ]}
+      >
+        <Button>Menu</Button>
+      </DropdownMenu>
     </>
   );
 };


packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -115,16 +127,54 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+  }, [nodeId, parentId, tree?.events, tree?.nodesRef]);
+
   const popover = usePopover({
-    visible: isSubmenuVisible || isNestedSubmenuVisible,
+    nodeId,
+    visible:
+      isSubmenuVisible ||
+      isNestedSubmenuVisible ||
+      // to keep the submenu open when mouse enters it and then hovers out
+      dropdownMenuContext.lastHoveredNode?.parentId === nodeId,
```
mayank99, line 177: I don't think this is quite what we discussed. By "hover ≈ focus" I meant literally focusing an item (by calling `.focus()`) when it is hovered. Arrow-key navigation should continue from the last hovered item.

mayank99: > But triggering :focus-visible and then hovering over Item 4_2 does not do that.

If by "triggering `:focus-visible`" you mean in dev-tools, then it's usually not representative of real interactions. Usually you want to check both `:focus` and `:focus-visible`. And even then it is mainly useful for debugging styles (CSS), not behavior (JS). For proper debugging, you can use the same snippet mentioned in https://github.com/iTwin/iTwinUI/issues/1924#issuecomment-2014097457.

Nothing stands out to me in code, but if I had to guess, it's likely the `hover`+`focus` triggers in `usePopover` that's messing with it.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -159,92 +206,129 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
         event.preventDefault();
         break;
       }
       case 'Escape': {
-        // focus might get lost if submenu closes so move it back to parent
-        parent.ref?.current?.focus();
+        dropdownMenuContext.close();
+
+        event.stopPropagation();
+        event.preventDefault();
```
mayank99, line 220: I think this whole branch might be unnecessary if we use [`bubbles`](https://floating-ui.com/docs/usedismiss#bubbles)

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: The `setFocusOnSubmenu` logic could also be conflicting with it. Ideally we wouldn't need to use this hack at all.

rohan-kadkol: > By "hover ≈ focus" I meant literally focusing an item (by calling `.focus()`) when it is hovered

In [`4b29b10`](https://github.com/iTwin/iTwinUI/pull/1942/commits/4b29b1077625712e9997760558386df0b339f3d8), I removed the `useHover` trigger and but kept the `useFocus` trigger. Now, when we hover over a `MenuItem`, we call `.focus()` on itself inside the `onMouseEnter`. Since the `useFocus` trigger exists, hovering still indirectly opens a submenu.

Since "hover ≈ focus", one change is that earlier, opening a `DropdownMenu` where the first item has a submenu used to not open that submenu. But not since "hover ≈ focus", that submenu is opened. Not sure if this change in behavior is unideal.

We can see this when opening the leftmost menu button using a mouse click in the below recording:

<details>
<summary>Screen-recording</summary>

Video parts:
1. Second DropdownMenu mouse: no noticeable behavior change
2. First DropdownMenu mouse: `Item 1_3`'s submenu is not automatically open
3. Second DropdownMenu keyboard: no noticeable behavior change
4. First DropdownMenu keyboard: no noticeable behavior change

https://github.com/iTwin/iTwinUI/assets/45748283/aae38f59-907b-4f10-9946-f89209c03cf7

</details>

In the above-mentioned commit, I temporarily hard-coded [`visibleOnly`](https://floating-ui.com/docs/usefocus#visibleonly) to `false` in `Popover.tsx` to make this work. I will changing the hard coded value to be general if we decide to continue down this path.

> I had tried to call `.focus()` in `onMouseEnter` but kept getting unexpected behavior similar to this:

Like you suggested and I had alluded to before, the hover and focus triggers were for some reason not cooperating. The issue I mentioned is no longer present after removing the `onHover` trigger.

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
```diff
@@ -265,20 +265,19 @@ it('should handle key press with sub menus', async () => {
...
   await userEvent.keyboard('{ArrowLeft}');
+  expect(subMenuItem).not.toBeVisible();
   expect(menuItem).toHaveFocus();
 
   // escape to close
   await userEvent.keyboard('{Escape}');
   expect(screen.queryByTestId('sub')).toBeFalsy();
 
-  // go right to open sub menu
+  // going right to should re-open the sub menu and move the focus to the first item in the submenu.
```
AdamMeza-Bentley, line 277: Extra 'to'

```suggestion
  // going right should re-open the sub menu and move the focus to the first item in the submenu.
```

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
```diff
@@ -288,3 +287,52 @@ it('should handle key press with sub menus', async () => {
...
+      Test item
+    </MenuItem>,
+  );
+
+  const menuItem = screen.getByTestId('parent');
+
+  // focus to open sub menu
+  act(() => menuItem.focus());
+
+  // go right to move focus to the first submenu item. The first submenu's submenu should open.
```
AdamMeza-Bentley, line 330: Similar to the above comment. I don't know if the wording is intentional but it feels a bit clunky.

rohan-kadkol: We could try combining the two sentences:

```suggestion
  // go right to move focus to the first submenu item and open its submenu.
```

… but I'm not sure if it makes it clear that the submenu's submenu should open. i.e. what are we testing for.

Or we could split it on two lines.

```diff
- // go right to move focus to the first submenu item. The first submenu's submenu should open.
+ // Go right to move focus to the first submenu item.
+ // The first submenu's submenu should open.
await userEvent.keyboard('{ArrowRight}');
expect(screen.getByTestId('sub-1')).toHaveFocus();
expect(screen.getByTestId('sub-2')).toBeVisible();

- // hover over the second submenu item. The first submenu's submenu should close.
+ // Hover over the second submenu item.
+ // The first submenu's submenu should close.
fireEvent.mouseEnter(screen.getByTestId('sub-3'));
expect(screen.queryByTestId('sub-4')).toBeFalsy();
```

Or maybe do you have any other suggestions? Because I'm not quite sure I understood what you meant by "clunky".

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: The first submenu automatically opening when the main menu is opened is not acceptable. The user only meant to open the main menu. They should be expected to hover or arrow-right on the first item to open the submenu.

Ideally we'd be able to use [`focus({ focusVisible: true })`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#focusvisible), but it has poor browser support.

So I think we'll need to look into the heuristics of how `visibleOnly` is implemented in floating-ui to figure out how we can emulate focusVisible.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Ohh, nice, hardcoded `bubbles: true` in `Popover.tsx` to get the same behavior as this branch. (https://github.com/iTwin/iTwinUI/pull/1942/commits/e6b628bd4cae8a3bd855b1107e31032faf378434).

Then in [`271c4bc`](https://github.com/iTwin/iTwinUI/pull/1942/commits/271c4bc791722d5acddba972c39c36e0e6c46465), I added a new prop to `usePopover` to add `bubbles: true` only for `MenuItem` and not for other uses of `usePopover`.

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
AdamMeza-Bentley: I think the first option is good! I guess more specially the grammar seemed a bit off. But the first option is great

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: The prop is not needed. Menu is the only place we use `FloatingTree`, so there are no "other uses". And even if we add `FloatingTree` in more places, I'd argue we want the bubbles behavior in all of those places too.

The one exception is tooltips. I don't think that dismissing a tooltip using <kbd>esc</kbd> should close other popovers. But since Tooltip has its own `useFloating` call, I believe this should already work fine.

rohan-kadkol: One more case I felt where `bubbles: true` was not working well was in `DatePickerInput`. Because <kbd>esc</kbd> was closing the entire filter instead of just the `DatePicker`.

<details>
<summary>bubbles: true/false</summary>

`bubbles: true`

https://github.com/iTwin/iTwinUI/assets/45748283/3041105d-0983-4f1d-967f-fbefbf01b9ba

`bubbles: false`

https://github.com/iTwin/iTwinUI/assets/45748283/4f3dedf6-1a6b-448b-9513-afc6f3820b7e

</details>

> The one exception is tooltips. I don't think that dismissing a tooltip using esc should close other popovers. But since Tooltip has its own `useFloating` call, I believe this should already work fine.

Yes, I just confirmed that `bubbles: true` does not affect the behavior when a tooltip (e.g. from an `IconButton`) is in a `Dialog`.

<details>
<summary>Screen-recording</summary>

https://github.com/iTwin/iTwinUI/assets/45748283/bd171f76-4967-4e42-aca0-48cc14c89151

</details>

mayank99: Oh interesting, in that case I think it should be dependent on whether there is a FloatingTree.

If usePopover called in a component within a FloatingTree, then it should bubble. If not, then don't bubble.

rohan-kadkol: Ohh, sounds good. Removed `triggerProps` and now only bubbling when inside a `<FloatingTree>` scope. (https://github.com/iTwin/iTwinUI/pull/1942/commits/5dac2d3aacf9ebec50c301cfa18e72857a30937d)

packages/itwinui-react/src/core/Menu/MenuItem.test.tsx
rohan-kadkol: Great, combined the two sentences similar to the first option (https://github.com/iTwin/iTwinUI/pull/1942/commits/c507a95f33585f9c25a646d8b4f0ff42c82e9fbb).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: I noticed that maintaining an `isHovered` state variable (https://github.com/iTwin/iTwinUI/pull/1942/commits/5ac32c6643f3f23c93b5c7b12b15b3f92ad0d953) solves the issue of unexpectedly closing all submenus upon hovering over "Item 4_2" as mentioned in https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1556174966.

In the same commit, I also brought back the hover trigger. This helped bring back the [`safePolygon`](https://floating-ui.com/docs/usehover#safepolygon). Everything still continues to work.

> The first submenu automatically opening when the main menu is opened is not acceptable.

I also was able to undo setting `visisbleOnly: false` in `useHover` in `Popover.tsx`. So, "Item_3" in the leftmost DropdownMenu now no longer opens its submenu upon opening the DropdownMenu after clicking the "Menu" button. It only opens upon hovering "Item 1_3".

<details>
<summary>Screen-recording</summary>

https://github.com/iTwin/iTwinUI/assets/45748283/3c7b68d0-f355-4232-b7dd-f531b16d048e

</details>

mayank99: > > The first submenu automatically opening when the main menu is opened is not acceptable.
> 
> I also was able to undo setting `visisbleOnly: false` in `useHover` in `Popover.tsx`. So, "Item_3" in the leftmost DropdownMenu now no longer opens its submenu upon opening the DropdownMenu after clicking the "Menu" button. It only opens upon hovering "Item 1_3".

I think the root cause may not be fixed. Because the same problem can happen inside a nested menu when the first item has another submenu.

<details><summary>Screen recording</summary>

In here, the third level menu automatically opens when Item #4 is focused.

https://github.com/iTwin/iTwinUI/assets/9084735/fa6491e2-3ca4-449f-8b5f-37740e09fda4

</details> 

To avoid these complications, I'm now rethinking whether we should keep `focus` as the trigger.

Arrow keys are already used as the trigger, and in the recent UX call we also discussed using click/<kbd>Enter</kbd> as the trigger (#879). Together, these two triggers might make the `focus` trigger not as important.

If this change is too complicated, it can also happen in a separate follow-up PR.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -159,92 +206,123 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+  nodeId: string;
+  parentId: string | null;
+};
+
+/**
+ * Wraps `children` with `<FloatingTree>` if a `FloatingTree` context is not found.
+ *
+ * This is useful when `MenuItem`s are used outside of a `DropdownMenu`.
+ */
+const ConditionalFloatingTreeWrapper = (props: {
```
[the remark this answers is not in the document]
rohan-kadkol: I noticed that `MenuItem`s were used outside of `DropdownMenu` in the unit tests in `MenuItem.test.tsx`. So, I thought to have something like `ConditionalFloatingTreeWrapper` to add a `FloatingTree` when it does not exist.

Open to other ideas to implement this.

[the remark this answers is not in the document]
mayank99: > `MenuItem`s were used outside of `DropdownMenu` in the unit tests in `MenuItem.test.tsx`.

Why is this a problem? It should be fine.

Also we should almost never need to change our code for the purpose of making unit tests pass.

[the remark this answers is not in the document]
rohan-kadkol: > Why is this a problem? It should be fine.

We are currently relying on `FloatingTree` + `FloatingNode` to send events when a submenu opens. These events help us enforce only one submenu open at a time for each menu. Since tree events are not possible without a `FloatingTree` context as an ancestor, I added `ConditionalFloatingTreeWrapper`.

> Also we should almost never need to change our code for the purpose of making unit tests pass.

True, but I was trying to catch even non-unit test cases where `MenuItem`s could be outside the `DropdownMenu`.

If this case doesn't exist, then I can remove `ConditionalFloatingTreeWrapper` and update the unit tests accordingly.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: > In here, the third level menu automatically opens when `Item #4` is focused.

Ohh, I thought that was expected since we decided hover ≈ focus. As a result, since hovering over `Item #4` would open the third level submenu, focusing it also does the same.

> To avoid these complications, I'm now rethinking whether we should keep `focus` as the trigger.

Sounds good, this was my preference too because that is how the menus work in other UI libraries (e.g. [Radix UI](https://www.radix-ui.com/primitives/docs/components/dropdown-menu), [React Aria](https://react-spectrum.adobe.com/react-aria/Menu.html#submenus), [AriaKit](https://ariakit.org/examples/menu-nested)). Even the menu bar menus and context menus on my Mac work the same way (i.e. doesn't open the submenu on focus and only opens them using the <kbd>←</kbd>/<kbd>→</kbd> keys).

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/060fcda7-a378-495c-a205-5e7bb0c21d7a

</details>

I guess they might have been working differently (i.e. using focus as a trigger) on your Mac because of some change in your settings.

> If this change is too complicated, it can also happen in a separate follow-up PR.

Great, so, I will try working on removing the focus trigger and seeing how it works and also see whether a new PR would be better.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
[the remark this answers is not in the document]
mayank99: It might be better to test these interactions in a playwright test, with the whole component. Unit tests are a bad place to test such important, complex logic.

I don't think `MenuItem` should be used anywhere else. The only place I can think of is in the custom item renderers of `Select` and `ComboBox`, but those should use `ListItem` imo. We might want to look into showing a runtime warning to enforce this.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: > Ohh, I thought that was expected since we decided hover ≈ focus. As a result, since hovering over Item #4 would open the third level submenu, focusing it also does the same.

You can think of it as a one-directional relationship. Hovering should focus the item, but focusing by itself is its own separate thing that does not need to behave.

The more important thing to remember is the user's intent. Opening a menu automatically focuses the first menu-item, but this focus was not moved by the user. So they didn't intend to open the sub-menu. Compare this to a mouse user who needs to intentionally move their mouse from the trigger element into menu-item.

> Even the menu bar menus and context menus on my Mac work the same way … I guess they might have been working differently (i.e. using focus as a trigger) on your Mac because of some change in your settings.

I couldn't find any setting that I changed. It's very possible that the behavior was changed in macOS 14. I prefer it this way, but it very much needs to be done properly (which currently it is not).

> Great, so, I will try working on removing the focus trigger and seeing how it works and also see whether a new PR would be better.

If you intend to do a separate PR for the click trigger, then I would keep the focus trigger unchanged in this PR.

rohan-kadkol: > You can think of it as a one-directional relationship. Hovering should focus the item, but focusing by itself is its own separate thing that does not need to behave identically. 

Sounds good, with the recent changes, focusing a MenuItem does not open its submenu until <kbd>→</kbd> is pressed:

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/f2e8b202-b2fb-4678-8648-3e80420d4bcf

</details>

In contrast, hovering a MenuItem opens its submenu:

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/c0e1c690-a20a-4951-b7ba-0e7559e69255

</details>

> If you intend to do a separate PR for the click trigger, then I would keep the focus trigger unchanged in this PR.

Yes, I hope to work on the click trigger in a separate PR since that required more work than just adding the trigger.

I have currently removed the focus trigger to stop opening the submenu on focus, as mentioned in the point above. Is there any strong reason to keep the focus trigger in this PR instead of removing it?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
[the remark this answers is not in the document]
rohan-kadkol: > I don't think `MenuItem` should be used anywhere else.

For now, I removed `ConditionalFloatingTreeWrapper` since one more disadvantage of using it is that the top-level menu will not have one FloatingTree. Instead, each top-level menu's MenuItems will have a FloatingTree each. So, enforcing just one active submenu for the first level menu will not be possible with this approach.

> Or even deprecate `MenuItem` altogether and replace it with something like `DropdownMenu.Item`.

If we deprecate `MenuItem`, do we still need to support it until it is removed? If so, then the problem will exist for users that are using `MenuItem` instead of `DropdownMenu.Item`.

> The only place I can think of is in the custom item renderers of `Select` and `ComboBox`, but those should use `ListItem` imo. We might want to look into showing a runtime warning to enforce this. 

`Select` and `ComboBox` would not have submenu items. And the floating tree events are only for submenu management. So, even if the `<FloatingTree>` wrapper is not present, I don't think it would be a problem.

To be safer, should/can we add a `<FloatingTree>` inside `Select` and `ComboBox`, similar to how we're doing for `DropdownMenu`?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -107,27 +106,119 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+      parentId,
+      subMenuItems.length,
+      tree?.events,
+      // TODO: How to handle this warning?
+      // warning  React Hook React.useEffect has an unnecessary dependency: 'tree?.nodesRef.current'. Either exclude it or remove the dependency array. Mutable values like 'tree.nodesRef.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
+      tree?.nodesRef.current,
+      children,
+      tree?.nodesRef,
+    ],
+  );
```
rohan-kadkol, line 205: I removed the `focusOnSubmenu` variable used to focus the first item in the submenu when <kbd>→</kbd> was pressed. This was because it was causing focus issues after the changes from this PR. (Related: https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1557993045)

As a replacement, I used this `useEffect`. This `useEffect` emits the `arrowRightPressed` event after certain conditions are met instead of directly emitting the `arrowRightPressed` event right after <kbd>→</kbd> is pressed. This is because we need to wait for the `FloatingTree`'s context to be updated with the new `FloatingNode`s from the submenu since `handleArrowRightPressed` needs the up-to-date tree to properly focus the first item.

This `useEffect` feels a bit hacky. Also, I disabled eslint for the deps array; not sure if that's alright. So, open to suggestions/alternatives.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99, line 1: I'm a little confused by what happened here. Last time I looked at this PR (last week), I think it was in a pretty ok state with some small changes pending. But now, there's suddenly 100+ new lines of (hacky) code that don't feel right. What's the main difficulty here that is necessitating such code? Are we picking the wrong abstractions? Do we instead need to rewrite the DropdownMenu/Menu/MenuItem components in a way that makes it less difficult to achieve the behavior we desire?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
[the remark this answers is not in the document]
mayank99: > To be safer, should/can we add a `<FloatingTree>` inside `Select` and `ComboBox`, similar to how we're doing for `DropdownMenu`?

No, there's no need. I don't see how that's safer, since `ComboBox`/`Select` don't have submenus.

> If we deprecate `MenuItem`, do we still need to support it until it is removed? If so, then the problem will exist for users that are using `MenuItem` instead of `DropdownMenu.Item`.

Yes we need to support it, but that's not a difficult thing (I'm not sure what you mean by "the problem").

The point of a deprecation here is mainly to enforce correct usage. Currently if you use `MenuItem` inside `ComboBox`/`Select`, it gets incorrect semantics and needs to execute a bunch of unrelated code that gets thrown away anyway (because it is only relevant for `DropdownMenu`).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: > What's the main difficulty here that is necessitating such code?

When testing with edge cases, there were a few issues coming up. And it was difficult to fix them without replacing some of the previous workaround/hacky code (e.g. `focusOnSubmenu`) with something else. That's why I had to change my approach in order to make it work in the edge cases.

> Are we picking the wrong abstractions? Do we instead need to rewrite the DropdownMenu/Menu/MenuItem components in a way that makes it less difficult to achieve the behavior we desire?

I don't think rewriting it any more is necessary, at least at the moment since everything seems to be working well at the moment. So, for now, some alternatives to the hacky parts would be sufficient.

However, when working on the Layered Dropdown Menu from #653, a re-write might be needed; but not at the moment.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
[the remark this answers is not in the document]
rohan-kadkol: > I'm not sure what you mean by "the problem"

By "the problem", I meant that even if we deprecate `MenuItem` inside`Select`/`ComboBox`, if a user still uses a `MenuItem` there, enforcing only one active submenu per menu will not be possible without wrapping the entire tree in `FloatingTree`. But I guess that since submenus should anyways not be there in `Select`s/`ComboBox`es, "this program" will not happen.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Adding back `focus` takes away some of the benefits from this PR. E.g. pressing <kbd>←</kbd> on a menu "X" will move focus to X's parent but still keep X open instead of closing it.

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/05dffd03-bedf-4299-8cf8-fc633eed2094

</details>

> When both focus and click are absent, arrow keys are the only trigger, which is not sufficient.

To confirm, why are arrow keys and hover insufficient triggers?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: The rewrite _is_ "alternative to the hacky parts". We can't keep adding more hacky code on top of the existing hacky code\*; that's how we end up with an unmaintainable codebase.

\* I've made it clear multiple times that the way submenus have been implemented (before this PR) is super hacky and might require a rewrite to truly fix.

---

At the very least, could you try to abstract away the hairy parts into private hooks/components to make the main components easier to understand?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: > To confirm, why are arrow keys and hover insufficient triggers?

With complex patterns like submenus, there should always be multiple affordances. A screen-reader or keyboard user might not even realize that they need to press the arrow keys to open submenus.

In an ideal scenario, we would have focus+click+arrow-keys as triggers (+hover for mouse users). The next best thing is click+arrow-keys (+hover).

("click" also implies <kbd>Enter</kbd>/<kbd>Space</kbd> keypress)

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: > The rewrite _is_ "alternative to the hacky parts". We can't keep adding more hacky code on top of the existing hacky code*; that's how we end up with an unmaintainable codebase.
> 
> * I've made it clear multiple times that the way submenus have been implemented (before this PR) is super hacky and might require a rewrite to truly fix.

I guess I'm not sure about the scope of the rewrite we were hoping for. Because I did remove the hacky parts from before with improvements, of which only the useEffect that I mentioned in https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1565664450 seems hacky.

mayank99: Almost the entirety of `Menu`/`MenuItem` was and continues to be hacky.

In addition to the multiple `useEffect`s, the other parts that can be improved include:
- The code structure and the split of responsibilities between `Menu`/`DropdownMenu`/`MenuItem` (each relying on implementation details of the others without being properly abstracted)
- The `isSubmenuVisible || <something-else>` condition (kinda ironic that `isSubmenuVisible` isn't the source of truth)
- The `rightArrowPressed` state variable (this is probably the biggest issue; it doesn't make sense to store this in state)
- The whole `isAncestor` function being executed on every render (performance issue, but also possibly indicative of a missing intermediate abstraction)

rohan-kadkol: In the last few commits, I tried to make it less hacky. I used ideas from [floating-ui/floating-ui#1943](https://redirect.github.com/floating-ui/floating-ui/discussions/1943) to help with it.

There are still some loose ends, but here are some of the code improvements so far while still mostly maintaining the previous behavior:
* `isSubmenuVisible` is now the source of truth.
* `rightArrowPressed` and the associated `useEffect` are removed.
* `isAncestor` is removed.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Sounds good, added the click trigger manually instead of using `useClick` from floating UI (https://github.com/iTwin/iTwinUI/pull/1942/files/1b5f5f2f1f88375ebd8fe76b45887dc241d3ab54..a9d865f26b10f759df61210091fb0c0d00a16108).

So, the triggers now are hover, click (including <kbd>Enter</kbd>/<kbd>Space</kbd>), and arrow keys.

testing/e2e/app/routes/DropdownMenu/spec.ts
```diff
@@ -0,0 +1,62 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('DropdownMenu', () => {
+  test('should support deep level submenus', async ({ page }) => {
```
rohan-kadkol, line 4: This e2e test seems to be flaky. Any suggestions on how to make it pass reliably?

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -187,7 +190,7 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
       case 'Enter':
       case ' ':
       case 'Spacebar': {
-        !disabled && onClick?.(value);
+        onActivate();
```
mayank99, line 193: Can you remind me what was the problem with `useClick`?

_(Continued from [previous conversation](https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1572609858))_

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -217,8 +220,20 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
     }
   };
 
+  const onClick = (e: React.MouseEvent<HTMLElement>) => {
+    onActivate();
+
+    e.preventDefault();
+    e.stopPropagation();
```
mayank99, line 225: It would be useful to have a code comment for every instance of `preventDefault()` and `stopPropagation()`, because both of these are potentially destructive actions.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -98,32 +107,76 @@ type PopoverInternalProps = {
...
+      interactionsProps?: {
+        listNavigation?: UseListNavigationProps;
+      };
+    };
+
+type PopoverInteractionProps = PopoverInteractionConditionalProps & {
+  interactionsProps?: {
+    hover?: UseHoverProps<ReferenceType>;
+  };
 };
```
mayank99, line 141: You shouldn't need all these complex types. It would be enough for every interaction to accept an object in addition to a boolean

Usage:
```js
interactions: {
  hover: { move: false },
  click: true,
}
```

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -98,32 +107,76 @@ type PopoverInternalProps = {
...
-   * `hover` and `focus` can be manually specified as triggers.
-   */
-  trigger?: Partial<Record<'hover' | 'click' | 'focus', boolean>>;
   role?: 'dialog' | 'menu' | 'listbox';
   /**
    * Whether the popover should match the width of the trigger.
    */
   matchWidth?: boolean;
+} & PopoverInteractionProps &
+  Record<string, any>;
```
mayank99, line 116: nit: This `any` was meant to be a placeholder during our pair programming session. Can be replaced with the "correct" type now (if possible).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +190,143 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+        size={size}
+        active={isSelected}
+        disabled={disabled}
+        ref={useMergedRefs(
+          menuItemRef,
+          forwardedRef,
+          subMenuItems.length > 0 ? popover.refs.setReference : null,
+          (el) => {
+            if (parent != null && parentTreeIndex != null) {
+              parent.listItemsRef.current[parentTreeIndex] = el as HTMLElement;
```
mayank99, line 254: Can you try to explain in some more detail the new changes in this file since last week? I don't see them explained anywhere, and I'm very confused about how you're handling indexes.

Alternatively, I can reach out privately if you prefer to explain 1:1. After that we can try to improve the PR description and add some code comments.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +190,143 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+                  onPointerMove: () => {
+                    // pointer might move into a nested submenu and set isSubmenuVisible to false,
+                    // so we need to flip it back to true when pointer re-enters this submenu.
+                    setIsSubmenuVisible(true);
+                  },
+                })}
+              >
+                {subMenuItems.map((item, index) =>
+                  React.cloneElement(item, {
+                    key: index,
```
mayank99, line 314: You shouldn't need to set `key` here. `key` generally belongs to the user of the component.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +190,143 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+                    ['data-menu-child-index']: index,
```
mayank99, line 315: Two comments here:
1. Data attributes should be namespaced. Something like `data-iui-index` would suffice.
2. While it works for `MenuItem`, cloning elements like this is not super reliable. It would be better to use something like context. We could maybe save this improvement for when we deprecate `MenuItem` in favor of `DropdownMenu.Item` _([related conversation](https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1561066189))_.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -106,28 +116,69 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+    onVisibleChange: setIsSubmenuVisible,
     placement: 'right-start',
-    trigger: { hover: true, focus: true },
+    interactions: {
+      hover: !hasFocusedNodeInSubmenu, // If focus is still inside submenu, don't close the submenu upon hovering out.
+      listNavigation: true,
+    },
+    interactionsProps: {
+      hover: {
+        move: false,
```
mayank99, line 173: Would this be better suited as the default for `usePopover`? Do we ever need `move: true`?

testing/e2e/app/routes/DropdownMenu/spec.ts
mayank99: Have you tried debugging it using [`--ui` mode](https://playwright.dev/docs/test-ui-mode)?

Full command:
```
pnpm run build --filter=e2e && pnpm --filter=e2e run test --ui
```

After that, you could play with some delays. Playwright has multiples ways of introduces delays in various places ([example](https://playwright.dev/docs/api/class-keyboard#keyboard-press-option-delay)).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: After adding the click trigger, I noticed that it required two clicks to close a submenu. But turns out that it was expected behavior ([floating-ui/floating-ui#1893 (comment)](https://redirect.github.com/floating-ui/floating-ui/issues/1893#issuecomment-1250161220)).

I worked around this by manually toggling the visible state in `onClick()` ([code](https://github.com/iTwin/iTwinUI/pull/1942/commits/b3c54a164a3f9e4c2760a4623031d6ae63f9e4de#diff-4410321a0ef9f99d8d2c38b7907baf2e10f7fdf8efcff3a9ca5245db96d35db9R223)).

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: Replaced with proper type (https://github.com/iTwin/iTwinUI/pull/1942/commits/99f3b3a580c8aaffb556f85b6c331d32e72a7dad).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Ohh, I see, removed (https://github.com/iTwin/iTwinUI/pull/1942/commits/e85db420d91fa8db90562d4d3ac398a2d5389496).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: For now, I scoped the data attribute to `iui` by renaming it to `data-iui-index` ([`c3f0946`](https://github.com/iTwin/iTwinUI/pull/1942/commits/c3f09463b8bc7313883a36527d56506080cf2adc) + https://github.com/iTwin/iTwinUI/pull/1942/commits/8016f5bfcd8961c67ddb6a968094324157cd48e6).

I added an after PR TODO to try replacing the `cloneElement` with contexts.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: I thought to avoid that to prevent causing any behavior changes for users already used to `move: true`. Also, I'm not sure in which cases would `move: true` be needed. But since Floating UI has set that as the default, I guess there is some reason for it?

I can add an after PR TODO to address this comment later or I can make the change in this PR itself if we think it won't impact users.

mayank99: Since it's a very small change, I'm ok with doing it in the same PR. Just make sure to check that there is no adverse behavioral effect, and add a short changeset as well.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: I meant proper types for all arbitrary props (i.e. `UseFloatingProps`), rather than explicitly listing `nodeId`.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: Why is it so complicated with the merge helper?

Something like this should be enough, I think (untested):

```js
useClick(
  floating.context,
  { enabled: !!interactions.click, ...interactions.click }
),
// …
```

The `!!` will coerce the objects into a boolean; and the `...` will automatically merge the objects (correctly overriding the defaults, while also being a noop if a boolean is used).

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: Even though `usePopover` is private, I still thought it would be good to keep the parameters list small. I thought that we can add more props as and when we need them.

But I can add all `useFloating`'s props (`UseFloatingProps`) now too (https://github.com/iTwin/iTwinUI/pull/1942/commits/82850b197e18f1b7386fc73c62705498a6284187).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Sounds good, I made that change in this PR (https://github.com/iTwin/iTwinUI/pull/1942/commits/93e5a33eed228c1cb43da0b50da9e4db062dc30d).

Since we don't expose the hover trigger in the public `Popover` and since we use the private `usePopover`'s `hover` trigger only in `MenuItem`, I think the existing changeset is sufficient ([code](https://github.com/iTwin/iTwinUI/pull/1942/files#diff-89ddc481dba81a1393f7122d869df11bb687a93faec74c18833d53f9399b297aR5)).

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: Since `interactions.click` can also be a `boolean` it gives a type error.

image

I can ignore the type error if we're okay with it because like you said, applying spread on a boolean will be no-op in JS.

testing/e2e/app/routes/DropdownMenu/spec.ts
rohan-kadkol: Ohh, I see, thanks. The UI mode helped me debug the issue and delays were one of the things that fixed the flakiness (https://github.com/iTwin/iTwinUI/pull/1942/files/7a41f31b492eb3ab656ecb1f8b43bb8bdbc2ac32..9e471ef813e96edd4ef841d7fae70f1c18b6e999).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: `cloneElement` and data attributes were replaced with context in [`142a531`](https://github.com/iTwin/iTwinUI/pull/1942/commits/142a531fa9b3dc358340aec0485405d217557419) in order to avoid reading from refs (and DOM elements) during render.

The other part of my comment (deprecating `MenuItem`) can be kept as an after-PR TODO.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -154,15 +202,32 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
...
+      outsidePress: closeOnOutsideClick,
+      bubbles: tree != null, // Only bubble when inside a FloatingTree
+      ...(mergedInteractions.dismiss as object),
+    }),
     useHover(floating.context, {
-      enabled: !!trigger.hover,
+      enabled: !!mergedInteractions.hover,
       delay: 100,
-      handleClose: safePolygon({ buffer: 1 }),
+      handleClose: safePolygon({ buffer: 1, requireIntent: false }),
```
mayank99, line 218: While not directly relevant for this PR, I noticed safe polygon only works for the very last item in a menu. I investigated further and found that it starts working correctly when [`position: relative`](https://github.com/iTwin/iTwinUI/blob/e2e6ba7abbb46ac99ed225775b56942b88d8da04/packages/itwinui-css/src/menu/mixins.scss#L46) is removed. 


Further investigation needed. (I believe the `position: relative` is necessary for [`ListItem.Action`](https://itwinui.bentley.com/docs/list#with-listitemaction) to work as intended).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: Upon testing, I realized that `e.preventDefault()` and `e.stopPropagation()` were not needed in this case. So, I removed them (https://github.com/iTwin/iTwinUI/pull/1942/commits/5ed937835d330eef33e3294e8d3ce9fe2f416506).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +197,134 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
+  };
+
+  const onClick = () => {
+    onActivate();
+  };
+
+  const onActivate = () => {
+    !disabled && subMenuItems.length === 0 && onClickProp?.(value);
+    setIsSubmenuVisible((prev) => !prev);
+  };
```
mayank99, line 232: 1. Since `onActivate` and now identical, they can be combined.
2. We should call `onClickProp` even when `subMenuItems` is provided, in case users were relying on that behavior.
3. We should not call `setIsSubmenuVisible` on disabled items because #1925`.
   - Additionally, we might also want to disable all other triggers on disabled items to make it consistent.

rohan-kadkol: 1. Combined (https://github.com/iTwin/iTwinUI/pull/1942/commits/a9ab28f140a4e8d0bb3d63e08735697a093f900e).
2. Ohh, sounds good, updated (https://github.com/iTwin/iTwinUI/pull/1942/commits/72058bb581bccdfc6867fc3ec4b0190faa5cd193).
3. I thought we planned to handle disabled `MenuItem`s later, but yes, reading your comment on the issue you linked, it feels better to handle it here. Handled (https://github.com/iTwin/iTwinUI/pull/1942/files/72058bb581bccdfc6867fc3ec4b0190faa5cd193..c556efa6d4ab39f823f80930084462a98303b7e7). These commits also disable all Floating UI triggers/interactions and our event handlers for disabled items.

packages/itwinui-react/src/core/Menu/Menu.tsx
```diff
@@ -89,6 +89,16 @@ export const Menu = React.forwardRef((props, ref) => {
     }
   };
 
+  // Keep focusedIndex in sync with the real focused node's index.
+  const onFocus = (event: React.FocusEvent<HTMLDivElement, Element>) => {
+    const items = getFocusableNodes();
+    const currentIndex = items.findIndex((el) => el === event.target);
```
mayank99, line 95: Running `getFocusableNodes` and `findIndex` every time focus moves is expensive.

Suggestion to improve performance:
1. In `MenuItem`, set `data-iui-index={indexInParentTree}`.
2. In this function, read `e.target.dataset.index`.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: Why do you need to disable everything in so many places? We don't need all this extra code. It would be enough just _not_ render the submenu when the triggering menu-item is disabled.

rohan-kadkol: > It would be enough just _not_ render the submenu when the triggering menu-item is disabled.

I did it this way since I thought we wanted to keep `isSubmenuVisible` as the source of truth (https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1567940451). But I guess that discussion was only for non-disabled `MenuItem`s.

So, I replaced that code with the safer approach of not rendering the submenu when the triggering `MenuItem` is disabled (https://github.com/iTwin/iTwinUI/pull/1942/commits/6e39132e65d42d9f6c972c0d729ecb755a7067b3).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +201,139 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
-        break;
-      }
-      case 'Escape': {
-        // focus might get lost if submenu closes so move it back to parent
-        parent.ref?.current?.focus();
-        break;
-      }
-      default:
-        break;
     }
```
rohan-kadkol, line 223: @mayank99 and I had removed this in a previous commit (https://github.com/iTwin/iTwinUI/pull/1942/commits/3861266a09f6d06d6a5f79e6e77261b56b12f309). But I brought it back since I now remember why I had it in the first place. It was because if we don't update the active index upon hover/focus, then Floating UI's `useListNavigation` keeps the stale `activeIndex`.

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/29fb5fc2-3ce0-4761-991f-6800f6edca9b

</details>

One small difference is that instead of updating the parent's activeIndex in `onFocus`, I am now doing it in `onMouseEnter`, where it is really needed because that is where we're manually changing the focus instead of letting `useListNavigation` handle it.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -139,112 +201,139 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
 
-          event.preventDefault();
-          event.stopPropagation();
-        }
-        break;
+      // Since we manually focus the MenuItem on hover, we need to manually update the active index for
+      // Floating UI's keyboard navigation to work correctly.
+      if (parent != null && indexInParentTree != null) {
+        parent.setActiveIndex(indexInParentTree);
       }
```
rohan-kadkol, line 222: @mayank99 and I had removed this in a previous commit (https://github.com/iTwin/iTwinUI/pull/1942/commits/3861266a09f6d06d6a5f79e6e77261b56b12f309). But I brought it back since I now remember why I had it in the first place. It was because if we don't update the active index upon hover/focus, then Floating UI's `useListNavigation` keeps the stale `activeIndex`.

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/29fb5fc2-3ce0-4761-991f-6800f6edca9b

</details>

One small difference is that instead of updating the parent's activeIndex in `onFocus`, I am now doing it in `onMouseEnter`, where it is really needed because that is where we're manually changing the focus instead of letting `useListNavigation` handle it.

packages/itwinui-react/src/core/Menu/Menu.tsx
rohan-kadkol: I believe I had made this change before using [`useListNavigation`](https://floating-ui.com/docs/useListNavigation). Back then, I was instead relying on `Menu` for the <kbd>↑</kbd>/<kbd>↓</kbd> arrow keys navigation.

But since I'm now using `useListNavigation` for keyboard navigation, the code changes in `Menu.tsx` are no longer necessary. So, I reverted all changes in `Menu.tsx` (https://github.com/iTwin/iTwinUI/pull/1942/commits/ffb4fa269f78a91bb87a21047e2156009ba4971d).

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: I realized the I was not using the `listNavigation` interaction in the `Menu` in `DropdownMenu`. Thus, the previously mentioned issue existed in the first-level `Menu` in the `DropdownMenu`.

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/bba7f760-a8ab-410e-b1e4-4589b1213fc1

</details>

This is now fixed in [`8a1bc78`](https://github.com/iTwin/iTwinUI/pull/1942/commits/8a1bc78647fe9c7d49a26f54989084279c858360). Since there are many contexts, open to suggestions (if any) for context names. In a follow up PR, I can try using [`jotai`](https://jotai.org/) to reduce the number of contexts, similar to https://github.com/iTwin/iTwinUI/pull/2011.

mayank99: Does this change mean you can remove the keyboard navigation logic from `Menu`?

mayank99: >  Since there now are many contexts

I'm not sure I understand why the contexts needed to be split? `setHasFocusedNodeInSubmenu` could be made optional instead?

rohan-kadkol: > Does this change mean you can remove the keyboard navigation logic from `Menu`?

Yes, it can be removed if the children of `Menu` is always going to be a `MenuItem`. This is because `MenuItem` handles setting the parent's `listItemsRef` ([code](https://github.com/iTwin/iTwinUI/pull/1942/files#diff-4410321a0ef9f99d8d2c38b7907baf2e10f7fdf8efcff3a9ca5245db96d35db9R270-R275)) which is needed for Floating UI's`useListNavigation` to work.

But since `Menu` supports non-`MenuItem` children too, I guess some cloning will be required to add a `ref` function that updates `listItemsRef`. So, we would need to add this to all internal uses of the `Menu` component and add this cloning logic to the children inside the `Menu`.

Since some more investigation might be required, I think this can be done after this PR. If so, I can add this as an after PR TODO.

mayank99: > But since Menu supports non-MenuItem children too

This is a contrived case that will never happen.

> I think this can be done after this PR

I'm ok with this, but it should be done sooner rather than later, since we now have duplicated logic for keyboard navigation.

rohan-kadkol: > This is a contrived case that will never happen.

I see, then should we update the JSDocs in a few places? Because it gives the notion that non-`MenuItem` are fully supported/allowed. Examples:

https://github.com/iTwin/iTwinUI/blob/d1af8d045f7117476abc06467d98b314b92a88c5/packages/itwinui-react/src/core/Menu/Menu.tsx#L17

https://github.com/iTwin/iTwinUI/blob/d1af8d045f7117476abc06467d98b314b92a88c5/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx#L120

https://github.com/iTwin/iTwinUI/blob/d1af8d045f7117476abc06467d98b314b92a88c5/packages/itwinui-react/src/core/Select/Select.tsx#L514

> I'm ok with this, but it should be done sooner rather than later

Sounds good, added an after PR TODO with the prefix "**(Important)**" to prioritize it.

mayank99: `ComboBox` and `Select` (public) have completely different semantics from `Menu` (private), so they should not even be using `MenuItem` in the first place. It makes more sense to use [`ListItem`](https://itwinui.bentley.com/docs/list#interactive-states).

No jsdoc changes necessary in this PR, but here's what I think makes the most sense as a plan for near-future:
1. Remove redundant logic from `Menu` and abstract some parts of `DropdownMenu`/`MenuItem` into `Menu`.
2. Replace incorrect usages of `Menu` (mainly in `ComboBox` and `Select`) with a new private component (could be called `Listbox`).
3. Deprecate `MenuItem` and replace with something like `DropdownMenu.Item` (and maybe `ComboBox.Item`/`Select.Item`).
4. Update all documentation to show the correct usage of all components.

rohan-kadkol:
Update: I think I've addressed all review comments, all tests are now passing, and I have also updated the PR description based on the latest developments in this PR.

.changeset/fuzzy-socks-return.md
```diff
@@ -0,0 +1,7 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+* Clicking a `MenuItem` with a non empty `submenuItems` array now toggles the submenu visibility.
+  * Thus, passing a non empty `submenuItems` array and `onClick` to `MenuItem` at the same time is not supported.
+  * If both props are passed, clicking the `MenuItem` will toggle the submenu visibility but also *still* call the `onClick` (to prevent a behavioral change).
```
mayank99, line 7: "non-empty" is implied, and "supported" should be reworded as "recommended", and the warning should be mentioned.

also slight formatting nit around bullet point (since changesets will already add a bullet point).

```suggestion
Clicking a `MenuItem` with `submenuItems` now toggles the submenu visibility.
  * If both `submenuItems` and `onClick` props are passed, then clicking the `MenuItem` will toggle the submenu visibility but also *still* call the `onClick`. However, this behavior can lead to a confusing UX and is not recommended, so a warning will be shown.
```

mayank99 (approved):
LGTM.

I'm not a fan of the `useEffect` and `getFocusableNodes` in [`4f30668`](https://github.com/iTwin/iTwinUI/pull/1942/commits/4f3066820ef9ceb310eba6c5ba4a0c432260bf11), but I'm ok with it as a stopgap since all tests are now passing and our [immediate plan](https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1581201279) likely involves removing all this code.

PR #1943, opened 2024-03-22
Fix `Table` resizer z-index

rohan-kadkol:
## Changes

I noticed that the `iui-table-resizer` appears above the sticky header cells. This small PR fixes that.

<details>

[Before](https://itwin.github.io/iTwinUI/react/?story=table--full2&theme=dark):

https://github.com/iTwin/iTwinUI/assets/45748283/3548b793-9848-4257-b7b5-c56fc10c3b5d

[After](https://itwin.github.io/iTwinUI/1943/react/?story=table--full2&theme=dark):

https://github.com/iTwin/iTwinUI/assets/45748283/dcf90a1f-6430-4ee5-b657-89046d1340b7

</details>

## Testing

Confirmed that the table resizer now appears below the sticky header cells.

## Docs

Added changesets.

## After PR TODOs:

* [ ] The table resizer in sticky columns gets cut off (https://github.com/iTwin/iTwinUI/pull/1943#discussion_r1537932830)

### DIFF
--- a/packages/itwinui-css/src/table/base.scss
+++ b/packages/itwinui-css/src/table/base.scss
@@ -66,6 +66,7 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
+      z-index: 1;
       opacity: 0;
 
       > .iui-table-resizer-bar {
@@ -417,7 +418,7 @@
 
   &.iui-table-cell-sticky {
     position: sticky;
-    z-index: 1;
+    z-index: 2;
     inset-inline-start: var(--iui-table-sticky-left, initial);
     inset-inline-end: var(--iui-table-sticky-right, initial);
   }
--- a/packages/itwinui-css/src/table/base.scss
+++ b/packages/itwinui-css/src/table/base.scss
@@ -17,14 +17,14 @@
     position: sticky;
     inset-block-end: 0;
     inset-inline-start: 0;
-    z-index: 2; // Above the table body and sticky columns/cells
+    z-index: 3; // Above the table body and sticky columns/cells
   }
 }
 
 @mixin iui-table-header-wrapper {
   position: sticky;
   inset-block-start: 0;
-  z-index: 2; // Above the table body and sticky columns/cells
+  z-index: 3; // Above the table body and sticky columns/cells
 
   display: flex;
   flex-shrink: 0;
--- a/packages/itwinui-css/src/table/base.scss
+++ b/packages/itwinui-css/src/table/base.scss
@@ -66,7 +66,6 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
-      z-index: 1;
       opacity: 0;
 
       > .iui-table-resizer-bar {
--- a/packages/itwinui-css/src/table/base.scss
+++ b/packages/itwinui-css/src/table/base.scss
@@ -418,7 +418,7 @@
 
   &.iui-table-cell-sticky {
     position: sticky;
-    z-index: 2;
+    z-index: 2; // Above the non-sticky columns and the table resizer
     inset-inline-start: var(--iui-table-sticky-left, initial);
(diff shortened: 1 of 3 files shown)


AdamMeza-Bentley (approved):
good catch!

packages/itwinui-css/src/table/base.scss
```diff
@@ -66,7 +66,6 @@
       transform: translateX(50%);
       touch-action: none;
       cursor: ew-resize;
-      z-index: 1;
```
mayank99, line 69: We need this z-index to prevent the resizer getting clipped by the next column (which appears later in DOM order).

For sticky columns, you could add `z-index: 2` to make it appear on top of the resizers.

rohan-kadkol: Ohh, I see. Then I bumped the z-index of `.iui-table-sticky-cell` instead and confirmed that it works (https://github.com/iTwin/iTwinUI/pull/1943/commits/770fa884c8c9c98e5274a026bb10b4c55e275131).

mayank99: Can you also add a comment to explain the magic number?

mayank99 (dismissed):

packages/itwinui-css/src/table/base.scss
rohan-kadkol: Sounds good, added (https://github.com/iTwin/iTwinUI/pull/1943/commits/5ce368724664efb76a181a754169b2d26777608e).

I also realized that it would be better to bump other z-indexes (paginator and the header wrapper) as a result of bumping the sticky column z-index. So, I made those changes too (https://github.com/iTwin/iTwinUI/pull/1943/commits/37891c36d3d770ca16c2139d1ec4e5584c7121dc).

packages/itwinui-css/src/table/base.scss
```diff
@@ -418,7 +418,7 @@
 
   &.iui-table-cell-sticky {
     position: sticky;
-    z-index: 1;
+    z-index: 2; // Above the non-sticky columns and the table resizer
```
mayank99, line 421: I just noticed that this doesn't work for resizers inside sticky columns.

mayank99: Screen recording shows how you can hover near left edge of the next non-sticky column to resize.

https://github.com/iTwin/iTwinUI/assets/9084735/02487d9c-7ee2-4761-a85e-1fca69141e52

rohan-kadkol: Actually, that issue is not originating from the changes in this PR. This issue can be seen in the currently hosted [v3 Full2 story](https://itwin.github.io/iTwinUI/react/?story=table--full2&theme=dark) and also in the [v2 Full2 story](https://itwin.github.io/iTwinUI/legacy/v2/react/?path=/story/core-table--full-2).

I will try fixing it in this PR if it's a small fix. But if it's a larger fix, it may be better to have a new PR for that. Investigating it further.

mayank99: Maybe we can try something like:
1. increase the resizer's z-index to be higher than sticky columns
2. decrease the column and resizer's z-indices when the column is being dragged (untested)

Not sure about 2 but I think 1 is definitely needed, because otherwise the resizer's hit target is like 8px (rather than 16px), which makes it difficult to use.

mayank99: Another suggestion: changing `:hover` selectors to `:hover:not(:active)` to avoid triggering hover styles when dragging.

mayank99 (approved):
Re-approving since there is no regression. Further improvements can come in future PRs.

packages/itwinui-css/src/table/base.scss
rohan-kadkol: As discussed, I will work on this in a future PR. Added this to the After PR TODOs.

PR #1944, opened 2024-03-22
Tooltips don't keep being displayed upon hovering out after a click

### LINKED ISSUE #1931: `IconButton` label is displayed after button is clicked

Clicking an `IconButton` and moving the cursor away keeps the button label visible.

### Expected Behavior

Button label should not be visible (unless button is explicitly focused).

### Link to minimal repro

https://itwin.github.io/iTwinUI/react/?story=iconbutton--add

### Steps To Reproduce

1. Click on `+` button
2. Move cursor away
3. Observe that the label stays open

### Anything else?

Label is hidden again after entering/leaving button with a cursor.

rohan-kadkol:
## Changes

Fixes #1931.

In `Tooltip.tsx`: 
* Removed the `useClick()` trigger in favor of adding `referencePress: true` and `referencePressEvent: "click"` in `useDismiss()`. This was done to prevent the tooltip from being shown after hovering out of a trigger after clicking the trigger. ([#1944 (comment)](https://github.com/iTwin/iTwinUI/pull/1944#discussion_r1537544635))
  * So now, just one mouse or keyboard click will dismiss the tooltip. The reasoning for this was that once the trigger is click/action is taken, the helper tooltip text will no longer be needed. ([#1944 (comment)](https://github.com/iTwin/iTwinUI/pull/1944#discussion_r1537981952))
* Also added `move: false` in `useHover` to prevent mouse movements inside the trigger region from causing the tooltip to reappear even after clicking the trigger region once (since clicking a trigger region should dismiss the tooltip). ([#1944 (comment)](https://github.com/iTwin/iTwinUI/pull/1944#discussion_r1538264536)).
  * To reshow the tooltip after clicking the trigger, users will need to hover/focus out and then back in the trigger.

## Testing

Confirmed that the tooltip no longer appears upon hovering out of the reference element after clicking it.

## Docs

Added changeset.

## After PR TODOs:

* [ ] Slider Tooltips behavior (https://github.com/iTwin/iTwinUI/pull/1944#pullrequestreview-1960430808)

### DIFF
--- a/.changeset/famous-hats-shout.md
+++ b/.changeset/famous-hats-shout.md
@@ -1,5 +1,5 @@
 ---
-"@itwin/itwinui-react": minor
+"@itwin/itwinui-react": patch
 ---
 
-`Tooltip`s no longer displayed upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
+`Tooltip`s no longer displayed upon clicking the trigger. They are now only shown upon hover or focus. This fixes issues like incorrectly displaying the `content` upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
--- a/.changeset/famous-hats-shout.md
+++ b/.changeset/famous-hats-shout.md
@@ -2,4 +2,6 @@
 "@itwin/itwinui-react": patch
 ---
 
-`Tooltip`s no longer displayed upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
+Fixed an issue where `Tooltip` would stay open after clicking on the trigger element. The tooltip will now be correctly dismissed upon click.
+
+Also affects other components that use tooltips internally: `IconButton`, `SideNavigation`, etc.
--- a/.changeset/famous-hats-shout.md
+++ b/.changeset/famous-hats-shout.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-`Tooltip`s no longer displayed upon clicking the trigger. They are now only shown upon hover or focus. This fixes issues like incorrectly displaying the `content` upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
+`Tooltip`s no longer displayed upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
--- a/.changeset/famous-hats-shout.md
+++ b/.changeset/famous-hats-shout.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+`Tooltip`s no longer displayed upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
--- a/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
+++ b/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
@@ -10,7 +10,6 @@ import {
   offset,
   flip,
   shift,
-  useClick,
   useHover,
   useFocus,
   useDismiss,
@@ -175,9 +174,6 @@ const useTooltip = (options: TooltipOptions = {}) => {
       handleClose: safePolygon({ buffer: -Infinity }),
     }),
     useFocus(floating.context),
-    useClick(floating.context, {
-      ignoreMouse: true,
-    }),
     useDismiss(floating.context),
   ]);
 
--- a/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
+++ b/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
@@ -10,6 +10,7 @@ import {
   offset,
   flip,
   shift,
+  useClick,
   useHover,
   useFocus,
   useDismiss,
@@ -174,6 +175,9 @@ const useTooltip = (options: TooltipOptions = {}) => {
       handleClose: safePolygon({ buffer: -Infinity }),
     }),
     useFocus(floating.context),
+    useClick(floating.context, {
+      ignoreMouse: true,
+    }),
     useDismiss(floating.context),
   ]);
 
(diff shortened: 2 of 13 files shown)


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -175,7 +175,9 @@ const useTooltip = (options: TooltipOptions = {}) => {
       handleClose: safePolygon({ buffer: -Infinity }),
     }),
     useFocus(floating.context),
-    useClick(floating.context),
+    useClick(floating.context, {
```
mayank99, line 178: Do we even need `useClick` here? What exactly is it doing for us?

rohan-kadkol: I tried to not remove it as far as possible because I assumed that it was there for a reason. But I feel having just hover and focus as triggers are sufficient in this case.

So, I removed `useClick` (https://github.com/iTwin/iTwinUI/pull/1944/commits/66d1900841727d5e4fdd8d0da2b1c87d0c8b8c2a). Removing `useClick` seems to be giving the same behavior as `useClick({ ignoreMouse: true })`, considering that the focus and hover triggers are also present and unchanged.

.changeset/famous-hats-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
```
mayank99, line 2: Should be patch?

rohan-kadkol: I thought it should be `minor` since when the tooltip is shown has changed. But I can rephrase the changeset message to be like a `fix` if we feel this PR is more of a bug fix than a change in the behavior.

mayank99: Yeah I think this is more of a small bug fix.

`minor` should be used for new features, API changes, and larger refactorings (like the Table scroll container change in #1891).

rohan-kadkol: Sounds good, changed to `patch`. Also reworded the message to be a patch changeset. (https://github.com/iTwin/iTwinUI/pull/1944/commits/a820fcf0bd927129a8d627e9901a0780cd6279cf)

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: Actually I just realized what `useClick` is doing for us: it's hiding the tooltip when the button is clicked.

So it's not needed for the purposes of _triggering_ (showing) the tooltip on click, but it's needed for hiding it.

This makes sense from a UX perspective — the user will first hover (or focus) the button to understand what it does, and then they will click (or press <kbd>Enter</kbd>) to trigger the action. At this point, they no longer need to see the tooltip.

rohan-kadkol: `referencePress` seems to dismiss upon click but not upon focus + <kbd>space</kbd>. `useClick({ ignoreMouse: true })` + `useDismiss({})` dismisses upon focus + <kbd>space</kbd> but not upon click.

This is why in [`f365335` (#1944)](https://github.com/iTwin/iTwinUI/pull/1944/commits/f3653355296b6710890befc2bd10825227bdde9a), I used both `ignoreMouse` in `useClick` and `referencePress` in `useDismiss`.

The issue now is that pressing <kbd>space</kbd> repeatedly upon focusing toggles the tooltip. But clicking repeatedly just closes the tooltip upon the first click and does not bring it back the second click onwards. Not sure if this is an issue:

https://github.com/iTwin/iTwinUI/assets/45748283/444f2134-ab35-40e1-9ebd-6137a772295a

mayank99: [floating-ui docs](https://floating-ui.com/docs/usedismiss#keyboard-press-dismissal) suggest adding a custom `onClick` (which is called on pressing <kbd>Enter</kbd>/<kbd>Space</kbd> when used on `<button>`). We could do that to avoid toggling repeatedly.

mayank99: Alternatively we could just change [`referencePressEvent`](https://floating-ui.com/docs/usedismiss#referencepressevent) to `click`?

rohan-kadkol: Earlier, in [`f365335` (#1944)](https://github.com/iTwin/iTwinUI/pull/1944/commits/f3653355296b6710890befc2bd10825227bdde9a), I used both `ignoreMouse` in `useClick` and `referencePress` in `useDismiss`.

The issue it gave was that pressing <kbd>space</kbd> repeatedly upon focusing toggles the tooltip. But clicking repeatedly just closes the tooltip upon the first click and does not bring it back the second click onwards.

https://github.com/iTwin/iTwinUI/assets/45748283/444f2134-ab35-40e1-9ebd-6137a772295a

But now, in [`1ae6d27` (#1944)](https://github.com/iTwin/iTwinUI/pull/1944/commits/1ae6d27b904fe4869dca1a5a8c5260a4816dc9eb), using `useDismiss({ referencePress: true, referencePressEvent: "click" })` + removing `useClick()` seems to give consistent behavior regardless of mouse or keyboard clicks.

So now, clicking just once using a mouse or a keyboard dismisses the tooltip.

https://github.com/iTwin/iTwinUI/assets/45748283/11d62b33-fd00-4544-a434-7fe9121e89ed

I also added `move: false` in `useHover` to prevent mouse movements within the trigger region from re-triggering the tooltip.

<details>

Without `move: false`:

https://github.com/iTwin/iTwinUI/assets/45748283/5ea5311d-3f23-4b08-a23b-93a881a0bebc

With `move: false`:

https://github.com/iTwin/iTwinUI/assets/45748283/aff213de-735d-4bda-b979-fa1c0d0307dd

</details>

mayank99 (approved):
LGTM.

Before merging, it would be nice to update PR description to reflect latest code

mayank99 (commented):
(Unrelated to this PR) I noticed weird behavior in the tooltips in `Slider`. The tooltip should stay open during dragging, whereas currently it closes when you hover out.


https://github.com/iTwin/iTwinUI/assets/9084735/9d939511-3775-4851-8d77-e0a65eae6593

.changeset/famous-hats-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+`Tooltip`s no longer displayed upon hovering out of the reference element after clicking it. Components that use `Tooltip` internally are also affected. (E.g.`IconButton`, `SideNavigation`, etc.)
```
rohan-kadkol, line 5: Just wanted to confirm before merging this, should the changeset mention that clicks no longer trigger the tooltip, or that clicking a trigger once now dismisses the tooltip and cannot be toggled again until re-hovered? But not sure if it will still remain a patch then.

mayank99: You can phrase it like this:

```suggestion
Fixed an issue where `Tooltip` would stay open after clicking on the trigger element. The tooltip will now be correctly dismissed upon click.

Also affects other components that use tooltips internally: `IconButton`, `SideNavigation`, etc.
```

PR #1948, opened 2024-03-25
create Badge component for css-workshop

mayank99:
## Changes

- Created reusable `Badge.astro` with props for backgroundColor and status.
- Used it in main `badge.astro` (page).
  - This required renaming the imported component to disambiguate between `badge.astro` and `Badge.astro`. (Workaround for bug in Astro)
- Used it in other pages: `tile.astro`

## Testing

All existing tests should pass.

## Docs

N/A

### DIFF
--- a/apps/css-workshop/src/components/Badge.astro
+++ b/apps/css-workshop/src/components/Badge.astro
@@ -1,0 +1,32 @@
+---
+type Props = {
+  backgroundColor?:
+    | 'skyblue'
+    | 'celery'
+    | 'froly'
+    | 'steelblue'
+    | 'sunglow'
+    | 'seabuckthorn'
+    | 'montecarlo'
+    | 'poloblue'
+    | 'bouquet'
+    | 'ash'
+    | 'oak';
+  status?: 'informational' | 'positive' | 'warning' | 'negative';
+} & astroHTML.JSX.HTMLAttributes;
+
+const { class: className, backgroundColor, status, ...props } = Astro.props;
+---
+
+<span
+  class:list={['iui-badge', className]}
+  style={{
+    '--iui-badge-background-color': backgroundColor
+      ? `var(--iui-color-background-${backgroundColor})`
+      : undefined,
+  }}
+  data-iui-status={status}
+  {...props}
+>
+  <slot />
+</span>
--- a/apps/css-workshop/src/pages/badge.astro
+++ b/apps/css-workshop/src/pages/badge.astro
@@ -1,5 +1,6 @@
 ---
 import Layout from './_layout.astro';
+import IuiBadge from '../components/Badge.astro';
 ---
 
 <Layout title='Badge'>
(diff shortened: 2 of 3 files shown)


apps/css-workshop/src/components/Badge.astro
```diff
@@ -0,0 +1,32 @@
...
+    | 'seabuckthorn'
+    | 'montecarlo'
+    | 'poloblue'
+    | 'bouquet'
+    | 'ash'
+    | 'oak';
+  status?: 'informational' | 'positive' | 'warning' | 'negative';
+} & astroHTML.JSX.HTMLAttributes;
+
+const { class: className, backgroundColor, status, ...props } = Astro.props;
```
mayank99, line 18: - Astro provides all externally passed as `Astro.props` ([See docs](https://docs.astro.build/en/basics/astro-components/#component-props)). We can define a type/interface for it using `type Props`, but it's not strictly necessary, because these components are for internal use. I only added the `type` to demonstrate it.
- The `{}` syntax is used for ["destructuring"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) specific properties from `Astro.props`. The `...props` is a catch-all syntax for any props that we don't want to manipulate directly.
- `class: className` is basically renaming the variable to `className` when destructuring, because `class` is a reserved JavaScript keyword.

apps/css-workshop/src/components/Badge.astro
```diff
@@ -0,0 +1,32 @@
...
+---
+
+<span
+  class:list={['iui-badge', className]}
```
mayank99, line 22: Here we are "merging" classes, so that any externally passed class names are correctly applied.

e.g. 
```
<Badge class="my-badge">
```
will output
```
<span class="iui-badge my-badge">
```

apps/css-workshop/src/components/Badge.astro
```diff
@@ -0,0 +1,32 @@
...
+  style={{
+    '--iui-badge-background-color': backgroundColor
+      ? `var(--iui-color-background-${backgroundColor})`
+      : undefined,
```
mayank99, line 26: This is just a [ternary](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator) expression for conditionally applying the inline CSS variable style only when `backgroundColor` is ["truthy"](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)

apps/css-workshop/src/components/Badge.astro
```diff
@@ -0,0 +1,32 @@
...
+  }}
+  data-iui-status={status}
+  {...props}
```
mayank99, line 29: This is using the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) for any delegated props that we do not directly care about in this component but may want to pass from "outside".

```
<Badge data-foo>
```
will output
```
<span class="iui-badge" data-foo>
```

apps/css-workshop/src/components/Badge.astro
```diff
@@ -0,0 +1,32 @@
...
+>
+  <slot />
```
mayank99, line 31: This allows placing content inside the badge. See docs for [slot](https://docs.astro.build/en/basics/astro-components/#slots).

apps/css-workshop/src/pages/badge.astro
```diff
@@ -1,5 +1,6 @@
 ---
 import Layout from './_layout.astro';
+import IuiBadge from '../components/Badge.astro';
```
mayank99, line 3: This is the error I get when using `Badge` inside `badge.astro`:

image

I don't like `IuiBadge`, it feels kinda ugly/long. Open to suggestions.

mayank99 (commented):
@FlyersPh9 @Ben-Pusey-Bentley @AdamMeza-Bentley Here are some comments explaining the basic structure of this component. Feel free to reference this when adding more components, and let me know if you have any questions.

apps/css-workshop/src/pages/badge.astro
AdamMeza-Bentley: maybe `AstroBadge` or `BadgeAst` ?

apps/css-workshop/src/components/Badge.astro
AdamMeza-Bentley: This was all really helpful! Thanks for taking the time to explain everything

apps/css-workshop/src/pages/badge.astro
mayank99: Astro is just the framework we are using today. The contents of the component have nothing to do with Astro.

Maybe `Badge_` could work?

image
image

rohan-kadkol (approved):

PR #1954, opened 2024-03-26
add `adoptedStyleSheets` fallback for older safari

mayank99:
## Changes

Slight refactoring of `ShadowRoot`, but the gist is that when `adoptedStyleSheets` is not supported, an inline `<style>` tag will be used instead. Fixes https://github.com/iTwin/iTwinUI/issues/1930#issuecomment-2020360095

All the `attachShadowRef` has been moved into a `useShadowRoot` hook, and all the client code itself has been moved into a separate local component. This hopefully makes it easier to maintain.

Also, I fixed some issues I noticed in the existing logic and made it more resilient to work with existing shadow-roots, re-renders, and popout windows.

## Testing

To emulate older Safari behavior, I manually set `supportsAdoptedStylesheets` to `false` and verified that 

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
+++ b/packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
+import { useLatestRef, useLayoutEffect } from '../hooks/index.js';
 
 const isBrowser = typeof document !== 'undefined';
 const supportsDSD =
@@ -24,34 +25,7 @@ export const ShadowRoot = ({
   children: React.ReactNode;
   css?: string;
 }) => {
-  const [shadowRoot, setShadowRoot] = React.useState<ShadowRoot>();
   const isFirstRender = useIsFirstRender();
-  const styleSheet = React.useRef<CSSStyleSheet>();
-
-  const attachShadowRef = React.useCallback(
-    (template: HTMLTemplateElement | null) => {
-      const parent = template?.parentElement;
-      if (!template || !parent) {
-        return;
-      }
-      if (parent.shadowRoot) {
-        parent.shadowRoot.replaceChildren(); // Remove previous shadowroot content
-      }
-      queueMicrotask(() => {
-        const shadow =
-          parent.shadowRoot || parent.attachShadow({ mode: 'open' });
-
-        if (css && supportsAdoptedStylesheets) {
-          styleSheet.current ||= new CSSStyleSheet();
-          styleSheet.current.replaceSync(css);
-          shadow.adoptedStyleSheets = [styleSheet.current];
-        }
-
-        ReactDOM.flushSync(() => setShadowRoot(shadow));
-      });
-    },
-    [css],
-  );
 
   if (!isBrowser) {
     return (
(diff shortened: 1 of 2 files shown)


packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -68,19 +38,90 @@ export const ShadowRoot = ({
...
+
+    return () => void setShadowRoot(null);
+  }, [templateRef, latestCss]);
+
+  // Synchronize the CSS contents of existing stylesheet
+  useLayoutEffect(() => {
+    if (css && supportsAdoptedStylesheets) {
+      styleSheet.current?.replaceSync(css);
+    }
+  }, [css]);
```
rohan-kadkol, line 118: Since `latestCss` is listed as a dependency in the above `useLayoutEffect`, will the above `useLayoutEffect` handle what this `useLayoutEffect` is trying to handle?

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -68,19 +38,90 @@ export const ShadowRoot = ({
...
+
+      // add the CSS immediately to avoid FOUC (one-time)
+      if (latestCss.current) {
+        styleSheet.current.replaceSync(latestCss.current);
+      }
+    }
+
+    queueMicrotask(() => {
+      ReactDOM.flushSync(() => setShadowRoot(shadow));
+    });
```
rohan-kadkol, line 108: A small comment explaining why this is wrapped in a `flushSync`/`queueMicrotask` would be helpful.

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -68,19 +38,90 @@ export const ShadowRoot = ({
...
+    if (parent.shadowRoot) {
+      parent.shadowRoot.replaceChildren(); // Remove previous shadowroot content
+    }
+
+    const shadow = parent.shadowRoot || parent.attachShadow({ mode: 'open' });
+
+    if (supportsAdoptedStylesheets) {
+      // create an empty stylesheet and add it to the shadowRoot
+      const _global = shadow.ownerDocument.defaultView || globalThis;
+      styleSheet.current = new _global.CSSStyleSheet();
```
rohan-kadkol, line 97: What is the benefit of doing `new _global.CSSStyleSheet()` instead of just `new CSSStyleSheet()`?

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: nope, it's a ref

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: it fixes popout window support.

see https://github.com/iTwin/iTwinUI/pull/1625

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: this good?

```suggestion
    queueMicrotask(() => {
      // Flush the state immediately to prevent flash of unslotted light DOM
      ReactDOM.flushSync(() => setShadowRoot(shadow));
    });
```

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
rohan-kadkol: Ohh, I see. A small comment explaining this might be helpful here too.

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
rohan-kadkol: Yes, that sounds good for the `flushSync()`. But I'm also just wondering, what is the reason for wrapping it in a `queueMicrotask`?

mayank99: you can't `flushSync` inside an Effect. i don't think that needs a comment, because React's error is clear enough when you try it

rohan-kadkol (approved):

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
mayank99: Updated the comment a little bit, because `flushSync` doesn't actually prevent the flash of unslotted content. [`c783d4e` (#1954)](https://github.com/iTwin/iTwinUI/pull/1954/commits/c783d4e9bcbba8f9b7e1a6a25062a072b794ba20)

To understand this more deeply, you can comment out the `flushSync` and call `setShadowRoot`, then visit `Dialog` stories.

rohan-kadkol: To confirm, which `Dialog` stories exactly are you suggesting me to visit? And what should I keep an eye out for?

mayank99: Actually I can't reproduce it anymore. But this `flushSync` was added in #1836, where I noticed that the measurements used by the draggable Dialog here were wrong:

https://github.com/iTwin/iTwinUI/blob/e3bdda14a320540036ee04a706c1b4d19783df03/packages/itwinui-react/src/core/Dialog/DialogMain.tsx#L139-L143

PR #1955, opened 2024-03-26
Adds inline-size to horizontal tabs wrapper.

Ben-Pusey-Bentley:
## Changes

Adds `inline-size: 100%` to horizontal tabs wrappers. This change is meant to counteract an issue caused by this [PR](https://github.com/iTwin/iTwinUI/pull/1884), where adding `contain: inline-size` caused the tabs wrapper to change it's implicit width.

## Testing

Ran css and react visual tests to make sure that there are no unintended visual changes.

## Docs

Added patch changesets to both the react and css packages.

### DIFF
--- a/.changeset/great-news-beg.md
+++ b/.changeset/great-news-beg.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Adds `inline-size` to `Tabs.Wrapper` to counteract visual change to the width of horizontal tabs wrappers.
--- a/.changeset/great-news-beg.md
+++ b/.changeset/great-news-beg.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Adds `inline-size` to `Tabs.Wrapper` to counteract visual change to the width of horizontal tabs wrappers.
+Fixed a regression in `Tabs` where the panel content was not occupying the full width of the container.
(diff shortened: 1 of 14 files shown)


packages/itwinui-css/src/tabs/base.scss
```diff
@@ -130,6 +130,7 @@ $borderless-horizontal-tab-min-height: calc(
   grid-template-columns: 1fr auto;
   grid-template-rows: auto 1fr;
   contain: inline-size;
+  inline-size: 100%;
```
mayank99, line 133: You can undo #1911 now

.changeset/great-news-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Adds `inline-size` to `Tabs.Wrapper` to counteract visual change to the width of horizontal tabs wrappers.
```
mayank99, line 5: Changesets should be worded in a way that emphasizes why the user would care about this change.

```suggestion
Fixed a regression in `Tabs` where the panel content was not occupying the full width of the container.
```

packages/itwinui-css/src/tabs/base.scss
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1955/commits/7f88dcd263e6519b38730bca28f5d150a2ca0483)

mayank99 (approved):

.changeset/great-news-beg.md
Ben-Pusey-Bentley: Changed along with the css changeset in this [commit.](https://github.com/iTwin/iTwinUI/pull/1955/commits/8177c00df7a68b47737f9f2b92017bb2ecb84aee)

PR #1956, opened 2024-03-26
add shim for `useSyncExternalStore`

mayank99:
## Changes

[`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore) is a hook added in React 18 that makes it easier to subscribe to DOM APIs and support server-side rendering.

This PR adds a shim that makes it possible for this hook to be used in React 17 (falls back to simple `useLayoutEffect`/`useEffect`). This is adapted from [React's source code](https://github.com/facebook/react/tree/main/packages/use-sync-external-store) and has been shortened to remove unnecessary parts like warnings and comments. I wanted to use the published [npm package](https://www.npmjs.com/use-sync-external-store), but it only supports CJS :/

To start off, `useMediaQuery` has been updated to make use of this hook.

In future PRs, we can add it in other places for more useful things.

## Testing

Tested that `<ThemeProvider theme="os">` works fine in playground.

## Docs

N/A

No changeset necessary because this change should be invisible to the user.

### DIFF
--- a/packages/itwinui-react/src/core/utils/hooks/useMediaQuery.ts
+++ b/packages/itwinui-react/src/core/utils/hooks/useMediaQuery.ts
@@ -3,36 +3,26 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { getWindow } from '../functions/index.js';
-import { useLayoutEffect } from './useIsomorphicLayoutEffect.js';
+import { useSyncExternalStore } from './useSyncExternalStore.js';
 
 export const useMediaQuery = (queryString: string) => {
-  const [matches, setMatches] = React.useState<boolean>();
+  const [getSnapshot, subscribe] = React.useMemo(() => {
+    const mediaQueryList = window.matchMedia(queryString);
 
-  useLayoutEffect(() => {
-    const mediaQueryList = getWindow()?.matchMedia?.(queryString);
-    const handleChange = ({ matches }: MediaQueryListEvent) =>
-      setMatches(matches);
-
-    if (mediaQueryList != undefined) {
-      setMatches(mediaQueryList.matches);
-      try {
-        mediaQueryList.addEventListener('change', handleChange);
-      } catch {
-        // Safari 13 fallback
-        mediaQueryList.addListener?.(handleChange);
-      }
-    }
-
-    return () => {
-      try {
-        mediaQueryList?.removeEventListener('change', handleChange);
-      } catch {
-        // Safari 13 fallback
-        mediaQueryList?.removeListener?.(handleChange);
-      }
-    };
+    return [
+      () => mediaQueryList.matches,
+      (onChange: () => void) => {
+        mediaQueryList.addEventListener?.('change', onChange);
+        return () => {
+          mediaQueryList.removeEventListener?.('change', onChange);
+        };
+      },
+    ];
   }, [queryString]);
 
-  return !!matches;
+  return useSyncExternalStore(
+    subscribe,
+    typeof window !== 'undefined' ? getSnapshot : () => undefined,
+    () => undefined,
+  );
 };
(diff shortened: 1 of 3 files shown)


AdamMeza-Bentley (approved):
super interesting hook. excited to see what other things it'll allow us to do. looks good

packages/itwinui-react/src/core/utils/hooks/useSyncExternalStore.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @see https://github.com/facebook/react/tree/main/packages/use-sync-external-store
+ */
+function useSyncExternalStoreShim<T>(
+  subscribe: (onSubscribe: () => void) => () => void,
+  getSnapshot: () => T,
+): T {
+  const value = getSnapshot();
+  const [{ instance }, forceUpdate] = React.useState({
+    instance: { value, getSnapshot },
+  });
```
rohan-kadkol, line 31: Just wondering, what's the benefit of using `useState({ instance: {…} })` instead of directly using `useState( {…} )`?

packages/itwinui-react/src/core/utils/hooks/useMediaQuery.ts
```diff
@@ -3,36 +3,23 @@
...
+    return typeof window !== 'undefined'
+      ? window.matchMedia?.(queryString).matches
+      : undefined;
   }, [queryString]);
 
-  return !!matches;
+  const subscribe = React.useCallback(
+    (onChange: () => void) => {
+      const mediaQueryList = window.matchMedia?.(queryString);
+      mediaQueryList?.addEventListener?.('change', onChange);
```
rohan-kadkol, line 18: The Safari fallback is no longer needed? If so, should it be mentioned in a changeset?

mayank99: nah, it's been supported since safari 14, which is long enough to not be important.

and it's still using `?.` so it will still work correctly for first mount, it just will never update.

packages/itwinui-react/src/core/utils/hooks/useSyncExternalStore.ts
mayank99: this is taken directly from the react source code. i'm not exactly sure why they did it this way, but i think it is so they get a new object each time (and thus force a re-render when needed).

mayank99: yup, my understanding is correct. see comment here: https://github.com/facebook/react/blob/2b036d3f1f016fbe8b121d223d96f09e785b97e1/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L82-L84

packages/itwinui-react/src/core/utils/hooks/useSyncExternalStore.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  React.useLayoutEffect(() => {
```
rohan-kadkol, line 33: I feel comments like [this](https://redirect.github.com/facebook/react/blob/2b036d3f1f016fbe8b121d223d96f09e785b97e1/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L104-L105) and [this](https://redirect.github.com/facebook/react/blob/2b036d3f1f016fbe8b121d223d96f09e785b97e1/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L116-L117) make it easy to understand the code. So, it would be helpful to have them here too.

rohan-kadkol (approved):

packages/itwinui-react/src/core/utils/hooks/useSyncExternalStore.ts
mayank99: i deliberately removed the comments because it's not code that will ever be touched. it's write-once-and-forget.

as mentioned in the PR description, i would have liked to use the `use-sync-external-store` npm package, but react doesn't publish an ESM version of it

PR #1957, opened 2024-03-27
Bump `@floating-ui/react`: `^0.26.3` → `^0.26.10`

rohan-kadkol:
## Changes

@mayank99 pointed out some unit tests that were failing on Mayank's end upon upgrading `@floating-ui/react`.

The issue didn't seem to be some breaking change from that package that was released in a patch release, as we previously thought. Mayank thinks the issue may have been with the way they define their dependencies that may have led to incompatible transitive deps ([#1957 (comment)](https://github.com/iTwin/iTwinUI/pull/1957#discussion_r1541294212)). But we're not completely sure.

Thus, the end result is that just bumping the package from `^0.26.3` → `^0.26.10` would likely be enough since all the tests are still passing. Another benefit from this PR is that during the debug process, I followed Mayank's suggestion to uninstall and then reinstall `@floating-ui/react`, and that removed the duplicate version of `@floating-ui/utils` ([code](https://github.com/iTwin/iTwinUI/pull/1957/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL2879-L2885)).

Thus, we decided to release 3.7.4 without this PR. Once this PR is merged, it would be released in 3.8.0. ([#1957 (comment)](https://github.com/iTwin/iTwinUI/pull/1957#discussion_r1541294212)).

<details>
<summary>Old changes section</summary>

As pointed out by @mayank99, `@floating-ui/react` seems to publish "breaking behavioral changes" in patch releases. Since we depended on `^0.26.3` (i.e. `0.26.x`), users might have been downloading the later patch versions of this package that might give unexpected results or errors.

To avoid unexpected results, this PR pins the dep version of this package to `0.26.3` instead of `0.26.x`.

After this quick PR, I will try updating to the latest patch version and investigate what changes we need to make in our package as a result.

</details>

## Testing

Confirmed that all tests still pass.

## Docs

Added changeset.

### DIFF
--- a/.changeset/spicy-plums-switch.md
+++ b/.changeset/spicy-plums-switch.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Changed our internal dependency on `@floating-ui/react` from `^0.26.3` to `0.26.3` to prevent unexpected results when users use later patch versions of `@floating-ui/react` that may contain behavioral breaking changes.
--- a/.changeset/spicy-plums-switch.md
+++ b/.changeset/spicy-plums-switch.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Changed our internal dependency on `@floating-ui/react` from `^0.26.3` to `0.26.3` to prevent unexpected results when users use later patch versions of `@floating-ui/react` that may contain behavioral breaking changes.
+Changed our internal dependency on `@floating-ui/react` from `^0.26.3` to `0.26.3` to prevent unexpected results when users use later patch versions of `@floating-ui/react` that may contain breaking behavioral changes.
--- a/.changeset/spicy-plums-switch.md
+++ b/.changeset/spicy-plums-switch.md
@@ -1,5 +1,5 @@
 ---
-"@itwin/itwinui-react": patch
+"@itwin/itwinui-react": minor
 ---
 
-Bumped the `@floating-ui/react` dependency from `^0.26.3` to `^0.26.10`.
+Bumped the minimum required version of `@floating-ui/react` from `^0.26.3` to `^0.26.10`. (Make sure to also update transitive deps!)
--- a/.changeset/spicy-plums-switch.md
+++ b/.changeset/spicy-plums-switch.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-The dependency on `@floating-ui/react` has been temporarily pinned to `0.26.3` to avoid potential breaking behavioral changes from future versions.
+Bumped the `@floating-ui/react` dependency from `^0.26.3` to `^0.26.10`.
--- a/.changeset/spicy-plums-switch.md
+++ b/.changeset/spicy-plums-switch.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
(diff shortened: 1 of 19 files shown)


.changeset/spicy-plums-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Changed our internal dependency on `@floating-ui/react` from `^0.26.3` to `0.26.3` to prevent unexpected results when users use later patch versions of `@floating-ui/react` that may contain breaking behavioral changes.
```
mayank99, line 5:
```suggestion
The dependency on `@floating-ui/react` has been temporarily pinned to `0.26.3` to avoid potential breaking behavioral changes from future versions.
```

mayank99 (approved):

pnpm-lock.yaml
```diff
@@ -2863,14 +2863,14 @@ packages:
       react-dom: 18.2.0(react@18.2.0)
     dev: false
 
-  /@floating-ui/react@0.26.9(react-dom@18.2.0)(react@18.2.0):
-    resolution: {integrity: sha512-p86wynZJVEkEq2BBjY/8p2g3biQ6TlgT4o/3KgFKyTWoJLU1GZ8wpctwRqtkEl2tseYA+kw7dBAIDFcednfI5w==}
+  /@floating-ui/react@0.26.3(react-dom@18.2.0)(react@18.2.0):
+    resolution: {integrity: sha512-iKH8WRR0L/nLiM6qavFZxkyegIZRMxGnM9aKEc71M4wRlUNkgTamjPsOQXy11oZbDOH37MiTbk/nAPn9M2+shA==}
```
rohan-kadkol, line 2867: Actually, should we pin the version to `0.26.9` instead of `0.26.3`, since that what we were using so far?

mayank99: I guess there is no science behind the exact version here. Would it be time-consuming to find out which exact version broke our tests?

mayank99: It looks like the problem cannot be reproduced anymore with latest versions of floating-ui deps. I think last week I may have ended up in a broken lockfile state with incompatible versions of transitive floating-ui deps (or maybe they fixed something in 0.26.10).

So then I would say there is no need to pin `@floating-ui/react` because it does not affect which transitive dep versions get installed. We can release 3.7.4 without it

And in 3.8.0, we can bump minimum required version to `^0.26.10`.

.changeset/spicy-plums-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Bumped the `@floating-ui/react` dependency from `^0.26.3` to `^0.26.10`.
```
rohan-kadkol, line 5: Just wanted to confirm before merging, is the new changeset alright? Also, should this be a `minor` changeset instead since we suspect that floating-ui's patch releases might contain breaking behavioral changes?

mayank99: something like this?

```suggestion
---
"@itwin/itwinui-react": minor
---

Bumped the minimum required version of `@floating-ui/react` from `^0.26.3` to `^0.26.10`. (Make sure to also update transitive deps!)
```

rohan-kadkol:
Since the direction of this PR changed drastically after https://github.com/iTwin/iTwinUI/pull/1957#discussion_r1541193878, I updated the PR description, PR title, and the changeset before merging this PR.

mayank99 (approved):

PR #1960, opened 2024-03-27
create Anchor component for css-workshop

FlyersPh9:
## Changes

- Created reusable `Anchor.astro` component.
- Update `Button.astro` and `Tag.astro` to use `HTMLTag` for `as` prop.
- Updated multiple demo pages to use said component.

## Testing

- Ran CSS image tests.

## Docs

n/a


apps/css-workshop/src/pages/alert.astro
```diff
@@ -39,14 +40,13 @@ import IconButton_ from '../components/IconButton.astro';
...
-          data-iui-status='informational'>Learn more</a
-        >
-      </span>
+      <!-- prettier-ignore -->
+      <span class='iui-alert-message'>
+        Informational message.<Anchor_
+          class='iui-alert-link'
+          underline='true'
+          status='informational'
+        >Learn more</Anchor_></span>
```
FlyersPh9, line 49: This feels pretty hacky, but it's the only way I've found to fix the white space issue.

apps/css-workshop/src/pages/anchor.astro
```diff
@@ -15,24 +16,37 @@ import Layout from './_layout.astro';
...
-      <a
-        href='https://www.bentley.com/en'
-        class='iui-anchor iui-anchor-external'
-        referrerpolicy='no-referrer'
-        target='_blank'
-        id='test-2'>Bentley Systems<span class='iui-visually-hidden'>(opens in new tab)</span></a
-      > - An anchor link that opens in a new tab / window.
+      <Anchor_ external='true' href='https://www.bentley.com/en'>
+        Bentley Systems<span class='iui-visually-hidden'>(opens in new tab)</span>
+      </Anchor_>
```
FlyersPh9, line 26: This is the only test image that is still failing and I am not sure how to solve it. The white space is being added before the pseudo element.
anchor-dif

apps/css-workshop/src/components/Anchor.astro
```diff
@@ -0,0 +1,18 @@
+---
+type Props = {
+  external?: 'true';
+  status?: 'informational' | 'positive' | 'warning' | 'negative';
+  underline?: 'true';
```
mayank99, line 5: This should be a boolean prop

```ts
underline?: boolean
```

implemented like this:

```jsx
data-iui-underline={underline ? 'true' : undefined}
```

and used like this:
```jsx
<Anchor_ underline>
```

mayank99: Same with `external`, which should be a boolean and should also be renamed to `isExternal` to match react.

apps/css-workshop/src/components/Anchor.astro
```diff
@@ -0,0 +1,18 @@
...
+} & astroHTML.JSX.HTMLAttributes;
+
+const { class: className, external, href, status, underline, ...props } = Astro.props;
+---
+
+<a
+  class:list={['iui-anchor', className, external ? 'iui-anchor-external' : '']}
```
mayank99, line 12: This should use object syntax for the conditional

```suggestion
  class:list={['iui-anchor', className, { 'iui-anchor-external': external }]}
```

apps/css-workshop/src/components/Anchor.astro
```diff
@@ -0,0 +1,18 @@
...
+  data-iui-status={status}
+  data-iui-underline={underline}
+  {...external ? { referrerpolicy: 'no-referrer', target: '_blank' } : {}}
```
mayank99, line 15: `referrerpolicy` and `target` should be added by the consumer, rather than being handled in here.

apps/css-workshop/src/components/Anchor.astro
```diff
@@ -0,0 +1,18 @@
...
+  href={href ? href : '#'}
```
mayank99, line 16: We shouldn't have such hacks here. Better to have the consuming components manually add `href='#'` because it's an anti-pattern

apps/css-workshop/src/components/Anchor.astro
mayank99, line 1: I think we should add the `as` prop, because there are a bunch of instance of `iui-anchor` which could be updated to use `<Anchor_ as="something">`

apps/css-workshop/src/pages/alert.astro
mayank99: Doesn't `prettier-ignore` ignore the whole file? Would it be better to use `prettier-ignore-start` and `prettier-ignore-end`?

apps/css-workshop/src/pages/anchor.astro
mayank99: What if you remove whitespace from start and end?

```jsx
<Anchor_ …>…</Anchor_>
```
rather than

```jsx
<Anchor_ …>
  …
</Anchor_>
```

apps/css-workshop/src/pages/anchor.astro
```diff
@@ -15,24 +16,37 @@ import Layout from './_layout.astro';
...
 
     <p>
-      <a
-        href='https://www.bentley.com/en'
-        class='iui-anchor iui-anchor-external'
-        referrerpolicy='no-referrer'
-        target='_blank'
-        id='test-2'>Bentley Systems<span class='iui-visually-hidden'>(opens in new tab)</span></a
-      > - An anchor link that opens in a new tab / window.
+      <Anchor_ external='true' href='https://www.bentley.com/en'>
```
Ben-Pusey-Bentley, line 24: This anchor is the one that should have the `id='test-2` set on it, not [this one](https://github.com/iTwin/iTwinUI/blob/97891d6ac107dc60547b9082b1c8496516dfa870/apps/css-workshop/src/pages/anchor.astro#L31). This change is breaking the hover and focus visual tests for this file.

apps/css-workshop/src/pages/anchor.astro
FlyersPh9: I've got this working:
```html
<!-- prettier-ignore -->
<Anchor_
  …
>Bentley Systems<span class='iui-visually-hidden'> (opens in new tab)</span></Anchor_>
```

apps/css-workshop/src/components/Anchor.astro
FlyersPh9: Both changes made.

apps/css-workshop/src/components/Anchor.astro
FlyersPh9: Removed from the component, inlined on anchor demo page.

apps/css-workshop/src/components/Anchor.astro
FlyersPh9: Removed and added `href="#"` to individual demos.

apps/css-workshop/src/pages/anchor.astro
FlyersPh9: Moved `id="test-2"` from the anchor with `this one` as a label to `Bentley Systems`.

apps/css-workshop/src/components/Anchor.astro
FlyersPh9: Added `as` prop and used for `label` within `file-upload`.  Unsure about the other instances of `iui-anchor`:
- Breadcrumbs are already using `<Button as='a' class='iui-anchor'>`.
- Tag is using `<Tag variant='basic' class='iui-anchor'>`.

Changing `<a>` to `<Element>` was introducing white space issues again so I added `<!-- prettier-ignore -->` within `Anchor.astro`.

Ben-Pusey-Bentley (approved):

apps/css-workshop/src/components/Anchor.astro
mayank99: You can use `<Button as={Anchor}>` and `<Tag as={Anchor}>`.

I think you might need to change the type of `as` from `string` to [`HTMLTag`](https://docs.astro.build/en/guides/typescript/#polymorphic-type) for that.

```diff
+ import type { HTMLTag } from 'astro/types';
```
```diff
- as?: string;
+ as?: HTMLTag;
```

FlyersPh9: Done for `Button.astro` and used in `breadcrumbs.astro`.  Looking into `Tag.astro`, the file is a bit different than what I've worked on so far.

FlyersPh9: Ok, got it done for Tag as well.

apps/portal/src/pages/index.astro
```diff
@@ -1,4 +1,6 @@
 ---
+import Anchor_ from '../../../css-workshop/src/components/Anchor.astro';
```
mayank99, line 2: undo this one, the `portal` is a completely different codebase from `css-workshop`

mayank99 (approved):
LGTM after https://github.com/iTwin/iTwinUI/pull/1960#discussion_r1594339595

apps/portal/src/pages/index.astro
FlyersPh9: Reverted.

PR #1962, opened 2024-03-28
add `useHydration` check for ShadowRoot

mayank99:
## Changes

This PR adds a new `useHydration` hook to help determine hydration state, which is then used in `VisuallyHidden` and `ShadowRoot` to prevent unnecessary `null` renders on client. *(Follow-up to #1936, #1954, #1956, for #1930)*

To make the hydration state available everywhere without re-running the same code over and over, `HydrationProvider` is used. `useHydration` will reuse the context value if the code has already executed. This optimization is also useful when the server render only happens for a small part of the application (e.g. initial page / loading state), and then every subsequent navigation happens entirely on client.

The implementation itself relies on `useSyncExternalStore` to determine `"hydrating"` state (see [`getServerSnapshot` docs](https://react.dev/reference/react/useSyncExternalStore#parameters)) and `useEffect` to determine `"hydrated"` state (see [`useEffect` docs](https://react.dev/reference/react/useEffect#displaying-different-content-on-the-server-and-the-client)).

## Testing

Tested `ProgressRadial` and `VisuallyHidden` in playgrounds:
- no flash during pure client render (vite)
- no errors during server render + hydration (next, astro)

Also tested that React 17 is functional (see [commit](https://github.com/iTwin/iTwinUI/pull/1962/commits/d0f3aa10d6ebaca34151fc59946b96750d00c65f)). I did get a hydration warning in Nextjs, but no serious errors.

Made sure all existing tests pass.

## Docs

N/A

Added changeset to briefly mention the overall effect on users.

### DIFF
--- a/.changeset/twenty-keys-roll.md
+++ b/.changeset/twenty-keys-roll.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed a regression in `Table` where the layout of a custom `Cell` unexpectedly changed from horizontal to vertical.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,14 +180,20 @@ jobs:
         uses: actions/github-script@v6
         with:
           script: |
-            github.rest.repos.createCommitStatus({
+            github.rest.checks.create({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              sha: '${{ github.event.pull_request.head.sha }}',
-              state: 'success',
-              context: 'Deploy preview',
-              description: 'Live preview of the current branch',
-              target_url: 'https://itwin.github.io/iTwinUI/${{ github.event.number }}',
+              head_sha: '${{ github.event.pull_request.head.sha }}',
+              conclusion: 'neutral',
+              name: 'Deploy preview',
+              output: {
+                title: 'Live preview of the current branch',
+                summary: '',
+                text: `\
+            - [css-workshop](https://itwin.github.io/iTwinUI/${{ github.event.number }}/css)
+            - [react-workshop](https://itwin.github.io/iTwinUI/${{ github.event.number }}/react)
+            - [docs website](https://deploy-preview-${{ github.event.number }}--itwinui-previews.netlify.app/)`
+              }
             })
 
   unit-test:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,6 +48,12 @@ To clone and build iTwinUI, you'll need [Git](https://git-scm.com), [Node 18+](h
 
 _Before running this command, make sure Docker is running. See [Visual testing](#visual-testing-css) (CSS and React) sections below for more details._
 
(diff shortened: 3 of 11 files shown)


packages/itwinui-react/src/core/utils/providers/HydrationProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <HydrationContext.Provider value={isHydrated}>
+      {!isHydrated ? <HydrationCheck onHydrate={onHydrate} /> : null}
+      {children}
+    </HydrationContext.Provider>
+  );
+};
+
+/** This is extracted into a child component to ensure it runs first. */
+const HydrationCheck = ({ onHydrate }: { onHydrate: () => void }) => {
+  React.useEffect(() => void onHydrate(), [onHydrate]);
```
mayank99, line 64: This is a react thing. Child effects always run first
image

packages/itwinui-react/src/core/utils/components/ShadowRoot.tsx
```diff
@@ -85,27 +86,34 @@ function useShadowRoot(
...
-      // Flush the state immediately to ensure layout measurements in parent component are correct
+      // Flush the state immediately after shadow-root is attached, to ensure that layout
+      // measurements in parent component are correct.
+      // Without this, the parent component may end up measuring the layout when the shadow-root
+      // is attached in the DOM but React hasn't rendered any slots or content into it yet.
       ReactDOM.flushSync(() => setShadowRoot(shadow));
+    };
+
+    queueMicrotask(() => {
+      setupShadowRoot();
```
mayank99, line 116: By fixing the first client render, I was able to reproduce the original issue related to incorrect measurements in Dialog. It was trying to read measurements when the shadow-root was attached but state wasn't updated.

image

To fix it, I moved both of these things (`attachShadow` browser call and `setShadowRoot` react call) closer to each other into queueMicrotask. (Same as it was before #1954)

packages/itwinui-react/src/core/utils/providers/HydrationProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * @private
+ */
+export const useHydration = () => {
+  // Returns true only if getServerSnapshot is called on the client.
+  // In React 18, this is true during hydration.
+  const hydrating = useSyncExternalStore(
+    noopSubscribe,
+    () => false,
+    () => !isServer,
+  );
```
rohan-kadkol, line 29: I'm a little confused about why we're using `useSyncExternalStore` here.

mayank99: `useSyncExternalStore` is the only way to check for hydration in react.

- `noopSubscribe` is just an empty function. It needs to be referentially stable, so it's declared outside of the component.
- `getSnapshot` returns false (called during non-hydration client renders)
- `getServerSnapshot` returns true on the server and during hydration. So I'm using a boolean check to make it only return true during hydration.

PR #1963, opened 2024-03-28
wrap `all.css` in `@layer`

mayank99:
## Changes

This PR wraps the distributed `all.css` inside a `@layer`. Previously this file was unlayered and was being explicitly layered by the importer (e.g. `css-workshop` and `itwinui-react`).

This is the first step towards introducing nested internal layers for managing cross-component cascade priority.

Note: Sass doesn't allow `@forward` inside `@layer` so I had to use [`meta.load-css`](https://sass-lang.com/documentation/modules/meta/#load-css).

## Testing

Everything works the same as before, just the layers have moved from one place to another.

## Docs

Added `major` changeset for itwinui-css, since this is technically a breaking change.


packages/itwinui-react/src/styles.js/styles.module.css
mayank99, line 1: for some reason the slider track isn't appearing in [`react-workshop`](https://itwin.github.io/iTwinUI/1963/react/?story=slider--basic&theme=dark) :/


this is puzzling because it looks fine in [`css-workshop`](https://itwin.github.io/iTwinUI/1963/css/slider) and [`website`](https://deploy-preview-1963--itwinui-previews.netlify.app/docs/slider/)

mayank99: added [temporary fix](https://github.com/iTwin/iTwinUI/pull/1963/commits/1d02d4744efb9142755d04ca28d70d3326a49437) but this might be worth investigating and finding a workaround for, in case users face the same issue

AdamMeza-Bentley (approved):

.changeset/dry-cups-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": major
+---
+
+All styles have been wrapped in a cascade layer named `itwinui`.
```
rohan-kadkol, line 5: Should we mention the v3 nested cascade layer (`itwinui.v3`)?

mayank99: Probably not? It's an implementation detail that we should be able to change at any time.

packages/itwinui-react/src/styles.js/styles.module.css
mayank99: **update**: It looks like this might actually a bug in `esbuild`; kinda similar to [esbuild#3620](https://redirect.github.com/evanw/esbuild/issues/3620). i was able to confirm that this issue does not happen when `lightningcss` is used as the minifier ([`59c6066`](https://github.com/iTwin/iTwinUI/pull/1963/commits/59c606688eaec193601719222e28af07bfde9068)).

Since `esbuild` is the default CSS minifier used by `vite`, it is possible that our users might run into the same issue. For this reason, I've slightly adjusted our CSS to work around the issue. [`afbad49`](https://github.com/iTwin/iTwinUI/pull/1963/commits/afbad496ead0343a634da3990d4e17ca4ed37709)

rohan-kadkol (approved):

PR #1964, opened 2024-03-29
Fix `InformationPanel` size on `orientation` change

Ben-Pusey-Bentley:
## Changes

Fixes an issue with the `InformationPanel` component where resizing the component and then changing the orientation would maintain the resized dimensions in a way that broke the intended style of the panel. This was fixed by adjusting the state logic so that only part of the style of the panel is effected by the state, dependent on the value of orientation. `width` is only applied when the orientation is `vertical`, and `height` is only applied when the orientation is `horizontal`.

## Testing

Tested in the vite playground to make sure the fix works. Ran react visual and unit tests to make sure there were no unintended changes. Also added e2e test for this issue.

## Docs

Added patch changeset to `@itwin/itwinui-react`.


packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -56,12 +56,36 @@ export const InformationPanel = React.forwardRef((props, ref) => {
     isOpen = false,
     orientation = 'vertical',
     resizable = true,
-    children,
     ...rest
   } = props;
 
+  return (
+    <InformationPanelInternal
```
mayank99, line 63: It would be better to pass all props directly to `InformationPanelInternal` instead of handling some of them here.

```jsx
<InformationPanelInternal
  key={props.orientation}
  ref={forwardedRef}
  {...props}
/>
```

testing/e2e/app/routes/InformationPanel/spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    await resizeInformationPanel(page);
+
+    expect(await infoPanel.getAttribute('style')).toBeTruthy();
+    await page.click('#OrientationButton');
+    expect(await infoPanel.getAttribute('style')).toBeFalsy();
+  });
+
+  async function resizeInformationPanel(page: Page) {
+    const resizer = page.locator('._iui3-resizer');
```
mayank99, line 29: Targeting these `_iui3` classes is a no-go. It's an implementation detail that should never be used anywhere.

You can instead try to find the edge of the panel and hover/mousedown on it.

.changeset/grumpy-bags-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixes bug with `InformationPanel` where changing orientation after resizing the panel would cause the panel to maintain it's resized width/height.
```
mayank99, line 5: Changesets should generally use past tense

"Fixes bug" → "Fixed a bug"

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
Ben-Pusey-Bentley: Okay, will do.

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -56,12 +56,36 @@ export const InformationPanel = React.forwardRef((props, ref) => {
...
+      )}
+      ref={ref}
+      key={orientation}
+      orientation={orientation}
+      resizable={resizable}
+      {...rest}
+    />
+  );
+}) as PolymorphicForwardRefComponent<'div', InformationPanelProps>;
+
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Will do.

testing/e2e/app/routes/InformationPanel/route.tsx
```diff
@@ -0,0 +1,152 @@
...
+              Elephants communicate using low-frequency infrasound.
+            </Blockquote>
+          </InformationPanelBody>
+        </InformationPanel>
+      </InformationPanelWrapper>
+    </>
+  );
+};
+
+function MyTable(): JSX.Element {
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Alright, will do.

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -95,26 +119,18 @@ export const InformationPanel = React.forwardRef((props, ref) => {
...
-        'iui-information-panel',
-        {
-          'iui-right': orientation === 'vertical',
-          'iui-bottom': orientation === 'horizontal',
-          'iui-visible': isOpen,
-        },
-        className,
-      )}
       ref={refs}
+      style={{ width: infoPanelWidth, height: infoPanelHeight }}
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Will do.

.changeset/grumpy-bags-whisper.md
Ben-Pusey-Bentley: Will do.

testing/e2e/app/routes/InformationPanel/spec.ts
Ben-Pusey-Bentley: Okay, I'll see if I can find the left edge of the panel instead.

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/7b041b9609f61f2c90fac1398436da887c49f3e6)

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/637bb2d31f64bd74e5489c31d84662e72a0fad63)

testing/e2e/app/routes/InformationPanel/spec.ts
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/f1ffdde84c52d4237d06660be4b449c56ddd3db9)

AdamMeza-Bentley (approved):

testing/e2e/app/routes/InformationPanel/spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(await infoPanel.getAttribute('style')).toBeTruthy();
+  });
+  test('should reset styles on orientation change', async ({ page }) => {
+    await page.goto('/InformationPanel', { waitUntil: 'domcontentloaded' });
+    const infoPanel = page.locator('#InformationPanel');
+    expect(await infoPanel.getAttribute('style')).toBeFalsy();
+
+    await resizeInformationPanel(page);
+
+    expect(await infoPanel.getAttribute('style')).toBeTruthy();
```
mayank99, line 23: Instead of checking for the `style` attribute, we should be checking that the dimensions actually changed.

When changing orientation from vertical to horizontal, the panel width should reset to be the same width as the wrapper. And when changing the orientation back to vertical, the height should reset to be the same as the wrapper.

Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/d272dbf14dd66a774a396cf860087905f17863d2)

testing/e2e/app/routes/InformationPanel/route.tsx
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Removed ids in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/d272dbf14dd66a774a396cf860087905f17863d2)

.changeset/grumpy-bags-whisper.md
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/2f362c6558331f5e67f75da885fa345549aafa2c)

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -60,8 +60,13 @@ export const InformationPanel = React.forwardRef((props, ref) => {
     ...rest
   } = props;
 
+  const [infoPanelSize, setInfoPanelSize] = React.useState<{
+    width: number | undefined;
+    height: number | undefined;
+  }>({ width: undefined, height: undefined });
```
mayank99, line 66: I just realized that we don't even need to store both dimensions because we are only using one at a time, based on the ternary styles

```js
  const [infoPanelSize, setInfoPanelSize] = React.useState<number>(undefined);
```

Ben-Pusey-Bentley: Okay, I'll update the state to be this instead.

Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/499b86a1893cdd28e63d08803276cdfc182e5f84) I want to note that this change causes the size to be maintained after orientation change, it just changes how the value is used. This behavior should be fine, since if a user reduces the size of a panel and then changes its location, they probably don't expect the size to reset.

testing/e2e/app/routes/InformationPanel/spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    await page.goto('/InformationPanel');
+    const infoPanel = page.locator('#InformationPanel');
+    const infoPanelWrapper = page.locator('#InformationPanelWrapper');
+    const wrapperWidth = (await infoPanelWrapper.boundingBox())!.width;
+    const wrapperHeight = (await infoPanelWrapper.boundingBox())!.height;
+    const orientationButton = page.getByRole('button');
+
+    const initialWidth = (await infoPanel.boundingBox())!.width;
+
+    {
```
mayank99, line 24: can you add short comments before every block to indicate what the block is doing/testing?

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
mayank99: can you also add the same explanation to a code comment in the test where you are testing this behavior?

testing/e2e/app/routes/InformationPanel/spec.ts
Ben-Pusey-Bentley: Sure.

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
Ben-Pusey-Bentley: Will do.

testing/e2e/app/routes/InformationPanel/spec.ts
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/ebc06196b0575adf72724056bcc9171ee8e18291)

packages/itwinui-react/src/core/InformationPanel/InformationPanel.tsx
Ben-Pusey-Bentley: Added a comment in this [commit.](https://github.com/iTwin/iTwinUI/pull/1964/commits/ebc06196b0575adf72724056bcc9171ee8e18291)

PR #1966, opened 2024-04-01
Astro Tag Component

AdamMeza-Bentley:
## Changes

- Created reusable `Tag_` astro component to use in both `tile.astro` and `tag.astro`

### `.prettierignore`

I ran into a formatting issue in the `tag.astro` file relatedto how astro handles white space sensitivity, which is [intentional](https://github.com/withastro/astro/issues/6893). In order to match the old snapshots I was forced to format the `Tag_` components in a way that prettier automatically undid on save. So I added a `.prettierignore` which includes a few files that use the `Tag_` component. This unfortunately prevents us from using prettier in those files.

## Testing

All existing tests are passing. This made no changes to any tests

If not applicable, you can write "N/A".

## Docs

N/A


apps/css-workshop/src/pages/tile.astro
```diff
@@ -1,6 +1,7 @@
 ---
 import Layout from './_layout.astro';
 import Badge from '../components/Badge.astro';
+import Tag_ from '../components/Tag.astro';
```
Ben-Pusey-Bentley, line 4: Is there a reason that the tag component is imported as `Tag_` while the badge component is imported as just `Badge`? It feels like all Astro components should use the same conventions when possible.

AdamMeza-Bentley: I think that's a mistake with the `Badge` component which I didn't work on. It was my understanding that we landed on the underscore https://github.com/iTwin/iTwinUI/pull/1948#discussion_r1538102242

Ben-Pusey-Bentley: It seems like that was only done in the `badge.astro` file because the name `Badge` caused a conflict. It doesn't here in the `tile.astro` file, so there is no reason to use `_` to differentiate the name. The same is true for the `Tag` component here, so it seems to me like it also shouldn't have the `_`.

Ben-Pusey-Bentley: Actually, rereading it, I think you might be right. In that case, the `Badge` here should be `Badge_`.

AdamMeza-Bentley: Yeah I get the same error Mayank did when trying to remove it. I'll add the underscore to `Badge`

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,41 @@
+---
+const { title, className, id, style, button = true, variant, href, ...props } = Astro.props;
+---
+
+<>
+  {
+    variant === 'basic' && !href && (
+      <span class='iui-tag-basic' id={id}>
+        {title}
```
mayank99, line 9: This `title` prop is definitely the wrong way to go about it. You should be using a `<slot />` instead.

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,41 @@
...
+      </span>
+    )
+  }
+  {
+    href && (
+      <a href={href} class='iui-anchor iui-tag-basic' id={id}>
```
mayank99, line 15: Instead of inlining the `iui-anchor` class here, it would be useful to create an `Anchor.astro` component (in a separate PR), since the `iui-anchor` class belongs to that component.

Until then, I think `iui-anchor` should be passed externally in the place where it used. For example:

```astro
<Tag_ as="a" class="iui-anchor" href="#the-url">Click me</Tag_>
```

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,41 @@
...
+        {title}
+      </a>
+    )
+  }
+
+  {
+    variant !== 'basic' && (
+      <span className='iui-tag' id={id} style={style} {...props}>
```
mayank99, line 23: You shouldn't need to create a new element for every condition. The way this has been written is all messed up too.
- Why is there a `className`? This is a React concept that doesn't apply in plain HTML or Astro.
- Why is `id` and `style` explicitly applied instead of being spread with the rest of the `...props`?
- Why does the basic tag not spread the `...props`?
- Why is the externally passed `class` prop not merged using `class:list`?

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,41 @@
...
+        <span class='iui-tag-label' title={title}>
```
mayank99, line 24: `title` attribute should not be used

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,41 @@
...
+          {title}
+        </span>
+        {button && (
```
mayank99, line 27: This `button` prop is too vague. What kind of "button"? What does it do? Can I pass my own button into it?

apps/css-workshop/src/pages/tile.astro
AdamMeza-Bentley: ok yeah i'll move that to a new pr

apps/css-workshop/src/components/Tag.astro
AdamMeza-Bentley: sweet i'll use slot instead

apps/css-workshop/src/components/Tag.astro
AdamMeza-Bentley: sounds good! I can make these changes and figure out a better way

apps/css-workshop/src/components/Tag.astro
AdamMeza-Bentley: I'll make it more specific.

apps/css-workshop/src/components/Tag.astro
AdamMeza-Bentley: I'm a bit lost on how else I would handle each use case. Simply adding classes doesn't account for the difference in HTML output. The basic tag is either a span or an anchor while the standard variant is a bit more complex. In a different scenario I might use `React.createElement`, which obviously wouldn't work here. Im sure there's a way to do it as you've said. Would you be able to elaborate?

mayank99: For basic vs standard variants, you can use more granular conditionals instead of branching off in two totally different directions.

1. **Root element**: This is just a span with a different class (`iui-tag` vs `iui-tag-basic`), which can be handled using [`class:list`](https://docs.astro.build/en/reference/directives-reference/#classlist).

2. **Label element**: The only other difference is that the slotted text should be wrapped in a span for standard tags.

Together, it might look like:

```astro
<span
  class:list={[
    {
     'iui-tag': variant !== 'basic',
     'iui-tag-basic': variant === 'basic'
    },
    className
  ]}
  {...props}
>
  {variant === 'basic' ? <slot /> : <span class='iui-tag-label'><slot /></span>}
</span>
```

---

The anchor case shouldn't be handled inside this component, as explained in my [other comment](https://github.com/iTwin/iTwinUI/pull/1966#discussion_r1548185081). When in doubt, look at the docs: [Dynamic HTML](https://docs.astro.build/en/basics/astro-syntax/#dynamic-html) + [Polymorphic type](https://docs.astro.build/en/guides/typescript/#polymorphic-type)

apps/css-workshop/src/components/Tag.astro
AdamMeza-Bentley: So i've been having trouble with this. I've tried implementing the solution you suggested but in order to render the a tag properly I've ended up using almost identical logic to the original i.e. conditional branching based off props. When I don't the tag renders as a span with the `as='a'` attribute. Would you be able to clarify how to avoid the conditional branching? or was your comment soley related to the class and not to the branching?

apps/css-workshop/src/pages/tag.astro
```diff
@@ -19,433 +20,141 @@ import Layout from './_layout.astro';
...
   <h2>Basic tag containers</h2>
   <section id='demo-basic'>
     <div class='iui-tag-container'>
-      <span class='iui-tag-basic'>tags</span><span class='iui-tag-basic'>go</span><span
-        class='iui-tag-basic'>here</span
-      ><span class='iui-tag-basic'>tags</span><span class='iui-tag-basic'>go</span><span
-        class='iui-tag-basic'>here</span
-      >
+      <Tag_ variant='basic' 
+        >tags</Tag_><Tag_ variant='basic'
```
AdamMeza-Bentley, line 82: this is the specific formatting that prettier isn't too fond of

AdamMeza-Bentley:
Additionally there are some changes to the snapshots as there is another spacing issue caused by astro that I can't quite figure out. the end result is a very small change. If anyone has ideas on how to solve that I would love some help

apps/css-workshop/src/pages/tag.astro
Ben-Pusey-Bentley: Why is this formatting only being used in some of the tag containers? For example, the container directly above this one does not use this formatting.

AdamMeza-Bentley: oh that was an accident. I thought I had committed those changes

Ben-Pusey-Bentley (approved):

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,35 @@
+---
+const { closeButton = true, variant, className, as, buttonId, ...props } = Astro.props;
+const Element = as ? 'a' : 'span';
```
mayank99, line 3: `a` shouldn't be hardcoded here.

I think you meant something like this:

```suggestion
const Element = as ?? 'span';
```

apps/css-workshop/src/components/Tag.astro
```diff
@@ -0,0 +1,35 @@
+---
+const { closeButton = true, variant, className, as, buttonId, ...props } = Astro.props;
```
mayank99, line 2: - `closeButton` should be renamed to `hasCloseButton` to make it obvious that it's a button.
- `className` is a react thing that shouldn't be used here. What you want instead is `class: className`, which keeps the prop named `class`.
- `buttonId` shouldn't be its own prop. Use `buttonProps` instead, and pass an `id` into it.

.prettierignore
mayank99, line 1: `prettierignore`-ing entire files is not acceptable. Can you highlight the exact instances (in the latest diff) where prettier was causing issues? And can you summarize all the other workarounds you tried and why they didn't work?

AdamMeza-Bentley: At pretty much every step of the process prettier/astro are causing issues. From the inlining happening in `Tag.astro`...
image

... to every container that has more than 1 basic tag (which is almost everysingle one) ... 

image

Without doing this type of forced inlining there are extra spaces which cause failing CSS tests. 

I've tried using [various forms](https://prettier.io/docs/en/ignore.html) of `prettier-ignore` to ignore the specific parts that prettier wants to rearrange but those haven't worked. I found [this article](https://medium.com/@popoever/use-prettier-ignore-to-solve-astro-vite-compiler-render-issue-57f1aef6e437) which addressed this and it didn't have any effect. I've also tried arranging the `Tags_` in a way that prettier won't change but those inevitably run into the extra white space issue

mayank99: Have you tried putting a `<!-- prettier-ignore -->` comment above all the HTML? This is better than using `.prettierignore` file because you get to keep prettier for formatting non-HTML parts of the file

AdamMeza-Bentley: Yeah I've tried both things and prettier will continue to format things

AdamMeza-Bentley: have you tested it in `tag.astro`?

AdamMeza-Bentley: so yes it seems like using the prettier ignore comments do work. I had using them incorrectly. I added those and removed the `.prettierignore` file.

mayank99 (approved):

PR #1970, opened 2024-04-02
Astro checkbox component

AdamMeza-Bentley:
## Changes

- Created astro checkbox component. Added it to the css-workshop. 

## Testing

Al existing tests are passing. No changes were made.

## Docs

N/A


apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,54 @@
...
+      <div
+        class='iui-progress-indicator-radial'
+        data-iui-indeterminate='true'
+        data-iui-size='x-small'
+      />
+    )
+  }
+  <span class='iui-checkbox-label'>
+    {label && label}
+    {variant === 'smile' && <svg-smiley-happy aria-hidden='true' style='display: flex;' />}
```
AdamMeza-Bentley, line 50: This is a bit janky. However after looking into options I found the issue is that astro doesn't support passing jsx via props. See [relevent stack overflow](https://stackoverflow.com/questions/75613671/passing-react-elements-as-props-in-astro-when-using-mui). Doing this via conditional rendering seems to be the most straight forward way to do it

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,54 @@
...
+type Props = {
+  isLoading?: boolean;
+  label?: React.ReactNode;
+  indeterminate?: boolean;
+  checked?: boolean;
+  disabled?: boolean;
+  variant?: 'eyeball' | 'smile' | 'frown' | 'big frown';
+  content?: React.ReactNode;
+  status?: 'negative' | 'positive' | 'warning' | 'negative';
+} & astroHTML.JSX.HTMLAttributes;
```
AdamMeza-Bentley, line 13: Chose to do using typing here as the props got a bit more complicated than expected

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: Found a better solution using the [slot component](https://docs.astro.build/en/basics/astro-components/#slots) provided by Astro

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -78,435 +55,228 @@ import Layout from './_layout.astro';
...
-          </span>
-        </label>
-        <label class='iui-checkbox-wrapper'>
-          <input class='iui-checkbox' type='checkbox' />
-          <span class='iui-checkbox-label'>
+      <div class='iui-input-grid iui-input-group-wrapper'>
+        <div class='iui-input-label'>Checkbox group label</div>
+        <div class='iui-input-group'>
+          <Checkbox_ checked>
             <svg-smiley-happy aria-hidden='true' style='display: flex;'></svg-smiley-happy>
```
rohan-kadkol, line 106: I think this should be `svg-smiley-happy-very`, as it was before. Not sure why the test didn't fail. Maybe the change amount was below the threshold?

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -78,435 +55,228 @@ import Layout from './_layout.astro';
...
-      </div>
-      <div class='iui-status-message'>
-        <div>Help message.</div>
+          </Checkbox_>
+        </div>
+        <div class='iui-status-message'>
+          <div>Help message.</div>
+        </div>
       </div>
     </div>
```
rohan-kadkol, line 122: I believe this should still be `</section>`. i.e. the `demo-default` section should end here, like it was previously.

A similar change would be needed for L281.

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,52 @@
...
+>
+  <input
+    class={cx('iui-checkbox', {
+      'iui-loading': isLoading,
+      'iui-checkbox-visibility': variant === 'eyeball',
+    })}
+    type='checkbox'
+    disabled={disabled}
+    checked={checked}
+    name={indeterminate ? 'indeterminate' : ''}
```
rohan-kadkol, line 37: Should the else part be `undefined` to prevent adding the `name` attribute when inderminate is `false`/`undefined`?

```suggestion
    name={indeterminate ? 'indeterminate' : undefined}
```

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -78,435 +55,228 @@ import Layout from './_layout.astro';
...
-
-      <label class='iui-checkbox-wrapper iui-warning'>
-        <input class='iui-checkbox' type='checkbox' />
-        <span class='iui-checkbox-label'>Checkbox</span>
-      </label>
-    </div>
+      >
+        <Checkbox_ indeterminate label='Half checkbox' />
+        <Checkbox_ label='Checkbox' />
+        <Checkbox_ label='Loading checkbox' isLoading={true} />
```
rohan-kadkol, line 209: nit: isLoading would also be sufficient.

```suggestion
        <Checkbox_ label='Loading checkbox' isLoading />
```

apps/css-workshop/src/pages/checkbox.astro
AdamMeza-Bentley: It didn't have the very happy variant before. I can add it tho

AdamMeza-Bentley: Yeah I can add that back

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -78,435 +55,228 @@ import Layout from './_layout.astro';
...
-        disabled
-      />
-    </label>
-
-    <label class='iui-checkbox-wrapper iui-disabled'>
-      <input class='iui-checkbox iui-checkbox-visibility' type='checkbox' disabled />
-    </label>
-  </section>
+      >
+        <Checkbox_ indeterminate label='Half checkbox' style={{ flexDirection: 'row-reverse' }} />
```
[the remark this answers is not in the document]
AdamMeza-Bentley: not sure how to change the the DOM order but I made that change along with the other suggestions in my last push

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,61 @@
...
+  label,
+  indeterminate,
+  disabled,
+  variant,
+  ...props
+} = Astro.props;
+---
+
+<label
+  class={cx('iui-checkbox-wrapper', {
```
rohan-kadkol, line 28: Instead of `cx`, you can use Astro's [`class:list`](https://docs.astro.build/en/reference/directives-reference/#classlist)

apps/css-workshop/src/pages/checkbox.astro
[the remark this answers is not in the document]
AdamMeza-Bentley: all good! I can convert it back to inline

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: yeah I'll make that switch

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -1,5 +1,6 @@
 ---
 import Layout from './_layout.astro';
+import Checkbox_ from '../components/Checkbox.astro';
```
rohan-kadkol, line 3: I believe there are checkboxes in other demo files too. E.g. `information-panel.astro`, `table.astro`, etc. Those would need to use the `Checkbox_` component too.

AdamMeza-Bentley: I'll switch those over!

apps/css-workshop/src/pages/checkbox.astro
AdamMeza-Bentley: addressed this in my last push

apps/css-workshop/src/pages/tree.astro
```diff
@@ -385,11 +386,7 @@ import Layout from './_layout.astro';
     <ul class='iui-tree' role='tree'>
       <li class='iui-tree-item' role='treeitem' tabindex='0' aria-expanded='true'>
         <div class='iui-tree-node'>
-          <input
-            class='iui-checkbox iui-checkbox-visibility iui-tree-node-checkbox'
-            type='checkbox'
-            checked
-          />
+          <Checkbox_ variant='basic eyeball' checked />
```
rohan-kadkol, line 389: Looks like there is a type error for `variant` here and other similar places in this file.

Instead of re-using the variant prop, maybe you can consider adding some other prop for with/without wrapper. This will help make the `variant` prop accept the same values as the `variant` prop in `Checkbox.tsx`.

apps/css-workshop/src/pages/table.astro
```diff
@@ -131,7 +132,7 @@ import Layout from './_layout.astro';
         <div class='iui-table-header'>
           <div class='iui-table-row'>
             <div class='iui-table-cell iui-slot'>
-              <input class='iui-checkbox' type='checkbox' id='select-all-checkboxes' />
+              <Checkbox_ variant='basic' indeterminate />
```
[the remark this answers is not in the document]
AdamMeza-Bentley: good catch, i added that back in

apps/css-workshop/src/pages/tree.astro
AdamMeza-Bentley: adde a new `basic` prop to avoid over using variant

rohan-kadkol (approved):
LGTM 🚀

apps/css-workshop/src/pages/transfer-list.astro
```diff
@@ -1,5 +1,7 @@
 ---
 import Layout from './_layout.astro';
+import Checkbox_ from '../components/Checkbox.astro';
+import Checkbox from './checkbox.astro';
```
mayank99, line 4: Looks like the checkbox page itself is being imported here

apps/css-workshop/src/pages/checkbox.astro
```diff
@@ -398,101 +226,37 @@ import Layout from './_layout.astro';
...
-      <label class='iui-checkbox-wrapper iui-negative'>
-        <span class='iui-checkbox-label'>Checkbox</span>
-        <input class='iui-checkbox' type='checkbox' />
-      </label>
-
-      <label class='iui-checkbox-wrapper iui-warning'>
-        <span class='iui-checkbox-label'>Checkbox</span>
-        <input class='iui-checkbox' type='checkbox' />
-      </label>
+      <Checkbox_ indeterminate style={{ flexDirection: 'row-reverse' }}>Half checkbox</Checkbox_>
```
mayank99, line 229: I think we should just remove these checkboxes since we don't support them in React (which means nobody is using it)

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,71 @@
+---
+type Props = {
+  isLoading?: boolean;
+  indeterminate?: boolean;
+  checked?: boolean;
+  disabled?: boolean;
+  variant?: 'eyeball';
+  basic?: boolean;
+  status?: 'negative' | 'positive' | 'warning';
+} & astroHTML.JSX.HTMLAttributes;
```
mayank99, line 10: I think some of these props don't need to be re-specified since they are part of the input itself.

`checked` and `disabled` can come from `astroHTML.JSX.InputHTMLAttributes`

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,71 @@
...
+
+const { status, isLoading, indeterminate, disabled, variant, basic, ...props } = Astro.props;
```
mayank99, line 12: `basic` doesn't need to be a prop at all.

It can be derived from [`Astro.slots`](https://docs.astro.build/en/reference/api-reference/#astroslots):

```js
const basic = Astro.slots.has('default');
```

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,71 @@
...
+    <label
+      class:list={[
+        'iui-checkbox-wrapper',
+        {
+          'iui-disabled': disabled,
+          'iui-loading': isLoading,
+          [`iui-${status}`]: !!status,
+        },
+      ]}
+      {...props}
```
mayank99, line 43: This one spreads `...props` in two different places. It should only spread on the input.

apps/css-workshop/src/components/Checkbox.astro
```diff
@@ -0,0 +1,71 @@
...
+      <input
+        class:list={[
+          'iui-checkbox',
+          {
+            'iui-loading': isLoading,
+            'iui-checkbox-visibility': variant?.includes('eyeball'),
+          },
+        ]}
+        type='checkbox'
+        name={indeterminate ? 'indeterminate' : undefined}
```
mayank99, line 54: `name` should not be set here.

apps/css-workshop/src/pages/checkbox.astro
AdamMeza-Bentley: Yeah i'll get rid of those. Should I change the "checkbox on left title" or just remove it?

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: Screenshot 2024-04-05 141900

The original script used `name` to add the `indeterminate` mask image so I did the same. 

The CSS selector is using the indeterminate pseudo class which I don't think is [directly assignable without Javascript](https://css-tricks.com/almanac/selectors/i/indeterminate/#aa-it-cannot-be-set-in-the-html). 
 
Using the name attribute seemed like a fairly simple way of doing it. But if there's better way to do it, I can do that instead.

apps/css-workshop/src/pages/transfer-list.astro
AdamMeza-Bentley: resolved this on a new branch. pr soon

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: resolved this on a new branch. pr soon

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: resolved this on a new branch. pr soon

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: resolved in new branch

apps/css-workshop/src/components/Checkbox.astro
mayank99: The [`name`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#name) attribute belongs to the individual checkboxes, and shouldn't be set at the component level.

`indeterminate` can totally be just be set as a regular attribute in HTML (even if it doesn't do anything by itself). The JavaScript code can then hook read the HTML attribute, and convert it to the `indeterminate` property.

PR #1971, opened 2024-04-02
add `stretch` prop to `Button`

mayank99:
## Changes

The new `stretch` prop essentially applies `inline-size: 100%`. This is most useful in narrow layouts (e.g. appui widgets) and mobile views.

image

See similar examples from other design systems: [1](https://www.lightningdesignsystem.com/components/buttons/#Stretch), [2](https://atlassian.design/components/button/examples#full-width), [3](https://carbondesignsystem.com/components/button/usage/)

---

**Implementation notes**: This is implemented using a private inline CSS variable. There is some global CSS applied on every element, where this CSS variable's value is used to convert it to `inline-size`. Benefits:
- The consumer can still specify a custom `width` in their CSS and it will be respected.
- There is no need to add a data attribute or anything like that to the button SCSS.
- Other components can add similar props in the future with no extra effort.

## Testing

Added story + visual test + unit test.

## Docs

Added jsdocs + documentation section + changeset.


packages/itwinui-react/src/styles.js/styles.module.css
```diff
@@ -13,3 +13,13 @@
     --_iui-v3-loaded: yes;
   }
 }
+
+@layer itwinui {
+  *,
+  *::before,
+  *::after {
+    /* When --_iui-width is set, inline-size will use it, otherwise it will be automatically reverted. */
+    --_iui-width: initial;
```
rohan-kadkol, line 22: If `--_iui-width` is initialized, will the fallback part of the `var()` (i.e. `revert-layer`) not be triggered?

mayank99: No, when a custom property is set to `initial`, it will trigger the fallback.

rohan-kadkol (approved):

Ben-Pusey-Bentley (approved):

PR #1976, opened 2024-04-03
Dialog: support uncontrolled mode and `show()`/`close()` methods

mayank99:
## Changes

WIP

### DIFF
--- a/apps/react-workshop/src/Dialog.stories.tsx
+++ b/apps/react-workshop/src/Dialog.stories.tsx
@@ -9,50 +9,33 @@ export default {
   title: 'Dialog',
 };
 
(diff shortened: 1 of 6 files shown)


mayank99:
Superseded by [#1983](https://github.com/iTwin/iTwinUI/pull/1983/files/ef7cb0d7a46e392e254a0d2b33cbc8a2b8a3db75..5021dba0cc53dfe518fa968f58511f4d04a5d66e#diff-e66bd6cea9157991b681796019e8930946763980a8ed8765a04842a0e0b77366)

The `ref` solution may feel more "react-y" but introduces difficulties:
- starts off as `null` because we cannot run custom code during initialization
- needs explicit typing using `typeof Dialog.Ref`
- you get a new ref every time the dialog mounts/unmounts from DOM.
- doesn't compose well. `useImperativeHandle` apparently needs to be defined to be defined for every single component, even for simple wrappers. (I couldn't even get the ref to work with Modal so far.)

PR #1981, opened 2024-04-04
use `jest` without `globalThis` when detecting it

mayank99:
## Changes

Fixes https://github.com/iTwin/iTwinUI/discussions/1978

## Testing

N/A. We do not use `jest` in this repo, so attempting to test this is futile.

## Docs

Added changeset.

### DIFF
--- a/packages/itwinui-react/src/core/utils/functions/dev.ts
+++ b/packages/itwinui-react/src/core/utils/functions/dev.ts
@@ -3,7 +3,9 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-const isJest = typeof (globalThis as any).jest !== 'undefined';
+// @ts-expect-error -- jest is not used in iTwinUI but may be used in consuming apps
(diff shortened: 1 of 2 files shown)


packages/itwinui-react/src/core/utils/functions/dev.ts
```diff
@@ -3,7 +3,9 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-const isJest = typeof (globalThis as any).jest !== 'undefined';
+// @ts-expect-error -- jest is not used in iTwinUI but may be used in consuming apps
+const isJest = typeof jest !== 'undefined';
+
```
rohan-kadkol, line 8: nit: Just noticed an extra line (unless it's intentional). If we work on this file later, maybe we can remove this line.

mayank99: It's intentional, since there is a comment above. The newline indicates (at least to me) that the comment does not relate to the code below the newline.

PR #1987, opened 2024-04-09
Adds `Button` Astro Component.

Ben-Pusey-Bentley:
## Changes

Adds `Button` Astro component to the `css-workshop`. Imported as `Button_` to all files in order to maintain consistency. Takes in props for all of the potential `data-iui` values.

## Testing

Ran visual tests and confirmed that the component did not cause any visual changes.

## Docs

N/A

### DIFF
--- a/apps/css-workshop/src/components/Button.astro
+++ b/apps/css-workshop/src/components/Button.astro
@@ -1,0 +1,21 @@
+---
+type Props = {
+  variant?: 'high-visibility' | 'borderless' | 'cta' | 'idea';
+  size?: 'small' | 'large';
+  active?: 'true';
+  shift?: 'left' | 'right';
+} & astroHTML.JSX.ButtonHTMLAttributes;
+
+const { class: className, variant, size, active, shift, ...props } = Astro.props;
+---
+
+<button
+  class:list={['iui-button', className]}
+  data-iui-variant={variant}
+  data-iui-size={size}
+  data-iui-active={active}
+  data-iui-shift={shift}
+  {...props}
+>
+  <slot />
+</button>
--- a/apps/css-workshop/src/pages/alert.astro
+++ b/apps/css-workshop/src/pages/alert.astro
@@ -1,5 +1,6 @@
 ---
 import Layout from './_layout.astro';
+import Button_ from '../components/Button.astro';
 ---
 
 <Layout title='Alert'>
@@ -46,15 +47,9 @@ import Layout from './_layout.astro';
           data-iui-status='informational'>Learn more</a
         >
       </span>
-      <button
-        aria-label='Close'
-        type='button'
-        class='iui-button'
-        data-iui-variant='borderless'
-        data-iui-size='small'
-      >
+      <Button_ aria-label='Close' type='button' variant='borderless' size='small'>
         <svg-close-small class='iui-button-icon' aria-hidden='true'></svg-close-small>
-      </button>
+      </Button_>
     </div>
 
     <div class='iui-alert' data-iui-status='positive'>
@@ -68,15 +63,9 @@ import Layout from './_layout.astro';
           data-iui-status='positive'>Learn more</a
         >
       </span>
-      <button
-        aria-label='Close'
-        type='button'
-        class='iui-button'
-        data-iui-variant='borderless'
-        data-iui-size='small'
-      >
+      <Button_ aria-label='Close' type='button' variant='borderless' size='small'>
         <svg-close-small class='iui-button-icon' aria-hidden='true'></svg-close-small>
-      </button>
+      </Button_>
     </div>
 
     <div class='iui-alert' data-iui-status='warning'>
@@ -90,15 +79,9 @@ import Layout from './_layout.astro';
           data-iui-status='warning'>Learn more</a
         >
       </span>
-      <button
-        aria-label='Close'
-        type='button'
-        class='iui-button'
-        data-iui-variant='borderless'
-        data-iui-size='small'
-      >
+      <Button_ aria-label='Close' type='button' variant='borderless' size='small'>
         <svg-close-small class='iui-button-icon' aria-hidden='true'></svg-close-small>
-      </button>
+      </Button_>
     </div>
 
     <div class='iui-alert' data-iui-status='negative' id='test-single-alert'>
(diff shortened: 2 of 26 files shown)


apps/css-workshop/src/components/Button.astro
```diff
@@ -0,0 +1,22 @@
+---
+type Props = {
+  variant?: 'high-visibility' | 'borderless' | 'cta';
+  size?: 'small' | 'large';
+  active?: 'true';
```
rohan-kadkol, line 5: Should this be a boolean?

Ben-Pusey-Bentley: I left this prop as a string to keep things consistent with how `data-iui-active` was used and to avoid having to perform a check in the component when setting `data-iui-active`.  Making active a boolean does make using the prop easier when using the component, so I have made a [commit](https://github.com/iTwin/iTwinUI/pull/1987/commits/fab6716cea1ea8fd5e11080d3a05f5a433db34eb) with this change. Let me know what you think.

rohan-kadkol: Great, I feel the boolean prop is better and more consistent with how we usually do it. We also did it similarly in https://github.com/iTwin/iTwinUI/pull/1970.

https://github.com/iTwin/iTwinUI/blob/d635a7cb09b04eb5d02f1fc648ffb23ac846065c/apps/css-workshop/src/components/Checkbox.astro#L36

rohan-kadkol (approved):

AdamMeza-Bentley (approved):

apps/css-workshop/src/components/Button.astro
```diff
@@ -0,0 +1,22 @@
+---
+type Props = {
+  variant?: 'high-visibility' | 'borderless' | 'cta';
+  size?: 'small' | 'large';
+  active?: boolean;
```
mayank99, line 5: In React, we call this one `isActive`

apps/css-workshop/src/components/Button.astro
```diff
@@ -0,0 +1,22 @@
...
+  shift?: 'left' | 'right';
```
mayank99, line 6: The `shift` prop probably shouldn't exist within `Button`. For one, we don't have this prop in React. But more importantly, it only makes sense for borderless icon buttons. I think we might want to create a separate `IconButton` astro component, similar to React

apps/css-workshop/src/components/Button.astro
```diff
@@ -0,0 +1,22 @@
...
+  iuiType?: 'feedback' | 'dropdown';
```
mayank99, line 7: This `iuiType` prop doesn't make any sense.
- The `iui-` prefix is only used as a form of namespacing in CSS and DOM elements.
- dropdown and ideas button have nothing to do with each other and shouldn't be grouped this way

apps/css-workshop/src/pages/alert.astro
```diff
@@ -46,15 +47,9 @@ import Layout from './_layout.astro';
...
         >
       </span>
-      <button
-        aria-label='Close'
-        type='button'
-        class='iui-button'
-        data-iui-variant='borderless'
-        data-iui-size='small'
-      >
+      <Button_ aria-label='Close' type='button' variant='borderless' size='small'>
```
mayank99, line 50: `type="button"` should be set by default within the Button component, so that it doesn't need to be repeated for every instance.

apps/css-workshop/src/pages/button.astro
```diff
@@ -53,296 +54,288 @@ import Layout from './_layout.astro';
 
   <section id='demo-high-visibility'>
     <div id='demo-high-visibility-small' class='demo-container'>
-      <button class='iui-button' data-iui-variant='high-visibility' data-iui-size='small'>
+      <Button_ variant='high-visibility' size='small'>
         <span>Small</span>
```
mayank99, line 58: I think this `<span>` should be added within the component, rather than by the consumer

apps/css-workshop/src/pages/button.astro
```diff
@@ -53,296 +54,288 @@ import Layout from './_layout.astro';
...
-      </button>
-      <button class='iui-button' data-iui-variant='high-visibility' data-iui-size='small'>
+      </Button_>
+      <Button_ variant='high-visibility' size='small'>
         <svg-placeholder class='iui-button-icon' aria-hidden='true'></svg-placeholder>
```
mayank99, line 61: start icon and end icon need to be handled within the component. `class="iui-button-icon"` shouldn't need to be specified outside of Button, otherwise it kinda defeats the purpose of the component. This is a major oversight that I'm surprised was missed before PR merge.

This can be achieved using [named slots](https://docs.astro.build/en/basics/astro-components/#named-slots).

mayank99 (commented):
I found some pending issues that can be handled in a follow-up PR

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: The `iuiType` prop was added to bring the properties in line with the `data-iui-type` attribute being added in this [PR](https://github.com/iTwin/iTwinUI/pull/1919). If you don't think that the props on the component should line up with the attributes like this then it can be removed.

apps/css-workshop/src/pages/alert.astro
Ben-Pusey-Bentley: Will do.

apps/css-workshop/src/pages/button.astro
Ben-Pusey-Bentley: Okay, will do.

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Alright, will rename.

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Okay, will add `IconButton` component.

apps/css-workshop/src/components/Button.astro
mayank99: Generally, there isn't a strong reason for components props to mimick how the CSS is implemented. Pick the abstractions that make more sense.

That being said, I left a [comment](https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1567596614) in Jon's PR, because the CSS implementation also feels questionable to me.

Ben-Pusey-Bentley: Okay, will remove.

PR #1990, opened 2024-04-10
fix `styleType` being passed to DOM

mayank99:
## Changes

`LabeledSelect` forces a `styleType` even when it's disallowed for non-native Select. This PR ensures that it's handled correctly by `SelectButton`, rather than being passed to the `InputWithIcon` DOM element along with `...rest` props.

## Testing

Tested that the prop doesn't appear in DOM

## Docs

N/A. Too small of a change for changeset

### DIFF
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -290,6 +290,8 @@ const CustomSelect = React.forwardRef((props, forwardedRef) => {
     triggerProps,
     status,
     popoverProps,
+    // @ts-expect-error -- this prop is disallowed by types but should still be handled at runtime
+    styleType,
     ...rest
   } = props;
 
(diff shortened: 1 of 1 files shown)


AdamMeza-Bentley (approved):

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -423,6 +425,7 @@ const CustomSelect = React.forwardRef((props, forwardedRef) => {
           aria-expanded={isOpen}
           aria-haspopup='listbox'
           aria-controls={`${uid}-menu`}
+          styleType={styleType}
```
rohan-kadkol, line 428: If `styleType` is not supported for `CustomSelect`, should we pass it to `<SelectButton>`? Or would just removing it out of `rest` be sufficient?

rohan-kadkol (approved):

PR #1985, opened 2024-04-05
Astro checkbox fixes

AdamMeza-Bentley:
## Changes

This PR addresses unresolved issues from #1970. 

## Testing

Updates Visual Snapshots.

## Docs

N/A

PR #1991, opened 2024-04-10
Fix CI Visual Test Checks

Ben-Pusey-Bentley:
## Changes

Updates the `runTests.mjs` script so that CI visual test checks correctly fail when the tests fail. Does this by exiting the parent node process with a non-zero status code when the subprocess exits with a non-zero status code. Also makes changes to the `css-workshop` to fix some failing css visual tests.

## Testing

Used this PR to test that checks correctly fail when the tests fail.

## Docs

N/A


scripts/runTests.mjs
```diff
@@ -4,28 +4,26 @@
...
+} else {
+  startProcess(`turbo run test`);
+}
+
+function startProcess(command) {
+  const subprocess = spawn(command, {
+    stdio: 'inherit',
+    shell: true,
+  });
+  subprocess.on('exit', (code) => {
```
mayank99, line 24: I think instead of listening to the `exit` event, you might be able to instead use [`spawnSync`](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options) and use the `status` from its return value

apps/css-workshop/src/components/Checkbox.astro
mayank99, line 1: You might want to coordinate these checkbox changes with @AdamMeza-Bentley / #1985

scripts/runTests.mjs
Ben-Pusey-Bentley: That does look like it should work. I'll work on implementing it.

apps/css-workshop/src/components/Checkbox.astro
AdamMeza-Bentley: The changes to the checkbox classes are good. I can pull those into my PR if this is merged first.

scripts/runTests.mjs
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/1991/commits/1358882931ae676dfc38781c3b138f55acaa18aa)

mayank99 (approved):

AdamMeza-Bentley (approved):

PR #1994, opened 2024-04-10
add `role="toolbar"` support to `ButtonGroup`

mayank99:
## Changes

This PR enhances the `ButtonGroup` component to support the [Toolbar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/), with arrow-key navigation support. It is off by default and can be turned on by specifying [`role="toolbar"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/toolbar_role).

I would have liked to make this pattern the default, but that is risky because users might be passing all kinds of nonsense into `ButtonGroup` (such as `<input>`, which already has arrow-keys reserved for moving the cursor between characters; see [MDN advice](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/toolbar_role)). Also, in some cases, multiple `ButtonGroup`s might need to be combined into a single toolbar.

I was actually implementing this from scratch, when I randomly found that `floating-ui` provides the keyboard navigation and focus management for free? Not sure why, but I won't complain if I have to write less code.

I did have to move some things around in `ButtonGroup.tsx`, since it has a bunch of unrelated code for overflow. Turn off whitespace for better diff when reviewing.

## Testing

Manually tested that it works fine.

https://github.com/iTwin/iTwinUI/assets/9084735/5dfc8df7-d7bc-44b4-be77-199f9d8aae7d

https://github.com/iTwin/iTwinUI/assets/9084735/f7d69ed2-7ee0-4171-a18c-322e2e8152ba

Added e2e test as well. And updated one of the existing stories.

## Docs

Added jsdocs + real docs + changeset.


apps/website/src/content/docs/buttongroup.mdx
```diff
@@ -19,7 +19,29 @@ Use a button group if you want to put similar functionality in one place. It hel
 
 Buttons in a button group can be active, disabled and both.
 
-## Appearance
+## Usage
+
+The `ButtonGroup` should almost always be used with [`IconButton`](button#iconbutton) components. Make sure to also use the `label` prop to each `IconButton`; this text will be used to label the button and will be displayed in a tooltip.
```
rohan-kadkol, line 24: Grammar?

```suggestion
The `ButtonGroup` should almost always be used with [`IconButton`](button#iconbutton) components. Make sure to also pass the `label` prop to each `IconButton`; this text will be used to label the button and will be displayed in a tooltip.
```

rohan-kadkol (approved):

testing/e2e/app/routes/ButtonGroup/spec.ts
```diff
@@ -0,0 +1,63 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('ButtonGroup', () => {
+  test("should support keyboard navigation when role='toolbar'", async ({
+    page,
+  }) => {
+    await page.goto('/ButtonGroup');
+
+    await page.keyboard.press('Tab');
+    await expect(page.getByLabel('Button 1')).toBeFocused();
```
mayank99, line 10: Tests failing, only got caught after #1991

AdamMeza-Bentley (approved):

PR #1995, opened 2024-04-11
Adds vite playground to deploy preview check

Ben-Pusey-Bentley:
## Changes

Deploys the vite playground and adds link to those deployment on the deploy preview check. The playground is deployed to `/vite`.

## Testing

Using this PR to check both the deployment and the link in the deploy preview check for correctness.

## Docs

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -137,6 +137,9 @@ jobs:
       - name: Build vite-playground
         run: pnpm run build --filter=vite-playground
 
+      - name: Build remix-playground
+        run: pnpm run build --filter=remix-playground
+
       - name: List folders to preserve (open PRs and 'legacy')
         id: clean_exclude_folders
         uses: actions/github-script@v4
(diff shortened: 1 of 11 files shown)


AdamMeza-Bentley (approved):

.github/workflows/build.yml
```diff
@@ -175,6 +181,24 @@ jobs:
           git-config-name: github-actions[bot]
           git-config-email: github-actions[bot]@users.noreply.github.com
 
+      - name: Host vite-playground under /vite
+        uses: JamesIves/github-pages-deploy-action@v4
```
mayank99, line 185: I think we should only run this one on branches that aren't `main`. Can be achieved using an `if` property

Ben-Pusey-Bentley: Okay, will do.

Ben-Pusey-Bentley: Added `if` property in this [commit.](https://github.com/iTwin/iTwinUI/pull/1995/commits/b155bb47f055e02832bd005964c8b24aa27162f3)

mayank99 (approved):

PR #2001, opened 2024-04-15
Layered Dropdown Menu

rohan-kadkol:
## Changes

WIP

Creating this draft PR with my exploration steps. Once I realize the full scope of this task, will most likely be splitting this PR into smaller PRs.

### DIFF
--- a/packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
+++ b/packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
@@ -15,8 +15,9 @@ import type {
   PolymorphicForwardRefComponent,
   PortalProps,
 } from '../../utils/index.js';
-import { Menu } from '../Menu/Menu.js';
+// import { Menu } from '../Menu/Menu.js';
 import { usePopover } from '../Popover/Popover.js';
+import { Surface } from '../Surface/Surface.js';
 
 export type DropdownMenuProps = {
   /**
@@ -36,6 +37,12 @@ export type DropdownMenuProps = {
    * Child element to wrap dropdown with.
    */
   children: React.ReactNode;
+  /**
+   * Whether the DropdownMenu is layered.
+   *
+   * @default false
+   */
+  layered?: boolean;
 } & Pick<
   Parameters<typeof usePopover>[0],
   'visible' | 'onVisibleChange' | 'placement' | 'matchWidth'
@@ -72,6 +79,7 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
     matchWidth = false,
     onVisibleChange,
     portal = true,
+    layered = false,
     ...rest
   } = props;
 
@@ -90,6 +98,12 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
 
   const menuContent = React.useMemo(() => {
     if (typeof menuItems === 'function') {
+      // console.log(menuItems(close));
+
+      menuItems(close).forEach((item) => {
+        console.log(item.props.subMenuItems);
+      });
+
       return menuItems(close);
     }
     return menuItems;
@@ -106,32 +120,67 @@ export const DropdownMenu = React.forwardRef((props, forwardedRef) => {
 
   return (
     <>
-      {cloneElementWithRef(children, (children) => ({
-        ...popover.getReferenceProps(children.props),
-        'aria-expanded': popover.open,
-        ref: mergeRefs(triggerRef, popover.refs.setReference),
-      }))}
-      {popover.open && (
-        <Portal portal={portal}>
-          <Menu
-            {...popover.getFloatingProps({
-              role,
-              ...rest,
-              onKeyDown: mergeEventHandlers(props.onKeyDown, (e) => {
-                if (e.defaultPrevented) {
-                  return;
-                }
-                if (e.key === 'Tab') {
-                  close();
-                }
-              }),
-            })}
-            ref={popoverRef}
-          >
-            {menuContent}
-          </Menu>
-        </Portal>
-      )}
+      <DropdownMenuContext.Provider
+        value={{
+          layered: layered,
+        }}
+      >
+        {cloneElementWithRef(children, (children) => ({
+          ...popover.getReferenceProps(children.props),
+          'aria-expanded': popover.open,
+          ref: mergeRefs(triggerRef, popover.refs.setReference),
+        }))}
+
+        {popover.open && (
+          <Portal portal={portal}>
+            {
+              <Surface
+                {...popover.getFloatingProps({
+                  role,
+                  ...rest,
+                  onKeyDown: mergeEventHandlers(props.onKeyDown, (e) => {
+                    if (e.defaultPrevented) {
+                      return;
+                    }
+                    if (e.key === 'Tab') {
+                      close();
+                    }
+                  }),
+                })}
+                ref={popoverRef}
+              >
+                <Surface.Header>ABC</Surface.Header>
+                <Surface.Body>{menuContent}</Surface.Body>
+              </Surface>
+            }
+            {/* <Menu
+              {...popover.getFloatingProps({
+                role,
+                ...rest,
+                onKeyDown: mergeEventHandlers(props.onKeyDown, (e) => {
+                  if (e.defaultPrevented) {
+                    return;
+                  }
+                  if (e.key === 'Tab') {
+                    close();
+                  }
+                }),
+              })}
+              ref={popoverRef}
+            >
+              {menuContent}
+            </Menu> */}
+          </Portal>
+        )}
+      </DropdownMenuContext.Provider>
     </>
   );
 }) as PolymorphicForwardRefComponent<'div', DropdownMenuProps>;
+
+// ----------------------------------------------------------------------------
+
+export const DropdownMenuContext = React.createContext<{
+  layered: DropdownMenuProps['layered'];
+}>({
+  layered: false,
+});
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -5,16 +5,17 @@
 import * as React from 'react';
 import {
   SvgCaretRightSmall,
-  Portal,
+  // Portal,
   useMergedRefs,
   useId,
 } from '../../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
-import { Menu } from './Menu.js';
+// import { Menu } from './Menu.js';
 import { ListItem } from '../List/ListItem.js';
 import type { ListItemOwnProps } from '../List/ListItem.js';
 import { flushSync } from 'react-dom';
 import { usePopover } from '../Popover/Popover.js';
+// import { DropdownMenuContext } from '../DropdownMenu/DropdownMenu.js';
 
 /**
  * Context used to provide menu item ref to sub-menu items.
@@ -107,10 +108,12 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   } = props;
 
   const menuItemRef = React.useRef<HTMLElement>(null);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [focusOnSubmenu, setFocusOnSubmenu] = React.useState(false);
   const submenuId = useId();
 
   const [isSubmenuVisible, setIsSubmenuVisible] = React.useState(false);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [isNestedSubmenuVisible, setIsNestedSubmenuVisible] =
     React.useState(false);
   const parent = React.useContext(MenuItemContext);
@@ -180,6 +183,10 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
     onKeyDown,
   };
 
+  // const dropdownMenuContext = React.useContext(DropdownMenuContext);
+
+  // const { layered } = dropdownMenuContext;
+
   return (
     <ListItem
       as='div'
@@ -223,7 +230,7 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
         </ListItem.Icon>
       )}
 
-      {subMenuItems.length > 0 && popover.open && (
+      {/* {subMenuItems.length > 0 && popover.open && (
         <Portal>
           <MenuItemContext.Provider
             value={{ ref: menuItemRef, setIsNestedSubmenuVisible }}
@@ -244,7 +251,7 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
             </Menu>
           </MenuItemContext.Provider>
         </Portal>
-      )}
+      )} */}
     </ListItem>
   );
 }) as PolymorphicForwardRefComponent<'div', MenuItemProps>;
--- a/playgrounds/vite/src/App.tsx
+++ b/playgrounds/vite/src/App.tsx
@@ -1,10 +1,66 @@
-import { Button } from '@itwin/itwinui-react';
+import { SvgMore } from '@itwin/itwinui-icons-react';
+import {
+  // Button,
+  DropdownMenu,
+  IconButton,
+  MenuItem,
+} from '@itwin/itwinui-react';
 
 const App = () => {
+  const onClick = (index: number, close: () => void) => () => {
+    console.log(`Item #${index} clicked!`);
+    close();
+  };
+  const dropdownMenuItems = (close: () => void) => [
+    <MenuItem key={1} onClick={onClick(1, close)}>
+      Item #1
+    </MenuItem>,
+    <MenuItem key={2} onClick={onClick(2, close)}>
+      Item #2
+    </MenuItem>,
+    <MenuItem
+      key={3}
+      subMenuItems={[
+        <MenuItem
+          key={4}
+          subMenuItems={[
+            <MenuItem key={7} onClick={onClick(7, close)}>
+              Item #7
+            </MenuItem>,
+            <MenuItem key={8} onClick={onClick(8, close)}>
+              Item #8
+            </MenuItem>,
+          ]}
+        >
+          Item #4
+        </MenuItem>,
+        <MenuItem key={5} onClick={onClick(5, close)}>
+          Item #5
+        </MenuItem>,
+        <MenuItem
+          key={6}
+          subMenuItems={[
+            <MenuItem key={9} onClick={onClick(9, close)}>
+              Item #9
+            </MenuItem>,
+            <MenuItem key={10} onClick={onClick(10, close)}>
+              Item #10
+            </MenuItem>,
+          ]}
+        >
+          Item #6
+        </MenuItem>,
+      ]}
+    >
+      Item #3
+    </MenuItem>,
+  ];
   return (
-    <>
-      <Button>Hello world</Button>
-    </>
+    <DropdownMenu menuItems={dropdownMenuItems}>
+      <IconButton>
+        <SvgMore />
+      </IconButton>
+    </DropdownMenu>
   );
 };
 


mayank99:
Before we get too far down this rabbithole, I just want to remind/note that this needs to be a generic pattern that shouldn't be tied to `DropdownMenu`. There are valid cases for a "sliding" panel that can be triggered by any arbitrary button within an `InformationPanel` or `SideNavigation` (submenu) or an AppUI widget.

Some other important requirements:
- It should likely be implemented as a disclosure pattern.
- It should probably move focus into the new panel when "opened".
- Since the "original" panel is off screen, it should be hidden from keyboard/AT users.
- It should contain a back button, which "closes" the panel and moves focus back.
- Maybe it should close on pressing <kbd>Esc</kbd>? Unsure.

It might be tricky to come up with the right abstraction for this, but that's part of the exploration work.

mayank99:
Sharing this example from Firefox as an inspiration. It's basically a multi-level popover (rather than a menu).

I like the subtle animation and I like that it moves focus (although it's questionable that the focus is so far down the second screen. I think it should be on the back button instead).

https://github.com/iTwin/iTwinUI/assets/9084735/3d3c2a5f-b5b0-47ea-b0a0-c83ce50658eb

mayank99:
During a pairing session, we experimented with a generic `Panels` component that handles all the "panel switching" logic, and can be used anywhere (whether that's `Popover` or `InformationPanel` or `Menu` or an AppUI widget). It uses a simple [disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/) and manages focus.

```tsx
import {
  Flex,
  IconButton,
  List,
  ListItem,
  Surface,
  Text,
  ToggleSwitch,
import { useAtom, useAtomValue, useSetAtom, atom } from 'jotai';
import { SvgChevronLeft } from '@itwin/itwinui-icons-react';
import { flushSync } from 'react-dom';

const App = () => {
  const basePanelId = React.useId();
  const qualityPanelId = React.useId();
  const repeatId = React.useId();

  // Note: Will not work, because BackButton currently relies on context.
  const { goBack } = Panels.useInstance();

  return (
    <Surface style={{ display: 'inline-block' }}>
      <Panels defaultActiveId={basePanelId}>
        <Panel id={basePanelId}>
          <List>
            <ListItem>
              <Flex>
                <label htmlFor={repeatId}>Repeat</label>
                <Flex.Spacer />
                <ToggleSwitch id={repeatId} />
              </Flex>
            </ListItem>
            <ListItem>
              <Panel.Trigger for={qualityPanelId}>
                <ListItem.Action>Quality</ListItem.Action>
              </Panel.Trigger>
            </ListItem>
            <ListItem>Speed</ListItem>
            <ListItem>Loop</ListItem>
          </List>
        </Panel>

        <Panel id={qualityPanelId}>
          <Surface.Header as={Panel.Header}>Quality</Surface.Header>
          <List>
            <ListItem>
              <ListItem.Action
                onClick={() => {
                  // setQuality('240p');
                  goBack();
                }}
              >
                240p
              </ListItem.Action>
            </ListItem>
            <ListItem>360p</ListItem>
            <ListItem>480p</ListItem>
            <ListItem>720p</ListItem>
            <ListItem>1080p</ListItem>
          </List>
        </Panel>
      </Panels>
    </Surface>
  );
};

const expandedIdAtom = atom<string | undefined>(undefined);
const triggersAtom = atom(
  new Map<string, { triggerId: string; panelId: string }>(),
);

const Panels = ({
  children,
  defaultActiveId,
}: React.PropsWithChildren<any>) => {
  const [expandedId, setExpandedId] = useAtom(expandedIdAtom);

  if (expandedId === undefined) {
    setExpandedId(defaultActiveId);
  }

  return <>{children}</>;
};

const Panel = ({ children, id, ...rest }: React.PropsWithChildren<any>) => {
  const [expandedId] = useAtom(expandedIdAtom);
  return (
    <PanelIdContext.Provider value={id}>
      <div id={id} hidden={id !== expandedId} {...rest}>
        {children}
      </div>
    </PanelIdContext.Provider>
  );
};

const PanelIdContext = React.createContext('');

Panel.Header = ({ children, ...props }: React.PropsWithChildren<any>) => {
  return (
    <Flex {...props}>
      <Panel.BackButton />
      <Text
        as='h2'
        tabIndex={-1}
        // TODO: Confirm that focus moves correctly to the Text after the next panel is opened.
        // When a keyboard user triggers the panel, they should be able to continue tabbing into the panel.
        // When a screen-reader user triggers the panel, they should hear the name of the panel announced.
        //
        // Alternate idea: maybe the Panel itself could be focused. But then the panel needs a role and a label.
        ref={React.useCallback((el: HTMLElement | null) => el?.focus(), [])}
      >
        {children}
      </Text>
    </Flex>
  );
};

Panel.BackButton = () => {
  const setExpandedId = useSetAtom(expandedIdAtom);
  const panelId = React.useContext(PanelIdContext);
  const trigger = useAtomValue(triggersAtom).get(panelId);

  const goBack = () => {
    flushSync(() => setExpandedId(trigger?.panelId));

    if (trigger?.triggerId) {
      document.getElementById(trigger?.triggerId)?.focus();
    }
  };

  return (
    <IconButton
      label='Back'
      styleType='borderless'
      onClick={goBack}
      size='small'
      data-iui-shift='left'
    >
      <SvgChevronLeft />
    </IconButton>
  );
};

Panels.useInstance = () => ({
  goBack: () => {},
});

Panel.Trigger = ({
  children: childrenProp,
  for: forProp,
}: React.PropsWithChildren<{ for: string }>) => {
  const [expandedId, setExpandedId] = useAtom(expandedIdAtom);
  const [triggers, setTriggers] = useAtom(triggersAtom);
  const panelId = React.useContext(PanelIdContext);

  const children = React.Children.only(childrenProp) as any;

  const triggerFallbackId = React.useId();
  const triggerId = children?.props?.id || triggerFallbackId;

  if (triggers.get(forProp)?.triggerId !== triggerId) {
    setTriggers(new Map(triggers.set(forProp, { triggerId, panelId })));
  }

  return (
    React.isValidElement(children) &&
    React.cloneElement(children, {
      id: triggerId,
      onClick: () => setExpandedId(forProp),
      'aria-expanded': expandedId === forProp,
      'aria-controls': forProp,
    } as any)
  );
};

export default App;
```

</details>

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+      }}
+    >
+      <Box
+        ref={useMergedRefs(setElement, forwardedRef)}
+        id={id}
+        aria-labelledby={`${id}-header`}
+        tabIndex={-1}
+        {...rest}
+        className={cx('iui-panel', className, {
+          'iui-panel-visible': id === currentPanelId?.id,
```
mayank99, line 427: This class looks unused? Can you review your own PR to make sure there is no leftover/dead code?

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+        })}
+        style={{
+          display: isHidden ? 'none' : undefined,
```
mayank99, line 430: Instead of setting `display` as an inline style, prefer setting the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute and reinforcing it in CSS (see `Tabs` for example).

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+        direction: 'next',
+      },
+    });
+  }, [forProp, panelWrapperContext]);
+
+  if (!React.isValidElement(children)) {
+    return null;
+  }
+
+  const triggersMatch = panelWrapperContext?.triggers.current?.get(forProp);
```
mayank99, line 503: Reading refs during rendering is an anti-pattern. Why can't this be a state?

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+          // i.e. showing the current panel for a split second between when the animations ends and the panel is hidden
+          ...(id === currentPanelId?.id &&
+          animations != null &&
+          animations[id] != null
+            ? animations[id][animations[id].length - 1]
+            : {}),
+
+          ...style,
+        }}
+        {...{ inert: isInert ? '' : undefined }}
```
mayank99, line 442: We need to also load the polyfill for `inert` attribute.

Also, in the description you said "When not animating, all panels except the non-active panel are marked as inert", which I'm not sure is accurate? Could you try to rephrase how/when `inert` and `display` (or `hidden`) are applied so I understand it better?

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+  instance?: PanelsInstance;
+  /**
+   * To customize the panel transition animations, pass one of the following:
+   * - `number`: Duration in milliseconds. Overrides the default `animationOptions`.
+   * - `object`: For more customization. Merges with the default `animationOptions`.
+   * - `null`: Disables animations completely.
+   *
+   * @see `HTMLElement.animate`
+   */
+  animationOptions?: Parameters<HTMLElement['animate']>[1] | null;
```
mayank99, line 61: I think we can also leave out animationOptions from the initial implementation. We can revisit based on user needs.

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,600 @@
...
+    id: initialActiveId,
+  });
+
+  // Reducer to manage panel animations
+  const [{ animations }, dispatch] = React.useReducer(panelAnimationReducer, {
+    currentPanelId,
+    setCurrentPanelId,
+  } satisfies PanelAnimationState);
+
+  const animateToPanel = React.useCallback(
```
mayank99, line 122: There is so much animation-related code in this component that it's making the main component code difficult to understand.

Is it possible to write the component in a way that the general flow of the component is similar, regardless of whether we are animating or not? e.g. important component code should not be so tightly coupled with the animation code.

testing/e2e/app/routes/Panels/route.tsx
mayank99, line 1: I know there are no docs in this PR but it would be good to still at least add an `examples` file, because it gets used to validate `a11y`.

packages/itwinui-css/src/panels/panels.scss
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+
+.iui-panel-wrapper {
+  position: relative;
+  overflow: hidden;
+}
+
+.iui-panel {
+  position: absolute;
```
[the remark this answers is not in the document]
mayank99: To make them stack horizontally instead of vertically, you can use regular flex/grid layout. The element would basically scroll horizontally during the animation (1. click, 2. add new inert panel to DOM, 3. scroll to new panel and un-inert it, 4. remove old panel from DOM).

packages/itwinui-react/src/core/Panels/Panels.tsx
mayank99: Based on your link, floating-ui is _not_ reading/writing refs during rendering. It is ok to read/write refs (via `.current`) inside Effects, callback refs and event handlers. It is *not* ok to read/write refs (via `.current`) directly in the rendering path. See [docs](https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs).

packages/itwinui-react/src/core/Panels/Panels.tsx
mayank99: > When not animating: All `hidden` panels will be made `inert`

This is redundant. `inert` does nothing on `hidden` elements, it's only useful on visible elements.

> When animating: _All_ panels are made inert. This prevents clicks in the panels during the animation

I think we should keep the triggering panel non-inert. The main reason for this is so that the user's focus doesn't randomly get lost; it should stay on the button that they clicked on. The animation shouldn't take so long that the user has a chance to accidentally click on something else in the panel. After the animation is complete, the old panel can be `hidden`.

packages/itwinui-react/src/core/Panels/Panels.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  } = render(<SvgStar />);
+
+  const { container } = render(
+    <Panels.BackButton>
+      <SvgStar />
+    </Panels.BackButton>,
+  );
+
+  expect(container.querySelector('button svg')).toEqual(starIcon);
+});
```
smmr-dn, line 53: I'm not sure if it's necessary to have a test to make sure that the `Panels.BackButton` can go back and forth between panels. More specifically, it should go back to the previous panel and go forth to another panel next.

testing/e2e/app/routes/Panels/spec.ts
```diff
@@ -0,0 +1,254 @@
...
+test('should display initialActiveId as the initial panel', async ({
+  page,
+}) => {
+  await page.goto('/Panels');
+
+  await expect(page.locator('#root')).toBeVisible();
+  await expect(page.locator('#more-info')).not.toBeVisible();
+});
+
+test('should animate to the next panel when the trigger is clicked', async ({
```
smmr-dn, line 12: You're right, it makes more sense to highlight the playwright test for my comment above. I was just thinking if it was necessary to have a test where you switch back and forth between panels to make sure the `Back` button works properly.

packages/itwinui-react/src/core/Panels/Panels.test.tsx
smmr-dn: Continued in this comment https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1710687739.

packages/itwinui-css/src/panels/panels.scss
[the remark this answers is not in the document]
mayank99: Did you try using the scrolling approach I described in my previous comment? If so, did you run into any problems with it?

The general process would go something like:
0. Panel 1 is the only thing in DOM
1. When panel 2 is triggered, add it to DOM but keep it outside the scrollport
3. Use [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to move panel 2 into scrollport and panel 1 out of scrollport
4. Remove panel 1 from DOM

I made a diagram to illustrate how this process. I'm using the word "animation", but really it's just smooth scrolling.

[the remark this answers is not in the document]
mayank99: I was able to make it work with the help of scroll snapping. Here are the changes I made against the [current code](https://github.com/iTwin/iTwinUI/tree/39054001a7eb5bd0b6123d9d00e40b5b70b9f23f/packages):

<details><summary>Diff</summary>

```diff
diff --git a/packages/itwinui-css/src/panels/panels.scss b/packages/itwinui-css/src/panels/panels.scss
index f30e35683..d6f5ca122 100644
--- a/packages/itwinui-css/src/panels/panels.scss
+++ b/packages/itwinui-css/src/panels/panels.scss
@@ -2,12 +2,8 @@
 // See LICENSE.md in the project root for license terms and full copyright notice.
 
 .iui-panel-wrapper {
-  overflow: auto;
-  overflow-anchor: auto;
-
-  @media (prefers-reduced-motion: no-preference) {
-    scroll-behavior: smooth;
-  }
+  overflow: hidden;
+  scroll-snap-type: x proximity;
 }
 
 // Intentional selector to increase specificity
@@ -19,4 +15,8 @@
 .iui-panel {
   inline-size: 100%;
   flex-shrink: 0;
+
+  &:not([data-iui-transitioning='true']) {
+    scroll-snap-align: end;
+  }
 }
diff --git a/packages/itwinui-react/src/core/Panels/Panels.tsx b/packages/itwinui-react/src/core/Panels/Panels.tsx
index 042636d51..63a9255fe 100644
--- a/packages/itwinui-react/src/core/Panels/Panels.tsx
+++ b/packages/itwinui-react/src/core/Panels/Panels.tsx
@@ -23,7 +23,6 @@ import cx from 'classnames';
 import {
   PanelsInstanceContext,
   PanelsInstanceProvider,
-  panelsReducer,
   PanelsWrapperContext,
 } from './helpers.js';
 import type { PanelsInstance, TriggerMapEntry } from './helpers.js';
@@ -107,18 +106,7 @@ const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
 
   const ref = React.useRef<HTMLDivElement | null>(null);
 
-  const [{ activePanel, nextPanel }, dispatch] = React.useReducer(
-    panelsReducer,
-    {
-      activePanel: initialActiveId,
-      nextPanel: undefined,
-      onActiveIdChange,
-    },
-  );
-
-  React.useEffect(() => {
-    console.log('panels', activePanel, nextPanel);
-  }, [activePanel, nextPanel]);
+  const [activePanel, setActivePanel] = React.useState(initialActiveId);
 
   const [_panelElements, setPanelElements] = React.useState<
     Record<string, HTMLElement | null>
@@ -129,33 +117,6 @@ const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
     Record<string, TriggerMapEntry>
   >({});
 
-  React.useEffect(() => {
-    const refCurrent = ref.current;
-
-    const scrollEndListenerCurrent = () => {
-      dispatch({ type: 'end-panel-transition' });
-    };
-
-    const scrollListener = () => {
-      clearTimeout((window as any).scrollEndTimer);
-      (window as any).scrollEndTimer = setTimeout(
-        scrollEndListenerCurrent,
-        100,
-      );
-    };
-
-    if ('onscrollend' in window) {
-      refCurrent?.addEventListener('scrollend', scrollEndListenerCurrent);
-    } else {
-      refCurrent?.addEventListener('scroll', scrollListener);
-    }
-
-    return () => {
-      refCurrent?.removeEventListener('scrollend', scrollEndListenerCurrent);
-      refCurrent?.removeEventListener('scroll', scrollListener);
-    };
-  }, [activePanel, nextPanel]);
-
   const changeActivePanel = React.useCallback(
     async (newActiveId: string) => {
       if (
@@ -167,29 +128,15 @@ const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
         return;
       }
 
-      ReactDOM.flushSync(() => {
-        // setNextPanel(newActiveId);
-        dispatch({ type: 'start-panel-transition', nextPanel: newActiveId });
+      ReactDOM.flushSync(() => setActivePanel(newActiveId));
+      panelElements.current?.[newActiveId]?.scrollIntoView({
+        block: 'nearest',
+        inline: 'center',
+        behavior: 'smooth',
       });
 
-      // setTimeout(() => {
-      //   console.log('HERE', panelElements.current[activePanel]);
-      //   panelElements.current[activePanel]?.scrollIntoView({
-      //     behavior: 'instant',
-      //     block: 'nearest',
-      //     inline: 'center',
-      //   });
-      // }, 1_000);
-
       await new Promise((resolve) => {
         setTimeout(() => {
-          console.log('Scrolling', panelElements.current?.[newActiveId]);
-
-          panelElements.current?.[newActiveId]?.scrollIntoView({
-            block: 'nearest',
-            inline: 'center',
-          });
-
           // WIP Focus
           // const activePanelElement = document.getElementById(activePanel);
           // const newActiveElement = document.getElementById(newActiveId);
@@ -214,9 +161,8 @@ const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
           //     trigger?.focus({ preventScroll: true });
           //   }
           // }
-
           resolve(null);
-        }, 4_000);
+        });
       });
 
       // TODO: focus
@@ -247,14 +193,13 @@ const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
       value={React.useMemo(
         () => ({
           activePanel,
-          nextPanel,
           changeActivePanel,
           triggers,
           setTriggers,
           panelElements: panelElements.current,
           setPanelElements,
         }),
-        [activePanel, changeActivePanel, panelElements, nextPanel, triggers],
+        [activePanel, changeActivePanel, panelElements, triggers],
       )}
     >
       <PanelsInstanceProvider instance={instance}>
@@ -295,7 +240,7 @@ type PanelProps = {
 const Panel = React.forwardRef((props, forwardedRef) => {
   const { id: idProp, children, className, ...rest } = props;
 
-  const { activePanel, nextPanel, triggers, panelElements, setPanelElements } =
+  const { activePanel, triggers, panelElements, setPanelElements } =
     React.useContext(PanelsWrapperContext) || {};
 
   const fallbackId = React.useId();
@@ -328,9 +273,14 @@ const Panel = React.forwardRef((props, forwardedRef) => {
     [id, triggers],
   );
 
-  const isMounted = [activePanel, nextPanel].includes(id);
+  const previousActivePanel = useDelayed(activePanel);
+
+  const isMounted = [activePanel, previousActivePanel].includes(id);
+  const isTransitioning =
+    activePanel === id && activePanel !== previousActivePanel;
+
   // TODO: Being back inert checks
-  // const isInert = [nextPanel].includes(id);
+  const isInert = previousActivePanel === id && activePanel !== id;
 
   const refs = useMergedRefs(setElement, forwardedRef);
 
@@ -348,7 +298,8 @@ const Panel = React.forwardRef((props, forwardedRef) => {
           className={cx('iui-panel', className)}
           aria-labelledby={`${id}-header`}
           tabIndex={-1}
-          // {...{ inert: isInert ? '' : undefined }}
+          {...{ inert: isInert ? 'true' : undefined }}
+          data-iui-transitioning={isTransitioning ? 'true' : undefined}
           {...rest}
         >
           {children}
@@ -520,3 +471,22 @@ export const Panels = {
   BackButton: PanelBackButton,
   useInstance: useInstance as () => PanelsInstance,
 };
+
+/**
+ * Returns a copy of value which reflects state changes after a set delay.
+ */
+function useDelayed<T>(value: T, { delay } = { delay: 500 }): T | undefined {
+  const [delayed, setDelayed] = React.useState<T | undefined>(undefined);
+  const timeout = React.useRef<ReturnType<typeof setTimeout> | undefined>(
+    undefined,
+  );
+
+  React.useEffect(() => {
+    timeout.current = setTimeout(() => setDelayed(value), delay);
+    return () => {
+      clearTimeout(timeout.current);
+    };
+  }, [value, delay]);
+
+  return delayed;
+}
diff --git a/packages/itwinui-react/src/core/Panels/helpers.tsx b/packages/itwinui-react/src/core/Panels/helpers.tsx
index 1107cf971..f85d9f889 100644
--- a/packages/itwinui-react/src/core/Panels/helpers.tsx
+++ b/packages/itwinui-react/src/core/Panels/helpers.tsx
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import { useSynchronizeInstance } from '../../utils/index.js';
-import type { PanelsWrapperProps } from './Panels.js';
 
 export type PanelsInstance = {
   /** Go back to the panel that has a trigger that points to the current panel. */
@@ -19,7 +18,6 @@ export const PanelsWrapperContext = React.createContext<
   | {
       // TODO: Change names to like activePanelId. But maybe we can use elements directly without ids?
       activePanel: string;
-      nextPanel?: string;
       triggers: Record<string, TriggerMapEntry>;
       setTriggers: React.Dispatch<
         React.SetStateAction<Record<string, TriggerMapEntry>>
@@ -83,47 +81,3 @@ export const PanelsInstanceProvider = (props: PanelInstanceProviderProps) => {
     </PanelsInstanceContext.Provider>
   );
 };
-
-// ----------------------------------------------------------------------------
-
-type PanelsReducerState = {
-  activePanel: string;
-  nextPanel?: string;
-  onActiveIdChange?: PanelsWrapperProps['onActiveIdChange'];
-};
-
-type PanelsReducerAction =
-  | {
-      type: 'start-panel-transition';
-      nextPanel: string;
-    }
-  | {
-      type: 'end-panel-transition';
-    };
-
-export const panelsReducer = (
-  state: PanelsReducerState,
-  action: PanelsReducerAction,
-) => {
-  if (action.type === 'start-panel-transition') {
-    return {
-      ...state,
-      nextPanel: action.nextPanel,
-    };
-  }
-
-  if (action.type === 'end-panel-transition') {
-    if (state.nextPanel == null) {
-      return state;
-    }
-
-    const newActivePanel = state.nextPanel;
-
-    state.onActiveIdChange?.(newActivePanel);
-    return {
-      activePanel: newActivePanel,
-    };
-  }
-
-  return state;
-};

```


</details>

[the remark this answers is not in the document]
mayank99: > Is the 500ms arbitrary? I added a check in [`e0c8590`](https://github.com/iTwin/iTwinUI/pull/2001/commits/e0c8590cfd148e938165ac1a4753797000565b9d) to prevent a page change from happening even if the scroll has finished before 500ms.

Yes it is arbitrary, because the exact number doesn't matter. I'm not sure why you want to prevent a panel change; is there a problem with letting the user quickly switch back and forth between panels?

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,488 @@
...
+  const [triggers, setTriggers] = React.useState<
+    Record<string, TriggerMapEntry>
+  >({});
+  const triggersRef = useLatestRef(triggers);
+
+  const changeActivePanel = React.useCallback(
+    async (newActiveId: string) => {
+      if (
+        // Panel transition in progress
+        activePanel !== previousActivePanel ||
```
mayank99, line 133: Continued from https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1744044480: this check is not necessary, and the `previousActivePanel` state does not need to be stored in context because it is not "real" state - it's derived from `activePanel` in the one place where it is actually needed.

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,488 @@
...
+
+      await new Promise((resolve) => {
+        setTimeout(() => {
+          const prevElement = document.getElementById(previousActivePanel);
+          const activeElement = document.getElementById(newActiveId);
+
+          if (prevElement != null && activeElement != null) {
+            const isActivePanelAfterPrevPanelInDom =
+              prevElement?.compareDocumentPosition(activeElement) &
+              Node.DOCUMENT_POSITION_FOLLOWING;
```
mayank99, line 157: Such hacks should not be necessary at all. I'm not sure what you're comparing the DOM position like this when we already have a dedicated `goBack` method which is separate from the normal trigger.

packages/itwinui-css/src/panels/panels.scss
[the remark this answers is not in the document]
mayank99: Yes, I think it is fine to not animate for rapid clicks. This is exactly what happens in `Carousel` too. The important part is to always respect the user's clicks. This is why `activePanel` should be the only source of truth.

packages/itwinui-react/src/core/Panels/Panels.tsx
mayank99: It is well-supported, but it is a hack - this is an API that we should almost never need to use.

Here's how I think about it:
1. When a `Panel.Trigger` is clicked:
   - We should store this trigger element's ID (let's call it `triggeredBy`). This state is only relevant for the lifecycle of the Panel.
   - We move focus into the `Panel.Header`.
2. When `goBack` is called.
   - We move focus to `triggeredBy`. This should work for both `Panel.BackButton` and also for our consumer's custom `goBack` calls (e.g. they might call `goBack` after an element is selected from the second Panel).
   - We clear `triggeredBy`, since the Panel has reached the end of its lifecycle. It can be repopulated when triggered again (by the same trigger or a different trigger).

packages/itwinui-react/src/core/Panels/Panels.tsx
mayank99: > code duplication is alright here?

I don't see any code duplication? If you're talking about the fact that hook will be called separately for every individual `Panel`, then yes I think that's ok because it makes sense this state is only used by the panel and this code is not expensive to re-run.

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,558 @@
...
+  value: T,
+  {
+    delay,
+  }: {
+    /** Default delay is 500ms, or 0ms when prefers-reduced-motion */
+    delay?: number;
+  } = {},
+): T {
+  const motionOk = getWindow()?.matchMedia?.(
+    '(prefers-reduced-motion: no-preference)',
```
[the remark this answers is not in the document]
mayank99: > So where do you suggest I create `motionOk`? In the components that call `useDelayed()` and then make `delay` a required parameter, or something along those lines?

It should be called only where it's needed. In this case, the main `Panels` component would call it, since it only needs to be used for changing the `scrollIntoView` `behavior`. I think changing the `delay` might not be necessary (unconditionally leaving the panel in the DOM for 500ms should not cause any adverse effect).

> So, should we replace the `getWindow()?.matchMedia?.('(prefers-reduced-motion: no-preference)',)?.matches` calls in other components with `useMediaQuery()` in a future PR?

Yeah that makes sense to me.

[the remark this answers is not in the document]
mayank99: > Sounds good, made that change ([c296e05](https://github.com/iTwin/iTwinUI/commit/c296e0541cbb416cad8e52bcd0d5120c01331e32)).

Not sure if it's on purpose but the `delay` is now `5000` (instead of `500`) and the initial value returned by `useDelayed` is no longer `undefined`.

smmr-dn (commented):
https://github.com/user-attachments/assets/1e2e50bb-56dd-4b97-a3dd-a7f2c1d0b16f

I tried out the main/default example and noticed this. When I click the back button, it doesn't show the animation but delays for a little bit then the root panel suddenly appears. Not sure yet if this has something to do with the component code or just the example, or the behavior is intentional.

packages/itwinui-react/src/core/Panels/Panels.tsx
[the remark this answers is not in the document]
smmr-dn: I just used what you have in the `Panels.main.jsx` example on Chrome, Edge and Firefox (the video above in my original comment was Edge). The delay seems to happen in Edge as well. My OS is Windows. I have pulled the latest change as well.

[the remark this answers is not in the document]
mayank99: Just to rule out the possibility that something is wrong with your local setup, can you also test out the deploy preview at this URL? https://itwin.github.io/iTwinUI/2001/react/?story=panels--basic

I tried it just now on Edge in a Windows VM, and it works fine for me.

<details>

https://github.com/user-attachments/assets/bbc4db39-e17f-436d-9061-012edc02cc5b

</details>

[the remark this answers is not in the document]
smmr-dn: I have tested on different browsers and you're right, there must be something wrong with my local setup. Things work fine on the deploy preview so I'm closing this comment.

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,568 @@
...
+
+  const ref = React.useRef<HTMLDivElement | null>(null);
+
+  const [activePanelId, _setActivePanelId] = React.useState<string | undefined>(
+    undefined,
+  );
+  const setActivePanelId = React.useCallback(
+    (newActivePanel: NonNullable<typeof activePanelId>) => {
+      _setActivePanelId(newActivePanel);
+      onActiveIdChange?.(newActivePanel);
```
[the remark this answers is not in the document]
mayank99: Can we use `useControlledState` here, or does it not work?

Side note: I'm realizing that this kind of code will lead to a chain of invalidations if the user doesn't memoize the callback passed to `onActiveIdChange`. We should use `useLatestRef` here (and probably also inside `useControlledState`).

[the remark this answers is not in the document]
mayank99: Ah yes, that's probably why didn't use `useControlledState` here initially. So I guess let's revert it and be explicit about it, like Uyen suggested. We can keep the `useLatestRef`.

packages/itwinui-react/src/utils/hooks/useInstance.ts
mayank99, line 1: Do you remember where we landed on this `useInstance` hook? Iirc, it was supposed to a temporary solution, because it still feels like a hacky, roundabout way of exposing imperative methods.

packages/itwinui-react/src/index.ts
```diff
@@ -111,6 +111,8 @@ export { NotificationMarker } from './core/NotificationMarker/NotificationMarker
 
 export { Overlay } from './core/Overlay/Overlay.js';
 
+export { Panels } from './core/Panels/Panels.js';
```
mayank99, line 114: Do we want to release this component as "unstable" to allow for API changes based on feedback?

Just the export name can be changed:

```suggestion
export { Panels as unstable_Panels } from './core/Panels/Panels.js';
```

And then it would be used as:

```jsx
import { unstable_Panels as Panels } from "@itwin/itwinui-react";
```

mayank99: CI is failing. All the imports will need to be updated.

Also, we will need docs on day one when we release this. You can start working on it while I'm taking a final look at the PR.

packages/itwinui-react/src/utils/hooks/useInstance.ts
mayank99: Oh yes, we want an Ariakit-like store implementation. We've been using Ariakit in Kiwi, and find it quite comfortable.

That's why we had chosen this `useInstance` hook + `instance` prop API. Even if we change the underlying implementation, the API will likely remain the same. So I think we're good for now.

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -0,0 +1,580 @@
...
+        () => ({ id, associatedTrigger }),
+        [associatedTrigger, id],
+      )}
+    >
+      {isMounted && (
+        <Box
+          ref={forwardedRef}
+          id={id}
+          className={cx('iui-panel', className)}
+          aria-labelledby={`${id}-header-title`}
```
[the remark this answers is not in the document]
mayank99: It might work in VoiceOver but it's not something we can take for granted to work everywhere, because labeling generic elements is not allowed.

The addition of `role="group"` looks good now. 👍

examples/Panels.main.jsx
```diff
@@ -0,0 +1,43 @@
...
+  Text,
+} from '@itwin/itwinui-react';
+
+export default () => {
+  const panelIdRoot = 'root';
+  const panelIdMoreInfo = 'more-info';
+
+  return (
+    <Panels.Wrapper as={Surface} className='demo-panels-wrapper'>
+      <Panels.Panel id={panelIdRoot} as={Surface} border={false} elevation={0}>
```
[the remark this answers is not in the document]
mayank99: I'll leave this comment open, in case we want to revisit this decision in the future  (I'm sure some edge cases will come up).

mayank99 (approved):
:shipit:

.changeset/sour-parrots-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added a new `Panels` component for easy setup of nested screens/panels. Example uses: multi-layered menus, wizards, settings screens, etc.
```
mayank99, line 5: We should mention that it's unstable in the changelog.

```suggestion
Added a new `unstable_Panels` component for easy setup of nested screens/panels. Example uses: multi-layered menus, wizards, settings screens, etc.
```

smmr-dn (approved):

packages/itwinui-react/src/core/Panels/Panels.tsx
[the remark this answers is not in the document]
mayank99: Like I said, it looks good. I was just clarifying that your assumption about `role="group"` being redundant might not be universally correct. We should definitely test on other screen-readers too.

Resolving.

PR #2009, opened 2024-04-17
correctly portal `Popover` into the main `portalContainer`

### LINKED ISSUE #2006: Table filter alignment issue for popup window on 3.x

We are experiencing problems with filter alignment when upgrading from 2.x to 3.x in a popup window that has a table component.

image

### Expected Behavior

Popover to table filter should be align 

### Link to minimal repro

https://codesandbox.io/p/sandbox/popout-window-table-5gw8sc?file=%2Fsrc%2FApp.tsx

### Steps To Reproduce

_No response_

### Anything else?

_No response_

mayank99:
## Changes

Fixes #2006

The main issue was the use of [`FloatingPortal`](https://floating-ui.com/docs/FloatingPortal) without `root` prop (defaults to `document.body` in the original window). I changed it to use the correct root element from `usePortalTo` (and therefore was also able to remove `Portal`).

Tangential changes:
- Added `display: contents` to extra wrapper divs to avoid potentially breaking layout.
- Added similar performance improvements as Tooltip from #2004

Also see comment about TimePicker/ColorPicker: https://github.com/iTwin/iTwinUI/pull/2009#discussion_r1569266375

## Testing

Tested by recreating the linked issue in local playground. The filter popover appears in the correct window.

image

Added a unit test for it as well.

Unrelated: I noticed we didn't have any existing unit tests for `Popover`, so I added a few. Can be improved further in the future.

## Docs

Added changeset.


apps/react-workshop/cypress-visual-screenshots/baseline/TimePicker.test.ts-Custom Renderers.png
mayank99, line 1: I'm not sure why this one changed. It _is_ using `Popover`, but there shouldn't be any difference in functionality.

mayank99: Some other images also changed ([`f27e696`](https://github.com/iTwin/iTwinUI/pull/2009/commits/f27e696c306a9652f89a78d107da141dac5b372c)), it's really strange.

Notice the double border around the swatch button.



It looks like when portaling was fixed, the floating focus manager started to work as intended, resulting in some random DOM elements being added next to the trigger. See [docs](https://floating-ui.com/docs/floatingfocusmanager#guards).



I would like to remove these DOM elements, but then it breaks focus management.

packages/itwinui-react/src/core/Popover/Popover.test.tsx
rohan-kadkol, line 1: non-blocking: I guess we can also test for cases like the controlled case. Since this PR is not aiming to write exhaustive tests, we can do it later too.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -294,36 +307,68 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
...
+    <FloatingPortal root={portalTo}>
+      <DisplayContents />
+      {children}
+    </FloatingPortal>
+  );
+};
+
+// ----------------------------------------------------------------------------
+
+/** Applies `display: contents` to the parent div. */
```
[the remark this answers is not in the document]
mayank99: The constraint here is that the parent needs to be one of the [elements that you can attach a shadow-root to](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#elements_you_can_attach_a_shadow_to), so it's not "regardless".

`div` is the most general case and it satisfies both usages of `DisplayContents` within this file. It's not important though, as this component is not exported.

packages/itwinui-react/src/core/Popover/Popover.test.tsx
mayank99: Yes, we should add more tests in the future.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -294,36 +307,68 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
...
+// ----------------------------------------------------------------------------
+
+const PopoverPortal = ({
+  children,
+  portal = true,
+}: React.PropsWithChildren<PortalProps>) => {
+  const portalTo = usePortalTo(portal);
+
+  return (
+    <FloatingPortal root={portalTo}>
```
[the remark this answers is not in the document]
mayank99: `FloatingPortal` is necessary for `FloatingFocusManager` (and maybe some other floating-ui features) to work.

`Portal` only does one thing, whereas `FloatingPortal` does everything `Portal` does (and more).

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -294,36 +307,68 @@ export const Popover = React.forwardRef((props, forwardedRef) => {
...
+
+// ----------------------------------------------------------------------------
+
+/** Applies `display: contents` to the parent div. */
+const DisplayContents = React.memo(() => {
+  return (
+    <ShadowRoot
+      css={`
+        :host {
+          display: contents;
```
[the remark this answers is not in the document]
mayank99: [`display: contents`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box) makes it so that the element becomes "see-through" for the purposes of layout and sizing.

So it shouldn't affect 99% of the cases, but I'm thinking of this one very specific case that I want to continue to support. If the user uses `flex` or `grid` layout on the ThemeProvider, the portaled elements shouldn't participate in the flex/grid layout.

```js
<ThemeProvider style={{ display: "grid" }}>
  <Header>…</Header>
  <SideNavigation>…</SideNavigation>
  <main>…</main>
  <Footer>…</Footer>
</ThemeProvider>
```

apps/react-workshop/cypress-visual-screenshots/baseline/TimePicker.test.ts-Custom Renderers.png
mayank99: I investigated the TimePicker issue a bit more and found that focus/scroll was not behaving correctly. Fixed in [`0438fa9`](https://github.com/iTwin/iTwinUI/pull/2009/commits/0438fa91a16210856dfea2fac406dae72e3db6ca).

We can leave the ColorPicker/ButtonGroup images as-is in this PR, since it is a pretty minor visual issue. The focus is behaving correctly.

apps/react-workshop/src/ColorPicker.stories.tsx
```diff
@@ -81,9 +81,10 @@ export const Basic = () => {
           </ColorPicker>
         }
         visible={isOpen}
+        onVisibleChange={setIsOpen}
         placement='bottom-start'
       >
-        <IconButton onClick={() => setIsOpen((open) => !open)}>
+        <IconButton>
```
rohan-kadkol, line 87: Should we make this change in `ColorPicker.advancedPopover.jsx` too?

rohan-kadkol (approved):

AdamMeza-Bentley (approved):

apps/react-workshop/src/ColorPicker.stories.tsx
mayank99: (Missed this comment somehow)
Yes, we should fix it, but honestly, all `ColorPicker`

mayank99: (Missed this comment somehow)

Yes, we should fix it, but honestly, all `ColorPicker` examples need to be updated to use `Popover`.

PR #2010, opened 2024-04-18
Add [x]Props to ToggleSwitch

AdamMeza-Bentley:
## Changes

Continues work on #1368 by adding optional `labelProps` to `ToggleSwitch`. I debated including a ``wrapperProps` as well but decided against it. I can always add those if they're wanted as well. 

## Testing

Added React testing for new `labelProps` functionality 

## Docs

Added a changeset


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -101,7 +108,11 @@ export const ToggleSwitch = React.forwardRef((props, ref) => {
         </Box>
       )}
       {label && (
-        <Box as='span' className='iui-toggle-switch-label'>
+        <Box
+          as='span'
+          className={cx('iui-toggle-switch-label', labelClassName)}
+          {...restLabelProps}
```
mayank99, line 114: Instead of creating `restLabelProps` like this, you can reorder this and use `labelProps` directly.

```suggestion
          {...labelProps}
          className={cx('iui-toggle-switch-label', labelProps?.className)}
```

AdamMeza-Bentley: adressed this in my last push

mayank99: That's not the same thing. You are still destructuring `className` from `labelProps`, which is unnecessary because you can just use `labelProps.className`.

AdamMeza-Bentley: gotcha. missed that part. fixed it

mayank99 (approved):

PR #2011, opened 2024-04-18
 fix `portalContainer` inheritance across different windows

mayank99:
## Changes

Follow-up to #2009. This attempts to reset `portalContainer` when it is in a different (popout) window than the `ThemeProvider` itself.

This is how it should work:

- When `portalContainer` _prop_ is passed, use it as-is.
- When inheriting theme, try to find the topmost portal container (recursively inheriting).
  - **If the inherited portal container is in a different window from the `ThemeProvider`, create a DOM element.** (This PR)
- If none of the above apply, create a new DOM element instead of `createPortal`ing into an existing element.
- Store the portal container for use in descendants.

### Implementation notes

After some iteration, I ultimately decided to add [`jotai`](https://jotai.org/), which is something we had already planned on adding ever since v3 was released. I created a thin wrapper over `useAtom` called `useScopedAtom` that falls back to reading the value from the nearest context if it hasn't been set in the current context. This unlocks the main inheritance functionality in this PR, while avoiding conflicts with global/application state. This wrapper is completely abstracted away in `ScopeProvider`, so the most interesting changes are in other files:
- `portalContainerAtom` now lives in and is exported from `Portal.tsx` and does not need to be stored in `ThemeContext`. (This is a great example of how `jotai`+`ScopeProvider` allows colocating code where it makes the most sense)
- `PortalContainer` is a helper component in `ThemeProvider.tsx` that handles everything, with the exception of `portalContainerFromParent`, which needs to be passed by `ThemeProvider`.
  - This exception is because `useScopedAtom` uses the parent scope in `ThemeProvider` but uses the current scope in `PortalContainer` (since it is a child of `ThemeProvider`'s own `ScopeProvider`).
- `ownerDocumentAtom` is a private atom only used in `ThemeProvider.tsx`. The read/write are separated using `useScopedAtom`/`useScopedSetAtom` in different components.

## Testing

Added an e2e test, and moved some existing unit tests to e2e for more reliability.

Manually tested in vite-playground using this code:

```tsx
import { useState } from 'react';
import {
  Button,
  Table,
  ThemeProvider,
  tableFilters,
import type { Column } from '@itwin/itwinui-react/react-table';
import { createPortal } from 'react-dom';

  const [popoutDocumentBody, setPopoutDocumentBody] =
    useState<HTMLElement | null>(null);

  const showPopOutWindow = () => {
    const childWindow = window.open(
      '',
      '_blank',
      'width=1280,height=720,menubar=no,resizable=yes,scrollbars=no,status=no,location=no',
    );

    setTimeout(() => {
      if (!childWindow) {
        console.log('childWindow is null??');
        return;
      }
      childWindow.document.title = 'Popout window!';
      childWindow.document.write(htmlShell);
      copyStyles(childWindow.document);
      setPopoutDocumentBody(childWindow.document.body);
    });
  };

  return (
    <div className='App'>
      <Button styleType='high-visibility' onClick={() => showPopOutWindow()}>
        Click to Show Pop out Window
      </Button>
      {popoutDocumentBody &&
        createPortal(
          <ThemeProvider>
            <Table
              columns={columns}
              data={data}
              emptyTableContent={'No Data'}
              isSortable={true}
              isSelectable={true}
              autoResetPage={false}
            />
          </ThemeProvider>,
          popoutDocumentBody,
        )}
    </div>
  );
}

/**
 * Copies the source CSS into the destination
 * @param targetDoc - target document
 * @param sourceDoc - source document
 * @internal
 */
export function copyStyles(
  targetDoc: Document,
  sourceDoc: Document = document,
) {
  const stylesheets = Array.from([
    ...sourceDoc.styleSheets,
    ...sourceDoc.adoptedStyleSheets,
  ]);
  stylesheets.forEach((stylesheet) => {
    const css = stylesheet;
    if (stylesheet.href) {
      const newStyleElement = targetDoc.createElement('link');
      newStyleElement.rel = 'stylesheet';
      newStyleElement.href = stylesheet.href;
      targetDoc.head.appendChild(newStyleElement);
    } else {
      if (css && css.cssRules && css.cssRules.length > 0) {
        const newStyleElement = targetDoc.createElement('style');
        Array.from(css.cssRules).forEach((rule) => {
          newStyleElement.appendChild(targetDoc.createTextNode(rule.cssText));
        });
        targetDoc.head.appendChild(newStyleElement);
      }
    }
  });

  // copy sprites
  const svgSymbolParent = sourceDoc.getElementById('__SVG_SPRITE_NODE__');
  if (svgSymbolParent) {
    targetDoc.body.appendChild(svgSymbolParent.cloneNode(true));
  }
}

const htmlShell = `<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8" />
  <style>
    body {
      margin: 0;
      height: 100dvh;
      overflow: auto;
    }
  </style>
</head>
<body></body>
</html>`;

const data = [
  {
    product: 'Product 1',
    quantity: 500,
    rating: '4/5',
  },
  {
    product: 'Product 2',
    quantity: 1200,
    rating: '1/5',
  },
  {
    product: 'Product 3',
    quantity: 1500,
    rating: '3/5',
  },
  {
    product: 'Product 4',
    quantity: 50,
    rating: '4/5',
  },
  {
    product: 'Product 5',
    quantity: 700,
    rating: '5/5',
  },
  {
    product: 'Product 6',
    quantity: 30,
    rating: '5/5',
  },
  {
    product: 'Product 7',
    quantity: 130,
    rating: '1/5',
  },
  {
    product: 'Product 8',
    quantity: 500,
    rating: '4/5',
  },
];

const columns = [
  {
    id: 'product',
    Header: 'Product',
    accessor: 'product',
    minWidth: 400,
    Filter: tableFilters.TextFilter(),
    fieldType: 'text',
  },
  {
    id: 'quantity',
    Header: 'Quantity',
    accessor: 'quantity',
    width: 400,
    Filter: tableFilters.TextFilter(),
    fieldType: 'text',
  },
  {
    id: 'rating',
    Header: 'Rating',
    accessor: 'rating',
    width: 400,
    Filter: tableFilters.TextFilter(),
    fieldType: 'text',
  },
] satisfies Column<(typeof data)[0]>[];

```

</details> 

The most important difference between this and the playground for the previous PR (#2009) is this:

```diff
 {popoutDocumentBody &&
   createPortal(
-    <ThemeProvider portalContainer={popoutDocumentBody}>
+    <ThemeProvider>
       …
     </ThemeProvider>,
     popoutDocumentBody,
   )
 }
```

(The user doesn't need to specify `portalContainer` anymore because this PR makes it so the correct window is used automatically)

## Docs

Added changesets.


packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -293,6 +300,63 @@ const useParentThemeAndContext = (rootElement: HTMLElement | null) => {
...
+    // Create a new portal container only if necessary:
+    // - not inheriting theme
+    // - no parent portal container to portal into
+    // - parent portal container is in a different window (#2006)
+    if (
+      !portalContainerProp && // bail if portalContainerProp is set, because it takes precedence
+      (!isInheritingTheme ||
+        !portalContainerFromParent ||
+        portalContainerFromParent.ownerDocument !== ownerDocument)
+    ) {
```
rohan-kadkol, line 342: If we're creating a new `portalContainer` div, I see that we're calling `setPortalContainer` to set the portal container in the store.

But if we're not creating a new `portalContainer` div, should we call `setPortalContainer(portalContainerFromParent)`? Because without it, the third level `ThemeProvider` will create a new `portalContainer` div since it sees that the second level `ThemeProvider` has no `portalContainer` set in its store. I think the correct thing would be for the third level `ThemeProvider` to reuse the div from the first level `ThemeProvider`.

So maybe something like:

```diff
if ( … ) {
  return <div>;
} else {
  setPortalContainer(portalContainerFromParent)
}

…
```

<details>
<summary>Some testing code I tried (App.tsx)</summary>

```jsx
import { Button, ThemeProvider, Tooltip } from '@itwin/itwinui-react';

const App = () => {
  return (
    <>
      <ThemeProvider theme='inherit'>
        <ThemeProvider theme='inherit'>
          <ThemeProvider theme='inherit'>
            <Tooltip content='Tooltip content'>
              <Button>Hello world</Button>
            </Tooltip>
          </ThemeProvider>
        </ThemeProvider>
      </ThemeProvider>
    </>
  );
};

export default App;
```

</details>

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -150,55 +163,45 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
...
           </ToastProvider>
-        </Root>
-      </ThemeContext.Provider>
-    </HydrationProvider>
+        </ThemeContext.Provider>
+      </HydrationProvider>
+    </ScopeProvider>
   );
 }) as PolymorphicForwardRefComponent<'div', ThemeProviderOwnProps>;
+ThemeProvider.displayName = 'ThemeProvider';
```
rohan-kadkol, line 204: nit: Was this necessary for some debugging? Is the `displayName` not set automatically?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -293,6 +300,63 @@ const useParentThemeAndContext = (rootElement: HTMLElement | null) => {
...
+    isInheritingTheme: boolean;
+  }) => {
+    const [ownerDocument] = useScopedAtom(ownerDocumentAtom);
+    const [portalContainer, setPortalContainer] =
+      useScopedAtom(portalContainerAtom);
+
+    // Synchronize atom with prop changes
+    if (portalContainerProp && portalContainerProp !== portalContainer) {
+      setPortalContainer(portalContainerProp);
+    }
```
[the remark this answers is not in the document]
mayank99: No, `useEffect` should be avoided as much as possible. See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

.changeset/spicy-buckets-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+Added dependency on `jotai`.
```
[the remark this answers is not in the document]
mayank99: Yes. The consumer (application) needs to be aware of which dependencies we require, because they are the ones managing all deps in the end. One example of an effect: If they already use `jotai` in their application, they will likely want to make sure that only one copy of `jotai` is included in the final client bundle.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: This should already be handled by the `useScopedAtom` inheritance logic? I will double check

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: Yes, it shows up as `ForwardRef` in React dev tools

packages/itwinui-react/src/utils/providers/ScopeProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+    >
+      {children}
+    </ScopeContext.Provider>
+  );
+};
+
+/**
+ * Wrapper over `useAtom` that uses the store from the nearest `ScopeProvider`.
+ *
+ * If the atom is not set in the current store, it will look for the atom in the parent store.
```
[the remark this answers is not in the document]
mayank99: good point, I changed it so that it actually `set`s the value instead of only returning it. [`9a66baf`](https://github.com/iTwin/iTwinUI/pull/2011/commits/9a66baf63bcbc0750f3e88c981b63cdfa11f70cb)

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
rohan-kadkol: Yes, looks like after [`9a66baf`](https://github.com/iTwin/iTwinUI/pull/2011/commits/9a66baf63bcbc0750f3e88c981b63cdfa11f70cb) from https://github.com/iTwin/iTwinUI/pull/2011#discussion_r1583370317, this issue is fixed.

packages/itwinui-react/src/utils/providers/ScopeProvider.tsx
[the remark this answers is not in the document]
mayank99: Pushed another change to make it more clear that only initial value is inherited + manually synchronize future updates. [`a1e2845`](https://github.com/iTwin/iTwinUI/pull/2011/commits/a1e2845d7745714920c0e625af218bca45a239e5)

While the future updates are not super important for this particular PR (because the inherited portal container shouldn't change), `useScopedAtom` is a general utility so this pattern might be important in the future.

rohan-kadkol (approved):

PR #2012, opened 2024-04-19
Updates `Button` Astro Component

Ben-Pusey-Bentley:
## Changes

Updates the `Button` Astro Component with various changes requested in the comments of this [PR](https://github.com/iTwin/iTwinUI/pull/1987). These changes include:

- Renaming the `active` prop to `isActive`
- Creating an `IconButton` component and moving the `shift` prop onto that component
- Removing the `iuiType` prop. The `idea` value has been added to `variant`, while `dropdown` has been made a separate boolean prop
- Adds `type='button'` to both the button components
- Inserts a `span` into the `Button` component for holding button text
- Uses named slots in both of the button components so that button icons no longer need the `iui-button-icon` class in order to be styled correctly

## Testing

Ran css visual tests to make sure that there were no unintended visual changes.

## Docs

N/A


apps/css-workshop/src/pages/tag.astro
```diff
@@ -22,31 +22,24 @@ import Button_ from '../components/Button.astro';
...
-      <Button_
-        class='iui-tag-button'
-        variant='borderless'
-        size='small'
-        aria-label='Delete tag'
-        type='button'
-      >
-        <svg-close-small class='iui-button-icon' aria-hidden='true'></svg-close-small>
-      </Button_>
+      <IconButton_ class='iui-tag-button' variant='borderless' size='small' aria-label='Delete tag'>
```
AdamMeza-Bentley, line 25: This file and the others that use the Tag component should use that instead of this.

Ben-Pusey-Bentley: Can you not just keep your original changes in the case of a merge conflict and then edit the Tag component with the updated Button component? I can just hold the merge until the Tag component is done, but I don't think the conflict here would be too difficult to solve without having to redo work.

apps/css-workshop/src/components/IconButton.astro
```diff
@@ -0,0 +1,29 @@
...
+<button
+  class:list={['iui-button', className]}
+  type='button'
+  data-iui-variant={variant}
+  data-iui-size={size}
+  data-iui-active={isActive ? 'true' : undefined}
+  data-iui-shift={shift}
+  {...props}
+  ><slot name='button-icon' />
+</button><script>
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Doing that adds an extra space after the component, which causes visual changes.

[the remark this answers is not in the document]
AdamMeza-Bentley: that makes sense

apps/css-workshop/src/pages/tag.astro
AdamMeza-Bentley: All I know is that last time I had to redo everything when I updated my branch to be in sync with main. You're probably right though

mayank99: #1966 is merged now, make sure to resolve merge conflicts

apps/css-workshop/src/components/Button.astro
```diff
@@ -1,22 +1,40 @@
 ---
 type Props = {
-  variant?: 'high-visibility' | 'borderless' | 'cta';
+  variant?: 'high-visibility' | 'borderless' | 'cta' | 'idea';
```
mayank99, line 3: Maybe `idea` could be another boolean prop similar to `dropdown`?

Otherwise you'll have to add extra conditional logic to account for the CSS changes in #1919

apps/css-workshop/src/components/Button.astro
```diff
@@ -1,22 +1,40 @@
...
+<script>
+  const buttonIcons = [
+    ...document.querySelectorAll<HTMLInputElement>('[slot="start-icon"]'),
+    ...document.querySelectorAll<HTMLInputElement>('[slot="end-icon"]'),
+    ...document.querySelectorAll<HTMLInputElement>('.iui-notification-marker'),
+  ];
+  buttonIcons.forEach((icon) => {
+    icon.classList.add('iui-button-icon');
+  });
+</script>
```
mayank99, line 40: You should never need a client-script to generate HTML. Instead, just create a wrapping span with `iui-button-icon` and `aria-hidden`, similar to how it is done in react (see [`Button.tsx`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/Buttons/Button.tsx#L89-L92)).

apps/css-workshop/src/pages/tag.astro
Ben-Pusey-Bentley: Okay, will do.

apps/css-workshop/src/components/IconButton.astro
```diff
@@ -0,0 +1,29 @@
...
+  const buttonIcons = [
+    ...document.querySelectorAll<HTMLInputElement>('[slot="button-icon"]'),
+    ...document.querySelectorAll<HTMLInputElement>('button.iui-button>span[aria-hidden="true"]'),
+  ];
+  buttonIcons.forEach((icon) => {
+    icon.classList.add('iui-button-icon');
+  });
+</script>
```
mayank99, line 29: Same comment about client-side script. Use a wrapping `<span>` instead.

apps/css-workshop/src/components/IconButton.astro
```diff
@@ -0,0 +1,29 @@
...
+
+<button
+  class:list={['iui-button', 'iui-button-base', className]}
+  type='button'
+  data-iui-variant={variant}
+  data-iui-size={size}
+  data-iui-active={isActive ? 'true' : undefined}
+  data-iui-shift={shift}
+  {...props}
+  ><slot name='button-icon' />
```
mayank99, line 20: This one should use the default slot instead of a named slot.

The usage of this component should look like this (notice the lack of extra attributes on the svg):

```jsx
<IconButton_ variant='borderless'>
  <svg-close-small></svg-close-small>
</IconButton_>
```

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Yeah, that makes sense because it will also be a class. Will change.

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Alright, will replace.

apps/css-workshop/src/components/IconButton.astro
Ben-Pusey-Bentley: Will replace this too.

apps/css-workshop/src/components/IconButton.astro
Ben-Pusey-Bentley: Will remove.

apps/css-workshop/src/pages/tag.astro
Ben-Pusey-Bentley: Merge conflicts resolved, and added `IconButton` to `Tag` in place of the `Button` component.

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Added in this [commit](https://github.com/iTwin/iTwinUI/pull/2012/commits/3de64f92c67bfd85550588ad938ada78cb1dc68a).

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Done in this [commit.](https://github.com/iTwin/iTwinUI/pull/2012/commits/b692e8a2547028a6c1c0ea1f40ade2c9795ffb0c)

apps/css-workshop/src/components/IconButton.astro
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2012/commits/2391a14be3c867e3f2d5f137ba6b53ea4feeef6d).

apps/css-workshop/src/components/Button.astro
mayank99: Why do you have `iui-button-icon` in some places on the button page?

apps/css-workshop/src/components/IconButton.astro
mayank99: You can remove `aria-hidden` from the various places where `IconButton_` is used, because it should already be handled within the main component.

apps/css-workshop/src/components/Button.astro
Ben-Pusey-Bentley: Removing the `iui-button-icon` class from the top level `iui-notification-marker` components caused the icon to shift up. However, testing again, it seems that the issue was being caused by removing that class while leaving the `iui-button-icon` on the interior `svg` since removing both does not cause the shift. Will update to remove the classes.

Ben-Pusey-Bentley: Removed in this [commit.](https://github.com/iTwin/iTwinUI/pull/2012/commits/5fb02d4fabeede1ab5c06df2bb9c5b3fb5e156ec)

apps/css-workshop/src/components/IconButton.astro
Ben-Pusey-Bentley: Okay, will remove.

Ben-Pusey-Bentley: Removed in this [commit](https://github.com/iTwin/iTwinUI/pull/2012/commits/718becec0a4308c59c77764db16f0bb98ebde9e8).

apps/css-workshop/src/pages/breadcrumbs.astro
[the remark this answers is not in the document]
Ben-Pusey-Bentley: No, that `.iui-button-icon` is not an accidental leftover. The wrapping `<button>` element does not have the class `iui-button`, so it has not been replaced with a `Button` component. As a result, the icon still needs to have it's class defined. Additionally, the [input](https://github.com/iTwin/iTwinUI/blob/718becec0a4308c59c77764db16f0bb98ebde9e8/apps/css-workshop/src/pages/input.astro#L399-L405) and [information-panel](https://github.com/iTwin/iTwinUI/blob/718becec0a4308c59c77764db16f0bb98ebde9e8/apps/css-workshop/src/pages/information-panel.astro#L187-L190) both have `iui-button-icon` declarations that are not wrapped in button components, and as a result, are necessary. [carousel](https://github.com/iTwin/iTwinUI/blob/718becec0a4308c59c77764db16f0bb98ebde9e8/apps/css-workshop/src/pages/carousel.astro#L444) has some being inserted into `IconButton` components through a script, so those will be removed. Could not find any declarations of `iui-button` on elements outside of the component files.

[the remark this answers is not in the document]
Ben-Pusey-Bentley: Removed the extra declarations in carousel with this [commit](https://github.com/iTwin/iTwinUI/pull/2012/commits/9cc874aa204e1f94409a5540ece88ffe49ce57c6).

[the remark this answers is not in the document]
mayank99: > The wrapping `<button>` element does not have the class `iui-button`

Sounds good for this PR. But I think after #1919, it *should* be replaced with `Button`.

mayank99 (approved):

AdamMeza-Bentley (approved):

Issue #1972, opened 2024-04-03
[Table] Expandable row selection toggle works incorrectly when there are disabled rows under it


juliusBaranauskas:
### Describe the bug (current behavior)

Expandable row inside a Table, that has disabled and enabled rows as its' sub-rows handles selection toggling incorrectly.

### Expected Behavior

That seems incorrect firstly because disabled rows (in our team's opinion) should not be counted when checking whether all sub-rows are selected, therefore the **checkbox state in Step 3 should have been that all sub-rows are selected**. Secondly, toggling the selection in Step 4 should in-turn disable all selected rows that are sub-rows of the expandable row.

### Link to minimal repro

https://itwin.github.io/iTwinUI/react/?story=table--full2

### Steps To Reproduce

1. Use Table component with selectable rows and expandable rows allowed
2. Add a row with at least one sub-row enabled and one disabled
3. Toggle selection of the expandable row (notice a minus icon appear inside the checkbox rectangle, indicating some, not all sub-rows are selected)
![image](https://github.com/iTwin/iTwinUI/assets/22618356/5c4d1520-0d83-40e2-a41c-6749b5fbafb3)

5. Toggle selection of the same expandable row again (notice that the selection stays the same but minus icon disappears)
![image](https://github.com/iTwin/iTwinUI/assets/22618356/4b401f94-ebc8-4076-a54a-01713a022cb9)

`Product 4` row in the repro link can be used as an example.

mayank99:
I think this issue might be the same as #928?

juliusBaranauskas:
@mayank99 seems similar in some regards of the issue, although in that case the reason for why not all sub-rows are selected is because they are filtered out, however they are still _selectable_. Where I see the difference - in my case the rows are not even selectable so I'd argue that since all of the _selectable_ rows are selected, the _parent_ row selection should indicate that they are **all** selected, whereas in the issue you mentioned some rows are just hidden, not selected, but they are still selectable (and some might even be selected off-screen before filtering happened) so the initial state seems correct for that scenario.

mayank99:
It's unlikely that we will change the parent row to be (visually) selected when some rows are not selected. It might fit your narrow use-case (in which case I think you might be able to set manual state?), but our Table is designed to support a wide variety of scenarios. Think of a case where some disabled rows might be pre-selected; these should be taken into account when calculating the total selected state.

[...]

Issue #928, opened 2021-08-19
Table: clicking on row's indeterminate checkbox makes it empty/false


bentleyvk:
![table-subrows-filtered-selection-bug](https://user-images.githubusercontent.com/36186912/130052947-3e215b69-ceb7-420f-a6c4-2a395342780d.gif)

##### Steps to reproduce
1. Filter table that only part of sub-rows are visible.
1. Press on parent row's checkbox. It now has indeterminate state.
1. Press on the same checkbox. Checkbox is in empty/false state.

#### Additional information
The cause of this is `TableRowMemoized` which doesn't re-render because nothing really changed and checkbox does what it naturally should do - deselects.

### Possible fix
Either remove memoization or use virtual scroll.
Or there is a hack that makes checkbox to always rerender:
![image](https://user-images.githubusercontent.com/36186912/130053444-b8f73512-b019-4b20-97d6-c0a4705dce5c.png)

mayank99:
This is finally fixed in [3.10.0](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.10.0) 🥳

PR #2013, opened 2024-04-23
Fixes `Table` row selection when sub rows are disabled or filtered

Ben-Pusey-Bentley:
## Changes

Fixes `Table` row selection when sub rows are disabled or filtered out by adding an `onChange` handler to the row `Checkbox` for the `selectionColumn`, similar to the one used on the header row. This on change handler applies different toggle logic to rows that have sub rows, where if all selectable(e.g. non disabled and not filtered out) rows are selected, then clicking the row deselects all sub rows. Otherwise, all sub rows that can be selected will become selected. Previously, it would only deselect sub rows if all sub rows were selected, but this behavior breaks when some sub rows are unselected and disabled or filtered out of the table, making them unable to be selected.

Screen recordings of new behavior:

https://github.com/iTwin/iTwinUI/assets/141063311/8d0fda9c-9f24-4731-9bd8-4b1f256b2409

https://github.com/iTwin/iTwinUI/assets/141063311/47991883-d8e8-4837-b037-96e4196a4a68

## Testing

Tested in vite-playground with this [code](https://github.com/iTwin/iTwinUI/blob/ed0b545cef47db320f621f43efbe29a24078e409/playgrounds/vite/src/App.tsx#L1), and added an e2e test to check for these issues, as well as some basic row selection tests.

## Docs

Added patch changeset.


AdamMeza-Bentley (approved):

packages/itwinui-react/src/core/Table/columns/selectionColumn.tsx
```diff
@@ -81,6 +81,13 @@ export const SelectionColumn = <T extends Record<string, unknown>>(
         title='' // Removes default title that comes from react-table
         disabled={isDisabled?.(row.original)}
         onClick={(e) => e.stopPropagation()} // Prevents triggering on row click
+        onChange={() => {
+          if (row.subRows.length > 0) {
+            row.toggleRowSelected(!row.isSomeSelected);
+          } else {
+            row.toggleRowSelected();
+          }
```
mayank99, line 89: I'm not sure I understand this logic, and the PR description feels too vague. Can you try to explain it in more detail?

Also:
1. What exactly do you mean by "similar to the one used on the header row"?
2. Can you add a screen recording and the code you used for testing it in vite-playground?
3. Have you verified whether the [behavior difference in expand vs collapse](https://github.com/iTwin/iTwinUI/issues/1972#issuecomment-2034728282) is fixed?
4. Have you investigated whether #928 is related?

testing/e2e/app/routes/Table/spec.ts
```diff
@@ -4,9 +4,10 @@ test.describe('Table sorting', () => {
   test('should work with keyboard', async ({ page }) => {
     await page.goto('/Table');
 
-    const firstColumnCells = page.locator('[role="cell"]:first-child');
+    const firstColumnCells = page.locator('[role="cell"]:nth-child(2)');
```
mayank99, line 7: There shouldn't be any big changes in existing tests. Keep the old component and test unchanged, and use search-params to enable additional features for new tests.

.changeset/happy-carrots-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed issue with `Table` row selection when the row has disabled sub rows.
```
mayank99, line 5: This changeset is insufficient. Fixed what issue?

packages/itwinui-react/src/core/Table/columns/selectionColumn.tsx
Ben-Pusey-Bentley: The logic here is that the row selection should be toggling to false instead of true if some sub rows are currently selected, since there are situations, like when the row has disabled sub rows, where not all of the sub rows can be selected. If you keep trying to select all sub rows when some sub rows can't be selected, then the checkbox can never toggle back to false. On retesting when adding back the vite-playground code, did notice an issue with using only `isSomeSelected` in the logic, since that value is `false` when all sub rows are selected. Reworking the logic to fix this, and will add a test for it.

1.  The checkbox for the selection column's header row has an `onChange` handler that uses similar logic, and never had these issues to begin with as a result.
2. 

https://github.com/iTwin/iTwinUI/assets/141063311/e5788835-417b-42bb-8c44-2e266d769680

Code has also been readded in latest commit for the vite-playground
3. The odd behavior with the expand and collapse was caused by the strange limbo state that the checkbox was being put in, where it was trying to become checked, but no sub rows were updating and all of the sub rows were not selected, so it would temporarily become unchecked. With the selection state fixed, this issue is also fixed.
4. This update to the state also seems to have fixed that issue. Relevant screen capture: 

https://github.com/iTwin/iTwinUI/assets/141063311/2d03338b-210d-4cf1-900a-581d7a4f3f43

testing/e2e/app/routes/Table/spec.ts
Ben-Pusey-Bentley: Okay, will fix.

.changeset/happy-carrots-chew.md
Ben-Pusey-Bentley: Updated changeset in this [commit](https://github.com/iTwin/iTwinUI/pull/2013/commits/997270c33fad447c71ac44fa8ad08053b02b8935).

packages/itwinui-react/src/core/Table/columns/selectionColumn.tsx
mayank99: > The logic here is that the row selection should be toggling to false instead of true if some sub rows are currently selected, since there are situations, like when the row has disabled sub rows, where not all of the sub rows can be selected. 

A condensed version of this explanation would be a useful thing to add as an in-code comment.

> If you keep trying to select all sub rows when some sub rows can't be selected, then the checkbox can never toggle back to false. 

This makes sense. So it should go between checked vs unchecked, and the indeterminate state should be independent of the checked state, correct?

> On retesting when adding back the vite-playground code, did notice an issue with using only `isSomeSelected` in the logic, since that value is `false` when all sub rows are selected. Reworking the logic to fix this, and will add a test for it.

So is this fixed or are you still working on it?

---

> 1. […] The checkbox for the selection column's header row has an `onChange` handler that uses similar logic, and never had these issues to begin with as a result.

When talking about code on github, you should [permalink it](https://github.com/iTwin/iTwinUI/assets/9084735/f5b79f95-52b1-4012-bdc6-53cfd72f8a66) to make it easier for review.

> 2. […] Code has also been readded in latest commit for the vite-playground

The playground code is useful for the deploy preview, but it should also be referenced in the PR description, because the playground will need to be reverted before merge.

> 4. This update to the state also seems to have fixed that issue.

If that issue is also being tackled, then you should [link it to this PR](https://github.com/iTwin/iTwinUI/assets/9084735/b664bb1c-6d8f-487d-bdb6-dff1c153af93).

However, I noticed some incorrect state when filtering. In the screen recording below, you can see that the checkbox does react to user action, but it goes to fully unchecked, when it should instead stay as indeterminate, because some sub-rows are still selected (even if they are not currently visible due to filtering).

https://github.com/iTwin/iTwinUI/assets/9084735/459445b6-b0c8-4e1f-954d-0d3136caf51d

This might be a difficult one to truly fix though. When some sub-rows are disabled and other sub-rows are filtered out, then we have no way to visually distinguish between these two states; they would both be indeterminate (unless I'm missing something).

Ben-Pusey-Bentley: > A condensed version of this explanation would be a useful thing to add as an in-code comment.

Will add.
> So it should go between checked vs unchecked, and the indeterminate state should be independent of the checked state, correct?

Yes, indeterminate is an independent state from checked and unchecked.
> So is this fixed or are you still working on it?

This is fixed, just wanted to add an explanation for why the check changed to include `row.isSelected`.
> The playground code is useful for the deploy preview, but it should also be referenced in the PR description, because the playground will need to be reverted before merge.

Added comment to PR description about removing vite-playground code.
> If that issue is also being tackled, then you should [link it to this PR](https://github.com/iTwin/iTwinUI/assets/9084735/b664bb1c-6d8f-487d-bdb6-dff1c153af93).

Added link to issue to this PR.

> However, I noticed some incorrect state when filtering.

I'll look into fixing this, since I think that maybe the `initalSubRows` value on row might give us a way to check if filtered sub rows are selected.

Ben-Pusey-Bentley: Fixed the filtered sub row deselection issue and added an e2e test for it with this [commit](https://github.com/iTwin/iTwinUI/pull/2013/commits/f7aead31c47928b1ae10be776b6f02c6a837f3b1).

mayank99: I'm still seeing some strange behavior / #928 is unfixed.

https://github.com/iTwin/iTwinUI/assets/9084735/f7235e02-4200-4d49-bf28-1e53f6addfbc

Ben-Pusey-Bentley: That seems to be the inverse of the original problem, where the parent row was trying to select sub rows that couldn't be selected. Now, it is trying to deselect rows that cannot be deselected. I think the best solution here is to have the selection logic ignore disabled and filtered out rows. I'll try to implement that and see how that works.

Ben-Pusey-Bentley: Updated selection logic to account for this in this [commit](https://github.com/iTwin/iTwinUI/pull/2013/commits/bf35a7379df047dfdf2b8448e07517314739d3f7). Also updated e2e tests and added a comment to the handler explaining the logic.

mayank99: The behavior looks good to me now. You should add a screen-recording (or two) in your PR description, along with a [permalink](https://github.com/iTwin/iTwinUI/blob/ed0b545cef47db320f621f43efbe29a24078e409/playgrounds/vite/src/App.tsx) to the playground code (after which you can revert the playground).

As for the code, I would suggest writing it like this for more clarity.
```jsx
onChange={() => {
  if (row.subRows.length > 0) {
    // We want to ignore any sub-rows that are not available (i.e. disabled or filtered out).
    // If all available sub-rows are selected, then we deselect them all, otherwise we select them all.
    const nextState = !row.subRows.every(
      (subRow) => subRow.isSelected || isDisabled?.(subRow.original),
    );
    row.toggleRowSelected(nextState);
  } else {
    row.toggleRowSelected();
  }
}}
```

Lastly, the tests could be made clearer too. It is easier to understand these tests in terms of nested indexes ("Row 1", "Row 1.2", "Row 1.2.1", etc), rather than vague names (`subRowCheckbox`/`secondRowCheckbox`) and vague selectors (`checkboxes.nth(2)`).

Ben-Pusey-Bentley: Alright, will do.

Ben-Pusey-Bentley: Code updates pushed in the latest batch of commits. Also updated the PR description with the screen recordings and the permalink to the playground code.

.changeset/happy-carrots-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fixed issue with `Table` row selection not correctly deselecting all sub rows when the row has disabled sub rows.
```
mayank99, line 5:
```suggestion
Fixed issue with `Table` row selection not correctly deselecting all sub rows when the row has disabled sub rows or when some rows are filtered out.
```

mayank99 (approved):

PR #2019, opened 2024-04-29
Add Astro Icon Component

AdamMeza-Bentley:
## Changes

This PR creates an Astro `Icon_` component for our CSS-workshop. It implements `Icon_` wherever it is applicable. 

This PR doesn't address icons which dont feature the `iui-svg-icon` class. After some iteration I found there are enough use cases for each that `svg` elements without that class should be handled by a separate `svg_` component. 

## Testing

All existing tests are passing 

## Docs

N/A


apps/css-workshop/src/components/Icon.astro
```diff
@@ -0,0 +1,43 @@
...
+const iconProps = {
+  ['class:list']: [
+    {
+      ['iui-svg-icon']: !buttonIcon,
+      ['iui-button-icon']: buttonIcon,
+    },
+    className,
+  ],
+  ['data-iui-icon-size']: size,
+  ['aria-hidden']: 'true',
```
Ben-Pusey-Bentley, line 27: Any wrapping component should provide `aria-hidden` to the icon as well, like #2012.

AdamMeza-Bentley: fixed this in my last push

apps/css-workshop/src/components/Icon.astro
```diff
@@ -0,0 +1,43 @@
...
+  buttonIcon = false,
+  wrapperProps,
+  ...rest
+} = Astro.props;
+
+const iconProps = {
+  ['class:list']: [
+    {
+      ['iui-svg-icon']: !buttonIcon,
+      ['iui-button-icon']: buttonIcon,
```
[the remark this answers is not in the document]
AdamMeza-Bentley: fixed this in the last push after merging in main

apps/css-workshop/src/components/Icon.astro
Ben-Pusey-Bentley: You shouldn't include this `aria-hidden` as a constant value in the `iconProps`, since only `basic` icons that aren't being placed in components like `IconButton` that wrap with `aria-hidden` will need it.

apps/css-workshop/src/components/Icon.astro
```diff
@@ -0,0 +1,46 @@
...
+    },
+    className,
+  ],
+  ['data-iui-icon-size']: size,
+  ['aria-hidden']: 'true',
+  ...rest,
+};
+
+const basic = !Astro.slots.has('default');
+const Element = `svg-${svg}`;
```
mayank99, line 31: Instead of the `basic` prop, I think it would be better to create two different components: `Icon` and `Svg`, because they do different things.
- `Icon`: Only apply `iui-svg-icon` class and `data-iui-icon-size` attributes.
- `Svg`: Only load the correct svg from `@itwin/itwinui-icons-elements`.

apps/css-workshop/src/components/Icon.astro
```diff
@@ -0,0 +1,46 @@
...
+---
+
+{basic && <Element {...iconProps} />}
+{
+  !basic && (
+    <div
+      class:list={['iui-text-spacing', wrapperProps?.class]}
```
mayank99, line 38: Why do you have this random `iui-text-spacing` class here?

apps/css-workshop/src/components/Icon.astro
mayank99: Actually I think I might be misunderstanding the `basic` prop. Why exactly is it necessary?

apps/css-workshop/src/pages/icon.astro
```diff
@@ -53,74 +51,40 @@ import Layout from './_layout.astro';
...
-      class='iui-svg-icon'
-      data-iui-icon-size='m'
-      data-iui-padded='true'
-      aria-hidden='true'></svg-imodel-hollow>
-    <svg-imodel-hollow
-      class='iui-svg-icon'
-      data-iui-icon-size='l'
-      data-iui-padded='true'
-      aria-hidden='true'></svg-imodel-hollow>
+    <Icon_ svg='imodel-hollow' size='s' data-iui-padded='true' />
```
mayank99, line 54: The `data-iui-padded` attribute should be handled within `Icon` using a boolean `padded` prop

apps/css-workshop/src/pages/icon.astro
```diff
@@ -53,74 +51,40 @@ import Layout from './_layout.astro';
...
+    <Icon_ svg='imodel-hollow' size='m' data-iui-padded='true' />
+    <Icon_ svg='imodel-hollow' size='l' data-iui-padded='true' />
   </section>
 
   <h2>Icons with status</h2>
   <section id='demo-status'>
-    <svg-info-circular class='iui-svg-icon' data-iui-icon-color='informational' aria-hidden='true'
-    ></svg-info-circular>
+    <Icon_ svg='info-circular' data-iui-icon-color='informational' />
```
mayank99, line 61: The `data-iui-icon-color` attribute should be handled within `Icon` using a `fill` prop (similar to react)

apps/css-workshop/src/components/Icon.astro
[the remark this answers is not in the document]
mayank99: You can also remove the `buttonIcon` prop and undo the changes in `Tag.astro`

apps/css-workshop/src/components/Icon.astro
AdamMeza-Bentley: It's to determine whether or not an Icon receives a wrapper. I made this choice when working on the examples in `icon.astro`, specifically the 'Auto scale with test' icons. This is also why I'm using `iui-text-spacing` as it was included in the original HTML as the aim was to match the HTML as it existed before. would it make more sense to just include the wrapping explicitly and not use the `basic` variable?

apps/css-workshop/src/pages/icon.astro
AdamMeza-Bentley: made that switch in last push

apps/css-workshop/src/components/Icon.astro
mayank99: The text and the wrapping div are not supposed to be part of the `Icon` component. It's just there for demo purposes, so you should remove it from `Icon.astro` and only keep it in `icon.astro`.

FlyersPh9: Removed.

apps/css-workshop/src/components/Icon.astro
FlyersPh9: @Ben-Pusey-Bentley, can you confirm this is fixed and as expected?  `basic` is no longer a variable, https://github.com/iTwin/iTwinUI/pull/2019#discussion_r1600177777.

Ben-Pusey-Bentley: It looks like the issue should be resolved, since the `aria-hidden` constant value has been removed. I haven't been able to test, though, because I haven't been able to get the branch to build locally. The CI checks also do not seem to be running correctly, so I can't currently check through deploy preview. Maybe solving the merge conflicts will help?

apps/css-workshop/src/components/Icon.astro
```diff
@@ -0,0 +1,22 @@
...
+type Props = {
+  svg: string;
+  size?: 's' | 'm' | 'l' | 'auto';
+  padded?: boolean;
+  fill?: string;
+} & astroHTML.JSX.HTMLAttributes;
+
+const { svg, size = 'm', class: className, padded, fill = 'default', ...rest } = Astro.props;
+
+const iconProps = {
```
mayank99, line 11: This intermediate variable is not necessary. The props can be used directly in the markup.

```jsx
<Element
  class:list={['iui-svg-icon', className]}
  …
```

mayank99 (commented):
Can you resolve the merge conflicts?

apps/css-workshop/src/components/Icon.astro
FlyersPh9: Nice, much cleaner, removed.

apps/css-workshop/src/components/Icon.astro
FlyersPh9: Merge conflicts resolved.

apps/css-workshop/src/components/Icon.astro
mayank99: You've only moved one (1) prop in total. The `…` in my suggestion is meant to imply that you can remove the whole `iconProps` object 😅

FlyersPh9: Ah, I misunderstood.  All `iconProps` have been inlined now.

mayank99 (approved):

apps/css-workshop/src/components/Icon.astro
Ben-Pusey-Bentley: Able to test now. Looks good!

Ben-Pusey-Bentley (approved):

Issue #1895, opened 2024-03-04
`Tooltip`s block click on `MenuItem`


HaveSpacesuit:
### Describe the bug (current behavior)

Wrapping a `Tooltip` around a `MenuItem` inside a `DropdownMenu` blocks the click on the item.
```tsx
<DropdownMenu
  menuItems={(close) => [
    <Tooltip content="content">
      <MenuItem onClick={onClickHandler}>Item title</MenuItem> // will not click
    </Tooltip>
  // ...
```

Nesting the `Tooltip` inside the `MenuItem` prevents the tooltip from appearing.

```tsx
<DropdownMenu
  menuItems={(close) => [
    <MenuItem onClick={onClickHandler}>
      <Tooltip content="content"> // will not appear
          Item title
    </Tooltip>
  // ...
```

Inserting a `<span>` between the `Tooltip` and `MenuItem` allows both the tooltip to appear and the menu item to handle its click, but it is not easy to find this solution.

```tsx
<Tooltip content="content">
  <span>
    <MenuItem onClick={onClickHandler}>Item title</MenuItem>
  </span>
</Tooltip>
```

### Expected Behavior

Wrapping a `Tooltip` around a `MenuItem` should show a tooltip on hover and not block the menu item click.

### Link to minimal repro

https://stackblitz.com/edit/github-lx2rzk

### Steps To Reproduce

Wrap a tooltip around a menu item in a dropdown menu.

### Anything else?

@itwin/itwinui-react v3.5.0

mayank99:
Thanks for reporting this.

Of all the cases you mentioned, these two are totally invalid and should not be used:

```jsx
<Tooltip>Text</Tooltip> // ❌
```

```jsx
<Tooltip>
  <span>…</span> // ❌
</Tooltip>
```

(See [docs](https://itwinui.bentley.com/docs/tooltip) for supported child elements)

The main issue about `MenuItem` is still valid and should be fixed:

```jsx
<Tooltip>
  <MenuItem>…</MenuItem>
</Tooltip>
```

However, for better UX, consider using a [sublabel](https://itwinui.bentley.com/docs/dropdownmenu#sublabel) instead of tooltip.

HaveSpacesuit:
That's a great point about the sublabel, thanks! I'll see if it makes sense in our context.

mayank99:
@HaveSpacesuit I spent some more time looking into it, and realized that you just need to pass a `value` to `MenuItem` make `onClick` work.

[...]

PR #2021, opened 2024-04-29
Private `Menu` component code optimization

rohan-kadkol:
## Changes

WIP

TODO:
- [ ] Undo all changes, make a commit, redo all the changes, then make another commit. This is because I temporarily committed with `--no-verify`.

### DIFF
--- a/packages/itwinui-react/src/core/Menu/Menu.tsx
+++ b/packages/itwinui-react/src/core/Menu/Menu.tsx
@@ -9,9 +9,14 @@ import {
   getFocusableElements,
   Box,
   mergeEventHandlers,
+  Portal,
 } from '../../utils/index.js';
-import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
-import { MenuItemContext } from './MenuItem.js';
+import type {
+  PolymorphicForwardRefComponent,
+  PortalProps,
+} from '../../utils/index.js';
+// import { MenuItemContext } from './MenuItem.js';
+import { usePopover } from '../Popover/Popover.js';
 
 type MenuProps = {
   /**
@@ -26,19 +31,20 @@ type MenuProps = {
    * @default true
    */
   setFocus?: boolean;
+  menuProps: ReturnType<typeof getMenuProps>;
 };
 
 /**
  * Basic menu component. Can be used for select or dropdown components.
  */
-export const Menu = React.forwardRef((props, ref) => {
-  const { setFocus = true, className, ...rest } = props;
+const MenuComponent = React.forwardRef((props, ref) => {
+  const { setFocus = true, className, menuProps, ...rest } = props;
 
   const [focusedIndex, setFocusedIndex] = React.useState<number | null>();
   const menuRef = React.useRef<HTMLElement>(null);
-  const refs = useMergedRefs(menuRef, ref);
+  const refs = useMergedRefs(menuRef, ref, menuProps.popover.refs.setFloating);
 
-  const menuItemContext = React.useContext(MenuItemContext);
+  // const menuItemContext = React.useContext(MenuItemContext);
 
   const getFocusableNodes = React.useCallback(() => {
     const focusableItems = getFocusableElements(menuRef.current);
@@ -50,13 +56,13 @@ export const Menu = React.forwardRef((props, ref) => {
 
   React.useEffect(() => {
     const focusableNodes = getFocusableNodes();
-    if (
-      menuItemContext != null &&
-      menuItemContext.focusableNodes.current !== focusableNodes
-    ) {
-      menuItemContext.focusableNodes.current = focusableNodes;
+
+    console.log('focusableNodes', focusableNodes);
+
+    if (menuProps.focusableNodes.current !== focusableNodes) {
+      menuProps.focusableNodes.current = focusableNodes;
     }
-  }, [getFocusableNodes, menuItemContext]);
+  }, [menuProps.focusableNodes.current, getFocusableNodes]);
 
   React.useEffect(() => {
     const items = getFocusableNodes();
@@ -74,42 +80,95 @@ export const Menu = React.forwardRef((props, ref) => {
   }, [setFocus, focusedIndex, getFocusableNodes]);
 
   const onKeyDown = (event: React.KeyboardEvent<HTMLElement>) => {
-    if (event.altKey) {
-      return;
-    }
+    console.log('onKeyDown', event);
 
-    const items = getFocusableNodes();
-    if (!items?.length) {
-      return;
-    }
+    // if (event.altKey) {
+    //   return;
+    // }
 
-    const currentIndex = focusedIndex ?? 0;
-    switch (event.key) {
-      case 'ArrowDown': {
-        setFocusedIndex(Math.min(currentIndex + 1, items.length - 1));
-        event.preventDefault();
-        event.stopPropagation();
-        break;
-      }
-      case 'ArrowUp': {
-        setFocusedIndex(Math.max(currentIndex - 1, 0));
-        event.preventDefault();
-        event.stopPropagation();
-        break;
-      }
-      default:
-        break;
-    }
+    // const items = getFocusableNodes();
+    // if (!items?.length) {
+    //   return;
+    // }
+
+    // const currentIndex = focusedIndex ?? 0;
+    // switch (event.key) {
+    //   case 'ArrowDown': {
+    //     setFocusedIndex(Math.min(currentIndex + 1, items.length - 1));
+    //     event.preventDefault();
+    //     event.stopPropagation();
+    //     break;
+    //   }
+    //   case 'ArrowUp': {
+    //     setFocusedIndex(Math.max(currentIndex - 1, 0));
+    //     event.preventDefault();
+    //     event.stopPropagation();
+    //     break;
+    //   }
+    //   default:
+    //     break;
+    // }
   };
 
   return (
-    <Box
-      as='div'
-      className={cx('iui-menu', className)}
-      role='menu'
-      ref={refs}
-      {...rest}
-      onKeyDown={mergeEventHandlers(props.onKeyDown, onKeyDown)}
-    />
+    menuProps.popover.open && (
+      <Portal portal={menuProps.portal}>
+        <Box
+          as='div'
+          className={cx('iui-menu', className)}
+          // role='menu'
+          ref={refs}
+          {...menuProps.popover.getFloatingProps({
+            role: 'menu',
+            ...rest,
+            onKeyDown: mergeEventHandlers(props.onKeyDown, onKeyDown),
+          })}
+          // {...rest}
+          // onKeyDown={mergeEventHandlers(props.onKeyDown, onKeyDown)}
+          // onKeyDown={mergeEventHandlers(onKeyDown)}
+        />
+      </Portal>
+    )
   );
 }) as PolymorphicForwardRefComponent<'div', MenuProps>;
+
+// ----------------------------------------------------------------------------
+
+const getMenuProps = ({
+  popoverProps,
+  ...rest
+}: {
+  popoverProps: Parameters<typeof usePopover>[0];
+  portal: PortalProps['portal'];
+}) => {
+  const [currentFocusedNodeIndex, setCurrentFocusedNodeIndex] = React.useState<
+    number | null
+  >(null);
+  const focusableNodes = React.useRef<Array<HTMLElement | null>>([]);
+
+  return {
+    popover: usePopover({
+      interactions: {
+        listNavigation: {
+          activeIndex: currentFocusedNodeIndex,
+          onNavigate: setCurrentFocusedNodeIndex,
+          listRef: focusableNodes,
+          focusItemOnOpen: true,
+          ...popoverProps.interactions?.listNavigation,
+        },
+        ...popoverProps.interactions,
+      },
+      ...popoverProps,
+    }),
+    ...rest,
+    currentFocusedNodeIndex,
+    setCurrentFocusedNodeIndex,
+    focusableNodes,
+  };
+};
+
+// ----------------------------------------------------------------------------
+
+export const Menu = Object.assign(MenuComponent, {
+  getMenuProps,
+});
--- a/packages/itwinui-react/src/core/Buttons/SplitButton.tsx
+++ b/packages/itwinui-react/src/core/Buttons/SplitButton.tsx
@@ -9,7 +9,7 @@ import type { ButtonProps } from './Button.js';
 import { IconButton } from './IconButton.js';
 import {
   Box,
-  Portal,
+  // Portal,
   SvgCaretDownSmall,
   SvgCaretUpSmall,
   useId,
@@ -21,7 +21,7 @@ import type {
 } from '../../utils/index.js';
 import type { Placement } from '@floating-ui/react';
 import { Menu } from '../Menu/Menu.js';
-import { usePopover } from '../Popover/Popover.js';
+// import { usePopover } from '../Popover/Popover.js';
 
 export type SplitButtonProps = ButtonProps & {
   /**
@@ -96,19 +96,39 @@ export const SplitButton = React.forwardRef((props, forwardedRef) => {
     return menuItems;
   }, [menuItems, close]);
 
-  const popover = usePopover({
-    visible,
-    onVisibleChange: (open) => (open ? setVisible(true) : close()),
-    placement: menuPlacement,
-    matchWidth: true,
+  // const [activeIndex, setActiveIndex] = React.useState<number | null>(null);
+  // const listItemsRef = React.useRef<Array<HTMLElement | null>>([]);
+
+  const menuProps = Menu.getMenuProps({
+    portal,
+    popoverProps: {
+      visible,
+      onVisibleChange: (open) => (open ? setVisible(true) : close()),
+      placement: menuPlacement,
+      matchWidth: true,
+    },
   });
+
+  // const popover = usePopover({
+  //   visible,
+  //   onVisibleChange: (open) => (open ? setVisible(true) : close()),
+  //   placement: menuPlacement,
+  //   matchWidth: true,
+  //   interactions: {
+  //     listNavigation: {
+  //       activeIndex,
+  //       onNavigate: setActiveIndex,
+  //       listRef: listItemsRef,
+  //     },
+  //   },
+  // });
 
   const labelId = useId();
 
   return (
     <Box
       {...wrapperProps}
-      ref={popover.refs.setPositionReference}
+      ref={menuProps.popover.refs.setPositionReference}
       className={cx(
         'iui-button-split',
         {
@@ -133,28 +153,29 @@ export const SplitButton = React.forwardRef((props, forwardedRef) => {
         size={size}
         disabled={props.disabled}
         aria-labelledby={props.labelProps?.id || labelId}
-        aria-expanded={popover.open}
-        ref={popover.refs.setReference}
-        {...popover.getReferenceProps(menuButtonProps)}
+        aria-expanded={menuProps.popover.open}
+        ref={menuProps.popover.refs.setReference}
+        {...menuProps.popover.getReferenceProps(menuButtonProps)}
       >
         {visible ? <SvgCaretUpSmall /> : <SvgCaretDownSmall />}
       </IconButton>
-      {popover.open && (
-        <Portal portal={portal}>
-          <Menu
-            {...popover.getFloatingProps({
-              onKeyDown: ({ key }) => {
-                if (key === 'Tab') {
-                  close();
-                }
-              },
-            })}
-            ref={popover.refs.setFloating}
-          >
-            {menuContent}
-          </Menu>
-        </Portal>
-      )}
+      {/* {popover.open && ( */}
+      {/* <Portal portal={portal}> */}
+      <Menu
+        menuProps={menuProps}
+        // {...popover.getFloatingProps({
+        //   onKeyDown: ({ key }) => {
+        //     if (key === 'Tab') {
+        //       close();
+        //     }
+        //   },
+        // })}
+        // ref={popover.refs.setFloating}
+      >
+        {menuContent}
+      </Menu>
+      {/* </Portal> */}
+      {/* )} */}
     </Box>
   );
 }) as PolymorphicForwardRefComponent<'button', SplitButtonProps>;


apps/react-workshop/cypress-visual-screenshots/baseline/Select.test.ts-With Selected Value (Open).png
Ben-Pusey-Bentley, line 1: Is there a reason that the styling on the selection state for this `Select` component is changing?

testing/e2e/app/routes/Menu/spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  test('should render menu with custom role', async ({ page }) => {
+    await page.goto('/Menu?role=listbox');
+
+    expect(page.locator('.Menu[role="listbox"]')).toBeVisible();
+
+    await page.waitForTimeout(50);
+  });
+
+  test('should focus selected item', async ({ page }) => {
+    // TODO: Do we need to focus the selected item?
```
Ben-Pusey-Bentley, line 47: Wanted to highlight this TODO, since I don't see it mentioned in the PR description.

apps/react-workshop/cypress-visual-screenshots/baseline/Select.test.ts-With Selected Value (Open).png
rohan-kadkol: Oops, I had a comment for that from last week but looks like I missed publishing the review (https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1589860524).

Not sure why that change is happening in the test image because the dev server version continues to look the same. We could continue the discussion on that thread since I've included some information there.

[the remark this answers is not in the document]
Ben-Pusey-Bentley: I think this is related to the failing focus selected item test. The thicker border may have been caused by the item being focused, as well as selected.

packages/itwinui-react/src/core/Menu/Menu.tsx
```diff
@@ -49,67 +59,45 @@ export const Menu = React.forwardRef((props, ref) => {
   }, []);
 
   React.useEffect(() => {
```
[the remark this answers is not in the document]
mayank99: You could probably use a callback `ref` instead of `useEffect`, but you would still need `getFocusableNodes`.

However, the main thing that this code is trying to solve is based on a flawed understanding of how the component is supposed to behave. Ideally, there would only be `MenuItem`s inside the `Menu`. But currently we support `MenuExtraContent` with any arbitrary focusable elements, which is not at all how a menu is supposed to work.

How difficult would it be to make it such that it uses regular <kbd>Tab</kbd> navigation for `MenuExtraContent`, and then only uses arrow-key navigation for the actual menu-items? Would likely need a wrapper around the menu-items (could be automatically added maybe?):

```jsx
<Menu>
  <MenuExtraContent>
    <button>tab to me</button>
  </MenuExtraContent>
  <Divider />
  <MenuList>
    <MenuItem>tab again</MenuItem>
    <MenuItem>now use arrow keys</MenuItem>
    <MenuItem>more arrows 🏹</MenuItem>
  </MenuList>
</Menu>
```

Additionally, we should probably deprecate `MenuExtraContent` in favor of `Popover`.

packages/itwinui-react/src/core/Menu/Menu.tsx
mayank99, line 1: While this is a net improvement over the previous code, this abstraction feels leaky and incomplete.
- The consumers of `Menu` shouldn't be required to use `useListNavigationProps` and `MenuContext`, both of which are implementation details of `Menu`.
- Additionally, this does not simplify `MenuItem` by much. There is a good amount of submenu-related code in there that feels like it should be within `Menu` instead. 🤔

One way to improve this abstraction would be by making more heavy use of `cloneElement`. The `Menu` component can clone the trigger element to add any necessary props to it. In the case of `DropdownMenu`, the trigger comes from the `children`, and in the case of `MenuItem` (submenus), it's the parent item, and in the case of `DropdownButton`/`SplitButton`, we render a `Button` ourselves. These trigger elements could be passed using a new prop to `Menu` (call it `trigger` or something; it's private so doesn't matter too much).

mayank99: It's also worth mentioning that you're currently missing `getItemProps`.

There are three pieces returned by `useFloating`:
- `getReferenceProps`: Goes on the trigger (cloned) element
- `getFloatingProps`: Goes on the menu itself
- `getItemProps`: Goes on the individual menu-items

rohan-kadkol: > While this is a net improvement over the previous code, this abstraction feels leaky and incomplete.

I have been working on improving this further using your suggestions. `Menu` now handles much more logic than it used to ([JSDocs](https://github.com/iTwin/iTwinUI/pull/2021/files#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R55-R83)).

I have some TODOs and some testing that are still left, but it may be good if you can take a look at the general picture to see if it aligns with what you had in mind.

packages/itwinui-react/src/core/Menu/Menu.tsx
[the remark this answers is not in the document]
rohan-kadkol: > How difficult would it be to make it such that it uses regular Tab navigation for `MenuExtraContent`, and then only uses arrow-key navigation for the actual menu-items?

I believe it would take a bit of thinking to get all aspects right and not introduce any regressions. So, I feel a simple solution that keeps the same behavior might be more suitable here, at least for this PR.

I played around with using `useSyncExternalStore` instead of the `useEffect` ([code](https://github.com/iTwin/iTwinUI/pull/2021/files#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R184-R208)). I think it works for most of the cases. Will be testing it more thoroughly.

packages/itwinui-react/src/core/Menu/Menu.tsx
mayank99: The `trigger` prop does help improve things, but the code in `MenuItem` does not look good. Most of that code should instead be in `Menu` (and maybe even in `DropdownMenu`). At best, `MenuItem` should pass some simple configuration props to `Menu`. It should be possible to refactor `Menu` in the future without requiring large changes in `MenuItem`.

It might help if you try to think of `MenuItem` is a simple component that renders something like this:
```jsx
const mainItem = <ListItem>…</ListItem>;

if (submenuItems.length > 0) {
  return <Menu trigger={mainitem}>{submenuItems}</Menu>
}

return mainItem;
```

rohan-kadkol: Sounds good, the last few commits (https://github.com/iTwin/iTwinUI/pull/2021/files/d56435938c3bc574875183c3c12888d037bf9874..af8c78e6bf833018e86cac264ec23671894866fb) try to abstract that more popover/submenu logic from `MenuItem` to `Menu`.

apps/react-workshop/cypress-visual-screenshots/baseline/Select.test.ts-With Selected Value (Open).png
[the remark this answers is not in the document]
rohan-kadkol: > I think this is related to the failing focus selected item test. The thicker border may have been caused by the item being focused, as well as selected.

Thanks, yes, I think that was the reason. After #2036, `Menu` is no longer used in `Select`, and as a result, this image change is no longer there. Thus, closing this comment.

testing/e2e/app/routes/Menu/spec.ts
rohan-kadkol: #2034 and #2036 removed uses of `Menu` from `ComboBox` and `Select`. As a result, focusing the selected item is no longer needed by any component that uses `Menu`.

Thus, I removed the code to focus the selected item and also removed this e2e test case.

Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Menu/Menu.tsx
rohan-kadkol: > It's also worth mentioning that you're currently missing `getItemProps`.

Ohh, right, added `getItemProps` ([code](https://github.com/iTwin/iTwinUI/blob/dae20abccde7720a171fce7e20395133c37c60d3/packages/itwinui-react/src/core/Menu/Menu.tsx#L250)).

rohan-kadkol:
Over the last few commits since May 20 (https://github.com/iTwin/iTwinUI/pull/2021/files/ed8d1fdc8cc4934d7092708787a3c58f859ec400..7986e482c24fa68ad05fd8556aec52fda487c53e), I have been fixing the last remaining bugs/regressions, adding/fixing the tests, and also addressing the remaining GitHub comments. I updated the PR description to reflect the current state of the PR.

There was a small regression that happened a few hours ago after [`dae20ab` (#2021)](https://github.com/iTwin/iTwinUI/pull/2021/commits/dae20abccde7720a171fce7e20395133c37c60d3)/https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1609930747. I am currently trying to fix it. But apart from that, the PR is ready for review.

packages/itwinui-react/src/core/Menu/Menu.tsx
rohan-kadkol: > Ohh, right, added `getItemProps` ([code](https://github.com/iTwin/iTwinUI/blob/dae20abccde7720a171fce7e20395133c37c60d3/packages/itwinui-react/src/core/Menu/Menu.tsx#L250)).

After making that change, I noticed that clicking on a `MenuItem` started giving an error:

<details>
<summary>Error</summary>

image

</details>

This was because the `MenuItem`'s `onClick`'s unique non-standard function signature. i.e. `(value?: unknown) => void` instead of something like `(event?: MouseEvent) => void`. Because of this, adding `getItemProps`'s `onClick` to `MenuItem` through cloning was not working. Thus, I created `MenuContext` to pass it down to `MenuItem` (some of these changes are in https://github.com/iTwin/iTwinUI/pull/2021/files/037431f98742b731fb3afbb1dce5597b2b7afa03..d3a42bf41772eccc83604284250e44d2f47d8816).

I believe this is alright since we mainly allow `MenuItem`s inside `Menu`.

packages/itwinui-react/src/core/Menu/Menu.tsx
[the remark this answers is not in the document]
rohan-kadkol: > I played around with using `useSyncExternalStore` instead of the `useEffect` ([code](https://github.com/iTwin/iTwinUI/pull/2021/files#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R184-R208)). I think it works for most of the cases. Will be testing it more thoroughly.

After properly applying `getItemProps` (https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1612095985), I realized that the extra `useSyncExternalStore` to keep `useListNavigation`'s activeIndex in sync whenever an item was hovered/focused was no longer needed. I, thus, removed it (https://github.com/iTwin/iTwinUI/pull/2021/commits/f75b3df2afa4f35f0a565e9ff2c476259f4a9eca).

However, I still could not remove the `useEffect`. If we have plans to only support `MenuItem`s and probably deprecate `MenuExtraContent`, I guess this `useEffect` is alright since it will be removed in the future.

packages/itwinui-react/src/core/Menu/Menu.tsx
mayank99: [This error](https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1612095985), along with https://github.com/iTwin/iTwinUI/issues/1895#issuecomment-1981579644, is another reason we should deprecate `MenuItem` (in a future PR). I also wonder if it would be safe to call `onClick` with the event object if a `value` prop is not passed.

apps/react-workshop/src/DropdownMenu.test.ts
```diff
@@ -23,10 +23,6 @@ describe('DropdownMenu', () => {
         cy.get('button').first().click();
       });
 
-      if (testName === 'Submenu') {
-        cy.get('[role=menuitem][aria-expanded=false]').trigger('mouseenter');
```
mayank99, line 27: We should fix this test. The original intent was to open the submenu, but for some reason it wasn't working. You can see in the image that the submenu is not even open.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -224,107 +145,70 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
         onClick,
         onKeyDown,
         onMouseEnter,
-        onFocus,
       }
     : {};
 
+  const trigger = (
+    <ListItem
+      as='div'
```
mayank99, line 165: Should we use `button` so that it automatically handles Enter/Spacebar keys?

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -215,19 +221,31 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
     ),
   });
 
+  const interactionsEnabledStates = React.useMemo(
```
mayank99, line 224: nit: this intermediate variable looks unnecessary

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -275,13 +296,70 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
...
+  const [currentFocusedNodeIndex, setCurrentFocusedNodeIndex] = React.useState<
+    number | null
+  >(null);
+  const focusableNodes = React.useRef<Array<HTMLElement | null>>([]);
+
+  const {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    activeIndex: activeIndexProp,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    listRef: listRefProp,
```
mayank99, line 341: I'm not sure I understand this. You're accepting these props only to then ignore them?

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -120,16 +120,19 @@ type PopoverInternalProps = {
    * When additional parameters are _required_ for an interaction/trigger, an object must be passed to enable it.
    * Booleans will not be allowed in this case.
    *
+   * When trying to add `listNavigation`, consider using `useListNavigationProps` to generate the necessary props.
```
mayank99, line 123: Any reason we can't call this hook automatically like the other interactions?

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: I was hoping to make sure that `useListNavigationProps`'s `activeIndex` and `listRef` are not overridden by what is passed in its props. I think this was important when I was experimenting with a `useEffect` inside `useListNavigationProps` that was editing the `listRef` internally. But since I don't have that `useEffect` anymore, I don't think this is too important.

So, upon re-reading that code, I believe this JS safeguard is unnecessary and that the TS types are sufficient to prevent overriding those two props. Thus, removed that JS safeguard (https://github.com/iTwin/iTwinUI/pull/2021/commits/8fd3dabe68212da396b6a71dfb6d890d458872e5).

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: I don't think adding the `useEffect` and `getFocusableNodes` to `usePopover` is acceptable, since it's a hack that needs to be removed in the future (as discussed [previously](https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1591006224)). Keep the hacks confined to `Menu` and/or a separate hook. I was just asking a question here, not making a suggestion.

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: I felt it would be safe to use this variable to clearly specify which interactions are enabled. Then wherever we need to run some logic conditionally on whether the interaction is enabled/disabled, we can use this variable instead of inlining an `if` condition (E.g. [1](https://github.com/iTwin/iTwinUI/pull/2021/files/10c41d4b745f0f636c2db9e5f731754a7559e364..a5d9f8f1eb6afbabc9c45f1a9d846c18bf66c6aa#diff-6dd66c2b86bfd08956a740b1aa4db0d9b0c5c5ae871748791e4ac6eb06ff6af9R247-R250), [2](https://github.com/iTwin/iTwinUI/pull/2021/files/10c41d4b745f0f636c2db9e5f731754a7559e364..a5d9f8f1eb6afbabc9c45f1a9d846c18bf66c6aa#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R233)).

That way, if the logic of when a variable is enabled is changed (e.g. `someInteraction === true` vs. just `someInteractions != null`), we can just change it in one place.

Moreover, since this was related to the private `usePopover` hook, I felt this approach would be alright.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -224,107 +145,70 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
     // Focus the item when hovered.
     if (e.target === e.currentTarget) {
       menuItemRef.current?.focus();
```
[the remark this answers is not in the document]
rohan-kadkol: Yes, I too felt that after spreading `getItemProps` in the missing places, FloatingUI will handle the focus on hover. However, upon testing, the focus was not reliable and kept jumping abruptly in some cases.

https://github.com/iTwin/iTwinUI/assets/45748283/de769ad2-b3ea-460c-927e-eed7af25a7a9

Thus, to be safe, I thought to handle it manually instead. But if this is an issue, we can investigate further why FloatingUI's `focusItemsOnHover` is not working with our implementation.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: So if it's just about using the default interactions, can't you return `mergedInteractions` directly? This intermediate variable does literally nothing extra, as far as I can tell.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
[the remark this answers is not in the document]
mayank99: It might have something to do with the way we set `tabIndex` in `MenuItem`.

Let's investigate this in the near-future (in a separate PR)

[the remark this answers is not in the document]
rohan-kadkol: Sounds good, added an after-PR TODO for this.

packages/itwinui-react/src/core/Menu/MenuItem.tsx
rohan-kadkol: > Should we use `button` so that it automatically handles Enter/Spacebar keys?

I tried using `as="button"`, but the styles broke:

<details>
<summary>Screenshot</summary>

image

</details>

> There's also `ListItem.Action` which renders a button.

I tried wrapping the `children` with a `ListItem.Action` but for some reason, could not get `ListItem.Action`'s `onClick` to trigger when pressing <kbd>Enter</kbd>/<kbd>Space</kbd>.

<details>
<summary>diff</summary>

```diff
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -123,24 +123,6 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
     } satisfies Parameters<typeof Menu>[0]['popoverProps'];
   }, [subMenuItems.length]);
 
-  const onKeyDown = (event: React.KeyboardEvent<HTMLElement>) => {
-    if (event.altKey) {
-      return;
-    }
-
-    switch (event.key) {
-      case 'Enter':
-      case ' ':
-      case 'Spacebar': {
-        onClick();
-        event.preventDefault();
-        break;
-      }
-      default:
-        break;
-    }
-  };
-
   const onMouseEnter = (e: React.MouseEvent<HTMLElement>) => {
     // Focus the item when hovered.
     if (e.target === e.currentTarget) {
@@ -148,14 +130,8 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
     }
   };
 
-  const onClick = () => {
-    onClickProp?.(value);
-  };
-
   const handlers = !disabled
     ? {
-        onClick,
-        onKeyDown,
         onMouseEnter,
       }
     : {};
@@ -185,7 +161,14 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
         </ListItem.Icon>
       )}
       <ListItem.Content>
-        <div>{children}</div>
+        <ListItem.Action
+          onClick={() => {
+            console.log('Testing the onClick event.');
+            onClickProp?.(value);
+          }}
+        >
+          {children}
+        </ListItem.Action>
         {sublabel && <ListItem.Description>{sublabel}</ListItem.Description>}
       </ListItem.Content>
       {!endIcon && subMenuItems.length > 0 && (
```

</details>

mayank99: > I tried using `as="button"`, but the styles broke

Yeah that's expected, because `button` comes with a bunch of default styles. The styles are fixable using something like `class=iui-button-base`.

> I tried wrapping the `children` with a `ListItem.Action` but for some reason, could not get `ListItem.Action`'s `onClick` to trigger when pressing <kbd>Enter</kbd>/<kbd>Space</kbd>.

Yeah I realized `ListItem.Action` might not be the best way to do it. It's designed for simple scenarios where a regular non-interactive list item contains an anchor or button. But in this case we are turning `ListItem` into an interactive `option` which gets focused. (Events only fire on the node that gets focus, which in this case is a parent of `ListItem.Action`)

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: `mergedInteractions` is the `interactions` from the user merged with our default interactions. However, to know if an interaction is really enabled, we have to do something like `mergedInteractions.click === true || (typeof mergedInteractions.click === 'object' && mergedInteractions.click.enabled !== false)`.

We can write this condition each time we want to check if an interaction is enabled. However, to make it easy, `interactionsEnabledStates` runs these conditions and exposes an object where the key is the interaction, and the value is a boolean that represents whether the interaction is enabled or not.

This is why I think `interactionsEnabledStates` is helpful.

---

I realized the previous if-conditions were insufficient. I thus updated those if-conditions to also check for the `.enabled` (https://github.com/iTwin/iTwinUI/pull/2021/commits/3c4289471f31e2f67b8f2a9b6a2d9e1ea3823001).

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: Ohh, I see. Reverted those changes to keep that `useEffect`/`getFocusableNodes` in `Menu` (https://github.com/iTwin/iTwinUI/pull/2021/commits/cffb1d0595c96bda2af2a7f8476e56f6c65c89c2).

> Any reason we can't handle this automatically based on the object passed to `listNavigation`?

In `usePopover`, I tried wrapping the `listNavigationProps` passed by `Menu` with useListNavigationProps()`. However, I could not get the arrow key navigation working correctly when doing this. I guess this has something to do with the stale `listRef` values being used since `listRef` a ref.

Thus, in that same commit, I reverted it such that `useListNavigationProps` needs to be called by `Menu` and its result passed to `usePopover`; and `listRef` needs to be updated by `Menu`.

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: > I don't understand why it needs to be this complicated.

Yes, it's unfortunately complicated, but I tried to keep it as simple as possible. I felt it would be better to keep this slight complexity in a helper object instead of keeping it inline each time we need to check if an interaction is enabled since I felt the latter is error prone.

> Where do you even have a need for such checks??

* This is needed to run some workaround logic in parallel with FloatingUI's `onClick`. This logic only needs to run when the click interaction is enabled ([code](https://github.com/iTwin/iTwinUI/pull/2021/files/10c41d4b745f0f636c2db9e5f731754a7559e364..cffb1d0595c96bda2af2a7f8476e56f6c65c89c2#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R262-R265)).
* Updating `listRef` only when the listNavigation interaction is enabled. This is helpful because useFocusableNodes is, like you said, an expansive operation ([code](https://github.com/iTwin/iTwinUI/pull/2021/files/10c41d4b745f0f636c2db9e5f731754a7559e364..6b0bc399d7e1a07f4fa65229e31543e6e508b88e#diff-eb1d53cfaf5ebd716d4feb107991cc2f5775bf2d4e9cc83b9de60e2206dea595R201-R204)).

mayank99: Your links are broken. Can you permalink to the actual code in your branches instead of the PR diff?

rohan-kadkol: Ohh, I see. Looks like that's because `Menu.tsx` is collapsed by default in the diff view. Updated the links to permalinks.

apps/react-workshop/src/DropdownMenu.test.ts
rohan-kadkol: Fixed (https://github.com/iTwin/iTwinUI/pull/2021/commits/6b77ac2a13b6938d7f61c2c1a7e0992a7a6f4366).

packages/itwinui-react/src/core/Menu/Menu.tsx
rohan-kadkol: > [This `onClick` error](https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1612095985), along with [#1895 (comment)](https://github.com/iTwin/iTwinUI/issues/1895#issuecomment-1981579644), is another reason we should deprecate `MenuItem` (in a future PR).

Sounds good, added an after-PR TODO to consider it.

> I also wonder if it would be safe to call `onClick` with the event object if a `value` prop is not passed.

Ohh, I'm not sure this will be clear to the user if they miss reading our documentation. Also, it might be considered a breaking change.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: > This is needed to run some workaround logic in parallel with FloatingUI's `onClick`. This logic only needs to run when the click interaction is enabled ([code](https://github.com/iTwin/iTwinUI/blob/cffb1d0595c96bda2af2a7f8476e56f6c65c89c2/packages/itwinui-react/src/core/Menu/Menu.tsx#L262-L265)).

If this is a workaround, can we expand the code comment to indicate what exactly it is working around? We should try to remove it in the future.

> Updating `listRef` only when the listNavigation interaction is enabled. This is helpful because useFocusableNodes is, like you said, an expansive operation ([code](https://github.com/iTwin/iTwinUI/blob/6b0bc399d7e1a07f4fa65229e31543e6e508b88e/packages/itwinui-react/src/core/Menu/Menu.tsx#L201-L204)).

Isn't `listNavigation` always going to be enabled in `Menu`? And if it's an expensive operation, the correct solution is to only update the focusable nodes when necessary. Putting it in a `useEffect` with a large dependency array is probably not a good idea.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: > However, I could not get the arrow key navigation working correctly when doing this. I guess this has something to do with the stale `listRef` values being used since `listRef` a ref.

Did you try using state instead of ref? Then it should be up-to-date, as it will re-render when the state changes.

rohan-kadkol: Yes, I tried it for a while initially but went the ref route since FloatingUI was expecting a ref for `listRef` ([docs](https://floating-ui.com/docs/useListNavigation#props)). I tried to make `useListNavigationProps` expose a custom `setListRef` function, but that caused some TS type complexities since `useListNavigationProps` was supposed to return `UseListNavigationProps`.

I will take another look at using states soon.

mayank99: A ref object is nothing special; it might be enough to just do something like this to satisfy the floating-ui API:

```js
useListNavigation(context, {
  listRef: { current: listElements },
});
```

where `listElements` is a prop stored in state.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: > > If this is a workaround, can we expand the code comment to indicate what exactly it is working around? We should try to remove it in the future.
> 
> To confirm, do you mean in addition to this ([code](https://github.com/iTwin/iTwinUI/blob/6b0bc399d7e1a07f4fa65229e31543e6e508b88e/packages/itwinui-react/src/core/Menu/Menu.tsx#L277-L281))?

I didn't realize that comment was applicable for referring to this case, since it comes after the early return. Maybe the two separate comments should be combined.

Regardless, I was looking into this issue and realized you might be able to avoid the workaround using [`ignoreMouse: true`](https://floating-ui.com/docs/useclick#ignoremouse).

---

> > Isn't `listNavigation` always going to be enabled in `Menu`? And if it's an expensive operation, the correct solution is to only update the focusable nodes when necessary. Putting it in a `useEffect` with a large dependency array is probably not a good idea.
> 
> I'm not quite sure what is the best time/condition to update `listNavigationProps.listRef`. Is there a good way to know when the focusable nodes have changed?

There is no nice way but you can use a [mutation observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver) to know when the child elements change.

In fact this could be abstracted away into a hook that always returns the latest focusable nodes, which you can then filter according to your needs.

```js
const allFocusableNodes = useFocusableNodes(rootNode);
const focusableNodes = React.useMemo(() => {
  return allFocusableNodes.filter(
    (i) => !focusableItems.some((p) => p.contains(i.parentElement)),
  );
}, [allFocusableNodes]);
```

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: > Would this change in types be allowed in a `minor` release? E.g. the event's type in the `onClick` might cause some errors if the users' code

We can leave the types unchanged (as `'div'` rather than `'button'`), it's not super important. Think of `button` as an implementation detail that users should not have to think about.

Since the plan is also to deprecate `MenuItem` in the future, we can have correct types for the replacement components.

packages/itwinui-react/src/core/Popover/Popover.tsx
mayank99: Two small suggestions:
1. Inside the hook, we should probably use `useLatestRef` to grab the latest instance of the filter function. Otherwise a new a mutation observer will be created on every single render.
2. (optional) The filter function can be made more intuitive to use. Try to avoid positional arguments to make the code more readable. The code below shows an object as the second argument of the hook and it lifts the `.filter` call to avoid the need for two arguments in the filter function.
   ```tsx
   const focusableElementsRef = useFocusableElementsRef(
     menuElement,
     {
       // Filter out focusable elements that are inside each menu item, e.g. checkbox, anchor
       filter: (allElements) => allElements.filter(
         (i) => !allFocusableElements?.some((p) => p.contains(i.parentElement)),
       ),
     },
   );
   ```

Also, should we move `useFocusableElementsRef` into `utils` into `focusable.ts` since we already have a file for it?

mayank99: the import in `focusable.ts` needs to be fixed to use namespace import for react (`import * as React`) and to import `useSyncExternalStore` from `utils` instead of react

packages/itwinui-react/src/core/Menu/MenuItem.tsx
mayank99: > In [`b8627fd`](https://github.com/iTwin/iTwinUI/pull/2021/commits/b8627fd8fe29045b7b55e02ce4653ddafeced97c), I also had to make some changes in the unit tests to have the keyboard event trigger the `onClick`. ([Relevant issue](https://redirect.github.com/testing-library/user-event/issues/179))

I'm not sure that test is even necessary anymore, it's just testing the default behavior of buttons. We can remove it, as long as we still have larger tests where we test arrow keys for cycling focus and <kbd>Enter</kbd> key for selection.

packages/itwinui-react/src/utils/functions/polymorphic.tsx
```diff
@@ -26,7 +26,13 @@ const _base = <As extends keyof JSX.IntrinsicElements = 'div'>(
...
       // Workaround for Safari refusing to focus links/buttons/non-text inputs.
       if (
-        Element === 'button' ||
+        (Element === 'button' &&
+          // Button should not be disabled
+          !(
+            String(props['aria-disabled']) === 'true' ||
+            String((props as any)['data-iui-disabled']) === 'true' ||
+            (props as React.ComponentPropsWithRef<'button'>).disabled
+          )) ||
```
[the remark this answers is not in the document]
mayank99: The first diff is generally ok, as long as the `tabIndex` can be overridden from outside. Think of it as a default/fallback, rather than something that works everywhere out-of-the-box.

The edited diff is not ok; specifically, there shouldn't be any internal focus state or even a focused prop. The `ButtonBase` part might be ok, though I'm worried that users could override the `as` prop too easily (maybe that's fine? it's their own fault).

The correct way to handle `tabIndex` would be to do it in the place where `MenuItem` is consumed. Specifically, the place where keyboard navigation is wired up, since that's where the tabIndex really comes into play.

Also, one thing to keep in mind is that when using multi-select, only the first selected option should have `tabIndex={0}`; everything else should use `-1`. The intent of using `0` is to indicate which item gets focused initially when focus moves into the list.

Ben-Pusey-Bentley:
> Optional: @Ben-Pusey-Bentley Since some parts of this PR have changed since your review, you could optionally take a quick look at the PR in case you have any comments.

Looks good to me. Tested the new behaviors with the `DropdownMenu`, `SplitButton`, and `Table` column manager, and everything is worked as expected based on the PR description.

packages/itwinui-react/src/utils/functions/polymorphic.tsx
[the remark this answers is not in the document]
mayank99: Did you try investigating why the arrow key was not working? [`a630f13`](https://github.com/iTwin/iTwinUI/pull/2021/commits/a630f13c2159d05901c8d39611b59b714ac4f164) does not look good. Specifically, you are reading a ref during render (to compute `focusableItemIndex`) and you have `onMouseEnter` spread across two different places.

For accessing the current index you might be better off using [`useListItem`](https://floating-ui.com/docs/floatinglist#uselistitem) (which requires `FloatingList`)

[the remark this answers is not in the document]
mayank99: > Update: After seeing the [failing unit test](https://github.com/iTwin/iTwinUI/actions/runs/9550447957/job/26322466413?pr=2021), I realized that `MenuItem` needs to be focused on hover even if it is not in a `Menu`. Thus, I brought back the `onMouseEnter` and the `.focus()` call in `MenuItem` ([7a2cad2](https://github.com/iTwin/iTwinUI/commit/7a2cad2881500e23bbad5a8ab3d39bf277ca63cb)).

This undoes the "fix" from the previous comment, leaving the issue unresolved.

You should be able to remove the `onMouseEnter` from `MenuItem` and consolidate it in `popoverGetItemProps.onMouseEnter` (using `event.currentTarget.focus()`)

testing/e2e/app/routes/Menu/route.tsx
```diff
@@ -0,0 +1,83 @@
+import { useSearchParams } from '@remix-run/react';
+import { Menu } from 'node_modules/@itwin/itwinui-react/esm/core/Menu/Menu.js';
```
mayank99, line 2: e2e tests should only test the public API surface. The `Menu` component is a private implementation detail of `DropdownMenu` and other components, so prefer testing those instead.

packages/itwinui-react/src/utils/functions/polymorphic.tsx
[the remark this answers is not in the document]
mayank99: > Update: After seeing the [failing unit test]

When looking at these test logs, I noticed a [warning](https://github.com/iTwin/iTwinUI/actions/runs/9550447957/job/26322466413?pr=2021#step:7:28) about nested `<button>`s that might be worth investigating. It's not related to this thread but still likely caused by this PR.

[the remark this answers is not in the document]
mayank99: > To confirm, are you saying the `popoverGetItemProps` from `Menu` should call `event.currentTarget.focus()` and since `MenuItem` inside a `Menu` will spread `popoverGetItemProps`, `.focus()` will be called in `MenuItem` too?

Yes that's correct. The latest changes look good!

> Just wondering, would this be considered a breaking change? (Will add a changeset depending on what we discuss)

No, it's not a problem at all. The "focus on hover" behavior only makes sense for `Menu` anyway, so this is actually more correct. Further, `MenuItem` should never be used independently of `Menu` (I believe this is something we have discussed before in #1942).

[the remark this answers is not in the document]
mayank99: Revisiting an earlier part of this thread:

> Regarding the `as="ButtonBase"`, I kept it as `as="button"` and with adding the `iui-button-base` class manually so that the user's `as` prop will be respected and the styles will not break. i.e. even if the user provides an `as` prop, the `iui-button-base` class will always exist.
> 
> Since I'm not using `as={ButtonBase}`, I confirmed that `MenuItem` prevents clicks when disabled. ([code](https://github.com/iTwin/iTwinUI/blob/7b2485e19692220673d771531463f6995c4c7102/packages/itwinui-react/src/core/Menu/MenuItem.tsx#L139-L144)).

I noticed two things in the current implementation:
1. We have an `onClick` handler that just guards against the `disabled` prop.
2. We are missing `type="button"` on the `<button>` element (defaults to `type="submit"`). 

For now, we should manually add `type='button'` ourselves.

In the future, we should investigate a way to use `ListItem` and `ButtonBase` together to reduce repetition.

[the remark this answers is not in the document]
mayank99: Let's think about `ButtonBase` later. The current code looks fine 👍

testing/e2e/app/routes/Menu/route.tsx
mayank99: > I thought about moving these `Menu` tests to `DropdownMenu` e2e tests but felt that they were not the right place for them since some tests are `Menu` specific and `Menu` is used in other components too apart from just the `DropdownMenu`.

I don't agree with this logic. `DropdownMenu` is a very thin wrapper over `Menu` so it's the best place to test it. If it helps, you can think of it as adding new tests to verify that `DropdownMenu` is behaving correctly. `Menu` is an implementation that literally does not matter.

> The only `Menu` e2e test now is one that checks that the focus moves correctly when opening/closing the `Menu` with a keyboard and mouse.

This needs to be removed. We should not have *any* e2e tests that directly use `Menu`.

mayank99: I'm ok with just testing `DropdownMenu`, but you can optionally also test the other components if you feel it's useful.

mayank99: The e2e tests can be improved and cleaned up a bit.
- Instead of rendering multiple menus, you can create local components (`<DropdownMenuWithSubmenus>`, `<DropdownMenuWithExtraContent>`) that can be conditionally rendered based on search params (see other e2e tests).
- Some of the test `expect`s are missing `await`s.
  - This probably causes flakiness which need to be resolved using hacks like the `keyboardPressOptions` object.
- Some of the test names don't match what is being tested. e.g. the "should respect click and keyboard enter/space triggers" doesn't test for clicks anymore.
  - Further, manually calling `.focus()` doesn't exactly match how the user would interact with it. The user would <kbd>Tab</kbd> to the button. Maybe this is ok though.

.changeset/pink-tables-explode.md
mayank99, line 1: This changeset could be shortened to be more readable?

Also we should have patch changesets in both react and css for the `<button>` change.

mayank99 (approved):
LGTM 🚀

PR #2024, opened 2024-05-01
fix vite dev servers by directly using TS source files

mayank99:
## Changes

This PR fixes the intermittent errors in vite dev servers (e.g. "Failed to resolve entry for package"). This is done using [`resolve.alias`](https://vitejs.dev/config/shared-options.html#resolve-alias) to resolve to the original TS paths (instead of the compiled JS paths), which can then be processed directly by Vite. I still kept the `swc` TS→JS dev transforms around for nextjs.

I also replaced turbo's broken `.env` implementation with an inline env variable (works on windows too, thanks to pnpm's [`shell-emulator`](https://pnpm.io/cli/run#shell-emulator)). This was necessary for aliases to be resolved correctly _only_ during dev.

Also includes all changes from #2025 (see [thread](https://github.com/iTwin/iTwinUI/pull/2024#discussion_r1586473809)).

## Testing

Tested that the dev servers for all workspaces work correctly.

## Docs

N/A


.github/workflows/build.yml
mayank99, line 1: This error is driving me insane: https://github.com/iTwin/iTwinUI/actions/runs/8911826029/job/24474001187?pr=2024

I didn't even touch cypress in this PR, so I don't understand why it's happening or how to fix it. Our setup looks correct and matches the [instructions from cypress](https://github.com/cypress-io/github-action#pnpm-workspaces).

Ben-Pusey-Bentley: The actions should [automatically be handling the binary caching](https://github.com/cypress-io/github-action/blob/d5edfb9a5f479852f4b9c6f405d63ff13542d49a/.github/workflows/example-start-and-pnpm-workspaces.yml#L51), so I'm not what the exact source of the issue is here. Maybe the automatic caching is dependent on setting up the [pnpm store cache](https://github.com/cypress-io/github-action/blob/d5edfb9a5f479852f4b9c6f405d63ff13542d49a/.github/workflows/example-start-and-pnpm-workspaces.yml#L30-L44)? But I don't know why this is only an issue now, given that cypress used to handle this caching correctly without it before.

apps/react-workshop/vite.config.mjs
rohan-kadkol, line 1: I keep getting the below error when opening the react-workshop and the vite-playground after running `pnpm run dev`.

<details>
<summary>Error</summary>

image

</details>

playgrounds/astro/astro.config.mjs
rohan-kadkol, line 1: I keep getting this error when opening the astro playground:

<details>
<summary>Error</summary>

image

</details>

I get a similar error when opening the remix playground:

<details>
<summary>Error</summary>

image

</details>

rohan-kadkol (commented):
I keep getting a few errors when opening some workspaces. I tried running `pnpm clean`, `pnpm i`, and then `pnpm run build --filter=@itwin/itwinui-react` before `pnpm run dev`, but that didn't fix the issue.

Is this happening only for me?

apps/react-workshop/vite.config.mjs
rohan-kadkol: Just confirmed, yes, I am on the latest commit of the branch (https://github.com/iTwin/iTwinUI/pull/2024/commits/b74d40ff75a60a1ad823759a16e2d3b1bafabf90).

I ran `pnpm store prune`, `rm -rf **/node_modules`, `pnpm i`, `pnpm run build --filter=@itwin/itwinui-react`, and then `pnpm run dev`. However, I still see this error and also the errors from https://github.com/iTwin/iTwinUI/pull/2024#discussion_r1587651407.

mayank99: Hmm I tried the same series of steps just now but cannot reproduce 🤔

rohan-kadkol: After discussing this together, realized that the order of the [`resolve.alias`](https://vitejs.dev/config/shared-options.html#resolve-alias) seems to matter ([docs](https://github.com/rollup/plugins/tree/master/packages/alias#entries)). After [`c8d9074`](https://github.com/iTwin/iTwinUI/pull/2024/commits/c8d9074ee1da545e24d6afb7b8cdab129b40a643), the all workspaces open correctly for me.

Ben-Pusey-Bentley (approved):

internal/helpers/index.js
rohan-kadkol, line 1: non-blocking: Just curious, any reason this is a JS file instead of a TS file?

.github/workflows/build.yml
rohan-kadkol: Not sure why the problem started from this PR. But here are some debugging approaches that come to mind:
* Manually confirming if `/home/runner/.cache/Cypress` exists. E.g. by adding a `ls /home/runner/.cache/Cypress`, or maybe some other better way.
* Trying the solutions mentioned in StackOverflow answers: [1](https://stackoverflow.com/a/62703115/11547064), [2](https://stackoverflow.com/a/58284070/11547064).
* At least temporarily, maybe we can try running `cypress install` in a step before this in case it does not add too much extra time to the run.
* Trying to get more output using the [debug mode](https://github.com/cypress-io/github-action#debugging).

internal/helpers/index.js
mayank99: TS files need extra processing. JS can be imported everywhere/anywhere

.github/workflows/build.yml
mayank99: Fixed in #2025

rohan-kadkol (approved):
LGTM 🚀

PR #2026, opened 2024-05-04
Floating UI `safePolygon` for non-last `MenuItem`s

rohan-kadkol:
## Changes

After PR TODO from #1942 (https://github.com/iTwin/iTwinUI/pull/1942#discussion_r1578323366).

To make the `safePolygon` work for even the non-last `MenuItem`s, looks like we needed to set [`blockPointerEvents: true`](https://floating-ui.com/docs/useHover#blockpointerevents). This PR makes that change.

This change is done in `usePopover` instead of `MenuItem` because `MenuItem` is the only component that has been using the hover trigger so far. So, making this change in `usePopover` is alright since it will not affect any other user component or our own internal component.

Additionally, to make it easier for users to hover over adjacent items instead of keeping the submenu open, [`requireIntent: false`](https://floating-ui.com/docs/usehover#requireintent) is removed (https://github.com/iTwin/iTwinUI/pull/2026#discussion_r1591009663).

## Testing

`safePolygon` works for non-last `MenuItem`s too.

<details>
<summary>Screen-recordings</summary>

Before:

https://github.com/iTwin/iTwinUI/assets/45748283/2dc9de71-81e1-498f-9a5c-9c8e5990229c

After:

https://github.com/iTwin/iTwinUI/assets/45748283/35d29f96-cb46-41d7-8bf4-f8c866682cbe

</details>

## Docs

Added changesets.

### DIFF
--- a/.changeset/smart-flies-reflect.md
+++ b/.changeset/smart-flies-reflect.md
@@ -2,4 +2,4 @@
 "@itwin/itwinui-react": patch
 ---
 
-Previously added leniency in mouse precision when hovering from one `MenuItem` to its submenu now even works for `MenuItem`s that are not at the end of the menu.
+Submenus within a `DropdownMenu` will now consistently require less precision when moving the mouse between the parent item and the submenu.
--- a/.changeset/smart-flies-reflect.md
+++ b/.changeset/smart-flies-reflect.md
@@ -1,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Previously added leniency in mouse precision when hovering from one `MenuItem` to its submenu now even works for `MenuItem`s that are not at the end of the menu.
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -179,7 +179,13 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   const popover = usePopover({
     nodeId,
     visible: isSubmenuVisible,
-    onVisibleChange: setIsSubmenuVisible,
+    onVisibleChange: (v) => {
+      if (!v) {
+        setHasFocusedNodeInSubmenu(false);
+      }
+
+      setIsSubmenuVisible(v);
+    },
     placement: 'right-start',
     interactions: !disabled
       ? {
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -235,7 +235,11 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   };
 
   const onFocus = () => {
-    parent.setHasFocusedNodeInSubmenu?.(true);
+    // Set hasFocusedNodeInSubmenu in a microtask to ensure the submenu stays open reliably.
+    // E.g. Even when hovering into it rapidly.
+    queueMicrotask(() => {
+      parent.setHasFocusedNodeInSubmenu?.(true);
+    });
 
     tree?.events.emit('onNodeFocused', {
       nodeId,
--- a/packages/itwinui-react/src/core/Menu/MenuItem.tsx
+++ b/packages/itwinui-react/src/core/Menu/MenuItem.tsx
@@ -179,13 +179,12 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
   const popover = usePopover({
     nodeId,
     visible: isSubmenuVisible,
-    onVisibleChange: (v) => {
-      if (!v) {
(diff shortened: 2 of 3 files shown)


packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -229,7 +229,11 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
     useHover(floating.context, {
       enabled: !!mergedInteractions.hover,
       delay: 100,
-      handleClose: safePolygon({ buffer: 1, requireIntent: false }),
+      handleClose: safePolygon({
+        buffer: 1,
+        requireIntent: false,
```
mayank99, line 234: Should we remove `requireIntent` now? I believe the main reason for making it `false` was because it was hard to tell if `safePolygon` was working at all.

.changeset/smart-flies-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Previously added leniency in mouse precision when hovering from one `MenuItem` to its submenu now even works for `MenuItem`s that are not at the end of the menu.
```
mayank99, line 5: This could be worded in a way that describes the final outcome more clearly.

```suggestion
Submenus within a `DropdownMenu` will now consistently require less precision when moving the mouse between the parent item and the submenu.
```

packages/itwinui-react/src/core/Popover/Popover.tsx
rohan-kadkol: I believe it was intentional when working on #1942 since we didn't want to take mouse speed in account at that time.

But now, yes, I think it's better to remove it to make it easier for users who want to actually hover over the adjacent item instead of keeping the submenu open because of the safePolygon. Removed (https://github.com/iTwin/iTwinUI/pull/2026/commits/01626db0c6e24cf0d0ed02957f9de62820988be2).

mayank99: Hmm, now I'm noticing some really weird behaviors (at least in Firefox). Screen recording shows that it doesn't work 2/4 times, and worse, it completely breaks the hover trigger after it fails. Maybe that's why we had `requireIntent: false`?

https://github.com/iTwin/iTwinUI/assets/9084735/a8f6332c-3818-402c-a59e-80b93b3b0fd8

rohan-kadkol: After a bit of investigation, I found a few issues.

> and worse, it completely breaks the hover trigger after it fails

Looks like this was a small bug. Fixed in [`8a65e07`](https://github.com/iTwin/iTwinUI/pull/2026/commits/8a65e07b419472c18f2ea85f21285a85af611ef2).

> it doesn't work 2/4 times

Not sure of the exact reasoning of why that's happening. But I think it's because of some race conditions in how `hasFocusedNodeInSubmenu` is being set and used.

So, using something something like `queueMicrotask` or `setTimeout(() => { … }, 0)` keeps the submenus open more reliably but could be considered hacky (https://github.com/iTwin/iTwinUI/pull/2026/commits/405bfec71f9d22cee8a74f5eb6c86ab5175a4079). Open to alternatives.

After those commits:

<details>
<summary>Firefox</summary>

https://github.com/iTwin/iTwinUI/assets/45748283/e3904df6-8148-4b9f-b535-21b2f133a410

</details>

<details>
<summary>Edge</summary>

https://github.com/iTwin/iTwinUI/assets/45748283/eac5e561-1aa5-4a97-9a4e-14cdd06792ae

</details>

packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -179,7 +179,13 @@ export const MenuItem = React.forwardRef((props, forwardedRef) => {
...
     nodeId,
     visible: isSubmenuVisible,
-    onVisibleChange: setIsSubmenuVisible,
+    onVisibleChange: (v) => {
+      if (!v) {
+        setHasFocusedNodeInSubmenu(false);
+      }
+
+      setIsSubmenuVisible(v);
+    },
```
mayank99, line 188: memoize + descriptive variable name

```suggestion
    onVisibleChange: React.useCallback((visible) => {
      if (!visible) {
        setHasFocusedNodeInSubmenu(false);
      }

      setIsSubmenuVisible(visible);
    }, []),
```

rohan-kadkol: Sounds good, done (https://github.com/iTwin/iTwinUI/pull/2026/commits/cf8b329c70f642100a517d94cb241ad3ed810b4b).

mayank99 (approved):

PR #2030, opened 2024-05-06
fix double borders in `ButtonGroup`

mayank99:
## Changes

This fixes https://github.com/iTwin/iTwinUI/pull/2009#discussion_r1569295877 using a combination of the following:
- button-group selectors simplified to only consider `iui-field` and `iui-input-flex-container`
  - `+ *` selector updated to be more like `~ .iui-field`
  - `:first-child`/`:last-child` updated to `:nth-child`/`:nth-last-child` (using [n of S syntax](https://developer.chrome.com/docs/css-ui/css-nth-child-of-s))
- `iui-input-flex-container::before` now inherits border-radius, to simplify the implementation in button-group

This will effectively ignore any direct children inside button-group that aren't one of the pre-sanctioned/expected items. Such rogue children may be added by third-party libraries (e.g. floating-ui) or in the future even added in our own code (e.g. native `[popover]` elements should be located right after the trigger element in DOM).

## Testing

Made sure that the floating-ui guards are ignored in `ColorPicker`. Updated affected visual tests.

| Before | After |
| --- | --- |
| image | image |

## Docs

Added changesets.


packages/itwinui-css/src/button/button-group.scss
```diff
@@ -63,43 +59,38 @@
...
       } @else {
         margin-block-start: -1px;
       }
-    }
 
-    // Add slight border between disabled items
-    &:where(:not(:first-child):not([data-iui-variant='borderless'])) {
-      #{$disabledSelector} {
+      // Add slight border between disabled items
+      &:where(:disabled, [aria-disabled='true']):where(:not([data-iui-variant='borderless'])) {
```
rohan-kadkol, line 93: Should we continue to exclude the first child for the disabled border?

mayank99: no, this is chained to the `~` selector, which means it will never match the first child

packages/itwinui-css/src/button/button-group.scss
```diff
@@ -51,7 +47,7 @@
       }
     }
 
-    #{$childBorderSelector} {
+    &:where(.iui-field:not([data-iui-variant='borderless']), .iui-input-flex-container) {
```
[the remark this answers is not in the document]
mayank99: no, `.iui-input-flex-container::before` now automatically inherits the border-radius.

rohan-kadkol (approved):

PR #2036, opened 2024-05-08
use `List` instead of `Menu` in `Select`

mayank99:
## Changes

Similar to #2034, this replaces the `Menu` component which does not belong in `Select`.

Unlike `ComboBox`, this one was relying on the focus management and keyboard navigation from `Menu`, so I replaced those parts with [`Composite`](https://floating-ui.com/docs/composite) from floating-ui. I had to add a little bit of code to tell `Composite` which indexes are active and that it should automatically move focus when the Select is opened.

## Testing

Manually tested all stories to make sure they're working as intended.

Had to slightly tweak a couple unit tests to make them pass.

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Select/Select.tsx
+++ b/packages/itwinui-react/src/core/Select/Select.tsx
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import cx from 'classnames';
-import { Menu } from '../Menu/Menu.js';
 import { MenuItem } from '../Menu/MenuItem.js';
 import type { MenuItemProps } from '../Menu/MenuItem.js';
 import {
@@ -27,6 +26,8 @@ import { SelectTag } from './SelectTag.js';
 import { SelectTagContainer } from './SelectTagContainer.js';
 import { Icon } from '../Icon/Icon.js';
 import { usePopover } from '../Popover/Popover.js';
+import { List } from '../List/List.js';
+import { Composite, CompositeItem } from '@floating-ui/react';
 
 // ----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -763,3 +774,55 @@ type MultipleSelectButtonProps<T> = {
...
+    });
+  }, []);
+
+  const children = React.useMemo(() => {
+    return React.Children.map(childrenProp, (child, index) =>
+      React.isValidElement(child) ? (
+        <CompositeItem
+          key={index}
+          render={child}
+          ref={child.props.ref || (child as any).ref}
```
rohan-kadkol, line 806: I'm a little confused by what is the difference between the two values. Could you elaborate?

packages/itwinui-react/src/core/Select/Select.tsx
```diff
@@ -763,3 +774,55 @@ type MultipleSelectButtonProps<T> = {
...
+      ),
+    );
+  }, [childrenProp]);
+
+  return (
+    <Composite
+      render={<List as='div' className={cx('iui-menu', className)} />}
+      orientation='vertical'
+      role='listbox'
+      aria-orientation={undefined}
```
rohan-kadkol, line 819: To confirm, is this needed? If so, why?

rohan-kadkol (approved):

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: In React 19, `ref` will be a prop. This code is basically future-proofing us while staying backwards compatible. See https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref

Separately, we should create a PR updating all other instances of `element.ref` to `element.props.ref || element.ref`.

packages/itwinui-react/src/core/Select/Select.tsx
mayank99: For some reason, `Composite` automatically adds `aria-orientation="vertical"`, even though it's not needed and doesn't always make sense. So I'm just overriding/unsetting it.

In the future, we might want to create our own `Composite` implementation to avoid such hacks.

rohan-kadkol: Ohh, I see. But upon inspecting the DOM, I still see `aria-orientation="vertical"` even with this added line. It only goes away after removing `orientation='vertical'` on L817.

mayank99: Huh, that shouldn't be happening. For now I've removed this line ([`2b5a7ad`](https://github.com/iTwin/iTwinUI/pull/2036/commits/2b5a7adc307b49dad869c39fb06728ce856f6011)), since `aria-orientation` is [valid](https://w3c.github.io/aria/#aria-orientation) on `role=listbox`. But we should be careful with this in the future.

rohan-kadkol (approved):

PR #2038, opened 2024-05-09
Fix docs site import error when running the `dev` script

rohan-kadkol:
## Changes

After #2024, @FlyersPh9 and I started noticing the below error when opening any docs site's pages with the path `/docs/…`

image

Seems like the `?url` in the import in LiveExample is causing some issue.

https://github.com/iTwin/iTwinUI/blob/f699eb774e74b1ee08a0edb7729f1e5446c2f8aa/apps/website/src/components/LiveExample.astro#L6

~This PR adds a [`resolve.alias`](https://vitejs.dev/config/shared-options.html#resolve-alias) for this import too.~ Instead changed the `resolve.alias` to be find and replacement rules (https://github.com/iTwin/iTwinUI/pull/2038#discussion_r1596099089).

## Testing

Confirmed that the docs site's `/docs/…` pages open without any errors. Also confirmed that other workspaces continue working.

## Docs

N/A

### DIFF
--- a/internal/helpers/index.js
+++ b/internal/helpers/index.js
@@ -6,15 +6,28 @@ import * as path from 'node:path';
 
 export const itwinuiReactAliases =
   process.env.NODE_ENV === 'development'
-    ? {
-        '@itwin/itwinui-react/styles.css?url': path.resolve(
-          '../../packages/itwinui-react/styles.css?url',
-        ),
-        '@itwin/itwinui-react/styles.css': path.resolve(
-          '../../packages/itwinui-react/styles.css',
-        ),
-        '@itwin/itwinui-react': path.resolve(
-          '../../packages/itwinui-react/src',
-        ),
-      }
-    : {};
+    ? [
+        {
+          find: /^@itwin\/itwinui-react$/,
+          replacement: path.resolve(
+            '../../packages/itwinui-react/src/index.ts',
+          ),
+        },
+        {
+          find: /^@itwin\/itwinui-react\/styles.css(.*)$/,
+          replacement: path.resolve(
+            '../../packages/itwinui-react/styles.css$1',
+          ),
+        },
+        {
+          find: /^@itwin\/itwinui-react\/react-table$/,
+          replacement: path.resolve(
+            '../../packages/itwinui-react/react-table.d.ts',
+          ),
+        },
+        {
+          find: /^@itwin\/itwinui-react\/(.*)$/,
+          replacement: path.resolve('../../packages/itwinui-react/src/$1'),
+        },
+      ]
+    : [];
--- a/internal/helpers/index.js
+++ b/internal/helpers/index.js
@@ -8,21 +8,13 @@ export const itwinuiReactAliases =
   process.env.NODE_ENV === 'development'
     ? [
         {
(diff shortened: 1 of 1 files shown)


Ben-Pusey-Bentley (approved):

internal/helpers/index.js
```diff
@@ -7,6 +7,9 @@ import * as path from 'node:path';
 export const itwinuiReactAliases =
   process.env.NODE_ENV === 'development'
     ? {
+        '@itwin/itwinui-react/styles.css?url': path.resolve(
+          '../../packages/itwinui-react/styles.css?url',
+        ),
         '@itwin/itwinui-react/styles.css': path.resolve(
           '../../packages/itwinui-react/styles.css',
```
mayank99, line 14: Since these two are very similar, maybe we could do a partial find/replace instead? https://www.npmjs.com/package/@rollup/plugin-alias#regular-expression-aliases

mayank99 (approved):

internal/helpers/index.js
mayank99: I just realized a similar problem would happen when importing from `@itwin/itwinui-react/react-table`, so it's probably better do a partial replace, which will ensure everything works.

rohan-kadkol: Sounds good, I switched to using the find and replacement (https://github.com/iTwin/iTwinUI/pull/2038/commits/853223eb78f045e167b9baa5b612ecbf2aa61943).

I tested that it works for react-table too by adding `import type { Column } from '@itwin/itwinui-react/react-table';` to `App.tsx` from the vite playground. This worked without a separate rule for react-table ([rule](https://github.com/iTwin/iTwinUI/pull/2038/commits/0fb67ef69e4e39a9197680ff4029ebb3b6732578#diff-f4d8f020910bbda54d6058e9d99cbdeba33ded5431bc2807fb6bcf45630f66c6L22-L27)). However, doing something like `import '@itwin/itwinui-react/react-table';` sometimes gives errors:

<details>
<summary>Error</summary>

image

</details>

So, I add a specific rule for `react-table` instead of a general catch-all rule (https://github.com/iTwin/iTwinUI/pull/2038/files/0fb67ef69e4e39a9197680ff4029ebb3b6732578..19e46b5901c4506ce1934d5758a589a2be99b532).

mayank99: To be clear, I'm suggesting something like this:

```js
{
  find: /^@itwin\/itwinui-react\/(.*)/,
  replacement: path.resolve(
    '../../packages/itwinui-react/$1',
  ),
},
{
  '@itwin/itwinui-react': path.resolve(
    '../../packages/itwinui-react/src',
  ),
},
```

rohan-kadkol: Ohh, I see, something like that works in the docs website but sometimes gives an error when doing `import '@itwin/itwinui-react/react-table';`. But the real case of importing types (e.g. `import type { Column } from '@itwin/itwinui-react/react-table';`) works.

So, I made changes similar to your suggestion (https://github.com/iTwin/iTwinUI/pull/2038/commits/3ff1f9581c8bd24383f17bc5f3a8c3781e073949).

mayank99 (approved):

PR #2042, opened 2024-05-10
Fix table empty state horizontal centering

### LINKED ISSUE #2037: Empty table state's elements are not centered

When a table is empty, the greyed-out elements (describing the empty table state) are not centered horizontally. This bug is seen clearer if the table's width is smaller

### Expected Behavior

When a table is empty, the greyed-out elements (describing the empty table state) should be centered horizontally

### Link to minimal repro

https://stackblitz.com/edit/github-girbn1-d2ghuj

### Steps To Reproduce

1. Simply open the code sandbox link provided, and it should display the empty table state to the right of the red box.
2. If the greyed-out elements appear to be centered, try to play with the sandbox's browser pane to minimize the table, that way the greyed-out elements could appear less centered.

Screenshot (10)

### Anything else?

_No response_

FlyersPh9:
## Changes

- Change empty state horizontal spacing method from using `inset-inline-start` & `transform` to using `margin-inline`.  Closes #2037.

## Testing

- Visual tests and manually checked in Chrome Canary.
- The previous method was implemented by @rohan-kadkol, so I checked with him if there were any reasons why `margin-inline` might not work:

> I had added it a while back when trying to add a dummy shadow dom in the table body such that the table body always has the same width as the header, even when there are no rows in the table body. That PR also introduced some minor CSS changes.
> But after the Table PR a few weeks ago that makes the table the scrollable container instead of the table body, we were able to remove hacks, such as the dummy shadow dom hack from that old PR.
> In that Table PR, I had removed the dummy shadow DOM but didn't revert those CSS styles since everything was working based on my testing. So I think there is no reason anymore for those CSS changes from the dummy shadow DOM PR. Thus, they can be removed/edited, unless they cause some regression upon testing.

## Docs

n/a


Ben-Pusey-Bentley (approved):
Looks good to me. You do have a failing react visual test, but I think it is just being flaky, based on the story related to test looking fine.

.changeset/fluffy-parents-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
```
Ben-Pusey-Bentley, line 2: This change will also affect the react package, so there should be a react changeset as well.

FlyersPh9: Added a React changeset.

.changeset/big-owls-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Fix table empty state horizontal spacing when table is very narrow.
```
rohan-kadkol, line 5: To confirm, is it "spacing" or "alignment" that has changed?

packages/itwinui-css/src/table/base.scss
```diff
@@ -371,8 +371,7 @@
   > .iui-table-empty {
     text-align: center;
     position: sticky;
```
rohan-kadkol, line 373: Since we're using margin and not `position: sticky` + `inset-inline-start: 50%`, is `position: sticky` needed any more?

packages/itwinui-css/src/table/base.scss
```diff
@@ -371,8 +371,7 @@
...
-    inset-inline-start: 50%;
-    transform: translateX(-50%);
+    margin-inline: auto;
```
rohan-kadkol, line 374: Although this fixes the problem mentioned in the attached issue, I guess there are some edge cases where the empty content is not centered.

Maybe those cases cannot be addressed because the table is narrower than the empty content width and thus, we do not need to cover this case?

<details>
<summary>Example un-centered case when the table is very narrow</summary>

https://github.com/iTwin/iTwinUI/assets/45748283/8e71a970-3a79-4bb8-9186-45c6fc319515

</details>

rohan-kadkol (commented):
Although it fixes the issue

packages/itwinui-css/src/table/base.scss
FlyersPh9: I tried dropping one of our illustration SVGs inside the empty state and it seemed to work ok.
table-empty-state-resize

Perhaps their SVG has a hardcoded size?  Just confirmed adding `width='400px'` to the `<svg>` does indeed break it.

.changeset/big-owls-flow.md
FlyersPh9: Updated to use "alignment".

packages/itwinui-css/src/table/base.scss
FlyersPh9: I've removed `position: sticky`.

rohan-kadkol (approved):

PR #2043, opened 2024-05-10
Disable zoom on double tap of button

FlyersPh9:
## Changes

- Add `touch-action: manipulation;` to `.iui-button-base` which [disables zoom on double tap of buttons](https://dbushell.com/2024/03/10/css-button-styles-you-might-not-know/#touch-actions).

Simulator Screen Recording - iPhone 15 Pro - 2024-05-10 at 10 09 30

## Testing

- Tested in iOS Simulator against `<button>` and `<a>`.

## Docs

n/a


packages/itwinui-css/src/button/base.scss
```diff
@@ -19,6 +19,7 @@ $disabledButtonSelectors: '[disabled], :disabled, [aria-disabled="true"], [data-
   padding-block: var(--_iui-button-padding-block);
   padding-inline: var(--_iui-button-padding-inline);
   text-decoration: none;
+  touch-action: manipulation;
```
mayank99, line 22: Maybe this should go in `iui-button-base`?

One thing I would like to verify is that it doesn't break links used as buttons, e.g. `<a class="iui-button-base iui-button">`

.changeset/strong-bags-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
+---
+
+Disable zoom on double tap of button.
```
mayank99, line 5: This changeset should be duplicated for `@itwin/itwinui-react`. Also it should be reworded to more clearly the describe *what* is impacted and *how*.

packages/itwinui-css/src/button/base.scss
FlyersPh9: Moved to `.iui-button-base` and tested with the breadcrumb anchors.

I'm wondering if this should actually be moved into `.iui-field` to be applied to `<input>` and `<textarea>` as well. 🤔

.changeset/strong-bags-type.md
FlyersPh9: Added changeset for React and updated the description.

packages/itwinui-css/src/button/base.scss
mayank99: I think it's fine in `iui-button-base`. Input/textarea likely have their own double tap behaviors (e.g. select text)

.changeset/strong-bags-type.md
mayank99: CSS classes should not be mentioned in the React changeset.

Also, could you word these in a way that frontloads the change, and *then* explains the reason?

FlyersPh9: That's a good point, updated to use your suggestion.

Was going to add "buttons and tabs" but tabs are `<button>`s, so left it as is.

mayank99 (approved):

PR #2048, opened 2024-05-14
ComboBox controlled state behavior bug fix

### LINKED ISSUE #2046: ComboBox controlled state not working

The controlled state in ComboBox does not work as in other components. i.e. if a user chooses a value from the combobox, the value in the combobox changes to it even though the value passed to the value prop is not changed.

### Expected Behavior

The value passed to the `value` prop should take priority.

### Link to minimal repro

https://stackblitz.com/edit/github-dudecs?file=src%2FApp.tsx

### Steps To Reproduce

1. Select any other item in the ComboBox.
2. Even though `value="C"`, the ComboBox value changes.
3. This doesn't happen in Select.

### Anything else?

_No response_

rohan-kadkol:
## Changes

Fixes #2046

WIP

### DIFF
--- a/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
@@ -14,12 +14,13 @@ import {
   useLayoutEffect,
   AutoclearingHiddenLiveRegion,
   useId,
+  useControlledState,
 } from '../../utils/index.js';
 import { usePopover } from '../Popover/Popover.js';
 import type { InputContainerProps, CommonProps } from '../../utils/index.js';
 import {
-  ComboBoxActionContext,
-  comboBoxReducer,
+  // ComboBoxActionContext,
+  // comboBoxReducer,
   ComboBoxRefsContext,
   ComboBoxStateContext,
 } from './helpers.js';
@@ -219,43 +220,73 @@ export const ComboBox = React.forwardRef(
     }
 
     // Get indices of selected elements in options array when we have selected values.
-    const getSelectedIndexes = React.useCallback(() => {
-      if (isMultipleEnabled(valueProp, multiple)) {
-        const indexArray: number[] = [];
-        valueProp?.forEach((value) => {
-          const indexToAdd = options.findIndex(
-            (option) => option.value === value,
-          );
-          if (indexToAdd > -1) {
-            indexArray.push(indexToAdd);
-          }
-        });
-        return indexArray;
-      } else {
-        return options.findIndex((option) => option.value === valueProp);
-      }
-    }, [multiple, options, valueProp]);
+    const getSelectedIndexes = React.useCallback(
+      (value: typeof valueProp) => {
+        if (isMultipleEnabled(value, multiple)) {
+          const indexArray: number[] = [];
+          value?.forEach((value) => {
+            const indexToAdd = options.findIndex(
+              (option) => option.value === value,
+            );
+            if (indexToAdd > -1) {
+              indexArray.push(indexToAdd);
+            }
+          });
+          return indexArray;
+        } else {
+          return options.findIndex((option) => option.value === value);
+        }
+      },
+      [multiple, options],
+    );
 
-    // Reducer where all the component-wide state is stored
-    const [{ isOpen, selected, focusedIndex }, dispatch] = React.useReducer(
-      comboBoxReducer,
-      {
-        isOpen: false,
-        selected: getSelectedIndexes(),
-        focusedIndex: -1,
+    const [value, setValue] = useControlledState(
+      multiple ? [] : undefined,
+      getSelectedIndexes(valueProp),
+      (newValue) => {
+        console.log('newValue', `${newValue}`);
+        // if (isSingleOnChange(onChange, multiple)) {
+        //   onChange?.(newValue);
+        // const q : React.SetStateAction<T>;
+        // }
       },
     );
+
+    // React.useEffect(() => {
+    //   setValue(valueProp);
+    // }, [valueProp, setValue]);
+
+    // Reducer where all the component-wide state is stored
+    // const [{ isOpen, selected, focusedIndex }, dispatch] = React.useReducer(
+    //   comboBoxReducer,
+    //   {
+    //     isOpen: false,
+    //     selected: getSelectedIndexes(),
+    //     focusedIndex: -1,
+    //   },
+    // );
+
+    const selected = value;
+
+    // const selected = getSelectedIndexes(value);
+
+    const [isOpen, setIsOpen] = React.useState(false);
+    const [focusedIndex, setFocusedIndex] = React.useState<number>(-1);
 
     const onShowRef = useLatestRef(onShowProp);
     const onHideRef = useLatestRef(onHideProp);
 
     const show = React.useCallback(() => {
-      dispatch({ type: 'open' });
+      // dispatch({ type: 'open' });
+      setIsOpen(true);
+
       onShowRef.current?.();
     }, [onShowRef]);
 
     const hide = React.useCallback(() => {
-      dispatch({ type: 'close' });
+      // dispatch({ type: 'close' });
+      setIsOpen(false);
+
       onHideRef.current?.();
     }, [onHideRef]);
 
@@ -266,13 +297,16 @@ export const ComboBox = React.forwardRef(
         // Reset the filtered list (does not reset when multiple enabled)
         if (!multiple) {
           setFilteredOptions(optionsRef.current);
-          dispatch({ type: 'focus', value: undefined });
+          // dispatch({ type: 'focus', value: undefined });
+          setFocusedIndex(-1);
         }
       }
       // When the dropdown closes
       else {
         // Reset the focused index
-        dispatch({ type: 'focus', value: undefined });
+        // dispatch({ type: 'focus', value: undefined });
+        setFocusedIndex(-1);
+
         // Reset the input value if not multiple
         if (!isMultipleEnabled(selected, multiple)) {
           setInputValue(
@@ -297,7 +331,9 @@ export const ComboBox = React.forwardRef(
       } else {
         setFilteredOptions(options);
       }
-      dispatch({ type: 'focus', value: undefined });
+      // dispatch({ type: 'focus', value: undefined });
+      setFocusedIndex(-1);
+
       // Only need to call on options update
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [options]);
@@ -321,44 +357,45 @@ export const ComboBox = React.forwardRef(
             ),
         );
         if (focusedIndex != -1) {
-          dispatch({ type: 'focus', value: -1 });
+          // dispatch({ type: 'focus', value: -1 });
+          setFocusedIndex(-1);
         }
         inputProps?.onChange?.(event);
       },
       [filterFunction, focusedIndex, inputProps, optionsRef, show],
     );
 
-    // When the value prop changes, update the selected index/indices
-    React.useEffect(() => {
-      if (isMultipleEnabled(valueProp, multiple)) {
-        if (valueProp) {
-          // If user provided array of selected values
-          const indexes = valueProp.map((value) => {
-            return options.findIndex((option) => option.value === value);
-          });
-          dispatch({
-            type: 'multiselect',
-            value: indexes.filter((index) => index !== -1), // Add available options
-          });
-        } else {
-          // if user provided one value or undefined
-          dispatch({
-            type: 'multiselect',
-            value: [], // Add empty list
-          });
-        }
-      } else {
-        dispatch({
-          type: 'select',
-          value: options.findIndex((option) => option.value === valueProp),
-        });
-      }
-    }, [valueProp, options, multiple]);
+    // // When the value prop changes, update the selected index/indices
+    // React.useEffect(() => {
+    //   if (isMultipleEnabled(valueProp, multiple)) {
+    //     if (valueProp) {
+    //       // If user provided array of selected values
+    //       const indexes = valueProp.map((value) => {
+    //         return options.findIndex((option) => option.value === value);
+    //       });
+    //       dispatch({
+    //         type: 'multiselect',
+    //         value: indexes.filter((index) => index !== -1), // Add available options
+    //       });
+    //     } else {
+    //       // if user provided one value or undefined
+    //       dispatch({
+    //         type: 'multiselect',
+    //         value: [], // Add empty list
+    //       });
+    //     }
+    //   } else {
+    //     dispatch({
+    //       type: 'select',
+    //       value: options.findIndex((option) => option.value === valueProp),
+    //     });
+    //   }
+    // }, [valueProp, options, multiple]);
 
     const isMenuItemSelected = React.useCallback(
       (index: number) => {
         if (isMultipleEnabled(selected, multiple)) {
-          return !!selected.includes(index as number);
+          return !!selected?.includes(index as number);
         } else {
           return selected === index;
         }
@@ -417,7 +454,10 @@ export const ComboBox = React.forwardRef(
             ? 'removed'
             : 'added';
           const newArray = selectedChangeHandler(__originalIndex, actionType);
-          dispatch({ type: 'multiselect', value: newArray });
+
+          // dispatch({ type: 'multiselect', value: newArray });
+          setValue(newArray);
+
           onChangeHandler(__originalIndex, actionType, newArray);
 
           // update live region
@@ -428,12 +468,15 @@ export const ComboBox = React.forwardRef(
               .join(', '),
           );
         } else {
-          dispatch({ type: 'select', value: __originalIndex });
+          // dispatch({ type: 'select', value: __originalIndex });
+          setValue(__originalIndex);
+
           hide();
           onChangeHandler(__originalIndex);
         }
       },
       [
+        setValue,
         selectedChangeHandler,
         isMenuItemSelected,
         multiple,
@@ -533,62 +576,63 @@ export const ComboBox = React.forwardRef(
       <ComboBoxRefsContext.Provider
         value={{ inputRef, menuRef, optionsExtraInfoRef }}
       >
-        <ComboBoxActionContext.Provider value={dispatch}>
-          <ComboBoxStateContext.Provider
-            value={{
-              id,
-              isOpen,
-              focusedIndex,
-              onClickHandler,
-              enableVirtualization,
-              filteredOptions,
-              getMenuItem,
-              multiple,
-              popover,
-              show,
-              hide,
-            }}
+        {/* <ComboBoxActionContext.Provider value={dispatch}> */}
+        <ComboBoxStateContext.Provider
+          value={{
+            id,
+            isOpen,
+            focusedIndex,
+            setFocusedIndex,
+            onClickHandler,
+            enableVirtualization,
+            filteredOptions,
+            getMenuItem,
+            multiple,
+            popover,
+            show,
+            hide,
+          }}
+        >
+          <ComboBoxInputContainer
+            ref={forwardedRef}
+            disabled={inputProps?.disabled}
+            {...rest}
           >
-            <ComboBoxInputContainer
-              ref={forwardedRef}
-              disabled={inputProps?.disabled}
-              {...rest}
-            >
-              <>
-                <ComboBoxInput
-                  value={inputValue}
-                  disabled={inputProps?.disabled}
-                  {...inputProps}
-                  onChange={handleOnInput}
-                  selectTags={
-                    isMultipleEnabled(selected, multiple)
-                      ? selected.map((index) => {
-                          const item = optionsRef.current[index];
-                          return (
-                            <SelectTag key={item.label} label={item.label} />
-                          );
-                        })
-                      : undefined
-                  }
-                />
-              </>
-              <ComboBoxEndIcon
-                {...endIconProps}
+            <>
+              <ComboBoxInput
+                value={inputValue}
                 disabled={inputProps?.disabled}
-                isOpen={isOpen}
+                {...inputProps}
+                onChange={handleOnInput}
+                selectTags={
+                  isMultipleEnabled(selected, multiple)
+                    ? selected?.map((index) => {
+                        const item = optionsRef.current[index];
+                        return (
+                          <SelectTag key={item.label} label={item.label} />
+                        );
+                      })
+                    : undefined
+                }
               />
+            </>
+            <ComboBoxEndIcon
+              {...endIconProps}
+              disabled={inputProps?.disabled}
+              isOpen={isOpen}
+            />
 
-              {multiple ? (
-                <AutoclearingHiddenLiveRegion text={liveRegionSelection} />
-              ) : null}
-            </ComboBoxInputContainer>
-            <ComboBoxMenu as='div' {...dropdownMenuProps}>
-              {filteredOptions.length > 0 && !enableVirtualization
-                ? filteredOptions.map(getMenuItem)
-                : emptyContent}
-            </ComboBoxMenu>
-          </ComboBoxStateContext.Provider>
-        </ComboBoxActionContext.Provider>
+            {multiple ? (
+              <AutoclearingHiddenLiveRegion text={liveRegionSelection} />
+            ) : null}
+          </ComboBoxInputContainer>
+          <ComboBoxMenu as='div' {...dropdownMenuProps}>
+            {filteredOptions.length > 0 && !enableVirtualization
+              ? filteredOptions.map(getMenuItem)
+              : emptyContent}
+          </ComboBoxMenu>
+        </ComboBoxStateContext.Provider>
+        {/* </ComboBoxActionContext.Provider> */}
       </ComboBoxRefsContext.Provider>
     );
   },
--- a/packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
@@ -15,7 +15,7 @@ import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 import { ComboBoxMultipleContainer } from './ComboBoxMultipleContainer.js';
 import {
   ComboBoxStateContext,
-  ComboBoxActionContext,
+  // ComboBoxActionContext,
   ComboBoxRefsContext,
 } from './helpers.js';
 
@@ -30,6 +30,7 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
     isOpen,
     id,
     focusedIndex,
+    setFocusedIndex,
     enableVirtualization,
     multiple,
     onClickHandler,
@@ -37,7 +38,7 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
     show,
     hide,
   } = useSafeContext(ComboBoxStateContext);
-  const dispatch = useSafeContext(ComboBoxActionContext);
+  // const dispatch = useSafeContext(ComboBoxActionContext);
   const { inputRef, menuRef, optionsExtraInfoRef } =
     useSafeContext(ComboBoxRefsContext);
   const refs = useMergedRefs(inputRef, popover.refs.setReference, forwardedRef);
@@ -72,12 +73,15 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
           if (focusedIndexRef.current === -1) {
             const currentElement =
               menuRef.current?.querySelector('[data-iui-index]');
-            return dispatch({
-              type: 'focus',
-              value: Number(
-                currentElement?.getAttribute('data-iui-index') ?? 0,
-              ),
-            });
+            // return dispatch({
+            //   type: 'focus',
+            //   value: Number(
+            //     currentElement?.getAttribute('data-iui-index') ?? 0,
+            //   ),
+            // });
+            return setFocusedIndex(
+              Number(currentElement?.getAttribute('data-iui-index') ?? 0),
+            );
           }
 
           // If virtualization is enabled, dont let round scrolling
@@ -101,7 +105,8 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
             nextIndex = Number(nextElement?.getAttribute('data-iui-index'));
 
             if (nextElement) {
-              return dispatch({ type: 'focus', value: nextIndex });
+              // return dispatch({ type: 'focus', value: nextIndex });
+              return setFocusedIndex(nextIndex);
             }
           } while (nextIndex !== focusedIndexRef.current);
           break;
@@ -127,12 +132,16 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
           }
 
           if (focusedIndexRef.current === -1) {
-            return dispatch({
-              type: 'focus',
-              value:
-                Object.values(optionsExtraInfoRef.current)?.[length - 1]
-                  .__originalIndex ?? -1,
-            });
+            // return dispatch({
+            //   type: 'focus',
+            //   value:
+            //     Object.values(optionsExtraInfoRef.current)?.[length - 1]
+            //       .__originalIndex ?? -1,
+            // });
+            return setFocusedIndex(
+              Object.values(optionsExtraInfoRef.current)?.[length - 1]
+                .__originalIndex ?? -1,
+            );
           }
 
           let prevIndex = focusedIndexRef.current;
@@ -146,7 +155,8 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
             prevIndex = Number(prevElement?.getAttribute('data-iui-index'));
 
             if (prevElement) {
-              return dispatch({ type: 'focus', value: prevIndex });
+              // return dispatch({ type: 'focus', value: prevIndex });
+              return setFocusedIndex(prevIndex);
             }
           } while (prevIndex !== focusedIndexRef.current);
           break;
@@ -173,7 +183,8 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
       }
     },
     [
-      dispatch,
+      // dispatch,
+      setFocusedIndex,
       enableVirtualization,
       focusedIndexRef,
       isOpen,
--- a/packages/itwinui-react/src/core/ComboBox/helpers.ts
+++ b/packages/itwinui-react/src/core/ComboBox/helpers.ts
@@ -6,61 +6,61 @@ import * as React from 'react';
 import type { SelectOption } from '../Select/Select.js';
 import type { usePopover } from '../Popover/Popover.js';
 
-type ComboBoxAction =
-  | { type: 'multiselect'; value: number[] }
-  | { type: 'open' }
-  | { type: 'close' }
-  | { type: 'select'; value: number }
-  | { type: 'focus'; value: number | undefined };
+// type ComboBoxAction =
+//   | { type: 'multiselect'; value: number[] }
+//   | { type: 'open' }
+//   | { type: 'close' }
+//   | { type: 'select'; value: number }
+//   | { type: 'focus'; value: number | undefined };
 
-export const comboBoxReducer = (
-  state: {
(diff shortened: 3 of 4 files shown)


packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -412,33 +408,43 @@ export const ComboBox = React.forwardRef(
...
           // update live region
+          // TODO: Confirm if need to show live region when controlled state and the onChangeProp does not update the value
           setLiveRegionSelection(
             newArray
-              .map((item) => optionsRef.current[item]?.label)
+              .map(
+                (item) =>
+                  optionsRef.current.find((option) => option.value === item)
+                    ?.label,
+              )
```
rohan-kadkol, line 431: I'm wondering when we need to set the live region. Do we need set it when the `value` changes externally (so that the user is aware that the value has changed) or whenever the user explicitly selects/deselects an option?

If it's the latter, how to implement this in the controlled state? Example case: the user selects/deselects an item but the application does not update the state in the `onChange`.

In this example case, should we set the live region? If not, then how do we know when the application has updated the `value` after the user has selected/deselected the item vs. when the `value` has changed externally without the user selecting/deselecting an option?

mayank99: The live region should update whenever the value actually changes.

The point of controlled state is not to allow applications to ignore the user's actions, but to instead have an _additional_ way to set the value. If the application does not respect the state provided by `onChange`, that's a horrendous bug (or a great way to annoy users).

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -699,12 +726,21 @@ it('should programmatically clear value', async () => {
...
+  // In controlled state, passed value should take priority
+  expect(input).toHaveValue('Item 1');
 
   rerender(
     <ComboBox options={options} onChange={mockOnChange} value={undefined} />,
   );
   expect(mockOnChange).not.toHaveBeenCalledTimes(2);
+
+  // value={undefined} = uncontrolled state. Thus, the value should be the selected value
+  expect(input).toHaveValue('Item 2');
```
mayank99, line 739: Can you explain how this works? Passing `value={undefined}` will randomly revert the component to some stale internal state?

mayank99: > Earlier, passing `value={undefined}` = no item selected. But after this PR fixed the controlled state issue, `value={undefined}` = uncontrolled state. After this PR, to force the "no item selected" state, `value={null}` needs to be passed. This is similar to `Select` ([Select code](https://github.com/iTwin/iTwinUI/blob/1232186b654be7769da7477663baf8e7ce239777/packages/itwinui-react/src/core/Select/Select.tsx#L578)).

I think this would be a breaking change. Passing `value={undefined}` should continue to reset the value. We can make it uncontrolled after resetting, but the resetting is important.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -266,38 +275,51 @@ export const ComboBox = React.forwardRef(
...
+            : '',
         );
-      } else {
-        setFilteredOptions(options);
       }
-      dispatch({ type: 'focus', value: undefined });
+
       // Only need to call on options update
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [options]);
```
mayank99, line 325: I'm a little confused by this particular `useEffect`. What do you mean by "reconfigure internals"?

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -321,84 +343,58 @@ export const ComboBox = React.forwardRef(
...
-          dispatch({ type: 'focus', value: -1 });
+          setFocusedIndex(-1);
         }
         inputProps?.onChange?.(event);
       },
       [filterFunction, focusedIndex, inputProps, optionsRef, show],
     );
 
-    // When the value prop changes, update the selected index/indices
-    React.useEffect(() => {
```
mayank99, line 332: Where is all this code handled now?

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
rohan-kadkol: > Can you explain how this works? Passing `value={undefined}` will randomly revert the component to some stale internal state?

Currently, passing `value={undefined}` reverts back to what is set in the uncontrolled state. Here, the uncontrolled state is `Item 2` since that was clicked earlier on [L726](https://github.com/iTwin/iTwinUI/pull/2048/files/6355ce0e8062070216e558d09a8cf361664c2a71#diff-669ab1d22f5bd28b68a345f18037f496716dea1e95bffba5ff13bf7d6d064b77R726).

I wasn't sure if it was alright that the uncontrolled state can be out of sync with the controlled state. But since `useControlledState` worked that way, I didn't spend too much time thinking about updating that code. But I can investigate this more if the behavior is not ideal.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: This is currently taken care of by the `useControlledState`. Whenever the `valueProp` changes, `useControlledState` updates the `value` accordingly.

Whenever the `value` changes, `selectedIndices` is updated because `value` is a part of the dependency array ([code](https://github.com/iTwin/iTwinUI/blob/6355ce0e8062070216e558d09a8cf361664c2a71/packages/itwinui-react/src/core/ComboBox/ComboBox.tsx#L234-L253)).

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: I was hoping to say something like "update the internal state (e.g. filteredOptions, focusedIndex, etc.) whenever the options change". I can change "internals" to state if you feel that's clearer. Or you can suggest a replacement too.

Unrelated: I will try to remove the `eslint-disable` by introducing a variable called something like `previousOptions` and run this logic whenever `options !== previousOptions`.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -97,7 +101,7 @@ export type ComboBoxProps<T> = {
   /**
    * Native input element props.
    */
-  inputProps?: React.ComponentProps<typeof Input>;
+  inputProps?: Omit<React.ComponentProps<typeof Input>, 'defaultValue'>;
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, looks like a leftover from something I was playing around with earlier. Removed the `Omit` (https://github.com/iTwin/iTwinUI/pull/2048/commits/aa454a5b7840487716521093cab1101d636e110c).

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -218,11 +223,22 @@ export const ComboBox = React.forwardRef(
...
 
-    // Get indices of selected elements in options array when we have selected values.
-    const getSelectedIndexes = React.useCallback(() => {
-      if (isMultipleEnabled(valueProp, multiple)) {
+    const [value, setValue] = useControlledState(
+      valueProp !== undefined ? valueProp : defaultValue,
+      valueProp !== undefined ? valueProp : undefined,
+
+      // we call onChangeProp manually after calling setSelected
+      () => {},
```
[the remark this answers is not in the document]
rohan-kadkol: I think we still need it to always give preference to the `valueProp` and only use the uncontrolled state when `valueProp={undefined}`?

I can use a regular `useState`, but then will need to duplicate code from `useControlledState` (e.g. `const state = controlledState !== undefined ? controlledState : uncontrolledState;`) and create extra variables in `ComboBox.tsx`.

Ben-Pusey-Bentley (approved):
Approving since I don't have any issues with this PR that haven't already been brought up in comments.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: Ah ok, then I think you can just leave this one out? It's an optional param

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: > I will try to remove the `eslint-disable` by introducing a variable called something like `previousOptions` and run this logic whenever `options !== previousOptions`.

This is a good idea. I'll try to suggest improvements (if any) after you do this.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
rohan-kadkol: Sounds good, made those changes in [`318a83d`](https://github.com/iTwin/iTwinUI/pull/2048/commits/318a83d8fbfc08cb5c55b0eedf5196466b65ff63) + [`2bbb064`](https://github.com/iTwin/iTwinUI/pull/2048/commits/2bbb064904cf0fbd39ec4b6ce54a1ec5927eaf38).

.changeset/happy-mirrors-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": minor
+---
+
+Fixed a bug in `ComboBox` where the controlled state (`value` prop) was not given priority over the uncontrolled state. As a result, passing `value={undefined}` now reverts to the uncontrolled state instead of resetting the value. To reset the value, you now need to pass `value={null}`.
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, right, updated the changeset (https://github.com/iTwin/iTwinUI/pull/2048/commits/2b27f4191c268a109a53d367a0d832b4a46b299e) to mention to use `defaultValue` instead.

I thought of combining the new `defaultValue` prop changeset with this one since this changeset also mentions `defaultValue`. But then I thought against it since I felt that the new `defaultValue` prop might be missed by the users who just scan the first few words/sentences of each bullet point.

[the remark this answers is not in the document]
mayank99: > * passing `value={undefined}` now reverts to the uncontrolled state instead of resetting the value. To reset the value, you now need to pass `value={null}`.

This part should be undone, as discussed in [above thread](https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1601987520)

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: > > I will try to remove the `eslint-disable` by introducing a variable called something like `previousOptions` and run this logic whenever `options !== previousOptions`.
> 
> This is a good idea. I'll try to suggest improvements (if any) after you do this.

Great, made this change (https://github.com/iTwin/iTwinUI/pull/2048/commits/f1a202da42e6309edb615767dce3e5bd3e829bad).

.changeset/happy-mirrors-stare.md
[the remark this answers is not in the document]
rohan-kadkol: Yes, handled in https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605156984.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: If you're storing the previous state, then I think you could remove the `useEffect` altogether? ([docs](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes))

.changeset/happy-mirrors-stare.md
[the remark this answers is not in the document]
mayank99: > passing `value={undefined}` or `value={null}` will continue to reset the value. However, `value={null}` will switch to/remain in the *controlled* state. To reset the value and switch to/remain in the *uncontrolled* state, use `value={undefined}`

Is this necessary to mention in the changeset? The `value={undefined}` behavior is the same as before, and `value={null}` already has its own changeset.

Instead of the changeset, it might be useful to mention the explicit `value={undefined}` behavior in the jsdocs.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
mayank99: Similar to https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605191007, you could remove the `useLayoutEffect`?

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: Ohh, yes, I think it can be removed. Just wondering which of the two is better:
* not have a `useEffect` so that the rendering is essentially restarted to avoid rendering stale values
* using a `useLayoutEffect` instead so that the state is updated before the render

mayank99: `useLayoutEffect` still runs after React completes the first render. It is basically equivalent to `useEffect`, with the only difference being that it runs before the _browser_ paints. So there is no benefit to using `useLayoutEffect` over just changing the value during the render.

rohan-kadkol: Ohh, I see, sounds good. In [`dad30d0`](https://github.com/iTwin/iTwinUI/pull/2048/commits/dad30d0cac5691fc0669734a2ac476e94cc10d71), I removed the `useEffect` from here and also the `useLayoutEffect` from where `previousValue` was being used ([related conversation](https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605156984)).

> I'm a little confused by this particular `useEffect`. What do you mean by "reconfigure internals"?

I renamed this to "reconfigure internal state" to try to make it clearer. Feel free to suggest a change if you had some other idea.

packages/itwinui-react/src/core/ComboBox/ComboBox.test.tsx
rohan-kadkol: Yes, handled in https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605315949.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: Since you're reading the previous state during render now, you should store it in a state rather than ref. Refs are not supposed to be read or written during render. https://react.dev/learn/referencing-values-with-refs#differences-between-refs-and-state

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: Sounds good. So, I will leave the `setLiveRegionSelection` where it is currently in the `onClickHandler` since users should always immediately update the controlled value whenever the `onChangeHandler` is called.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: Upon testing after the changes from https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605315949, I noticed many unexpected issues.

For example, in the below screen recording, if we select "AAA" and then change the `options` every 5 seconds:
* When the menu is open, the new options are not displayed automatically all the times. Manually re-clicking the ComboBox (something which I do in the second half of the recording) shows the latest `options`.
* Sometimes when the `options` changesm unselected options are shown as selected. I guess this is because 	`optionsExtraInfoRef` is not updating properly and so is returning incorrect/stale indexes for multiple options. So, doing `selected = _originalIndex === selectedIndices` returns `true` for multiple options.

<details>

https://github.com/iTwin/iTwinUI/assets/45748283/bf562df3-478d-47e1-a6af-b378f5e639b2

</details>

Apart from these visual issues, I also came across some *runtime* errors in the browser that weren't highlighted in VSCode during *development* time. So, this could lead to unknown regressions if we don't test it extensively.

Most of these errors were "accessing variable before initialization" errors as a result of calling functions in the render function and not in the `useEffect`.

<details>
<summary>Some of the errors</summary>

image

</details>

For now, I re-added back the `useLayoutEffect` since it works well and doesn't cause any issues (https://github.com/iTwin/iTwinUI/pull/2048/commits/7abdfd9a3d8f24a0cb94a35552e316c669cd173d). Also, from what I understand, `useLayoutEffect` updates the state and blocks browser painting until that state is updated ([Docs](https://react.dev/reference/react/useLayoutEffect#examples)). So, the user will not see any stale state. So, I think `useLayoutEffect` will be okay here.

<details>
<summary>Docs</summary>

image

</details>

I tried some debugging with the `useEffect` and the `useLayoutEffect` removed (https://github.com/iTwin/iTwinUI/pull/2048/commits/cbf36bbb3d0d2b53535e5326b29dd0e39fd3573e). But since it's not related to what this PR is trying to solve, we can continue this later, if we indeed want to remove the `useLayoutEffect`s, since the docs says to avoid `useLayoutEffect` whenever possible.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -412,33 +419,42 @@ export const ComboBox = React.forwardRef(
...
+          setValue(newArray);
           onChangeHandler(__originalIndex, actionType, newArray);
 
           // update live region
           setLiveRegionSelection(
             newArray
-              .map((item) => optionsRef.current[item]?.label)
+              .map(
+                (item) =>
+                  optionsRef.current.find((option) => option.value === item)
```
mayank99, line 439: We should undo all the changes regarding `value` vs `index`. It's not related to the controlled state bug, and it can degrade performance. I believe using `index` was a deliberate design decision to avoid having to repeatedly call `find`/`findIndex`.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: Did you try the previous state pattern mentioned [above](https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605326112)? The issues you were facing might be because you were using ref instead of state.

.changeset/happy-mirrors-stare.md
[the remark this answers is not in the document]
mayank99: Bumping [this comment](https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605198373).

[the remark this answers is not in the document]
rohan-kadkol: > `value={null}` already has its own changeset

To confirm, which changeset are you referring to? Because this PR has mentioned `value={null}` only in this changeset. For now, I just rephrased the changeset. (https://github.com/iTwin/iTwinUI/pull/2048/commits/9d2fd03149653ec8fb8c406661713a8ea5f5a1e1).

> Instead of the changeset, it might be useful to mention the explicit `value={undefined}` behavior in the jsdocs.

Explained this in the JSDocs (same commit).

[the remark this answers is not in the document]
mayank99: Hmm yeah I'm not exactly sure what I was referring to either. 😅 

Just to confirm, are `value={null}` and `value={undefined}` behaving identically now? That's not what I expected. `undefined` means "absence of value" and `null` means "intentionally empty value". So `value={undefined}` should flip to uncontrolled, but `value={null}` should stay controlled.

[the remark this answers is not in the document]
rohan-kadkol: > Just to confirm, are `value={null}` and `value={undefined}` behaving identically now? That's not what I expected. `undefined` means "absence of value" and `null` means "intentionally empty value". So `value={undefined}` should flip to uncontrolled, but `value={null}` should stay controlled.

Yes, that is how it works currently. I added some screen recordings of it from https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1605156984 to the PR description for better visibility.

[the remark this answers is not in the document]
mayank99: Hmm okay I think I understand now. The behavior is correct, but the confusion was coming from the way the changeset is phrased.

Here's a suggestion where I rewrote the last point to make it more clear:

> Fixed a bug in `ComboBox` where the controlled state (`value` prop) was not given priority over the uncontrolled state.
> 
> As a result:
> * Setting the default value using `value={myDefaultValue}` will no longer work. Instead, use the new `defaultValue` prop.
> * Resetting the value using `value={null}` will now force the ComboBox to remain in *controlled* mode. If you want to reset the value and also switch to *uncontrolled* mode, then use `value={undefined}` instead.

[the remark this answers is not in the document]
rohan-kadkol: Ohh, sounds good, changed to your suggestion with just a small change (https://github.com/iTwin/iTwinUI/pull/2048/commits/d1671a0e1ece526378468cc7415bd39a12f2a626).

```diff
- * Resetting the value using `value={null}` will now force the ComboBox to remain in _controlled_ mode. If you want to reset the value and also switch to _uncontrolled_ mode, then use `value={undefined}` instead.
+ * Resetting the value using `value={null}` will now force the ComboBox to switch to/remain in *controlled* mode. If you want to reset the value and also switch to/remain in *uncontrolled* mode, then use `value={undefined}` instead.
```

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: Ohh, I see. Reverted back to using index over value (https://github.com/iTwin/iTwinUI/pull/2048/files/ff7b08b79d001463af7c3f636bac854ca8d234fa..e9b231507291fa11e7b5f305a2bca08e831c1c5e). Since this was a big change, also updated the PR description significantly.

rohan-kadkol:
https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1621543479 makes big changes to the PR by reverting back to using indices over values as the source of truth for the selected options. Updated the PR description accordingly.

packages/itwinui-react/src/utils/hooks/useControlledState.ts
```diff
@@ -4,6 +4,25 @@
...
+ * const onClickHandler = (clickedValue) => {
+ *   const newState = getControlledState(props.value, clickedValue);
+ *   setState(newState);
+ * };
+ */
+export const getControlledState = <T>(
+  controlledState: T,
+  uncontrolledState: T,
+) => {
+  return controlledState !== undefined ? controlledState : uncontrolledState;
```
mayank99, line 23: Why the abstraction?

This is straightforward and immediately super clear:

```js
const newState = props.value !== undefined ? props.value : clickedValue;
```

while this is the same amount of code but now with added indirection:

```js
import { getControlledState } from "../../utils.js";

const newState = getControlledState(props.value, clickedValue)
```

rohan-kadkol: I felt that if we change the logic like:

```diff
- const newState = controlledState !== undefined ? controlledState : uncontrolledState;
+ const newState = controlledState != undefined ? controlledState : uncontrolledState;
```

we can do it in just one place. Additionally, I felt that writing it inline might be error prone. E.g. if we accidentally do `!= undefined`/`!= null` instead of `!== undefined`.

---

Maybe this function can have a better name? i.e. Something that signifies "returns the first non-undefined arg from ...args"?

mayank99: The real problem is that you're sidestepping the `useControlledState` hook that does the heavylifting.

I'm trying to review your recent changes from https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1621543479 and I'm not really sure why you brought back the reducer. I was just asking for the `value` vs `index` parts to be changed.

rohan-kadkol: > I'm trying to review your recent changes from [#2048 (comment)](https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1621543479) and I'm not really sure why you brought back the reducer. I was just asking for the `value` vs `index` parts to be changed.

Ohh, I see. When I started working on the PR, I think I faced issues with using `useControlledState` with indexes. That is why I had gone with using values instead of indexes. So, when I was working on moving from values back to indexes, I thought to again use the reducer instead of `useControlledState`.

To confirm, since I have now fixed the bug while using the reducer, is that alright? Or I can try going back to using `useControlledState`, but this time with indexes. If so, then I could do this in this PR or maybe another one to avoid delay to the bug fix.

> The real problem is that you're sidestepping the `useControlledState` hook that does the heavylifting.

So, are you suggesting keeping it inline? I'm not sure that's the best approach as I feel it's error prone and can lead to regressions in the future if our tests don't catch the accidental changes in the future in the inline if-statement.

Changed it to inline for now (https://github.com/iTwin/iTwinUI/pull/2048/commits/d8ae400d20c899e147b3e158657d6bc6bfdd1c32) until we come with a better way (if we do so).

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
rohan-kadkol: I tried using the "previous state" pattern but still couldn't get it to work without it giving any errors. My trial code is in [`8eb779e`](https://github.com/iTwin/iTwinUI/pull/2048/commits/8eb779e8f0858ab90298ad70dd2f651ac60ab5de). Here is what my trial includes:
* Move the `onOptionsChange()` and `onValuePropChange()` calls at the bottom of the render function to prevent the "variable used before initialization" runtime error. Doing this worked and didn't give any such runtime errors anymore, at least in the little that I tested it.
* Created a `#region` with explanatory comments to always keep the `onOptionsChange()` and `onValuePropChange()` calls at the bottom of the render function and not to move it elsewhere accidentally.
* I tested the case where the `options` passed to the ComboBox changes such that some of the options in the new `options` were not there in the old `options`. I constantly kept getting an error because `optionsExtraInfoRef` was updating too soon. i.e. `getMenuItem` was trying to render the old `options` but `optionsExtraInfoRef` now represented the new options.
  * To solve this error, one of the things I tried was moving around the logic that populates `optionsExtraInfoRef`. This didn't work.
  * To reproduce this error, you can `git checkout` the commit, open the vite playground, and then wait for a few seconds when the `options` changes automatically in `App.tsx`.

<details>
<summary>Error screenshot</summary>

image

</details>

To solve the above-mentioned error when new `options` is passed, I thought of changing `optionsExtraInfoRef` from a ref to a state (https://github.com/iTwin/iTwinUI/pull/2048/commits/e18672648d1caf948a849e245775730319eda569). However, this too gave the same error. This commit also has the same repro steps as the ones mentioned above.

packages/itwinui-react/src/utils/hooks/useControlledState.ts
mayank99: > To confirm, since I have now fixed the bug while using the reducer, is that alright? Or I can try going back to using `useControlledState`, but this time with indexes. If so, then I could do this in this PR or maybe another one to avoid delay to the bug fix.

I think the code has become harder to follow as a result of extending the reducer. I'm not sure if it would be simpler with `useControlledState` but it's worth trying, just to evaluate. Let me know if this takes too long or if you need help with this.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -197,65 +202,68 @@ export const ComboBox = React.forwardRef(
...
-    ) {
-      options.forEach((option, index) => {
-        optionsExtraInfoRef.current[getOptionId(option, id)] = {
-          __originalIndex: index,
-        };
-      });
-    }
+    const [selectedIndexes, setSelectedIndexes] = useControlledState(
+      getSelectedIndexes(defaultValue),
+      valueProp !== undefined ? getSelectedIndexes(valueProp) : undefined,
```
mayank99, line 235: Why not just `getSelectedIndexes(valueProp)`?

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -536,67 +555,96 @@ export const ComboBox = React.forwardRef(
...
     );
   },
 ) as <T>(
   props: ComboBoxProps<T> & { ref?: React.ForwardedRef<HTMLElement> },
 ) => JSX.Element;
+
+// ----------------------------------------------------------------------------
+
+/** Returns either `option.id` or derives a stable id using `idPrefix` and `option.label` (without whitespace) */
+const getOptionId = (option: SelectOption<unknown>, idPrefix: string) => {
```
mayank99, line 633: why is this randomly moved to the bottom now? move it back up to reduce the diff

packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -127,12 +127,16 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
           }
 
           if (focusedIndexRef.current === -1) {
-            return dispatch({
-              type: 'focus',
-              value:
-                Object.values(optionsExtraInfoRef.current)?.[length - 1]
-                  .__originalIndex ?? -1,
-            });
+            // return dispatch({
```
mayank99, line 130: leftover comments need to be cleaned up

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -536,67 +555,96 @@ export const ComboBox = React.forwardRef(
...
+  return option.id ?? `${idPrefix}-option-${option.label.replace(/\s/g, '-')}`;
+};
+
+const getOptionsExtraInfo = <T,>(
```
mayank99, line 637: This one could be made part of the component.

Instead of useState + this helper, you can directly use useMemo.

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -536,67 +555,96 @@ export const ComboBox = React.forwardRef(
...
+                {...inputProps}
+                onChange={handleOnInput}
+                selectTags={
+                  isMultipleEnabled(selectedIndexes, multiple)
+                    ? (selectedIndexes
+                        ?.map((index) => {
+                          const item: SelectOption<T> | undefined =
+                            optionsRef.current[index];
+
+                          if (item == null) {
```
mayank99, line 594: Why would any of the items be null?

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -264,43 +272,66 @@ export const ComboBox = React.forwardRef(
...
+            selectedIndexes >= 0
+              ? optionsRef.current[selectedIndexes]?.label ?? ''
               : '',
           );
         }
       }
-    }, [isOpen, multiple, optionsRef, selected]);
+    }, [isOpen, multiple, optionsRef, selectedIndexes]);
 
     // Update filtered options to the latest value options according to input value
```
mayank99, line 295: this comment looks outdated now

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: The problem might be that we were using `optionsRef` during rendering (discouraged in [react docs](https://react.dev/learn/referencing-values-with-refs#differences-between-refs-and-state) probably to avoid this exact issue).

Can we directly use `options[index]` instead?

packages/itwinui-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -173,13 +164,14 @@ export const ComboBoxInput = React.forwardRef((props, forwardedRef) => {
       }
     },
     [
-      dispatch,
+      // dispatch,
```
mayank99, line 167: one more accidentally leftover

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -438,16 +472,16 @@ export const ComboBox = React.forwardRef(
...
         hide,
+        setSelectedIndexes,
       ],
     );
 
     const getMenuItem = React.useCallback(
       (option: SelectOption<T>, filteredIndex?: number) => {
         const optionId = getOptionId(option, id);
-        const { __originalIndex } = optionsExtraInfoRef.current[optionId];
+        const { __originalIndex } = optionsExtraInfo[optionId];
```
mayank99, line 484: unit tests failing here

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -196,66 +206,82 @@ export const ComboBox = React.forwardRef(
...
+      [multiple, options],
     );
 
+    const [selectedIndexes, setSelectedIndexes] = useControlledState(
+      getSelectedIndexes(defaultValue) ?? (multiple ? [] : -1),
+      getSelectedIndexes(valueProp),
+    );
+
+    const previousValue = React.useRef(valueProp);
+    React.useLayoutEffect(() => {
```
mayank99, line 257: https://github.com/iTwin/iTwinUI/blob/2d415899691ab14f869fa68aef68923120ffb470/STYLEGUIDE.md?#L162

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: Related: https://github.com/iTwin/iTwinUI/discussions/2070

mayank99: In https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1630182367, you said:

> I believe that updating `optionsExtraInfo` directly inside the render function or a `useMemo` is a problem as that immediately updates `optionsExtraInfo` based on the _new_ options even though the `filteredOptions` being used in the return part of the render function is still referring to the _old_ options. This is why when `getMenuItem` tries to search for _old_ options in `optionsExtraInfo` during the render, no match is found.

…which makes me wonder: Shouldn't we fix `filteredOptions` to update eagerly instead of adding a workaround for it (i.e. more stale state)?

I can understand that fixing `filteredOptions` could be considered out-of-scope for this PR, but then it begs the question: why did we even introduce `optionsExtraInfo` in the first place? Was there a problem with keeping it as `optionsExtraInfoRef`?

mayank99: > Yes, with a ref, it updates it immediately in the render function itself instead of after the first render.

This does not explain why it's a problem now and wasn't a problem before.
 
> > Shouldn't we fix `filteredOptions` to update eagerly instead of adding a workaround for it (i.e. more stale state)?
> 
> I think I had tried doing this by calling `setFilteredOptions` directly inside the render function instead of in a `useEffect`/`useLayoutEffect`. 

No, that's still indirect/stale state. To do it eagerly, you can use `useMemo` (same as my `optionsExtraInfo` suggestion).

```jsx
const filteredOptions = React.useMemo(() => {
  return filterFunction(options, value);
}, [options, inputValue, filterFunction]);
```

mayank99: This was fixed during pairing session where we moved both `optionsExtraInfo` and `filteredOptions` into `useMemo`s that re-compute whenever `options` change [`b768fc5`](https://github.com/iTwin/iTwinUI/pull/2048/commits/b768fc5ca8b3d42a49986cc9a93915e7562d7936)

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
mayank99: Fixed after https://github.com/iTwin/iTwinUI/pull/2048#discussion_r1631566750

mayank99 (approved):
LGTM

Ben-Pusey-Bentley (approved):
Tested everything out, and it looks good to me. I do see that your CI is failing, but it looks like it is due to one of the flaky `Table` React visual tests, so it should pass on re-run.

PR #2049, opened 2024-05-14
increase contrast of disabled borders

mayank99:
## Changes

Adjusted the value of `--iui-color-border-disabled` to make it contrast more against the background. This makes disabled controls more perceivable (previously it was hard to even tell that a control is present). Text, background, and icon colors are unaffected and continue to serve as the primary way to identify disabled controls.

I deliberately chose `fg-` colors instead of `bg-`, because they contrast more. At the same time I made sure that `--iui-color-border-disabled` contrasts less than `--iui-color-border-foreground` (which is the strongest border we currently have).

## Testing

Updated all failing visual test images in both `css-workshop` and `react-workshop`. See diff to understand how/what changed visually.

Deploy preview: https://itwin.github.io/iTwinUI/2049/ (e.g. [`button.html`](https://itwin.github.io/iTwinUI/2049/css/button))

## Docs

Added changesets for both `-variables` and `-react` packages.


.changeset/fifty-rats-brake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/itwinui-variables": patch
```
Ben-Pusey-Bentley, line 2: Should the `-css` package also get a changeset?

mayank99: Nope, this doesn't change anything in `-css`.

`-react` needs one because it vendors `-variables`. (See first point under [FAQ](https://github.com/iTwin/iTwinUI/wiki/FAQ))

Ben-Pusey-Bentley (approved):

rohan-kadkol (approved):

FlyersPh9 (approved):

PR #2050, opened 2024-05-14
add `await`s to e2e tests to reduce flakiness

mayank99:
## Changes

Self-explanatory title. The lack of `await` makes tests randomly fail in CI.

See https://playwright.dev/docs/test-assertions

## Testing

All tests continue to pass, hopefully more predictably now.

## Docs

N/A

### DIFF
--- a/testing/e2e/app/routes/DropdownMenu/spec.ts
+++ b/testing/e2e/app/routes/DropdownMenu/spec.ts
@@ -7,29 +7,29 @@ test.describe('DropdownMenu', () => {
     const trigger = page.getByTestId('trigger');
     await trigger.click();
 
-    expect(page.getByTestId('Item 1_1')).toBeFocused();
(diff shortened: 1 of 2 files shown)


testing/e2e/app/routes/DropdownMenu/spec.ts
rohan-kadkol, line 1: I don't think the DropdownMenu tests were flaky on my end. Just curious, we're they flaky on your end?

But yes, having the `await`s is alright too.

mayank99: They've always been flaky. See CI run from `main` branch, where it fails before it passes: https://github.com/iTwin/iTwinUI/actions/runs/9084616075/job/24965994888#step:7:132

rohan-kadkol (approved):

PR #2053, opened 2024-05-15
New `oklch` non-gray colors

FlyersPh9:
> [!IMPORTANT]  
> **Use this draft PR only to [review the new colors](https://itwin.github.io/iTwinUI/2053/css/color) visually. Feel free to branch and tweak. Do not bother reviewing the code at this moment.**

## Changes

- Use `oklch` to create more visually consistent blues, greens, oranges, reds.  [Article on `oklch`](https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl).
  - Blue, green, orange are more vibrant.
  - Red is less vibrant.
- Add an accent hue slider in the settings corner of CSS demo pages. _**By default shows as empty until slider is adjusted, needs to be fixed.**_

## Testing

- Confirmed colors passing.

## Docs

n/a


packages/itwinui-variables/src/themes/dark.scss
rohan-kadkol, line 1: I'm curious about how we would expose the standardized `l` and `c` for users to use. Would we offer new css variables?

Also, I'm wondering for what use cases would users need to create new colors. Because I thought that our variables are sufficient for most standard use cases?

rohan-kadkol (commented):
Took a brief look at some css demos pages. Even the slight changes in the colors give such a crisper and refreshing feel imo! The `informational` color especially seems like a big improvement.

Talking about the format, it's nice that this can be a precursor into standardizing the `l` and the `c` to be able to generate more colors. It's also good to know that the browser support for [`oklch`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch) is good, at least for the basic uses of `oklch` ([Caniuse](https://caniuse.com/?search=oklch)).

Once you approve the failing images, it will be easier to understand all the visual changes related to this PR.

packages/itwinui-variables/src/themes/dark.scss
FlyersPh9: I'm thinking we could use `oklch` in the next major release of iTwinUI-Variables, but maybe we could convert the `oklch` to `hsl` and release the new colors as a minor version.

packages/itwinui-variables/src/themes/light.scss
mayank99, line 1: While this is a fun exploration, I want to emphasize that `oklch` is a fairly new feature to rely on today. Colors form the basis of the design system and therefore need very good browser support.

Additionally, I'm hesitant about changing the very widely used and identity-defining colors (such as `-accent`); these colors should be kept stable. Our users often have multiple versions of iTwinUI on the page, which could result in two blue buttons looking very different (in other words, contributing to the very problem that a design system is supposed to solve). For less prominent colors, I'm ok with small adjustments (`-negative` looks much better now, for example).

packages/itwinui-variables/src/themes/light.scss
```diff
@@ -121,48 +123,49 @@ $opacities: (
...
-  --iui-color-informational-hsl: #{get-hsl($fg-informational-2)};
-  --iui-color-positive-hsl: #{get-hsl($fg-positive-2)};
-  --iui-color-warning-hsl: #{get-hsl($fg-warning-2)};
-  --iui-color-negative-hsl: #{get-hsl($fg-negative-2)};
+  // TODO: These need to be fixed
+  // --iui-color-accent-hsl: var(--iui-color-informational-hsl);
+  // --iui-color-informational-hsl: #{get-hsl($fg-informational-2)};
+  // --iui-color-positive-hsl: #{get-hsl($fg-positive-2)};
+  // --iui-color-warning-hsl: #{get-hsl($fg-warning-2)};
+  // --iui-color-negative-hsl: #{get-hsl($fg-negative-2)};
```
mayank99, line 168: The `-hsl` variables could probably be deprecated if we decide to move to `oklch`.

There's also [`color-mix`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix), which can completely replace these variables. For example:

```diff
  background-color:
-   hsl(var(--iui-color-accent-hsl) / var(--iui-opacity-5));
+   color-mix(in srgb, transparent,
+     var(--iui-color-background-accent) calc(var(--iui-opacity-5) * 100%));
```

packages/itwinui-variables/src/themes/light.scss
FlyersPh9: How about a separate PR using `hsl` that changes:

- Background warning & negative.
- Foreground informational, positive, warning, negative.

Since foreground is used by thin elements (text, borders) it might be less jarring.

I've got a branch with background warning & negative ready to go, but the changes in dark theme foreground colors are so nice… I want it. 🥹

FlyersPh9: #2057, foreground changes will be branched from that PR.

PR #2054, opened 2024-05-15
Fix unnecessary `row-gap` from inline input grid

### LINKED ISSUE #2051: Inline `LabeledInput` has an unnecessary row gap

`LabeledInput` and `LabeledSelect` with `displayStyle="inline"` have a row gap at the bottom (visible when inspecting in [docs](https://itwinui.bentley.com/docs/input#inline)). I suppose it is there because of a message that can be displayed below the input but the gap is also there when there is no message. This makes the input off-center.

### Expected Behavior

Inline `LabeledInput` and `LabeledSelect` should not have a row gap at the bottom if there is no message. It should be just like non-inline `LabeledInput` and `LabeledSelect` which do not have a gap at the bottom when message does not exist.

FlyersPh9:
## Changes

- Removes the unnecessary `row-gap` from inline input grid if only 1 row is needed by checking if `.iui-status-message` is present. Closes #2051.

## Testing

- Ran visual tests in React & CSS.

## Docs

n/a


packages/itwinui-css/src/input-container/input-container.scss
```diff
@@ -13,6 +13,10 @@
     grid-template-rows: 1fr auto;
     grid-template-columns: auto 1fr;
 
+    &:not(:has(.iui-status-message)) {
+      row-gap: unset;
+    }
```
rohan-kadkol, line 18: Instead of removing the `row-gap`, would it be better to modify the `grid-template-rows` appropriately? Because I think the real issue is that we're telling it that there are two rows even when there's just one row.

.changeset/breezy-otters-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-react": patch
+---
+
+Remove row gap from inline input grid when no secondary line is present.
```
rohan-kadkol, line 5: I think we should also mention at least a few components that are affected. E.g.:

```suggestion
Remove row gap from the inline `<InputGrid>` when no secondary line (`StatusMessage`) is present. Some other affected components: `LabeledInput`, `LabeledSelect`, `LabeledTextarea`.
```

Ben-Pusey-Bentley (approved):

packages/itwinui-css/src/input-container/input-container.scss
FlyersPh9: Updated to use `grid-template-rows` instead of `row-gap`.

.changeset/breezy-otters-rule.md
FlyersPh9: Updated both changesets.  Wasn't sure if I should be using React component names in the CSS changeset or not, did for now.

rohan-kadkol: The react changeset LGTM. For the css changeset, I think classnames would be better. So, something like:

```md
Remove unnecessary gap below the `.iui-input-grid` when `data-iui-label-placement='inline'` and no secondary line (`iui-status-message`) is present.
```

FlyersPh9: @mayank99 said to not use class names within changesets: https://github.com/iTwin/iTwinUI/pull/2043#discussion_r1599072350

FlyersPh9: Change made.

.changeset/poor-turtles-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/itwinui-css": patch
+---
+
+Remove unnecessary gap below `.iui-input-grid` when `data-iui-label-placement='inline'` and no secondary line (`iui-status-message`) is present.
```
rohan-kadkol, line 5: nit/optional: After re-reading this, I felt my suggestion could lead to two interpretations. So, I think we can make it clearer (optional)

```suggestion
Remove unnecessary gap below `.iui-input-grid` when `data-iui-label-placement='inline'` and when no secondary line (`iui-status-message`) is present.
```

rohan-kadkol (approved):

PR #2055, opened 2024-05-16
`Breadcrumbs.Item` now gets anchor styling even when passing `as={…}`

rohan-kadkol:
## Changes

I noticed an issue with the `Breadcrumbs.Item` as links. When adding `as={…}`, the `Anchor` styling (`.iui-anchor`) seems to be missing. Small NextJS [repro](https://stackblitz.com/edit/nextjs-xjqawb?file=app%2Fpage.tsx) using `as={Link}`.

image

Additionally, @mayank99 noticed that spans are also getting button hover styles.

image

These issues stem from https://github.com/iTwin/iTwinUI/pull/1919#discussion_r1576799908. This PR tries to fix these issues.

---

This PR fixes these issues by improving logic in `Breadcrumbs.tsx` to render a:
* `Box as={"span"}` if `as="span"` or `props.href == null && props.onClick == null && asProp == null`
* `Button`
  * `as={Anchor}` when `asProp={"a"}` or `asProp=null && !!props.href`
  * `as={asProp}` in all other cases

It also makes some changes in `breadcrumbs.scss` to properly style the `as={Link}` case.

As a result, this PR lets `Button`, `Anchor`, and `Link` get the button hover background but prevents `span` from getting it.

Upon testing the breadcrumbs items in v3.9.1 (the version before v3.10.0 where the regression started from #1919), I also realized that doing `as={Link}` was not showing the anchor styling since `.iui-anchor` was missing (screen recordings and playgrounds in the testing section below). This PR fixes that issue too.

Additionally, this PR gives anchor styling to a link even when it does not have `href`. E.g. [`Link`](https://remix.run/docs/en/main/components/link) from RemixJS uses `to` instead of `href`. (https://github.com/iTwin/iTwinUI/pull/2055#discussion_r1608325179)

## Testing

* Added a unit test to confirm that breadcrumbs item links get anchor styling even when doing `as={Link}`
* Testing in the Remix (using `<Link to={} />`) and NextJS (using `<Link href={} />`) playgrounds:
  * v3.11.0 (version before this PR): [NextJS](https://stackblitz.com/edit/nextjs-yjodea?file=app%2Fpage.tsx), [RemixJS](https://stackblitz.com/edit/remix-run-remix-6nbtbt?file=README.md)
  * v3.9.1 (version before the regression started from #1919): [NextJS](https://stackblitz.com/edit/nextjs-7htile?file=app%2Fpage.tsx), [RemixJS](https://stackblitz.com/edit/remix-run-remix-zfsuvc?file=package.json)
  * this PR: [NextJS](https://gist.github.com/rohan-kadkol/ab300ec0772c31c8a46e912ecdb151e0), [RemixJS](https://gist.github.com/rohan-kadkol/139187b0ce222f99536284521e7991ed)

<details>
<summary>Screen recordings from NextJS. RemixJS results are identical.</summary>

#### v3.11.0 (version before this PR):

https://github.com/iTwin/iTwinUI/assets/45748283/d052e70e-bacd-479a-b341-3ac17b37cb44

#### v3.9.1 (version before the regression started from #1919):

https://github.com/iTwin/iTwinUI/assets/45748283/01eece1e-0c5f-4684-89ca-987e7be60edf

#### this PR

https://github.com/iTwin/iTwinUI/assets/45748283/a45c79bc-1a02-416b-922e-2e6f28473c47

</details>

## Docs

- Added changesets


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -230,18 +230,33 @@ const Separator = ({ separator }: Pick<BreadcrumbsProps, 'separator'>) => (
...
+        styleType='borderless'
+        {...(commonProps as Partial<React.ComponentProps<typeof Button>>)}
+      >
+        {children}
+      </Button>
+    );
+  }
+
+  if (resolvedAs === 'a') {
+    return <Anchor {...commonProps}>{children}</Anchor>;
```
mayank99, line 253: This will not work if the custom `Link` component does not have a `href` prop (which is not an unusual scenario; I've seen `<Link to>`).

mayank99: There are two parts to this issue:

1. `<Breadcrumbs.Item as={Link}>` renders `<Link class="iui-button iui-breadcrumbs-content">` which does not have anchor styling
2. `<Breadcrumbs.Item as="span">` renders `<span class="iui-button iui-breadcrumbs-content">` which has hover styling.

It makes sense to change the rendered element (from `Button` to `Box`) for fixing 2.

```jsx
defaultAs === 'span' ? <Box /> : <Button />
```


But for 1, we could leave the current JS and instead fix the CSS. You can try something like this diff:

<details><summary>breadcrumbs.scss</summary>

```diff
@@ -43,11 +43,15 @@
     text-overflow: ellipsis;
   }
 
-  &:where(a):focus {
+  &:where(:any-link) {
     outline-offset: -1px;
+
+    &:hover {
+      text-decoration: underline;
+    }
   }
 
-  &:where(button) {
+  &:where(button, :any-link) {
     @include base.iui-button;
     @include borderless.iui-button-borderless;
     color: var(--iui-color-text-accent);
```

</details> 

(The `:any-link` selector is basically `a` but actually ensures that `href` is present.)

mayank99: While waiting for your response I took a look at your current code and thought it could be simplified like this:

```tsx
  const { as: asProp, ...rest } = props;

  const commonProps = {
    ...rest,
    className: cx('iui-breadcrumbs-content', props.className),
    ref: forwardedRef,
  };

  if (String(asProp) === 'span') {
    return <Box as='span' {...commonProps} />;
  }

  return (
    <Button
      as={(asProp === 'a' ? Anchor : asProp) as any}
      styleType='borderless'
      {...commonProps}
    />
  );
```

.changeset/soft-rice-visit.md
mayank99, line 1: Is this changeset accurate now?

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
rohan-kadkol: Ohh, I think this would work if we ask users to explicitly use the `asProp` and not rely on us choose what to render based on the `href`/`onClick`. To confirm, are we hoping to make that change?

When testing in the Remix playground after making the suggested change to `Breadcrumbs.tsx`, all items were rendered in HTML as `<button>`s, except the third item that used the `asProp`:

<details>
<summary>Remix playground</summary>


https://github.com/iTwin/iTwinUI/assets/45748283/2f3c49cc-b8e3-429a-8381-08742f23fec3

```jsx
import type { MetaFunction } from '@remix-run/node';
import { Breadcrumbs } from '@itwin/itwinui-react';
import { Link } from '@remix-run/react';

export const meta: MetaFunction = () => {
  return [
    { title: 'Remix Playground' },
    {
      name: 'description',
      content: 'Playground for testing iTwinUI components in Remix.',
    },
  ];
};

export default function Index() {
  return (
    <>
      <Breadcrumbs>
        <Breadcrumbs.Item key={0} onClick={() => console.log('Button clicked')}>
          button
        </Breadcrumbs.Item>
        <Breadcrumbs.Item key={1} href='/?path=page-refresh'>
          link (as=undefined)
        </Breadcrumbs.Item>
        <Breadcrumbs.Item
          as={Link}
          className='my-special-link'
          key={2}
          to='/?path=no-page-refresh'
        >
          link (as=Link)
        </Breadcrumbs.Item>
        <Breadcrumbs.Item key={3}>span 1</Breadcrumbs.Item>
        <Breadcrumbs.Item key={4}>span 2</Breadcrumbs.Item>
      </Breadcrumbs>
    </>
  );
}

```

</details>

rohan-kadkol: I was in progress with testing and fixing the unit tests and so didn't reply after making the relevant commits yesterday. But I believe the latest code work now.

> This will not work if the custom `Link` component does not have a `href` prop (which is not an unusual scenario; I've seen `<Link to>`).

Made changes to render `<Button as={asProp} />` whenever an `asProp` is passed. I also updated the CSS styles like you suggested to add anchor type styling even when rendering using `<Button as={Link} />`.

> There are two parts to this issue:
> 
> 1. `<Breadcrumbs.Item as={Link}>` renders `<Link class="iui-button iui-breadcrumbs-content">` which does not have anchor styling
> 2. `<Breadcrumbs.Item as="span">` renders `<span class="iui-button iui-breadcrumbs-content">` which has hover styling.

I modified the code `Breadcrumbs.Item` code to render a:
* `Button`
  * `as={undefined}` when `onClick` is passed
  * `as={Anchor}` when `href` is passed
  * `as={asProp}` when `as: asProp` is passed
* `Box as={"span"}` in all other cases

As a result:
* even links have the button hover styling, as it is in the current version (v3.11.0). However, with this PR, the anchor styling is properly added when using `as={Link}` since the css has been modified to target `:any-link`.
* `span`s no longer get the button hover styling.

I tested all this logic in NextJS (using `<Link href={} />`) and RemixJS (`<Link to={} />`):
* v3.10.1 (version before this PR): [NextJS](https://stackblitz.com/edit/nextjs-yjodea?file=app%2Fpage.tsx), [RemixJS](https://stackblitz.com/edit/remix-run-remix-6nbtbt?file=README.md)
* v3.9.1 (version before the regression started from #1919): [NextJS](https://stackblitz.com/edit/nextjs-7htile?file=app%2Fpage.tsx), [RemixJS](https://stackblitz.com/edit/remix-run-remix-zfsuvc?file=package.json)
* this PR: [NextJS](https://gist.github.com/rohan-kadkol/ab300ec0772c31c8a46e912ecdb151e0), [RemixJS](https://gist.github.com/rohan-kadkol/139187b0ce222f99536284521e7991ed)

---

All these discussed changes are part of https://github.com/iTwin/iTwinUI/pull/2055/files/6402e0d17323d3d9091a83c0e1fd0fa4e5eb1525..1df2df5e134843ac81c42e78ef5b520e4948b808#diff-8e450321e4016954cde3ba7c086d86f6286650a88b49a9bfc57c39feead6155c.

mayank99: > Ohh, I think this would work if we ask users to explicitly use the `asProp` and not rely on us choose what to render based on the `href`/`onClick`. To confirm, are we hoping to make that change?

Whoops, no. We need to keep the auto behavior (though I think in the future we should try to handle it in `Button` or `ButtonBase` or a different utility component, since it's useful to have it in many places).

Updated code with slight adjustment:

```jsx
  const { as: asProp, ...rest } = props;

  const commonProps = {
    ...rest,
    className: cx('iui-breadcrumbs-content', props.className),
    ref: forwardedRef,
  };

  if (String(asProp) === 'span') {
    return <Box as='span' {...commonProps} />;
  }

  return (
    <Button
      as={(asProp === 'a' || !!props.href ? Anchor : asProp) as any}
      styleType='borderless'
      {...commonProps}
    />
  );
BreadcrumbsItem.displayName = 'Breadcrumbs.Item';
```

.changeset/soft-rice-visit.md
rohan-kadkol: Ohh, right, it was outdated. Updated it (https://github.com/iTwin/iTwinUI/pull/2055/commits/3ef76b457cb388a28f022f5759b3cc685fcc76d6)

mayank99 (dismissed):
Pre-approving since behavior is now (mostly) correct and the remaining comments ([1](https://github.com/iTwin/iTwinUI/pull/2055#discussion_r1613722662), [2](https://github.com/iTwin/iTwinUI/pull/2055#discussion_r1613646832), [3](https://github.com/iTwin/iTwinUI/pull/2055#discussion_r1613755369)) are pretty minor.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
rohan-kadkol: Ohh, I see, sounds good. I flipped the order of the returns to try to simplify the if condition (https://github.com/iTwin/iTwinUI/pull/2055/commits/45dd08d53aa1864aa1e99d5bdfc50dd925e91eb9).

To make it work, I had to add some more logic in the first `if` statement. But this is still lesser code than what was there before.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -11,6 +11,7 @@ import * as UseOverflow from '../../utils/hooks/useOverflow.js';
 import { IconButton } from '../Buttons/IconButton.js';
 import { Button } from '../Buttons/Button.js';
 import { userEvent } from '@testing-library/user-event';
+import cx from 'classnames';
```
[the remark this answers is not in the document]
rohan-kadkol: Sounds good, removed `cx` from the unit test file (https://github.com/iTwin/iTwinUI/pull/2055/commits/08d49c6996459f1a1b26627807a86071f35ec4aa).

packages/itwinui-css/src/breadcrumbs/breadcrumbs.scss
```diff
@@ -43,11 +43,16 @@
     text-overflow: ellipsis;
   }
 
-  &:where(a):focus {
+  &:where(:any-link) {
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, I see, made that change (https://github.com/iTwin/iTwinUI/pull/2055/commits/9f62bf2729b5f4f791d8b1d905cfb7b836e27930). Was able to revert the `css-workshop` test image changes too.

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -228,20 +228,27 @@ const Separator = ({ separator }: Pick<BreadcrumbsProps, 'separator'>) => (
...
 
-  const defaultAs = !!props.href ? Anchor : !!props.onClick ? 'button' : 'span';
+  const commonProps = {
+    ...rest,
+    className: cx('iui-breadcrumbs-content', props.className),
+    ref: forwardedRef,
+  };
+
+  if (
+    asProp === 'span' ||
```
mayank99, line 240: In my original suggestion this was wrapped in `String`. Not sure why you removed that but it makes the build fail.

```suggestion
    String(asProp) === 'span' ||
```

Is it possible that your IDE is not surfacing types correctly?

rohan-kadkol: Ohh, not sure why I didn't get that error before (or maybe I missed it somehow), but I am getting it now. Wrapped `asProp` with `String()` (https://github.com/iTwin/iTwinUI/pull/2055/commits/6959296eaf75bd8db79e6b61d1648d3c7740c031).

When re-testing it, I spotted a mistake where if a `as={Link}` and `href="…"` is passed in NextJS, an `Anchor` is rendered. Fixed this error (https://github.com/iTwin/iTwinUI/pull/2055/commits/68d232f50e41c7b14dd45977c2e4e60732712d20).

PR #2058, opened 2024-05-17
add `loading` prop to `Button`

mayank99:
## Changes

This adds the ability to display a progress indicator inside a `Button`, without causing layout shift.

New APIs (bikesheddable):
- **CSS**: Added `data-iui-loading` attribute and `iui-button-spinner` classes
  - Also added `iui-button-label` class to be applied on the existing text label. See [comment](https://github.com/iTwin/iTwinUI/pull/2058#discussion_r1608548036)
- **React**: Added `loading` prop (currently a boolean).

### Future tasks

- This PR currently just disables the button, which is not enough to make it accessible. I was playing around with live regions to announce the "Loading" text, but that will require some larger changes, which can come in a follow-up PR.
- We might want to allow showing the loading spinner and some visible "Loading" text inside the button at the same time. Again, this can come in a follow-up PR.

## Testing

Added css visual tests + react visual tests + react unit tests.

Tested in vite playground using this code:

```jsx
import { Button } from '@itwin/itwinui-react';

  const [isLoading, setIsLoading] = React.useState(false);

  return (
    <>
      <Button
        loading={isLoading}
        onClick={async () => {
          ReactDOM.flushSync(() => setIsLoading(true));

          // simulate async operation
          await new Promise((resolve) => setTimeout(resolve, 2000));

          setIsLoading(false);
        }}
      >
        Click me
      </Button>
    </>
  );
};
```

</details> 

https://github.com/iTwin/iTwinUI/assets/9084735/97994d3c-948e-4b1d-8a59-524be30e2289

Added the above code into `Button.loading` example, which can be imported in playgrounds. (This required installing `examples` as a global workspace.)

## Docs

Added changesets and jsdocs.

Proper docs to come in a follow-up PR, after we have live regions implemented.


packages/itwinui-css/src/button/base.scss
```diff
@@ -32,6 +32,18 @@ $disabledButtonSelectors: '[disabled], :disabled, [aria-disabled="true"], [data-
   &:is(#{$disabledButtonSelectors}) {
     --_iui-button-icon-fill: var(--iui-color-icon-disabled);
   }
+
+  &:is([data-iui-loading='true']) {
+    > :where(span) {
```
mayank99, line 37: This feels a little brittle. I'm thinking we should add a class for the button-label

Ben-Pusey-Bentley: Yeah, adding a `.iui-button-label` class for the `<span>` elements inside of the button makes sense to me.

Ben-Pusey-Bentley (approved):

apps/react-workshop/src/Button.stories.tsx
```diff
@@ -124,3 +124,7 @@ Stretch.decorators = [
     </div>
   ),
 ];
+
+export const Loading = () => {
+  return <Button loading>Click me</Button>;
+};
```
rohan-kadkol, line 130: I think this demo will be better if clicking the button toggles the loading state.

packages/itwinui-react/src/core/Buttons/Button.tsx
```diff
@@ -77,6 +84,8 @@ export const Button = React.forwardRef((props, ref) => {
       className={cx('iui-button', 'iui-field', className)}
       data-iui-variant={styleType !== 'default' ? styleType : undefined}
       data-iui-size={size}
+      data-iui-loading={loading}
```
rohan-kadkol, line 87: Would it be better if it is `{loading ? true : undefined}`? Because `loading` might be controlled by a state variable. And so `false` will be passed instead of `undefined` during the non-loading cases.

apps/css-workshop/backstop/results/bitmaps_reference/iTwinUI_button_Loading_dark_0_demo-loading_0_desktop.png
rohan-kadkol, line 1: The borderless button with the loading state looks like a non-interactive `ProgressRadial` and not an interactive button. To confirm, is that alright?

If yes, maybe we can elaborate on this case in the future docs regarding when to use `loading` on a borderless button and when not to.

apps/css-workshop/src/components/Button.astro
```diff
@@ -35,6 +38,9 @@ const {
   data-iui-size={size}
   data-iui-variant={idea ? 'high-visibility' : variant}
   data-iui-active={isActive ? 'true' : undefined}
+  data-iui-loading={loading ? 'true' : undefined}
+  tabindex='0'
```
rohan-kadkol, line 42: nit: Just curious, was this needed for something?

packages/itwinui-react/src/core/Buttons/Button.tsx
```diff
@@ -108,6 +117,14 @@ export const Button = React.forwardRef((props, ref) => {
           {endIcon}
         </Box>
       )}
+
+      {loading && (
+        <ProgressRadial
+          size='small'
+          className='iui-button-spinner'
+          aria-hidden
```
rohan-kadkol, line 125: > * This PR currently just disables the button, which is not enough to make it accessible. I was playing around with live regions to announce the "Loading" text, but that will require some larger changes, which can come in a follow-up PR.

Would removing `aria-hidden` and relying on the visually hidden "Loading." text in the ProgressRadial not be sufficient?

I guess a using a live region will be able to announce when the loading is complete. Because with the regular visually hidden text, when the loading is complete, VoiceOver just reannounces the button label instead of saying something like "finished loading".

apps/css-workshop/backstop/results/bitmaps_reference/iTwinUI_button_Loading_dark_0_demo-loading_0_desktop.png
mayank99: Yeah I think that's fine. It looks weird in isolation, but usually the spinner is only shown for a brief duration after the button is clicked.

apps/css-workshop/src/components/Button.astro
mayank99: This makes it consistent with React and allows focusing disabled buttons. See https://github.com/iTwin/iTwinUI/pull/1844 and https://github.com/iTwin/iTwinUI/pull/1461

packages/itwinui-react/src/core/Buttons/Button.tsx
mayank99: fixed in [`677e8da`](https://github.com/iTwin/iTwinUI/pull/2058/commits/677e8da93e301c6fd651aaa03f59f466e7110063)

packages/itwinui-react/src/core/Buttons/Button.tsx
mayank99: We cannot rely on text inside the button to be announced at the right time. 

The expected flow should go something like:
1. "click me, button")
2. (user clicks)
3. "processing…" (notice the lack of "click me")
4. (wait N seconds)
5. "done!"

apps/react-workshop/src/Button.stories.tsx
rohan-kadkol: Sounds good. It's just that I wanted to test how the transition between the non-loading and loading states looks like and so I had to manually update this story.

packages/itwinui-css/src/button/base.scss
rohan-kadkol: > This feels a little brittle. I'm thinking we should add a class for the button-label

Yes, sounds good, since you're also adding `.iui-button-spinner`. But that can come in another PR if that was your plan.

rohan-kadkol (approved):

apps/react-workshop/src/Button.stories.tsx
mayank99: I added a `Button.loading` example, not used anywhere currently. [`3937d9b`](https://github.com/iTwin/iTwinUI/pull/2058/commits/3937d9b898dd7bfce10fe59b68e983f86d49c9f8)

You can use it in the playground like this:

```jsx
export { ButtonLoadingExample as default } from 'examples';
```

rohan-kadkol (approved):

packages/itwinui-css/src/button/base.scss
mayank99: added in [`1c87486`](https://github.com/iTwin/iTwinUI/pull/2058/commits/1c87486e095e194280dcc6d18066f47978419c7a)

apps/css-workshop/src/components/Button.astro
```diff
@@ -58,4 +64,13 @@ const {
       </span>
     )
   }
+  {
+    loading && (
+      <div
+        class='iui-progress-indicator-radial iui-button-spinner'
+        data-iui-indeterminate='true'
+        data-iui-size={size === 'small' ? 'x-small' : 'small'}
```
rohan-kadkol, line 72: If the size of the `ProgressRadial` depends on the size of the Button, should this be done in React as well?

mayank99: fixed in [`a67ebd3`](https://github.com/iTwin/iTwinUI/pull/2058/commits/a67ebd341abb9a17f297b25ea3faa0c3fa85211e)

examples/Button.loading.jsx
```diff
@@ -0,0 +1,25 @@
...
+          // simulate async operation
+          await new Promise((resolve) => setTimeout(resolve, 3000));
+
+          setIsLoading(false);
+        }}
+      >
+        Click me
+      </Button>
+    </>
+  );
```
rohan-kadkol, line 24: nit: Is the `flushSync` needed here?

packages/itwinui-react/src/core/Buttons/Button.test.tsx
```diff
@@ -220,3 +221,25 @@ it('should respect `stretched` prop', () => {
   const { container } = render(<Button stretched>Do not click</Button>);
   expect(container.querySelector('button')).toHaveStyle('--_iui-width: 100%');
 });
+
+it.each(['default', 'small', 'large'] as const)(
```
rohan-kadkol, line 225: nit/optional: Maybe this test can also confirm that the label is still present in the DOM.

apps/css-workshop/src/components/Button.astro
mayank99: added unit test and tested manually using https://github.com/iTwin/iTwinUI/commit/b2d0a1f0a43de938e0ab1d507330e4afc5db37c7

image

rohan-kadkol (approved):

examples/Button.loading.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default function App() {
+  const [isLoading, setIsLoading] = React.useState(false);
+
+  return (
+    <>
+      <Button
+        loading={isLoading}
+        onClick={async () => {
+          ReactDOM.flushSync(() => setIsLoading(true));
```
rohan-kadkol, line 13: nit: Is the `flushSync` needed here?

packages/itwinui-react/src/core/Buttons/Button.test.tsx
mayank99: Nope, it's not important here. There will be proper tests for the semantics later.

examples/Button.loading.jsx
mayank99: Yes, we want to show the loading state immediately before performing the action.

mayank99: You can see similar code here: https://github.com/iTwin/iTwinUI-migration-tool/blob/32924f564041761236002166b500f9a3f36b6170/app/(routes)/v3/preview/RunCodemodsInAllFilesButton.tsx#L31-L33

rohan-kadkol: To confirm, do you mean that showing the loading state in the next frame after the action has started has some issue?

mayank99: Yes. When you write code like this, React will batch all state updates, so you might not even see the loading state:

```jsx
setIsLoading(true);
doSomething();
setIsLoading(false);
```

PR #2060, opened 2024-05-22
use `popover` API in `Tooltip` if supported

mayank99:
## Changes

When supported, [`popover=manual`](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using#using_manual_popover_state) will be used to show the tooltip in the [top-layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer). In non-supported browsers, the attribute is ignored.

In addition to the `popover` attribute itself, there are only two changes:
- **JS**: Calling [`togglePopover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/togglePopover) whenever the React state changes.
  - This is guarded using `?.` for supporting older browsers.
  - This also required adding a custom TS type, which can be removed in the future.
- **CSS**: Hiding the tooltip when [`:popover-open`](https://developer.mozilla.org/en-US/docs/Web/CSS/:popover-open) doesn't match.

I kept the existing mechanism (`Portal` component, `z-index` property, `hidden` attribute), so `Tooltip` will continue to work in browsers that don't support popover.

## Testing

All existing tests pass 🚀

Tested manually that hover/focus behaviors are intact and that the tooltip does indeed show in the top layer (This can be verified in Chrome dev tools, for example). Also emulated older browsers by commenting out `popover` and testing that it works fine.

Updated unit tests to simply check for the presence of the `popover` attribute. I spent some time looking into adding tests for `togglePopover` without success, since [JSDOM doesn't support it](https://redirect.github.com/jsdom/jsdom/issues/3721). This is not a big deal imo; the visual tests already serve as a way to ensure that the tooltip is working.

## Docs

Added changesets for both CSS and React.

Updated Tooltip documentation to mention that portaling is useful only in browsers that don't support `popover`.

### DIFF
--- a/.changeset/quick-ligers-serve.md
+++ b/.changeset/quick-ligers-serve.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': minor
+---
+
+Added support for `popover` attribute in `iui-tooltip`.
--- a/apps/website/src/content/docs/tooltip.mdx
+++ b/apps/website/src/content/docs/tooltip.mdx
@@ -37,7 +37,7 @@ There are some advanced props available for more granular control over positioni
 
 ### Portals
 
-It is important to know that before calculating the position, the tooltip gets [portaled](https://react.dev/reference/react-dom/createPortal) into the nearest [`ThemeProvider`](themeprovider) to avoid [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) issues. This behavior can be controlled using the Tooltip's `portal` prop or the ThemeProvider's [`portalContainer`](themeprovider#portals) prop.
+It is important to know that before calculating the position, the tooltip gets [portaled](https://react.dev/reference/react-dom/createPortal) into the nearest [`ThemeProvider`](themeprovider). This is done to avoid [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) issues in browsers where the [`popover` API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) is not supported. This portaling behavior can be controlled using the Tooltip's `portal` prop or the ThemeProvider's [`portalContainer`](themeprovider#portals) prop.
 
 ## Accessibility
 
--- a/packages/itwinui-css/src/tooltip/tooltip.scss
+++ b/packages/itwinui-css/src/tooltip/tooltip.scss
@@ -23,7 +23,7 @@
 
   @include mixins.iui-blur($hsl: 0 0% 0%, $opacity: 3);
 
-  &[hidden] {
+  &:where([hidden], [popover]:not(:popover-open)) {
     display: none !important;
   }
 }
--- a/packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
+++ b/packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
@@ -20,7 +20,7 @@ it('should toggle the visibility of tooltip on hover', () => {
 
   const tooltip = getByText('some text');
   expect(tooltip).not.toBeVisible();
-  expect(tooltip).toHaveAttribute('popover');
+  expect(tooltip).toHaveAttribute('popover', 'manual');
 
   fireEvent.mouseEnter(trigger);
   expect(tooltip).toBeVisible();
@@ -46,7 +46,7 @@ it('should toggle the visibility of tooltip on focus', async () => {
 
   const tooltip = getByText('some text');
   expect(tooltip).not.toBeVisible();
-  expect(tooltip).toHaveAttribute('popover');
+  expect(tooltip).toHaveAttribute('popover', 'manual');
 
   fireEvent.focus(trigger);
   act(() => void vi.advanceTimersByTime(50));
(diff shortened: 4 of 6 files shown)


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -242,13 +266,15 @@ const useTooltip = (options: TooltipOptions = {}) => {
...
-    () =>
-      interactions.getFloatingProps({
+    () => ({
+      ...interactions.getFloatingProps({
         hidden: !open,
         'aria-hidden': 'true',
         ...props,
         id,
       }),
+      popover: 'manual',
```
rohan-kadkol, line 276: Don't think this would matter, but would it be better to put `popover: 'manual'` inside `getFloatingProps`? Or maybe it's outside intentionally to force `popover: 'manual'`?

apps/website/src/content/docs/tooltip.mdx
```diff
@@ -37,7 +37,7 @@ There are some advanced props available for more granular control over positioni
 
 ### Portals
 
-It is important to know that before calculating the position, the tooltip gets [portaled](https://react.dev/reference/react-dom/createPortal) into the nearest [`ThemeProvider`](themeprovider) to avoid [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) issues. This behavior can be controlled using the Tooltip's `portal` prop or the ThemeProvider's [`portalContainer`](themeprovider#portals) prop.
+It is important to know that before calculating the position, the tooltip gets [portaled](https://react.dev/reference/react-dom/createPortal) into the nearest [`ThemeProvider`](themeprovider). This is done to avoid [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) issues in browsers where the [`popover` API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) is not supported. This portaling behavior can be controlled using the Tooltip's `portal` prop or the ThemeProvider's [`portalContainer`](themeprovider#portals) prop.
```
rohan-kadkol, line 40: Should we mention "top-layer" somewhere? Because it may not be clear that the Popover API uses the "top-layer".

I feel that mentioning the "top-layer" might be important for users since that will reassure them that stacking context or overlap issues are less likely now.

packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
```diff
@@ -20,6 +20,7 @@ it('should toggle the visibility of tooltip on hover', () => {
 
   const tooltip = getByText('some text');
   expect(tooltip).not.toBeVisible();
+  expect(tooltip).toHaveAttribute('popover');
```
rohan-kadkol, line 23: Should it also check that the attribute value is `manual`?

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
mayank99: I kept it outside so I don't have to deal with types. (`popover` types are not currently available)

packages/itwinui-css/src/tooltip/tooltip.scss
```diff
@@ -23,7 +23,7 @@
 
   @include mixins.iui-blur($hsl: 0 0% 0%, $opacity: 3);
 
-  &[hidden] {
+  &:where([hidden], [popover]:not(:popover-open)) {
```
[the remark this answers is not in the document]
mayank99: It's more correct and yes serves as an additional safeguard. Maybe in the future we could even remove `hidden`.

apps/website/src/content/docs/tooltip.mdx
mayank99: The link to the "`popover` API" explains top layer too. That's literally its purpose after all.

This section is mainly concerned with describing the portaling behavior.

packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
mayank99: The value is not important here. The `popover` attribute is what really matters. Every value of `popover` puts it in the top layer.

rohan-kadkol: optional: Since there are some differences between `auto` and `manual` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/popover#value)), I thought it might not hurt to confirm that `manual` is selected.

.changeset/quick-ligers-serve.md
rohan-kadkol, line 1: Should we also modify the `<script>` ([code](https://github.com/iTwin/iTwinUI/blob/d7ef90cb4e6ac83b761b454c5bd14dd81208d837/apps/css-workshop/src/pages/tooltip.astro#L90-L100)) in the tooltip HTML demo to use the popover API?

mayank99: No need. The HTML demos are not meant to be perfectly indicative of the behavior.

packages/itwinui-react/src/core/Tooltip/Tooltip.test.tsx
mayank99: Sure, done ([`1ce6774`](https://github.com/iTwin/iTwinUI/pull/2060/commits/1ce67744f52effe5c68739f20977faeb40c6a4ad)). But I'll reemphasize that it's _not_ important, it's almost like testing implementation details. And improving unit tests is a waste of everyone's time in most cases.

rohan-kadkol (approved):

PR #2061, opened 2024-05-22
Adds Support for Dynamic Sizing to `ComboBox` Virtualization

Ben-Pusey-Bentley:
## Changes

Adds support for dynamic sizing to `ComboBox` virtualization. This is achieved through adding the [tanstack virtual](https://tanstack.com/virtual/latest) package, and adding the [measureElement](https://tanstack.com/virtual/latest/docs/api/virtualizer#measureelement-1) ref to a wrapper around the virtualized items. This fixes an issue where the list box would not be sized correctly for the content if virtualization was enabled and some options had a `subLabel` value. 
Before:
image
After:
image

The spacing between options on the virtualized `ComboBox` is also now slightly different, as a result of using the virtualizer's [gap](https://tanstack.com/virtual/latest/docs/api/virtualizer#gap). The react workshop screenshot has been updated accordingly.

- [ ] After PR TODO: Add tanstack virtual to `Tree` and `Table` components in order to remove older virtualization code.

## Testing

Tested with this [vite playground code](https://github.com/iTwin/iTwinUI/blob/137c8d369bdf4df4e0c1bc91f838d5e70dd3d7b6/playgrounds/vite/src/App.tsx). Also ran all automated tests. Replaced the failing unit test for keyboard navigation on virtualized `ComboBox` with an e2e test. Added an e2e test for the checking the dynamic sizing as well. Also updated react workshop test for virtualized `ComboBox` with it's new spacing.

## Docs

Added react changesets for the dynamic sizing support and the added dependency on `@tanstack/virtual-react`.


packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,18 +58,42 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
+      as='div'
+      style={{
+        minBlockSize: virtualizer.getTotalSize(),
+        minInlineSize: '100%',
+      }}
+      {...rest}
+    >
+      <div {...innerProps}>
+        {virtualizer.getVirtualItems().map((virtualItem) => (
+          <div
```
mayank99, line 82: I don't think there should be an extra div needed here. All these props could be easily applied on the items directly.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,18 +58,42 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
-    // whenever there is no elements, to show empty state message
-    itemsLength: filteredOptions.length || 1,
-    itemRenderer: virtualItemRenderer,
-    scrollToIndex: focusedVisibleIndex,
-  });
+  useLayoutEffect(() => {
+    if (focusedVisibleIndex) {
+      virtualizer.scrollToIndex(focusedVisibleIndex, { align: 'start' });
+    }
+  }, [focusedVisibleIndex, virtualizer]);
```
mayank99, line 65: This useLayoutEffect feels like a hack and it doesn't work when you select an item that's in the middle of the list (below the initial view) and open the dropdown. How were we handling this  previously?

mayank99: Also the `start` behavior is not ideal. Previously we were using `nearest` (see [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#block)).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,18 +58,42 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
-        {visibleChildren}
+    <Box
+      as='div'
+      style={{
+        minBlockSize: virtualizer.getTotalSize(),
+        minInlineSize: '100%',
+      }}
+      {...rest}
+    >
+      <div {...innerProps}>
```
mayank99, line 80: Is this nested div necessary? Previously there were some props coming from the virtualizer, which seems to not be the case anymore

pnpm-lock.yaml
mayank99, line 1: This file shouldn't have so many changes. You might have misformatted it.

.changeset/wet-trainers-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Added dependency on `@tanstack/react-virtual`.
```
mayank99, line 5: I think we should test out that this library works fine for `Table` and `Tree` cases before we finalize it. It would be problematic if we add this dependency and make a release, then later find out that it doesn't handle everything we need.

So while it's fine for this PR to be its own separate thing, the `Table`/`Tree` parts need to be investigated sooner rather than later.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,18 +58,42 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
+          <div
+            key={virtualItem.key}
+            data-index={virtualItem.index}
+            ref={virtualizer.measureElement}
+            style={{
+              position: 'absolute',
+              top: 0,
+              left: 0,
+              width: '100%',
+              transform: `translateY(${virtualItem.start}px)`,
```
mayank99, line 91: I feel like there should be a way to keep the negative margin we had before. Maybe you could adjust the `transform` (and the total size too)

The collapsing margins are important in the case of multi-select, to avoid showing multiple adjacent outlines. `Table` and `Tree` also have similar design decisions that would be nice to preserve when using virtualization.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: Removed the wrapping div in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/89a25d2099cb669a376560aff8e61a42b45d6c3f), and also removed the code accounting for it from the keyboard navigation handling. Also removed the mention of change to the keyboard handling from the PR description.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: It might not be. Wanted to try and the structure as similar as possible to what was already here, but I can try removing this and see if that causes any issues.

pnpm-lock.yaml
Ben-Pusey-Bentley: Yeah, seems like the file somehow got reformatted. Will fix.

.changeset/wet-trainers-heal.md
Ben-Pusey-Bentley: I do have branches testing this library with the `Table` and `Tree` components. There were some keyboard navigation issues with the `Tree` component, and the `Table` component started failing multiple unit tests because of the added state hook for holing the `table` element. I believe these issues should be fixable, but I will finish investigating them before merging this PR.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: Okay, I will give that a try.

pnpm-lock.yaml
Ben-Pusey-Bentley: Fixed in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/058b7991a0c975c2f6ab97d893a9c1d02625a9b4).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: So, updating the `transform` and total size didn't work. I think this is because the items are rendered based on the scroll distance and the reported element sizes, so when the list is scrolled to the bottom, the final few elements aren't rendered in. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/88e0793c-e7e5-49ba-8f2b-8a0a2e6f4afe

I did find a workaround for the spacing issue, though. A custom `measureElement` function can be passed into tanstack virtual [docs](https://tanstack.com/virtual/latest/docs/api/virtualizer#measureelement). This function allows us to treat the elements after the first in the list as shorter than they actually are. This impacts both the `translate` and total size, since they are based on the size results returned by `measureElement`, and avoids the scrolling issue, since the element sizes coming from the function now match the positioning we want them rendered at. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/9b509627-85ec-4a92-9bb6-c6d982c9e3e7

Added this function in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/6e7911a0cffc3a8b1b60213f41b8c4d2c9debd62).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: This screen-recording shows the problem where arrow-key navigation doesn't work if the selected item is below the initial view.

https://github.com/iTwin/iTwinUI/assets/9084735/e77b9cdb-2498-4893-9010-ff9fa0c87401

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: One more improvement: You can handle these things directly in `virtualItemRenderer`.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,19 +62,35 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
+      style={{
+        minBlockSize: virtualizer.getTotalSize(),
+        minInlineSize: '100%',
+      }}
+      {...rest}
+    >
+      {virtualizer.getVirtualItems().map((virtualItem) => {
+        return React.cloneElement(virtualItemRenderer(virtualItem.index), {
+          key: virtualItem.key,
+          'data-index': virtualItem.index,
```
mayank99, line 83: Is this index different from the already-present `data-iui-index`?

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: Thanks for the screen-recording. I've been able to recreate the issue on my machine and am currently looking for a fix. Definitely seems to be coming from the `scrollToIndex` function, since using the mouse to select causes the same broken state to occur when the menu is reopened. Screen-recording: 

https://github.com/iTwin/iTwinUI/assets/141063311/9eabcb70-c470-429b-a2a1-ddf108f306de

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: Yeah, that `data-index` value is what tanstack virtual uses for its default `measureElement` function [(docs)](https://tanstack.com/virtual/latest/docs/api/virtualizer#measureelement-1). Since we are now using a custom function for `measureElement`, though, we may be able to use our `data-iui-index` instead. I'll test and see if this works the way I think it should.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: Yes, you can pass the `virtualItem` to `virtualItemRenderer` and apply all the props there.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: Did you try using a negative [`gap`](https://tanstack.com/virtual/latest/docs/api/virtualizer#gap) instead of manually measuring it?

Ben-Pusey-Bentley: Yeah, using a gap of -1 gives slightly different spacing than using the custom `measureElement`. The custom `measureElement` is more accurate to what we originally have. Gap spacing: 
image
Measure Element spacing: 
image
I'm not completely sure why these are different, but the second option is a pixel closer when using gap, which throws off the spacing by a pixel for all of the options. The overall difference is minor, so it might be better to use gap instead for readability.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/d50f0c5947cfcf5ef00c593d1819692d2f75b98c).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,19 +73,24 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
-      <div {...innerProps} ref={innerProps.ref}>
-        {visibleChildren}
-      </div>
+    <Box
+      as='div'
+      style={{
+        minBlockSize: virtualizer.getTotalSize(),
+        minInlineSize: '100%',
+      }}
+      {...rest}
```
mayank99, line 89: We should try to merge the `style` prop here.

```jsx
{...rest}
style={{
  minBlockSize: virtualizer.getTotalSize(),
  minInlineSize: '100%',
  ...props.style,
}}
```

Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/6c0ea72969b885b7eff88077dbd08933e7a81151).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: This issue is pretty frustrating, since it seems to be coming from [here](https://github.com/TanStack/virtual/issues/216). The core issue comes from the virtualizer scrolling before the element measurement has finished, causing the scroll to go to the wrong place. There are a couple of workarounds proposed in the linked issue, using functions like `setTimeout` and `requestAnimationFrame` to delay the scroll until the measurement has happened. Both of these seem hacky to me, but there is a different solution. Adjusting the the size estimate to `36px` for the options prevents size adjustments for any options that do not include sub labels.. This fixes the issue for our specific example of a virtualized ComboBox, but wouldn't fix the problem for any ComboBoxes that use a large amount of options with sublabels.
Also started using `auto` as the alignment value, as that seems to be the closest to `nearest`. 
The keyboard navigation issues only seem to occur when the `scrollToIndex` function doesn't scroll the option into view correctly, which shouldn't happen anymore with the latest commit.
Changed the alignment of the scroll and updated the estimate size value in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/4e02ee85a627764fcc97a82ed650f2e581ab503f).

mayank99: Are you sure about the cause of the issue? The issue you linked was fixed in tanstack-virtual v3, so it shouldn't apply anymore.

I would think the actual cause is that you're using `useLayoutEffect`, which is never going to be called at a reliable time. You might want to instead try using a [callback ref](https://react.dev/reference/react-dom/components/common#ref-callback), which gets invoked when the DOM node is actually present. Another idea is to update some state when the measurement completes.

I also think it's okay to delay the scroll using `queueMicrotask` (for a micro-task at the end of the current task) or `setTimeout` (for next task). For this to work well, you know _what_ you're waiting for and _when_ it happens.

As for the size estimate thing, I think we could conditionally use 36 vs 48 based on some heuristics (e.g. if 3 of the first 5 items have sublabels, then use 48). But this is meant to be an "estimate", so we shouldn't rely on it.

Ben-Pusey-Bentley: Tried both adding a callback ref and updating state with the `measureElement` prop passed into `useVirtualizer`, neither worked. The React components for both the outer portion of the `VirtualizedComboBoxMenu1 component and the list box that renders it both rerender whenever they are scrolled, which causes the callback ref to be constantly called whenever scrolling happens, not allowing users to freely scroll once an item has been selected. Video:

https://github.com/iTwin/iTwinUI/assets/141063311/f4edda30-f1a9-442e-8fc6-d830ba7ae86e

Updating the state does not work because updating the state with every completed measure causes too many rerenders to happen, and limiting the state update to just when the focused item is measured doesn't work because elements are only measured when they are rendered, and closing and reopening the box automatically returns to the top, causing elements far enough down the list to not be scrolled to.

Wrapping the code in the `useLayoutEffect` in a `setTimeout` does seem to work aligning items correctly on scroll regardless of the estimate size. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/3bdf6b9c-91e2-470c-9979-e336ff4871bd

Added the wrapping `setTimeout` and also added a heuristic for the estimate size in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/00296e8ad3db213124cd5d26b78e061b31721d1e).

.changeset/wet-trainers-heal.md
Ben-Pusey-Bentley: I have created draft PRs for both the [`Table`](https://github.com/iTwin/iTwinUI/pull/2092) and [`Tree`](https://github.com/iTwin/iTwinUI/pull/2091) components. They should be far enough along at this point to show that the `tanstack` virtual package will work for these components as well.

mayank99 (commented):
Adding @r100-stack since this is the first of the three PRs that creates the `useVirtualScroll` hook used in all three places.

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: There's still an issue. If you scroll away from the current item using your mouse, arrow keys stop working.

https://github.com/iTwin/iTwinUI/assets/9084735/febb04c5-5de6-4c01-a025-d63d7225697a

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,19 +53,50 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
+    },
+    [filteredOptions, getMenuItem, children, virtualizer.measureElement],
+  );
+
   return (
-    <Box as='div' {...outerProps} {...rest}>
-      <div {...innerProps} ref={innerProps.ref}>
-        {visibleChildren}
-      </div>
+    <Box
```
mayank99, line 88: Since we are now exploring the idea of hiding this `<div>` inside shadow DOM in `Table` (https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1644627752) and also in `Tree` (https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1649220865), maybe it makes sense to also do the same in `ComboBox`? It would keep the light DOM consistent regardless of whether virtual scroll is enabled.

/cc @r100-stack

Ben-Pusey-Bentley: Yeah, I think that makes sense. Wrapped that `<div>` in `ShadowRoot` in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/7b20c9d63b680634cd7e56ddd77482644eb4cfb5).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,20 +54,58 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
     );
   }, [focusedIndex, menuRef]);
 
-  const { outerProps, innerProps, visibleChildren } = useVirtualization({
+  const virtualizer = useVirtualScroll({
     // 'Fool' VirtualScroll by passing length 1
     // whenever there is no elements, to show empty state message
```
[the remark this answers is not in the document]
mayank99: This is always going to be a problem with virtual scroll tbh. It's part of why virtualization shouldn't be universally recommended.

This tradeoff is not a dealbreaker, as long as the other important functionality works fine (filtering items, navigating between items, selecting items, etc)

packages/itwinui-react/src/utils/hooks/useVirtualScroll.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useVirtualizer, type Virtualizer } from '@tanstack/react-virtual';
+import { useLayoutEffect } from './useIsomorphicLayoutEffect.js';
+import * as React from 'react';
+
+/**
+ * Wrapper over `useVirtualizer` from `@tanstack/react-virtual` that sets the index attribute to
+ * `data-iui-index` and adds a `useLayoutEffect` for scrolling to a provided index on render.
+ *
+ * @example
+ * const virtualizer = useVirtualScroll(item.length, () => parentRef.current, () => 30, scrollToIndex);
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/326a2e474da2cd23e0b8bda108d18de2734ba370).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -8,13 +8,15 @@ import { Menu } from '../Menu/Menu.js';
...
   useMergedRefs,
-  useVirtualization,
   Portal,
   Box,
+  useVirtualScroll,
+  ShadowRoot,
 } from '../../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 import { ComboBoxStateContext, ComboBoxRefsContext } from './helpers.js';
 import { List } from '../List/List.js';
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated the imports in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/eb471a0448cc4a839a609d367f967793baee2bd4).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -49,19 +53,50 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
+  const virtualizer = useVirtualScroll({
     // 'Fool' VirtualScroll by passing length 1
     // whenever there is no elements, to show empty state message
-    itemsLength: filteredOptions.length || 1,
-    itemRenderer: virtualItemRenderer,
+    count: filteredOptions.length || 1,
+    getScrollElement: () => menuRef.current,
+    estimateSize: () => (mostlySubLabeled ? 48 : 36),
     scrollToIndex: focusedVisibleIndex,
+    gap: -1,
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: There is a margin of `-1px` that is intended to exist between options on the `ComboBox`, but because the tanstack virtualizer relies on absolute positioning, setting the margin doesn't change the spacing between the items. Setting the `gap` to `-1` fixes this issue by having the virutalizer remove a pixel of space between each item, thus maintaining our original spacing.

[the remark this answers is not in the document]
mayank99: More discussion in [this thread](https://github.com/iTwin/iTwinUI/pull/2061#discussion_r1612062062).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -28,13 +29,16 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
...
 
-  const virtualItemRenderer = React.useCallback(
-    (index: number) =>
-      filteredOptions.length > 0
-        ? getMenuItem(filteredOptions[index], index)
-        : (children as JSX.Element), // Here is expected empty state content
-    [filteredOptions, getMenuItem, children],
-  );
+  //Checks the first five (or all, if there are less than 5) filtered options. If at least three (or all, if there are less than 3 options) have sub labels, sets estimate to 48 instead of 36.
+  const mostlySubLabeled = React.useMemo(() => {
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated comment in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/848ab0e557b1d767e8d8185551ecbcb657151e10).

packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
Ben-Pusey-Bentley: This issue was being caused by the virtualization, since the keyboard navigation code relied on being able to find the element that the `focusedIndex` pointed to, which would not be the case if users scroll away from focused elements. Fixed this by having a check for if the `ComboBox` is virtualized and if the `focusedIndex` element is not rendered, then select either the first or last rendered option, if the pressed down or up, respectively. [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/5cb826a6538b5512ef8de73ad9ec8d260a9a8c0e). Screen recording of new behavior:
https://github.com/iTwin/iTwinUI/assets/141063311/65de104d-7aa6-4c08-ab98-2b24bb98b55b

mayank99: I don't think that behavior is right. When pressing <kbd>↑</kbd> or <kbd>↓</kbd>, it should do the same thing it does for non-virtualized comboboxes, i.e. scroll back to the focused item and move focus to the corresponding adjacent item.

If this is too difficult to fix, then we can list this out as a "known issue" somewhere. After all, it is somewhat of an edge case, requiring the use of mouse and keyboard together.

Ben-Pusey-Bentley: Okay. To do that, the `ComboBoxInput` component would need to be able to call the `virtualizer.scrollToIndex` function, is currently being instantiated  in `ComboBoxMenu`. I could try to instantiate the `virualizer` in the `ComboBox` component and then pass it down to `ComboBoxInput` and `ComboBoxMenu` through context or props, but that would be a pretty large refactor. Because of that, I think it might be better to create an issue for it and come back to it later. Do you agree? Also, removed the previous changes to keyboard navigation behavior in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/6387a29580c44968c78457adf9b9832453f44bfb).

mayank99: I agree it could be too big of a change and should be done in a separate PR. Not sure it needs an open issue though. I've created a [draft issue](https://github.com/orgs/iTwin/projects/52/views/1?sliceBy%5Bvalue%5D=iTwin%2FiTwinUI&pane=issue&itemId=68716376) where you can list out this problem. We can convert it to a public issue in the future if necessary.

Ben-Pusey-Bentley: Okay, added a short description of the problem to the draft issue.

packages/itwinui-react/src/utils/hooks/useVirtualScroll.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @example
+ * const virtualizer = useVirtualScroll({count: item.length, getScrollElement: () => parentRef.current, estimateSize: () => 30, scrollToIndex});
+ */
+export const useVirtualScroll = (params: {
+  count: number;
+  getScrollElement: () => HTMLElement | null;
+  estimateSize: (index: number) => number;
+  scrollToIndex?: number;
+  gap?: number;
+  getItemKey?: (index: number) => React.Key;
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated the `params` type and used `...rest` to spread parameters into `useVirtualizer` in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/34e3b8311626d89cedc140a8f4d17437f878e626).

testing/e2e/app/routes/ComboBox/spec.ts
```diff
@@ -0,0 +1,175 @@
...
+    //stay on last
+    await page.keyboard.press('ArrowDown');
+    await expect(comboBoxInput).toHaveAttribute(
+      'aria-activedescendant',
+      'test-component-option-Item-11',
+    );
+    await expect(items.last()).toHaveAttribute('data-iui-focused', 'true');
+
+    //select last
+    await page.keyboard.press('Enter');
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Lines checking for the [aria-controls attribute](https://github.com/iTwin/iTwinUI/blob/bf2b3e3b1a47d50210f5d6b3b24944af67b3d98c/testing/e2e/app/routes/ComboBox/spec.ts#L65) were meant to check if the menu was open or not, but that relies on the `aria-controls` value being correctly setup. Added additional check for the presence of the `ComboBoxMenu` by looking for elements with the `listbox` role in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/bf2b3e3b1a47d50210f5d6b3b24944af67b3d98c).

testing/e2e/app/routes/ComboBox/spec.ts
```diff
@@ -0,0 +1,175 @@
...
+      has: page.locator('slot'),
+    });
+    const items = page.getByRole('option');
+
+    let totalItemsHeight = 0;
+    for (const item of await items.all()) {
+      if (totalItemsHeight > 0) {
+        // accounts for height lost due to custom measure element function
+        totalItemsHeight = totalItemsHeight - 1;
+      }
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated the comment and moved the math outside of the loop in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/516b83ef7c800fbc60f460335a4295fe1b257421).

testing/e2e/app/routes/ComboBox/spec.ts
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Replace `toHaveCount` with `toBeVisible` in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/02f43e165bf52fae15dcde9036c8077b103e1ada). `toBeTruthy`/`toBeFalsy` doesn't seem to work when evaluating locator, and I think that is because the locator is still defined, even if the element that it is pointing to is no longer being rendered.

packages/itwinui-react/src/utils/hooks/useVirtualScroll.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useVirtualizer } from '@tanstack/react-virtual';
+import type { Virtualizer } from '@tanstack/react-virtual';
+import { useLayoutEffect } from './useIsomorphicLayoutEffect.js';
+
+/**
+ * Wrapper over `useVirtualizer` from `@tanstack/react-virtual` that sets the index attribute to
+ * `data-iui-index` and adds a `useLayoutEffect` for scrolling to a provided index on render.
+ *
+ * @example
+ * const virtualizer = useVirtualScroll({count: item.length, getScrollElement: () => parentRef.current, estimateSize: () => 30, scrollToIndex});
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Reformatted the example in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/6d37c01e5209fc2a7106e00bb66088b18a522307).

packages/itwinui-react/src/utils/hooks/useVirtualScroll.ts
```diff
@@ -0,0 +1,40 @@
...
+    overscan: 10,
+    ...rest,
+  });
+  useLayoutEffect(() => {
+    setTimeout(() => {
+      if (scrollToIndex) {
+        virtualizer.scrollToIndex(scrollToIndex, { align: 'auto' });
+      }
+    });
+  }, [scrollToIndex, virtualizer]);
```
mayank99, line 38: Thinking about https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1665813754, can we improve the way scrollToIndex is handled?

Instead of using a layout effect that runs when the passed index changes,we could expose a wrapper over `virtualizer.scrollToIndex` which adds the `setTimeout`, and adds the default `align` while allowing it to be overridden. This way, we could use `align: "auto"` in `ComboBox` and maybe `align: "center"` in `Table`.

```jsx
const scrollToIndex = React.useCallback((index, options) => {
  setTimeout(() => {
    virtualizer.scrollToIndex(index, { align: 'auto', ...options });
  });
}, [virtualizer.scrollToIndex]);

return React.useMemo(() => ({
  ...virtualizer,
  scrollToIndex,
}, [virtualizer, scrollToIndex]);
```

mayank99: Actually it looks like we don't even need to manually create `scrollToIndex`. We can pass a custom function to [`scrollToFn`](https://tanstack.com/virtual/latest/docs/api/virtualizer#scrolltofn), where we simply wrap the exported `elementScroll` with `setTimeout`.

Ben-Pusey-Bentley: For some reason, passing in a `scrollToFn` that is the `elementScroll` wrapped with `setTimeout` didn't work. Exposing the wrapper around the `virtualizer.scrollToIndex` function does work, however. Is the plan to still call this function from within a layout effect, though? Is there some better event handler that we can call this function inside of?

mayank99: We can use `useLayoutEffect` (in `ComboBoxMenu` and `Table`) for now, but in the future I would like to try to remove the `useLayoutEffect` because it is [discouraged](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers).

Ben-Pusey-Bentley: Okay, I've changed the hook to pass back the wrapped function in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/a1adf575c050c874ec79d1f40a146118df99dad3) and updated the other PRs to account for the changes made.

packages/itwinui-react/package.json
```diff
@@ -104,6 +104,7 @@
     "@floating-ui/react": "^0.26.18",
     "@itwin/itwinui-illustrations-react": "^2.1.0",
     "@swc/helpers": "^0.5.11",
+    "@tanstack/react-virtual": "^3.5.0",
```
mayank99, line 107: lets use latest (3.8.2)

https://github.com/TanStack/virtual/releases

Ben-Pusey-Bentley: Updated the package in this [commit](https://github.com/iTwin/iTwinUI/pull/2061/commits/4975651c8ba1606b3ff0d3effdbc292ca9ad48a1). There was a change in the `VirtualItem` type that meant that all instances of it across the branches had to be changed to `VirtualItem<Element>`. I did this in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/3ebe316f8cabcacf49030ad35de19d23989c8cfc) for `Table` and this [commit](https://github.com/iTwin/iTwinUI/pull/2091/commits/afcbd2df67e43f34dac0153688dcbb3447bb211c) for `Tree`.

mayank99 (approved):
LGTM! I re-confirmed that #1107 is fixed

PR #1848, opened 2024-02-13
Move jsdom to devDependencies

Woyken:
## Changes

Package should not include dependencies that are not used during runtime

## Testing

Unit tests pass

## Docs

"N/A"

PR #2065, opened 2024-05-23
bump `lightningcss` to `1.25.0`

mayank99:
## Changes

1.24.1 → 1.25.0. See [release notes](https://github.com/parcel-bundler/lightningcss/releases).

This shouldn't change anything but we should try to be on the latest, since there are often bug fixes.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -28,7 +28,7 @@
     "astro-expressive-code": "^0.33.5",
     "examples": "workspace:*",
     "htm": "^3",
-    "lightningcss": "^1.22.1",
+    "lightningcss": "^1.25.0",
     "nanoid": "*",
     "nanostores": "^0.9.0",
     "outdent": "^0.8.0",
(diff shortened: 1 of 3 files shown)


apps/website/package.json
```diff
@@ -28,7 +28,7 @@
     "astro-expressive-code": "^0.33.5",
     "examples": "workspace:*",
     "htm": "^3",
-    "lightningcss": "^1.22.1",
+    "lightningcss": "^1.25.0",
```
rohan-kadkol, line 31: Unrelated: to confirm, should `lightningcss` be a dev dependency?

mayank99: It doesn't matter, this is not in a published package (e.g. see #1848). And there is no difference for a website built using a framework like astro; everything is a dev dependency.

Ben-Pusey-Bentley (approved):

PR #2066, opened 2024-05-23
Changed tooltip triggers in stories to buttons

rohan-kadkol:
## Changes

In our Tooltip docs ([link](https://itwinui.bentley.com/docs/tooltip#best-practices)), we mention that the trigger should be interactable elements, like buttons.

image

However, our `Tooltip` stories use text as triggers. This PR updates the stories to use `Button`s as triggers.

Also, removed some inline styles in the `Tooltip` stories that were no longer unnecessary.

## Testing

Tested that all `Tooltip` stories look and work correctly.

## Docs

N/A


apps/react-workshop/src/Tooltip.stories.tsx
```diff
@@ -3,91 +3,55 @@
...
 
 export const Left = () => {
   return (
     <Tooltip placement='left' content='Here I am!'>
-      <div
+      <Button
         id='tooltip-target'
         style={{
-          marginTop: 40,
           marginLeft: 100,
```
Ben-Pusey-Bentley, line 41: Why does this `Button` still have some of the inline styles?

rohan-kadkol: I left the inline style for `marginLeft` because without it, the tooltip shows up on the right since there isn't enough space on the left.

rohan-kadkol: But I just realized that the leftover `width` style was not necessary and so removed it (https://github.com/iTwin/iTwinUI/pull/2066/commits/0cb85e23d9e3342de264000b11e9452db8797961).

Ben-Pusey-Bentley (approved):

mayank99 (approved):

PR #2068, opened 2024-05-24
`EditableCell` properly merges user's `className`/`style` with inner props

rohan-kadkol:
## Changes

I noticed that `EditableCell` was not merging the user's `className` and `style` properly with `cellElementProps`. As a result, when users would pass a `className` or `style`, it would override (instead of merging with) the `className` and `style` from `cellElementProps` and thus break the layout.

This PR fixes it by merging the user's `className` and `style` with `cellElementProps`, similar to [DefaultCell](https://github.com/iTwin/iTwinUI/blob/75a7a5ff70927059a14182021015822fcf5c6acf/packages/itwinui-react/src/core/Table/cells/DefaultCell.tsx#L79-L82).

## Testing

Added unit tests for `DefaultCell` and `EditableCell` to confirm that the `className` and `style` are properly merged with `cellElementProps`.

Tested in the vite playground:
* [Current version (3.11.0)](https://stackblitz.com/edit/github-p3k8jw?file=package.json)
* [This PR](https://gist.github.com/rohan-kadkol/3103eac1cde30b822f7d63a88829da99)

<details>
<summary>custom className and style</summary>

```jsx
className='my-custom-cell'
style={{
  backgroundColor: 'black',
}}
```

</details>

|Current version (3.11.0)|This PR|
|-|-|
|image|image|

## Docs

Added changeset

### DIFF
--- a/packages/itwinui-react/src/core/Table/cells/DefaultCell.test.tsx
+++ b/packages/itwinui-react/src/core/Table/cells/DefaultCell.test.tsx
@@ -1,0 +1,44 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { render } from '@testing-library/react';
+import * as React from 'react';
+import type { CellRendererProps } from '@itwin/itwinui-react/react-table';
+import { DefaultCell } from './DefaultCell.js';
+
+const renderComponent = (
+  props?: Partial<React.ComponentProps<typeof DefaultCell>>,
+) => {
+  const dummyTableCellRenderedProps = {
+    cellElementProps: {
+      className: 'iui-table-cell',
+      style: {
+        boxSizing: 'border-box',
+        flex: '1 1 145px',
+        minWidth: '72px',
+        width: '0px',
+      },
+    },
+    cellProps: {
+      column: {
+        id: 'id',
+      },
+      value: 'cell value',
+    },
+  } as CellRendererProps<any>;
+
+  return render(<DefaultCell {...dummyTableCellRenderedProps} {...props} />);
+};
+
+it('should merge className/style with cellElementProps', () => {
+  const { container } = renderComponent({
+    className: 'test-class',
+    style: { color: 'red' },
+  });
+
+  const cellElement = container.querySelector('.iui-table-cell') as HTMLElement;
+  expect(cellElement).toBeTruthy();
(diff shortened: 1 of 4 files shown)


packages/itwinui-react/src/core/Table/cells/EditableCell.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { container } = renderComponent({
+    className: 'test-class',
+    style: { color: 'red' },
+  });
+
+  const cellElement = container.querySelector('.iui-table-cell') as HTMLElement;
+  expect(cellElement).toBeTruthy();
+  expect(cellElement.classList.contains('test-class')).toBe(true);
+  expect(cellElement.style.color).toBe('red');
+});
```
mayank99, line 52: This can be simplified a lot. Avoid introducing unnecessary abstractions and irrelevant details in unit tests. We only need to test one very specific thing here, so pass only that, along with the minimum _required_ props.

```jsx
it('should merge className/style with cellElementProps', () => {
  const { container } = renderComponent(
    <EditableCell
      cellElementProps={{
        key: 'key',
        className: 'original-class',
        style: { width: 100 },
      }}
      className='test-class'
      style={{ color: 'red' }}
      cellProps={{ value: 'value' } as any}
      onCellEdit={() => {}}
      children={null}
    />
  );

  const cellElement = container.querySelector('.original-class') as HTMLElement;
  expect(cellElement).toHaveClass('test-class');
  expect(cellElement.style.width).toBe('100px');
  expect(cellElement.style.color).toBe('red');
});```

packages/itwinui-react/src/core/Table/cells/DefaultCell.test.tsx
mayank99, line 1: Same comment as EditableCell.test

mayank99 (approved):

Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Table/cells/EditableCell.test.tsx
rohan-kadkol: Sounds good, removed some unnecessary fields and functions, like your suggested code (https://github.com/iTwin/iTwinUI/pull/2068/commits/528847fa36cb8b50042482c5edafc7c7f56dd76e).

packages/itwinui-react/src/core/Table/cells/DefaultCell.test.tsx
rohan-kadkol: Removed (https://github.com/iTwin/iTwinUI/pull/2068/commits/528847fa36cb8b50042482c5edafc7c7f56dd76e).

PR #2073, opened 2024-05-29
Added explicit import paths for 'react-table'

### LINKED ISSUE #2072: Missing package.json mappings for itwinui-react/react-table while using "moduleResolution": "node16"

After updating to v3, the package we are consuming iTwinUI in cannot resolve paths to import anything from '@itwin/itwinui-react/react-table':
image

Looks like we cannot resolve the import when the current file is detected to use CJS based on tsconfig. We are using `"moduleResolution": "node16"` which appears to be causing the issue as this `moduleResolution` cannot find `itwinui-react/react-table` paths.

### Expected Behavior

We should be able to use 'import' syntax for '@itwin/itwinui-react/react-table' elements.

### Link to minimal repro

https://stackblitz.com/edit/github-lw2hd6?file=src%2FApp.tsx

### Steps To Reproduce

1. Copy the tsconfig provided in the minimal repo (with `"moduleResolution": "node16"`)
2. Try to import anything from '@itwin/itwinui-react/react-table' 

### Anything else?

_No response_

Antanas-Bentley:
## Changes

Added explicit paths to support `"moduleResolution": "node16"` for `react-table`.
The change fixes #2072 

## Testing

Can be tested with a consuming application using the specified module resolution. No tests should be required otherwise.

## Docs

- None

### DIFF
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -20,12 +20,10 @@
     },
     "./react-table": {
       "import": {
-        "types": "./esm/react-table/react-table.d.ts",
-        "default": "./esm/react-table/react-table.js"
+        "types": "./esm/react-table/react-table.d.ts"
       },
       "require": {
-        "types": "./cjs/react-table/react-table.d.ts",
-        "default": "./cjs/react-table/react-table.js"
+        "types": "./cjs/react-table/react-table.d.ts"
       }
     },
     "./styles.css": "./styles.css",
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -19,7 +19,14 @@
       }
     },
     "./react-table": {
-      "types": "./react-table.d.ts"
+      "import": {
+        "types": "./esm/react-table/react-table.d.ts",
+        "default": "./esm/react-table/react-table.js"
+      },
+      "require": {
+        "types": "./cjs/react-table/react-table.d.ts",
+        "default": "./cjs/react-table/react-table.js"
+      }
     },
     "./styles.css": "./styles.css",
     "./cjs": "./cjs/index.js",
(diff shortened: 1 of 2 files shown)


packages/itwinui-react/package.json
```diff
@@ -19,7 +19,12 @@
       }
     },
     "./react-table": {
-      "types": "./react-table.d.ts"
+      "import": {
+        "types": "./esm/react-table/react-table.d.ts"
+      },
+      "require": {
+        "types": "./cjs/react-table/react-table.d.ts"
+      }
```
mayank99, line 27: Would this also do the same thing?

```suggestion
      "import": {
        "types": "./react-table.d.ts"
      },
      "require": {
        "types": "./react-table.d.ts"
      }
```

There was never meant to be a `react-table` folder inside `esm/` and `cjs/` folder, so it would be nice to not depend on it.

Antanas-Bentley: Just tried it out with your suggestion and it doesn't get picked up, I still get the error about unresolved path :(

mayank99: Ok, lets go with this one for now:

```suggestion
      "types": "./react-table.d.ts",
      "import": {
        "types": "./esm/react-table/react-table.d.ts"
      },
      "require": {
        "types": "./cjs/react-table/react-table.d.ts"
      }
```

packages/itwinui-react/package.json
mayank99, line 1: Can you also add a `patch` changeset using `pnpm changeset`?

Ben-Pusey-Bentley: Was able to run some tests with the vite playground that this change is solving the intended issue without causing issues for other `moduleResolution` types like `Bundler`, so once this changeset is added, this PR should be good to merge.

Antanas-Bentley: Added changeset

mayank99 (approved):

Ben-Pusey-Bentley (approved):

PR #2074, opened 2024-05-30
Fix indentation of tree node without expander button

### LINKED ISSUE #2069: Tree: nodes with and without expander do not align

Nodes that do not have an expander do not align with nodes that do:
newCatCropped

Adding an empty icon forces the nodes to align but adds unnecessary spacing:
newWithIconCropped

### Expected Behavior

Nodes should align regardless of whether they have an expander or not.

### Link to minimal repro

https://stackblitz.com/edit/github-sfphuz?file=src%2FApp.tsx

### Steps To Reproduce

1. Open repro link
2. You can see nodes do not align when icon is not used.

### Anything else?

_No response_

FlyersPh9:
## Changes

When a tree node does not have an expander button we added a `margin-inline-start` equal to that of the expander button's width. However, we forgot to include the `column-gap` which caused misalignment of tree nodes with the expander button VS tree nodes without the expander button.  Closes #2069.

## Testing

- Tested with pre-existing demos.
- Temporarily added a tree node at the root level without an expander button, confirmed before & after the CSS change.
- Confirmed in small and default size tree.
- Updated `Basic tree` story to better test this scenario.

## Docs

n/a


mayank99 (commented):
Should we remove icons from the [Basic story](https://itwin.github.io/iTwinUI/react/?story=tree--basic&theme=dark)? This will serve as a more proper test

Ben-Pusey-Bentley:
> Should we remove icons from the [Basic story](https://itwin.github.io/iTwinUI/react/?story=tree--basic&theme=dark)? This will serve as a more proper test

Yeah, I agree that the icons should be removed from that story. There is some testing through the css-workshop, but I think the react-workshop should also be updated to test for this change as well, since the react package will also be affected.

FlyersPh9:
> > Should we remove icons from the [Basic story](https://itwin.github.io/iTwinUI/react/?story=tree--basic&theme=dark)? This will serve as a more proper test
> 
> Yeah, I agree that the icons should be removed from that story. There is some testing through the css-workshop, but I think the react-workshop should also be updated to test for this change as well, since the react package will also be affected.

I've removed icons and sublabels from the `Basic tree` story.

Ben-Pusey-Bentley (approved):

rohan-kadkol (approved):

PR #2075, opened 2024-05-30
use cyclic state toggles in field CSS

mayank99:
## Changes

This is a very small CSS refactor of `iui-field` that incorporates the [cyclic space toggles](https://kizu.dev/cyclic-toggles/) technique to allow for easy switching between default vs hover vs disabled states.

Extracted out of #2076 for easier review.

## Testing

Everything works exactly the same.

## Docs

N/A.

### DIFF
--- a/packages/itwinui-css/src/field.scss
+++ b/packages/itwinui-css/src/field.scss
@@ -16,10 +16,23 @@
   --_iui-field-focus-outline-width: 2px;
   --_iui-field-font-size: var(--iui-font-size-1);
 
-  background-color: var(--_iui-field-color-background);
-  border: 1px solid var(--_iui-field-color-border);
+  // Cyclic toggles for default/hover/disabled states. See https://kizu.dev/cyclic-toggles/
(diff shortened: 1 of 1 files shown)


packages/itwinui-css/src/field.scss
mayank99, line 1: @FlyersPh9 I noticed this file sets color fill but doesn't set icon fill, which leads to the hover/disabled state colors being spread across two different files (`field.scss` and `button/base.scss`). Do you remember the reasoning behind this choice?

mayank99: That shouldn't be a problem though. Input and textarea will just ignore the icon fill if it's set in `field`.

Ben-Pusey-Bentley (approved):

rohan-kadkol (approved):

packages/itwinui-css/src/field.scss
FlyersPh9: #2088

PR #2076, opened 2024-05-30
add "open" popover styling to buttons

mayank99:
## Changes

This PR adds "open" state styling to buttons that have a popover (or dropdown menu) that is currently open. Right now, it just reuses hover state styling, which is a subtle, non-offensive choice. (Can be made more noticeable in the future if desired)

Builds upon #2075.

### CSS

The hover state is toggled on when the popover is open. This is done at the `iui-button` level (we can think about adding it to other fields in the future if needed).

I used a data attribute (`data-iui-has-popover`) and also added `:has(+ :popover-open)` as a hook that we can rely on in the future if we use the `popover` API in more places.

### React

A new `PopoverOpenContext` is exported. Its value is provided in multiple components where a trigger element is present.

`Button` and `IconButton` read from this context and set the data attribute accordingly.

## Testing

Added unit tests that either directly or indirectly check the affected code paths.

Updated affected visual test images. (Not sure why the focus moved in some cases; probably some flakyness).

Manually tested a whole bunch of stories that are affected. I encourage reviewers to do the same. (Start with [DropdownButton](https://itwin.github.io/iTwinUI/2076/react/?story=dropdownmenu--basic) for example)

https://github.com/iTwin/iTwinUI/assets/9084735/3dc04b2d-5684-40fd-8688-d52d2ecf88cc

## Docs

Added changesets for CSS and React.


apps/react-workshop/cypress-visual-screenshots/baseline/DatePicker.test.ts-Basic.png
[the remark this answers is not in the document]
mayank99: Yeah we should add `Popover` in all places, similar to #1880

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Full2.png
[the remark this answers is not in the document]
mayank99: It's unrelated but looks more accurate now so I approved the image.

apps/react-workshop/cypress-visual-screenshots/baseline/DatePicker.test.ts-Basic.png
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Should this be a separate PR?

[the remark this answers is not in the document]
mayank99: Yes, it's unrelated. Feel free to open a PR

Ben-Pusey-Bentley (approved):
Manually tested the affected react stories, only found the issue with most of the `DatePicker` stories not using the `Popover` component. LGTM.

packages/itwinui-css/src/button/button.scss
```diff
@@ -10,6 +10,11 @@
 .iui-button {
   @include base.iui-button;
 
+  // Reusing hover styling when an associated popover/dropdown is open.
+  &:is([data-iui-has-popover='open'], :has(+ :popover-open)) {
```
rohan-kadkol, line 14: To confirm, what is the `:has(+ :popover-open)` for? Is it a safeguard for when we use the Popover API in more places in the future?

packages/itwinui-react/src/core/Buttons/DropdownButton.test.tsx
```diff
@@ -41,11 +41,12 @@ it('should render in its most basic state', () => {
   expect(container.querySelector('.iui-button-icon')).toBeTruthy();
 });
 
-it('should update icon when menu opens or closes', async () => {
+it('should update when menu opens or closes', async () => {
   const { container } = renderComponent();
 
   const button = container.querySelector('.iui-button') as HTMLButtonElement;
   expect(button).toBeTruthy();
+  expect(button).not.toHaveAttribute('data-iui-has-popover', 'open');
```
rohan-kadkol, line 49: Would it be safer to have `expect(button).not.toHaveAttribute('data-iui-has-popover')`, i.e. confirm that the attribute doesn't exist regardless of the value? So, if `data-iui-has-popover: opened` is present instead of `data-iui-has-popover: open`, the test should fail?

mayank99: No, because it would also fail unnecessarily if we do something like `data-iui-has-popover="closed"` in the future.

rohan-kadkol: Ohh, I see. Yes, then as-is is good.

nit/optional: I was just suggesting this to confirm that `data-iui-has-popover` is not in the DOM, because the current check will not catch that. Also, maybe fixing a few tests may not be a big issue if we add `data-iui-has-popover="closed"` in the future.

packages/itwinui-css/src/button/button.scss
rohan-kadkol: Ohh, I see, I was also hoping to understand the DOM for this. Since the floating content will likely be portaled to the `ThemeProvider`, will this sibling selector have any effect?

mayank99: It should not be portaled. We kept the portal implementation in tooltip, but ideally it should be a sibling.

```html
<button popover popovertarget="my-popover">…</button>
<div id="my-popover">…</div>
```

This is important to allow keyboard/AT users to easily navigate back and forth between the trigger and the popover, without needing to write a lot of JS to manage keyboard focus and virtual focus.

rohan-kadkol (approved):

packages/itwinui-react/src/core/Buttons/DropdownButton.test.tsx
mayank99: Leaving it as-is because checking whether `data-iui-has-popover` is removed from the DOM is not important. The CSS only cares about the exact `="open"` value, so it will not match even if a different value is used.

FlyersPh9:
> Waiting for @FlyersPh9 to create a PR moving icon fills to field CSS (see [thread](https://github.com/iTwin/iTwinUI/pull/2075#discussion_r1621400705)), because that change will likely affect visuals in this PR.

#2088 is merging any minute.

mayank99 (commented):
[`231481a`](https://github.com/iTwin/iTwinUI/pull/2076/commits/231481aa94119cbd9a2d1e07e783476549380aba) updates all affected images after merging latest changes.

PR #2077, opened 2024-05-31
Added popovers to `DatePicker` stories.

Ben-Pusey-Bentley:
## Changes

Added popovers to all the `DatePicker` react stories to bring them inline with our recommendations on how to use the `DatePicker` [docs](https://itwinui.bentley.com/docs/datepicker#usage). This change also insures that any changes to `Popover` styles are reflected in all `DatePicker` stories, and not just the basic story. Noticed in this [comment](https://github.com/iTwin/iTwinUI/pull/2076#discussion_r1622360770).

## Testing

Made sure that react-workshop tests continue to pass. Updated the image for the range example, since the `Popover` element is positioned based on the button, and not the text, causing the image to change.

## Docs

N/A


apps/react-workshop/src/DatePicker.stories.tsx
```diff
@@ -58,21 +58,24 @@ export const WithTime = () => {
 
   return (
     <>
-      <IconButton onClick={() => setOpened(!opened)} id='picker-button'>
-        <SvgCalendar />
-      </IconButton>
+      <Popover
+        content={
+          <div style={{ marginTop: 4 }}>
```
mayank99, line 63: Can you also remove the `marginTop`? You can use [`middleware.offset`](https://itwinui.bentley.com/docs/popover#positioning) if you want to keep the gap.

mayank99: Although, I wonder if we even need the gap. 🤔

Opinions? @rohan-kadkol

Ben-Pusey-Bentley: I'll remove the `<div>` and use `middlewar.offeset` to maintain the gap for now.

rohan-kadkol: > Although, I wonder if we even need the gap. 🤔

I'm not sure of the reason, but looking at the git history, looks like this was added back in April 2021 ([iTwin/iTwinUI-react #4 (files)](https://redirect.github.com/iTwin/iTwinUI-react/pull/4/files#diff-d92e6fc01405e3a187caa82f9fb11ed2bc03ba56c71ff84750c3b22885e16709L48-R48)) and so may be outdated now with the new `Popover` component.

mayank99: Yeah lets get rid of the gap then. We don't show a gap in the website demos https://itwinui.bentley.com/docs/datepicker

Ben-Pusey-Bentley: `div` removed and `offset` added in this [commit](https://github.com/iTwin/iTwinUI/pull/2077/commits/f66d1b6a478a9a471995cecd614c3b5f6d0fed5d).

Ben-Pusey-Bentley: Sorry, github didn't show me the follow up comments until now. Removed `offset` gap and updated test images to match in this [commit](https://github.com/iTwin/iTwinUI/pull/2077/commits/c9357d0f11a014cb0a7d990197e7960166fe8b05).

mayank99 (approved):

apps/react-workshop/src/DatePicker.stories.tsx
```diff
@@ -58,21 +58,22 @@ export const WithTime = () => {
...
             onChange={onChange}
             setFocus
             showTime
             showDatesOutsideMonth={false}
           />
-        </div>
-      )}
+        }
+      >
+        <IconButton onClick={() => setOpened(!opened)} id='picker-button'>
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: I tested this and they don't seem to be needed, so I removed them in this [commit](https://github.com/iTwin/iTwinUI/pull/2077/commits/171533b6aad628845121f5ad072654c0ce5c4973).

PR #2080, opened 2024-06-01
Small improvements to `ColorPicker` examples and stories

rohan-kadkol:
## Changes

Based on https://github.com/iTwin/iTwinUI/pull/2009#discussion_r1580160704, this PR:
* Adds a `Popover` in the `ColorPicker.main` example.
  * Didn't do the same in the `ColorPicker.advanced` example since not having a Popover seemed intentional in this case (there is a `ColorPicker.advancedPopover` example that shows mostly the same example, but with a `Popover`).
* Uses the `Popover`'s `onVisibleChange` instead of toggling the open state using the `IconButton`'s `onClick` in the `ColorPicker.main` example.
* Removes the unnecessary `margin: 0` on the `ColorSwatch` in some of the stories.

## Testing

Confirmed that the result of the changes in the examples and stories look and work correctly.

## Docs

N/A

### DIFF
--- a/examples/ColorPicker.main.jsx
+++ b/examples/ColorPicker.main.jsx
@@ -8,6 +8,9 @@ import {
   ColorInputPanel,
   ColorPicker,
   ColorValue,
+  IconButton,
+  ColorSwatch,
+  Popover,
 } from '@itwin/itwinui-react';
 
 export default () => {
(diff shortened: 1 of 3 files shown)


examples/ColorPicker.main.jsx
```diff
@@ -20,9 +23,20 @@ export default () => {
...
+        <ColorPicker
+          selectedColor={activeColor}
+          onChangeComplete={onColorChanged}
+        >
+          <ColorBuilder />
+          <ColorInputPanel defaultColorFormat={'hsl'} />
+        </ColorPicker>
+      }
+    >
+      <IconButton>
```
Ben-Pusey-Bentley, line 37: This `IconButton` should have a label, like the on in the advanced popover example. The example doesn't meet accessibility requirements without it.

mayank99 (approved):

Ben-Pusey-Bentley (approved):

PR #2083, opened 2024-06-03
Improved `Button`'s and `IconButton`'s `loading` state

r100-stack:
## Changes

Addition to #2058.

Some issues I noticed:
* `IconButton` accepts the `loading` prop but does nothing with it
  * Fixed by: removing the `loading` prop from `IconButton`'s types. Actual functionality of `loading` in `IconButton` can be brought later (https://github.com/iTwin/iTwinUI/pull/2083#discussion_r1626326297).
* When `loading={true}`, only the `.iui-button-label` was getting hidden. The `startIcon` and `endIcon` were not getting hidden.
  * Fixed by: Updating the button CSS styles for `data-iui-loading` to hide *all* children of the loading button, except `.iui-button-spinner`.

## Testing

* Created testing playground ([before this PR](https://stackblitz.com/edit/github-amgyvy?file=src%2FApp.tsx), [after this PR](https://gist.github.com/r100-stack/49d43b3900bd3ecbc1b39b30fa300030))

|`loading={false}`|`loading={true}` in versions before this PR|`loading={true}` in this PR|
|-|-|-|
|image|image|image|

* Added/Edited unit tests to confirm:
  * `loading` and `size` props work in `Button`
  * `size` prop works in `IconButton`
* Added a visual test of `Button` with `loading` and `startIcon` by updating the `Button → Loading` story to use a `startIcon`.

## Docs

Added changesets

### DIFF
--- a/.changeset/nasty-seals-fry.md
+++ b/.changeset/nasty-seals-fry.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-css': patch
 ---
 
-Fixed a visual bug where elements like `iui-button-icon` inside the `iui-button` where not hidden when the button was in a loading state. Now, `data-iui-loading="true"` on `iui-button` hides *all* its children except `iui-button-spinner`.
+Fixed a visual bug where elements like `iui-button-icon` inside the `iui-button` were not hidden when the button was in a loading state. Now, `data-iui-loading="true"` on `iui-button` hides *all* its children except `iui-button-spinner`.
--- a/.changeset/silver-jokes-nail.md
+++ b/.changeset/silver-jokes-nail.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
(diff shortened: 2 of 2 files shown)


Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/utils/components/ButtonBase.tsx
mayank99, line 1: I don't think this prop should be added to `ButtonBase`. It's only designed to be used in `Button`.

`ButtonBase` is a lower-level primitive that should not bring any visuals with it.

mayank99: I'd suggest reverting all changes except the small CSS fix.

mayank99: Can we leave `IconButton` out of this PR? It's unrelated to the bug fix and I don't think it's even necessary (we can revisit in the future if someone requests it).

mayank99 (approved):

PR #2088, opened 2024-06-04
`.iui-field` handles icon fill

FlyersPh9:
## Changes

- Add 3 `.iui-field` variables:
  - `--_iui-field-color-icon`
  - `--_iui-field-color-icon-hover`
  - `--_iui-field-color-icon-disabled`
- Colors inherited in button and select.

## Testing

- Ran visual tests.

## Docs

n/a


packages/itwinui-css/src/select/select.scss
```diff
@@ -35,7 +35,7 @@
   }
 
   &:where([disabled], :disabled, [aria-disabled='true'], [data-iui-disabled='true']) {
-    --_iui-list-item-icon-fill: var(--iui-color-icon-disabled);
+    --_iui-list-item-icon-fill: var(--_iui-field-color-icon-disabled);
```
FlyersPh9, line 38: It's weird that we are setting disabled fill color, but not setting default / hover state fill color within `select.scss`.

Ben-Pusey-Bentley (commented):
I am a bit confused about the purpose of this PR. The conversation that this PR is in response to seems to be about adding the icon fill logic into the field file, but this PR only adds a few variables to that file, and leaves all of the logic setting icon fill in the button and select files. Is this what we want?

mayank99 (commented):
Adding to @Ben-Pusey-Bentley's comment, this PR looks half-baked. `iui-field` needs to handle the actual icon fill, most importantly the fact that it changes between default, `:hover`, and `disabled` states.

packages/itwinui-css/src/select/select.scss
Ben-Pusey-Bentley: It seems like some of the icon fill logic is currently being handled by this [menu mixin](https://github.com/iTwin/iTwinUI/blob/dac60976090fb86069bb0a4807a0ecd0a550e1fa/packages/itwinui-css/src/menu/mixins.scss#L181-L189). Should probably also move this logic into field as well.

FlyersPh9:
> `iui-field` needs to handle the actual icon fill

Would it be acceptable to add an `.iui-field-icon` class?  Button uses `.iui-button-icon` and Select uses `.iui-icon`, I think we'll need a shared class to apply the fill correctly.

mayank99:
> Would it be acceptable to add an `.iui-field-icon` class?

@FlyersPh9 I don't like the idea of a `iui-field-icon` class. We already have too many icon classes that need to be consolidated before we add another one.

For this PR, it might be simpler to set the `fill` in `iui-field` and inherit it wherever it needs to be used. This way, we get to keep the markup unchanged.

```scss
.iui-field {
  // …
  fill: var(--_iui-field-state--default, var(--_iui-field-color-icon))
    var(--_iui-field-state--hover, var(--_iui-field-color-icon-hover))
    var(--_iui-field-state--disabled, var(--_iui-field-color-icon-disabled));
}
```
```scss
.iui-button-icon svg {
  fill: inherit;
}
```

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -353,19 +352,24 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
...
-    &[data-iui-active='true']:enabled,
-    &[data-iui-active='true']:enabled:hover,
-    &[data-iui-active='true']:active {
-      background-color: rgba(0, 0, 0, var(--iui-opacity-4));
+      &:hover,
+      &[data-iui-active='true']:enabled,
+      &[data-iui-active='true']:enabled:hover,
+      &[data-iui-active='true']:active {
+        background-color: rgba(0, 0, 0, var(--iui-opacity-4));
+      }
```
FlyersPh9, line 372: Tile was failing React tests.  Looked into it and our CSS demo's HTML structure was not matching React's.

**React:**
```html
<div class='iui-tile-thumbnail-quick-action'>
  <button class="iui-button iui-button-base iui-field">…</button>
</div>
```

**CSS:**
```html
<button class="iui-button iui-button-base iui-field iui-tile-thumbnail-quick-action">…</button>
```

Made the changes in CSS demos to match React & had to update the CSS to work properly.  The hover state of the buttons overlaying the thumbnail aren't even working properly in prod right now: https://itwin.github.io/iTwinUI/react/?mode=preview&story=tile--all-props

FlyersPh9:
> For this PR, it might be simpler to set the `fill` in `iui-field` and inherit it wherever it needs to be used. This way, we get to keep the markup unchanged.

This is done.  @mayank99 & @Ben-Pusey-Bentley, I think this could use another review now.

packages/itwinui-css/src/button/base.scss
```diff
@@ -73,8 +66,7 @@ $disabledButtonSelectors: '[disabled], :disabled, [aria-disabled="true"], [data-
   }
 
   &:is(#{$disabledButtonSelectors}) {
-    --_iui-button-active-stripe-background: var(--iui-color-icon-disabled);
-    --_iui-button-icon-fill: var(--iui-color-icon-disabled);
+    --_iui-button-active-stripe-background: var(--_iui-field-color-icon-disabled);
```
mayank99, line 69: Undo this one, it's unrelated to field

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -353,19 +352,24 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
     justify-self: end;
   }
 
-  .iui-tile-thumbnail-picture ~ & {
-    @include mixins.iui-blur($opacity: 5);
-    --_iui-button-icon-fill: var(--iui-color-white);
+  .iui-button[data-iui-variant='borderless'] {
```
mayank99, line 355: It might be better to create a new class for this instead of referencing iui-button selectors here.

mayank99: Also this needs a changeset.

FlyersPh9: > It might be better to create a new class for this instead of referencing iui-button selectors here.

@r100-stack said the user can place anything in this slot.  So we would need the user to pass this new class for it to work correctly, no?

mayank99: Technically yes, the user can place anything, but it's mainly designed to work with `Tile.IconButton`, where we can add this class automatically.

Ben-Pusey-Bentley (commented):
Looking at your CI output, it looks like the changes that you have made to the CSS tile have also introduced some minor visual changes. It seems to be something with the border of the icon buttons. May want to tweak the CSS tile some more to avoid these visual changes, or you may want to approve some new test images, if you feel that the visual change is fine to keep. Diff: 
image

packages/itwinui-css/src/tile/tile.scss
mayank99: If you mean like `> *`, yes that would work for most cases, but not for this:

```jsx
<Tile.QuickAction>
  <div>
    <Tile.IconButton>…</Tile.IconButton>
  </div>
</Tile.QuickAction>
```

It could also have unintended consequences for this scenario (probably rare):

```jsx
<Tile.QuickAction>
  <Tile.IconButton>…</Tile.IconButton>
  <div>…</div>
</Tile.QuickAction>
```

I think it makes sense to have a dedicated class for `Tile.IconButton`, since it has its own distinct look and feel. `Tile.QuickAction` and `Tile.TypeIndicator` should only handle positioning.

One thing to consider, however, is that this could be a potential breaking change. Maybe that's ok? We already require users to [change their usage](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-react-v3-migration-guide#tile) going from v2→v3.

packages/itwinui-css/src/select/select.scss
```diff
@@ -35,13 +35,17 @@
...
 
-  .iui-icon svg {
-    @include menu.iui-menu-icon;
+  .iui-icon {
+    flex: 0 0 auto;
+
+    svg {
+      @include mixins.iui-icon-style('m');
+      fill: inherit;
+    }
```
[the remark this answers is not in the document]
mayank99: It's because this is using `fill: inherit`.

packages/itwinui-css/src/button/base.scss
FlyersPh9: Reverted.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: Ahh, I see.  I've added a new `.iui-tile-thumbnail-button` and updated both the CSS demos and React.

packages/itwinui-css/src/select/select.scss
FlyersPh9: Removed the menu mixin.

packages/itwinui-css/src/tile/tile.scss
FlyersPh9: I added this to the changesets.  Unsure if this warrants a major or not.

packages/itwinui-css/src/tile/tile.scss
```diff
@@ -353,19 +352,24 @@ $thumbnail-selector: ':is(.iui-thumbnail-icon, .iui-tile-thumbnail-picture)';
     justify-self: end;
   }
 
-  .iui-tile-thumbnail-picture ~ & {
-    @include mixins.iui-blur($opacity: 5);
-    --_iui-button-icon-fill: var(--iui-color-white);
+  .iui-tile-thumbnail-button {
+    border-radius: 50%;
```
FlyersPh9, line 356:
```suggestion
    border-radius: var(--iui-border-radius-round);
```

Feel like this should be changed, but will likely effect visuals tests.  If we like this change, I'll make it right at the end to keep image test changes until the last push.

packages/itwinui-css/src/tile/tile.scss
mayank99: Should we move the Tile changes into a separate PR? I think it warrants additional discussion.

FlyersPh9: Ok, working on that when I get back to my desk tonight.

FlyersPh9: Moved tile content into #2101.  Images are failing now on this PR, I didn't feel comfortable approving test images knowing they are incorrect, unsure how to proceed without making some sort of change within tile.

mayank99: This PR should preferably have the minimum changes necessary for making the images match the `main` branch. Is it possible to do that without large changes in `tile.scss`?

FlyersPh9: https://github.com/iTwin/iTwinUI/pull/2088/commits/73cb4ed3c00fc2d41c3149c4a756a9706ab14c73 seems to get the job done for now.

mayank99 (approved):
LGTM

apps/react-workshop/cypress-visual-screenshots/baseline/Tile.test.ts-All Props.png
Ben-Pusey-Bentley, line 1: Are there any visual changes that this screenshot is being updated for? If not, it would be nice to overwrite this with the screenshot file from the main branch so that it doesn't appear in the diff for not reason.

FlyersPh9: I tried downloading the source image from GitHub and replacing it in a push but it still showed as changed.  @r100-stack showed me a way to restore the original image using GitLense, hopefully images will be passing after https://github.com/iTwin/iTwinUI/pull/2088/commits/25ff29a6663d69b31b6f1734b96033a0c25fbfdf

Ben-Pusey-Bentley (approved):

.changeset/small-otters-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': minor
+---
+
+- `.iui-field` handles icon fill, used within button and select.
```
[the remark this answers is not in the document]
FlyersPh9: Fixed.

PR #2092, opened 2024-06-06
Adds tanstack virtualization to `Table` Component

Ben-Pusey-Bentley:
## Changes

Updates `Table` component to use tanstack virtualization instead of the old virtualization implementation. This PR is a part of the larger process of removing the old virtualization implementation from iTwinUI entirely. Removes references in the `Table` component to the `VirtualScroll` component, which will be removed from the entirely in the [`Tree`](https://github.com/iTwin/iTwinUI/pull/2091) PR. This property is optional to avoid having to make changes to the `ComboBox` and `Tree` components in this PR that would just be undone by the PRs for those components anyways. Also wraps the `ShadowRoot` on the column headers in a fragment to resolve the react key warnings.

After PR TODOs

- [ ] Fix `border-radius` for the top left and right corners of `Table`. [discussion](https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1664229041)
- [ ] Investigate adding `contain: strict` to the outer most container `<div>` on virtualized components to improve performance. [discussion](https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1658842656)
- [ ] Memoize column resizing ref function to fix layout thrashing [discussion](https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1672615254)

## Testing

Updated some virtual scroll unit tests, and replaced others with `Table` e2e tests. Confirmed that all other tests pass.

## Docs

Added changeset.


mayank99 (commented):
Can you [stack](https://timothya.com/blog/git-stack/) the three PRs to make them easier to review?

Ben-Pusey-Bentley:
> Can you [stack](https://timothya.com/blog/git-stack/) the three PRs to make them [easier to review](https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/)?

I've rebased the PRs so that main <- ComboBox PR <- Table PR <- Tree PR.  Should I also merge the new bases into the Table and Tree PRs?

mayank99:
> Should I also merge the new bases into the Table and Tree PRs?

Yeah, all three branches need to be up-to-date. Currently this branch is 30 commits behind its base.

packages/itwinui-react/src/utils/components/VirtualScroll.tsx
mayank99, line 1: I'm not sure I understand the way you've implemented this PR. In `ComboBox` and `Tree` you're using `useVirtualizer` directly in the component. Why not do the same in `Table`?

The `VirtualScroll` component and the old hook `useVirtualization` hook could be removed completely in the end.

Ben-Pusey-Bentley: I implemented it this way to maintain the `VirtualScroll` component, since the tanstack package doesn't provide a component for wrapping the virtualization, just the `useVirtualizer` hook. The idea was to have the `VirtualScroll` component remain identical, while replacing the old `useVirtualization` with `useVirtualizer`. Reconsidering this though, it doesn't make  a lot of sense to maintain a util component that will only be used by one of our components. The `VirtualScroll` component can be removed in this PR, but because of how the stack is ordered, the `Tree` PR would have to be the one to remove the `useVirtualization` hook. Is that fine? Should I try to reorder the stack so that both parts can be removed in this PR?

mayank99: No need to reorder the stack. In this PR you can simply change `Table` to *not* rely on `VirtualScroll` and leave the utils touched. You can then remove all dead code in the final PR in the stack (whether that's the `Tree` PR or a separate fourth PR).

**Note**: I'm not necessarily against having abstractions, even if it's only used in one place. But there needs to be a good reason to do so. Using the custom `VirtualScroll` component just seemed odd to me, since the other two PRs were directly consuming tanstack-virtual.

Ben-Pusey-Bentley: Removed `VirtualScroll` from the `Table` component in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/b21e592496579cafbc9044cab0e68a7d229bbfbf), and replaced the last remaining `VirtualScroll` unit test with a `Table` e2e test in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/dcbe24243419651bb43d3b57190cda4e01ffcc4c). Also made a commit to the `Tree` PR to remove the dead code [here](https://github.com/iTwin/iTwinUI/pull/2091/commits/00ddeb87cda3633182ba048c6d4a4e33ef4be96f).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +823,41 @@ export const Table = <
     }
   });
 
+  const virtualizer = useVirtualizer({
+    count: page.length,
+    getScrollElement: () => tableElement ?? null,
+    estimateSize: () => 62, //Set to 62px since that is the default height of a table row.
```
mayank99, line 829: this estimate should probably differ based on the `density` prop

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +823,41 @@ export const Table = <
...
+    style: {
+      minBlockSize: virtualizer.getTotalSize(),
+      minInlineSize: '100%',
+      ...style,
+    },
+    ...rest,
+  } as React.HTMLAttributes<HTMLElement>;
+
+  const innerProps = {
+    style: { willChange: 'transform' },
```
mayank99, line 844: I believe the `<div {...innerProps}>` is no longer needed anywhere. The new virtualizer only requires a single wrapping div instead of two.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -903,6 +949,9 @@ export const Table = <
         ref={useMergedRefs(tableRef, (element) => {
           ownerDocument.current = element?.ownerDocument;
           resizeRef(element);
+          if (enableVirtualization) {
+            setTableElement(element);
```
mayank99, line 953: nit: this does not need to be guarded under `if`. It can be passed directly into `useMergedRefs`

```jsx
ref={useMergedRefs(tableRef, setTableElement, (element) => {
```

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1114,11 +1163,19 @@ export const Table = <
           {data.length !== 0 && (
             <>
               {enableVirtualization ? (
-                <VirtualScroll
-                  itemsLength={page.length}
-                  itemRenderer={virtualizedItemRenderer}
-                  scrollToIndex={scrollToIndex}
-                />
+                <div {...outerProps}>
```
mayank99, line 1166: I'm thinking we should move this outer wrapper div into a `<ShadowRoot>` (attached to the table-body), so that it stays "hidden" from the light DOM.

packages/itwinui-react/src/utils/components/VirtualScroll.tsx
Ben-Pusey-Bentley: Added an abstraction in the `ComboBox` PR [here](https://github.com/iTwin/iTwinUI/blob/a75174f637f7fa73ec833440900695320c08276b/packages/itwinui-react/src/utils/hooks/useVirtualScroll.ts#L16). This `useVirtualScroll` hook handles setting up the `useLayoutEffect` for scrolling to a provided index and initializing the virtualizer with some given parameters. Started using it for `Table` with this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/507a69420bea6908e9c2555928febf5057e50be7).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Updated the estimate to vary on the `density` prop in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/0650ccddd1ec8800db5b6107ddcfd923ffe37095).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Removed the `innerProps` object and `<div>` element in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/105098f1819da3cfb8b47a090754d51ccf22bf51). For the `Tree` component, the `innerProps` are going on an `<ul>` inside of the outer `<div>` component. On top of that, trying to put the styles from the `innerProps` onto this outer `<div>` causes the entire tree to take the height of the `virtualizer.getTotalSize()` value, which defeats the point of the virtualization. Is it fine to leave the `Tree` component with its `innerProps` for now?

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: I have the function being passed directly into `useMergedRefs` in the latest commit, but this change is causing some maximum update depth exceeded errors in the `Table` unit tests.
image
Any ideas on how to work around this without guarding the `setTableElement` function? Should I try to replace these unit tests with e2e tests?

mayank99: These errors are a red herring and usually indicative of a larger problem somewhere else. You should try to reproduce this issue in a playground and spend some time debugging it.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: I'm not sure I understand the problem. Didn't we talk about not having an outer `<div>` in `Tree`? Can you share the code you're trying to use that's causing the problem?

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Added `<ShadowRoot>` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/b6afd1397941969e87f9f3911001c6c2e32e842e).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: There was the outer `<div>` in the react workshop example that was outside of the `Tree` element itself, which is what made the recursive search for a scrollable parent necessary. That was removed, but there is another outer `<div>` inside of the `Tree` component that is now acting as the scrollable parent for the virtualization. Putting the full size on the scrollable parent breaks the virtualization, since the virtualizer tries to load all of the items at the same time. This isn't an issue for the `Table` and `ComboBox` component, since they have higher level scrollable elements that the scroll ref can point to instead. `Tree` doesn't have an element like that, though, so it has to wrap the `TreeElement` in a `<div>` that is scrollable. There can't be an inner `<div>` with the height while the `TreeElement` is the scrollable element because the `TreeNode` elements are `<li>` elements at their top level. Therefore, it seems like the `innerProps` and `outerProps` structure is still needed on `Tree`, unless I am completely missing something. Broken code that loads in every element at once:

```jsx
 const outerProps = {
      className: className,
      style: {
        minInlineSize: '100%',
        overflow: 'auto',
        minBlockSize: virtualizer.getTotalSize(),
        willChange: 'transform',
        ...style,
      },
    } as React.HTMLAttributes<HTMLElement>;

    const innerProps = {
      style: {},
    } as React.HTMLAttributes<HTMLElement>;

    return (
      <Box {...outerProps} ref={parentRef}>
        <TreeElement {...innerProps} {...rest} ref={ref}>
          {virtualizer
            .getVirtualItems()
            .map((virtualItem) =>
              itemRenderer(virtualItem.index, virtualItem, virtualizer),
            )}
        </TreeElement>
      </Box>
    );
```

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: This doesn't do anything, because your wrapper div is not inside the shadow root.

I'm suggesting something like this:

```jsx
<ShadowRoot>
  {enableVirtualization ? (
    <div {...outerProps}><slot /></div>
  ) : (
    <slot />
  )}
</ShadowRoot>

{data.length !== 0 && (
  {enableVirtualization ? (
    virtualizer
      .getVirtualItems()
      .map((virtualItem) =>
        virtualizedItemRenderer(
          virtualItem.index,
          virtualItem,
          virtualizer,
        ),
      )
  ) : (
    page.map((_, index) => getPreparedRow(index))
  )}
)}
```

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: I don't think we should be adding a wrapper div around `TreeElement`. The `...rest` props and forwarded `ref` should almost always be spread on the topmost element of the component.

But we should be able to add an extra div *inside* `TreeElement`, and even hide it inside shadow DOM, similar to https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1644627752. Can you explain why this is a problem: "the `TreeNode` elements are `<li>` elements at their top level"?

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Fixed in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/65c2229e35592be539d44d33cc0f9e923dd4f6fb).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: So the `TreeElement` component is a polymorphic `<ul>`, and the `TreeNode` component returns a `<li>` element, so, from my understanding, putting a `<div>` between the two elements would create an invalid DOM, since the `<li>` needs a list element as its parent, and the `<div>` would be the parent in that case. Wrapping the inner `<div>` in the shadow DOM would make the light DOM valid, though, so I've been trying to implement that. However, I am currently getting these errors when I try to attach a `ShadowRoot` to the `TreeElement`:
image
I am not sure what the cause of this is. Could it be because the `TreeElement` is being defined by `polymorphic`? code:
```jsx
const TreeElement = polymorphic.ul('iui-tree', {
  role: 'tree',
  tabIndex: 0,
});


return (
      <TreeElement
        {...outerProps}
        {...rest}
        ref={useMergedRefs(ref, parentRef)}
      >
        <ShadowRoot>
          <div {...innerProps}>
            <slot />
          </div>
        </ShadowRoot>
        <>
          {virtualizer
            .getVirtualItems()
            .map((virtualItem) =>
              itemRenderer(virtualItem.index, virtualItem, virtualizer),
            )}
        </>
      </TreeElement>
```

mayank99: That's a very valid point about invalid DOM. We can change the `<ul>` to be `<div>` though. The `<ul>`+`<li>` doesn't give us anything useful, because we are manually setting their respective roles to [`"tree"`](https://github.com/iTwin/iTwinUI/blob/3b5fe97caf19d4314e0f91645c9ec6b9cdb1c8d8/packages/itwinui-react/src/core/Tree/Tree.tsx#L358) (overriding [implicit `"list"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary)) and [`"treeitem"`](https://github.com/iTwin/iTwinUI/blob/3b5fe97caf19d4314e0f91645c9ec6b9cdb1c8d8/packages/itwinui-react/src/core/Tree/TreeNode.tsx#L256) (overriding [implicit `"listitem"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#technical_summary)). You can [read more about lists](https://htmhell.dev/adventcalendar/2023/13/#lessulgreater-and-lessolgreater) if you're curious.

Coincidentally, the `<ul>` is also the reason we can't attach a shadow-tree (see [list of allowed elements](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#elements_you_can_attach_a_shadow_to)). Changing `<ul>`+`<li>` to `<div>`s would solve this issue as well.

Ben-Pusey-Bentley: Okay, change the elements to `<div>`s and wrapped the inner div in a `ShadowRoot` in this [commit](https://github.com/iTwin/iTwinUI/pull/2091/commits/8629bd125abce0a4a10b0826f88d2f8ee4c0098c).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: I was eventually able to reproduce this issue, but only by not memoizing the `Table` `data`. Therefore, I don't think this issue reflects a problem in the `Table` component that needs to be fixed. Reproduction code:
```tsx
import { Button, Table } from '@itwin/itwinui-react';
import { Column } from '@itwin/itwinui-react/react-table';
import React from 'react';

const App = () => {
  const [reverseColumnOrder, setReverseColumnOrder] = React.useState(false);

  type TestDataType = {
    name: string;
    description: string;
    accessor?: string;
    subRows?: TestDataType[];
    booleanValue?: boolean;
  };

  const mockColumns: Column<TestDataType>[] = [
    {
      id: 'name',
      Header: 'Name',
      accessor: 'name',
      width: 90,
    },
    {
      id: 'description',
      Header: 'Description',
      accessor: 'description',
      maxWidth: 200,
    },
    {
      id: 'view',
      Header: 'View',
      Cell: () => {
        return <span>View</span>;
      },
    },
  ];

  const mockColumnsReverse = [...mockColumns].reverse();

  const mockedData: TestDataType[] = [...new Array(3)].map((_, index) => ({
    name: `Name${index + 1}`,
    description: `Description${index + 1}`,
  }));

  return (
    <>
      <Button
        onClick={() => {
          setReverseColumnOrder(!reverseColumnOrder);
        }}
      >
        Reverse Columns
      </Button>
      <Table
        columns={reverseColumnOrder ? mockColumnsReverse : mockColumns}
        data={mockedData}
        emptyTableContent='Empty table'
        emptyFilteredTableContent='No results. Clear filter.'
        enableColumnReordering
        isSortable
      />
    </>
  );
};

export default App;

```
Investigated the failing tests and found that they were making multiple calls to a `mockedData` function that were unnecessary. Removing those extra calls fixed the issues and caused the tests to pass. Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/26b4b6e0f06e24f4bca901599550a4661bba3ff0).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +820,36 @@ export const Table = <
...
+      return 62;
+    }, //Set to the height of the table row based on the value of the density prop.
+    scrollToIndex,
+  });
+
+  const outerProps = {
+    style: {
+      minBlockSize: virtualizer.getTotalSize(),
+      minInlineSize: '100%',
+      willChange: 'transform',
```
mayank99, line 841: I believe the `willChange: "transform"` can be removed from all three components. We are no longer using a `transform` on the wrapping div

mayank99: Also I noticed in the tanstack-virtual docs that they add [`contain: strict`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) to the outer div. It probably helps with performance so we should do the same

packages/itwinui-react/src/core/Table/Table.tsx
mayank99, line 1: Have you done a perf test on `ComboBox` and `Table` similar to you did for `Tree` with 4x-20x CPU slowdown?

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
mayank99, line 1: Why did the images change? Looks unintentional/unrelated to virtual-scroll changes

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Just did. `ComboBox` seems to have pretty similar performance overall. The tanstack virtualization takes a little longer to load, its only really noticeable at 20x CPU slowdown. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/02bade09-d6ad-4aee-8097-5bfe9330d427

`Table` by contrast has significantly worse performance at high CPU slowdown. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/8e455e94-f51f-4a42-b233-699fc2efb7b0

Looking at the console, I did find that there is a React keys error coming from `Table` now that previously wasn't there. 
image

I think this might be the cause of the performance issues and am currently working to solve it.

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
mayank99: This seems to be an issue elsewhere too. Worth investigating separately and submitting a separate PR.

packages/itwinui-react/src/utils/components/VirtualScroll.tsx
```diff
@@ -118,21 +131,65 @@ export type VirtualScrollProps = {
  *    This is my item #{index}
  *  </div>
  * ), [])
+ * <div style={{overflow: 'auto'}} ref={ref}>
  * <VirtualScroll
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: The `VirtualScroll` component is being removed in the [`Tree`](https://github.com/iTwin/iTwinUI/pull/2091) component's PR, so I don't think it makes much sense to update the docs for this component. I realized that the PR description was out of date, and referenced some things that were previously being done with this component that have since been changed. I have updated the PR description to mention that `VirtualScroll` is being removed.

packages/itwinui-react/src/utils/components/VirtualScroll.tsx
```diff
@@ -118,21 +131,65 @@ export type VirtualScrollProps = {
  *    This is my item #{index}
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Similar to this [comment](https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1661525933), I don't think it makes sense the update the JSDocs for a component that is being removed.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -70,8 +71,8 @@ import {
   onTableResizeEnd,
   onTableResizeStart,
 } from './actionHandlers/index.js';
-import { VirtualScroll } from '../../utils/components/VirtualScroll.js';
 import { SELECTION_CELL_ID } from './columns/index.js';
+import { type VirtualItem, Virtualizer } from '@tanstack/react-virtual';
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Fixed in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/0a3dffa7b54c050cb559093f4cd066dba3cb9d91).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +820,36 @@ export const Table = <
     }
   });
 
+  const virtualizer = useVirtualScroll({
+    count: page.length,
+    getScrollElement: () => tableElement ?? null,
+    estimateSize: () => {
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: `rowHeight` value added, which holds the if-condition list and is memoized, in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/489e64d737438b9490e1e67f48c1798504354606). `estimateSize` now simply returns this value.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +820,36 @@ export const Table = <
     }
   });
 
+  const virtualizer = useVirtualScroll({
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Added `getItemKey` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/d8145b0ef5d8b1be62d2e01ee2c0cb8fe50dc256).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -863,7 +897,11 @@ export const Table = <
   );
 
   const virtualizedItemRenderer = React.useCallback(
-    (index: number) => getPreparedRow(index),
+    (
+      index: number,
+      virtualItem?: VirtualItem,
+      virtualizer?: Virtualizer<Element, Element>,
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Removed in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/72e807de10ad5338282f8c9140a8d1081d508896).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -816,8 +820,36 @@ export const Table = <
...
+  });
+
+  const outerProps = {
+    style: {
+      minBlockSize: virtualizer.getTotalSize(),
+      minInlineSize: '100%',
+      contain: 'strict',
+      ...style,
+    },
+    ...rest,
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Removed in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/b3551341c9b407ee50088dedea82ff1ad877f2df).

.changeset/violet-rats-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Replaced old virtualization implementation with `@tanstack/react-virtual` for the `Table` component.
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated changeset in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/b834dd418966ac5c7f110b2da019f415249d93ec).

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
Ben-Pusey-Bentley: This issue for this PR turned out to be caused by the `border-radius: inherit` on `iui-table-row` not inheriting its styles from the `iui-table-body`, due to the added `ShadowRoot` between the two elements. Added some css to the `ShadowRoot` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/cb07805f8e93b5250783c1fac2cb68a8083e8bf5), which fixes the issue. Also reverted changes to the images, and removed mention of the changed images from the PR description.

packages/itwinui-react/src/core/Table/Table.tsx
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Moved the comment in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/5a15c0738c744295e4a360ca597d043bf21d47e6).

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
mayank99: Good finding! But I'm not sure about the solution. `:last-child` doesn't make sense inside the shadow-root.

There are at-most two shadow elements inside the table-body: the virtualization `<div>` and the `<slot>`. We would want both of them to inherit the `border-radius` in order for the table-row to inherit it correctly.

Unrelated: I noticed that the `border-radius` on the top left and right corners is not even working correctly. We should fix that (in a separate PR).

Generally, I'm leaning towards doing away with inheritance. We should just set the `border-radius` value directly, to make it less fragile.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99, line 1: When I visit the [With Paginator story](https://itwin.github.io/iTwinUI/2092/react/?story=table--with-paginator) in Firefox, the Table crashes upon scroll. There is an infinite re-render happening somewhere.

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
Ben-Pusey-Bentley: Rewrote the css block so the `<div>` and `<slot>` both inherit `border-radius` from above in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/c537efc1e1f9068b6c72fc3c9210807e3f3868df).

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: It might produce a bigger performance gain on the outer div? Lets add an after-PR TODO linking to this thread to investigate more.

For now, we can use `contain: strict` on the inner div in all three components.

Also, I think the variable names are confusing. You should rename them all to be something generic like `virtualWrapperProps`, since we only have a single div, not two. Alternatively, you could even inline the props directly instead of using an intermediate variable.

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Basic.png
mayank99: Nit: For multi-line strings, you should always use backticks instead of regular quotes. As a bonus, this will also allow prettier to format the CSS code properly.

```jsx
<ShadowRoot
  css={`
    div,
    slot {
      border-radius: inherit;
    }
  `}
>
```

Ben-Pusey-Bentley: Replaced the quotes with backticks in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/9a3e504301c63a1b4c2a989501526a6c148cae31). Also added the after PR TODO to the PR description.

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: I inlined the props directly to the `<div>` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/a95f14d684f51bcb33977b13457be9dc14782561). Also added the after PR TODO and moved the `contain: strict` style to the inner `<div>` on the `Tree` PR in this [commit](https://github.com/iTwin/iTwinUI/pull/2091/commits/a3a119fde05d570ad86aeb3c515186d244615398).

mayank99: Can you use the same approach everywhere? If you're inlining it in `Table`, then `ComboBox` and `Tree` should also inline it.

Also I noticed a mistake: the `style` prop should not be merged on L1166. That `style` prop is only meant for the root `Table` element (L950).

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: I wonder if properly memoizing the refs will help:

```jsx
ref={useMergedRefs(
  tableRef,
  resizeRef,
  setTableElement,
  React.useCallback((element) => {
    ownerDocument.current = element?.ownerDocument;
  }, []),
)}
```

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Removed the spread `style` prop in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/adbc5c3b76a0b00ca5aa3255f0f1416e2ae58fe7). Also inlined the styles for the `Tree` component in this [commit](https://github.com/iTwin/iTwinUI/pull/2091/commits/4d91fd1a8ea51f87dc7c27dbb0871bf205e530b0). The `ComboBox` was already inlining this [here](https://github.com/iTwin/iTwinUI/blob/74211fd2247e3a49e0269557e676edd11fe99e2a/packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx#L97), so that PR hasn't changed.

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Memoizing the refs on its own did not fix this issue. I did decide to retest passing in the `tableRef.current` into the `useVirtualScroll` hook, and that actually works. There must have been something wrong with how I was handling passing in the parent element back when I was trying to adapt the `VirtualScroll` component into this virtualization. This means that I can remove the state `tableElement`, and removing that seems to fix the rerender issues. Removed the `tableElement` state in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/c26637578de211b74b790f9ca309d2660cd49699) and removed mention of the `tableElement` state from the PR description.

mayank99: This is still unresolved, right?

mayank99, line 1: Not sure if it's related to the [most recent change](https://github.com/iTwin/iTwinUI/pull/2092/files#r1664763597), but now the ["Scroll to row" story](https://itwin.github.io/iTwinUI/2092/react/?story=table--scroll-to-row&theme=dark) isn't working properly. The target row is _just_ below the viewport.

It's possible that this might be because of a combination of:
- the Table header being in the same scroll container (thus offsetting the row by 1), and 
- the `scrollToIndex` function using `align: "auto"`.

In either case, we should fix this and add an e2e test to confirm it works.

Ben-Pusey-Bentley: Yeah, this is still unresolved. The keys error has been difficult to track down, but I don't think it is the cause of the performance issues anymore, since I was able to recreate it on the main branch when I pulled it and ran the workshop locally. I think the key issue might be related to the `ShadowRoot` that wraps the column headers. The removal of the `tableElement` state does seem to have improved performance, but it still is worse than the old implementation.

https://github.com/iTwin/iTwinUI/assets/141063311/402cde66-99e1-4b82-92fd-297207f719b1

Going to continue trying to resolve the missing key warning. I do want to ask, since the `Table` component has the density prop, does it make sense to potentially remove dynamic resizing? Is it possible for different rows on an iTwinUI `Table` to have different heights? Because removing the dynamic resize measuring could improve performance.

mayank99: > Is it possible for different rows on an iTwinUI `Table` to have different heights?

Yes, the row can be as tall as the user wants. And rows can have expandable content of varying heights. See also the "Virtualized sub-rows" story.

Ideally these things should be automatically handled by tanstack-virtual (the whole point of using the library). What exactly is it about dynamic sizing that is degrading the performance? You can still disable the dynamic resizing locally just to confirm that it is indeed the cause of bad performance.

Ben-Pusey-Bentley: Changed the scroll alignment to `center` and updated the preexisting e2e test for `scrollTo` to check that the row being scrolled to is visible in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/286b9f307563edb49ac5e9d6846c20e6ca98b31d). The `scrollPaddingStart` and `scrollPaddingEnd` parameters have some odd behavior, where adding `scrollPaddingStart` causes the virtualizer to scroll less and `scrollPaddingEnd` is what causes the virtualizer to scroll more, which seems backwards to me. Also some of the alignments don't seem to be affected by the padding, which I find strange. Auto alignment + `scrollPaddingEnd` (padding makes target row visible): 
image
Center alignment with no padding: 
image
Center alignment + `scrollPaddingEnd` (padding is ignored):
image
For now, I have gone with the center alignment with no padding, since that makes the row most clearly visible.

Ben-Pusey-Bentley: Tanstack requires us to pass a [`measureElement`](https://tanstack.com/virtual/latest/docs/framework/react/examples/dynamic) ref function into the items that are being virtualized in order to handle dynamic sizing. My logic was that not adding this ref function would improve performance, since that would be one less function running on render for the `TableRow`. However, this doesn't seem to help, since removing the ref doesn't improve performance substantially. video: 

https://github.com/iTwin/iTwinUI/assets/141063311/46632ed5-9fe2-4cb5-be73-5f3a9f04ceed

Since that isn't the source, I tried removing the optional `virtualItem` and `Virtualizer` parameter from the `TableRow` memoization, and that did seem to significantly improve performance, though it is still not quite as good as the old implementation. Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/7640f371-ecea-4d20-ac92-aa738403d4d5

 Removed those values from the memoization in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/da66fdf0f1316beebd0555e8618c8e1415cfc398).

mayank99: > I tried removing the optional `virtualItem` and `Virtualizer` parameter from the `TableRow` memoization, and that did seem to significantly improve performance

Is `virtualItem` and `virtualizer` properly memoizable? Can you inspect the source code of tanstack-virtual to see if the returned values are memoized? Additionally you can log it on every render to see if it's changing.

```tsx
function usePreviousLog<T>(value: T) {
  const previous = React.useRef<T>(undefined);

  React.useEffect(() => {
    if (previous.current !== value) {
      console.log("changed", value);
      previous.current = value;
    }
  });
}

usePreviousLog(virtualizer);
usePreviousLog(virtualItem);
```

If it's not properly memoizable, then we'll need to remove it from all dependency arrays.

Ben-Pusey-Bentley: I was able to find that various parts of the `virtualizer` have been memoized, including the [`getVirtualItems`](https://github.com/TanStack/virtual/blob/9f8921171e556e55ecbb00a5c95d405ae3539822/packages/virtual-core/src/index.ts#L813) function, but the `vitualItem`s themselves are not. See in the video bellow that the `virtualizer` only changes on a rows first render, while the `virtualItem`s change constantly as the user scrolls. 

https://github.com/iTwin/iTwinUI/assets/141063311/032e6264-50ef-4390-ad72-d9e736bb33a6

We can see in this second video that readding the `virtaulizer` to the `TableRow` memoization does not have a negative impact on performance. 

https://github.com/iTwin/iTwinUI/assets/141063311/5c20ca88-11f9-429c-9cef-77ff039d84f5

Therefore, I have readded the `virtualizer` to the memoization in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/29dc879194ece3dde0c621aabcfd703d60d3d579). Since all of the dependency arrays rely on `virtualizer` and not `virtualItem`, none of those will have to change. Also fixed the previously mentioned react keys issue in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/04b7a1b14a2eabc6027f1d920a24c233931fb812).

I did realize while working on this that virtualization doesn't seem to be a great fit for our `Table` component. The `TableRow`s are expensive to render, to the point where we are memoizing the rows on all of their props, but the virtualizer will for them to rerender anyways if users scroll away and scroll back. Maybe there is some way we can cache the rows so that we don't completely rerender rows that we have already rendered before.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1012,7 +1049,7 @@ export const Table = <
                           }
                         }}
                       >
-                        <ShadowRoot>
+                        <ShadowRoot key={`${columnProps.key}_shadow_root`}>
```
mayank99, line 1052: Are you sure you need to key all the children separately? Usually keys are only required for lists of repeated elements of the same type.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: > I did realize while working on this that virtualization doesn't seem to be a great fit for our `Table` component. The `TableRow`s are expensive to render, to the point where we are memoizing the rows on all of their props, but the virtualizer will for them to rerender anyways if users scroll away and scroll back. Maybe there is some way we can cache the rows so that we don't completely rerender rows that we have already rendered before.

Have you tried to investigate how long it really takes to render the rows when they scroll into view? One of the benefits of virtualization is that there is a very small number of rows, so I think that should counteract the fact that it needs to re-initialize when entering the view.

It's worth looking into this and #2114 together. As I mentioned before, you can use the Performance tab in Chrome dev-tools and the [long animation frames API](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Long_animation_frame_timing) to find out anything that takes too long. There is also some [layout thrashing](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing) seen in #2114 which we should try to avoid.

packages/itwinui-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -223,7 +223,5 @@ export const TableRowMemoized = React.memo(
       nextProp.state.sticky.isScrolledToLeft &&
     prevProp.state.sticky.isScrolledToRight ===
       nextProp.state.sticky.isScrolledToRight &&
-    prevProp.density === nextProp.density &&
-    prevProp.virtualItem === nextProp.virtualItem &&
```
mayank99: I think this is a bit problematic. We are accepting `virtualItem` as a prop but we aren't using it during memoization, which could result in a stale version of this prop being used.

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Yes, the children need to be keyed separately to resolve the react key warnings, since the children are rendered as siblings of the `ShadowRoot` component in the react component tree. 
image
I do wonder why react thinks this list needs to be keyed, though. The components are of different types, even if they are siblings.

packages/itwinui-react/src/core/Table/TableRowMemoized.tsx
Ben-Pusey-Bentley: I found that we were only using the `vitualItem.index` and `virtualItem.start` values in the `TableRow`, so I tried memoizing on those values specifically. I found that memoizing on these values does not cause unnecessary rerenders and does not negatively impact performance. I added these values to the memoization in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/d4b2e3cb4734afdb4225da19ed1e03c650f5ab6d).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: It depends on how many rows are being scrolled into view at once. I have had it take as long as 250 ms with no throttling. 
image
Looking closer at the event scroll scripts, there is a warning marker over one of the render tasks called "Recalculate Style"
image
image

It highlights this [ref function](https://github.com/iTwin/iTwinUI/blob/d4b2e3cb4734afdb4225da19ed1e03c650f5ab6d/packages/itwinui-react/src/core/Table/Table.tsx#L1029) as the reason for the layout thrashing. This function seems to be related to the column resizing, and will need to be replaced to resolve the layout thrashing.

Additionally, I did try reducing the `overscan` for the virtualized `Table`, since that would limit the ammount of rows that the `virtualizer` would render at once, and setting the value to `0` substantially improved the performance, making it very close to the old implementation's performance. 

https://github.com/iTwin/iTwinUI/assets/141063311/4d9e0302-c64b-449d-aa8d-a2c85ee1473f

I have set the `overscan` to `0` and updated e2e tests to account for this change in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/3f0e7476a481230198c923669f541728b1632fd0).

mayank99: > It highlights this [ref function](https://github.com/iTwin/iTwinUI/blob/d4b2e3cb4734afdb4225da19ed1e03c650f5ab6d/packages/itwinui-react/src/core/Table/Table.tsx#L1029) as the reason for the layout thrashing. This function seems to be related to the column resizing, and will need to be replaced to resolve the layout thrashing.

Oh wow. Can we memoize it in a `useCallback` for now? It is needlessly re-running on every render and continuously forcing layout (using `getBoundingClientRect`).

> I have set the `overscan` to `0` and updated e2e tests to account for this change in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/3f0e7476a481230198c923669f541728b1632fd0).

`0` could be problematic if the user reaches the edge of the Table and no other rows are loaded yet. Lets set it to at least `1`.

pnpm-lock.yaml
mayank99, line 1: Lockfile shouldn't have changed in this PR. Can you run `pnpm install` in the base PR (#2061) to ensure it is correct?

testing/e2e/app/routes/Table/route.tsx
```diff
@@ -13,6 +13,26 @@ export default function Resizing() {
...
+        arr[i] = {
+          index: i,
+          name: `Name${i}`,
+          description: `Description${i}`,
+          id: i,
+        };
+      }
+    }
+    return arr;
+  }, []);
```
mayank99, line 35: This is missing `oneRow` and `empty` from the dependency array

testing/e2e/app/routes/Table/route.tsx
```diff
@@ -121,14 +141,21 @@ export default function Resizing() {
...
         isRowDisabled={isRowDisabled}
         isSelectable={isSelectable}
         isSortable
         columnResizeMode={columnResizeMode as 'fit' | 'expand' | undefined}
         selectSubRows={selectSubRows}
+        enableVirtualization={enableVirtualization}
+        style={enableVirtualization ? { maxHeight: '90vh' } : undefined}
+        scrollToRow={
+          scroll
+            ? (rows, data) => rows.findIndex((row) => row.original === data[50])
```
mayank99, line 156: nit: It would be better if the number `50` was passed using the query params, so that the test case reads more naturally.

testing/e2e/app/routes/Table/spec.ts
```diff
@@ -379,3 +377,66 @@ test.describe('Table row selection', () => {
...
+    }); //Need to wait until the virtual rows are able to be rendered for the tests to work.
+
+    const rows = page.getByRole('rowgroup').getByRole('row');
+    const row50NameCell = page.getByText('Name50');
+    expect((await rows.all()).length).toBe(12);
+    await expect(rows.nth(0)).toContainText('Name44');
+    await expect(rows.nth(6)).toContainText('Name50');
+    await expect(rows.nth(11)).toContainText('Name55');
+
+    await expect(row50NameCell).toBeVisible();
```
mayank99, line 432: I believe [`toBeInViewport`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-in-viewport) is more appropriate here. `toBeVisible` would match even when the row is not actually scrolled to.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: It looks like using a fragment would avoid the need for separate keys.

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Wrapped the ref function in a `useCallback` and set the `overscan` to `1` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/ee85a78ad3c6ef08eedd78ae6008bb2890b73347). Chrome dev tools no longer show the layout thrashing warning on the "Recalculate Style" task. 
image

mayank99: The actual ref function is still not memoized. It will be recreated and re-run on every render.


I was thinking something like this:

```jsx
ref={React.useCallback((el) => {
  if (el) {
    columnRefs.current[column.id] = el;
    column.resizeWidth =
      el.getBoundingClientRect().width;
  }
}, [column])}
```

Ben-Pusey-Bentley: The React lint rules throw errors when I try to do this. 
image
It does seem to run well in dev though. Should I ignore the lint error?

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Wrapped the `ShadowRoot` and its children in fragment and removed their keys in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/de6dbda7a39f0048763514b26b0521d8b7a085a4).

pnpm-lock.yaml
Ben-Pusey-Bentley: Running `pnpm install` in the base PR didn't change the file at all, so I manually removed the changes on this branch and ran `pnpm install` here, and it seems to work as intended. Removed pnpm-lock file changes in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/63aeffbc68b0ad5dcb96dc6c3436de3c80439d3a).

testing/e2e/app/routes/Table/route.tsx
Ben-Pusey-Bentley: Added to the dependency array in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/2901d08b7970d1a6d2a40eb49b36be5badb79d44).

testing/e2e/app/routes/Table/spec.ts
Ben-Pusey-Bentley: Replaced with `toBeInViewport` in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/f97de88987b126fea5adcdae5ae2c736389711af).

testing/e2e/app/routes/Table/route.tsx
mayank99: i meant replacing the boolean param with a number param, but i guess we can have both

mayank99 (approved):
LGTM

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -991,11 +1039,7 @@ export const Table = <
                         key={columnProps.key}
                         title={undefined}
                         ref={(el) => {
-                          if (el) {
-                            columnRefs.current[column.id] = el;
-                            column.resizeWidth =
-                              el.getBoundingClientRect().width;
-                          }
+                          columnResizeRef(el, column);
                         }}
```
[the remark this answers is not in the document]
mayank99: Lets not do this. It changes the meaning and is not allowed in React 19.

packages/itwinui-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -44,6 +45,8 @@ export const TableRow = <T extends Record<string, unknown>>(props: {
   scrollContainerRef: HTMLDivElement | null;
   tableRowRef?: React.Ref<HTMLDivElement>;
   density?: 'default' | 'condensed' | 'extra-condensed';
+  virtualItem?: VirtualItem<Element>;
```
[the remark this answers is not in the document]
mayank99: Make sure to run `pnpm install`. The version was bumped in a recent commit.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -865,7 +896,11 @@ export const Table = <
   );
 
   const virtualizedItemRenderer = React.useCallback(
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2092/commits/5c97d0af5b7eee6f5b1490648842e9eb22e385f9).

packages/itwinui-react/src/core/Table/TableRowMemoized.tsx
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Talked about this on the `ComboBox` PR here: [discussion](https://github.com/iTwin/iTwinUI/pull/2061#discussion_r1672977686).

PR #2120, opened 2024-06-27
Fix `ComboBox` options' tag flicking

### LINKED ISSUE #2112: ComboBox: Selected options jittering

When `multiple` property is set to true and more options are selected than can fit inside the `ComboBox` input component, occasionally the selected options will start jittering by switching back and forth between displaying `n` and `n+1` options. 

The issue only happens under some sizes of the `ComboBox` and does not always replicate with the same settings on different machines or browsers. Seems like it occurs when the available space is almost enough to fit another tag in.

### Expected Behavior

Displayed selected option amount does not rapidly change.

### Link to minimal repro

https://stackblitz.com/edit/github-sfphuz-laltrk?file=src%2FApp.tsx

### Steps To Reproduce

1. Open sandbox.
2. You should see the selected options rapidly changing for some of the `ComboBox` components.
3. Toggling repeatedly between light / dark theme sometimes fixes / creates this issue.

### Anything else?

_No response_

r100-stack:
## Changes

Fixes #2112. As mentioned in https://github.com/iTwin/iTwinUI/issues/2112#issuecomment-2191857965, I believe the flickering is because of an infinite loop:

> `visibleCount` from our `useOverflow` internal hook changes on each resize of `iui-select-tag-container` and each resize changes the `visibleCount`.

I think this is because our `useOverflow` hook makes the assumption that all items are around the same size and thus we can take an *average*:

https://github.com/iTwin/iTwinUI/blob/4f431e59614ce50af39fc502765159efc7dd0b19/packages/itwinui-react/src/utils/hooks/useOverflow.tsx#L79

While this may work for cases like the TablePaginator, it doesn't seem to work for the ComboBox case where tags could have different widths.

This PR refactors `useOverflow` with a new algorithm that now takes the *actual* widths/heights into consideration.

<details><summary>New algorithm for useOverflow</summary>
<p>

Logic:
- Have a guess range for `visibleCount`. e.g. `[0, 32]` (32 is an arbitrary choice).
- Keep doubling the max guess until the container overflows.
  i.e. the max guess should always be `≥` the correct `visibleCount`.
  - With each such doubling, the new min guess is the current max guess (since underflow = we guessed low).
- Set `visibleCount` to the `maxGuess`.
- Repeat the following by calling `guessVisibleCount()` (keep re-rendering but not painting):
  - Each time the container overflows, new max guess is the average of the two guesses.
  - Each time the container does not overflow, new min guess is the average of the two guesses.
- Stop when the average of the two guesses is the min guess itself. i.e. no more averaging possible.
- The min guess is then the correct `visibleCount`.

Why this works:
* This approach takes the *real* widths/heights into account. So, the previous flickering is no longer there.
* Even if there are 10k items (for example), the maximum number of items we render starts from 32 and keeps doubling until we reach the first overflow. After the first overflow, the number of items we render always goes down. So, performance is not affected because we don't render any more items than needed. 
* Since this approach involves halving the guess in each render, I believe this approach is of the order of `log(n)`.

</p>
</details> 

While working on `useOverflow`, it seemed hard to get it to work by using `ref`s and also not knowing the children. So, this PR creates a new private component called `OverflowContainer` that calls `useOverflow` internally. `useOverflow` is this now no longer used by any other component.

OverflowContainer:
* We just pass it the `children`, `overflowTag` and `overflowTagPlacement` (e.g. overflow from start or end). It automatically places the `overflowTag` at the `overflowTagPlacement`.
* For custom cases (e.g. where overflow doesn't happen exactly in the start or end), a function could be passed (`(visibleCount) => React.ReactNode`) as the child of `OverflowContainer`. Now what to render based be customized based on the visibleCount. `itemLength` is required in this case.

---

I also tested that this bug does not regress some related prior bug fixes. Examples:
* [x] https://github.com/iTwin/iTwinUI/issues/1633#issuecomment-1839132399: Added an e2e test in `Select` and `ComboBox` for this case.

## Testing

* Since `useOverflow` now renders multiple times, having unit tests to mock rendered sizes is now unreasonable. So, I replaced all failing unit tests related to `useOverflow` with e2e tests.

<details>

Before:

https://github.com/iTwin/iTwinUI/assets/45748283/4d254f5b-f51c-45d5-8250-36b23c8cc7b0

After:

https://github.com/iTwin/iTwinUI/assets/45748283/2b50e71c-9950-4f0a-a43d-0ec0e5b0a1e7

</details>

## Docs

* Added JSDocs for the new algorithm, components, hooks, etc. to better explain this new algorithm.
* Added changesets.


mayank99 (commented):
Is it possible to split this into multiple [stacked](https://www.stacking.dev/) PRs? It's extremely difficult to understand/review it right now (unless you're looking for a quick LGTM)

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,363 @@
...
+        setVisibleCount(doubleOfMaxGuess);
+
+        return;
+      }
+
+      let newVisibleCountGuessRange = visibleCountGuessRange;
+
+      if (isOverflowing) {
+        // overflowing = we guessed too high. So, new max guess = half the current guess
+        newVisibleCountGuessRange = [visibleCountGuessRange[0], visibleCount];
```
Ben-Pusey-Bentley, line 295: I don't think that I understand how this part of the algorithm works. This is apparently halving the `visibleCountGuesssRange`, but I don't know how `visibleCount` is half of the max for the guess range. I must be missing something, since I am able to test the overflow container out with the deploy preview and see that it is working.

PR #2128, opened 2024-07-03
Add `resize` to `Breadcrumbs` and `ButtonGroup` overflow stories

r100-stack:
## Changes

To help me in testing `useOverflow` from #2120, I thought of adding `resize` to the overflow stories of components that use `useOverflow`.

Some additional improvements:
* Initial widths try to always start the story with an overflow. Because earlier, the width was something like `50%`. So, on larger screens, the story used to not overflow in the beginning.
* Added helper text suggesting resizing the container to see the overflow behavior.
* Moved the overflow container to decorators for cleaner story code.

## Testing

* Manually verified that the stories still look correct.
* Changed the `.test.ts` files to hide the helper text and then approved the failing images.

## Docs

N/A

### DIFF
--- a/.changeset/twelve-deers-laugh.md
+++ b/.changeset/twelve-deers-laugh.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Fixed an issue in `iui-dialog`, where long titles were not wrapping properly.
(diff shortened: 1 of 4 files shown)


apps/react-workshop/src/ButtonGroup.stories.tsx
```diff
@@ -60,53 +60,60 @@ export const Overflow = () => {
...
-        >
-          {items}
-        </ButtonGroup>
+        <Story />
       </div>
     </>
-  );
-};
+  ),
+] satisfies StoryDecorator[];
```
mayank99, line 116: nit: Can we not rely on ladle-specific APIs? It makes it harder to migrate to a different tool in the future. You can type the `Story` param as `Story: () => React.ReactNode`

mayank99 (approved):

Issue #783, opened 2022-10-17
Column manager aka Configure grid


DineshBodewar:
Provide a column manager modal dialog box which provide below features to manage table
A] Search by column name
B] Change column sequence using drag handle
C] Stick/Unstick columns - So user can always see sticked columns in table even in horizontal scroll
D] Show/Hide columns
E] Reset to product default settings

1] User will get configure icon next to search box on top of table like show below
![image](https://user-images.githubusercontent.com/115967995/196094766-27efb20d-f5df-4158-b4db-5f263871b916.png)

2] user click on it and he gets dropdown with options like below
![image](https://user-images.githubusercontent.com/115967995/196097529-21f851e3-6b75-4372-867f-bd7370e79338.png)

3] User clicks on "Configure grid" option and below modal dialog box will open.
For start we allow user to stick only 2 columns in table.

![image](https://user-images.githubusercontent.com/115967995/196101772-d092dba8-5dec-4c1c-af22-9e7de5c1ed8c.png)

Once the 2 column limit reached we will disable the stick(lock) icon for other columns and give a tooltip for the same as shown below.

![image](https://user-images.githubusercontent.com/115967995/196101834-d9dc310b-61f5-4a5a-8422-a2596f3ccda7.png)

![image](https://user-images.githubusercontent.com/115967995/196101878-a89e1620-8de8-418d-b1e2-fcd0d05f28a2.png)

![image](https://user-images.githubusercontent.com/115967995/196101937-95a2a93e-fbfb-4b81-b57a-d13be735a194.png)

FlyersPh9:
The option for a user to stick / unstick a column from within the column manager was raised by another UX team member today.

rohan-kadkol:
Technically, users can pass your custom ColumnManger type UIs using the `Header` property in the column object. For UI ideas, you can refer to a similar UI implementation [here](https://github.com/iTwin/iTwinUI/discussions/1672#discussioncomment-7491351).

While this is technically possible, some additional functionality from our side will likely be missing if users directly pass the `Header` property. E.g.:

https://github.com/iTwin/iTwinUI/blob/9182303b132ea7bf48439c1c24c0b5345b12ca10/packages/itwinui-react/src/core/Table/columns/actionColumn.tsx#L82-L90

So ideally, if there is still such a use case, we could consider easier ways for users to pass custom UIs and perform column manager type modifications easily.

@DineshBodewar Please give an update if this is still a valid use-case.

[...]

PR #2135, opened 2024-07-09
Fix markup in `Table`'s column manager

r100-stack:
## Changes

Fixes #2134. Upon investigating that issue, I noticed that there were a few issues in the markup. E.g.:
* `<input>` (checkbox) as a descendant of a `<button>`
* `<div>` not allowed as a child of `<button>` in this context

<details>
<summary>MenuItem markup in the column manager</summary>

image

</details>

Furthermore, @mayank99 added that even if each menu item was a div, no interactable elements are allowed in `role="menuitem"`.

As a result, we decided to replace the `DropdownMenu` + `MenuItem startIcon={<Checkbox>}` with a `Popover` + floating `<fieldset>` with a `<Checkbox label={…}>` for each column. Using a `Popover` also makes it easy to expand on column manager functionality to address https://github.com/iTwin/iTwinUI/issues/783.

Created a new `iui-table-column-manager` class mainly for padding and max-block-size styles. Like `iui-menu`, this also shows a maximum of 8.5 items (hardcoded without any calculation (https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1671383815)). The `0.5` part is to show that there are more items in the list.

Unrelated: Added `@default` to the JSDocs and a default value to the deconstructed object for the Surface's `isPadded` prop.

## Testing

* Fixed failing unit tests
* Approved failing test images

## Docs

Added changesets

## After PR TODOs

* [ ] Consider offering a way localize the "Column Manager" label ([#2135 (comment)](https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1670724463))
* [ ] Try preventing "`<SomeComponent as={Flex}>` not override the rendered element" ([#2135 (comment)](https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1672595007)). Related: [#2152 (comment)](https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1684836241)

### DIFF
--- a/packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
+++ b/packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
@@ -114,15 +114,15 @@ export const ActionColumn = <T extends Record<string, unknown>>({
         <Popover
           applyBackground
           content={
-            <FieldsetBase
-              as={Flex}
+            <Flex
+              as={FieldsetBase}
               className='iui-table-column-manager'
               flexDirection='column'
               alignItems='flex-start'
             >
               <VisuallyHidden as='legend'>Show/hide columns</VisuallyHidden>
               {headerCheckBoxes()}
-            </FieldsetBase>
+            </Flex>
           }
           {...popoverProps}
         >


packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -94,45 +102,41 @@ export const ActionColumn = <T extends Record<string, unknown>>({
...
           ? columnManager.dropdownMenuProps
           : {};
 
       return (
-        <DropdownMenu
-          {...dropdownMenuProps}
-          menuItems={headerCheckBoxes}
+        <Popover
+          content={
+            <Surface as={'fieldset'}>
```
mayank99, line 124: 1. There is no need for `Surface` here. You can use `<Popover applyBackground>`.
2. You are missing a `<legend>`. A fieldset without legend is useless.
3. We need to add some CSS to reset the default styles of `<fieldset>`.


Something like:

```jsx
<Popover
  applyBackground
  {...popoverProps}
  content={
    <fieldset>
      <VisuallyHidden as='legend'>Show/hide columns</VisuallyHidden>
      {headerCheckBoxes}
    </fieldset>
  }
>
```

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -94,45 +102,41 @@ export const ActionColumn = <T extends Record<string, unknown>>({
...
-                }
-                onClick={onClick}
+                checked={checked}
                 disabled={column.disableToggleVisibility}
-              >
-                <div id={`iui-column-${column.id}`}>
-                  {column.render('Header')}
-                </div>
-              </MenuItem>
+                onClick={(e) => e.stopPropagation()}
```
mayank99, line 109: This `onClick` can be removed.

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -94,45 +102,41 @@ export const ActionColumn = <T extends Record<string, unknown>>({
...
+                {headerCheckBoxes()}
+              </Flex>
+            </Surface>
+          }
+          {...popoverProps}
           onVisibleChange={(open) => {
             setIsOpen(open);
-            dropdownMenuProps?.onVisibleChange?.(open);
+            popoverProps?.onVisibleChange?.(open);
           }}
```
mayank99, line 134: I think we don't need controlled open state.

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -94,45 +102,41 @@ export const ActionColumn = <T extends Record<string, unknown>>({
...
         >
           <IconButton styleType='borderless' isActive={isOpen} ref={buttonRef}>
```
mayank99, line 136: We should add `label` here. And we don't need to set the `isActive` prop, since `Popover` now provides popover-open styling (#2076).

```jsx
<IconButton
  styleType='borderless'
  ref={buttonRef}
  label='Column manager'
>
```

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -5,24 +5,32 @@
...
+          /**
+           * ARIA role. Role of menu. For menu use 'menu', for select use 'listbox'.
+           * @default 'menu'
+           */
+          role?: string;
+        } & Pick<
+            Parameters<typeof usePopover>[0],
+            'visible' | 'onVisibleChange' | 'placement' | 'matchWidth'
+          > &
+          Pick<PortalProps, 'portal'>;
```
mayank99, line 33: I think we can just use `Popover` props with some of them omitted.

```ts
dropdownMenuProps: Omit<
  React.ComponentPropsWithoutRef<typeof Popover>,
  'children' | 'content'
>;
```

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
mayank99: Re: 2, the `<legend>` needs to be a direct child of `<fieldset>` for it to be properly associated. It shouldn't be inside `Flex`.

Re: 3, I don't think we need to maintain this complex CSS for sizing it to be exactly 8.5 items, especially since we want to add more things in this popover in the future. I would say lets just pick an arbitrary max-height and remove all those CSS variables.

Ben-Pusey-Bentley (commented):
Should [this example](https://github.com/iTwin/iTwinUI/blob/c70bfbf0888c67693628b79e1c30eb19b88e7d52/packages/itwinui-react/src/core/Table/columns/actionColumn.tsx#L40) be updated to use the `Popover` component instead of the `DropdonwMenu` component?

Ben-Pusey-Bentley (approved):

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
mayank99: I'm mostly ok with this, except `matchWidth` does nothing on `Popover` and will show up in the DOM props if set.

Even previously, `matchWidth` was totally unnecessary, I'm not sure why were exposing it.

mayank99: > So, are you suggesting no longer exposing `matchWidth` in `ActionColumn`'s `dropdownMenuProps`?

Yes lets remove it.

> Also, when setting `matchWidth={true}`, I didn't see any difference ([playground](https://stackblitz.com/edit/github-few12s?file=src%2FApp.tsx)).

It only does something when the menu-items are narrower than the trigger element.



Since this is an impossible case for column manager, we don't need this prop.

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -94,45 +96,45 @@ export const ActionColumn = <T extends Record<string, unknown>>({
...
-          onVisibleChange={(open) => {
-            setIsOpen(open);
-            dropdownMenuProps?.onVisibleChange?.(open);
-          }}
+        <Popover
+          applyBackground
+          content={
+            <Flex
+              as='fieldset'
+              className='iui-fieldset-base iui-table-column-manager'
```
mayank99, line 120: Lets use this:

```suggestion
            <FieldsetBase
              as={Flex}
              className='iui-table-column-manager'
```

or


```suggestion
            <Flex
              as={FieldsetBase}
              className='iui-table-column-manager'
```

The `iui-fieldset-base` class is an implementation detail of `FieldsetBase`. Of course right now `FieldsetBase` just sets the class, but in the future we might add more things to it (similar to `ButtonBase`) so it's good to use the component.

.changeset/violet-donkeys-help.md
mayank99, line 1: We could add a `patch` changeset mentioning that the "open" state color has changed.

mayank99: The other `minor` changeset about the `matchWidth` should also be `patch` (or removed entirely).

`minor` changesets are for big changes, lets not turn the release notes into something the user feels annoyed by. (e.g. The user might react as "Why are you telling me this? I dont care about this prop that did nothing previously.")

mayank99: If the user is overriding the `Header` like that, then there is no point even using `actionColumn`. We should try to be pragmatic.

mayank99 (approved):

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
mayank99: Actually this seems to not be working

mayank99 (approved):

packages/itwinui-react/src/core/Table/columns/actionColumn.tsx
mayank99: Side note: I believe this is an easy mistake to make and in the future we should do something to make `<SomeComponent as={Flex}>` not override the rendered element.

Issue #2015, opened 2024-04-24
Shift in `Table` component when resizing column with Expandable Row


Ben-Pusey-Bentley:
### Describe the bug (current behavior)

The `Table` component shifts to the the left by a few pixels when resizing the first resizable column on tables that have an Expandable Row. Video of issue: 

https://github.com/iTwin/iTwinUI/assets/141063311/40b0d193-bf51-4bc2-bd7c-46b68fab40d7

The cause of the issue seems to stem from a lack of responsiveness in the column resizer for columns that have the `SubRowExpander` component inside of them. Very small movements do not cause a resize at all, and when the resizing does start, size is initially taken away from the neighboring column, but not added to the column being resized. That is the cause of the shift, but it is not clear to me what is causing the column resizer to not respond to initial input.

### Expected Behavior

The column resizer should be immediately responsive and the table should not shift.

### Link to minimal repro

https://stackblitz.com/edit/github-lvepul?file=src%2FApp.tsx

### Steps To Reproduce

1. Resize the name column, and notice how the resizer follows the cursor exactly. This is the expected behavior.
2. Try to make very small movements with both resizers, and see that the index resizer does not immediately react.
3. Resize the index column, and notice the shift in the table and how the resizer lags behind the cursor.

Ben-Pusey-Bentley:
After some more investigation, it turns out that this issue can be reproduced without the expand button. The source of the issue stems from the width of the column being set to less than the min width for a column, which is 72px for normal columns and 108px for columns with the expand button. Reproduction [stackblitz](https://stackblitz.com/edit/github-nr8has?file=src%2FApp.tsx). Video: 

https://github.com/iTwin/iTwinUI/assets/141063311/16d57846-ffe3-4002-8dce-8079b2b62168


 Setting the inline styles `width` and `flex` values so that they are not less than the min width does not fix this issue. Seems that the resize logic is relying on the `column.width` value directly, which creates issues when our styles set the width to a different value.

PR #2138, opened 2024-07-12
Fixes bug with `Table` shifting on column resize

Ben-Pusey-Bentley:
## Changes

Adds check during the default `minWidth` being set for `Table` columns to set the `minWidth` value to the user provided `width` if `width` is less than `minWidth`. This change fixes https://github.com/iTwin/iTwinUI/issues/2015. The mentioned issue happens when `width` is set to a number less than `minWidth`, due to how the resizing logic assumes that the current column width, if it is a number, must be valid. This change only affects `width`s set to a number value, since string value `width`s do not affect the resizing logic.

## Testing

Tested in vite playground with this [code](https://github.com/iTwin/iTwinUI/blob/7e259906fb6b4833d117ea3bb94b25980295264a/playgrounds/vite/src/App.tsx#L1). Also added an e2e test for this issue.

## Docs

Added patch react changeset.


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1001,6 +1001,14 @@ export const Table = <
...
                         ? COLUMN_MIN_WIDTHS.withExpander
                         : COLUMN_MIN_WIDTHS.default;
+
+                      if (
+                        column.width &&
+                        typeof column.width === 'number' &&
+                        column.minWidth > column.width
+                      ) {
+                        column.width = column.minWidth;
+                      }
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: I could set the `minWidth` to the user provided `width` value instead. I've made the change in this [commit](https://github.com/iTwin/iTwinUI/pull/2138/commits/9bac52b5851f802ed38713ad2afd23cd00281d89).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1001,6 +1001,14 @@ export const Table = <
                       column.minWidth = columnHasExpanders
                         ? COLUMN_MIN_WIDTHS.withExpander
                         : COLUMN_MIN_WIDTHS.default;
+
+                      if (
+                        column.width &&
+                        typeof column.width === 'number' &&
+                        column.minWidth > column.width
+                      ) {
+                        column.minWidth = column.width;
```
mayank99, line 1010: Maybe I'm misunderstanding something, but wasn't the problem deemed to be within our resizing logic? Then why does this PR not touch the resizing logic at all? Iirc you said something along the lines of: our resize logic updates the `style` but does not update the actual `column.width`.

Ben-Pusey-Bentley: I think this is coming from a misunderstanding. The logic correctly updates both the `style` and the `column.width`, but the issue is that the logic breaks if the `width` is less than the `minWidth`, which causes the `column.width` to not update until the user resizes by enough to make up the difference. This is what causes the shift, since the actual width of the column itself doesn't change, but the resize logic still takes the difference from later columns on the table anyways. Since the issue is coming from the `column.width` value being less than the `column.minWidth`, it makes sense to prevent this from happening at initialization.

mayank99: It makes sense when you say "the resize logic still takes the difference from later columns on the table anyways", but can you point to where in code this is happening? I'm having trouble finding it.

Ben-Pusey-Bentley: It starts with the [`dispatchMove()`](https://github.com/iTwin/iTwinUI/blob/714b874246b458231b5ceddc8acd551c022b4743/packages/itwinui-react/src/core/Table/hooks/useResizeColumns.tsx#L101) function. This takes in the amount that the [mouse has moved on the x-axis](https://github.com/iTwin/iTwinUI/blob/714b874246b458231b5ceddc8acd551c022b4743/packages/itwinui-react/src/core/Table/hooks/useResizeColumns.tsx#L111) and passes it to the column resizing action. In this issue, this initially does nothing, since the resizing action recognizes that the new width that the first column is trying to be given is invalid. However, once the first columns new width becomes valid, the action is still observing how much the mouse has [moved to decide how much the later columns need to have taken from their widths](https://github.com/iTwin/iTwinUI/blob/714b874246b458231b5ceddc8acd551c022b4743/packages/itwinui-react/src/core/Table/hooks/useResizeColumns.tsx#L295). This causes a shift, since the first column's width isn't actually being updated, it is being "increased" to the size that it already was.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1001,6 +1001,14 @@ export const Table = <
                       column.minWidth = columnHasExpanders
                         ? COLUMN_MIN_WIDTHS.withExpander
                         : COLUMN_MIN_WIDTHS.default;
+
+                      if (
+                        column.width &&
```
mayank99, line 1006: very small suggestion: this `column.width` check can be removed, since it will be automatically covered by `typeof column.width === 'number'`

Ben-Pusey-Bentley: Removed in this [commit](https://github.com/iTwin/iTwinUI/pull/2138/commits/f7f8548bb8c105617130c109185382b8f8761126).

packages/itwinui-react/src/core/Table/Table.tsx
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Moved the first comment and added a comment to the new if block in this [commit](https://github.com/iTwin/iTwinUI/pull/2138/commits/df2c0990777b2fb6a67ea244ac3a90fab40552f4).

PR #2141, opened 2024-07-15
add development checks for `displayName`

mayank99:
## Changes

Follow-up to #2131, all `displayName`s are now conditionally added only during development.

While this shouldn't make a big difference in the bundle size, the code is only useful for React developer tools, and does add up over time when there are lots of components, so it's good to not have it in the prod build.

## Testing

Manually inspected the build outputs to verify that `.displayName` assignments are correctly present only in the dev build and removed from the prod build.

## Docs

Added patch changeset.

### DIFF
--- a/.changeset/proud-hornets-itch.md
+++ b/.changeset/proud-hornets-itch.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Component `displayName`s have been eliminiated from the prod build.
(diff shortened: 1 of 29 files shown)


Ben-Pusey-Bentley (commented):
A couple of components, [`Alert.Message`](https://github.com/iTwin/iTwinUI/blob/77d0d6cc959bb50c8366cac054aa186c27773f45/packages/itwinui-react/src/core/Alert/Alert.tsx#L134), and [`ListItem.Content`](https://github.com/iTwin/iTwinUI/blob/77d0d6cc959bb50c8366cac054aa186c27773f45/packages/itwinui-react/src/core/List/ListItem.tsx#L79) still have their `displayName` values set in the prod build output.

mayank99 (commented):
@Ben-Pusey-Bentley Good catch, thanks. Fixed in [`61bbc8b`](https://github.com/iTwin/iTwinUI/pull/2141/commits/61bbc8b1b52deda39f1b9148b6460c429f635099).

I did a global "find" in the `esm/` folder to verify there are no remaining instances.

Ben-Pusey-Bentley (approved):

PR #2143, opened 2024-07-15
Fixes border radius on top corners of `Table` component

Ben-Pusey-Bentley:
## Changes

This PR fixes an after PR TODO on this PR: #2092 Adds `border-radius` to the `.iui-table` class so that the border radius values are correctly applied to the `Table`s corners. Also removes the `border-radius` from the `.iui-table-header-wrapper` and `.iui-table` classes, and replaces the `border-radius: inherit` on the last `.iui-table-row` with a direct declaration. 
Before: 
image
After: 
image

After PR TODOs
- [ ]  Remove cuts from corners of expandable row when expanded. [discussion](https://github.com/iTwin/iTwinUI/pull/2143#discussion_r1687126744)

## Testing

Updated failing test images for the react visual tests.

## Docs

Added patch changesets for CSS and React packages.


packages/itwinui-css/src/table/base.scss
mayank99, line 1: I think you might have misunderstood the task, because this PR does not fix the issue and it also uses inheritance (which I explicitly asked to avoid in https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1664229041).

The issue is that the Table currently has square corners when it should have rounded corners. Even in the new story from this PR, you can see that the corner is a square edge.



For comparison, you can look at the [Table from v2](https://itwin.github.io/iTwinUI/legacy/v2/react/?path=/story/core-table--basic) which does have properly rounded corners.

Ben-Pusey-Bentley: Okay, I will move this PR to draft and rework it to work on the actual issue.

packages/itwinui-css/src/table/base.scss
```diff
@@ -39,6 +39,7 @@
   flex-shrink: 0;
   font-weight: var(--iui-font-weight-semibold);
   border-block-end: 1px solid var(--iui-color-border);
+  border-radius: inherit;
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Adding `border-radius` to `.iui-table` definitely seems to be the way to go, since the `.iui-table-header-wrapper` class doesn't have the full width of the table anymore.
The `.iui-table-header-wrapper` in iTwinUI v2:
image
The `.iui-table-header-wrapper` in iTwinUI v3:
image
The scrollbar seems to look okay with the curved corners on Firefox and Chrome.
Fireforx:
image
Chrome:
image

Border radius inheritance is a trickier issue here, since we do draw borders around the expandable sub content on `Table`s, and if those borders don't also have rounded corners, then they will slightly clip out of `Table` See the bottom left corner of this screenshot: 

image

It probably makes more sense to define the border radius on that corner directly, instead of relying on a chain of inheritance like we currently do, since that chain can be broken. It might make sense for the bottom right corner of this expandable content to have a square corner, though, since it isn't directly on the rounded corner of the `Table` like the bottom left corner is. What do you think? Should the bottom right corner of the expandable content be squared, or rounded?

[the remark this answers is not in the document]
Ben-Pusey-Bentley: I've added `border-radius` to the `.iui-table` class in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/8c863354915af873335a9d64779a851867aa3e19). Also updated changesets and the PR description to reflect this change.

packages/itwinui-css/src/table/base.scss
Ben-Pusey-Bentley: I have reworked the PR to add `border-radius` to the corners of the `Table` component and marked it as ready for review again.

packages/itwinui-css/src/table/base.scss
```diff
@@ -154,7 +153,8 @@
   display: flex;
 
   &:last-of-type {
-    border-radius: inherit;
+    border-end-start-radius: var(--iui-border-radius-1);
+    border-end-end-radius: var(--iui-border-radius-1);
```
mayank99, line 157: This is also getting applied to the table-header row, which is not desirable.

I believe we only want it for rows inside the table-body, and only when there is no paginator.

mayank99: Also there might be some remaining instances of `border-radius: inherit` that could be removed.

Ben-Pusey-Bentley: Updated the selectors so that the border radius is only applied in table-body when there is not paginator in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/452e3fe26f0daf7e83d6e5dad3d41f8a141a948f).

Trying to remove the last instance of `border-radius: inherit` from the file as well, and currently causing unexpected visual changes. The instance is on the first table cell of a row, currently trying to create a selector so that the border radius appears when we expect it to.

Ben-Pusey-Bentley: The last instance of `border-radius: inherit` has been correctly replaced in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/9089064cb8da806af525cef40b967c10bc36b16e). Also updated a react screenshot to remove a curved bottom left corner from the header row.

packages/itwinui-css/src/table/base.scss
```diff
@@ -397,9 +394,14 @@
   position: relative;
   word-break: break-word;
 
+  .iui-table-body:not(:has(+ .iui-table-paginator)) > .iui-table-row:last-of-type & {
+    &:first-of-type {
+      border-end-start-radius: var(--iui-border-radius-1);
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Yes, the cell also needs the `border-radius` style. Not having it causes the border to become straight along the cell. Video (watch the bottom left corner of the table): 

https://github.com/user-attachments/assets/a52170a9-59fc-4f98-a494-46c29650f4cd

The rows also need the styling on them, since the expandable content rows do not have cells inside of them, and rely on the `border-radius` provided by the rows.

packages/itwinui-css/src/table/base.scss
```diff
@@ -308,6 +303,13 @@
     }
     // #endregion Statuses & states
   }
+
+  .iui-table-body:not(:has(+ .iui-table-paginator)) & {
+    &:last-of-type {
+      border-end-start-radius: var(--iui-border-radius-1);
+      border-end-end-radius: var(--iui-border-radius-1);
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: I'm not sure what you mean. The `border-radius` on the row class is visible through the border provided to the expanded content. Video (watch the bottom corners of the table): 

https://github.com/user-attachments/assets/38e2bac3-10fe-42e9-ae81-063c9943f3f8

packages/itwinui-css/src/table/base.scss
mayank99: Could you explain this long selector? You lost me here 😅



Also there is another instance of `border-radius` left in the React code.

https://github.com/iTwin/iTwinUI/blob/e4b624be6e9e3baf08962c99a1c3c42a22400401/packages/itwinui-react/src/core/Table/Table.tsx#L1156-L1161

Ben-Pusey-Bentley: Okay, so after selecting `.iui-table-body`s where there is no `.iui-table-paginator` present, we select the last `.iui-table-row` child, and select the first `.iui-table-cell` child of the row to add the bottom left border radius. Added a comment to explain what the selector is doing in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/2cb214c19ad85e29a973a0e52c3d2583f1eb1c80).

Also removed the inherited `border-radius` from the React code in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/9d99d206a8b5d48ab5560a0617ab8d00ad79218b)

packages/itwinui-css/src/table/base.scss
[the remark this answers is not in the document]
Ben-Pusey-Bentley: I see. I have rounded that corner and updated the test images in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/50abbf8079e05a0a33746eb65dc92d4fefd35a9a).

packages/itwinui-css/src/table/base.scss
```diff
@@ -397,9 +394,19 @@
   position: relative;
   word-break: break-word;
 
+  //Adds the border-radius only to the first and last cells in the last row of the table body when the table has no paginator.
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Updated in this [commit](https://github.com/iTwin/iTwinUI/pull/2143/commits/d2b86a202bc996d9c2599cd39fff758757a26b2e).

apps/react-workshop/cypress-visual-screenshots/baseline/Table.test.ts-Expandable.png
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Added an after PR TODO for this to the PR description, since it doesn't seem to be related to `border-radius` as `border-radius` isn't defined for those corners.

mayank99 (approved):

PR #2150, opened 2024-07-18
`useOverflow` argument change: `items` → `itemsCount`

r100-stack:
## Changes

Pulled out of #2120 as PR 2 in the PR stack ([PR stack order](https://github.com/iTwin/iTwinUI/pull/2120#issuecomment-2386574990)). Didn't include this change in #2154 to prevent that PR from getting large and also because of https://github.com/iTwin/iTwinUI/pull/2150#pullrequestreview-2340986769.

In the `OverflowContainer` from later in the stack, it could be confusing why `items` are passed as `children` and as the `items` prop (@mayank99 pointed this out a while back). Thus, this PR renames `useOverflow()`'s `items` argument to `itemsCount`.

## Testing

Tests still passing.

## Docs

No changeset since not a user facing change.


mayank99 (approved):
Changes look good, but I feel like this PR should probably be earlier in the stack, since it does not really depend on the large code changes from #2154.

smmr-dn (approved):

mayank99 (approved):

Issue #894, opened 2022-10-17
Option to customize the display of selected items within ComboBox


FlyersPh9:
User is requesting the ability to display selected items in a different way that currently offered.  In the mockup it shows the selected items displaying within a tag container underneath the combobox.

![Selected items appear within a tag container underneath a combobox](https://user-images.githubusercontent.com/115967995/196093772-124521c0-9590-4dfc-8c13-73d276103396.png)

Request comes from step 3 of [this closed issue](https://github.com/iTwin/iTwinUI/issues/782).

S-Bra:
We are using a TagContainer with a ComboBox for editing a given set of tags. This is very close the original request. 

Our version adds the ability to use the input of the ComboBox to add new items if they not exist - just type the new item and press enter. Here is the Stackblitz for it: https://stackblitz.com/edit/github-21dait-t6c5zo?file=src%2FApp.tsx

FlyersPh9:
iTwinUI is now in maintenance mode per [the support policy](https://github.com/iTwin/iTwinUI/wiki/Support-policy), with [StrataKit](https://github.com/iTwin/stratakit) replacing it as the primary design system.

PR #2151, opened 2024-07-18
e2e tests for `useOverflow` logic

r100-stack:
## Changes

Pulled out of #2120 as PR 1 in the PR stack ([PR stack order](https://github.com/iTwin/iTwinUI/pull/2120#issuecomment-2386574990)).

To make that PR easy to review, according to @mayank99's suggestion, I created this PR to only add new e2e tests that test the existing `useOverflow` logic. These tests will thus act as a baseline that the new `useOverflow` logic will also have to pass to make sure that there is no change in the resulting behavior.

Some e2e tests (e.g. `MiddleTextTrucation`) seem to be failing only on GitHub (e.g. [action run](https://github.com/iTwin/iTwinUI/actions/runs/9996764547/job/27631858596?pr=2151#step:7:339)) but not locally. I guess this is a small testing environment specific issue that I am currently trying to fix. Thus, the PR is ready for review despite the failing tests.

TODO:
* [x] Fix the flaky/unreliable e2e tests.

## Testing

* Ensured that the new e2e tests are passing.

## Docs

N/A

## After PR TODOs

- [ ] Consider a lint rule for `await`ing the `expect`s in e2e tests ([#2151 (comment)](https://github.com/iTwin/iTwinUI/pull/2151#discussion_r1684484281))
- [ ] Don't rely on implementation details when it's possible to pass props to the tag container/tags in ComboBox ([#2151 (comment)](https://github.com/iTwin/iTwinUI/pull/2151#discussion_r1684394649)) and Select ([#2151 (comment)](https://github.com/iTwin/iTwinUI/pull/2151#discussion_r1684395240)).

### DIFF
--- a/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
+++ b/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
@@ -9,79 +9,63 @@ import {
   LineClamp,
   SvgSortDown,
   SvgSortUp,
+  useMergedRefs,
+  type PolymorphicForwardRefComponent,
 } from '../../utils/index.js';
 import type {
-  ColumnInstance,
   HeaderGroup,
   TableKeyedProps,
-  TableState,
 } from '../../react-table/react-table.js';
-import { SELECTION_CELL_ID } from './columns/index.js';
 import { FilterToggle } from './filters/FilterToggle.js';
-import { getCellStyle, getSubRowStyle, getStickyStyle } from './utils.js';
+import {
+  getCellStyle,
+  getSubRowStyle,
+  getStickyStyle,
+  TableInstanceContext,
+} from './utils.js';
 import cx from 'classnames';
 
-type ColumnHeaderProps<
-  T extends Record<string, unknown> = Record<string, unknown>,
-> = TableKeyedProps & {
-  columnRefs: React.MutableRefObject<Record<string, HTMLDivElement>>;
-  column: HeaderGroup<T>;
-  index: number;
+type ColumnHeaderProps = TableKeyedProps & {
+  column: HeaderGroup<Record<string, unknown>>;
   areFiltersSet: boolean;
-  hasAnySubRows: boolean;
-  headers: HeaderGroup<T>[];
-  state: TableState<T>;
-  data: T[];
   isResizable: boolean;
   columnResizeMode: 'fit' | 'expand';
   enableColumnReordering: boolean;
   density: string | undefined;
-  visibleColumns: ColumnInstance<T>[];
+  columnHasExpanders: boolean;
+  isLast: boolean;
+  isTableEmpty: boolean;
 };
 
-export const ColumnHeader = <
-  T extends Record<string, unknown> = Record<string, unknown>,
->(
-  props: ColumnHeaderProps<T>,
-): JSX.Element => {
+export const ColumnHeader = React.forwardRef((props, forwardedRef) => {
   const {
-    columnRefs,
     column,
-    index,
     areFiltersSet,
-    hasAnySubRows,
-    headers,
-    state,
-    data,
     isResizable,
     columnResizeMode,
     enableColumnReordering,
     density,
-    visibleColumns,
+    columnHasExpanders,
+    isLast,
+    isTableEmpty,
     ...rest
   } = props;
 
   const isHeaderDirectClick = React.useRef(false);
+  const instance = React.useContext(TableInstanceContext);
 
   const COLUMN_MIN_WIDTHS = {
     default: 72,
     withExpander: 108, // expander column should be wider to accommodate the expander icon
   };
 
-  const showFilterButton = (column: HeaderGroup<T>) =>
-    (data.length !== 0 || areFiltersSet) && column.canFilter && !!column.Filter;
+  const showFilterButton = (column: HeaderGroup<Record<string, unknown>>) =>
+    (!isTableEmpty || areFiltersSet) && column.canFilter && !!column.Filter;
 
-  const showSortButton = (column: HeaderGroup<T>) =>
-    data.length !== 0 && column.canSort;
+  const showSortButton = (column: HeaderGroup<Record<string, unknown>>) =>
+    !isTableEmpty && column.canSort;
 
   const { onClick, ...restSortProps } = column.getSortByToggleProps();
-
-  const columnHasExpanders =
-    hasAnySubRows &&
-    index ===
-      headers.findIndex(
-        (c) => c.id !== SELECTION_CELL_ID, // first non-selection column is the expander column
-      );
 
   if ([undefined, 0].includes(column.minWidth)) {
     // override "undefined" or zero min-width with default value
@@ -107,9 +91,9 @@ export const ColumnHeader = <
       column.columnClassName,
     ),
     style: {
-      ...getCellStyle(column, !!state.isTableResizing),
+      ...getCellStyle(column, !!instance?.state.isTableResizing),
       ...(columnHasExpanders && getSubRowStyle({ density })),
-      ...getStickyStyle(column, visibleColumns),
+      ...getStickyStyle(column, instance?.visibleColumns ?? []),
       flexWrap: 'wrap',
       columnGap: 'var(--iui-size-xs)',
     },
@@ -121,14 +105,16 @@ export const ColumnHeader = <
       {...rest}
       key={columnProps.key}
       title={undefined}
-      ref={React.useCallback(
-        (el: HTMLDivElement) => {
-          if (el) {
-            columnRefs.current[column.id] = el;
-            column.resizeWidth = el.getBoundingClientRect().width;
-          }
-        },
-        [column, columnRefs],
+      ref={useMergedRefs(
+        React.useCallback(
+          (el?: HTMLDivElement) => {
+            if (el) {
+              column.resizeWidth = el.getBoundingClientRect().width;
+            }
+          },
+          [column],
+        ),
+        forwardedRef,
       )}
       onMouseDown={() => {
         isHeaderDirectClick.current = true;
@@ -183,7 +169,7 @@ export const ColumnHeader = <
         )}
         {isResizable &&
           column.isResizerVisible &&
-          (index !== headers.length - 1 || columnResizeMode === 'expand') && (
+          (!isLast || columnResizeMode === 'expand') && (
             <Box
               {...column.getResizerProps()}
               className='iui-table-resizer'
@@ -195,13 +181,15 @@ export const ColumnHeader = <
         {enableColumnReordering && !column.disableReordering && (
           <Box className='iui-table-reorder-bar' slot='resizers' />
         )}
-        {column.sticky === 'left' && state.sticky.isScrolledToRight && (
-          <Box className='iui-table-cell-shadow-right' slot='shadows' />
-        )}
-        {column.sticky === 'right' && state.sticky.isScrolledToLeft && (
-          <Box className='iui-table-cell-shadow-left' slot='shadows' />
-        )}
+        {column.sticky === 'left' &&
+          instance?.state.sticky.isScrolledToRight && (
+            <Box className='iui-table-cell-shadow-right' slot='shadows' />
+          )}
+        {column.sticky === 'right' &&
+          instance?.state.sticky.isScrolledToLeft && (
+            <Box className='iui-table-cell-shadow-left' slot='shadows' />
+          )}
       </>
     </Box>
   );
-};
+}) as PolymorphicForwardRefComponent<'div', ColumnHeaderProps>;
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -25,7 +25,6 @@ import type {
   ActionType,
   TableInstance,
   Column,
-  ColumnInstance,
 } from '../../react-table/react-table.js';
 import { ProgressRadial } from '../ProgressIndicators/ProgressRadial.js';
 import {
@@ -40,7 +39,7 @@ import {
   useVirtualScroll,
 } from '../../utils/index.js';
 import type { CommonProps } from '../../utils/index.js';
-import { TableColumnsContext } from './utils.js';
+import { TableInstanceContext } from './utils.js';
 import { TableRowMemoized } from './TableRowMemoized.js';
 import type { TableFilterValue } from './filters/index.js';
 import { customFilterFunctions } from './filters/customFilterFunctions.js';
@@ -639,7 +638,6 @@ export const Table = <
     gotoPage,
     setPageSize,
     flatHeaders,
-    visibleColumns,
     setGlobalFilter,
   } = instance;
 
@@ -950,7 +948,9 @@ export const Table = <
   }, []);
 
   return (
-    <TableColumnsContext.Provider value={instance.columns as ColumnInstance[]}>
+    <TableInstanceContext.Provider
+      value={instance as TableInstance<Record<string, unknown>>}
+    >
       <Box
         ref={useMergedRefs<HTMLDivElement>(
           tableRef,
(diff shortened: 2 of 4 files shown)


mayank99 (commented):
Are you also planning to delete the existing unit tests? Because otherwise it begs the question: why even add the e2e tests?

testing/e2e/app/routes/Breadcrumbs/spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    // should restore hidden items when space is available again
+    await setContainerSize(undefined);
+
+    await expectOverflowState({
+      expectedItemLength: 5,
+      expectedOverflowButtonVisibleCount: undefined,
+    });
+  });
+
+  test(`should at minimum always show one overflow tag and one item`, async ({
```
mayank99, line 33: nit: "overflow button",  not "overflow tag"

testing/e2e/app/routes/Breadcrumbs/spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    await expectOverflowState({
+      expectedItemLength: 1,
+      expectedOverflowButtonVisibleCount: 1,
+    });
+  });
+});
+
+// ----------------------------------------------------------------------------
+
+const getSetContainerSize = (page: Page) => {
```
mayank99, line 57: optional: one layer of abstraction could be removed by simply creating a `setContainerSize` that accepts both the `page` object and the actual value.

same comment applies to `expectOverflowState` and any other similar utils in other test scenarios

testing/e2e/app/routes/Breadcrumbs/spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  return async (dimension: string | undefined) => {
+    await page.getByTestId('container').evaluate(
+      (element, args) => {
+        if (args.dimension != null) {
+          element.style.setProperty('width', args.dimension);
+        } else {
+          element.style.removeProperty('width');
+        }
+      },
+      { dimension },
```
mayank99, line 67: nit: `dimension` makes it sound like the possible values would be `horizontal` or `vertical`. It might be better to name this `size` or even just `value`

testing/e2e/app/routes/ButtonGroup/route.tsx
```diff
@@ -1,23 +1,90 @@
-import { ButtonGroup, IconButton } from '@itwin/itwinui-react';
+import { Button, ButtonGroup, Flex, IconButton } from '@itwin/itwinui-react';
 import { SvgPlaceholder } from '@itwin/itwinui-icons-react';
 import { useSearchParams } from '@remix-run/react';
+import React from 'react';
```
mayank99, line 4: nit: always use `* as React` namespace import, not default import

same comment applies to all other instances

testing/e2e/app/routes/ButtonGroup/route.tsx
```diff
@@ -1,23 +1,90 @@
...
-  return (
-    <ButtonGroup role='toolbar' orientation={orientation as any}>
-      <IconButton label='Button 1'>
-        <SvgPlaceholder />
-      </IconButton>
-      <IconButton label='Button 2' isActive>
-        <SvgPlaceholder />
-      </IconButton>
-      <IconButton disabled label='Button 3'>
+  const Default = () => {
```
mayank99, line 20: components should never be created in a nested way like this, or they will be unnecessarily recreated on every render and might cause some issues with React. If data needs to be shared from the parent component, use props.

testing/e2e/app/routes/DropdownMenu/spec.ts
mayank99, line 1: this file shouldn't have changed?

testing/e2e/app/routes/ComboBox/route.tsx
```diff
@@ -1,24 +1,51 @@
...
-      id='test-component'
-      enableVirtualization={virtualization}
-    />
+    <div data-testid='container'>
+      <Button
+        data-testid='change-value-to-first-option-button'
+        onClick={() => setValue(multiple ? [0] : 0)}
+      >
+        Change value to just the first option
+      </Button>
```
mayank99, line 40: all the code for new tests should be placed in a separate component and conditionally rendered by the default component

in other words: adding new tests should not affect existing tests where possible. right now, you need to change the existing tests to account for the extra button

testing/e2e/app/routes/ComboBox/spec.ts
```diff
@@ -1,181 +1,382 @@
...
-
-    await page.keyboard.press('Tab');
-    const comboBoxInput = page.locator('#test-component').locator('input');
-    const comboBoxMenu = page.getByRole('listbox');
-    await expect(comboBoxInput).toHaveAttribute(
-      'aria-controls',
-      'test-component-list',
-    );
-    await expect(comboBoxMenu).toBeVisible();
+  test.describe('General', () => {
```
mayank99, line 14: nit: do not nest `describe` blocks; instead create sibling `describe` blocks at the top-level. The existing `describe` block could be renamed to something like "ComboBox (general)" if you want.

same comment applies to other tests that use this pattern

testing/e2e/app/routes/MiddleTextTruncation/spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    await setContainerSize('200px');
+
+    await expect(middleTextTruncation.first()).toHaveText(
+      'MyFileWithAReallyLon…2.html',
+    );
+
+    await setContainerSize(undefined);
+
+    // should restore hidden items when space is available again
+    expect(middleTextTruncation.first()).toHaveText(longItem);
```
mayank99, line 26: don't forget to `await`

testing/e2e/app/routes/Select/route.tsx
```diff
@@ -0,0 +1,47 @@
...
+   */
+  const searchParamValue = searchParams.get('value') as
+    | ('all' & string & {})
+    | null;
+  const value =
+    searchParamValue != null
+      ? searchParamValue === 'all'
+        ? options.map((option) => option.value)
+        : (JSON.parse(searchParamValue) as number | number[])
+      : undefined;
```
mayank99, line 36: nit: never write nested ternaries; regular `if` conditions are always more readable

testing/e2e/app/routes/ComboBox/spec.ts
```diff
@@ -1,181 +1,382 @@
...
 
-    //stay on first
-    await page.keyboard.press('ArrowUp');
-    await expect(comboBoxInput).toHaveAttribute(
-      'aria-activedescendant',
-      'test-component-option-Item-0',
-    );
-    await expect(items.first()).toHaveAttribute('data-iui-focused', 'true');
+      const tags = await page
+        .locator('#test-component-selected-live > span')
```
mayank99, line 26: this locator is relying on an implementation detail which could change at any time?

testing/e2e/app/routes/Select/spec.ts
```diff
@@ -0,0 +1,143 @@
...
+const getExpectOverflowState = (page: Page) => {
+  return async ({
+    expectedItemLength,
+    expectedLastTagTextContent,
+  }: {
+    expectedItemLength: number;
+    expectedLastTagTextContent: string | undefined;
+  }) => {
+    const tags = await page
+      .locator('div[class$="-select-tag-container"] > span')
```
mayank99, line 131: this locator is also relying on an implementation detail?

mayank99 (approved):
Some minor comments to improve tests and reduce diff.

Approving on the assumption that you can make CI pass.

testing/e2e/app/routes/ButtonGroup/route.tsx
mayank99: we have had this rule for a while now, it's in the style guide

the default `import React` is considered legacy at this point, it's from before hooks and before the new JSX transform was introduced. i believe the React team wants to remove it in a future version, so they encourage namespace (`import * as React`) or named imports (`import {} from`)

Ben-Pusey-Bentley (approved):
Approving so that this isn't held up. Added some comments that I hope will help with the failing e2e tests on the CI.

testing/e2e/app/routes/ComboBox/spec.ts
mayank99: The `select-tag` is still coming from an internal class which can change at any time.

Right now we don't have a way to pass classes/attributes to the tag container or the tags (#894). This test could probably be rewritten to avoid relying on such implementation details (e.g. by only comparing the text inside the main container), but I'll leave it upto you if you want to rewrite it or add a TODO to wait until we allow customizing the tag container.

testing/e2e/app/routes/Select/spec.ts
mayank99: Same as https://github.com/iTwin/iTwinUI/pull/2151#discussion_r1684394649, this is still relying on an implementation detail.

testing/e2e/app/routes/ButtonGroup/route.tsx
mayank99: i noticed it in some other tests too (e.g. Table)

testing/e2e/app/routes/MiddleTextTruncation/spec.ts
mayank99: i noticed a few more `expect`s that were missing `await`s.

we might need a lint rule to help avoid this situation

testing/e2e/app/routes/ButtonGroup/route.tsx
mayank99: this one is a pretty gnarly mistake imo, it should be fixed before merging

mayank99: you can copy-paste this if you'd like:

<details><summary>route.tsx</summary>

```tsx
import {
  Table,
  tableFilters,
  TablePaginator,
  TablePaginatorRendererProps,
import { useSearchParams } from '@remix-run/react';

export default function TableTest() {
  const { exampleType } = useConfig();

  return exampleType === 'withTablePaginator' ? (
    <WithTablePaginator />
  ) : (
    <Default />
  );
}

function useConfig() {
  const [searchParams] = useSearchParams();

  const exampleType = searchParams.get('exampleType') || 'default';
  const disableResizing = searchParams.get('disableResizing') === 'true';
  const columnResizeMode = searchParams.get('columnResizeMode') || 'fit';
  const maxWidths = searchParams.getAll('maxWidth');
  const minWidths = searchParams.getAll('minWidth');
  const density = (searchParams.get('density') || undefined) as
    | 'default'
    | 'condensed'
    | 'extra-condensed'
    | undefined;
  const isSelectable = searchParams.get('isSelectable') === 'true';
  const subRows = searchParams.get('subRows') === 'true';
  const filter = searchParams.get('filter') === 'true';
  const selectSubRows = !(searchParams.get('selectSubRows') === 'false');
  const enableVirtualization = searchParams.get('virtualization') === 'true';
  const empty = searchParams.get('empty') === 'true';
  const scroll = searchParams.get('scroll') === 'true';
  const oneRow = searchParams.get('oneRow') === 'true';
  const stateReducer = searchParams.get('stateReducer') === 'true';
  const scrollRow = Number(searchParams.get('scrollRow'));

  return {
    exampleType,
    disableResizing,
    columnResizeMode,
    maxWidths,
    minWidths,
    density,
    isSelectable,
    subRows,
    filter,
    selectSubRows,
    enableVirtualization,
    empty,
    scroll,
    oneRow,
    stateReducer,
    scrollRow,
  };
}

function Default() {
  const {
    disableResizing,
    columnResizeMode,
    maxWidths,
    minWidths,
    density,
    isSelectable,
    subRows,
    filter,
    selectSubRows,
    enableVirtualization,
    empty,
    scroll,
    oneRow,
    stateReducer,
    scrollRow,
  } = useConfig();

  const virtualizedData = React.useMemo(() => {
    const size = oneRow ? 1 : 100000;
    const arr = new Array(size);
    if (!empty) {
      for (let i = 0; i < size; ++i) {
        arr[i] = {
          index: i,
          name: `Name${i}`,
          description: `Description${i}`,
          id: i,
        };
      }
    }
    return arr;
  }, [oneRow, empty]);

  const data = subRows
    ? [
        {
          index: 1,
          name: 'Name1',
          description: 'Description1',
          id: '111',
        },
        {
          index: 2,
          name: 'Name2',
          description: 'Description2',
          subRows: [
            {
              index: 2.1,
              name: 'Name2.1',
              description: 'Description2.1',
              id: '223',
            },
            {
              index: 2.2,
              name: 'Name2.2',
              description: 'Description2.2',
              id: '224',
            },
          ],
          id: '222',
        },
        {
          index: 3,
          name: 'Name3',
          description: 'Description3',
          subRows: [
            {
              index: 3.1,
              name: 'Name3.1',
              description: 'Description3.1',
              id: '334',
            },
            {
              index: 3.2,
              name: 'Name3.2',
              description: 'Description3.2',
              id: '335',
            },
          ],
          id: '333',
        },
      ]
    : [
        {
          index: 1,
          name: 'Name1',
          description: 'Description1',
          id: '111',
        },
        {
          index: 2,
          name: 'Name2',
          description: 'Description2',
          id: '222',
        },
        {
          index: 3,
          name: 'Name3',
          description: 'Description3',
          id: '333',
        },
      ];

  const isRowDisabled = React.useCallback(
    (rowData: Record<string, unknown>) => {
      return rowData.name === 'Name3.2';
    },
    [],
  );

  return (
    <>
      <Table
        columns={[
          {
            Header: '#',
            accessor: 'index',
            width: 100,
            maxWidth: parseInt(maxWidths[0]) || undefined,
            minWidth: parseInt(minWidths[0]) || undefined,
          },
          {
            Header: 'Name',
            accessor: 'name',
            maxWidth: parseInt(maxWidths[1]) || undefined,
            minWidth: parseInt(minWidths[1]) || undefined,
            disableResizing,
            Filter: filter ? tableFilters.TextFilter() : undefined,
          },
          {
            Header: 'Description',
            accessor: 'description',
            width: '200px',
          },
          {
            Header: 'ID',
            accessor: 'id',
            width: '8rem',
          },
        ]}
        data={enableVirtualization ? virtualizedData : data}
        emptyTableContent='No data.'
        isResizable
        isRowDisabled={isRowDisabled}
        isSelectable={isSelectable}
        isSortable
        density={density}
        columnResizeMode={columnResizeMode as 'fit' | 'expand' | undefined}
        selectSubRows={selectSubRows}
        enableVirtualization={enableVirtualization}
        style={enableVirtualization ? { maxHeight: '90vh' } : undefined}
        scrollToRow={
          scroll
            ? (rows, data) =>
                rows.findIndex((row) => row.original === data[scrollRow])
            : undefined
        }
        stateReducer={
          stateReducer
            ? (newState, action, previousState, instance) => {
                if (action.type === 'toggleRowSelected') {
                  console.log(action.value);
                }
                return newState;
              }
            : undefined
        }
      />
    </>
  );
}

function WithTablePaginator() {
  const { density } = useConfig();

    () => [
      {
        id: 'name',
        Header: 'Name',
        accessor: 'name',
        Filter: tableFilters.TextFilter(),
      },
      {
        id: 'description',
        Header: 'Description',
        accessor: 'description',
        maxWidth: 200,
        Filter: tableFilters.TextFilter(),
      },
    ],
    [],
  );

  const data = React.useMemo(
    () =>
      Array(505)
        .fill(null)
        .map((_, index) => ({
          name: `Name ${index}`,
          description: `Description ${index}`,
        })),
    [],
  );

  const paginator = React.useCallback(
    (props: TablePaginatorRendererProps) => (
      <TablePaginator id='paginator' {...props} />
    ),
    [],
  );

  return (
    <>
      <div id='container' style={{ height: '80vh' }}>
        <Table
          style={{ height: '100%' }}
          emptyTableContent='No data.'
          columns={columns}
          data={data}
          pageSize={50}
          density={density}
          paginatorRenderer={paginator}
        />
      </div>
    </>
  );
}
```

</details>

mayank99 (approved):

PR #2152, opened 2024-07-18
New `OverflowContainer` wrapper around `useOverflow`

r100-stack:
## Changes

Pulled out of #2120 as PR 3 in the PR stack ([PR stack order](https://github.com/iTwin/iTwinUI/pull/2120#issuecomment-2386574990)).

This just adds the `OverflowContainer` component from the combined PR but continues to use the existing `useOverflow` algorithm. Additionally, all components now use `OverflowContainer` instead of `useOverflow`.

* `OverflowContainer`: Takes the `items` prop (would likely be removed in later PRs in the PR stack), passes it to `useOverflow()`, and put the returned `visibleItems` along with the `items.length` in a context.
* `useOverflowContainerContext()`: Can be used in components under `OverflowContainer` to access the context values in OverflowContainer. ([#2152 (comment)](https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1737238710))
* `OverflowContainer.OverflowNode`: Wrap overflow content in this component to conditionally render it when overflowing.

I also had to make a small css change to make the `iui-breadcrumbs-list` stretch to the width of `iui-breadcrumbs`. This was needed to constrain the width and thus trigger an overflow when the width of the container is smaller than the required width.

## Testing

- All tests continue to pass with some minimal changes

## Docs

- Added pertinent JSDocs.
  - Since `OverflowContainer`'s API will change, the JSDocs are currently concise. They will be expanded upon in later PRs in this PR stack.
- Added changeset for the CSS change.

## After PR TODOs:
* [ ] Re-think the polymorphic as-prop ([#2152 (comment)](https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1684836241), [#2135 (comment)](https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1672595007))
* [ ] Investigate why next page button doesn't work when the ellipses show up. ([#2135 (comment)](https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1797302884))

### DIFF
--- a/packages/itwinui-react/src/core/Table/TablePaginator.tsx
+++ b/packages/itwinui-react/src/core/Table/TablePaginator.tsx
@@ -430,13 +430,13 @@ const TablePaginatorPageButtons = (props: TablePaginatorPageButtonsProps) => {
       {pageList.slice(startPage, endPage)}
       {endPage !== totalPagesCount && !isLoading && (
         <>
-          {ellipsis}
+          {showEndEllipsis ? ellipsis : null}
           {pageButton(totalPagesCount - 1, 0)}
         </>
       )}
       {isLoading && (
         <>
-          {showEndEllipsis ? ellipsis : null}
+          {ellipsis}
           <ProgressRadial indeterminate size='small' />
         </>
       )}


packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    if (visibleCount >= children.length) {
+      return children;
+    }
+
+    if (overflowLocation === 'start') {
+      return (
+        <>
+          {overflowTag(visibleCount - 2)}
```
mayank99, line 122: Why -2? The code comments don't help understand this, and actually add to the confusion (why would there ever be `children.length + 1`?)

I feel like the `overflowButton` logic in `ButtonGroup` is easier to reason about, both from the inside (component code) and from the outside (consumer-facing API).

mayank99: > `visibleCount = 11` = 10 items and 0 overflow tags

I don't agree with the `visibleCount` returning false information, but I will save that discussion for the final PR. In this PR, it looks like the code comment is totally unnecessary/distracting, since there is nothing special happening yet, so I would just remove it.

---


> Are you suggesting changing `OverflowContainer`'s `overflowTag` in some way? 

Yes, I'm suggesting calling the function with `overflowStart` instead of `visibleCount`. Or you could even provide both.

---

I would write it like this:


```tsx
const visibleItems = React.useMemo(() => {
  if (overflowLocation === 'start') {
    return (
      <>
        {overflowTag({ visibleCount, overflowStart, overflowCount })}
        {children.slice(children.length - (visibleCount - 1))}
      </>
    );
  }

  return (
    <>
      {children.slice(0, visibleCount - 1)}
      {overflowTag({ visibleCount, overflowStart, overflowCount })}
    </>
  );
}, [children, overflowTag, overflowLocation, visibleCount]);
```

mayank99: > `visibleCount = 11` = 10 items and 0 overflow tags

I don't agree with the `visibleCount` returning false information, but I will save that discussion for the final PR. In this PR, it looks like the code comment is totally unnecessary/distracting, since there is nothing special happening yet, so I would just remove it.

---


> Are you suggesting changing `OverflowContainer`'s `overflowTag` in some way? 

Yes, I'm suggesting calling the function with `overflowStart` instead of `visibleCount`. Or you could even provide both.

---

I would write it like this:


```tsx
const visibleItems = React.useMemo(() => {
  const overflowStart = overflowLocation === 'start'
    ? children.length - visibleCount
    : visibleCount - 1;
  const overflowCount = children.length - visibleCount;

  if (overflowLocation === 'start') {
    return (
      <>
        {overflowTag({ visibleCount, overflowStart, overflowCount })}
        {children.slice(overflowStart + 1)}
      </>
    );
  }

  return (
    <>
      {children.slice(0, Math.max(0, overflowStart))}
      {overflowTag({ visibleCount, overflowStart, overflowCount })}
    </>
  );
}, [children, overflowTag, overflowLocation, visibleCount]);
```

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <>
+        {children.slice(0, visibleCount - 1)}
+        {overflowTag(visibleCount)}
+      </>
+    );
+  }, [children, overflowTag, overflowLocation, visibleCount]);
+
+  return (
+    <Box ref={useMergedRefs(ref, containerRef)} {...rest}>
+      {typeof children === 'function' ? children(visibleCount) : visibleItems}
```
mayank99, line 138: This render prop branching isn't nice, and kinda defeats the purpose of extracting a component (it's basically a hook with extra steps). What is the main reason for calling `children` as a function?

I'm wondering if it would be better to expose a subcomponent instead.

```jsx
<OverflowContainer>
  <>{/* regular nodes */}</>

  {/* this is rendered only when overflowing */}
  <OverflowNode>…</OverflowNode>
</OverflowContainer>
```

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+ *   }
+ * </OverflowContainer>
+ */
+export const OverflowContainer = React.forwardRef((props, ref) => {
+  const {
+    overflowTag,
+    overflowLocation = 'end',
+    items,
+    children,
+    overflowDisabled = false,
```
mayank99, line 94: The `overflowDisabled` prop is not necessary, since we can just conditionally render the component.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+    if (visibleCount >= children.length) {
+      return children;
+    }
+
+    if (overflowLocation === 'start') {
+      return (
+        <>
+          {overflowTag(visibleCount)}
+          {children.slice(children.length - (visibleCount - 1))}
```
mayank99, line 123: Whenever doing array operations on children (`.slice`, `.length`), it is recommended to use [`React.Children.toArray`](https://react.dev/reference/react/Children#children-toarray) so that empty and conditional nodes are automatically filtered out

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -186,14 +188,12 @@ const OverflowGroup = React.forwardRef((props, forwardedRef) => {
...
-
-  return (
-    <BaseGroup
+  return overflowButton != null ? (
+    <OverflowContainer
+      as={BaseGroup}
+      items={items}
+      overflowOrientation={orientation}
+      overflowLocation={overflowPlacement}
       orientation={orientation}
```
mayank99, line 197: Looks like you have both `orientation` and `overflowOrientation`. Is this a mistake?

packages/itwinui-react/src/utils/components/MiddleTextTruncation.tsx
```diff
@@ -46,33 +46,38 @@ export type MiddleTextTruncationProps = {
...
-    if (visibleCount < text.length) {
-      return `${text.substring(
-        0,
-        visibleCount - endCharsCount - ELLIPSIS_CHAR.length,
-      )}${ELLIPSIS_CHAR}${text.substring(text.length - endCharsCount)}`;
-    } else {
-      return text;
-    }
-  }, [endCharsCount, text, visibleCount]);
+  const truncatedText = React.useCallback(
```
mayank99, line 49: It's worth noting that converting a `useMemo` into a `useCallback` changes the meaning significantly: previously, the nodes were being memoized, whereas now only the function is memoized, and the nodes are recreated on every render (because the function is called on every render).

It might not be a problem in this case, but I wonder if we could avoid it somehow 🤔

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
mayank99: It's fine for now, but I'm starting to believe we need to rethink our polymorphic `as` prop. I'm reminded of the [other recent issue](https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1672595007) we had in `VisuallyHidden`. Let's make a note about this.

Adding an [`asChild` prop](https://www.radix-ui.com/primitives/docs/guides/composition#composing-multiple-primitives) could be one alternative to explore in the future.

Ben-Pusey-Bentley (approved):
I don't have any issues with this PR that don't already have comments, so I am approving this.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: Yeah, the new context API makes sense. I think we should also have a hook called something like `useOverflowContext`. It would just be a wrapper over `useSafeContext(OverflowContainerContext)`, with the main goal of allowing to easily destructure the return value without needing to import `useSafeContext` or `OverflowContainerContext`.

My main concern with the context API is the need to extract child components. While it's generally a good idea to have lots of smaller components rather than one large component, it makes the diff too large in some cases (e.g. TablePaginator).

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: I believe this code was moved out of `OverflowContainer` so the change was undone.

For example in breadcrumbs, we currently only rename the `children` prop `items`. We should instead use `React.Children.toArray` (even if we weren't previously using it). 

```jsx
const items = React.Children.toArray(children);
```

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: Looks like this was resolved via https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1686518959

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+type OverflowContainerProps = {
+  /**
+   * The orientation of the overflow in container.
+   * @default 'horizontal'
+   */
+  overflowOrientation?: 'horizontal' | 'vertical';
+  /**
+   * TODO: Will be removed in a later PR in the stacked PRs.
+   */
+  items: React.ReactNode[] | string;
```
mayank99, line 20: This prop can be removed now that we don't have children as a function?

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const [containerRef, visibleCount] = useOverflow(
+    items,
+    false,
+    overflowOrientation,
+  );
+
+  return (
+    <OverflowContainerContext.Provider
+      value={{ visibleCount, itemCount: items.length }}
```
mayank99, line 72: context values should be memoized

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -172,6 +176,12 @@ const BaseGroup = React.forwardRef((props, forwardedRef) => {
 
 // ----------------------------------------------------------------------------
 
+type OverflowGroupProps = Pick<
+  ButtonGroupProps,
+  'children' | 'orientation' | 'overflowButton' | 'overflowPlacement'
+>;
+
+/** Note: If `overflowButton == null`, it behaves like a `BaseGroup`. */
```
mayank99, line 184: This component doesn't need to handle `overflowButton == null` because it's already taken care of in the parent component. We can just assume that `overflowButton` will not be null; the types can be changed to enforce this.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const isOverflowing =
+    overflowContainerContext != null &&
+    overflowContainerContext.visibleCount < overflowContainerContext.itemCount;
+
+  return isOverflowing && children;
+};
+
+// ----------------------------------------------------------------------------
+
+export const OverflowContainer = Object.assign(OverflowContainerComponent, {
```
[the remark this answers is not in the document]
mayank99: try to combine them with the existing `utils/index.js` imports

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { useOverflow } from '../hooks/useOverflow.js';
+import { useSafeContext } from '../hooks/useSafeContext.js';
+
+type OverflowContainerProps = {
+  /**
+   * The orientation of the overflow in container.
+   * @default 'horizontal'
+   */
+  overflowOrientation?: 'horizontal' | 'vertical';
+  itemsCount: number;
```
mayank99, line 18: This could use some JSDocs. The prop name is not self-explanatory (I think it represents the *original* items count?).

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const OverflowContainerOverflowNode = (
+  props: OverflowContainerOverflowNodeProps,
+) => {
+  const { children } = props;
+
+  const { visibleCount, itemsCount } = useOverflowContainerContext();
+  const isOverflowing = visibleCount < itemsCount;
+
+  return isOverflowing && children;
```
mayank99, line 64: This should return `null` for the falsy case.

```suggestion
  return isOverflowing ? children : null;
```

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -0,0 +1,92 @@
...
+      visibleCount: number;
+      itemsCount: number;
+    }
+  | undefined
+>(undefined);
+if (process.env.NODE_ENV === 'development') {
+  OverflowContainerContext.displayName = 'OverflowContainerContext';
+}
+
+export const useOverflowContainerContext = () => {
```
mayank99, line 89: nit/question: Could this be exposed as `OverflowContainer.useContext()`?

packages/itwinui-react/src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -114,70 +114,41 @@ type BreadcrumbsProps = {
...
     ...rest
   } = props;
 
-  const [overflowRef, visibleCount] = useOverflow(items.length);
-  const refs = useMergedRefs(overflowRef, ref);
+  const items = React.useMemo(
+    () => React.Children.toArray(childrenProp),
+    [childrenProp],
+  );
+  const currentIndex = currentIndexProp || items.length - 1;
```
mayank99, line 129: It's confusing to set default values in the middle of the component definition; can we use `React.Children.count(childrenProp) - 1` when destructuring it?

packages/itwinui-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -203,40 +210,50 @@ const OverflowGroup = React.forwardRef((props, forwardedRef) => {
...
+  const { overflowButton, overflowPlacement, items } = props;
+  const { visibleCount } = useOverflowContainerContext();
+
+  const overflowStart =
+    overflowPlacement === 'start'
+      ? items.length - visibleCount
+      : visibleCount - 1;
+
+  return !(visibleCount < items.length) ? (
+    items
```
mayank99, line 243: nit: Can we use `if` conditions (with early `return`s) instead of ternaries for better readability?

packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -401,3 +280,191 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
+  if (startPage < 0) {
+    endPage = Math.min(totalPagesCount, endPage + Math.abs(startPage)); // If no room at the beginning, show extra pages at the end
+    startPage = 0;
+  }
+  if (endPage > totalPagesCount) {
+    startPage = Math.max(0, startPage - (endPage - totalPagesCount)); // If no room at the end, show extra pages at the beginning
+    endPage = totalPagesCount;
+  }
+
+  const ellipsis = (
```
mayank99, line 342: Can some of this code be moved back down to reduce diff?

packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -302,64 +221,24 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
+            needFocus={needFocus}
+            onPageChange={onPageChange}
+            setFocusedIndex={setFocusedIndex}
+            currentPage={currentPage}
+            localization={localization}
+            buttonSize={buttonSize}
+            pageListRef={pageListRef}
+            pageButton={pageButton}
+            pageList={pageList}
+            isLoading={isLoading}
```
mayank99, line 239: Passing this huge list of props to the child component is not nice. Can we reduce the number of props through some restructuring? Maybe some of the rendered elements could be moved back into the main component, and some of the logic could be moved into the child component.

mayank99 (commented):
While this isn't exactly what I had envisioned from an `OverflowContainer` component (esp. the whole context thing), I don't want to block the PR and I don't have the time to look more deeply and make better suggestions. So I'm ok with this, especially since it's an internal component.

What follows is mostly minor comments around code structure.

packages/itwinui-react/src/core/Table/TablePaginator.tsx
mayank99: I was able to remove most of the props by moving things around.

<details><summary>Patch</summary>

```diff
--- a/packages/itwinui-react/src/core/Table/TablePaginator.tsx
+++ b/packages/itwinui-react/src/core/Table/TablePaginator.tsx
@@ -169,34 +169,52 @@ export const TablePaginator = (props: TablePaginatorProps) => {
   }, [focusedIndex]);
 
   const buttonSize = size != 'default' ? 'small' : undefined;
+  const totalPagesCount = Math.ceil(totalRowsCount / pageSize);
 
-  const pageButton = React.useCallback(
-    (index: number, tabIndex = index === focusedIndex ? 0 : -1) => (
-      <Button
-        key={index}
-        className='iui-table-paginator-page-button'
-        styleType='borderless'
-        size={buttonSize}
-        data-iui-active={index === currentPage}
-        onClick={() => onPageChange(index)}
-        aria-current={index === currentPage}
-        aria-label={localization.goToPageLabel(index + 1)}
-        tabIndex={tabIndex}
-      >
-        {index + 1}
-      </Button>
-    ),
-    [focusedIndex, currentPage, localization, buttonSize, onPageChange],
-  );
+  const onKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    // alt + arrow keys are used by browser/assistive technologies
+    if (event.altKey) {
+      return;
+    }
 
-  const totalPagesCount = Math.ceil(totalRowsCount / pageSize);
-  const pageList = React.useMemo(
-    () =>
-      new Array(totalPagesCount)
-        .fill(null)
-        .map((_, index) => pageButton(index)),
-    [pageButton, totalPagesCount],
-  );
+    const focusPage = (delta: number) => {
+      const newFocusedIndex = getBoundedValue(
+        focusedIndex + delta,
+        0,
+        totalPagesCount - 1,
+      );
+
+      needFocus.current = true;
+      if (focusActivationMode === 'auto') {
+        onPageChange(newFocusedIndex);
+      } else {
+        setFocusedIndex(newFocusedIndex);
+      }
+    };
+
+    switch (event.key) {
+      case 'ArrowRight': {
+        focusPage(+1);
+        event.preventDefault();
+        break;
+      }
+      case 'ArrowLeft': {
+        focusPage(-1);
+        event.preventDefault();
+        break;
+      }
+      case 'Enter':
+      case ' ':
+      case 'Spacebar': {
+        if (focusActivationMode === 'manual') {
+          onPageChange(focusedIndex);
+        }
+        break;
+      }
+      default:
+        break;
+    }
+  };
 
   const [paginatorResizeRef, paginatorWidth] = useContainerWidth();
 
@@ -204,6 +222,19 @@ export const TablePaginator = (props: TablePaginatorProps) => {
   const showPageSizeList =
     pageSizeList && !!onPageSizeChange && !!totalRowsCount;
 
+  const hasNoRows = totalPagesCount === 0;
+  const noRowsContent = (
+    <>
+      {isLoading ? (
+        <ProgressRadial indeterminate size='small' />
+      ) : (
+        <Button styleType='borderless' disabled size={buttonSize}>
+          1
+        </Button>
+      )}
+    </>
+  );
+
   if (!showPagesList && !showPageSizeList) {
     return null;
   }
@@ -220,23 +251,45 @@ export const TablePaginator = (props: TablePaginatorProps) => {
         )}
       </Box>
       {showPagesList && (
-        <OverflowContainer className='iui-center' itemsCount={pageList.length}>
-          <TablePaginatorCenterContent
-            size={size}
-            focusedIndex={focusedIndex}
-            focusActivationMode={focusActivationMode}
-            totalPagesCount={totalPagesCount}
-            needFocus={needFocus}
-            onPageChange={onPageChange}
-            setFocusedIndex={setFocusedIndex}
-            currentPage={currentPage}
-            localization={localization}
-            buttonSize={buttonSize}
-            pageListRef={pageListRef}
-            pageButton={pageButton}
-            pageList={pageList}
-            isLoading={isLoading}
-          />
+        <OverflowContainer className='iui-center' itemsCount={totalPagesCount}>
+          <IconButton
+            styleType='borderless'
+            disabled={currentPage === 0}
+            onClick={() => onPageChange(currentPage - 1)}
+            size={buttonSize}
+            aria-label={localization.previousPage}
+          >
+            <SvgChevronLeft />
+          </IconButton>
+          <Box
+            as='span'
+            className='iui-table-paginator-pages-group'
+            onKeyDown={onKeyDown}
+            ref={pageListRef}
+          >
+            {hasNoRows ? (
+              noRowsContent
+            ) : (
+              <TablePaginatorPageButtons
+                size={size}
+                focusedIndex={focusedIndex}
+                totalPagesCount={totalPagesCount}
+                onPageChange={onPageChange}
+                currentPage={currentPage}
+                localization={localization}
+                isLoading={isLoading}
+              />
+            )}
+          </Box>
+          <IconButton
+            styleType='borderless'
+            disabled={currentPage === totalPagesCount - 1 || hasNoRows}
+            onClick={() => onPageChange(currentPage + 1)}
+            size={buttonSize}
+            aria-label={localization.nextPage}
+          >
+            <SvgChevronRight />
+          </IconButton>
         </OverflowContainer>
       )}
       <Box className='iui-right'>
@@ -282,92 +335,57 @@ export const TablePaginator = (props: TablePaginatorProps) => {
 
 // ----------------------------------------------------------------------------
 
-type TablePaginatorCenterContentProps = Pick<
+type TablePaginatorPageButtonsProps = Pick<
   TablePaginatorProps,
   'onPageChange'
 > &
-  Required<
-    Pick<
-      TablePaginatorProps,
-      'localization' | 'size' | 'focusActivationMode' | 'isLoading'
-    >
-  > & {
+  Required<Pick<TablePaginatorProps, 'localization' | 'size' | 'isLoading'>> & {
     focusedIndex: number;
     totalPagesCount: number;
-    needFocus: React.MutableRefObject<boolean>;
-    setFocusedIndex: React.Dispatch<React.SetStateAction<number>>;
     currentPage: number;
-    buttonSize: 'small' | undefined;
-    pageListRef: React.MutableRefObject<HTMLDivElement | null>;
-    pageButton: (index: number, tabIndex?: number) => React.ReactNode;
-    pageList: React.ReactNode[];
   };
 
-const TablePaginatorCenterContent = (
-  props: TablePaginatorCenterContentProps,
-) => {
+const TablePaginatorPageButtons = (props: TablePaginatorPageButtonsProps) => {
   const {
     focusedIndex,
-    focusActivationMode,
     totalPagesCount,
-    needFocus,
     onPageChange,
-    setFocusedIndex,
     currentPage,
     localization,
-    buttonSize,
-    pageListRef,
-    pageButton,
-    pageList,
     isLoading,
     size,
   } = props;
-  const { visibleCount } = OverflowContainer.useContext();
 
-  const onKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    // alt + arrow keys are used by browser/assistive technologies
-    if (event.altKey) {
-      return;
-    }
+  const { visibleCount } = OverflowContainer.useContext();
 
-    const focusPage = (delta: number) => {
-      const newFocusedIndex = getBoundedValue(
-        focusedIndex + delta,
-        0,
-        totalPagesCount - 1,
-      );
+  const buttonSize = size != 'default' ? 'small' : undefined;
 
-      needFocus.current = true;
-      if (focusActivationMode === 'auto') {
-        onPageChange(newFocusedIndex);
-      } else {
-        setFocusedIndex(newFocusedIndex);
-      }
-    };
+  const pageButton = React.useCallback(
+    (index: number, tabIndex = index === focusedIndex ? 0 : -1) => (
+      <Button
+        key={index}
+        className='iui-table-paginator-page-button'
+        styleType='borderless'
+        size={buttonSize}
+        data-iui-active={index === currentPage}
+        onClick={() => onPageChange(index)}
+        aria-current={index === currentPage}
+        aria-label={localization.goToPageLabel?.(index + 1)}
+        tabIndex={tabIndex}
+      >
+        {index + 1}
+      </Button>
+    ),
+    [focusedIndex, currentPage, localization, buttonSize, onPageChange],
+  );
 
-    switch (event.key) {
-      case 'ArrowRight': {
-        focusPage(+1);
-        event.preventDefault();
-        break;
-      }
-      case 'ArrowLeft': {
-        focusPage(-1);
-        event.preventDefault();
-        break;
-      }
-      case 'Enter':
-      case ' ':
-      case 'Spacebar': {
-        if (focusActivationMode === 'manual') {
-          onPageChange(focusedIndex);
-        }
-        break;
-      }
-      default:
-        break;
-    }
-  };
+  const pageList = React.useMemo(
+    () =>
+      new Array(totalPagesCount)
+        .fill(null)
+        .map((_, index) => pageButton(index)),
+    [pageButton, totalPagesCount],
+  );
 
   const halfVisibleCount = Math.floor(visibleCount / 2);
   let startPage = focusedIndex - halfVisibleCount;
@@ -381,8 +399,6 @@ const TablePaginatorCenterContent = (
     endPage = totalPagesCount;
   }
 
-  const hasNoRows = totalPagesCount === 0;
-
   const ellipsis = (
     <Box
       as='span'
@@ -394,76 +410,31 @@ const TablePaginatorCenterContent = (
     </Box>
   );
 
-  const noRowsContent = (
-    <>
-      {isLoading ? (
-        <ProgressRadial indeterminate size='small' />
-      ) : (
-        <Button styleType='borderless' disabled size={buttonSize}>
-          1
-        </Button>
-      )}
-    </>
-  );
+  if (visibleCount === 1) {
+    return pageButton(focusedIndex);
+  }
 
   return (
     <>
-      <IconButton
-        styleType='borderless'
-        disabled={currentPage === 0}
-        onClick={() => onPageChange(currentPage - 1)}
-        size={buttonSize}
-        aria-label={localization.previousPage}
-      >
-        <SvgChevronLeft />
-      </IconButton>
-      <Box
-        as='span'
-        className='iui-table-paginator-pages-group'
-        onKeyDown={onKeyDown}
-        ref={pageListRef}
-      >
-        {(() => {
-          if (hasNoRows) {
-            return noRowsContent;
-          }
-          if (visibleCount === 1) {
-            return pageButton(focusedIndex);
-          }
-          return (
-            <>
-              {startPage !== 0 && (
-                <>
-                  {pageButton(0, 0)}
-                  {ellipsis}
-                </>
-              )}
-              {pageList.slice(startPage, endPage)}
-              {endPage !== totalPagesCount && !isLoading && (
-                <>
-                  {ellipsis}
-                  {pageButton(totalPagesCount - 1, 0)}
-                </>
-              )}
-              {isLoading && (
-                <>
-                  {ellipsis}
-                  <ProgressRadial indeterminate size='small' />
-                </>
-              )}
-            </>
-          );
-        })()}
-      </Box>
-      <IconButton
-        styleType='borderless'
-        disabled={currentPage === totalPagesCount - 1 || hasNoRows}
-        onClick={() => onPageChange(currentPage + 1)}
-        size={buttonSize}
-        aria-label={localization.nextPage}
-      >
-        <SvgChevronRight />
-      </IconButton>
+      {startPage !== 0 && (
+        <>
+          {pageButton(0, 0)}
+          {ellipsis}
+        </>
+      )}
+      {pageList.slice(startPage, endPage)}
+      {endPage !== totalPagesCount && !isLoading && (
+        <>
+          {ellipsis}
+          {pageButton(totalPagesCount - 1, 0)}
+        </>
+      )}
+      {isLoading && (
+        <>
+          {ellipsis}
+          <ProgressRadial indeterminate size='small' />
+        </>
+      )}
     </>
   );
 };
```

</details>

packages/itwinui-react/src/core/Table/TablePaginator.tsx
mayank99, line 1: Something about the last commit seems to have broken the paginator (hence the CI failure). The "next page" button stops working after a few pages.

mayank99: The "next page" button stops responding after you click it enough times. Screen recording below using Chrome.


https://github.com/user-attachments/assets/3bb45e3d-3904-4793-8733-026adbb00e93

mayank99 (approved):

smmr-dn (approved):
LGTM! I have no further suggestions :)

PR #2154, opened 2024-07-22
New and improved `useOverflow` algorithm

r100-stack:
## Changes

Pulled out of #2120 as PR 4 in the PR stack ([PR stack order](https://github.com/iTwin/iTwinUI/pull/2120#issuecomment-2386574990)).

As mentioned in https://github.com/iTwin/iTwinUI/issues/2112#issuecomment-2191857965, I believe the flickering is because of an infinite loop:

> `visibleCount` from our `useOverflow` internal hook changes on each resize of `iui-select-tag-container` and each resize changes the `visibleCount`.

I think this is because our `useOverflow` hook makes the assumption that all items are around the same size and thus we can take an *average*:

https://github.com/iTwin/iTwinUI/blob/4f431e59614ce50af39fc502765159efc7dd0b19/packages/itwinui-react/src/utils/hooks/useOverflow.tsx#L79

While this may work for cases like the TablePaginator, it doesn't seem to work for the ComboBox case where tags could have different widths.

Thus, `useOverflow` should be refactored to use real sizes.

---

Logic of new algorithm for `useOverflow`:
- Have a guess range for `visibleCount`. e.g. `[0, 32]` (32 is an arbitrary choice).
- Keep doubling the max guess until the container overflows.
  i.e. the max guess should always be `≥` the correct `visibleCount`.
  - With each such doubling, the new min guess is the current max guess (since underflow = we guessed low).
- Set `visibleCount` to the `maxGuess`.
- Repeat the following by calling `guessVisibleCount()` (keep re-rendering but not painting):
  - Each time the container overflows, new max guess is the average of the two guesses (since overflow = we guessed high).
  - Each time the container does not overflow, new min guess is the average of the two guesses (since underflow = we guessed low).
- Stop when the average of the two guesses is the min guess itself. i.e. no more averaging possible.
- The min guess is then the correct `visibleCount`.

Why this works:
* This approach takes the *real* widths/heights into account. So, the previous flickering is no longer there.
* Even if there are 10k items (for example), the maximum number of items we render starts from 32 and keeps doubling until we reach the first overflow. After the first overflow, the number of items we render always goes down. So, performance is not affected because we don't render any more items than needed. 
* Since this approach involves halving the guess in each render, I believe this approach is of the order of `log(n)`.

---

Others:
* `useOverflow` removes `disabled` argument since we can just conditionally render `OverflowContainer`. (https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1684710209)
* `useOverflow` returns the same values that is used to return before this PR. I didn't re-implement the WIP logic where `useOverflow` returned one more than the total number of items when all items are visible. (https://github.com/iTwin/iTwinUI/pull/2152#discussion_r1684706078)

## Testing

* Remove failing unit tests since mocking `scrollWidth` and `offsetWidth` was no longer reasonable with the new algorithm implementation. Equivalent e2e already added in #2151 from the PR chain.
* Testing `overflowStart` is correct:
  * e2e test
  * Comparing the storybook `overflowStart` with a [playground](https://stackblitz.com/edit/github-wsvnha?file=src%2FApp.tsx) from `3.13.1` with the same code.
* Added an e2e to confirm that #2112 is indeed fixed.

When testing, I realized that in addition to fixing this bug, this PR also might improve performance. Specifically, the old algorithm made infinite changes in the DOM even when there is no flickering visually. But the new algorithm doesn't make these continuous DOM updates.

<p>

```jsx
import { ComboBox, Flex } from '@itwin/itwinui-react';
import React from 'react';

const App = () => {
  const data = new Array(15).fill(0).map((_, i) => ({
    label: `option ${i}`,
    value: i,
  }));
  const widths = new Array(10).fill(0).map((_, i) => 790 + i * 3);

  const [
    selectTagContainersDomChangeCount,
    setSelectTagContainersDomChangeCount,
  ] = React.useState(0);

  const observer = new MutationObserver(() =>
    setSelectTagContainersDomChangeCount((count) => count + 1),
  );

  React.useEffect(() => {
    const selectTagContainers = document.querySelectorAll(
      "[role='combobox'] + div:first-of-type",
    );
    selectTagContainers.forEach((container) => {
      observer.observe(container, {
        attributes: false,
        childList: true,
        subtree: false,
      });
    });

    return () => {
      observer.disconnect();
    };
  }, []);

  return (
    <Flex flexDirection='column' alignItems='flex-start'>
      <p data-testid='dom-changes-counter'>
        Select tag containers DOM changes: {selectTagContainersDomChangeCount}
      </p>
      {widths.slice(0, 10).map((width) => (
        <ComboBox
          key={width}
          style={{ width: `${width}px`, maxWidth: '80vw' }}
          multiple={true}
          options={data}
          value={data.map((x) => x.value)}
        />
      ))}
    </Flex>
  );
};

export default App;

```

</p>
</details> 

<details><summary>Result</summary>

Before:

Halfway through the video, I increased the throttling percentage.

https://github.com/user-attachments/assets/23d50d77-612b-44bc-91d2-3616c754c246

After:

https://github.com/user-attachments/assets/c0abb20e-0752-437b-8789-983fd8b5128b

</details> 

## Docs

* Added necessary JSDocs to explain the logic of the new algorithm and to the new props.
* Added changeset since #2112 is now fixed.

## After PR TODOs:

* [ ] Try making the tags-container in ComboBox expand back when more width is available ([#2154 (comment)](https://github.com/iTwin/iTwinUI/pull/2154#discussion_r1813408275)).
* [ ] Try reducing the number of dependencies in the `useLayoutEffect()` ([#2154 (comment)](https://github.com/iTwin/iTwinUI/pull/2154#discussion_r1813476727))


mayank99 (commented):
Before I review this, could you list out (in order) all PRs in the stack somewhere? It's a little difficult to tell at a glance which PR goes where. I'm also not sure of the current status of #2152 (it's been open for 2 months).

Also it would be good to start merging some of the earlier PRs in the stack, unless there is a strong reason to keep them around. Long-lived branches can be difficult to keep up-to-date.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -90,3 +92,222 @@ export const useOverflowContainerContext = () => {
...
+
+        setVisibleCountGuessRange([visibleCount, doubleOfMaxGuess]);
+        setVisibleCount(doubleOfMaxGuess);
+
+        return;
+      }
+
+      let newVisibleCountGuessRange = visibleCountGuessRange;
+
+      if (isOverflowing) {
```
smmr-dn, line 232: I'm wondering if we can simplify this if-else statement into one ternary such as:
```jsx
const newVisibleCountGuessRange = isOverflowing
      ? [visibleCountGuessRange[0], visibleCount]
      : [visibleCount, visibleCountGuessRange[1]];
```

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+  useLayoutEffect(() => {
+    if (isStabilized || isUnitTest) {
+      return;
+    }
+
+    if (
+      visibleCount !== previousVisibleCount.current ||
+      JSON.stringify(visibleCountGuessRange) !==
+        (previousVisibleCountGuessRange.current != null
+          ? JSON.stringify(previousVisibleCountGuessRange.current)
```
smmr-dn, line 294: I have the same question for this check as in [comment](https://github.com/iTwin/iTwinUI/pull/2154/files#r1806739124).

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+    itemsCount,
+    orientation,
+    setVisibleCount,
+    visibleCount,
+    visibleCountGuessRange,
+  ]);
+
+  const previousVisibleCount = useLatestRef(visibleCount);
+  const previousVisibleCountGuessRange = useLatestRef(visibleCountGuessRange);
+  const previousContainer = useLatestRef(containerRef.current);
```
mayank99, line 279: I'm not sure why we're using `useLatestRef` here, if we're manually changing these things in the `useLayoutEffect` anyway.

I also think these variables might not even be necessary. The `useLayoutEffect` dependency array could automatically take care of these things. To make that work, `guessVisibleCount` is what should be passed through `useLatestRef` to ensure that it doesn't re-trigger the Effect every time it changes.

And the `containerRef` shouldn't ever change. If it does, we should probably reset everything.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+  );
+  const initialVisibleCountGuessRange = React.useMemo(
+    () => [0, initialVisibleCount] satisfies GuessRange,
+    [initialVisibleCount],
+  );
+
+  const [visibleCount, _setVisibleCount] =
+    React.useState<number>(initialVisibleCount);
+  const [visibleCountGuessRange, setVisibleCountGuessRange] =
+    React.useState<GuessRange>(initialVisibleCountGuessRange);
```
mayank99, line 151: This whole "range" idea feels confusing and even misleading to me. The first thing I would suggest is to use an object instead of array, so that we are not dealing with indexes.

---

Further, I think we can reduce the number of different states (while keeping the same algorithm).

There are four (three and a half?) different states right now:
1. min guess
2. max guess
3. `visibleCount`
4. `isStabilized` indicated by the "range" being `null`.

The "max guess" makes sense, but the "min guess" and `null` parts are what feel unnecessary. The "min guess" is basically the "stale" version of the "max guess", and also often overlaps with the `visibleCount`. And setting `null` also feels unnecessary — we could probably derive it from other existing states.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+        const newVisibleCount =
+          typeof setStateAction === 'function'
+            ? setStateAction(prev)
+            : setStateAction;
+
+        return Math.min(newVisibleCount, itemsCount);
+      });
+    },
+    [itemsCount],
+  );
```
mayank99, line 170: This kind of wrapper makes it confusing to follow the logic, because we can't ever trust that our `setVisibleCount` call is actually setting the value that we passed. It would be better to remove this wrapper, and call `Math.min` at the place where we are setting the state.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+  // - `visibleCountGuessRange`
+  // - `containerRef`
+  useLayoutEffect(() => {
+    if (isStabilized || isUnitTest) {
+      return;
+    }
+
+    if (
+      visibleCount !== previousVisibleCount.current ||
+      JSON.stringify(visibleCountGuessRange) !==
```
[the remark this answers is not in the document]
smmr-dn: Hmm, I went through the code locally and saw that the type of `visibleCountGuessRange` used to be just `[number, number]` so I just mentioned `undefined` or `null` just to clarify. The new approach makes these clear to me so I'm closing this comment :)

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
smmr-dn: Closing this comment since it has been mentioned in [comment](https://github.com/iTwin/iTwinUI/pull/2154#discussion_r1806822405).

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +104,224 @@ if (process.env.NODE_ENV === 'development') {
...
+    },
+    [itemsCount],
+  );
+
+  const [resizeRef] = useResizeObserver(
+    React.useCallback(() => {
+      setVisibleCount(initialVisibleCount);
+      setVisibleCountGuessRange(initialVisibleCountGuessRange);
+    }, [initialVisibleCount, initialVisibleCountGuessRange, setVisibleCount]),
+  );
```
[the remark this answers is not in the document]
mayank99: To keep the overflow code centralized and reduce repetition, you might need a wrapper component in `OverflowContainer.tsx`.

The nested hierarchy would be something like:

```
- Breadcrumbs/ButtonGroup/etc
- OverflowContainerWrapper
  - useContainerWidth
  - OverflowContainerMain (key)
    - useOverflow
```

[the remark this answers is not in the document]
mayank99: There are two issues in that commit: `keyProp` and `refProp`. We cannot destructure these things because `key` and `ref` are not regular props; they are reserved get removed from the props object by React.

I don't see proper docs for this; the closest thing I found was the [representation of a JSX node](https://react.dev/reference/react/createElement#returns), which lists out `key` and `ref` separately from `props`. You can also verify the behavior at runtime; I believe `keyProp` and `refProp` will always be `undefined`.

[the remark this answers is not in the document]
mayank99: Another issue I've noticed is that in multi-select ComboBox, if you select enough items then make the window narrower, the tag container correctly collapses based on available width, but then when you resize it back, it doesn't expand back. Maybe this is ok, because it's a rare scenario?

To handle this correctly, we would need to move the ResizeObserver back into the ComboBox.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: > > I recently noticed that this logic does not work properly for `Select`.
> 
> I found that this was because we weren't handling the `itemsCount` change case. Now, that case is handled where an `itemsCount` change resets the guess related variables. ([6a85122](https://github.com/iTwin/iTwinUI/commit/6a85122861515002b1d8c845b92fce4ad1f47121)).

I think we can handle this using `key` as well, and not have to maintain the `previousItemsCount` state.

<details><summary>Patch</summary>

```tsx
diff --git a/packages/itwinui-react/src/utils/components/OverflowContainer.tsx b/packages/itwinui-react/src/utils/components/OverflowContainer.tsx
index 03e28d68c..36067f912 100644
--- a/packages/itwinui-react/src/utils/components/OverflowContainer.tsx
+++ b/packages/itwinui-react/src/utils/components/OverflowContainer.tsx
@@ -53,7 +53,7 @@ type OverflowContainerProps = {
  *   );
  * };
  */
-const OverflowContainerMain = React.forwardRef((props, ref) => {
+const OverflowContainerMain = React.forwardRef((props, forwardedRef) => {
   const { itemsCount, children, overflowOrientation, ...rest } = props;
 
   const [containerRef, visibleCount] = useOverflow(
@@ -68,7 +68,7 @@ const OverflowContainerMain = React.forwardRef((props, ref) => {
 
   return (
     <OverflowContainerContext.Provider value={overflowContainerContextValue}>
-      <Box ref={useMergedRefs(ref, containerRef)} {...rest}>
+      <Box ref={useMergedRefs(forwardedRef, containerRef)} {...rest}>
         {children}
       </Box>
     </OverflowContainerContext.Provider>
@@ -103,18 +103,23 @@ const OverflowContainerOverflowNode = (
 
 const OverflowContainerComponent = React.forwardRef((props, forwardedRef) => {
-  const { overflowOrientation = 'horizontal', ...rest } = props;
+  const { itemsCount, overflowOrientation = 'horizontal', ...rest } = props;
 
   const [size, setSize] = React.useState<DOMRectReadOnly | null>(null);
   const [resizeRef] = useResizeObserver(setSize);
 
   const ref = useMergedRefs(resizeRef, forwardedRef);
 
+  const key =
+    `${itemsCount}` +
+    `${overflowOrientation === 'vertical' ? size?.height : size?.width}`;
+
   return (
     <OverflowContainerMain
       {...rest}
-      key={overflowOrientation === 'vertical' ? size?.height : size?.width}
+      key={key}
       ref={ref}
+      itemsCount={itemsCount}
       overflowOrientation={overflowOrientation}
     />
   );
@@ -278,15 +283,6 @@ const useOverflow = (
     }
   }, [isStabilized, itemsCount, maxGuess, minGuess, orientation, visibleCount]);
 
-  const [previousItemsCount, setPreviousItemsCount] =
-    React.useState(itemsCount);
-
-  // If itemsCount changes, we have to restart guessing.
-  if (itemsCount !== previousItemsCount) {
-    dispatch({ type: 'itemCountChanged', itemsCount });
-    setPreviousItemsCount(itemsCount);
-  }
-
   // Guess the visible count until stabilized.
   // To prevent flicking, use useLayoutEffect to paint only after stabilized.
   useLayoutEffect(() => {

```
</details> 

---

> This didn't work as the `useLayoutEffect` doesn't call `guessVisibleCount()` when the guess related variables change. So, I didn't implement this.
> 
> The `useLayoutEffect()` may be called a few more times because of it, I think it's alright since once it's stabilized, `guessVisibleCount()` is not called unnecessarily.

I meant to respond to this earlier: Relying on a large deps array of `useLayoutEffect` is usually an anti-pattern. One way of fixing this would be to recursively call `guessVisibleCount` in certain scenarios where we know for sure that we want to re-execute all the logic. This would be worth investigating, maybe in the future, so as to not hold up this PR for too long.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: The reducer does make things cleaner, so I like that. I need some more time to think about the other points before I respond.

In the meantime, can you investigate the failing e2e tests?

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+
+      const dimension = orientation === 'horizontal' ? 'Width' : 'Height';
+      const availableSize = containerRef.current[`offset${dimension}`];
+      const requiredSize = containerRef.current[`scroll${dimension}`];
+
+      const isOverflowing = availableSize < requiredSize;
+
+      if (
+        // We have already found the correct visibleCount
+        (visibleCount === itemsCount && !isOverflowing) ||
```
mayank99, line 239: I think this comment is wrong. When `visibleCount` is `itemsCount`, it has nothing to do with "correctness". It's just an early return case because the overflow never happens.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+        // if the new average of visibleCountGuessRange will never change the visibleCount anymore (infinite loop)
+        (maxGuess - minGuess === 1 && visibleCount === minGuess)
```
mayank99, line 241: Could you look through all comments to make sure they're up-to-date? `visibleCountGuessRange` doesn't exist anymore, and some of the comments below might be in the wrong place, since the related code was moved into the reducer.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+        dispatch({ type: 'stabilize' });
+        return;
+      }
+
+      // Before the main logic, the max guess MUST be >= the correct visibleCount for the algorithm to work.
+      // If not:
+      // - double the max guess and visibleCount: since we need to overflow.
+      // - set min guess to current visibleCount: since underflow means correct visibleCount >= current visibleCount.
+      if (maxGuess === visibleCount && !isOverflowing) {
+        dispatch({ type: 'shiftGuessRangeForward' });
```
mayank99, line 252: When does this condition happen? i.e. what are the values for `visibleCount`, `minGuess`, and `maxGuess` when this `if`-condition matches?

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+
+      let newMinGuess = minGuess;
+      let newMaxGuess = maxGuess;
+
+      if (isOverflowing) {
+        // overflowing = we guessed too high. So, new max guess = half the current guess
+        newMaxGuess = visibleCount;
+      } else {
+        // not overflowing = maybe we guessed too low? So, new min guess = half of current guess
+        newMinGuess = visibleCount;
```
mayank99, line 264: I'm not sure when this condition would hit. We already check for `!isOverflowing` above. What is the difference between the two conditions, and when does this one match?

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+      }
+
+      // Next guess is always the middle of the new guess range
+      const newVisibleCount = Math.floor((minGuess + maxGuess) / 2);
```
mayank99, line 268: I'm not sure that this calculation is right. It's still using the old `minGuess` and `maxGuess` values, so it wouldn't even change (since the old `visibleCount` likely was already in the middle).

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -96,6 +157,263 @@ if (process.env.NODE_ENV === 'development') {
...
+
+      // If newVisibleCount is 0 (impossible state as min is 1), end guessing
+      if (newVisibleCount === 0) {
+        dispatch({ type: 'stabilize' });
+        return;
```
mayank99, line 273: Why do we need to handle this case again randomly here? Shouldn't it already be handled by one of the `if` checks earlier?

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: The only remaining suggestion I have is to split the `"update"` action into two.

```jsx
if (isOverflowing) {
  // overflowing = we guessed too high. So, decrease max guess
  dispatch({ type: 'decreaseMaxGuess' });
  return;
} else {
  // not overflowing = maybe we guessed too low? So, increase min guess
  dispatch({ type: 'increaseMinGuess' });
  return;
}
```

All the other code could be moved into the reducer, since we already have the old state values there.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
mayank99: After all the recent changes, I feel good about this. The reducer makes things a lot easier to understand, since it abstracts multiple states into well-named actions. 👍

mayank99 (approved):
LGTM. Sorry about the extra-prolonged review cycles 😅

smmr-dn (approved):

PR #2161, opened 2024-07-23
Fixes virtualized `Table` empty content

Ben-Pusey-Bentley:
## Changes

Fixes the virtualized `Table` components `emptyTableContent` value. This is done by making sure that the `<div>` meant to wrap around the virtual items is only added when there is data being passed into the `Table`, since this `<div>` has a height of 0 when there are no virtual items, causing the `emptyTableContent` to not display.

## Testing

Updated the e2e test for empty virtualized table to check for the `emptyTableContent`. Also tested with this code in the vite playground: 
```tsx
import { Table } from '@itwin/itwinui-react';
import { Column } from '@itwin/itwinui-react/react-table';

const App = () => {
  type CustomStoryDataType = {
    name: string;
    description: string;
  };

    () => [
      {
        id: 'name',
        Header: 'Name',
        accessor: 'name',
      },
      {
        id: 'description',
        Header: 'Description',
        accessor: 'description',
        maxWidth: 200,
      },
    ],
    [],
  ) satisfies Column<CustomStoryDataType>[];

  return (
    <Table<CustomStoryDataType>
      columns={columns}
      data={[]}
      enableVirtualization={true}
      isLoading={false}
      emptyTableContent='No data.'
    />
  );
};

export default App;

```

## Docs

Added patch react changeset.

### DIFF
--- a/.changeset/quiet-kangaroos-teach.md
+++ b/.changeset/quiet-kangaroos-teach.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed issue where `emptyTableContent` would not appear on virutalized `Table` components.
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -1153,7 +1153,7 @@ export const Table = <
           }
         >
           <ShadowRoot>
-            {enableVirtualization ? (
+            {enableVirtualization && data.length !== 0 ? (
               <div
                 style={{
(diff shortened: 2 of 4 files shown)


mayank99 (approved):
LGTM!

Ben-Pusey-Bentley:
> LGTM!

Should I also add Jon as a reviewer and get his approval for this since Rohan is currently out, or is it fine to merge this with one approval?

mayank99 (commented):
Lets merge it. This is purely React so Jon wouldn't be much help here.

PR #2163, opened 2024-07-23
use only `layout` containment in virtualizers

mayank99:
## Changes

This fixes horizontal scroll in virtualized `Table`.

My understanding is that when using [`paint` containment](https://css-tricks.com/lets-take-a-deep-dive-into-the-css-contain-property/#aa-paint-containment), the browser does not paint any elements that are overflowing or aren't initially visible. Removing it fixes the issue. (`contain: strict` is short for `contain: layout paint size`). I wanted to use `contain: layout size` but it was causing issues in Firefox, so I went with only `contain: layout`. This does mean we lose the performance benefits of `contain: size`, but at least the layout is isolated.

**Note**: I also made the change in `ComboBox` and `Tree`, since all three components are supposed to work similarly (and it might help with [horizontal scroll issues in `Tree`](https://github.com/iTwin/iTwinUI/discussions/2146)).

**Note 2**: I also added `position: relative`. This is supposed to be implied when using `contain`, but I noticed some [weird behavior in Safari](https://github.com/user-attachments/assets/d178d7c6-9857-47c1-af71-d1c4e410a7a1) which gets fixed when using it explicitly.

## Testing

Manually tested that the [Full2 story](https://itwin.github.io/iTwinUI/2163/react/?story=table--full2&theme=dark) looks fine.

Updated the existing visual test to take a screenshot after horizontally scrolling.

## Docs

Added changeset.


Ben-Pusey-Bentley (approved):
LGTM! I have tested out the `Tree` issues with this, and this should help. Will also need to remove the `width: 100%` from the `Tree` virtual items and make some other adjustments to the virtualized `Tree`. Currently, when I can get the horizontal scroll to appear, it appears on the table itself instead of the container like the user wants. Video: 

https://github.com/user-attachments/assets/d73bfd4f-a3e6-4135-9ea8-c92ec56ed74f

Will work some more on this and create a new PR once the fix is ready.

PR #2166, opened 2024-07-23
move `contain` to virtual-scroll parent

mayank99:
## Changes

Builds upon #2165. Handles a TODO item from #2092.

Instead of `contain: layout` being set on the internal virtualizer element, it is now set on the parent element, just as originally intended.

The reason we're able to make this change now is that we no longer use `contain: size`, which was causing the parent element to sometimes collapse to 0x0 dimensions. `contain: layout` does not affect size, so it is safe to use on the outer element.

I also added `background-color` after reading [this](https://css-tricks.com/lets-take-a-deep-dive-into-the-css-contain-property/#aa-an-interesting-accidental-discovery). While `Tree` and `ComboBox` already have a `background-color` specified in CSS, this becomes useful for `Table` whose body doesn't have a background.

## Testing

Manually tested the stories of all three components and inspected them in dev-tools to verify that `:host` styles are present.

I didn't notice any difference in behavior, though I admittedly haven't tried to measure performance before/after these changes.

## Docs

N/A. Not sure that this needs a changeset?

### DIFF
--- a/packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
+++ b/packages/itwinui-react/src/core/ComboBox/ComboBoxMenu.tsx
@@ -58,7 +58,7 @@ const VirtualizedComboBoxMenu = (props: React.ComponentProps<'div'>) => {
     );
   }, [focusedIndex, menuRef]);
 
-  const virtualizer = useVirtualScroll({
+  const { virtualizer, css: virtualizerCss } = useVirtualScroll({
     // 'Fool' useVirtualScroll by passing length 1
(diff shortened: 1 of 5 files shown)


Ben-Pusey-Bentley (approved):
LGTM. The performance seems to be almost identical for the virtualized `Table` before and after this specific PR. Makes more sense to me for the `contain` style to be on the parent instead of the virtualized element. Video of performance(First is this PR, second is main):

https://github.com/user-attachments/assets/a891c595-d9b2-4f48-8213-793bd3208d5a

PR #2170, opened 2024-07-24
Replaces the `width` on `Tree` virtual items with `--_iui-width`

Ben-Pusey-Bentley:
## Changes

Replaces the `width: 100%` style that was added to virtual items on the `Tree` component with `--_iui-width: 100%`. This change preserves the style of the virtualized `Tree` where tree nodes stretched the full width of the `Tree` while allowing nodes to overflow the `Tree`s width. This overflow will allow users to add horizontal scroll to their virtualized `Tree` components. [relevant discussion](https://github.com/iTwin/iTwinUI/discussions/2146). Also fixes a bug where the style prop passed into `TreeNode` for virtualized `Tree` components was not respected.

After PR TODOs:

- [ ] Makes sure that all virtualized `Tree` items have the same width: [discussion](https://github.com/iTwin/iTwinUI/pull/2170#discussion_r1695466900)

## Testing

Added new story and test image for a virtualized `Tree` with horizontal scroll content. Also added a test image for the original virtualized `Tree`. Also confirmed that modification to the user code in the discussion works in the vite playground. Adding `height: 100%` to their `.my-tree` class is the only needed change to get their component working correctly.
 
App.tsx:
```tsx
import { ComponentPropsWithoutRef, useCallback, useState } from 'react';
import { Tree, TreeNode } from '@itwin/itwinui-react';
import './App.css';

function App() {
  return (
    <div
      style={{
        height: 400,
        width: 250,
        overflow: 'hidden',
        border: '1px solid red',
      }}
    >
      <MyTree />
    </div>
  );
}

type NodeData = {
  id: string;
  label: string;
  children: NodeData[];
};

const createNodes = (parentId: string, depth = 0) => {
  return depth === 10
    ? []
    : Array(5)
        .fill(0)
        .map(
          (_, i): NodeData => ({
            id: `${parentId}-${i}`,
            label: `Some long node label to take space ${parentId} ${i}`,
            children: createNodes(`${parentId}-${i}`, depth + 1),
          }),
        );
};
const data = createNodes('root');

type TreeProps<T> = ComponentPropsWithoutRef<typeof Tree<T>>;

function MyTree() {
  const [expanded, setExpanded] = useState<Record<string, boolean>>({});

  const getNode = useCallback<TreeProps<NodeData>['getNode']>(
    (node) => {
      return {
        node,
        nodeId: node.id,
        hasSubNodes: node.children.length > 0,
        subNodes: node.children,
        isExpanded: expanded[node.id],
      };
    },
    [expanded],
  );

  const nodeRenderer = useCallback<TreeProps<NodeData>['nodeRenderer']>(
    ({ node, ...nodeProps }) => {
      return (
        <TreeNode
          {...nodeProps}
          label={node.label}
          onExpanded={(nodeId, isExpanded) => {
            setExpanded((prev) => ({ ...prev, [nodeId]: isExpanded }));
          }}
        />
      );
    },
    [],
  );

  return (
    <div className='my-tree-container'>
      <Tree
        data={data}
        nodeRenderer={nodeRenderer}
        getNode={getNode}
        enableVirtualization={true}
        className='my-tree'
      />
    </div>
  );
}

export default App;
```

App.css:
```css
.my-tree {
  width: fit-content;
  min-width: 100%;
  height: 100%;
}

.my-tree-container {
  width: 100%;
  height: 100%;
  overflow: auto;
}
```

## Docs

Added patch react changesets.


packages/itwinui-react/src/core/Tree/Tree.tsx
```diff
@@ -297,7 +297,7 @@ export const Tree = <T,>(props: TreeProps<T>) => {
                 'data-iui-virtualizer': 'item',
                 ref: virtualizer.measureElement,
                 style: {
-                  width: '100%',
+                  minInlineSize: '100%',
```
mayank99, line 300: Can this be set in the `useVirtualScroll` css or is it not common between the three components?

apps/react-workshop/src/Tree.stories.tsx
```diff
@@ -682,3 +682,122 @@ export const Small = () => {
     />
   );
 };
+
+export const VirtualizedWithHorizontalScroll = () => {
```
mayank99, line 686: Some changes to the story:
- Lets increase the max depth, because that's when horizontal scroll is most useful.
- Lets remove extra features like `sublabel`s and disabled nodes so that the code becomes simpler. These features are not relevant for virtualization or horizontal scroll.

packages/itwinui-react/src/core/Tree/Tree.tsx
mayank99: I'm also wondering if this is a good default. Won't this always result in a horizontal overflow, even when the user wants truncation?

Ben-Pusey-Bentley: I hadn't considered it common since the `TableRow` doesn't include `width` with it's virtualization styles specifically, but it does have `minWidth: 100%` on all rows. Moving the style to `useVirtualScroll` doesn't seem to create any visual changes, so it can be moved there. However, this style does always result in horizontal overflow, like you mentioned. Do you think that it might be better to ask users to add this style to their `TreeNodes` instead, so that truncation remains the default behavior?

mayank99: I think the `Tree` should behave consistently, whether virtualization is enabled or not. So yes, lets ask users to add custom CSS. In other words, just remove `width: 100%` from the inline style?

Ideally, the exact same CSS can be used to create horizontal scroll for both `enableVirtualization={false}` and `enableVirtualization={true}`.

Ben-Pusey-Bentley: I've removed the inline width style in this [commit](https://github.com/iTwin/iTwinUI/pull/2170/commits/82304f067c2acd05f5b852bf8a1cbcd0dc030300). Also updated the stories to add inline styles to the `TreeNode` components, including the `Virtualized` story, since the nodes become much thinner than the `Tree` without `width: 100%`. Screenshot of `Virtualized` story without `width: 100%`:
image

Also fixed a bug with the virtualized `Tree` where user defined inline styles weren't being merge with the styles being added to the virtual items, causing them to be lost.

mayank99: > the nodes become much thinner than the Tree without `width: 100%`

Hang on, this is a problem because it introduces another difference in behavior between virtualized and non-virtualized trees. The virtualized tree-nodes should fill the container by default, while allowing the user to easily override it.

So I'd suggest setting `--_iui-width: 100%` in the inline style instead of removing it altogether. The reason for using this variable over just `width: 100%` is that the user's CSS will be able to still easily set `width`.

After that, the "virtualized with horizontal scroll" story can set `width: fit-content` instead of `minInlineSize: 100%`.

Ben-Pusey-Bentley: Setting `width: fit-content` on the virtualized horizontal scroll story causes the nodes to shrink to their content's width., which causes them to not fill the container. Screenshot: 
image

Is this okay? It feels like it should have `minWidth: 100%` as well, so that the minimum width of the nodes is the container's width.

mayank99: > It feels like it should have `minWidth: 100%` as well, so that the minimum width of the nodes is the container's width.

Yeah we can do both `width: fit-content` and `min-width: 100%`.

apps/react-workshop/src/Tree.stories.tsx
Ben-Pusey-Bentley: Removed the sublabels and disabled nodes, and changed how the sub items are generated so that every node has a deeper sub node tree in this [commit](https://github.com/iTwin/iTwinUI/pull/2170/commits/49e554d67de2d05f1a5cc4f02d286a27f446194c).

packages/itwinui-react/src/core/Tree/Tree.tsx
Ben-Pusey-Bentley: Set the `--_iui-width: 100%` style for the virtual items and updated the react stories accordingly in this [commit](https://github.com/iTwin/iTwinUI/pull/2170/commits/07f0c639a0951e5a00f8b0f06728767e40aa1ca2).

.changeset/perfect-parents-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Replaces the `width: 100%` added to `Tree` virtual items with `minInlineSize: 100%` so that virtualized `Tree` components can have horizontal scroll.
```
mayank99, line 5: Changeset could be rephrased to be less specific.

```suggestion
Fixed an issue in virtualized `Tree` where horizontal scroll was impossible to achieve.
```

Also another changeset could be added to mention the `style` prop fix.

```md
---
'@itwin/itwinui-react': patch
---
Fixed an issue in virtualized `Tree` where the `style` prop passed to `TreeNode` was not being respected.
```

mayank99 (approved):

.changeset/perfect-parents-dress.md
Ben-Pusey-Bentley: Updated changesets in this [commit](https://github.com/iTwin/iTwinUI/pull/2170/commits/55aae4785a28857c9356612b4205c2cb6445a734).

apps/react-workshop/cypress-visual-screenshots/baseline/Tree.test.ts-Virtualized With Horizontal Scroll.png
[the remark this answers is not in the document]
mayank99: That comment is a bit different; it's about stretching to fill the _initial_ container width (i.e. the width of the parent of the scrolling element). Ideally it would also stretch to fill the entire container width (i.e. the width of the scrolling element).

If this requires more work, maybe it's fine to use the current version and we can fix it in a subsequent PR.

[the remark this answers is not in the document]
Ben-Pusey-Bentley: Did some investigating on this, and it seems like the issue is related to the inner `<div>` component not having the width of it's children the way that it did in the previous virtualization implementation. 
image

Tried setting width to various values and seeing if the `<slot>` component might be responsible. The width values did not fix the issue, and removing the `<slot>` didn't fix it either. The best solution that I can currently think of is to pass down some state into the virtual items render function and use ref functions on the items to update that state if the passed width is too short. This solution would also have to be able to react to the scroll element getting narrower, since that's how the old implementation worked. I think this makes more sense as a separate PR. Let me know if you agree, or if you think that I am missing something.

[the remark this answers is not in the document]
Ben-Pusey-Bentley: Okay, I've added the after PR TODO to the PR description.

Issue #888, opened 2022-11-10
Table: layout trashing when scrolling


medikornov:
#### Describe the bug (actual behavior)
When CPU usage is limited due to other tasks, the table's virtualized scrolling is slower and seems to lag (White space appears).
![image](https://user-images.githubusercontent.com/75395229/201165244-de15ec15-8cf3-438f-9443-979264973004.png)
This is an issue because UI is inconsistent and it is visible in production environment.
Video attached.

#### Expected behavior
Scrolling shouldn't cause layout trashing. If it is not fixable, then any suggestions on how to solve it?

#### Reproduction
[Link to a minimal repro: 
](https://codesandbox.io/s/itwinui-react-example-forked-tmxv45?file=/src/App.tsx)
Reduce CPU performance in dev tools -> performance tab to experience similar behaviour.

Video:

https://user-images.githubusercontent.com/75395229/201165024-5051447a-ec0d-4885-abae-2c8bf47dfaeb.mp4

![image](https://user-images.githubusercontent.com/75395229/201165088-48c2e254-2037-4e09-99d3-53e8050e66ba.png)

juliusBaranauskas:
Hi @mayank99, this issue is also affecting my team and users. We are using the table to show long lazy-loaded lists without virtualization but after reaching around 2000 rows browser just crashes. We cannot find any solution to the memory issue from our side and we cannot use virtualization because the performance with it enabled on chrome is plain terrible.

Is there any plan to fix this or release a new version of table sometime soon?

mayank99:
@juliusBaranauskas Can you share more on this?

> we cannot use virtualization because the performance with it enabled on chrome is plain terrible.

In the recent releases, we have reworked virtualization entirely, and also specifically reduced layout thrashing (though it's not completely fixed). What kind of performance issue are you seeing on Chrome?

juliusBaranauskas:
@mayank99 

Attaching a screen capture of how the virtualized table performs for our use case with and without virtualization (localhost has virtualization enabled and itwinui-react v3.11.0)

![virtualizedTableThrashingLocal](https://github.com/user-attachments/assets/64c3f393-c847-43b8-891d-5a3bce9b1ea9)

Below is itwinui-react v3.15.1 (dont mind the styling differences, they are caused by updating itwinui-react)
![virtualizedTableThrashing_v3151](https://github.com/user-attachments/assets/07bb1fea-e434-43a8-b5e9-1ed40cedecf5)

[...]

PR #2178, opened 2024-07-30
Adds `ColumnHeader` component to `Table`

Ben-Pusey-Bentley:
## Changes

Adds a `ColumnHeader` component to the `Table` component. The primary purpose of this change is to make it possible to wrap the column resizing ref function in `React.useCallback` in order to prevent layout thrashing. This PR addresses an after PR TODO from #2092

## Testing

Ran automated tests and confirmed that the new component did not add any visual changes. Compared performance and confirmed that the `Recalculate Style` task performs significantly better after this change.

Before:
Screenshot 2024-07-30 142111

After:
Screenshot 2024-07-30 142045
 
It should be noted that the overall performance of the virtualized `Table` component remains around the same on high levels of throttling. While the rendering itself has become inexpensive, the scripting for the `Table` component seems to take a very long time.
image

## Docs

Added react patch changeset.

### DIFF
--- a/.changeset/lucky-goats-kneel.md
+++ b/.changeset/lucky-goats-kneel.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Reduced layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
+Reduced layout thrashing on `Table` component by memoizing an expensive `ref` function.
--- a/.changeset/lucky-goats-kneel.md
+++ b/.changeset/lucky-goats-kneel.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': patch
 ---
 
-Fixed layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
+Reduced layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
--- a/.changeset/lucky-goats-kneel.md
+++ b/.changeset/lucky-goats-kneel.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
--- a/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
+++ b/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
@@ -4,7 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
 import { Box } from '../../utils/index.js';
-import type { ColumnInstance, HeaderGroup, TableKeyedProps } from 'react-table';
+import type {
+  ColumnInstance,
+  HeaderGroup,
+  TableKeyedProps,
+} from '../../react-table/react-table.js';
 import { SELECTION_CELL_ID } from './columns/index.js';
 import { getCellStyle, getSubRowStyle, getStickyStyle } from './utils.js';
 import cx from 'classnames';
--- a/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
+++ b/packages/itwinui-react/src/core/Table/ColumnHeader.tsx
@@ -17,11 +17,9 @@ type ColumnHeaderProps<
   index: number;
   hasAnySubRows: boolean;
   headers: HeaderGroup<T>[];
-  columnMinWidths: { default: number; withExpander: number };
   isTableResizing: boolean | undefined;
   density: string | undefined;
   visibleColumns: ColumnInstance<T>[];
-  isHeaderDirectClick: React.MutableRefObject<boolean>;
   showSortButton: (column: HeaderGroup<T>) => boolean;
   children: React.ReactNode;
 };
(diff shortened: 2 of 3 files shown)


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1033,30 +1024,15 @@ export const Table = <
                     });
```
mayank99, line 1024: I think all this code above can also be moved into `ColumnHeader`

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1033,30 +1024,15 @@ export const Table = <
...
-                            column.toggleSortBy();
-                          }
-                        }}
+                        onClick={onClick}
+                        columnRefs={columnRefs}
+                        column={column}
+                        isHeaderDirectClick={isHeaderDirectClick}
+                        showSortButton={showSortButton}
                       >
                         <>
```
mayank99, line 1037: All this could below can also probably be moved into `ColumnHeader`. You can do that in a follow-up PR or in the same PR.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Box } from '../../utils/index.js';
+import type {
+  HeaderGroup,
+  TableHeaderProps,
+  TableKeyedProps,
+} from 'react-table';
```
mayank99, line 11: These types should come from `'../../react-table/react-table.js'`.

.changeset/lucky-goats-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
```
mayank99, line 5: Does this mean #888 is completely fixed?

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Moved this code into `ColumnHeader` with this [commit](https://github.com/iTwin/iTwinUI/pull/2178/commits/0198db023ca2cec3cad5eed4b1162ee5b3275cc0).

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
Ben-Pusey-Bentley: Fixed in this [commit](https://github.com/iTwin/iTwinUI/pull/2178/commits/4f3195083be8be54edc49a5862adf7ee21a19ccd).

packages/itwinui-react/src/core/Table/Table.tsx
Ben-Pusey-Bentley: Moved the `ColumnHeader`s children inside of the component in this [commit](https://github.com/iTwin/iTwinUI/pull/2178/commits/e153e92bdcd2cdcff79721d638f2d71f77875f10).

.changeset/lucky-goats-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Reduced layout thrashing on `Table` component by wrapping an expensive ref function in a callback.
```
mayank99, line 5: nit: "callback" is just a generic term; better to use the word "memoize". (`useCallback` is actually more like `useMemoizedCallback`)

```suggestion
Reduced layout thrashing on `Table` component by memoizing an expensive `ref` function.
```

mayank99 (approved):
LGTM! The reduced layout thrashing is a really good outcome 👍

There are ways to clean up the code in `ColumnHeader` to not rely on so many different props, but we can always do that in a future PR

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -987,155 +956,24 @@ export const Table = <
...
                     return (
-                      <Box
-                        {...columnProps}
+                      <ColumnHeader<T>
                         {...column.getDragAndDropProps()}
-                        key={columnProps.key}
-                        title={undefined}
-                        ref={(el) => {
-                          columnResizeRef(el, column);
-                        }}
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: That didn't work because the component is being instantiated inside of a map function, so it isn't at the top level of the component. [discussion](https://github.com/iTwin/iTwinUI/pull/2092#discussion_r1672610769).

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -987,155 +956,24 @@ export const Table = <
...
-                        flexWrap: 'wrap',
-                        columnGap: 'var(--iui-size-xs)',
-                      },
-                    });
-
                     return (
-                      <Box
-                        {...columnProps}
+                      <ColumnHeader<T>
                         {...column.getDragAndDropProps()}
```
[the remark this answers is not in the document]
Ben-Pusey-Bentley: Done in this [commit](https://github.com/iTwin/iTwinUI/pull/2178/commits/b0e168b6ef2baaf58452d26bdd2675fa86600da1).

PR #2184, opened 2024-08-05
Set up tasks for VSCode

smmr-dn:
## Changes

Setup VSCode tasks to help with workflow without having to repeat certain CLI commands.

## Testing
These tasks have been tested and confirmed to work. There are two ways to run these tasks. 

The tasks can be opened on the command top bar.
run-tasks

VSCode Extension `Tasks` to have these tasks displayed and easily accessible in the bottom taskbar.
image

Below is an example after the task `Build` is run.
image

## Docs
`CONTRIBUTION.md` has been updated with instructions to run tasks.

## After PR TODO

- [ ] Edit `createComponent.js` to add the new component in the list of `pickString` of components. (https://github.com/iTwin/iTwinUI/pull/2184#discussion_r1705703548)

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -14,19 +14,19 @@
       "detail": "Build all"
     },
     {
-      "label": "Custom build",
+      "label": "Workspace build",
       "type": "shell",
       "command": "pnpm run build --filter=${input:workspaceName}",
       "detail": "Build specific package or application"
     },
     {
-      "label": "Run tests",
+      "label": "Test",
       "type": "shell",
       "command": "pnpm run test",
-      "detail": "Run all tests without filter"
+      "detail": "Run all tests"
     },
     {
-      "label": "Run tests for a specific component",
+      "label": "Component test",
       "type": "shell",
       "command": "pnpm run test ${input:componentName}",
       "detail": "Run all tests for a specific component"
@@ -38,7 +38,7 @@
       "detail": "Open all development servers"
     },
     {
-      "label": "Run tests with filter",
+      "label": "Test with filter",
       "type": "shell",
       "command": "pnpm run test --filter=${input:testFilter}",
       "detail": "Run different test types"
@@ -52,10 +52,11 @@
       "options": [
         "css-workshop",
         "react-workshop",
-        "portal",
         "@itwin/itwinui-css",
         "@itwin/itwinui-react",
-        "@itwin/itwinui-variables"
+        "@itwin/itwinui-variables",
+        "vite-playground",
+        "e2e"
       ]
     },
     {
@@ -133,11 +134,10 @@
     {
       "type": "pickString",
       "id": "testFilter",
-      "description": "Choose a test type",
+      "description": "Choose a workspace to run tests for",
       "options": [
         "css-workshop",
         "react-workshop",
-        "@itwin/itwinui-css",
         "@itwin/itwinui-react",
         "e2e",
         "a11y"
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -10,25 +10,25 @@
     {
       "label": "Build",
       "type": "shell",
-      "command": "pnpm build",
+      "command": "pnpm run build",
       "detail": "Build all"
     },
     {
       "label": "Custom build",
       "type": "shell",
-      "command": "pnpm build --filter=${input:monorepoName}",
+      "command": "pnpm run build --filter=${input:workspaceName}",
       "detail": "Build specific package or application"
     },
     {
       "label": "Run tests",
       "type": "shell",
-      "command": "pnpm test",
+      "command": "pnpm run test",
       "detail": "Run all tests without filter"
     },
     {
       "label": "Run tests for a specific component",
       "type": "shell",
-      "command": "pnpm test ${input:componentName}",
+      "command": "pnpm run test ${input:componentName}",
       "detail": "Run all tests for a specific component"
     },
     {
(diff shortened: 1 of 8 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,134 @@
...
+      "type": "shell",
+      "command": "pnpm run test --filter=e2e",
+      "detail": "Run e2e tests"
+    }
+  ],
+  "inputs": [
+    {
+      "type": "pickString",
+      "id": "monorepoName",
+      "description": "What monorepo would you like to build?",
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

.vscode/tasks.json
```diff
@@ -0,0 +1,134 @@
...
+      "detail": "Run all tests for a specific component"
+    },
+    {
+      "label": "Open development servers",
+      "type": "shell",
+      "command": "pnpm dev",
+      "detail": "Open all development servers"
+    },
+    {
+      "label": "Run e2e tests",
```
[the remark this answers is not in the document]
smmr-dn: I have added more test options.

.vscode/tasks.json
```diff
@@ -0,0 +1,147 @@
...
+      "type": "pickString",
+      "id": "workspaceName",
+      "description": "Choose a worskspace",
+      "options": [
+        "css-workshop",
+        "react-workshop",
+        "portal",
+        "@itwin/itwinui-css",
+        "@itwin/itwinui-react",
+        "@itwin/itwinui-variables"
```
mayank99, line 58: We should include `vite-playground` and `e2e` in this list, because those are some of the most commonly used workspaces that we'd need to build for. And we can exclude `portal` as it's not important.

.vscode/tasks.json
```diff
@@ -0,0 +1,147 @@
...
+        "TransferList",
+        "Tree",
+        "Typography",
+        "VisuallyHidden"
+      ]
+    },
+    {
+      "type": "pickString",
+      "id": "testFilter",
+      "description": "Choose a test type",
```
mayank99, line 136: nit: This description can be improved to mention that it's asking to pick from a list of _workspaces_ to run tests for. ("test type" is too vague and unhelpful)

.vscode/tasks.json
```diff
@@ -0,0 +1,147 @@
...
+      "options": [
+        "css-workshop",
+        "react-workshop",
+        "@itwin/itwinui-css",
```
mayank99, line 140: I believe `"@itwin/itwinui-css"` can be excluded from this list, because it doesn't have any tests.

Also, I don't know if it's possible but it would be nice if we could associate custom labels with the options. For example, `"css-workshop"` and `"react-workshop"` would be more clearly described as "css-workshop (visual tests)" and "react-workshop (visual tests)".

.vscode/tasks.json
mayank99, line 1: 1. Could you please add some details to the PR description about how you tested that these tasks are working?
   - In general, you should _not_ remove the "Testing" and "Docs" sections from the PR description. In cases where these are genuinely not applicable, you can write "N/A" instead of deleting the section. For examples/inspiration, you can take a look at descriptions from some older PRs (e.g. [#2168](https://github.com/iTwin/iTwinUI/pull/2168), [#2106](https://github.com/iTwin/iTwinUI/pull/2106), [#1903](https://github.com/iTwin/iTwinUI/pull/1903)).

2. Can you briefly mention how to run tasks somewhere in `CONTRIBUTING.md`? Right now, contributors might not even realize that this repo has tasks. (This change should also be mentioned in the "Docs" section of the PR description)

.vscode/tasks.json
smmr-dn: I have made the change.

.vscode/tasks.json
smmr-dn: I have made the change.

.vscode/tasks.json
smmr-dn: I have found a way to add labels yet so in the meantime, I would include the description between the parentheses.

.vscode/tasks.json
smmr-dn: I will keep this in mind, and I have updated the PR description.

.vscode/tasks.json
```diff
@@ -0,0 +1,147 @@
...
+      "detail": "Build all"
+    },
+    {
+      "label": "Custom build",
+      "type": "shell",
+      "command": "pnpm run build --filter=${input:workspaceName}",
+      "detail": "Build specific package or application"
+    },
+    {
+      "label": "Run tests",
```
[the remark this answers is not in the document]
smmr-dn: I have made the change for these commands.

.vscode/tasks.json
smmr-dn: > (Edited) I haven't found a way to add labels yet so in the meantime, I would include the description between the parentheses.

Sorry, I think I made a typo yesterday. I meant I haven't found a way to add labels yet so I might have to just go with parentheses for now.

mayank99: This doesn't work, because the parentheticals will be passed verbatim into the terminal command.

mayank99: This doesn't work, because the parentheticals will be passed verbatim into the terminal command.



---

If you look at the [docs for input variables](https://code.visualstudio.com/docs/editor/variables-reference#_input-variables), it states this:

> An option can be a string value or an object with both a label and value. The dropdown will display label: value.

So I believe you should be able something like this:

```json
      "options": [
        { "label": "visual tests", "value": "css-workshop" },
        { "label": "visual tests", "value": "react-workshop" },
        { "label": "unit tests", "value": "@itwin/itwinui-react" },
        { "label": "end-to-end tests", "value": "e2e" },
        { "label": "accessibility tests", "value": "a11y" }
      ]
```

(feel free to copy paste this)

smmr-dn: Oh, I totally missed that while going through the doc. I have made the change.

apps/css-workshop/src/components/Radio.astro
mayank99, line 1: I think you meant to push this change to a different branch.

mayank99 (approved):
LGTM!

PR #2191, opened 2024-08-09
Add Astro Radio Component

smmr-dn:
## Changes

This PR created an Astro `Radio_` component for our CSS workshop and applied it wherever this component was used.

## Testing

All visual tests passed.

## Docs

N/A


apps/css-workshop/src/pages/toast.astro
```diff
@@ -47,27 +48,27 @@ import IconButton_ from '../components/IconButton.astro';
     <div class='placement-control'>
       <h3>Toast placement:</h3>
       <label class='placement-choice' id='top-start-option'>
-        <input type='radio' name='radio-demo-6' value='top-start' />
+        <Radio_ name='radio-demo-6' value='top-start' />
```
mayank99, line 51: These radios do not rely on our CSS (there is no `"iui-radio"` class) so they do not need to be componentized. You can revert the changes in this file, and simplify `Radio.astro` to remove the `basic` handling.

smmr-dn: I have made the change.

mayank99: Can you review the current state of your own PR to make sure there are no leftovers in `toast.astro`?

Also, the `inputProps` in `Radio.astro` does not need to be a variable anymore; the props can be used directly on the `<input>`

apps/css-workshop/src/pages/radio.astro
```diff
@@ -388,45 +224,27 @@ import Icon_ from '../components/Icon.astro';
         <span class='iui-radio-label'>Option 1</span>
         <input class='iui-radio' type='radio' name='radio-demo-12' value='option-1' checked />
       </label>
-
       <label class='iui-radio-wrapper'>
         <span class='iui-radio-label'>Option 2</span>
         <input class='iui-radio' type='radio' name='radio-demo-12' value='option-2' />
       </label>
```
FlyersPh9, line 230: Is there no way to handle radio on the right hand side with the Astro component?

smmr-dn: I have added a flag prop called `isFlipped` (I feel like it should have a better name) so the `input` and `span` label can be swapped in the radio on right layout. Not sure if this is the right way to do it.

apps/css-workshop/src/pages/toast.astro
smmr-dn: I have checked `toast` page and other pages and cleaned up the `Radio` component.

apps/css-workshop/src/pages/radio.astro
FlyersPh9: Since these Astro components are for our use only I am ok with that prop name, but I'll leave that decision up to @mayank99.

mayank99: In `ToggleSwitch`, we call it `labelPosition` so we could reuse that name.

https://github.com/iTwin/iTwinUI/blob/c977e9f39d3c05ce8278d0d3b72df3a7c5ee662a/packages/itwinui-react/src/core/ToggleSwitch/ToggleSwitch.tsx#L23

As for the code, I think the `<input>` can stay where it is. Only the `<span>` needs to be conditionally rendered before or after the `<input>`.

mayank99 (approved):
Pre-approving this, assuming https://github.com/iTwin/iTwinUI/pull/2191#discussion_r1719860218 is taken care of.

apps/css-workshop/src/components/Radio.astro
```diff
@@ -0,0 +1,43 @@
...
+        <span class='iui-radio-label'>
+          <slot />
+        </span>
+      </>
+    )
+  }
+  <input class:list={['iui-radio', className]} type='radio' disabled={disabled} {...props} />
+  {
+    labelPosition === 'right' ||
+      (!labelPosition && (
```
mayank99, line 35: This if condition looks wrong. When `labelPosition` is set to `'right'`, it will *not* show this label.

Please wait for a second approval before merging the PR next time. Such mistakes would be easily caught by a second review.

Additionally please try to respond to the comments after committing code. If you don't respond, the reviewers may not even realize that you acted on their comments.

smmr-dn: I apologize for this and will wait for another round of review next time as well as make sure to notify every commentor about changes. I will open another PR to fix this issue. I will open another PR to fix this issue.

PR #2192, opened 2024-08-12
Fix menu item hover state in CSS demo page

FlyersPh9:
## Changes

- Remove unused class `.iui-menu-item` from CSS.
- Add `data-iui-actionable='true'` to list items in menu CSS demo page to fix hover issue.

## Testing

Checked hover state manually in both CSS and React demo pages.

## Docs

n/a

### DIFF
--- a/.changeset/fair-cats-tickle.md
+++ b/.changeset/fair-cats-tickle.md
@@ -1,5 +1,0 @@
----
-'@itwin/itwinui-css': patch
----
-
-Remove unused CSS class.
--- a/.changeset/fair-cats-tickle.md
+++ b/.changeset/fair-cats-tickle.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-css': patch
 ---
 
-Fix hover state on menu item by targeting the correct class name.
+Remove unused CSS class.
--- a/.changeset/fair-cats-tickle.md
+++ b/.changeset/fair-cats-tickle.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Fix hover state on menu item by targeting the correct class name.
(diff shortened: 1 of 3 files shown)


packages/itwinui-css/src/menu/mixins.scss
```diff
@@ -51,7 +51,7 @@ $iui-active-outline: 1px solid var(--iui-color-border-accent);
     margin-block-start: -1px;
   }
 
-  &:where([data-iui-actionable='true'], .iui-menu-item):where(:hover),
+  &:where([data-iui-actionable='true'], .iui-list-item):where(:hover),
```
mayank99, line 54: I believe this was just leftover from #1295 so it should be removed altogether.

```suggestion
  &:where([data-iui-actionable='true']:hover),
```

By default, `iui-list-item` should *not* have a hover state. See [docs](https://itwinui.bentley.com/docs/list#actionable).

To fix the `menu.html` page, `data-iui-actionable="true"` should be added to all instances of `<div class="iui-list-item">`. See [React implementation](https://github.com/iTwin/iTwinUI/blob/e1045a8ea4e2d36f4b67342e50dfb8278c264217/packages/itwinui-react/src/core/Menu/MenuItem.tsx#L148-L152) for comparison.

FlyersPh9: Ah, this makes sense.  Committed your suggestion and updated the CSS demo page to use `data-iui-actionable='true'`.

.changeset/fair-cats-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': patch
+---
+
+Remove unused CSS class.
```
mayank99, line 5: I think we can just omit this changeset, it's too small to list out.

mayank99 (approved):

.changeset/fair-cats-tickle.md
FlyersPh9: Sounds good, deleted.

smmr-dn (approved):

PR #2193, opened 2024-08-15
Add Astro ListItem Component

smmr-dn:
## Changes

This PR created an Astro `ListItem_` component as list items for our CSS workshop and applied it wherever this component was used, mainly in `menu.astro`. List content is componentized into named slots.

## Testing

All visual tests passed.

N/A

## Docs

N/A

## After-PR TODO

- [ ] Remove all attribute classes from CSS and HTML (such as `iui-large` or `iui-active`) and use data attributes instead. Comment https://github.com/iTwin/iTwinUI/pull/2193#discussion_r1742396416


apps/css-workshop/src/components/Menu.astro
mayank99, line 1: This is a `ListItem` component, not a `Menu` component. You can compare it to the `ListItem` and `MenuItem` components in React.

apps/css-workshop/src/components/Menu.astro
```diff
@@ -0,0 +1,68 @@
...
+  class:list={[
+    { 'iui-list-item': !isSkeleton },
+    { 'iui-menu-item': isSkeleton },
+    { 'iui-active': isActive },
+    { 'iui-disabled': disabled },
+    { 'iui-menu-item-skeleton': isSkeleton },
+    { 'iui-large': isLarge },
+    className,
+  ]}
+  aria-checked={checked}
```
mayank99, line 35: `aria-checked` is not valid on list items; see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#associated_roles).

pnpm-lock.yaml
mayank99, line 1: There shouldn't be any lockfile changes in this PR. You can revert it using this command:

```
git checkout origin/main pnpm-lock.yaml
```

apps/css-workshop/src/pages/menu.astro
```diff
@@ -38,239 +39,134 @@ import Icon_ from '../components/Icon.astro';
...
-      </li>
-      <li class='iui-list-item' data-iui-actionable='true' role='menuitem' tabindex='0'>
-        <span class='iui-list-item-content'>Option 2</span>
-      </li>
-      <li class='iui-list-item iui-disabled' data-iui-actionable='true'>
-        <span class='iui-list-item-content'>Option 3</span>
-      </li>
+      <Menu_ tabindex='0' id='test-menu-1' role='menuitem'>Option 1</Menu_>
+      <Menu_ tabindex='0' role='menuitem'>Option 2</Menu_>
+      <Menu_ disabled>Option 3</Menu_>
```
mayank99, line 44: All items should have `role='menuitem'`. Some of them are currently missing it.

apps/css-workshop/src/components/Menu.astro
```diff
@@ -0,0 +1,68 @@
+---
+type Props = {
+  isActive?: boolean;
+  checked?: boolean;
+  isLarge?: boolean;
+  disabled?: boolean;
+  isSkeleton?: boolean;
+  iconStyle?: string;
+  iconClassName?: string;
```
mayank99, line 9: Instead of `iconStyle` and `iconClassName`, you can have a single prop that does both. And it should be named appropriately, as `startIconProps`.

```suggestion
  startIconProps?: astroHTML.JSX.HTMLAttributes;
```

There should probably also be `endIconProps`.

apps/css-workshop/src/pages/menu.astro
```diff
@@ -38,239 +39,134 @@ import Icon_ from '../components/Icon.astro';
...
-        class='iui-list-item iui-active'
-        data-iui-actionable='true'
-        role='menuitem'
-        aria-checked='true'
-        tabindex='0'
-      >
-        <span class='iui-list-item-content'>Option 3</span>
-        <svg-checkmark class='iui-list-item-icon' aria-hidden='true'></svg-checkmark>
-      </li>
+      <Menu_ isActive tabindex='0' role='menuitem' checked>
```
mayank99, line 50: `checked` does not make sense here, was it accidentally added?

pnpm-lock.yaml
smmr-dn: I have reverted `pnpm-lock.yaml`.

apps/css-workshop/src/pages/menu.astro
smmr-dn: I have added `role=menuitem` to all items.

apps/css-workshop/src/pages/menu.astro
smmr-dn: Right, some `checked`'s were mistakenly placed. I have removed the redundant ones.

apps/css-workshop/src/components/Menu.astro
smmr-dn: Noted this and I will look more into the differences between the two.

apps/css-workshop/src/components/Menu.astro
smmr-dn: I have removed all `aria-checked` from the items.

apps/css-workshop/src/components/Menu.astro
smmr-dn: I have switched to using `ListItem` and for skeletons, I have created a `MenuItemSkeleton` as structured similarly in the React package. I'm still a little iffy about as `ListItem` also contains `ListItem.Icon`, `ListItem.Description`, `ListItem.Content` and `ListItem.Action`, should I separate these all into different Astro components and combine them in a single component called `MenuItem` instead of using slots like what I'm currently doing?

apps/css-workshop/src/components/Menu.astro
smmr-dn: I have combined the props into the `startIconProps` and `endIconProps`.

apps/css-workshop/src/components/Menu.astro
mayank99: `MenuItemSkeleton` is not necessary I think. We can use `ListItem` and add any extra classes on top of it. If we decide that we need to make a` MenuItemSkeleton` component in the future, we can always do it in another PR.

As for all the subcomponents of `ListItem`, those are also not necessary. We can add props instead (e.g. `<ListItem icon={…} description={…}>`. The subcomponents are only there to provide full customizability to the user.

apps/css-workshop/src/components/Menu.astro
mayank99: I don't see `startIconProps` and `endIconProps` in your code. I see `iconProps` but it is not even used fully (only `iconProps.class` and `iconProps.style` are being used).

It needs to be spread like this:

```jsx
<span
  aria-hidden='true'
  {...startIconProps}
  class:list={['iui-list-item-icon', startIconProps?.class]}
>
```

apps/css-workshop/src/components/Menu.astro
smmr-dn: I have removed `MenuItemSkeleton` and used `ListItem` for skeletons instead.

mayank99: I think you misunderstood. The `isSkeleton` prop should not exist. `ListItem` is a lower-level component that does not need to be aware of specific features like skeleton. 

I was suggesting passing all skeleton-related classes from outside. e.g. in `menu.astro`:

```astro
<ListItem
  class="iui-menu-item-skeleton"
  startIconProps={{ class: "iui-icon" }}
  contentProps={{ class: "iui-content" }}
```

smmr-dn: I apologize for the misunderstanding. I have added `contentProps` as you suggested. However, I noticed that the skeletons' classes are all "iui-menu-item iui-menu-item-skeleton" while the others are all "iui-list-item". I'm not quite sure how I should handle this case.

mayank99: I think `"iui-list-item iui-menu-item-skeleton"` should also work. The `"iui-menu-item"` class is probably leftover/dead code.

You can leave `"iui-list-item"` inside `ListItem.astro` and only pass `"iui-menu-item-skeleton"` in `menu.astro` like suggested in my previous comment.

apps/css-workshop/src/pages/menu.astro
```diff
@@ -38,239 +39,249 @@ import Icon_ from '../components/Icon.astro';
...
+      <ListItem_
+        contentProps={{ class: 'iui-list-item-content' }}
+        isActive
+        actionable
+        tabindex='0'
         role='menuitem'
-        aria-checked='true'
+      >
+        Option 1
+        <svg-checkmark slot='end-icon' aria-checked='true'></svg-checkmark>
```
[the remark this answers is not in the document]
smmr-dn: Right, this is my mistake when adding and removing things.

apps/css-workshop/src/pages/menu.astro
```diff
@@ -490,41 +507,44 @@ import Icon_ from '../components/Icon.astro';
...
+        <div>Model Lorem ipsum dolor sit amet, consectetur adipiscing elit</div>
+        <div class='iui-list-item-description'>0x0192837465</div>
+      </ListItem_>
       <li class='iui-menu-divider' role='separator'></li>
-      <li class='iui-list-item' data-iui-actionable='true' role='menuitem' tabindex='0'>
-        <span class='iui-list-item-content'>
-          <div>My projects</div>
-        </span>
-      </li>
+      <ListItem_ actionable tabindex='0' role='menuitem'>My projects</ListItem_>
```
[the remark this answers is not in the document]
smmr-dn: I have added the `contentProps` here. I think in the skeletons `iui-content` is used instead of `iui-list-item-content`.

apps/css-workshop/src/pages/menu.astro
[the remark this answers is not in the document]
smmr-dn: I have had the redundant `aria-checked` removed.

apps/css-workshop/src/components/Menu.astro
mayank99: looks like tests were fixed by [`a9235d1`](https://github.com/iTwin/iTwinUI/pull/2193/commits/a9235d12e960cc6bf548401e5447ee349e51b34a)

mayank99 (approved):
LGTM! Make sure to update PR title and description to reflect the latest changes.

apps/css-workshop/src/components/ListItem.astro
```diff
@@ -0,0 +1,57 @@
...
+    { 'iui-disabled': disabled },
+    { 'iui-large': size === 'large' },
+    className,
+  ]}
+  data-iui-actionable={actionable ? 'true' : undefined}
+  {...props}
+>
+  {
+    Astro.slots.has('start-icon') && (
+      <span class={startIconProps?.class} {...startIconProps} aria-hidden='true'>
```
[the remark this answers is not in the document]
smmr-dn: I tried with slots alone and the icons seem off-centered and greyed out. I checked the devtools and realized the styling that comes with the base class `iui-list-item-icon` as well as other attributes are gone. I'm assuming once the svg is assigned to the slot, it overwrites the existing slot's attributes by itself.
image

apps/css-workshop/src/pages/menu.astro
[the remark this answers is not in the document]
smmr-dn: I noticed these `tabindex` in the `MenuItem` React component are set to 0 based on whether the menu item is selected or not. Not sure if these here in `menu.astro` are the same thing.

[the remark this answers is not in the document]
mayank99: Menu is supposed to use [roving tabindex](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#technique_1_roving_tabindex) to enable arrow-key navigation. Since we do not have such JS in the `menu.astro` page, we were using `tabindex=0` everywhere. You can either re-add `tabindex=0` to ensure that we can test visual focus states, or write some JS to implement roving tabindex.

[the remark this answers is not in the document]
smmr-dn: I have added `tabindex=0` back to the items.

apps/css-workshop/src/components/ListItem.astro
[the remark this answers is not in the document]
smmr-dn: I'm okay with this idea 👍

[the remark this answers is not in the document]
mayank99: It's not necessary, at least not right now. The main purpose of using Astro is to expose components for all the CSS classes. `MenuItem` has the exact same styles as `ListItem` (there is no `"iui-menu-item"` class), so it would do nothing useful.

Also, the `MenuItem` React component is supposed to be deprecated in the near future.

apps/css-workshop/src/pages/menu.astro
```diff
@@ -1082,58 +1275,66 @@ import Icon_ from '../components/Icon.astro';
...
+        <svg-placeholder slot='start-icon'></svg-placeholder>
+        <div>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
+          ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+          ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in
+          reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur
+          sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id
+          est laborum.
+        </div>
+        <div class='iui-list-item-description'>
```
mayank99, line 1319: Looks like description is not being handled by `ListItem` right now

smmr-dn: I have put description and list content base class inside `ListItem`.

apps/css-workshop/src/components/ListItem.astro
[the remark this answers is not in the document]
mayank99: Can we please not block this PR? What we have right now is definitely not an ideal situation, but it's ok for this PR to not be perfect, especially since we don't have all the answers right now. This PR is still making very important improvements that make it much easier to maintain our CSS demo pages. If we want to add more components on top of `ListItem`, we can always do that in future PRs.

apps/css-workshop/src/components/ListItem.astro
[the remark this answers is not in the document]
mayank99: The wrapper is fine, and it's actually more consistent with the React implementation. There are two differences:
1. We use `<div>`s, not `<span>`s in the React version. We can use `<div>`s in Astro too, there is no good reason to use `<span>`s anywhere in this component.
2. We directly use `<svg>`s (not `<svg-placeholder>` and similar custom elements) in the React version. This is just one of those differences between the `icons-react` and `icons-elements` libraries - there is no way around it and we need to support both.

| Astro | React |
| --- | --- |
|  |  |

apps/css-workshop/src/components/ListItem.astro
```diff
@@ -0,0 +1,69 @@
...
+  ...props
+} = Astro.props;
+---
+
+<li
+  class:list={[
+    'iui-list-item',
+    { 'iui-active': isActive },
+    { 'iui-disabled': disabled },
+    { 'iui-large': size === 'large' },
```
mayank99, line 32: In a future PR, we should remove all these classes from the CSS and HTML, and instead use data attributes (similar to how it is done in `ListItem.tsx`).

The current version of the CSS supports both classes and data attributes, but that was mainly done for backwards compatibility (which is no longer a concern).

This can be mentioned in the PR description as an "after-PR TODO".

smmr-dn: I have added this as an after-PR TODO. I apologize for the late response since I have stopped receiving comment notification emails for some reason but I should have checked PRs often.

PR #2196, opened 2024-08-20
allow chaining polymorphic components

mayank99:
## Changes

This updates the `polymorphic` implementation so that the `as` prop can accept an array of components. Each of the components in the array will be rendered sequentially, forming a chain of components that build upon each other.

This array implementation is completely hidden from users, since they should not be passing arrays into `as`. Think of it as an internal feature to help to us compose multiple primitives together.

This is needed to address a TODO item from #2021 (see [thread](https://github.com/iTwin/iTwinUI/pull/2021#discussion_r1643449120)). We are now able to use `ButtonBase` and `ListItem` together.

(Also related but not fully addressed: https://github.com/iTwin/iTwinUI/pull/2135#discussion_r1672595007).

Further thoughts: The `as` prop seems to have fallen out of fashion in the React world. The alternatives are [`asChild`](https://www.radix-ui.com/primitives/docs/guides/composition) and [`render`](https://ariakit.org/guide/composition) props, both of which allow chaining components without any TS issues. However, it is too late to change our component API.

## Testing

Pending.

## Docs

N/A


packages/itwinui-react/src/core/Menu/MenuItem.tsx
```diff
@@ -206,6 +195,11 @@ if (process.env.NODE_ENV === 'development') {
   MenuItem.displayName = 'MenuItem';
 }
 
+const MenuItemBase = polymorphic(
+  ButtonBase,
+  ListItem,
+) as PolymorphicForwardRefComponent<'button', ListItemOwnProps>;
```
mayank99, line 201: @r100-stack Do you think this `polymorphic` call makes sense?

An alternative would be to do this:

```jsx
<ListItem
  as={[ButtonBase, asProp] as any}
  // …
/>
```

In both cases, it's difficult or impossible to get proper types, but the `polymorphic` call at least allows defining custom types.

mayank99: > The two approaches look good from a usage perspective! 

I'm asking if it makes sense to keep `polymorphic` or do we want to use arrays. I'm leaning towards the latter to simplify things.

> In the above example, what will be the rendered element? `li`, `button`, or `asProp`?

The last one should win. `ListItem` will first render `ButtonBase` which will then render `asProp` if passed. If there is no `asProp`, then `ButtonBase` will render its own element (i.e. `<button>`).

> Could you elaborate about the types you mentioned?

When using the `as` prop, normally the prop types are automatically merged. So `<ListItem as={ButtonBase}>` provides autocomplete for props from both `ListItem` and `ButtonBase`.

But since there can be any number of components in the `as` chain, we cannot automatically merge the types. But since it is for internal implementation, I think that's ok. We can live without the autocomplete or use type assertions when necessary.

> props exposed will be a combination of a native button + ListItemOwnProps but without the props from `ButtonBase`?

Correct. We are explicitly setting the types, so if we want `ButtonBase` props, we'll have to add it manually.

> Does this approach help in the case where two props have the same name but different types?

It doesn't. For that we'd have to use a different approach, like the `asChild` or `render` prop approaches mentioned in the PR description.

PR #2202, opened 2024-08-22
encourage using the `label` prop in `IconButton`

mayank99:
## Changes

- When `title` prop is passed, it will be rerouted to `label`. (related: #2182)
- When no `label`/`aria-label` is found, a dev-only warning will be shown.

This change helps nudge users to make their `IconButton`s more accessible. Ideally, the `label` prop would just be required, but that would be a breaking change, so this is the next best solution.

## Testing

Manually verified the behavior in vite playground. `title` and `label` behave identically, and a warning is correctly displayed when there is no `label`.

Added unit test to verify that `title` is not present in the DOM.

## Docs

Updated docs for `IconButton` to mention the `label` prop.

Added `patch` changesets for both code changes.

## After-PR TODO

- [ ] Fix `createWarningLogger` (see [comment](https://github.com/iTwin/iTwinUI/pull/2202#discussion_r1729170659)).

### DIFF
--- a/packages/itwinui-react/src/core/Buttons/IconButton.tsx
+++ b/packages/itwinui-react/src/core/Buttons/IconButton.tsx
@@ -4,13 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 import cx from 'classnames';
 import * as React from 'react';
-import { Box, ButtonBase } from '../../utils/index.js';
+import { Box, ButtonBase, createWarningLogger } from '../../utils/index.js';
 import { Tooltip } from '../Tooltip/Tooltip.js';
 import type { ButtonProps } from './Button.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 import { VisuallyHidden } from '../VisuallyHidden/VisuallyHidden.js';
 import { ButtonGroupContext } from '../ButtonGroup/ButtonGroup.js';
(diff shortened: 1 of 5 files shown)


packages/itwinui-react/src/core/Buttons/IconButton.tsx
```diff
@@ -4,14 +4,16 @@
...
-import { Box, ButtonBase } from '../../utils/index.js';
+import { Box, ButtonBase, createWarningLogger } from '../../utils/index.js';
 import { Tooltip } from '../Tooltip/Tooltip.js';
 import type { ButtonProps } from './Button.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 import { VisuallyHidden } from '../VisuallyHidden/VisuallyHidden.js';
 import { ButtonGroupContext } from '../ButtonGroup/ButtonGroup.js';
 import { PopoverOpenContext } from '../Popover/Popover.js';
 
+const logWarning = createWarningLogger();
```
mayank99, line 15: There is a slight issue with our `createWarningLogger` utility. It only runs once. Not once per instance of component but once per the file where it is called.

This issue is not specific to this PR, but in the future we should fix it to run once per instance of a component, so that two different `<IconButton>`s will trigger two different warnings.

smmr-dn (approved):

PR #2213, opened 2024-08-28
create `useWarningLogger` hook

mayank99:
## Changes

Fixes https://github.com/iTwin/iTwinUI/pull/2202#discussion_r1729170659

I've converted `createWarningLogger` into a `useWarningLogger` hook, and made some changes along the way.
1. Immediately I noticed that warnings were showing up twice, so I added a workaround using `setTimeout` to only show them once.
2. I also noticed that `console.warn` doesn't allow looking at the stack trace, so I changed it to `console.error`. This also makes the warnings more prominent and better distinguishes them from the browser's own warnings.
3. I then noticed that the stack trace doesn't include the original component name that's calling our component, so I added some very cursed code to get the component name. I think it is fine because this code is only present during development and property access is guarded using `?.`, so it will not crash even if React internals change.

## Testing

Tested that warnings are showing in the vite playground.

```jsx
const App = () => {
  return (
    <>
      <IconButton>🥝</IconButton>
      <Test />
    </>
  );
};

const Test = () => {
  return <IconButton>💩</IconButton>;
};
```

</details> 

## Docs

Added `patch` changeset.

### DIFF
--- a/packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
+++ b/packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
@@ -6,6 +6,8 @@
 import * as React from 'react';
 import { isUnitTest } from '../functions/dev.js';
 
+const internalsKey = '__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED';
+
 /**
  * Returns a function that can be used to log one-time warnings in dev environments.
  *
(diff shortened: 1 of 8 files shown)


packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        // https://stackoverflow.com/a/71685253
+        const stack =
+          ReactInternals?.ReactDebugCurrentFrame?.getCurrentStack?.();
+
+        // Second line in the stack is the component name.
+        const componentName = stack?.trim().split('\n')[1]?.trim();
+
+        const prefix = componentName
+          ? `Warning (${componentName}):`
```
mayank99, line 41: One of the core maintainers of React [said](https://redirect.github.com/reactjs/react.dev/issues/3896#issuecomment-914036200):

> If you use `console.error` and the user has React DevTools then they should already see the component stack in development automatically. That’s the intended solution.

It would be nice to not rely on these hacks if it works out-of-the-box. However, the component stack was not showing up for me. It was identical with or without React Dev Tools.



Can someone else also confirm this?

packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        React.useEffect(() => {
+          // Clearing timeout on unmount to avoid double execution in StrictMode.
+          // The warning should be logged only once per component instance.
+          return () => {
+            if (timeoutRef.current) {
+              window.clearTimeout(timeoutRef.current);
+            }
+          };
+        }, []);
```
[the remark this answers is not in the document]
mayank99: Not sure what kind of boolean flag you're thinking of, but everything in the component gets reset when it unmounts. And if we move it out of the component, then we'd have the same problem of it not being unique per component instance.

packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
mayank99: Maybe it used to work but no longer does, in which case we can't rely on it.

I'm ok with this workaround for now. It's worth noting that it only works for React 18. We'll need to investigate how to make it work in React 19 when we add support for it.

smmr-dn (approved):

PR #2214, opened 2024-08-29
Add `#region` in long files like `Tabs`

r100-stack:
## Changes

Since Tabs is a big file (850+ lines), it could be hard to follow, even with the "Outline" panel in VSCode. Thus, this quick nit PR just adds some `#region` + `#endregion` comments to make the regions show in the minimap and also make the regions collapsable.

<details>
<summary>Screenshots</summary>

image

image

</details>

We could consider adding regions to other similarly large files with subcomponents (e.g. Panels from #2001), or we can consider splitting them across different files (iirc, I think we decided against that for some reason in Tabs).

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Tabs/Tabs.tsx
+++ b/packages/itwinui-react/src/core/Tabs/Tabs.tsx
@@ -22,7 +22,7 @@ import { Icon } from '../Icon/Icon.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 
 // ----------------------------------------------------------------------------
-// TabsWrapper
+// #region TabsWrapper
 
 type TabsOrientationProps =
   | {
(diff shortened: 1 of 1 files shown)


mayank99 (approved):
I'm fine with this change if you find it helpful, but I have mixed feelings on this and would not want to _require_ this verbose set of comments to be used everywhere.

- Lately I've just been leaning towards *only* using the separator comments, since it's super low maintenance (comments don't need to be kept up-to-date with code). See [`ThemeProvider.tsx`](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx) for example.
- I also don't like that we now need an extra `#endregion` comment which only exists for the sake of VSCode.

smmr-dn (approved):

PR #2218, opened 2024-09-02
Dont add line clamp, if column id is not defined

veekeys:
## Changes

~~Added validation if column id is present before checking for `Cell` property. If column has no id, we cannot know about the `Cell` property. As a result, not adding line clamp styling.~~

Fixed the column `id` check to use `instance.columns` instead of `props.column`. This ensures that columns always have ids (`react-table` generates a backup id if the user-specified column object does not have one).

This fixes an issue where a column with a `Cell` but without an `id` wasn't being detected properly.

## Testing

Added unit test;

~~TODO: validate in storybook, failing local build for now.~~

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/core/Table/cells/DefaultCell.tsx
+++ b/packages/itwinui-react/src/core/Table/cells/DefaultCell.tsx
@@ -48,7 +48,6 @@ export const DefaultCell = <T extends Record<string, unknown>>(
   const columnsProp = React.useContext(TableColumnsContext);
   const isCustomCell = React.useMemo(
     () =>
-      !!props.cellProps.column.id &&
       columnsProp
         .find(({ id }) => props.cellProps.column.id === id)
         ?.hasOwnProperty('Cell'),
@@ -69,7 +68,9 @@ export const DefaultCell = <T extends Record<string, unknown>>(
     className,
     style,
     status,
-    clamp = typeof cellProps.value === 'string' && !isCustomCell,
+    clamp = typeof cellProps.value === 'string' &&
+      !!props.cellProps.column.id &&
+      !isCustomCell,
     ...rest
   } = props;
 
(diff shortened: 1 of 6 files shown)


packages/itwinui-react/src/core/Table/cells/DefaultCell.tsx
```diff
@@ -68,7 +68,9 @@ export const DefaultCell = <T extends Record<string, unknown>>(
     className,
     style,
     status,
-    clamp = typeof cellProps.value === 'string' && !isCustomCell,
+    clamp = typeof cellProps.value === 'string' &&
+      !!props.cellProps.column.id &&
```
mayank99, line 72: This fix works here, but I'm wondering if some other things are also broken when `id` is not available. I always thought react-table would assign a default `id` to it. Maybe we should try to assign a fallback id using `accessor`?

veekeys: so what would you suggest?

mayank99: Instead of store the `columns` prop into context ([code](https://github.com/iTwin/iTwinUI/blob/1a93a397e7ceab7bce6e8eb6a7158852c8bb063b/packages/itwinui-react/src/core/Table/Table.tsx#L910)), store `instance.columns`?

veekeys: fixed

mayank99 (approved):
LGTM!

I verified locally by changing one of the stories and comparing the outcome of this branch vs `main`. The fix works!

```diff
@@ -63,9 +63,10 @@ export const Basic = () => {
   const columns = React.useMemo(
     () => [
       {
-        id: 'name',
+        // id: 'name',
         Header: 'Name',
         accessor: 'name',
+        Cell: ({ value }) => value,
       },
       {
         id: 'description',

```

mayank99:
🚀 This is released in ⁠[3.14.2](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.14.2), thanks for the PR!

PR #2221, opened 2024-09-03
Remove type hacks after built-in types from newer TS versions

r100-stack:
## Changes

Removed some type hacks that we had done when older TS versions didn't provide up-to-date types:

* `ScrollBehavior`: In TypeScript 5.1, `"instant"` was added to the `ScrollBehavior` type ([comment](https://redirect.github.com/microsoft/TypeScript/issues/47441#issuecomment-1514519335)). Thus, a type assertion in our internal code is no longer necessary.
* `HTMLElementWithPopover`: https://github.com/iTwin/iTwinUI/pull/2221#discussion_r1743752687

## Testing

Confirmed that there are no type errors after the changes from this PR.

## Docs

N/A

### DIFF
--- a/.changeset/spotty-comics-carry.md
+++ b/.changeset/spotty-comics-carry.md
@@ -1,5 +1,0 @@
----
-'@itwin/itwinui-react': patch
----
-
-Fixed the fallback CSS logic in `ThemeProvider` to load the correct stylesheet version.
--- a/apps/css-workshop/src/components/ListItem.astro
+++ b/apps/css-workshop/src/components/ListItem.astro
@@ -1,0 +1,69 @@
+---
+type Props = {
+  isActive?: boolean;
+  size?: string;
+  disabled?: boolean;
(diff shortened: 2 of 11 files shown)


mayank99 (approved):

packages/itwinui-react/src/core/Carousel/CarouselSlider.tsx
mayank99, line 1: This reminded me that we can remove our custom [`HTMLElementWithPopover`](https://github.com/iTwin/iTwinUI/blob/1a93a397e7ceab7bce6e8eb6a7158852c8bb063b/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx#L121-L125) type and instead use `HTMLElement` directly, since latest TS has types for popover.

mayank99 (approved):

smmr-dn (approved):

PR #2224, opened 2024-09-06
Added missing `useInertPolyfill()` calls

r100-stack:
## Changes

In https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1705990042, @mayank99 pointed out a missing `useInertPolyfill()` in `Panels`. That made me realize that it was missing in `CarouselSlide.tsx` too.

To prevent forgetting to call `useInertPolyfill()` in the future, the `useInertPolyfill()` is now moved to the `ThemeProvider` (https://github.com/iTwin/iTwinUI/pull/2224#discussion_r1750919395).

Related: https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1746157525, https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1706876166

## Testing

* Tests still passing.
* Tested in Firefox 110 (where `inert` is [not supported](https://caniuse.com/mdn-html_global_attributes_inert)) that `inert` is still working as expected in `Overlay` and `CarouselSlide`.

## Docs

Didn't add changeset as we felt there's no need to communicate this small change (https://github.com/iTwin/iTwinUI/pull/2224#discussion_r1750940832).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,4 +1,4 @@
-name: Bug Report 🐛 
+name: Bug Report 🐛
 description: Create a bug report to help us improve
 labels: ['bug']
 assignees: []
@@ -14,7 +14,7 @@ body:
       label: Describe the bug (current behavior)
       description: A concise description of what you're experiencing.
     validations:
-      required: false
+      required: true
   - type: textarea
     attributes:
       label: Expected Behavior
@@ -25,6 +25,8 @@ body:
     attributes:
       label: Link to minimal repro
       description: 'Create a minimal reproduction by forking [this sandbox](https://stackblitz.io/github/iTwin/iTwinUI/tree/main/minimal-sandbox?file=/src/App.tsx)'
+    validations:
+      required: true
   - type: textarea
     attributes:
       label: Steps To Reproduce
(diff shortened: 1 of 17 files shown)


packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
```diff
@@ -52,6 +53,8 @@ export const CarouselSlide = React.forwardRef((props, ref) => {
 
   const refs = useMergedRefs(intersectionRef, ref);
 
+  useInertPolyfill();
```
mayank99, line 56: Since we use `inert` in multiple places, I wonder if we should do this in `ThemeProvider` instead?

mayank99 (approved):

smmr-dn (approved):

packages/itwinui-react/src/core/Carousel/CarouselSlide.tsx
mayank99: Yeah it's a tradeoff. I'm ok with doing it in `ThemeProvider` because:
- it's less code for us to maintain
- it's a small download and we're lazy importing it (so it doesn't block the initial render)
- it doesn't affect the majority of users (who are likely using an up-to-date browser)

.changeset/silver-items-refuse.md
mayank99, line 1: no need for changeset, i think it's fine to not inform users of this small change

mayank99 (approved):

smmr-dn (approved):

PR #2225, opened 2024-09-06
Added documentation for `Divider` component

smmr-dn:
## Changes

Added documentation for `Divider` component. 
## Testing

N/A

## Docs

N/A

## After-PR TODO

- [ ] Remove all `propsPath` from all components ([comment](https://github.com/iTwin/iTwinUI/pull/2225#discussion_r1750804352))

### DIFF
--- a/apps/website/src/content/docs/divider.mdx
+++ b/apps/website/src/content/docs/divider.mdx
@@ -1,0 +1,28 @@
+---
+title: Divider
+description: A divider can be used to separate a layout into different sections of content or element.
+propsPath: '@itwin/itwinui-react/esm/core/Divider/Divider.d.ts'
+thumbnail: Divider
+---
+
+<p>{frontmatter.description}</p>
+
+<LiveExample src='Divider.main.jsx' truncate={false}>
+  <AllExamples.DividerMainExample client:load />
+</LiveExample>
+
+## Usage
+
+<p>
+  The divider is set to be displayed both horizontally (by default) or vertically. Additionally, the
+  divider can be customized to suit different design needs, including adjustable styles such width
+  or height.
+</p>
+
+<LiveExample src='Divider.vertical.jsx' truncate={false}>
+  <AllExamples.DividerVerticalExample client:load />
+</LiveExample>
+
+## Props
+
+<PropsTable path='@itwin/itwinui-react/esm/core/Divider/Divider.d.ts' />
--- a/apps/website/src/content/docs/divider.mdx
+++ b/apps/website/src/content/docs/divider.mdx
@@ -2,7 +2,6 @@
 title: Divider
 description: A divider can be used to separate a layout into different sections of content or element.
 propsPath: '@itwin/itwinui-react/esm/core/Divider/Divider.d.ts'
-thumbnail: Divider
 ---
 
 <p>{frontmatter.description}</p>
@@ -25,4 +24,4 @@ thumbnail: Divider
 
 ## Props
 
-<PropsTable path='@itwin/itwinui-react/esm/core/Divider/Divider.d.ts' />
+<PropsTable path={frontmatter.propsPath} />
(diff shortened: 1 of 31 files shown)


examples/Divider.main.jsx
```diff
@@ -0,0 +1,16 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from 'react';
+import { Divider, Text } from '@itwin/itwinui-react';
+
+export default () => {
+  return (
+    <div className='demo-container'>
+      <Text variant='subheading'>Section 1</Text>
+      <Divider style={{ width: '500px' }} orientation='horizontal' />
```
mayank99, line 12: it would be good to not have to pass inline `style`s to `Divider` in these examples. it should automatically fill up the available width of the container (or height in the case of vertical)

smmr-dn: I have removed `width`, `height` and `orientation='horizontal'` since it is set by default already.

apps/website/src/content/docs/divider.mdx
```diff
@@ -0,0 +1,27 @@
...
+  or height.
+</p>
+
+<LiveExample src='Divider.vertical.jsx' truncate={false}>
+  <AllExamples.DividerVerticalExample client:load />
+</LiveExample>
+
+## Props
+
+<PropsTable path={frontmatter.propsPath} />
```
mayank99, line 27: The frontmatter variable is not necessary, we can use the string directly.

```suggestion
<PropsTable path='@itwin/itwinui-react/esm/core/Divider/Divider.d.ts' />
```

We should actually remove these `propsPath` variables from all places (even in other components). Feel free to open a separate PR for that.

apps/website/src/content/docs/divider.mdx
```diff
@@ -0,0 +1,27 @@
...
+  <AllExamples.DividerMainExample client:load />
+</LiveExample>
+
+## Usage
+
+<p>
+  The divider is set to be displayed both horizontally (by default) or vertically. Additionally, the
+  divider can be customized to suit different design needs, including adjustable styles such width
+  or height.
+</p>
```
mayank99, line 19: This Usage section feels superfluous and disconnected from the vertical example. Maybe we can skip it and instead create a section for "Vertical". Also, the `<p>` tags are not necessary because this is markdown.

```suggestion
## Vertical

The `Divider` component is displayed horizontally by default. To display it vertically, use the `orientation` prop.
```

apps/website/src/content/docs/divider.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Divider
+description: A divider can be used to separate a layout into different sections of content or element.
+propsPath: '@itwin/itwinui-react/esm/core/Divider/Divider.d.ts'
+---
```
mayank99, line 5: We should put this component under Utilities in the sidebar. Also we should add a thumbnail in the future. For now lets use a TODO comment.

```suggestion
group: utilities
thumbnail: #TODO
---
```

apps/website/src/content/docs/divider.mdx
smmr-dn: I will add this as an after-PR TODO and will commit the suggested change in this PR.

smmr-dn: I have made the change.

apps/website/src/content/docs/divider.mdx
smmr-dn: I have reworded this section.

apps/website/src/content/docs/divider.mdx
smmr-dn: I have added this component under Utilities. Just a quick question, how are thumbnails generated? I noticed some components do have thumbnails while some remain "#TODO".

mayank99 (approved):

apps/website/src/content/docs/divider.mdx
smmr-dn: I see, thank you.

PR #2231, opened 2024-09-10
Automatically assigned author to PR

smmr-dn:
## Changes

CI workflow has been updated to automatically assign the author who opens the PR to be the assignee. Additionally, I have created a workflow item on the "Design Systems" project board to automatically move newly opened PR to "In Progress/Review" (https://github.com/orgs/iTwin/projects/52/workflows/20699676)

## Testing

Confirmed that assignee is the PR author and PR is moved to "In Progress/Review".

## Docs

N/A


smmr-dn (commented):
I'm not sure if I should separate these into different PRs since they are somewhat similar tasks.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -9,12 +9,27 @@ body:
...
+  - type: dropdown
+    attributes:
+      label: Are you using the latest version of iTwinUI?
+      options:
+        - 'Yes'
+        - 'No'
+    validations:
+      required: true
+    id: version
+  - type: textarea
```
smmr-dn, line 21: Is asking for what version (if not the latest) they are using necessary?

mayank99 (commented):
yeah these should be separate PRs

smmr-dn:
> yeah these should be separate PRs

I have separate these into 2 PRs.

.github/workflows/build.yml
mayank99, line 1: In the PR description you said:

> Additionally, I have created a workflow item on the "Design Systems" project board to automatically move newly opened PR to "In Progress/Review" (https://github.com/orgs/iTwin/projects/52/workflows/20699676)

I'm not sure that this will be enough. Doesn't this replace the existing handling of issues? Iirc, the limitation was that we can't have automation for both issues and PRs using github's UI

smmr-dn: You're right, it does replace the existing handling of issues. I have put back the original workflow for issue and will find a way to automate this through CI.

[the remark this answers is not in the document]
smmr-dn: I just opened a draft one and if the assignee is successfully updated github will notify both on the PR and via email.

smmr-dn (commented):
I'm going to convert this PR back to draft to work on the CI script to move PRs to `In Progress` status.

.github/workflows/build.yml
mayank99: I do wonder if automating it using github's API could be too difficult/tedious. Even if we make it work, we'd have to make the same change in four different repos.

Could we live without automation of issues? i.e. Only use the built-in automation for PRs? We need to manually triage issues anyway, so the Status is not super important there.

mayank99: @r100-stack The "current state" does not use Github's API for setting the status for PRs. It only sets the assignee field to the PR author.

What I'm suggesting is that we move forward with the current state, just like @smmr-dn intended in the beginning of the PR.

I initially said "the limitation was that we can't have automation for both issues and PRs using github's UI" and now I'm saying I'm ok with that limitation, because automation for PR status is much more important than automation for Issue status (which needs manual triage anyway).

smmr-dn: So I think what I understand from Mayank's suggestion is that we can replace the issue automation with PR automation on Github's UI (since we can't have both at the same time) so that we won't have to write a tedious script using Github API just for this automation workflow.

mayank99 (approved):

.github/workflows/build.yml
```diff
@@ -50,6 +50,23 @@ jobs:
       - run: pnpm install
       - run: pnpm audit
 
+  assign-author:
```
mayank99, line 53: It looks like this step is being run whenever we merge a PR to `main`. https://github.com/iTwin/iTwinUI/actions/runs/10910728547/job/30281834580

We can add some `if` check here, or even move this step to a different job or workflow altogether, since it's unrelated to all the other `build` steps.

smmr-dn: I will open another PR addressing this issue.

PR #2232, opened 2024-09-10
Updated Bug Issue template

smmr-dn:
## Changes

Changed **Bug Issue** template to require the link to a minirepo as well as current version information from users.

## Testing

Confirmed that the template turned out correctly.

## Docs

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -39,8 +39,9 @@ body:
   - type: checkboxes
     attributes:
       label: Package Version
+      description: Please confirm if you are using the [latest version](https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions) of iTwinUI.
       options:
-        - label: I'm using the [latest version](https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions).
+        - label: I'm using the latest version.
           required: false
     validations:
       required: false
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -40,7 +40,7 @@ body:
     attributes:
       label: Package Version
       options:
-        - label: I'm using the [latest version](https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions) of iTwinUI.
+        - label: "I'm using the [latest version](https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions) of iTwinUI."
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -9,12 +9,27 @@ body:
...
+        - 'No'
+    validations:
+      required: true
+    id: version
+  - type: textarea
+    attributes:
+      label: Version Details
+      description: If you are not using the latest version, please specify the version you are using.
+    validations:
+      required: false
```
mayank99, line 26: I think we could instead show a checkbox at the end?

[ ] I am using the latest version of iTwinUI.

smmr-dn: Does this change okay? [commit](https://github.com/iTwin/iTwinUI/pull/2232/commits/4b0bc8baf4d81c60312a3079c921cdb0b083c023). Then users can specify their current version, if not the latest, in the last "Anything else?" text box.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -34,6 +36,17 @@ body:
...
+    attributes:
+      label: Package Version
+      description: Confirm if you are using the latest version of iTwinUI.
+      options:
+        - label: I'm using the latest version of iTwinUI.
+          required: false
+        - label: I'm using other version of iTwinUI.
+          required: false
+    validations:
+      required: true
```
mayank99, line 49: I was thinking a single checkbox should be enough. If it is unchecked, it implies they are using a different version. Also we should include a link so that they know what the latest version is.

```suggestion
  - type: checkboxes
    attributes:
      label: Package Version
      options:
        - label: I'm using the [latest version](https://www.npmjs.com/package/@itwin/itwinui-react?activeTab=versions) of iTwinUI.
          required: false
    validations:
      required: false
```

smmr-dn: I'm not sure why the link markdown is not supported in `label` for checkbox `option`. I have looked around, but I haven't found a way to do as suggested. I have added a description to include the link in your suggestion but not sure this is okay.

mayank99: Yeah I guess this is a limitation in Github, nothing we can do about it.

mayank99 (approved):
LGTM!

mayank99 (commented):
Let's merge this, unless we're waiting for something.

(The "audit" failure doesn't block the PR)

PR #2233, opened 2024-09-11
Enable `hide` middleware in `Popover` by default

### LINKED ISSUE #2235: Issue with dropdown menu when the dropdown trigger is scrolled out of the view

With scroll the dropdown menu stays visible beyond the height of the scrollable div, even if the trigger gets out of the view .

image

### Expected Behavior

The dropdown menu should not be visible if the trigger is not visible.

### Link to minimal repro

https://stackblitz.com/edit/github-sogghk?file=src%2FItwinTree.tsx,src%2Fmain.tsx

### Steps To Reproduce

1. Create a Tree with specific height , so that it becomes scrollable
2. Add dropdown menu on each tree nodes 
3. Open any dropdown menu which gets out of the view on scroll 
4. Observe that even if the trigger item is not present on the screen, the dropdown menu stays visible

### Anything else?

_No response_

r100-stack:
## Changes

Fixes #2235.

This PR enables the [`hide` middleware](https://floating-ui.com/docs/hide) in `Popover` and popover-like components (e.g. `Select`, `ComboBox`, `DropdownMenu`).

@mayank99 pointed out some issues that we faced with the `hide` modifier in old versions of itwinui-react that used to use [popper](https://popper.js.org/docs/v2/) instead of [floating-ui](https://floating-ui.com/). I made sure that the previous issues no longer happen after this PR or are no longer related to this PR:
* https://github.com/iTwin/iTwinUI-react/issues/783: I'm not sure what was the issue here. But iiuc, I don't think it relates to this PR since the trigger of a dropdown menu is usually a button that doesn't get `width: 0; height: 0;` or `display: none;`.
* https://github.com/iTwin/iTwinUI-react/issues/947: Confirmed that temporarily enabling the `hide` modifier in `ComboBox`'s `usePopover()` call still does not give the problem demoed in the issue.

Since there is a change that the `hide` middleware could hide the floating content even when it shouldn't, this PR also edits types or adds props to expose a way to disable the `middleware.hide`.

Unrelated fixes:

* Fixed the `hide` middleware in `Popover` to actually apply `visibility: hidden`. Up till now, that middleware used to unintentionally be no-op. ([sandbox](https://stackblitz.com/edit/github-6mg7sd?file=src%2FApp.tsx))
* Fixed merging interactions logic in `Popover`.
  * e.g. suppose we want to enable `foobar` by default. Doing `{foo: true, ...userProps}` actually results in `{foobar: undefined}` if `userProps.foobar = undefined`. The fix I used was `{foobar: userProps.foobar ?? true}`.

## Testing

Added e2e test.

## Docs

Added changesets and docs.

### DIFF
--- a/.changeset/slimy-vans-lick.md
+++ b/.changeset/slimy-vans-lick.md
@@ -3,5 +3,5 @@
 ---
 
 `Popover` now enables the [`hide` middleware](https://floating-ui.com/docs/hide) to hide the floating content when the trigger is hidden.
-* This also affects other popover-like components (e.g. `Select`, `ComboBox`, `DropdownMenu`).
+* This also affects other popover-like components (e.g. `Select`, `ComboBox`, `DropdownMenu`, `SplitButton`).
 * If the floating content gets hidden even when it shouldn't (e.g. due to some custom styles interfering with the trigger's hide detection), consider disabling the `hide` middleware.
--- a/packages/itwinui-react/src/core/Popover/Popover.tsx
+++ b/packages/itwinui-react/src/core/Popover/Popover.tsx
@@ -74,7 +74,10 @@ type PopoverOptions = {
   /**
    * Middleware options.
    *
-   * By default, `flip`, `shift` and `size` are enabled.
+   * By default, `flip`, `shift`, `size`, and `hide` are enabled.
+   *
+   * If the floating content gets hidden even when it shouldn't (e.g. some custom styles interfering with the trigger's
+   * hide detection) consider disabling the `hide` middleware.
    *
    * @see https://floating-ui.com/docs/middleware
    */


packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
```diff
@@ -87,6 +98,14 @@ const DropdownMenuContent = React.forwardRef((props, forwardedRef) => {
     onVisibleChange,
   );
 
+  const middleware = React.useMemo(
+    () => ({
+      hide: false,
```
mayank99, line 103: I was actually hoping that we would enable `hide` by default. #2235 is clearly a bug, and we shouldn't require users to opt-in; it should just do the right thing by default. And since `hide` has been problematic in the past, we can still allow users to opt-out when it actually causes issues.

packages/itwinui-react/src/core/Popover/Popover.tsx
```diff
@@ -287,17 +287,24 @@ export const usePopover = (options: PopoverOptions & PopoverInternalProps) => {
                 maxInlineSize: `min(${referenceWidth * 2}px, 90vw)`,
               }
             : {}),
+          ...(middleware.hide && {
+            visibility: floating.middlewareData.hide?.referenceHidden
+              ? 'hidden'
+              : 'visible',
+          }),
```
mayank99, line 294: We shouldn't set `visibility: visible`; we should just *do nothing* when `referenceHidden` doesn't match.

e.g. something like:

```js
...(middleware.hide
    && floating.middlewareData.hide?.referenceHidden && {
      visibility: 'hidden'
   })
```

apps/react-workshop/src/DropdownMenu.stories.tsx
```diff
@@ -336,3 +339,60 @@ WithContent.decorators = [
     </div>
   ),
 ] satisfies StoryDecorator[];
+
+export const HideMenuWhenTriggerHidden = () => {
```
mayank99, line 343: I feel like this story is unnecessary when we already have docs + e2e test. Why do we need all three?

apps/website/src/content/docs/dropdownmenu.mdx
mayank99, line 1: `popover.mdx` also needs to be updated to mention that `hide` is enabled by default

.changeset/slimy-vans-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`DropdownMenu` exposes `middleware` prop to customize `hide`, similar to `Popover`.
```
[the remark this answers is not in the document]
mayank99: Actually this affects more than just `DropdownMenu` and `Popover`. We should give it the same treatment that we did to `size` middleware in [3.14.0](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.14.0)

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
```diff
@@ -30,6 +30,17 @@ export type DropdownMenuProps = {
...
+  /**
+   * Middleware options.
+   *
+   * By default, `hide` is enabled. If the menu gets hidden even when it shouldn't (e.g. some custom styles interfering
+   * with the trigger's hide detection) consider disabling the `hide` middleware.
+   *
+   * @see https://floating-ui.com/docs/middleware
+   */
+  middleware?: {
+    hide?: boolean;
```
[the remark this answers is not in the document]
mayank99: It should be under `dropdownMenuProps` in `ComboBox` and under `popoverProps` in `Select`.

testing/e2e/app/routes/DropdownMenu/route.tsx
```diff
@@ -2,23 +2,33 @@ import {
...
+  IconButton,
+  List,
+  ListItem,
   MenuDivider,
   MenuExtraContent,
   MenuItem,
+  Surface,
 } from '@itwin/itwinui-react';
 import { useSearchParams } from '@remix-run/react';
+import { SvgMore } from 'node_modules/@itwin/itwinui-react/esm/utils/index.js';
```
mayank99, line 14: Do not rely on these internal paths; instead import from `@itwin/itwinui-icons-react` if you need this icon.

mayank99 (approved):

packages/itwinui-react/src/core/Buttons/DropdownButton.test.tsx
```diff
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { render } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
```
smmr-dn, line 6: Is `waitFor` being used in this unit test?

mayank99 (commented):
To help with failing unit tests, you might consider guarding the default value of `hide` with `isUnitTest`. This middleware does not make any sense in a fake DOM environment so it's ok to ignore it.

packages/itwinui-react/src/core/DropdownMenu/DropdownMenu.tsx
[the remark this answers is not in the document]
mayank99: > Changed ([92a8517](https://github.com/iTwin/iTwinUI/commit/92a851724ec51843141f0ac73381bde1ce3be122)).

Just a nit: we shouldn't rely on `DropdownMenuProps` in `ComboBox`; we should use `usePopover` props instead. The `ComboBox` has nothing to do with the `DropdownMenu` (yes, the naming of `dropdownMenuProps` is unfortunate, but we cannot change that now).

> Do we want to expose props to pass down to `Menu`, `DropdownMenu`, etc.?

Yeah let's do that at least for `SplitButton`. For `HeaderButton` we can leave it out, since it's always in one spot (and the component is built in a questionable way and should be replaced in the future).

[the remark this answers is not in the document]
mayank99: Maybe `dropdownMenuProps` instead of `menuProps`?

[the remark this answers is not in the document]
mayank99: Another option is `popoverProps` which I guess would be more appropriate. I don't feel strongly either way, so I'd say use your best judgement.

mayank99 (approved):
LGTM (regardless of https://github.com/iTwin/iTwinUI/pull/2233#discussion_r1768892606)

smmr-dn (approved):

PR #2217, opened 2024-08-30
Mark all `Footer` translations as optional

r100-stack:
## Changes

`Footer`'s `TitleTranslations` type used to have all title translations as required:

https://github.com/iTwin/iTwinUI/blob/608887d59825abceb30e15902a40e1f4d002293c/packages/itwinui-react/src/core/Footer/Footer.tsx#L14-L20

Thus, it used to give a type error when passing just one of the translations. This PR makes all those title translations as optional.

## Testing

The `Basic` story passes only one translation. Thus, added `Footer.stories.tsx` in `tsconfig.build.json` so that the file will be checked for any type errors, including any type error for not passing all the translations.

## Docs

Added changesets

## After PR TODO:

* [ ] Type check all stories (https://github.com/iTwin/iTwinUI/pull/2217#discussion_r1742020099)
  * PR: https://github.com/iTwin/iTwinUI/pull/2234

PR #2234, opened 2024-09-11
Enable type checks for all stories

r100-stack:
## Changes

After PR TODO from #2217 (https://github.com/iTwin/iTwinUI/pull/2217#discussion_r1742609279).

Adds type checking to all stories in `react-workshop`.

## Testing

All tests continue to pass.

## Docs

No changeset needed as this is an internal change.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,4 +1,4 @@
-name: Bug Report 🐛 
+name: Bug Report 🐛
 description: Create a bug report to help us improve
 labels: ['bug']
 assignees: []
@@ -14,7 +14,7 @@ body:
       label: Describe the bug (current behavior)
       description: A concise description of what you're experiencing.
     validations:
-      required: false
+      required: true
   - type: textarea
     attributes:
       label: Expected Behavior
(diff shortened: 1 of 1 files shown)


apps/react-workshop/src/InformationPanel.stories.tsx
```diff
@@ -39,7 +39,7 @@ export const Basic = () => {
       },
     ],
     [],
-  );
+  ) satisfies Column<{ name: string }>[];
```
mayank99, line 42: Does `satisfies Column[]` not work? because this somewhat defeats the purpose of using `satisfies`

apps/react-workshop/src/Select.stories.tsx
```diff
@@ -157,7 +157,10 @@ export const Custom = () => {
     <Select
       options={options}
       value={selectedValue}
-      onChange={setSelectedValue}
+      onChange={React.useCallback(
+        (value: string) => setSelectedValue(value),
```
mayank99, line 161: The explicit type annotation is usually unnecessary for params. Does it not get automatically inferred?

```suggestion
        (value) => setSelectedValue(value),
```

mayank99 (approved):

smmr-dn (approved):

PR #2239, opened 2024-09-12
Fix `scrollToRow` for non-virtualized `Table`s

r100-stack:
## Changes

A user reported issue was that the `scrollToRow` always scrolled a bit further than the intended row and thus the intended row would be overlapped by the sticky header.

I added [`scroll-margin-block-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-block-start) to `iui-table-row` with a value of approximately the maximum possible height of the header (min height of the table header (`--_iui-table-header-size`) + some extra height) ([suggestion](https://github.com/iTwin/iTwinUI/pull/2239#discussion_r1757303996) by @mayank99).

I also considered a JS solution like `scrollIntoView()` (fallback) + `scrollTo(row.offsetTop)` (when the table's root element is scrollable). But due to potential performance issues, unnecessary complexity, and [other reasons](https://github.com/iTwin/iTwinUI/pull/2239#discussion_r1757470898), I moved away from this.

To keep the scroll behavior consistent, the scroll align for the virtualized table is now also set to top:
|Virtualization|scroll align (before)|scroll align (after)|
|-|-|-|
|enabled|center|**top (changed)**|
|disabled|top|top|

## Testing

Removed the old unit test in favor of a react-workshop image test and e2e test.

Tested [user sandbox](https://stackblitz.com/edit/github-z6nfry?file=src%2FApp.tsx,src%2FSampleTable.tsx,src%2Fmain.tsx) without virtualization. Notice the new extra space at the top of the scrolled row. This would likely look better when adding the "highlight the scrolled row" feature.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/6804af04-faea-4d6a-ac8c-1c6cf8b34881"/>|<video src="https://github.com/user-attachments/assets/8ab94a82-3a87-4dbd-bfbf-a84dab4f76e1" />|

Tested with virtualization enabled. Here, that extra space is not added regardless of the sticky header height.

|Before|After|
|-|-|
|image|image|

## Docs

Added patch changeset.

## After PR TODOs:

* [ ] Add highlight when scrolled to a row

### DIFF
--- a/packages/itwinui-react/src/core/Overlay/Overlay.tsx
+++ b/packages/itwinui-react/src/core/Overlay/Overlay.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import * as React from 'react';
-import { Box, isUnitTest, polymorphic } from '../../utils/index.js';
+import { Box, polymorphic } from '../../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 
 type OverlayComponentProps = {
@@ -39,7 +39,6 @@ if (process.env.NODE_ENV === 'development') {
 
 const OverlayHiddenContent = React.forwardRef((props, ref) => {
   const { children, ...rest } = props;
-  useInertPolyfill();
   return (
     <Box {...{ inert: '' }} ref={ref} {...rest}>
       {children}
@@ -104,22 +103,3 @@ export const Overlay = Object.assign(OverlayComponent, {
    */
   Overlay: OverlayOverlay,
 });
-
-const useInertPolyfill = () => {
-  const loaded = React.useRef(false);
-  const modulePath =
-    'https://cdn.jsdelivr.net/npm/wicg-inert@3.1.2/dist/inert.min.js';
-
-  React.useEffect(() => {
-    (async () => {
-      if (
-        !HTMLElement.prototype.hasOwnProperty('inert') &&
-        !loaded.current &&
-        !isUnitTest
-      ) {
-        await new Function('url', 'return import(url)')(modulePath);
-        loaded.current = true;
-      }
-    })();
-  }, []);
-};
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -149,6 +149,8 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
     ...rest
   } = props;
 
+  useInertPolyfill();
+
   const [rootElement, setRootElement] = React.useState<HTMLElement | null>(
     null,
   );
@@ -485,4 +487,25 @@ const useIuiDebugRef = () => {
   }
 
   _globalThis.__iui.versions.add(JSON.stringify(meta));
+};
+
+// ----------------------------------------------------------------------------
+
+const useInertPolyfill = () => {
+  const loaded = React.useRef(false);
+  const modulePath =
+    'https://cdn.jsdelivr.net/npm/wicg-inert@3.1.2/dist/inert.min.js';
+
+  React.useEffect(() => {
+    (async () => {
+      if (
+        !HTMLElement.prototype.hasOwnProperty('inert') &&
+        !loaded.current &&
+        !isUnitTest
+      ) {
+        await new Function('url', 'return import(url)')(modulePath);
+        loaded.current = true;
+      }
+    })();
+  }, []);
 };


packages/itwinui-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -54,8 +56,12 @@ export function useScrollToRow<T extends Record<string, unknown>>({
       return;
     }
 
-    rowRefs.current[pageRef.current[scrollToIndex]?.id]?.scrollIntoView();
-  }, [enableVirtualization, scrollToIndex]);
+    setTimeout(() => {
+      tableRef.current?.scrollTo({
```
mayank99, line 60: We cannot rely on `scrollTo`. It requires an explicit value, which is not always feasible to calculate. `offsetTop` might work for simple cases but will fail when the Table itself is not scrollable (e.g. when the page is scrollable and Table has `overflow: visible`).

Like I suggested earlier, let's use `scrollIntoView` and pick an approximate `scroll-margin` value that's bigger than the Table header (in its tallest wrapped view).

mayank99: It doesn't make sense to have two different implementations (three if you count virtualization), but I won't block the PR if you strongly before this. I just hope you understand that there is an implementation complexity to this, without any real UX benefit.

- Previously it was deemed acceptable that the row would be somewhere within the view, not necessarily at the top. So picking an approximate `scroll-margin` is not worse than before. It fixes the bug.
- Scrolling to the top might make sense to a developer, but the user might still be unaware of which row to look at (there's more than one row on the page).
- The highlight effect is a must for usable UX, and we should also recommend developers to move the focus (for keyboard users). Any solution we pick today must be in service of this end goal.

Other points to be aware of:
- Reading `offsetTop` forces [layout](https://gist.github.com/paulirish/5d52fb081b3570c81e3a), which could lead to a performance issue.
- Not having a buffer/padding around the scrolled row might cause an issue if we go with a highlight effect that extends outwards (e.g. if we use a box-shadow, it might get clipped by the Table header).

testing/e2e/app/routes/Table/spec.ts
```diff
@@ -485,10 +485,10 @@ test.describe('Virtual Scroll Tests', () => {
...
-    expect((await rows.all()).length).toBe(14);
-    await expect(rows.nth(0)).toContainText('Name43');
-    await expect(rows.nth(7)).toContainText('Name50');
-    await expect(rows.nth(13)).toContainText('Name56');
+    expect((await rows.all()).length).toBe(13);
+    await expect(rows.nth(0)).toContainText('Name49');
+    await expect(rows.nth(1)).toContainText('Name50');
+    await expect(rows.nth(12)).toContainText('Name61');
 
     await expect(row50NameCell).toBeInViewport();
```
mayank99, line 493: Should we have an e2e test for the non-virtualized case?

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -822,7 +826,7 @@ export const Table = <
 
   useLayoutEffect(() => {
     if (scrollToIndex) {
-      virtualizer.scrollToIndex(scrollToIndex, { align: 'center' });
+      virtualizer.scrollToIndex(scrollToIndex, { align: 'start' });
```
mayank99, line 829: Correct me if I'm wrong, but this is what happens when using virutalization:
1. `scrollIntoView` (from `useScrollToRow`) executes first, but it fails because the row does not exist in the DOM.
2. `virtualizer.scrollToIndex` is what actually scrolls the row to the top, but it still stays hidden underneath the Table header.
3. `scrollTo` (within `setTimeout` in `useScrollToRow`) executes and finally shows the Table.

Feels a bit indirect, no? See tanstack-virtual's [`scrollMargin`](https://tanstack.com/virtual/latest/docs/api/virtualizer#scrollmargin)  option for a more direct solution.

packages/itwinui-react/src/core/Table/hooks/useScrollToRow.ts
mayank99: Make sure the PR description is up-to-date.

testing/e2e/app/routes/Table/route.tsx
```diff
@@ -136,11 +145,17 @@ function EverythingElse() {
...
+        style={{ maxHeight: '90vh' }}
         scrollToRow={
           scroll
-            ? (rows, data) =>
-                rows.findIndex((row) => row.original === data[scrollRow])
+            ? (rows, data) => {
+                const q = rows.findIndex(
+                  (row) => row.original === data[scrollRow],
+                );
+                console.log(q, scrollRow);
```
mayank99, line 155: stray console.log

mayank99 (approved):

smmr-dn (approved):

PR #2240, opened 2024-09-12
Made expandable content a separate component

smmr-dn:
## Changes

Separate expandable content in `Table` to another component, independent from the original `TableRowMemoized`. This PR prepares for the fixing of the bug that is caused by sub-component not being recognized by the virtualizer when being rendered within the main row.

## Testing

Confirmed the expandable contents visually turned out correctly.

## Docs

N/A

## After-PR TODO:
- [ ] Change AIRA attributes into data attributes ([comment](https://github.com/iTwin/iTwinUI/pull/2240#discussion_r1758912905))

### DIFF
--- a/packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
+++ b/packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
@@ -8,13 +8,11 @@ import { Box } from '../../utils/index.js';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 
 type TableExpandableContentProps = {
-  isDisabled: boolean;
   children: React.ReactNode;
-  isSelected?: boolean;
 };
 
 const TableExpandableContent = React.forwardRef((props, ref) => {
-  const { isDisabled, children, isSelected, className, style, ...rest } = props;
+  const { children, className, style, ...rest } = props;
   return (
     <Box
       className={cx('iui-table-row', 'iui-table-expanded-content', className)}
@@ -23,8 +21,6 @@ const TableExpandableContent = React.forwardRef((props, ref) => {
         minWidth: '100%',
         ...style,
       }}
-      aria-disabled={isDisabled}
-      aria-selected={isSelected}
       ref={ref}
       {...rest}
     >
--- a/packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
+++ b/packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
@@ -5,43 +5,33 @@
 import cx from 'classnames';
 import * as React from 'react';
 import { Box } from '../../utils/index.js';
-import type { VirtualItem } from '@tanstack/react-virtual';
 import type { PolymorphicForwardRefComponent } from '../../utils/index.js';
 
 type TableExpandableContentProps = {
-  virtualItem?: VirtualItem<Element>;
   isDisabled: boolean;
   children: React.ReactNode;
   isSelected?: boolean;
 };
 
-const TableExpandableContent = React.forwardRef((props, ref) => {
-  return (
-    <Box
-      className={cx('iui-table-row', 'iui-table-expanded-content')}
-      style={{
-        flex: '0 0 auto',
-        minWidth: '100%',
-        ...(props.virtualItem != null
-          ? { transform: `translateY(${props.virtualItem.start}px)` }
-          : {}),
-      }}
-      aria-disabled={props.isDisabled || undefined}
-      data-iui-index={props.virtualItem?.index}
-      aria-selected={props.isSelected}
-      {...(props.virtualItem != null && { 'data-iui-virtualizer': 'item' })}
-      ref={ref}
-    >
-      {props.children}
-    </Box>
-  );
-}) as PolymorphicForwardRefComponent<'div', TableExpandableContentProps>;
-
 export const TableExpandableContentMemoized = React.memo(
-  TableExpandableContent,
-  (prevProp, nextProp) =>
-    prevProp.children === nextProp.children &&
-    prevProp.virtualItem === nextProp.virtualItem &&
-    prevProp.isDisabled === nextProp.isDisabled &&
-    prevProp.isSelected === nextProp.isSelected,
+  React.forwardRef((props, ref) => {
+    const { isDisabled, children, isSelected, className, style, ...rest } =
+      props;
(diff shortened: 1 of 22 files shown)


packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const TableExpandableContent = React.forwardRef((props, ref) => {
+  return (
+    <Box
+      className={cx('iui-table-row', 'iui-table-expanded-content')}
+      style={{
+        flex: '0 0 auto',
+        minWidth: '100%',
+        ...(props.virtualItem != null
+          ? { transform: `translateY(${props.virtualItem.start}px)` }
```
mayank99, line 26: Virtualization doesn't currently work with expanded content, right? So let's remove all virtualization-specific code from this component.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
```diff
@@ -0,0 +1,47 @@
...
+  virtualItem?: VirtualItem<Element>;
+  isDisabled: boolean;
+  children: React.ReactNode;
+  isSelected?: boolean;
+};
+
+const TableExpandableContent = React.forwardRef((props, ref) => {
+  return (
+    <Box
+      className={cx('iui-table-row', 'iui-table-expanded-content')}
```
mayank99, line 21: Even though we're not exposing this component publicly, I think `props` should be properly spread and merged where needed.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
```diff
@@ -0,0 +1,47 @@
...
+}) as PolymorphicForwardRefComponent<'div', TableExpandableContentProps>;
+
+export const TableExpandableContentMemoized = React.memo(
+  TableExpandableContent,
+  (prevProp, nextProp) =>
+    prevProp.children === nextProp.children &&
+    prevProp.virtualItem === nextProp.virtualItem &&
+    prevProp.isDisabled === nextProp.isDisabled &&
+    prevProp.isSelected === nextProp.isSelected,
+);
```
mayank99, line 47: Let's not have a second parameter, it's usually needed only for advanced cases. Instead, we can just wrap `TableExpandableContent` in a simple `React.memo()` call.

See docs: https://react.dev/reference/react/memo#parameters

smmr-dn: I have wrapped `React.memo` around the component.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
smmr-dn: I have made the change.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
smmr-dn: I have removed virtualization-relating code.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
mayank99: I wasn't asking to remove the `TableExpandableContent` component 😅

```jsx
export const TableExpandableContentMemoized = React.memo(TableExpandableContent);
```

smmr-dn: Ah I thought I could have just wrapped the whole component inside the `memo` call since we didn't need the logic of the second parameter anymore. I have made the change.

mayank99 (approved):
LGTM!

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <Box
+      className={cx('iui-table-row', 'iui-table-expanded-content', className)}
+      style={{
+        flex: '0 0 auto',
+        minWidth: '100%',
+        ...style,
+      }}
+      aria-disabled={isDisabled}
+      aria-selected={isSelected}
```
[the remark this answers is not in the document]
mayank99: Good catch. Neither of these ARIA attributes are being used correctly, because the expandable-content is a generic `<div>`. I think we only use `aria-disabled` in our CSS (which we should probably change to a data attribute). The selection state is handled using the `+` selector.

https://github.com/iTwin/iTwinUI/blob/55377685ee5f2f361ae762a840e7e72c321d5026/packages/itwinui-css/src/table/base.scss#L241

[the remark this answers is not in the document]
smmr-dn: Should I remove this aria attribute in `TableRowMemoized` as well since this is also a generic `<div>`?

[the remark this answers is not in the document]
mayank99: No, we rely on `aria-selected` to style `TableRow`. Let's make a note to look into changing these to data attributes in the future.

https://github.com/iTwin/iTwinUI/blob/55377685ee5f2f361ae762a840e7e72c321d5026/packages/itwinui-css/src/table/base.scss#L220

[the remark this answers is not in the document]
smmr-dn: I have this as an after-PR todo and removed the aria attributes in the expandable content.

mayank99:
I'm good with merging it tbh. No reason to leave it hanging. It's a small, standalone PR that does exactly what it says.

packages/itwinui-react/src/core/Table/TableExpandableContentMemoized.tsx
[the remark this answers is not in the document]
mayank99: Yes, we never suggested removing `aria-disabled`, only `aria-selected` (because it wasn't present before this PR). Please re-read the thread and feel free to ask any questions. The changes in your latest commit look fine, but just make sure to revert the playground.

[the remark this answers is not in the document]
smmr-dn: I have re-read and everything is clear to me. Merged the PR.

PR #2244, opened 2024-09-16
Remove outdated `STYLEGUIDE` point

r100-stack:
## Changes

Removed outdated point ([#2244 (comment)](https://github.com/iTwin/iTwinUI/pull/2244#discussion_r1771365755)) from the `STYLEGUIDE`:

https://github.com/iTwin/iTwinUI/blob/cbd8aa1b1f21b390f395aff7288ce585a9f08b7a/STYLEGUIDE.md?plain=1#L83-L114

Unrelated: Reused some prop types in `SearchBox.tsx` to prevent duplication.

## Testing

Tests still passing.

## Docs

N/A

### DIFF
--- a/packages/itwinui-react/src/utils/providers/ScopeProvider.tsx
+++ b/packages/itwinui-react/src/utils/providers/ScopeProvider.tsx
@@ -11,8 +11,6 @@ const ScopeContext = React.createContext({
   parentStore: null as null | ReturnType<typeof createStore>,
 });
 
-type ScopeProviderProps = { children: React.ReactNode };
-
 /**
  * Provider that creates a fresh, isolated jotai store for its children.
  *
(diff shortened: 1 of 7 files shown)


packages/itwinui-react/src/utils/providers/ScopeProvider.tsx
```diff
@@ -11,14 +11,18 @@ const ScopeContext = React.createContext({
   parentStore: null as null | ReturnType<typeof createStore>,
 });
 
+type ScopeProviderProps = { children: React.ReactNode };
```
mayank99, line 14: I'm not sure how helpful it is to have these `type`s defined separately. It adds a lot of unnecessary top-level noise to the file.

In private components, we often just have one or two props. And in public components, we usually need `PolymorphicForwardRefComponent`. So I'm thinking that we have outgrown the convention defined in `STYLEGUIDE.md`.

mayank99: The `STYLEGUIDE` is meant to be an evolving document based on the team's (current) agreed-upon preferences. Many points in it were written quite some time ago, so when in doubt, you can always ask internally before changing code.

mayank99 (approved):

smmr-dn (approved):

PR #2246, opened 2024-09-17
`dev` script properly builds all outputs

r100-stack:
## Changes

Fixes Next playground giving a "Failed to compile" error:

image

The error was because [#2246 (comment)](https://github.com/iTwin/iTwinUI/pull/2246#discussion_r1763732973):

> Next.js (Webpack) is trying to resolve the [`"development"` entrypoint](https://github.com/iTwin/iTwinUI/blob/b243bd22fd70282d5e59f8bf07cb2c9fc3694887/packages/itwinui-react/package.json#L14), which `swc` does not currently produce.

The fix involves properly generating the package code for the `DEV-cjs` and `DEV-esm` folders. To do this, I added two sub-scripts that use swc to build package code for the two `DEV-*` folders.

|`dev` before|`dev` after|
|-|-|
|image|image|

Types are only generated for the `esm` and `cjs` folders ([#2246 (comment)](https://github.com/iTwin/iTwinUI/pull/2246#discussion_r1763970385)).

I had also thought about just building the `DEV-esm` and DEV-cjs` folders when running the `dev` script. But looks like we need to always build all four folders for the reasons mentioned in [#2246 (comment)](https://github.com/iTwin/iTwinUI/pull/2246#discussion_r1763926380).

Unrelated: I noticed that `styles.css` was a build output but was not added to the `build:clean`. So, I added it.

## Testing

All playgrounds except the next playground used to work. Confirmed that now all playgrounds work. 

## Docs

N/A

## After PR TODOs:

- [ ] Consider not including the next playground in the top-level dev script ([#2246 (comment)](https://github.com/iTwin/iTwinUI/pull/2246#discussion_r1763926380)). This makes things easiers and reduces resource consumption when using the top level `dev` script. Not having the `next` playground in the top level `dev` script is alright as we usually use the vite playground. Even if we want to use the next playground, we can do so manually using own `dev` script.
- [ ] See if the next playground can directly use tsx files ([#2246 (comment)](https://github.com/iTwin/iTwinUI/pull/2246#discussion_r1763926380)).

### DIFF
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -99,7 +99,7 @@
     "dev:dev-esm": "swc src -d DEV-esm --env-name='development' --watch --strip-leading-paths",
     "dev:cjs": "swc src -d cjs --env-name='production' --watch --strip-leading-paths -C module.type=commonjs",
     "dev:dev-cjs": "swc src -d DEV-cjs --env-name='development' --watch --strip-leading-paths -C module.type=commonjs",
-    "dev:types": "concurrently \"tsc -p tsconfig.build.json --outDir DEV-esm --watch --preserveWatchOutput\" \"tsc -p tsconfig.build.json --outDir DEV-cjs --watch --preserveWatchOutput\"",
+    "dev:types": "concurrently \"tsc -p tsconfig.build.json --outDir esm --watch --preserveWatchOutput\" \"tsc -p tsconfig.build.json --outDir cjs --watch --preserveWatchOutput\"",
     "dev:styles": "pnpm build:styles --watch"
   },
   "dependencies": {


packages/itwinui-react/src/styles.js/vite.config.mjs
mayank99, line 1: I'm not sure that I understand the changes in this file. Can you elaborate on your reasoning?

The `DEV-esm`/`DEV-cjs` folders are for our users, and not necessarily related to our local dev servers. The problem that we're running into is that Next.js (Webpack) is trying to resolve the `"development"` entrypoint, which `swc` does not currently produce.

So it should be enough to only fix the `swc` script in `package.json` to produce all four folders (`DEV-esm`/`DEV-cjs`/`esm`/`cjs`). We should never *not* produce the `esm`/`cjs` directories.

mayank99: > But just curious, why would we need the production `esm` and `cjs` folders in the development environment? Is it in case we ever need to check and debug the build output with hot/fast reload enabled?

The main reason is that it's an invalid configuration to only have `DEV-cjs` and `DEV-esm` folders. There is no guarantee that all tools will use the `"development"` entrypoint (it's a convention, not a standard).

Also, `esm` and `cjs` folders are where the types are located, so we need it anyway (for things like proper IDE support). I don't think the `dev:types` script should be changed.

---

It's also worth mentioning that ideally we could make the `next` playground use our TypeScript source directly (similar to #2024), but I haven't yet found a way to do that. These things can be time-consuming to do properly and it's one of the reasons I was suggesting excluding `next` playground from our top-level `dev` script.

mayank99 (approved):
LGTM! next-playground working fine on my machine

smmr-dn (approved):

PR #2247, opened 2024-09-17
Removed Menu mixins

smmr-dn:
## Changes

Mixins for `Menu` have been removed. Since the styles are not used anywhere else, this change does not affect other components' styling.

## Testing

N/A

## Docs

N/A


mayank99 (approved):
LGTM

PR #2248, opened 2024-09-17
Allowed assigning PR author in PRs only

smmr-dn:
## Changes

Only allowed assigning author in PR and disabled this CI item from running when the PR is merged (mentioned in this comment https://github.com/iTwin/iTwinUI/pull/2231#discussion_r1764014043)

## Testing

Confirmed that PR author is still assigned in PRs.

## Docs

N/A


.github/workflows/assign-PR-author.yml
```diff
@@ -0,0 +1,22 @@
+name: Assign PR author
```
mayank99, line 1: I would recommend renaming this file to `PR.yml` and naming the workflow "PR" as well. In the future, we can add any other PR-related automation in this file too.

.github/workflows/assign-PR-author.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  pull_request:
+    types: [opened, reopened, ready_for_review]
```
mayank99, line 4: I think the `types` is not necessary?

.github/workflows/assign-PR-author.yml
smmr-dn: I have renamed both.

.github/workflows/assign-PR-author.yml
smmr-dn: I have removed `types`.

mayank99 (approved):
LGTM

PR #2250, opened 2024-09-18
Add veracode scan

ben-polinsky:
## Changes

Add Veracode static code analysis. Gather, zip, and upload files to Veracode for analysis.

## Testing

Tested zipping locally. Tested zip and upload in CI and validated in Veracode dashboard.


scripts/zip-for-veracode.mjs
mayank99, line 1: It would be nice to not need this manual script with multiple new dependencies. It looks like all it does is zip up the source code; can't we just do that from the command line? 

We could even get rid of all the excluded paths, by only zipping up the `packages` and `apps` directories.

mayank99 (commented):
Hey, thanks for creating this PR. It generally looks ok but I'm wondering if we can reduce maintenance burden.

scripts/zip-for-veracode.mjs
ben-polinsky: Yeah that makes sense. We may need to ignore some paths, but if we don't install deps then no node_modules to exclude. I'll circle back today

.github/workflows/build.yml
```diff
@@ -367,3 +367,64 @@ jobs:
...
+          excludedPaths=(
+            "**/node_modules/*"
+            ".git/*"
+            "**/backstop/*"
+            '**/dist/*'
+            '**/build/*'
+            "**/cypress*/*"
+            "**/.turbo/*"
+            "**/.astro/*"
+            "**/.next/*"
```
mayank99, line 396: We can reduce the number of excluded paths, since we don't run `install` or `build`.

```suggestion
            ".git/*"
            "**/backstop/*"
            "**/cypress*/*"
```

package.json
mayank99, line 1: The `package.json` and `pnpm-lock.yaml` changes can be reverted entirely.

.github/workflows/build.yml
```diff
@@ -367,3 +367,64 @@ jobs:
           name: e2e-report
           path: testing/e2e/playwright-report/
           retention-days: 30
+
+  veracode:
+    name: Zip and Upload files to Veracode
+    runs-on: ubuntu-latest
+    # if: github.ref == 'refs/heads/main'
```
mayank99, line 374: so we don't forget

```suggestion
    if: github.ref == 'refs/heads/main'
```

mayank99 (commented):
Some small suggestions

.github/workflows/build.yml
ben-polinsky: Did this while you were suggesting.

package.json
ben-polinsky: yep did this

mayank99: There's some whitespace changes in `pnpm-lock.yaml`. Can you try reverting it entirely?

```
git checkout origin/main pnpm-lock.yaml
```

ben-polinsky:
I see your audit has been failing for some time, do you want me to fix these?

mayank99 (approved):
LGTM! Thanks again for creating this PR

We are aware of the `audit` issue, don't worry about that. I believe we were waiting for a fix from one of our dependencies; it might be a good time to check again. /cc @r100-stack

PR #2251, opened 2024-09-18
`useWarningLogger` only runs on the client

r100-stack:
## Changes

An example from `examples` was failing on the Astro docs site:

image

This is because `useWarningLogger` uses `window` (not available on the server), but in Astro, that component was running on the server.

https://github.com/iTwin/iTwinUI/blob/e4c16731c0dc3ee2e85c8956ea9a61720650f84d/packages/itwinui-react/src/utils/hooks/useWarningLogger.ts#L47

So, this PR uses `getWindow()` instead of `window`. `getWindow()` returns `undefined` when `window` does not exist instead of giving an error.

https://github.com/iTwin/iTwinUI/blob/e4c16731c0dc3ee2e85c8956ea9a61720650f84d/packages/itwinui-react/src/utils/functions/dom.ts#L14-L20

## Testing

* Confirmed that the warnings (e.g. `IconButton is missing the ``label`` prop.` warning) are still working.
* Confirmed that the new example I was adding that was giving the error on the docs site is now working.

## Docs

No changeset needed imo as not a significant change for the user.


packages/itwinui-react/src/utils/hooks/useWarningLogger.ts
```diff
@@ -43,6 +43,11 @@ export const useWarningLogger =
 
         const logWarning = React.useCallback(
           (message: string) => {
+            // Bail if window not available (e.g. on server)
+            if (typeof window === 'undefined') {
+              return;
```
mayank99, line 48: we do have a `getWindow` helper if you want to use that instead

mayank99 (approved):

smmr-dn (approved):

PR #2252, opened 2024-09-18
re-add shadow DOM styles when reparented

### LINKED ISSUE #2230: Preview popout reparent: iTwinUI ProgressRadial shows "Loading" text when popped out
### Describe the bug

When `ProgressRadial` is popped out, it's shadow dom loses it's styling and "Loading" text is being displayed even though it was supposed to be hidden:
image

This doesn't happen when component is rendered initially in new window, only when transition from main to new window is happening and widget dom is being moved.

I reproduced everything on appui storybook (on release/4.15.0 tag) by changing widget content to `ProgressRadial` in `ReparentPopoutWidgets.tsx` file.

### To Reproduce

1. Have a widget with `ProgressRadial` spinner.
2. Pop out widget.
3. Notice "Loading" text being shown inside spinner. 

### Expected Behavior

Expected for loading text to remain hidden when transitioning.

### Screenshots

image

### Desktop (please complete the applicable information)

**OS:** Windows 11
**Browser:** Microsoft Edge
**Browser version:** Version 126.0.2592.68 (Official build) (64-bit)
**Appui version:** 4.15.0 (reproduced on source)

### Additional context

Would need this fixed before our Modeler's release, our feature freeze is at beggining of next month. At worst we would need a workaround or to opt out of iTwinUI's accessibility shadow dom usage

mayank99:
## Changes

This PR fixes #2230 by re-creating the `CSSStyleSheet` inside `ShadowRoot` whenever the `appui:reparent` event is emitted. This event was added in https://github.com/iTwin/appui/pull/946 specifically to be consumed by our `ShadowRoot` component, in order to support the "reparent widget" feature (there is more info in https://github.com/iTwin/appui/issues/893 if you're curious).

It's not perfect and there is room for improvement, but it fixes the bug. The main issue right now is that we are unable to properly narrow down the reparented target, which means we are calling `createStyleSheet` every time `appui:reparent` is dispatched anywhere in the window. We do have additional checks for stylesheet, so we're not unnecessarily recreating lots of stylesheets. Just something to be aware of and improve in the future. 

## Testing

Since this is AppUI-specific functionality, it is hard for us to test directly. But it was confirmed to work in AppUI's own test app (see [thread below](https://github.com/iTwin/iTwinUI/pull/2252#discussion_r1765882605)).

I did add a small e2e test to ensure we are not regressing on the most basic use of popout.

## Docs

Added changeset.


packages/itwinui-react/src/utils/components/ShadowRoot.tsx
```diff
@@ -127,5 +135,13 @@ function useShadowRoot(
     }
   }, [css]);
 
+  // Re-create stylesheet if the element is moved to a different window (by AppUI)
+  React.useEffect(() => {
+    window.addEventListener('appui:reparent', createStyleSheet);
+    return () => {
+      window.removeEventListener('appui:reparent', createStyleSheet);
+    };
+  }, [createStyleSheet, latestShadowRoot]);
```
mayank99, line 144: @GerardasB What do you think would be the best way to test this? It would be difficult to add a proper test for this in iTwinUI without manually recreating the custom event / reparenting logic.

I'm thinking we first manually verify that it works within the AppUI repo (by editing `@itwin/itwinui-react` in `node_modules`) and then after this is released, we can update the existing [AppUI test scenario](https://github.com/iTwin/appui/blob/6d28327aed77154e0814fcde526f78fee4c18cf2/e2e-tests/tests/popout-widget.test.ts#L207) to ensure the styles look fine without any special handling on top of `ProgressRadial`.

GerardasB: Sounds good, I'll verify if this change works and let you know.

GerardasB: I can confirm, this works correctly in the reparenting use-case (I made sure to remove the custom `useEffect` workaround https://github.com/iTwin/appui/blob/b1c1d0c95a58ebbf2985bb2e61ad01d639e7f795/apps/test-app/src/frontend/appui/frontstages/TestPopoutFrontstage.tsx#L90)

**However**, this change breaks the regular use-case, where no reparenting is done and the component is simply mounted in a new window. (`reparentPopoutWidgets=0` to configure in AppUI test-app: https://github.com/iTwin/appui/blob/b1c1d0c95a58ebbf2985bb2e61ad01d639e7f795/e2e-tests/tests/popout-widget.test.ts#L210)

This happens because `useEffect` that is scheduled in `useLatestRef(shadowRoot)` runs **after** the `createStyleSheet` called after `ReactDOM4.flushSync()`.

mayank99: Thanks for testing it. I believe I've fixed the regular popout case in my latest changes. I'm passing the already existing `shadow` into `createStyleSheet` (rather than relying on `useLatestRef`).

I also added a basic test for it, since this test does not require us to reimplement any reparenting logic.

GerardasB: Looks good now!

GerardasB (approved):

PR #2263, opened 2024-09-25
Fix missing default `Tooltip` delay

r100-stack:
## Changes

This PR fixes the `Tooltip` by properly adding the intended default delay.

`Tooltip.tsx`'s previous delay logic:

https://github.com/iTwin/iTwinUI/blob/6639cb9f44116114f460ca72f46cc8359c1170c9/packages/itwinui-react/src/core/Tooltip/Tooltip.tsx#L196-L200

Since `useDelayGroup().delay` is `0` ([docs](https://floating-ui.com/docs/floatingdelaygroup#troubleshooting)) when `FloatingDelayGroup` doesn't exist, the part to the right of the `??` was unreachable.

This PR fixes that `??` logic by using the `FloatingDelayGroup` delay if it exists (i.e. `delay !== 0`). Else, it uses the default delay. One down-side to this simple approach is that if `FloatingDelayGroup` exists and its delay is set to `0`, the delay for open will use the default delay instead of `0`. However, the close delay will still correctly be `0`. (https://github.com/iTwin/iTwinUI/pull/2263#discussion_r1775617600)

<details>
<summary>Old approach</summary>

To add the intended default delay, this PR uses the delay from creates a wrapper component called `FloatingDelayGroup` over floating ui's component with the same name. This wrapper sets a context which we then read in `Tooltip`. I went this way as I didn't see a way to know whether floating ui's `FloatingDelayGroup` exists up the tree or not using just floating ui's API.

</details>

## Testing

[Tooltip](https://itwin.github.io/iTwinUI/react/?story=tooltip--bottom&theme=dark):

|Old|New|
|-|-|
|<video src="https://github.com/user-attachments/assets/19e8eb4e-4cef-413d-8e20-76069f41663b" />|<video src="https://github.com/user-attachments/assets/4426c264-146e-45d0-aa63-27aacc0a8878" />|

[ButtonGroup + Tooltip](https://itwin.github.io/iTwinUI/react/?story=buttongroup--overflow&theme=dark): unchanged

https://github.com/user-attachments/assets/e7703474-0355-467f-ae69-02a5c08dbf97

## Docs

Added changeset.

### DIFF
--- a/apps/react-workshop/src/Breadcrumbs.test.ts
+++ b/apps/react-workshop/src/Breadcrumbs.test.ts
@@ -27,6 +27,7 @@ describe('Breadcrumbs', () => {
         cy.get('button').eq(1).click();
       } else if (testName === 'Custom Overflow Back Button') {
         cy.get('button').eq(1).trigger('mouseenter');
+        cy.wait(60);
       }
 
       cy.compareSnapshot(testName);


packages/itwinui-react/src/utils/components/FloatingDelayGroup.tsx
mayank99, line 1: Can't we just check the value against `0` instead of creating this abstraction?

mayank99: You can test it out. I believe it will never be `0` (see [code](https://github.com/floating-ui/floating-ui/blob/ecd627e4676a50643edcdf8915a6acefa50ee3d8/packages/react/src/components/FloatingDelayGroup.tsx#L143-L146)).

mayank99: Yeah I don't think we would ever want to use `0` explicitly, so it's ok to not handle this case.

mayank99 (approved):
LGTM. Let's make a patch release after merging this

smmr-dn (approved):

PR #2270, opened 2024-09-26
Add missing labels to `IconButton`s

r100-stack:
## Changes

Adds `label`/`aria-label` to `IconButton` in stories, examples, and one in iTwinUI component code.

## Testing

Manually verified that no more `IconButton`s without labels exist.

## Docs

Didn't add a changeset since I don't think adding one missing `aria-label` is significant enough to mention.

## After PR TODOs:

- [x] See if we can improve the label for the Table row expand button for expandable sub-rows and for expandable subComponent. (https://github.com/iTwin/iTwinUI/pull/2270#discussion_r1778725135).

### DIFF
--- a/apps/react-workshop/src/ButtonGroup.stories.tsx
+++ b/apps/react-workshop/src/ButtonGroup.stories.tsx
@@ -61,33 +61,30 @@ export const Overflow = () => {
   return (
     <ButtonGroup
       orientation='horizontal'
-      overflowButton={(overflowStart) => {
-        return (
-          <DropdownMenu
-            menuItems={(close) => {
-              const length = items.length - overflowStart;
-
-              return Array.from({ length }, (_, _index) => {
+      overflowButton={(overflowStart) => (
+        <DropdownMenu
+          menuItems={(close) =>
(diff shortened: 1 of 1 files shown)


packages/itwinui-react/src/core/Table/columns/expanderColumn.tsx
```diff
@@ -63,6 +63,7 @@ export const ExpanderColumn = <T extends Record<string, unknown>>(
       } else {
         return (
           <IconButton
+            aria-label='Toggle expandable content'
```
mayank99, line 66: These labels should ideally be unique, so that assistive technology users are able to differentiate between each instance more easily. Typically, this is done by combining two labels, i.e. "Toggle expandable content, [name of the row]".

How do we handle this for regular expandable sub-rows?

mayank99: Ok, then let's go with this for now, since it's better than nothing. But we should improve it in the future.

mayank99 (approved):

smmr-dn (approved):

PR #2271, opened 2024-09-27
Updated documentation for Progress Indicators

smmr-dn:
## Changes

Updated documentation for both Linear and Radial progress indicators.

## Testing

N/A

## Docs

N/A

## After-PR TODO:
- [x] Fix bug in `ProgressLinear` to properly merge `style` prop https://github.com/iTwin/iTwinUI/pull/2271#discussion_r1785243984. Fixed in https://github.com/iTwin/iTwinUI/pull/2285
- [ ] Investigate why `client:visible` does not work for `isAnimated` example https://github.com/iTwin/iTwinUI/pull/2271#discussion_r1786789359


apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+This prop is only available for radial indicators.
+
+<LiveExample src='ProgressRadial.size.jsx'>
+  <AllExamples.ProgressRadialSizeExample client:load />
+</LiveExample>
+
+### Content
+
+The radial progress indicator can be wrapped around JSX elements, such as icons or texts. This content helps provide flexibility to enhance the progress display with additional visual appearance. This prop is only available for radial indicators.
```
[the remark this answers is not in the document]
smmr-dn: I have mentioned this in the `Content` description.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
-### Props
+<LiveExample src='ProgressRadial.main.jsx' truncate={false}>
+  <AllExamples.ProgressRadialMainExample client:load />
+</LiveExample>
 
-<PropsTable path='@itwin/itwinui-react/esm/core/ProgressIndicators/ProgressLinear.d.ts' />
+### Value
 
-## Progress radial
+To specify a value for indeterminate indicator, pass a number into the `value` prop. This prop is available for both linear and radial indicators. If the `indeterminate` prop is set to `true`, this value will be ignored. We recommend keeping the value inclusively between 0 and 100.
```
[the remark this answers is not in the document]
smmr-dn: I have rephrased this.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
 thumbnail: ProgressIndicator
 ---
 
 <p>{frontmatter.description}</p>
 
-## Progress linear
+## Usage
 
-<Placeholder componentPageName='progresslinear--determinate' />
+There are two types of progress indicators which are linear and radial. Linear indicators show progress on a linear bar, while radial indicators display progress in a circular form.
```
mayank99, line 11: We should use the exact component names here, and maybe use a list for readability.

```suggestion
There are two types of progress indicators available as separate components:

- `ProgressLinear`: shows progress on a linear bar
- `ProgressRadial`: shows progress in a circular form.
```

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+The default progress indicator is indeterminate. If a [value](#value) is passed, the indicator becomes determinate.
 
 <LiveExample src='ProgressLinear.main.jsx' truncate={false}>
```
mayank99, line 15: Could some of these examples be combined? The demos are taking up too much space.

I'm thinking `ProgressLinear.main` and `ProgressRadial.main` could be combined. Similarly, `ProgressLinear.value` and `ProgressRadial.value` could be combined.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
-### Props
+<LiveExample src='ProgressRadial.main.jsx' truncate={false}>
+  <AllExamples.ProgressRadialMainExample client:load />
+</LiveExample>
 
-<PropsTable path='@itwin/itwinui-react/esm/core/ProgressIndicators/ProgressLinear.d.ts' />
+### Value
 
-## Progress radial
+To specify a value for indeterminate indicator, pass a number into the `value` prop. This prop is available for both linear and radial indicators. If the `indeterminate` prop is set to `true`, this value will be ignored. This value must be inclusively between 0 and 100.
```
mayank99, line 25: Is "indeterminate" a typo?

Also some slight rewording/moving things around for readability and clarification around the purpose of `value`

```suggestion
To specify a value for determinate progress indicator, the `value` prop can be used. This prop is available for both linear and radial indicators. The value must be inclusively between `0` and `100`, representing what percentage of the task has completed.

If the `indeterminate` prop is set to `true`, this value will be ignored.
```

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+Exclusively for linear indicators, the `isAnimated` flag is helpful when applying animation to the indicator as the value changes.
+
+<LiveExample src='ProgressLinear.isAnimated.jsx'>
+  <AllExamples.ProgressLinearAnimatedExample client:load />
+</LiveExample>
+
+### Status
+
+Three available statuses for both linear and radial progress indicators are:
```
mayank99, line 43: We should clarify the purpose of these statuses. It's not something that makes sense by default. I imagine that the desired UX flow is to show a default progress indicator, and then set a status after it completes or fails.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+Indicator labels can be passed into the `labels` prop as an array. If one label is passed, it will be centered. If there are two labels, each is pushed to one end of the progress indicator. This prop is only available for linear indicators.
+
+<LiveExample src='ProgressLinear.label.jsx'>
+  <AllExamples.ProgressLinearLabelExample client:load />
+</LiveExample>
+
+### Size
+
+Four size options available are:
```
mayank99, line 69: The `size` prop needs to be explicitly mentioned.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+- `"positive"`
+- `"negative"`
+- `"warning"`
+
+<LiveExample src='ProgressLinear.status.jsx'>
+  <AllExamples.ProgressLinearStatusExample client:load />
+</LiveExample>
+
+Each radial indicator with status is accompanied by an icon, providing a visual cue to inform the current progress state better. The icon can be replaced by a [content](#content).
```
mayank99, line 53: Is this paragraph redundant? Feels a bit out of place in the "Status" section.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
+
+This prop is only available for radial indicators.
+
+<LiveExample src='ProgressRadial.size.jsx'>
+  <AllExamples.ProgressRadialSizeExample client:load />
+</LiveExample>
+
+### Content
+
+The radial progress indicator can be wrapped around JSX elements, such as icons or texts. This content helps provide flexibility to enhance the progress display with additional visual appearance. Moreover, indicators of `"x-small"` size will not display its content. This prop is only available for radial indicators.
```
mayank99, line 84: The "moreover" sentence could be moved below the demo and turned into a "Note:", so that it doesn't distract from the main point here.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,29 +1,90 @@
...
 
 <p>{frontmatter.description}</p>
 
-## Progress linear
+## Usage
 
-<Placeholder componentPageName='progresslinear--determinate' />
+There are two types of progress indicators which are linear and radial. Linear indicators show progress on a linear bar, while radial indicators display progress in a circular form.
+
+The default progress indicator is indeterminate. If a [value](#value) is passed, the indicator becomes determinate.
```
mayank99, line 13: This paragraph could probably be removed/rephrased to fit into the "Value" section.

apps/website/src/content/docs/progressindicator.mdx
mayank99: This comment is actually applicable throughout the page. Most of the references to "linear indicator" and "radial indicator" could be replaced with the component names.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: Hmmm, probably I should just split it and move the second sentence where value is mentioned to the "Value" section?

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have moved it to a "Note".

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have made the change.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have added more wording to this. Does it sound okay?

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have made the change.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have explicitly mentioned these components.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have combined the examples. There are two problems I encountered:
- The linear indicator has to be specified its width so that it could show up with the radial one.
- When I passed width and value, the linear indicator does not have the correct filled value.

I'm not sure what to do to correct the styling when having these together.

apps/website/src/content/docs/progressindicator.mdx
mayank99: I actually meant it should be _below the demo_ (at the end of the section). In my mind, the demo should come first because it's not really affected by the note.

apps/website/src/content/docs/progressindicator.mdx
mayank99: Similar to https://github.com/iTwin/iTwinUI/pull/2271#discussion_r1784868852, this note about `indeterminate` could be prefixed with "Note:" and moved below the demo? Since there are multiple demos in this section, we can put the note below everything, at the bottom of this section.

apps/website/src/content/docs/progressindicator.mdx
mayank99: Hmm, I think if you use `flex-direction: column` on the container, then it should work better? We shouldn't need to pass an explicit `width` to the progress indicators.

apps/website/src/content/docs/progressindicator.mdx
mayank99: Slight rewording and cleanup:

> The `status` prop can be used to convey the outcome of a process. The recommended UX flow is to display the default progress indicator during the loading process, and then update the `status` to reflect whether the process was successful, encountered an error, or requires attention.
>
> Three available statuses are:

apps/website/src/content/docs/progressindicator.mdx
mayank99: How about we keep it in the section but move it *below* the demo and expand on it to clarify what "indeterminate" is. This segues nicely into the next section.

> By default, the progress indicators are considered to be "indeterminate". This is useful for short processes, where it isn't necessary (or possible) to indicate the completion percentage. For longer processes, a "determinate" progress indicator should be used.

examples/ProgressLinear.isAnimated.jsx
```diff
@@ -0,0 +1,23 @@
...
+import { ProgressLinear } from '@itwin/itwinui-react';
+
+export default () => {
+  const [value, setValue] = React.useState(0);
+
+  React.useEffect(() => {
+    setTimeout(() => {
+      const newValue = value + 20;
+      setValue(newValue > 100 ? 0 : newValue);
+    }, 3000);
```
mayank99, line 15: This demo can be improved in a few ways.

1. Shorten the timeout from 3s to 2s (or 1.5s) so that the animation is more obvious.
2. Don't loop it; it can be a bit annoying when you're just trying to read the docs. Instead, display a completion message and set the `status`.
3. Provide a "Reset" button for restarting the timeout, in case the user missed the animation.
4. In the MDX, change the client directive from `client:load` to `client:visible`, so that it gets triggered when the demo comes into view. (See [docs](https://docs.astro.build/en/reference/directives-reference/#clientvisible))

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have put this below the demo.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have made the change.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have moved this below the demo.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have made the change.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: `flex-direction: column` seems to have the same problem :(

examples/ProgressLinear.isAnimated.jsx
smmr-dn: I have added a message and status as well as a reset button.

apps/website/src/content/docs/progressindicator.mdx
mayank99: So there are two problems here:
1. `align-items: center` tries to collapse the items. To fix this, we can add `.demo-container > * { width: 100% }` in the CSS, instead of the inline style.
2. There is actually a bug in `ProgressLinear`: it does not correctly merge the `style` prop. We should fix this in a separate PR.
   https://github.com/iTwin/iTwinUI/blob/76613b5ffbe04a7ad899cd7c39a4dbc9dd0b99b1/packages/itwinui-react/src/core/ProgressIndicators/ProgressLinear.tsx#L85-L90

examples/ProgressLinear.isAnimated.jsx
mayank99: It doesn't seem to work in the [deploy preview](https://deploy-preview-2271--itwinui-previews.netlify.app/docs/progressindicator/). It's probably because of `useEffect` running too often (since it clears the timeout every time the `value` changes).

To fix this, you can create a variable (outside the `useEffect`): `const isDone = value === 100;` and use that variable (instead of `value`) in the `useEffect`.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: I have added this as an after-PR todo. Can I use `:first-child` instead of grabbing all direct children? It caused a pretty odd visual error
image

examples/ProgressLinear.isAnimated.jsx
smmr-dn: I have used `isDone` instead of `value` but for some reason it didn't work. I tried switching back to `client:load` and the demo ran so I'm not sure if that's the problem.

apps/website/src/content/docs/progressindicator.mdx
mayank99: `:first-child` is not reliable because the order of elements might change. Can you try removing `align-items: center` and centering using `.demo-container > * { margin-inline: auto }`?

examples/ProgressLinear.isAnimated.jsx
mayank99: You are still using `value` in the dependency array of `useEffect`, it needs to be just `[isDone]`.

apps/website/src/content/docs/progressindicator.mdx
smmr-dn: Hmm, I still have the same problem as above.

mayank99: Sorry I should have clarified: the `width: 100%` also needs to be removed. It was only necessary when `align-items: center` was used.

But it looks like `margin-inline: auto` reintroduces the problem where the linear indicator collapses entirely.

For the sake of simplicity, I would say let's just *not* center it. So this would be the _entire_ CSS:

```css
.demo-container {
  display: flex;
  flex-direction: column;
  width: 200px;
  gap: var(--iui-size-s);
}
```

smmr-dn: I have made the change.

examples/ProgressLinear.isAnimated.jsx
mayank99: The other issue I noticed is that `setTimeout` should be `setInterval` so that it continuously re-runs until it's cleared.

If it still does not work, it might be a bug in `client:visible`. Let's use `client:load` for now and investigate later.

smmr-dn: I have switched to `setInterval` and added the `client:visible` issue as an after-PR todo item.

mayank99: Some more suggestions:

- We don't need to use a ref for the timeout; it can be a local variable inside the Effect.
- We should turn off `isAnimated` when value reaches 100, to avoid unnecessary/artificially delayed value. (It looks strange when the green progress bar animates).
- We should change the `key` when resetting the process, to force a remount and avoid unnecessary animation when the value goes from 100 to 0.

<details><summary>Full code</summary>

```jsx
export default () => {
  const [value, setValue] = React.useState(0);
  const isDone = value === 100;

  React.useEffect(() => {
    let interval;

    if (!isDone) {
      interval = setInterval(
        () => setValue((prevValue) => prevValue + 20),
        1500,
      );
    }

    return () => clearTimeout(interval);
  }, [isDone]);

  return (
    <div className='demo-container'>
      <ProgressLinear
        value={value}
        isAnimated={!isDone}
        labels={!isDone ? ['Loading...', `${value}%`] : ['Upload succeeded.']}
        status={isDone ? 'positive' : undefined}
        key={isDone.toString()}
      />
      <Button onClick={() => setValue(0)}>Reset process</Button>
    </div>
  );
};
```

</details>

smmr-dn: I have done as suggested.

mayank99: `clearTimeout` → `clearInterval` (typo in my original suggestion, sorry)

mayank99 (approved):
LGTM 🚀

examples/ProgressLinear.isAnimated.jsx
smmr-dn: Oh I just noticed that when you commented too. I have fixed it!

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,37 +1,105 @@
...
-## Progress linear
+## Usage
 
-<Placeholder componentPageName='progresslinear--determinate' />
+There are two types of progress indicators available as separate components:
 
-<LiveExample src='ProgressLinear.main.jsx' truncate={false}>
-  <AllExamples.ProgressLinearMainExample client:load />
+- `ProgressLinear`: shows progress on a linear bar
+- `ProgressRadial`: shows progress in a circular form.
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

apps/website/src/content/docs/progressindicator.mdx
```diff
@@ -1,37 +1,105 @@
...
 </LiveExample>
 
-### Props
+By default, the progress indicators are considered to be "indeterminate". This is useful for short processes, where it isn't necessary (or possible) to indicate the completion percentage. For longer processes, a "determinate" progress indicator should be used.
 
-<PropsTable path='@itwin/itwinui-react/esm/core/ProgressIndicators/ProgressLinear.d.ts' />
+### Value
 
-## Progress radial
+To specify a value for determinate progress indicator, the `value` prop can be used. This prop is available for both `ProgressLinear` and `ProgressRadial`. The value must be inclusively between `0` and `100`, representing what percentage of the task has completed.
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

PR #2274, opened 2024-10-01
`TransferList.Toolbar` implements toolbar pattern

r100-stack:
## Changes

After #2263, tooltips now have a default delay. This, in addition to the vertical placement of the tooltips in `TransferList.Toolbar` make it difficult to go from one button to another.

This PR fixes this by using a `<ButtonGroup orientation="vertical" />` for `TransferList.Toolbar` under the hood. The tooltips now show to the right (#1709) and with no delay when going from one button to the adjacent one (#1524).

Additionally, this also improves accessibility since `<ButtonGroup role="toolbar">` implements the [toolbar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) with arrow-key navigation functionality.

## Testing

||Before|After|
|-|-|-|
|Tooltips|<video src="https://github.com/user-attachments/assets/b0c4ef76-28d3-4838-a983-900bb1128192"/>|<video src="https://github.com/user-attachments/assets/4d9f2997-f4f0-4bd5-b550-cbeed86f7777"/>|
|arrow-key functionality|<video src="https://github.com/user-attachments/assets/1c2e8293-3cc4-476f-94f2-86761c373bd7" />|<video src="https://github.com/user-attachments/assets/e84ecccc-b459-42e2-a485-b8659e427d0e" />|

## Docs

Added minor changesets.


.changeset/orange-cats-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
```
mayank99, line 2: Why are this changesets `minor`? Can't we release this PR as `patch`?

testing/e2e/app/routes/TransferList/spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+test('should support toolbar keyboard navigation in TransferList.Toolbar', async ({
+  page,
+}) => {
+  await page.goto('/TransferList');
+  const buttons = page.locator('button');
+
+  await page.keyboard.press('Tab');
+
+  await page.keyboard.press('Tab');
```
mayank99, line 11: After pressing <kbd>Tab</kbd>, we should validate focus. This will help with readability (knowing what we are tabbing to) and with maintainability (if we add more tab stops, this test will be easier to modify).

Also, it would be good to validate `role="toolbar"` somewhere. It feels brittle that we are looking for _all_ buttons on the page.

testing/e2e/app/routes/TransferList/route.tsx
mayank99, line 1: nit: Could this file be simplified a bit?

mayank99: Ok, here's an alternative suggestion: Could we import one of the `examples` instead of redefining the entire component? My main concern is that there is a lot of code in the test route, which isn't necessarily relevant or unique for testing.

mayank99: I meant actually importing from the `examples` package, not using relative path.

Also we should still define a function to make it easier to modify the test in the future.

```jsx
import { TransferListMainExample } from 'examples';

export default function Page() {
  return <TransferListMainExample />
}
```

mayank99 (approved):

PR #2275, opened 2024-10-01
`Table`'s `ActionType.type` type made more specific

r100-stack:
## Changes

A user wanted to know all the actions we support. This was difficult to know since some actions are defined by react-table while others are defined by iTwinUI. So there is no one place to find all types.

The react-table types listed `ActionType.type` as just a `string` instead of a union of string literals.

https://github.com/iTwin/iTwinUI/blob/090aa3d18e03970ecd2740193a8a4c7ef01ffb0f/packages/itwinui-react/src/react-table/react-table.ts#L1197-L1198

To make it clear what actions we use, this PR replaces `string` with a union of string literals. This union also includes [`AnyString`](https://github.com/iTwin/iTwinUI/blob/090aa3d18e03970ecd2740193a8a4c7ef01ffb0f/packages/itwinui-react/src/utils/types.ts#L9) just as a safety in case somehow an action is used (e.g. we add a new hook to react-table and forget to update `PossibleActionType`).

Split the actions into react-table actions and custom actions. This makes it easier to maintain. (https://github.com/iTwin/iTwinUI/pull/2275#discussion_r1783681186).

To generate the string literals, I just did a `console.log(Object.keys(actions))` in Table related code and glanced through [react-table v7 code](https://redirect.github.com/TanStack/table/tree/v7).

## Testing

Confirmed that the type shows up when using the `Table` component.

|Before|After|
|-|-|
|image|image|

## Docs

Added patch changeset since it's more like a fix of the type instead of adding something new.

### DIFF
--- a/packages/itwinui-react/src/react-table/react-table.ts
+++ b/packages/itwinui-react/src/react-table/react-table.ts
@@ -1233,7 +1233,15 @@ type BuiltInAction =
   | 'toggleSortBy'
   | 'clearSortBy';
 
-type CustomAction = 'setScrolledLeft' | 'setScrolledRight';
+type CustomAction =
+  | 'setScrolledLeft'
+  | 'setScrolledRight'
+  | 'columnDragStart'
+  | 'columnDragEnd'
+  | 'singleRowSelected'
+  | 'shiftRowSelected'
+  | 'tableResizeStart'
+  | 'tableResizeEnd';
 
 type PossibleActionType = BuiltInAction | CustomAction | AnyString;
 


packages/itwinui-react/src/react-table/react-table.ts
```diff
@@ -1193,9 +1193,52 @@ export interface SortingRule<D> {
 
 //#endregion
 
+type ActionTypes =
```
mayank99, line 1196: Can we split this into two types (built-in actions and custom actions) and define `ActionType` using a union `BuiltInAction | CustomAction | AnyString`?

It doesn't change anything for the user, but would help with our own maintainability.

also nit: the type name should be singular, not plural, since it is equal to *one* of these values.

packages/itwinui-react/src/react-table/react-table.ts
```diff
@@ -1193,9 +1193,53 @@ export interface SortingRule<D> {
...
+  | 'toggleAllPageRowsSelected'
+  | 'setRowState'
+  | 'setCellState'
+  | 'resetRowState'
+  | 'resetSortBy'
+  | 'setSortBy'
+  | 'toggleSortBy'
+  | 'clearSortBy';
+
+type CustomAction = 'setScrolledLeft' | 'setScrolledRight';
```
mayank99, line 1236: There are a few more custom actions defined here:

https://github.com/iTwin/iTwinUI/blob/090aa3d18e03970ecd2740193a8a4c7ef01ffb0f/packages/itwinui-react/src/core/Table/Table.tsx#L71-L74

mayank99 (approved):
LGTM

smmr-dn (approved):

PR #2278, opened 2024-10-02
Add custom table actions to `react-table`'s `actions`

r100-stack:
## Changes

After https://github.com/iTwin/iTwinUI/pull/2275#discussion_r1784802956, I realized that some of our custom actions were not added to the `actions` object from `react-table`. This PR adds those remaining custom actions to the `actions` object from `react-table`.

## Testing

Did a `console.log(Object.keys(actions))` from Table code to confirm that those remaining actions now exist in the `actions` object.

## Docs

N/A since not user-facing.

### DIFF
--- a/packages/itwinui-react/src/core/Table/Table.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.tsx
@@ -696,7 +696,7 @@ export const Table = <
           (selectionMode === 'single' || !ctrlPressed)
         ) {
           dispatch({
-            type: actions.shiftRowSelected,
+            type: actions.singleRowSelected,
             id: row.id,
           });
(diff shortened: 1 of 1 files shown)


mayank99 (changes requested):
This change is not necessary. It's fine for `actions` from `react-table` to not include custom actions.

We should never modify anything that is imported from dependencies. The dependency should be considered pure and opaque. 

Changing imported objects like this can make the flow of code hard to understand, both from a maintainability perspective and from a static analysis perspective (e.g. bundlers might have difficulties with tree-shaking). Plus it could cause potential issues for consumers if they are also relying on `react-table` directly.

PR #2283, opened 2024-10-04
Workaround for unscrollable sidenav in react-workshop

r100-stack:
## Changes

@mayank99 noticed that the sidenav is _not_ scrollable in newer versions of Chrome (e.g. 129.0, Canary 131). But it works in Edge 128.0, Firefox 130.0, and Safari 16.6.

Ladle already has an issue for this: [tajo/ladle#574](https://redirect.github.com/tajo/ladle/issues/574)

Until they investigate why it doesn't work in Chrome and until either we or Ladle do a proper fix, this PR adds custom styles to make the sidenav scrollable. This is _important_ as users using the `react-workshop` on Chrome might not be able to see all demos.

## Testing

The sidenav is now scrollable in Chrome 129.0:

|Before|After|
|-|-|
|image|image|

## Docs

N/A

### DIFF
--- a/apps/react-workshop/.ladle/global.css
+++ b/apps/react-workshop/.ladle/global.css
@@ -28,7 +28,7 @@ main {
   transform: translateZ(0);
 }
 
-/* TODO: Remove when the sidenav scroll works out-of-the-box in newer versions of Chrome. See #2283 */
+/* TODO: Remove when the sidenav scroll is fixed in Chrome. See https://github.com/tajo/ladle/issues/574 */
 nav.ladle-aside {
   min-height: unset;
 }


apps/react-workshop/.ladle/global.css
```diff
@@ -27,3 +27,8 @@ main {
   position: relative;
   transform: translateZ(0);
 }
+
+/* TODO: Remove when the sidenav scroll works out-of-the-box in newer versions of Chrome. See #2283 */
```
mayank99, line 31: I found a bug report in the ladle report which we can link to instead.

```suggestion
/* TODO: Remove when the sidenav scroll is fixed in Chrome. See https://github.com/tajo/ladle/issues/574 */
```

mayank99 (approved):

PR #2287, opened 2024-10-07
Reduced props accepted by `ColumnHeader`

smmr-dn:
## Changes

Reduced the number of props accepted by `ColumnHeader` as some of these props are not specific to individual columns. (mentioned here https://github.com/iTwin/iTwinUI/pull/2178#pullrequestreview-2211294338)

## Testing

N/A

## Docs

N/A


packages/itwinui-react/src/core/Table/ColumnHeader.tsx
```diff
@@ -47,16 +45,15 @@ export const ColumnHeader = <
   const {
     columnRefs,
     column,
-    index,
     areFiltersSet,
-    hasAnySubRows,
-    headers,
     state,
```
mayank99, line 49: Can `state` also be removed? I assume we can get it from `instance`.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
```diff
@@ -47,16 +45,15 @@ export const ColumnHeader = <
   const {
     columnRefs,
```
mayank99, line 46: Can `columnRefs` be removed? It doesn't feel quite right for a `ColumnHeader` to mutate its parent component's variables.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: Would you mind elaborating more on getting `state` from `instance`? Just want to make sure I get it right.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I have removed this prop.

mayank99: You have removed the whole functionality, not just the prop. What I was asking for is to move the `columnRefs`-related code from `ColumnHeader` into `Table`. 

You can pass a `ref` from the outside.

```js
ref={(el) => {
  if (el) {
    columnRefs.current[column.id] = el;
  }
}}
```

And the refs will need to be merged inside `ColumnHeader`.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I have stored `instance` within the `TableInstanceContext` but there are some points that I'm not sure about. I'm feeling really uneasy about passing `any` its type parameter but I'm currently having a bit of a hard time finding a way to declare/extend generic type `T` in the context. What should I do to assign type appropriately here? Also, should I create a `useContext` hook and wrap provider around a wrapper component so I won't have to pass `instance` as a prop?

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I have the same concern regarding generic type `T` when making this a `forwardRef` component to pass in `ref` from its parent.

mayank99: I don't think we even need to use a generic in `ColumnHeader`. We can get rid of it because it's an internal component. In the places where `T` is used inside the component, we can use `Record<string, unknown>`.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
mayank99: Similar to https://github.com/iTwin/iTwinUI/pull/2287#discussion_r1792467871, the generic types are not super important. We can use `Record<string, unknown>`.

I'm not sure what you mean by "create a `useContext` hook and wrap provider around a wrapper component". You can remove the `instance` prop and replace it with this:

```js
const instance = React.useContext(TableInstanceContext);
```

smmr-dn: I have made this change.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I have replaced `T` with `Record<string, unknown>`.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
mayank99: There is one part of this thread still pending. I said earlier:

> One thing that we'll need to verify is whether `instance` is a stable/memoized object, i.e. it doesn't change on every re-render.

Can you verify this? You can check the source code for `react-table` v7 and also try something like this code locally:

```js
useEffect(() => { console.log("changed", instance) }, [instance])
```

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
mayank99: I noticed two things in `ColumnHeader.tsx`:
1. The ref callback is no longer memoized. This undoes the performance improvement from https://github.com/iTwin/iTwinUI/pull/2178.
2. (nit) The forwarded ref is vaguely named `ref`, when it should instead be `forwardedRef`.

The code should look something like this:

```jsx
ref={useMergedRefs(
  React.useCallback((el: HTMLDivElement) => {
    if (el) {
      column.resizeWidth = el.getBoundingClientRect().width;
    }
  }, [column]),
  forwardedRef
)}
```

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I apologize I got sidetracked by the type issue so I totally missed this :(. I have checked the source code (https://github.com/TanStack/table/blob/v7/src/hooks/useTable.js) and seen that they use either `useMemo` or `useCallback` for the instance's properties, which I believe should result in the instance object being memoized. I have also tried your suggestion above and tried something like this to double check.

``` jsx
const prevInstanceRef = React.useRef(instance);

React.useEffect(() => {
  if (prevInstanceRef.current && prevInstanceRef.current !== instance) {
    console.log('Instance object has changed');
  }
  prevInstanceRef.current = instance;
}, [instance]);
```

So, I believe it should be stable.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
smmr-dn: I have made this change.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -1000,22 +1000,29 @@ export const Table = <
...
+                        }
+                        isLast={index === headerGroup.headers.length - 1}
+                        isTableEmpty={data.length === 0}
                         isResizable={isResizable}
                         columnResizeMode={columnResizeMode}
                         enableColumnReordering={enableColumnReordering}
                         density={density}
-                        visibleColumns={visibleColumns}
+                        ref={(el) => {
+                          if (el) {
```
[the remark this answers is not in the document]
mayank99: Can you clarify what this comment is about? Your comment is on the ref.

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
```diff
@@ -140,6 +115,17 @@ export const ColumnHeader = <
           isHeaderDirectClick.current = false;
         }
       }}
+      ref={useMergedRefs(
+        React.useCallback(
+          (el?: HTMLDivElement) => {
```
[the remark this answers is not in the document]
mayank99: It is intentional. React will not always call this callback with an element. That's why we check `if (el)` below.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -64,6 +62,7 @@ import {
   onTableResizeStart,
 } from './actionHandlers/index.js';
 import { SELECTION_CELL_ID } from './columns/index.js';
+import { TableInstanceContext } from './utils.js';
```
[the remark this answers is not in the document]

packages/itwinui-react/src/core/Table/ColumnHeader.tsx
```diff
@@ -140,6 +115,17 @@ export const ColumnHeader = <
           isHeaderDirectClick.current = false;
         }
       }}
+      ref={useMergedRefs(
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

mayank99 (approved):

PR #2288, opened 2024-10-07
Added console warning for `Table` when both `subRows` and `subComponent` exist

smmr-dn:
## Changes

Added console warning when users pass in `subRows` and `subComponent` for the `Table` at the same time, mentioned in one of the comments in this PR: https://github.com/iTwin/iTwinUI/pull/2208.

## Testing

Verified that the warning appears when there are both `subRows` and `subComponent` in the `Table`.

## Docs

Added changeset


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -656,6 +656,12 @@ export const Table = <
     }
   }
 
+  if (data.some((item) => (item.subRows as T[]).length > 0) && subComponent) {
```
mayank99, line 659: Can you also add a check for `process.env.NODE_ENV === 'development'`?

smmr-dn: I have added the if condition to check for the dev environment.

mayank99: it should be made part of the same `if` condition instead of a nested one.

mayank99 (approved):
LGTM. I verified that the warning appears when using both `subRows` and `subComponent`.




You should also mention in the PR description how you tested it.

Also, this is a user-facing change, so I think it needs a `patch` changeset.

smmr-dn:
> LGTM. I verified that the warning appears when using both `subRows` and `subComponent`.
> 
> 
> 
> You should also mention in the PR description how you tested it.
> 
> Also, this is a user-facing change, so I think it needs a `patch` changeset.

I have modified the PR description and added a changeset.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -656,6 +656,16 @@ export const Table = <
     }
   }
 
+  if (
+    process.env.NODE_ENV === 'development' &&
+    data.some((item) => (item.subRows as T[]).length > 0) &&
+    subComponent
```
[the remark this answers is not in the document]

mayank99 (approved):

PR #2289, opened 2024-10-07
Force bump `cookie` to fix audit

r100-stack:
## Changes

Fix failing [audit](https://github.com/iTwin/iTwinUI/actions/runs/11224444759/job/31201196122?pr=2001) by addressing https://github.com/advisories/GHSA-pxg6-pf52-xh8x.

Had to use an override since `express`'s latest version [4.21.0](https://unpkg.com/browse/express@4.21.0/package.json#L36) and `@remix-run/server-runtime`'s latest version [2.12.1](https://unpkg.com/browse/@remix-run/server-runtime@2.12.1/package.json#L22) don't support `cookie>=0.7.0` in their dep version range.

Similar to https://github.com/iTwin/kiwi/pull/69.

## Testing

Audit passes.

## Docs

Since the deps are from `e2e` (not completely sure since I think `pnpm why cookie` doesn't show _all_ paths), i.e. from a non-package workspace, didn't add any changeset.

`pnpm why cookie` from before this PR:

image

## After PR TODOs

* [ ] (Not important) Maybe change [`resolutions`](https://pnpm.io/package_json#resolutions) to [`pnpm.overrides`](https://pnpm.io/package_json#pnpmoverrides) in `package.json`. ([#2289 (comment)](https://github.com/iTwin/iTwinUI/pull/2289#discussion_r1791963954)).

### DIFF
--- a/.changeset/cuddly-gorillas-type.md
+++ b/.changeset/cuddly-gorillas-type.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Added warning log in browser when both `subRows` and `subComponent` are passed into `Table`.
--- a/apps/website/src/content/docs/progressindicator.mdx
+++ b/apps/website/src/content/docs/progressindicator.mdx
@@ -1,30 +1,92 @@
 ---
 title: Progress indicator
-description: Progress indicators express an unspecified wait time or display the length of a process.
+description: Progress indicators express the length of in-progress operations.
 thumbnail: ProgressIndicator
 ---
 
 <p>{frontmatter.description}</p>
 
-## Progress linear
+## Usage
 
-<Placeholder componentPageName='progresslinear--determinate' />
+There are two types of progress indicators available as separate components:
 
-<LiveExample src='ProgressLinear.main.jsx' truncate={false}>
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -85,5 +85,10 @@
     "node": ">=18.14.1"
   },
   "packageManager": "pnpm@9.1.0",
-  "prettier": "configs/prettier-config"
+  "prettier": "configs/prettier-config",
+  "pnpm": {
+    "overrides": {
+      "cookie@<0.7.0": ">=0.7.0"
```
mayank99, line 91: This needs to be under the `resolutions` field along with the rest of the overrides

pnpm-lock.yaml
mayank99, line 1: This file needs to be formatted, I think. It shouldn't have so many changes.

package.json
mayank99: Yes, we should use either `pnpm.overrides` or `resolutions`, but not both. We can change `resolutions` to `pnpm.overrides` in a future PR if we want, but it's not important.

mayank99 (approved):

smmr-dn (approved):

PR #2294, opened 2024-10-09
ComboBox filter reset behavior (`clearFilterOnOptionToggle`)

r100-stack:
## Changes

Adds on to the patch fix in #2295 to completely address #2276.

Added a new `clearFilterOnOptionToggle` prop (name bikesheddable) only when `multiple` is `true`. Possible values:
- `true` (default): Filter is cleared when an option is toggled. Useful when users would likely want to re-filter after toggling an option.
- `false`: Filter is _not_ cleared when an option is toggled. This is useful when users would likely want to toggle multiple options from the same filtered results.

Since the default value is `true`, which is the existing behavior before this PR, there is no change in the default behavior (no breaking change).

This prop allows the user to choose the behavior since as we see in #2276, both behaviors have benefits.

## Testing

* Edited existing story (could create a separate one if needed).
* Added docs. Didn't add an example since that didn't feel necessary and would unnecessarily take up space imo.
* Added e2e tests to test `clearFilterOnOptionToggle`.

## Docs

Added minor changeset.

### DIFF
--- a/.changeset/fuzzy-bikes-move.md
+++ b/.changeset/fuzzy-bikes-move.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Updated the default value of `Dialog`'s `setFocus` prop to be based on the value of the `trapFocus` prop.
--- a/packages/itwinui-react/CHANGELOG.md
+++ b/packages/itwinui-react/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 3.15.5
+
+### Patch Changes
+
+- [#2306](https://github.com/iTwin/iTwinUI/pull/2306): Fixed broken expandable content (`subComponent`) when `Table` is asynchronously re-rendered.
+- [#2306](https://github.com/iTwin/iTwinUI/pull/2306): Fixed `Table`'s max depth reached error caused when passing the `getRowId` prop.
+
 ## 3.15.4
 
 ### Patch Changes
--- a/packages/itwinui-react/package.json
+++ b/packages/itwinui-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/itwinui-react",
-  "version": "3.15.4",
+  "version": "3.15.5",
   "author": "Bentley Systems",
   "license": "MIT",
   "type": "module",
--- a/packages/itwinui-react/src/core/Dialog/Dialog.tsx
+++ b/packages/itwinui-react/src/core/Dialog/Dialog.tsx
@@ -25,7 +25,7 @@ type DialogProps = {
 const DialogComponent = React.forwardRef((props, ref) => {
   const {
     trapFocus = false,
-    setFocus = false,
+    setFocus = trapFocus,
     preventDocumentScroll = false,
     isOpen = false,
     isDismissible = true,
--- a/packages/itwinui-react/src/core/Dialog/DialogContext.tsx
+++ b/packages/itwinui-react/src/core/Dialog/DialogContext.tsx
@@ -35,13 +35,17 @@ export type DialogContextProps = {
    */
   closeOnEsc?: boolean;
   /**
-   * Traps the focus inside the dialog. This is useful when the dialog is modal.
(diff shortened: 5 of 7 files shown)


packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -82,12 +82,25 @@ export type ComboboxMultipleTypeProps<T> =
...
+       *
+       * If `true`, toggling an option will clear the filter.
+       * Useful when users would likely want to re-filter after toggling an option.
+       *
+       * If `false`, the filter will remain as-is after toggling an option.
+       * Useful when users would likely want to toggle multiple options from the _same_ filtered results.
+       *
+       * @default true
+       */
+      clearFilterOnOptionToggle?: never;
```
smmr-dn, line 96: I'm quite uncertain about the type `never` being used here instead of `boolean`. Would you mind elaborating a little more on this?

mayank99: I believe this TS pattern is referred to as "discriminated unions".
- https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions
- https://basarat.gitbook.io/typescript/type-system/discriminated-unions
- https://www.totaltypescript.com/discriminated-unions-are-a-devs-best-friend

smmr-dn: This is really helpful. Thank you!

packages/itwinui-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -481,22 +494,26 @@ export const ComboBox = React.forwardRef(
               .join(', '),
           );
 
-          // Clear the filter whenever an item is selected
-          setInputValue('');
+          // If clearFilterOnOptionToggle, clear the filter
+          setIsInputDirty(!clearFilterOnOptionToggle);
```
mayank99, line 498: Should this be written as an `if` condition? Because otherwise, the comment is a lie by omission (it's also calling `setIsInputDirty(true)` when `clearFilterOnOptionToggle` is false).

```suggestion
          if (clearFilterOnOptionToggle) {
            setIsInputDirty(false);
          }
```

This can probably be combined with the if-condition below.

apps/react-workshop/src/ComboBox.stories.tsx
```diff
@@ -519,16 +522,28 @@ export const MultipleSelect = () => {
...
-        setSelectedOptions(selected);
-      }}
-    />
+    <Flex flexDirection='column' alignItems='stretch'>
+      <Checkbox
+        checked={clearFilterOnOptionToggle}
+        onChange={(e) => setClearFilterOnOptionToggle(e.target.checked)}
+        label='clearFilterOnOptionToggle'
+      />
+      <Divider />
```
mayank99, line 535: There's some subjectiveness here, but I think this checkbox and divider should be:
1. defined using a decorator
2. be below the combobox and hidden from the visual test

It would be nice to keep the test image unchanged.

apps/website/src/content/docs/combobox.mdx
```diff
@@ -60,6 +60,11 @@ Set the property `multiple` to true to allow users to select more than one value
   <AllExamples.ComboBoxMultipleSelectExample client:load />
 </LiveExample>
 
+When `multiple` is `true`, a `clearFilterOnOptionToggle` prop is available. Possible values:
+
+- `true` (default): Filter is cleared when an option is toggled. Useful when users would likely want to re-filter after toggling an option.
+- `false`: Filter is _not_ cleared when an option is toggled. This is useful when users would likely want to toggle multiple options from the same filtered results.
+
```
mayank99, line 67: Should we add a demo here?

smmr-dn (approved):
I didn't have any further comments on this PR at the moment so I'm approving this. I will take a look back when you make changes later.

mayank99 (approved):

smmr-dn (approved):

PR #2297, opened 2024-10-10
Re-add nested portals

r100-stack:
## Changes

Reverts #2209 as the fix for https://github.com/iTwin/iTwinUI/pull/2205#discussion_r1731714448. Did this revert after internal discussion after we realized that a simpler alternative fix also works.

The simpler alternative to #2209 is to move focus to the DatePicker's today's date (second popover) only _after_ it has been portaled to be a descendant of `DateRangeFilter`. Since floating-ui's `FloatingFocusManager` uses DOM events instead of React events ([floating-ui/floating-ui#3060](https://redirect.github.com/floating-ui/floating-ui/issues/3060)), waiting for the second popover to be portaled within the first is important to prevent a focus out being triggered on the first popover.

Additionally, I was unable to figure out how #2209 fixed the bug. Because after #2209, the second popover was moved _outside_ the DOM of the first popover. So, focusing the second popover should be treated as a _focus out_ of the first popover prompting the first popover to close. But it didn't. Since we're unable to figure out why it happens, reverting is safer as it brings a more reliable and predictable state.

## Testing

* Fix failing unit tests
* Approved failing Table test images. Some of them may be reverts of the test images approved in [#2209](https://github.com/iTwin/iTwinUI/pull/2209).
  * Even though the new images show that focus does not move to the correct element in the floating content (even with `cy.wait()`), I manually confirmed that focus moves correctly to the expected element in the floating content.
  * Additionally, the images with the incorrect focus were approved before [#2209](https://github.com/iTwin/iTwinUI/pull/2209). So, it's okay to re-approve these images that don't have the focus on the correct element?

## Docs

Not sure if a changeset is needed since I don't think it's a significant user facing change.

## After PR TODO:

* [ ] Could consider a general "overridable initial focus" mechanism. Would be useful in Dialog too. ([#2297](https://github.com/iTwin/iTwinUI/pull/2297#discussion_r1801172533))

### DIFF
--- a/testing/e2e/app/routes/Table/spec.ts
+++ b/testing/e2e/app/routes/Table/spec.ts
@@ -448,13 +448,17 @@ test.describe('Table Paginator', () => {
   test(`should render data in pages`, async ({ page }) => {
     await page.goto(`/Table?exampleType=withTablePaginator`);
 
+    const paginatorButtons = page.locator('#paginator button', {
+      hasText: /[0-9]+/,
+    });
+
     await expect(page.locator(`[role="cell"]`).first()).toHaveText('Name 0');
     await expect(page.locator(`[role="cell"]`).last()).toHaveText(
       'Description 49',
     );
 
     // Go to the 6th page
-    await page.locator('button').last().click();
+    await paginatorButtons.nth(5).click();
 
     await expect(page.locator(`[role="cell"]`).first()).toHaveText('Name 250');
     await expect(page.locator(`[role="cell"]`).last()).toHaveText(
@@ -465,21 +469,22 @@ test.describe('Table Paginator', () => {
   test('should render truncated pages list', async ({ page }) => {
     await page.goto(`/Table?exampleType=withTablePaginator`);
 
-    await setContainerSize(page, '800px');
-
-    // Go to the 6th page
-    await page.locator('button').last().click();
-
     const paginatorButtons = page.locator('#paginator button', {
       hasText: /[0-9]+/,
     });
+
+    await setContainerSize(page, '800px');
+
+    // Go to the 5th page
+    await paginatorButtons.nth(4).click();
+
     await expect(paginatorButtons).toHaveText([
       '1',
+      '3',
       '4',
       '5',
       '6',
       '7',
-      '8',
       '11',
     ]);
     await expect(paginatorButtons.nth(3)).toHaveAttribute(


packages/itwinui-react/src/core/DatePicker/DatePicker.tsx
```diff
@@ -699,11 +699,18 @@ export const DatePicker = React.forwardRef((props, forwardedRef) => {
...
-                        element?.focus()
-                      }
+                      ref={(element) => {
+                        if (
+                          isSameDay(weekDay, focusedDay) &&
+                          needFocus.current
+                        ) {
+                          // Wait for DateRangeFilter's portaling to finish before focusing
+                          // See: https://github.com/iTwin/iTwinUI/pull/2297
+                          setTimeout(() => {
```
mayank99, line 709: Why not `queueMicrotask`?

Also I think the `if` check should be inside the microtask/timeout.

mayank99: > This is because I had noticed some failing tests with `queueMicrotask()`. I think this is because it conflicts with `FloatingFocusManager`'s focusing.

We might need a way to tell `FloatingFocusManager` to not do anything if we are already handling it.

Also, if only unit tests were failing, it could be because microtasks don't run reliably in unit tests. I had to add some hacky code when working on shadow DOM related tasks (see `setupTests`). You could try to convert these tests into e2e tests if it's giving you too much trouble.

> But actually after making these screen recordings, I noticed that the focus with `setTimeout()` starts from the trigger, goes to the previous month icon button, and then to the current date. I believe the focus should go from the trigger directly to the current date to prevent focus jumps and multiple callouts? Investigating further.

Yes, this was my concern with `setTimeout`. It might execute "too late" leading to focus jumping multiple times.

mayank99: > I went with the context approach to do this ([227827f](https://github.com/iTwin/iTwinUI/commit/227827f9668c1dd5727b3e265eec1d2a07faa593)). `Popover` adds a context containing the initialFocus around the floating content. The floating content can access this context and disable the initialFocus.

1. Whenever using context, make sure to memoize the `value` to prevent an unnecessary chain of re-renders.
2. Have you tested that this does not break the normal case where the first interactive element inside the popover is focused? (e.g. #2168)
3. (Maybe not relevant for this PR) We need to start thinking about a more general "overridable initial focus" solution that also works for Dialog.

mayank99: > Just wondering, if I directly put `setInitialFocus` instead of `{ setInitialFocus }` as the value in the context provider, would it still need to be memoized?

No, because `setInitialFocus` is already a stable reference. Whereas wrapping it in `{}` creates a new, unstable object during render (`{} === {}` is `false`).

smmr-dn (approved):
LGTM!

jason-crow:
@mayank99 when can we expect this PR to merge in and a patch published including it? This is needed to resolve a blocker for our pending release and it seems like the PR is approved, so I'm hoping this can complete and publish today?

PR #2301, opened 2024-10-14
Fix `TablePaginator` arrow keys page navigation

r100-stack:
## Changes

Fixed the arrow keys navigation in `TablePaginator` that wasn't working ever since `3.0.0`. The reason it wasn't working was because we didn't take into account the new `"_iui3-"` class prefix.

## Testing

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/b45a9b31-8817-4973-845a-b362c8cd9565" />|<video src="https://github.com/user-attachments/assets/946cd74f-02be-498a-986a-7836a34085f2" />|

## Docs

Added changeset

## After PR TODOs:

* [ ] To prevent errors like these, consider reworking the `iui-` class prefix logic in the future ([#2301 (review)](https://github.com/iTwin/iTwinUI/pull/2301#pullrequestreview-2369337431)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "typescript": "~5.5.2",
     "@types/react": "18.2.14",
     "@types/react-dom": "18.2.18",
-    "astro": "~4.15.8",
+    "astro": "~4.16.3",
     "cypress": "13.8.1",
     "ws@8": "^8.17.1",
     "cookie@<0.7.0": ">=0.7.0"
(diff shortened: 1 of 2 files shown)


mayank99 (approved):
Yikes. This is highlighting why we need to avoid using `iui-` classes. Let's make a note to rework this logic in the future.

smmr-dn (approved):

PR #2303, opened 2024-10-14
`TablePaginator` only shows ellipses when needed

r100-stack:
## Changes

Shows ellipses only when there is a gap of one or more page in between the extremities and the middle pages.

## Testing

* Added e2e test
* Manually confirmed that the ellipses are shown only when there is at least one page in between:

|Before|After|Explanation|
|-|-|-|
|image|image|`"…"` between 1 & 2|
|image|image|`"…"` between 100 & 101|

## Docs

Added changeset

## After PR TODOs:

* [ ] Try to replace all `&&` with ternary expressions. Also update the style guide and try adding a lint rule. ([#2303 (comment)](https://github.com/iTwin/iTwinUI/pull/2303#discussion_r1803542107)).

### DIFF
--- a/minimal-sandbox/package.json
+++ b/minimal-sandbox/package.json
@@ -7,6 +7,7 @@
   "main": "src/index.tsx",
   "dependencies": {
     "@itwin/itwinui-react": "3",
+    "@itwin/itwinui-icons-react": "2",
     "react": "18",
     "react-dom": "18"
   },


packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -326,18 +330,19 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
                 return pageButton(focusedIndex);
               }
+
               return (
                 <>
                   {startPage !== 0 && (
                     <>
                       {pageButton(0, 0)}
-                      {ellipsis}
+                      {showStartEllipsis && ellipsis}
```
mayank99, line 339:
```suggestion
                      {showStartEllipsis ? ellipsis : null}
```

packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -326,18 +330,19 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
                     </>
                   )}
                   {pageList.slice(startPage, endPage)}
                   {endPage !== totalPagesCount && !isLoading && (
                     <>
-                      {ellipsis}
+                      {showEndEllipsis && ellipsis}
```
mayank99, line 345:
```suggestion
                      {showEndEllipsis ? ellipsis : null}
```

mayank99 (approved):

smmr-dn (approved):

packages/itwinui-react/src/core/Table/TablePaginator.tsx
mayank99: Yes, we want to avoid rendering `undefined` or `false`, even if it works in the newer React versions. Using the ternary expression is "more correct". See also: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

We can update all instances in a separate PR, and maybe update our style guide. A lint rule would also be nice to help avoid these issues before committing.

PR #2305, opened 2024-10-15
Add `itwinui-icons-react` dep to `minimal-sandbox`

r100-stack:
## Changes

Since the [recommended](https://github.com/iTwin/iTwinUI/blob/2029579ef09478925668bb5ed3d8d5e5a6d36638/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1#L27) iTwinUI starter [sandbox](https://stackblitz.com/github/iTwin/iTwinUI/tree/main/minimal-sandbox) is based on the `minimal-sandbox` workspace and since `@itwin/itwinui-icons-react` package is commonly needed in sandboxes, this PR adds a dependency to `@itwin/itwinui-icons-react` to `minimal-sandbox`.

When testing in the new resulting stackblitz [sandbox](https://stackblitz.com/github/iTwin/iTwinUI/tree/r/add-icons-dep-to-sandbox/minimal-sandbox), I noticed that for some reason, the auto import for `@itwin/itwinui-icons-react` was not working unless an import was already present.

~So, I added an import and usage of it.~ Reverted those changes to keep `App.tsx` simple (https://github.com/iTwin/iTwinUI/pull/2305#discussion_r1802052891). Thus, the auto-import does not work.

image

But at least this PR saves time by not having to specify the dependency in the sandbox and then restart the sandbox.

## Testing

Testing that the dependency to `@itwin/itwinui-icons-react` exists in the new resulting stackblitz [sandbox](https://stackblitz.com/github/iTwin/iTwinUI/tree/r/add-icons-dep-to-sandbox/minimal-sandbox).

<details><summary>Old testing</summary>
<p>

Tested that the auto import for `@itwin/itwinui-icons-react` works in the new resulting stackblitz [sandbox](https://stackblitz.com/github/iTwin/iTwinUI/tree/r/add-icons-dep-to-sandbox/minimal-sandbox).

image

</p>
</details> 

## Docs

N/A

## After PR TODOs

- [ ] Try fixing auto-import issue in stackblitz sandbox ([#2305 (comment)](https://github.com/iTwin/iTwinUI/pull/2305#discussion_r1802072551))

### DIFF
--- a/minimal-sandbox/src/App.tsx
+++ b/minimal-sandbox/src/App.tsx
@@ -1,10 +1,10 @@
+import * as React from 'react';
 import { Button } from '@itwin/itwinui-react';
-import { SvgPlaceholder } from '@itwin/itwinui-icons-react';
 
 export default function App() {
   return (
     <>
-      <Button startIcon={<SvgPlaceholder />}>Hello world</Button>
+      <Button>Hello world</Button>
     </>
   );
 }


minimal-sandbox/src/App.tsx
mayank99, line 1: Let's revert this file, the component should be kept simple and minimal. And the `React` import helps with using hooks.

Originally, the `@itwin/itwinui-icons-react` package was intentionally excluded, to keep the number of deps low. Adding `@itwin/itwinui-icons-react` makes sense, to reduce the number of steps required for using icons, but that doesn't mean we need to use it in the example.

mayank99: The auto-import is a different issue which we should fix in a future PR. We might need to update our tsconfig.

mayank99 (approved):

smmr-dn (approved):

Issue #2300, opened 2024-10-11
Version 3.15.0 has broken table with custom expandable content


veekeys:
### Describe the bug (current behavior)

I have spent some time to even be able to reproduce this on isolated environment, but after updating to version 3.15.0 (and above), our Table has broken down.
Mainly we faced 2 issues:
- We got max depth reached error (which we managed to fix by getting rid of `getRowId` prop on Table, but that is a workaround..)
- After first fix, we were seeing expandable content on main rows rather than the main data rows..

In original code, we are using react-query to fetch some data asynchronously, where we do expect table to recalculate some stuff..
For reproduction, I have added just simple state which is updated after some time just to show what is enough to break the UI.

I have seen some changes done on expandable content code, so pretty sure some regressions was delivered there.

### Expected Behavior

Working same as with 3.14.x

### Link to minimal repro

https://stackblitz.com/edit/github-6vfacf-d3rg1m?file=src%2FApp.tsx

### Package Version

- [x] I'm using the latest version of iTwinUI.

smmr-dn:
This now has been fixed in [3.15.5](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.15.5)

PR #2306, opened 2024-10-16
Fixed expandable content bug when asynchronously re-rendering `Table`

smmr-dn:
## Changes

As reported in this issue (https://github.com/iTwin/iTwinUI/issues/2300), the expandable contents were broken due to the `WeakSet` data structure internally that we used in this [PR](https://github.com/iTwin/iTwinUI/pull/2208) to fix the issue regarding virtualization. However, due to the problem of stale data within the `WeakSet` between re-renders, another approach is used in this PR, where we added a `Symbol` property to mark each sub-row generated so that our `getSubRowsWithSubComponents` will not get caught in an infinite loop.

Additionally, as the user passed in the `getRowId`, this function got into a loop due to the duplicated IDs of both main rows and sub-rows. Therefore, we added a wrapper function to append the ID passed by users into the row's according ID conditionally generated based on the information of its parent. 

## Testing

Before this PR, when data is asynchronous fetched, the sub-components show up in place of the main rows as `Table` is re-rendered.

before

After the fix, the sub-components now are correctly rendered.

image

https://github.com/user-attachments/assets/a4a0fe87-2a63-40a3-9a0b-ce3121182cca

## Docs

Added patch changeset.


packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +579,42 @@ export const Table = <
...
 
-  const [rowHasParent] = React.useState(() => new WeakSet<T>());
-
   const getSubRowsWithSubComponents = React.useCallback(
-    (originalRow: T) => {
-      if (!rowHasParent.has(originalRow)) {
-        rowHasParent.add(originalRow);
-        return [originalRow];
+    (originalRow: T, relativeIndex: number) => {
+      const symbol = Symbol.for('iui-id');
```
mayank99, line 584: I think we want to use a private a symbol rather than a global one. To create a private symbol, use the [`Symbol()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/Symbol) constructor, rather than [`Symbol.for()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for), and save a reference to it.

So at the top of the file, you can declare this:

```js
const iuiId = Symbol('iui-id');
```

and use it everywhere in the file, e.g. `originalRow[iuiId]`

packages/itwinui-react/src/core/Table/Table.tsx
mayank99, line 1: Can we add a test for the scenario reported in #2300? It doesn't have to be the full repro code, just a simplified version of it should be enough.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +579,42 @@ export const Table = <
...
+      // Check if the current row is a sub-row that holds data for the according sub-component.
+      if (originalRow[symbol as keyof typeof originalRow]) {
+        return [];
       }
 
-      return (originalRow.subRows as T[]) || [];
+      // If rows have sub-components (generated as sub-rows), we need to return that sub-rows. Otherwise, row selection will get into a loop as it needs the row ID to set the sub-component to be selected.
+      // If the main row has not had any sub-rows, we need an object that holds a Symbol as a form of flag to prevent infinite loop.
+      return (
+        (originalRow.subRows as T[]) ?? [
```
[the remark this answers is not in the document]
smmr-dn: The reason why I check if `subRows` are present is because when the row is selected, we already have all sub-rows generated. The react-table `useRowSelect` hook keeps getting through all sub-rows via IDs using this `getSubRowsWithSubComponents` function to set main row and all nested sub-rows selected (https://github.com/TanStack/table/blob/v7/src/plugin-hooks/useRowSelect.js#L166C8-L167C71). However, since we 
 return the new object with the spread `originalRow`, it kept accessing the same `row.id` over again. Therefore, I had to return the sub-rows (our sub-component data) so it could set the sub-component row selected and get out of the loop.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +579,42 @@ export const Table = <
...
-
   const getSubRowsWithSubComponents = React.useCallback(
-    (originalRow: T) => {
-      if (!rowHasParent.has(originalRow)) {
-        rowHasParent.add(originalRow);
-        return [originalRow];
+    (originalRow: T, relativeIndex: number) => {
+      const symbol = Symbol.for('iui-id');
+
+      // Check if the current row is a sub-row that holds data for the according sub-component.
```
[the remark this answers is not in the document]

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +579,42 @@ export const Table = <
...
+      if (originalRow[symbol as keyof typeof originalRow]) {
+        return [];
       }
 
-      return (originalRow.subRows as T[]) || [];
+      // If rows have sub-components (generated as sub-rows), we need to return that sub-rows. Otherwise, row selection will get into a loop as it needs the row ID to set the sub-component to be selected.
+      // If the main row has not had any sub-rows, we need an object that holds a Symbol as a form of flag to prevent infinite loop.
```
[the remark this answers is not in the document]

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +579,42 @@ export const Table = <
...
+            [symbol]: `subcomponent-${relativeIndex}`,
+            ...originalRow,
+          },
+        ]
+      );
+    },
+    [],
+  );
+
+  const getRowIdWithSubComponents = React.useCallback(
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -857,32 +883,10 @@ export const Table = <
     ) => {
       const row = page[index];
       prepareRow(row);
+      const isRowWithSubComponent =
```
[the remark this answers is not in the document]

packages/itwinui-react/src/core/Table/Table.tsx

mayank99: It needs to be outside the component, otherwise a different symbol will be re-created on every render.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -857,32 +886,10 @@ export const Table = <
...
-            expanderCell={expanderCell}
-            scrollContainerRef={tableRef.current}
-            tableRowRef={enableVirtualization ? undefined : tableRowRef(row)}
-            density={density}
-            virtualItem={virtualItem}
-            virtualizer={virtualizer}
-          />
-        );
-      } else {
+      if (isRowASubComponent && !!subComponent) {
```
mayank99, line 892: This might be a nit but i think the `!!subComponent` check can be moved up with `isRowASubComponent`.

```js
      const isRowASubComponent =
        !!row.original[iuiId as keyof typeof row.original] && !!subComponent;

      if (isRowASubComponent) {
```

mayank99: Also, I don't think we need to set `as keyof typeof row.original`. It's basically lying to TS.

It would be better to just use `as any`.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,44 @@ export const Table = <
...
-      return (originalRow.subRows as T[]) || [];
+      // If originalRow represents the top-most level row, add itself as a subrow that will represent a subComponent.
+      // Add a symbol as a key on this subrow to distinguish it from the real row.
+      // This distinguishment is needed as the subrow needs to have all fields of the row since react-table expects even subrows to be rows (Row<T>).
+      return (
+        (originalRow.subRows as T[]) ?? [
+          {
+            [iuiId]: `subcomponent-${relativeIndex}`,
+            ...originalRow,
+          },
```
mayank99, line 598: Spreading can be expensive/slow. I think we could do this instead:

```js
if (originalRow.subRows) {
  return originalRow.subRows;
}

originalRow[iuiId] = `subcomponent-${relativeIndex}`;
return originalRow;
```

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: I have moved it outside.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: image

Hmm, I got this error while trying to assign the value. Is there any other way to approach this?

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: Noted this. I will work on adding the test.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: This is just a TS issue, which we can ignore (assuming the JS code still works correctly).

```js
(originalRow as any)[iuiId] = `subcomponent-${relativeIndex}`;
```

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: I have made the change.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: image

I run into this issue where it creates the right sub-rows in the first render but in the next one, since the original rows has the Symbol key and each symbol is unique, it keeps adding the symbols into the original data. When it loops back and hits the first if statement, it returns an empty sub-row array.

mayank99: That doesn't look right. Each symbol is _not_ unique; it's the same symbol.



Did you test this change *after* moving the symbol out of the component (based on the [other comment](https://github.com/iTwin/iTwinUI/pull/2306#discussion_r1806552426))?

smmr-dn: Right, I made this commit after moving the Symbol out commit.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: I have added the test.

packages/itwinui-react/src/core/Table/Table.tsx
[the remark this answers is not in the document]
smmr-dn: I have made the change.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: I looked into this and realized that my suggestion was flawed. We don't want to modify the original row, we want to return a copy of it, because the expanded-content is technically a new row.

So let's go with the original solution:

```js
if (originalRow.subRows) {
  return originalRow.subRows as T[];
}

return [
  {
    [iuiId]: `subcomponent-${relativeIndex}`,
    ...originalRow,
  },
];
```

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: I think you missed the [first comment](https://github.com/iTwin/iTwinUI/pull/2306#discussion_r1806575479). I'm still seeing `!!subComponent` in its original place.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: I have reverted back to the original solution.

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: The test doesn't seem representative of the bug. I'm not seeing a `getRowId` in the test.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: You're right, I missed the first comment. I have made the change.

mayank99: Now you've undone the second comment, where I suggested using `as any`.

The correct snippet should be:

```ts
      const isRowASubComponent = !!row.original[iuiId as any] && !!subComponent;

      if (isRowASubComponent) {
```

smmr-dn: I apologize for now reviewing this carefully. I have made the change.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,47 @@ export const Table = <
...
+    (originalRow: T, relativeIndex: number, parent?: Row<T>) => {
+      const mainRowId =
+        getRowId?.(originalRow, relativeIndex, parent) ?? `${relativeIndex}`;
+
+      // If the row contains the Symbol, it indicates that the current row is a sub-component row. We need to append the ID passed by user with its according sub-component ID.
+      return originalRow[iuiId as any]
+        ? `${mainRowId}-${
+            parent ? `${parent.id}.${relativeIndex}` : relativeIndex
+          }`
+        : mainRowId;
```
mayank99, line 621: Avoid nested ternaries. This can be made more readable by using an `if` condition.

```suggestion
      const mainRowId =
        getRowId?.(originalRow, relativeIndex, parent) ?? `${relativeIndex}`;

      // If the row contains the Symbol, it indicates that the current row is a sub-component row.
      // We need to append the ID passed by user with its according sub-component ID.
      if (originalRow[iuiId as any]) {
        const suffix = parent ? `${parent.id}.${relativeIndex}` : relativeIndex;
        return `${mainRowId}-${suffix}`;
      }

      return mainRowId;
```

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,47 @@ export const Table = <
...
 
-  const [rowHasParent] = React.useState(() => new WeakSet<T>());
-
   const getSubRowsWithSubComponents = React.useCallback(
-    (originalRow: T) => {
-      if (!rowHasParent.has(originalRow)) {
-        rowHasParent.add(originalRow);
-        return [originalRow];
+    (originalRow: T, relativeIndex: number) => {
+      console.log('row', originalRow);
```
mayank99, line 585: Accidentally left over

```suggestion
```

packages/itwinui-react/src/core/Table/Table.tsx
mayank99: Actually I'm not sure that this function is even how it should be. Looking at the [docs](https://react-table-v7-docs.netlify.app/docs/api/useTable#table-options), the default value is: `(row, relativeIndex, parent) => parent ? [parent.id, relativeIndex].join('.') : relativeIndex`, whereas this function falls back to just `relativeIndex`.

```suggestion
      const defaultRowId = parent
        ? [parent.id, relativeIndex].join('.')
        : relativeIndex;

      const mainRowId = getRowId?.(originalRow, relativeIndex, parent);

      // If the row contains the Symbol, it indicates that the current row is a sub-component row.
      // We need to append the ID passed by user with its according sub-component ID.
      if (mainRowId && originalRow[iuiId as any]) {
        return `${mainRowId}-${defaultRowId}`;
      }

      return mainRowId ?? defaultRowId;
```

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: Noted this and I will modify the test.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,47 @@ export const Table = <
...
+      // If originalRow represents a sub-component, don't add any sub-rows to it
+      if (originalRow[iuiId as any]) {
+        return [];
+      }
+
+      // If originalRow represents the top-most level row, add itself as a sub-row that will represent a subComponent.
+      // Add a symbol as a key on this sub-row to distinguish it from the real row.
+      // This distinction is needed as the sub-row needs to have all fields of the row since react-table expects even sub-rows to be rows (Row<T>).
```
mayank99, line 593: I think this comment is in the wrong place. It should be a few lines further down.

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: Not sure if I understand it correctly but this might cause an error due to race condition since `getRowId` takes precedence before the sub-rows are generated. It seems that it doesn't gets inside the if statement since the id passed by user is defined while the sub-rows are not yet generated, which ends up in a loop.
image
 
Additionally, since the original return type is `string`, and the current return type in this function is `string | number`, I think I also have to make `defaultRowId` a string so it won't error out.

Can I do something like this?

````jsx
const getRowIdWithSubComponents = React.useCallback(
    (originalRow: T, relativeIndex: number, parent?: Row<T>) => {
      const defaultRowId = `${
        parent ? [parent.id, relativeIndex].join('.') : relativeIndex
      }`;
      const mainRowId = getRowId?.(originalRow, relativeIndex, parent);
      // If the row contains the Symbol, it indicates that the current row is a sub-component row.
      // We need to append the ID passed by user with its according sub-component ID.
      if (originalRow[iuiId as any]) {
        return mainRowId ? `${mainRowId}-${defaultRowId}` : defaultRowId;
      }
      return mainRowId ?? defaultRowId;
    },
    [getRowId],
  );
````

Comment (mayank99) [reply packages/itwinui-react/src/core/Table/Table.tsx:621]: I'm not sure I understand your comment about race conditions. Doesn't it fall back to `defaultRowId` in all cases?

Comment (smmr-dn) [reply packages/itwinui-react/src/core/Table/Table.tsx:621]: I tried printing out `rowIdFromUser` and the original rows from `getSubRowsWithSubComponent` and saw that it kept returning `rowIdFromUser`, not the expected fallback value, before the sub-rows got to be created.
image

File: packages/itwinui-react/src/core/Table/Table.tsx (hunk 10 above)
Comment (smmr-dn) [reply packages/itwinui-react/src/core/Table/Table.tsx]: I have removed this.

File: packages/itwinui-react/src/core/Table/Table.tsx (hunk 11 above)
Comment (smmr-dn) [reply packages/itwinui-react/src/core/Table/Table.tsx]: I have moved this down and added another comment here.

File: packages/itwinui-react/src/core/Table/Table.tsx (hunk 9 above)
Comment (mayank99) [reply packages/itwinui-react/src/core/Table/Table.tsx]: I wonder if this is a problem with `0` being a falsy value. Based on your ternary condition, the equivalent fallback return value would need to use `||` instead of `??`.

<details><summary>Full code</summary>

```tsx
const defaultRowId = parent
  ? `${parent.id}.${relativeIndex}`
  : `${relativeIndex}`;
const rowIdFromUser = getRowId?.(originalRow, relativeIndex, parent);
// If the row contains the Symbol, it indicates that the current row is a sub-component row.
// We need to append the ID passed by user with its according sub-component ID.
if (rowIdFromUser && originalRow[iuiId as any]) {
  return `${rowIdFromUser}-${defaultRowId}`;
}
return rowIdFromUser || defaultRowId;
```

</details> 

This does raise the question: is `0` a valid value that we want to use, or do we want to discard it? If it's valid, then we should check using `??` and `!== undefined` instead of loose boolean checks.

smmr-dn: I used `rowIdFromUser !== undefined` for the if condition, put `??` back to the final return and it behaved correctly. Hmm, I think it should be kept since it's the index of the first row (?).

packages/itwinui-react/src/core/Table/Table.tsx
smmr-dn: I have added `getRowId` to the test.

mayank99 (approved):
LGTM 🚀

We can't make a patch release because #2299 was already merged, so this will have to wait until `3.16.0`.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,49 @@ export const Table = <
...
+          [iuiId]: `subcomponent-${relativeIndex}`,
+          ...originalRow,
+        },
+      ];
+    },
+    [],
+  );
+
+  /**
+   * Gives `subComponent` a unique id since `subComponent` has the same react-table id as its parent row. Avoiding duplicate react-table ids prevents react-table errors.
```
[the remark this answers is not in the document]
smmr-dn: I have split the sentence.

packages/itwinui-react/src/core/Table/Table.tsx
```diff
@@ -578,18 +580,49 @@ export const Table = <
...
+      }
+
+      // If sub-rows already exist, return them as they are.
+      if (originalRow.subRows) {
+        return originalRow.subRows as T[];
+      }
+
+      // If originalRow represents the top-most level row, add itself as a sub-row that will represent a subComponent.
+      // Add a symbol as a key on this sub-row to distinguish it from the real row.
+      // This distinction is needed as the sub-row needs to have all fields of the row since react-table expects even sub-rows to be rows (Row<T>).
```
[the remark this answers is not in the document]
smmr-dn: I have split the sentence.

smmr-dn:
Co-authored-by: Mayank <9084735+mayank99@users.noreply.github.com>
Co-authored-by: Vykis <81580355+veekeys@users.noreply.github.com>
Co-authored-by: Rohan <45748283+r100-stack@users.noreply.github.com>

PR #2256, opened 2024-09-20
Update deps to fix audit

r100-stack:
## Changes

Fixed the [following](https://github.com/iTwin/iTwinUI/actions/runs/10961807709/job/30439620264?pr=2233) audit issues:
* https://github.com/advisories/GHSA-f6v4-cf5j-vf3w
* https://github.com/advisories/GHSA-9wv6-86v2-598j
* https://github.com/advisories/GHSA-gp8f-8m3g-qvj9
* https://github.com/advisories/GHSA-m6fv-jmcg-4jfg
* https://github.com/advisories/GHSA-9cwx-2883-4wfx
* https://github.com/advisories/GHSA-64vr-g452-qvp3

The notable dep bumps:
* `vite`: `^5.1.7` → `~5.1.8`
* `next`: `^14.2.3` → `^14.2.13`
* `astro`: `~4.5.6` → `~4.15.8`

The newer astro versions recommends making some changes in the `env.d.ts` file ([changelog](https://github.com/withastro/astro/releases/tag/astro%404.14.0)). This PR makes those changes (by running `build` in `@itwin/itwinui-react`) and adds [`.astro`](https://docs.astro.build/en/guides/content-collections/#the-astro-directory) to the gitignore.

<details>
<summary>Changelog excerpt</summary>

> Although src/env.d.ts will continue to work as is, we recommend you update it:
> 
> ```diff
> - /// <reference types="astro/client" />
> + /// <reference path="../.astro/types.d.ts" />
> - /// <reference path="../.astro/env.d.ts" />
> - /// <reference path="../.astro/actions.d.ts" />
> ```

</details>

Since these dep bumps resulted in a lot of changes in the lock file, I also ran [`pnpm dedupe`](https://pnpm.io/cli/dedupe).

## Testing

* Audit now passes 🎉
* Confirmed that all workspaces open up correctly.
* Confirmed that `pnpm run build` and `pnpm run test` continue working correctly.
* CI passes

## Docs

Since the three main package bumps are related to dep workspaces, no changesets necessary.

## After PR TODOs:

* [ ] Update `vite` to `5.4.7` (or later) in all places ([#2256 (comment)](https://github.com/iTwin/iTwinUI/pull/2256#discussion_r1771870422)).

PR #2312, opened 2024-10-23
Bump vite to `5.4.14`

r100-stack:
## Changes

After PR TODO from #2256 (https://github.com/iTwin/iTwinUI/pull/2256#discussion_r1771870422).

Fixes https://github.com/advisories/GHSA-vg6x-rcgg-rjx6. Closed #2409 in favor of this PR.

---

Bumps `vite` from `~5.1` to `~5.4.14` (after deduping, only `5.4.14` is used atm).

After the bump, the `styles.css` was not getting generated in `styles.js/dist`. Looks like it is because of [vitejs/vite/#16051](https://redirect.github.com/vitejs/vite/pull/16051) from [5.2.0-beta.0](https://redirect.github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#520-beta0-2024-03-12).

To fix it, I added the vite plugin suggested in [vitejs/vite#16051 (comment)](https://redirect.github.com/vitejs/vite/pull/16051#issuecomment-2076616816). I had tried it when I had opened this PR too. However, it didn't work back then (which is why I had tried another fix which posed other problems; more info below). But since this vite plugin works now, I went ahead with this.

<details><summary>Old fix</summary>
<p>

To fix it, I renamed `styles.module.css` to `styles.css`. If there is a reason why the file has to be a `module.css` file, then we'd have to import _and use_ the styles somewhere, or something along these lines.

One problem with using `styles.css`, however, is that the `iui-` prefix is not there in the generated styles.css file because the css module part of the vite config file is not applied.

</p>
</details> 

Since I noticed multiple `vite` versions, especially some that were still in the affected range of the security advisory, I also deduped it to only use `5.4.14` (https://github.com/iTwin/iTwinUI/pull/2312/commits/b50db2a99f72dafc477087414ae8b28006271799). But this caused the css-workshop tests to all fail. To fix it, I changed the test script to replace `astro preview` with `serve` and added a [`wait-on`](https://www.npmjs.com/package/wait-on) step so that the tests start up only after `serve` has started up the server. ([more info](https://github.com/iTwin/iTwinUI/pull/2312#discussion_r1932451134)).

## Testing

* [x] CI still passing
* [x] The `build` script is working in the `itwinui-react` workspace
* [x] The generated `styles.css` in `packages/itwinui-react` is unchanged.

## Docs

No changeset as `vite` is used in dev workspaces or listed as `devDependencies` in the package workspaces (`itwinui-react`).

### DIFF
--- a/apps/css-workshop/package.json
+++ b/apps/css-workshop/package.json
@@ -21,7 +21,8 @@
     "clean": "rimraf dist && rimraf node_modules && rimraf .turbo",
     "approve": "backstop --config=./backstop/backstop.cjs approve",
     "test:only": "backstop --config=./backstop/backstop.cjs test --docker",
-    "test": "npm-run-all -p -r \"preview\" \"test:only {1}\" --"
+    "test:only:wait-for-preview": "npx -y wait-on http://localhost:3050 && pnpm run test:only",
+    "test": "npm-run-all -p -r \"preview\" \"test:only:wait-for-preview {1}\" --"
   },
   "prettier": "configs/prettier-astro-config.js"
 }


packages/itwinui-react/src/styles.js/vite.config.mjs
```diff
@@ -86,6 +87,21 @@ function postcssAddIuiVersion() {
...
+    enforce: 'post',
+    transform(_, id) {
+      if (id.endsWith('.module.css')) {
+        return {
+          moduleSideEffects: 'no-treeshake',
+        };
+      }
+    },
+  };
+}
```
mayank99, line 103: Nice find. My only suggestion would be to inline this plugin in the config above, so that we get better types.

mayank99 (approved):

mayank99:
> Update: I recreated the lock file without deduping `vite`.

Can you clarify which versions of `vite` we are using now? The PR title is misleading if we don't have a single `5.4.14` version.

pnpm-lock.yaml
mayank99, line 1: _Continuing https://github.com/iTwin/iTwinUI/pull/2312#issuecomment-2616903324_

Regardless of where it's coming from, I don't think we should have four different versions of `vite`, three of which fall in the range of affected versions for https://github.com/advisories/GHSA-vg6x-rcgg-rjx6?rgh-link-date=2024-10-23T22%3A27%3A59Z. If it weren't for the vulnerable version, I would feel less strongly about this.

Can we try to investigate why deduping is breaking things?

mayank99: Assuming that the issue is only in the preview command (i.e. the actual build is fine), we can work around it by replacing `astro preview` with something like `npx -y serve` (similar to how it's done in [`react-workshop`](https://github.com/iTwin/iTwinUI/blob/55cc09ed775de69c1ee787cc8a9cc7dac5b30512/apps/react-workshop/scripts/entrypoint.sh#L6))

mayank99: How frequent is this issue? Maybe we can live with it for now.

mayank99 (approved):
LGTM, just make sure to update PR description

I'm noticing CI is failing for react visual tests, but I think that's caused by #2391; maybe one of the Toast images needs to be updated to be less flaky.

smmr-dn (approved):

PR #2316, opened 2024-10-25
Prevent `TablePaginator` page buttons from being overlapped

r100-stack:
## Changes

After PR TODO from #2152.

The next button was not clickable after going to the next pages a few times because the right content was overlapping the page buttons. This overlap was happening because when the `scrollWidth` was increasing, the `useOverflow()` algorithm was not recalculating.

The best way would be to re-run the `useOverflow()` calculations when the `scrollWidth` changes. But that was not straightforward to do ([thread](https://github.com/iTwin/iTwinUI/pull/2316#discussion_r1819879252)).

So for now, I just went with giving the `iui-center` in the table paginator a higher z-index ([#2316 (comment)](https://github.com/iTwin/iTwinUI/pull/2316#discussion_r1821561229)). I also added an after-PR TODO to search for a better approach.

<details><summary>Old PR description subset that could be helpful when working on the better approach</summary>

I have been trying to find another way to trigger the `useOverflow()` re-calculations. E.g. moving the `key` inside OverflowContainerComponent from #2154:

https://github.com/iTwin/iTwinUI/blob/528cf5459760128d88bd3648b2217b396033526c/packages/itwinui-react/src/utils/components/OverflowContainer.tsx#L113-L115

I felt like doing that to keep similar code together and also so that other components that use OverflowContainer can also listen to their `scrollWidth` changes. However, I wasn't able to find a way that works well. Open to ideas for any improvements.

</details> 

## Testing

Could not think of a good way to test the overlapping. So, did not add any tests.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/1c119a38-e3f8-459e-afb3-7f01261fe790" />|<video src="https://github.com/user-attachments/assets/4996c37a-e3ce-435f-bd98-f539e3b17b9f" />|

## Docs

Added one react changeset and one css changeset.

## After PR TODOs

* [ ] Find a better solution than the one in this PR ([#2316 (comment)](https://github.com/iTwin/iTwinUI/pull/2316#discussion_r1821446820)).

### DIFF
--- a/.changeset/strange-rockets-taste.md
+++ b/.changeset/strange-rockets-taste.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed duplicate data returned from the `onSelect` prop passed into `Table`.
--- a/packages/itwinui-react/src/core/Table/Table.test.tsx
+++ b/packages/itwinui-react/src/core/Table/Table.test.tsx
@@ -1981,6 +1981,44 @@ it('should select and filter rows', async () => {
   expect(checkboxInputs[3].checked).toBe(false);
 });
 
+it('should only select main row when subComponent is present', async () => {
+  const logSpy = vitest.spyOn(console, 'log');
+  const columns = [
+    {
+      id: 'name',
+      Header: 'Name',
+      accessor: 'name',
+    },
+    {
+      id: 'description',
+      Header: 'description',
+      accessor: 'description',
+    },
+  ];
+
+  const data = mockedData();
+
+  const { container } = renderComponent({
+    columns,
+    data,
+    isSelectable: true,
+    onSelect: (selectedRows) => console.log(selectedRows),
+    subComponent: (row) => <div>{`Name ${row.original.name}`}</div>,
+  });
+
+  const checkboxCells = container.querySelectorAll('.iui-slot .iui-checkbox');
+  expect(checkboxCells.length).toBe(4);
+
+  await userEvent.click(checkboxCells[0]);
+  expect(logSpy).toHaveBeenCalledWith(data);
+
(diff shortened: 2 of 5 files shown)


packages/itwinui-react/src/core/Table/TablePaginator.tsx
```diff
@@ -252,7 +254,12 @@ export const TablePaginator = (props: TablePaginatorProps) => {
         )}
       </Box>
       {showPagesList && (
-        <OverflowContainer className='iui-center' itemsCount={totalPagesCount}>
+        <OverflowContainer
+          key={centerContainerRef.current?.scrollWidth}
```
mayank99, line 258: ref is being read during render.

Can we handle this some other way? Would `paginatorWidth` work as a key?

mayank99: Maybe you can replace `useContainerWidth` with a `useResizeObserver` that lets you save both the full width and the `scrollWidth` into state?

mayank99: I'm also suggesting removing `useContainerWidth` (which itself uses `useResizeObserver`). We shouldn't set up two resize observers, we can use `useResizeObserver` to save all the states that we need.

mayank99: So then this approach will not work. I'd say let's revisit this after #2154.

For now, can we use `z-index` as a workaround? The center portion can be given a higher `z-index` than the right portion, to ensure the buttons are interactable even when they overlap. It's not perfect, but better than current situation.

mayank99 (approved):

smmr-dn (approved):

PR #2326, opened 2024-11-04
Created documentation for `SkipToContentLink`

smmr-dn:
## Changes

Created documentation for the `SkipToContentLink` component.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/apps/website/src/content/docs/skiptocontentlink.mdx
+++ b/apps/website/src/content/docs/skiptocontentlink.mdx
@@ -1,0 +1,17 @@
+---
+title: Skip to content link
+description: A skip-to-content link can be used to navigate directly to certain desired contents.
+thumbnail: #TODO
+---
+
+<p>{frontmatter.description}</p>
+
+## Usage
+
+`SkipToContentLink` is for screen reader and keyboard users and will not be visible unless tabbed to. Provides a shortcut to the main content of the page without navigating through the header, etc.
+
+<LiveExample src='SkipToContentLink.main.jsx' truncate={true}>
+  <AllExamples.SkipToContentLinkMainExample client:visible />
+</LiveExample>
+
+## Props
--- a/apps/website/src/content/docs/skiptocontentlink.mdx
+++ b/apps/website/src/content/docs/skiptocontentlink.mdx
@@ -6,12 +6,26 @@ thumbnail: #TODO
 
 <p>{frontmatter.description}</p>
 
+<Placeholder componentPageName='table--basic' />
+
 ## Usage
 
-`SkipToContentLink` is for screen reader and keyboard users and will not be visible unless tabbed to. Provides a shortcut to the main content of the page without navigating through the header, etc.
+`SkipToContentLink` is designed for screen reader and keyboard users and will only become visible when focused. It provides a shortcut to the main content of the page, bypassing the header and other navigation elements. The `href` prop should be set to the `id` of the main content section, and the value must include the '#' prefix.
 
 <LiveExample src='SkipToContentLink.main.jsx' truncate={true}>
-  <AllExamples.SkipToContentLinkMainExample client:visible />
+  <AllExamples.SkipToContentLinkMainExample client:load />
+</LiveExample>
+
+**Note**: `SkipToContentLink` must be placed inside a `<body>`.
+
+### Custom Text
+
+The link component accepts custom text through its `children` prop, allowing for localization. If no custom text is provided, the default text displayed is "Skip to main content".
+
+<LiveExample src='SkipToContentLink.customText.jsx' truncate={true}>
+  <AllExamples.SkipToContentLinkCustomTextExample client:load />
 </LiveExample>
 
 ## Props
+
+<PropsTable path='@itwin/itwinui-react/esm/core/SkipToContentLink/SkipToContentLink.d.ts' />
--- a/examples/SkipToContentLink.main.jsx
+++ b/examples/SkipToContentLink.main.jsx
@@ -1,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { SkipToContentLink, Text, Kbd, Anchor } from '@itwin/itwinui-react';
+
+export default () => {
+  return (
+    <>
+      <SkipToContentLink href='http://localhost:1700/docs/skiptocontentlink#props' />
+      <Text as='p'>
+        This is not the main content. Press&nbsp;
+        <Kbd>tab</Kbd>&nbsp;to see skip-to-content-link component. You might
+        need to click on the top of the page or the URL first. Press&nbsp;
+        <Kbd>&crarr;</Kbd>
+        &nbsp;after focusing on skip-to-content-link to skip to the main content
+        below. <Anchor href='#'>This link will be skipped.</Anchor> <br />
+      </Text>
+      <div
+        style={{
+          border: 'solid 1px var(--iui-color-border)',
+          height: 300,
+          marginTop: 10,
+          padding: 12,
+        }}
+        tabIndex={0}
+        id='main-content'
+      >
+        <Text as='p'>
+          This is the main content. Focus will be directed here from the
+          skip-to-content-link component.&nbsp;
+          <Anchor href='#'>Tab again to focus on this link.</Anchor>
+          &nbsp;
+        </Text>
+      </div>
+    </>
+  );
+};
--- a/examples/SkipToContentLink.customText.jsx
+++ b/examples/SkipToContentLink.customText.jsx
@@ -7,7 +7,7 @@ import { SkipToContentLink, Text, Kbd, Anchor } from '@itwin/itwinui-react';
 export default () => {
   return (
     <>
-      <SkipToContentLink href='#main-content' />
+      <SkipToContentLink href='#props'>Go to main section</SkipToContentLink>
       <Text variant='body'>
         This is not the main content. Press&nbsp;
         <Kbd>tab</Kbd>&nbsp;to see skip-to-content-link component. You might
@@ -17,10 +17,19 @@ export default () => {
         below.{' '}
         <Anchor target='_blank' rel='noopener noreferrer' href='#'>
           This link will be skipped.
-        </Anchor>
+        </Anchor>{' '}
         <br />
       </Text>
-      <div className='main-content' id='main-content'>
+      <div
+        style={{
+          border: 'solid 1px var(--iui-color-border)',
+          height: 300,
+          marginTop: 10,
+          padding: 12,
+        }}
+        tabIndex={0}
+        id='main-content'
+      >
         <Text variant='body'>
           This is the main content. Focus will be directed here from the
           skip-to-content-link component.&nbsp;
--- a/examples/SkipToContentLink.main.css
+++ b/examples/SkipToContentLink.main.css
@@ -1,0 +1,6 @@
+.main-content {
+  border: solid 1px var(--iui-color-border);
+  height: 300px;
+  margin-top: 10px;
+  padding: 12px;
+}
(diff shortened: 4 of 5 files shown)


smmr-dn (commented):
The examples are currently not working since the page uses the "#" identifier to move focus to its section (e.g. Usage, Props, ...) but not the actual `div` that contains the main content's id, which causes the page not focusing on the main content of each example. I notice the v2 storybook opens the links in new tab so maybe we can do something similar for this doc? Other than that, I'm not sure how I should handle the examples.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+<Placeholder componentPageName='table--basic' />
+
+## Usage
+
+`SkipToContentLink` is designed for screen reader and keyboard users and will only become visible when focused. It provides a shortcut to the main content of the page, bypassing the header and other navigation elements. The `href` prop should be set to the `id` of the main content section, and the value must include the '#' prefix.
+
+<LiveExample src='SkipToContentLink.main.jsx' truncate={true}>
+  <AllExamples.SkipToContentLinkMainExample client:load />
+</LiveExample>
```
mayank99, line 17: If the link is not working anyway, can we simplify the examples? The main purpose of the demo would be to reveal the link upon tabbing.

The instructions can be placed outside the demo if needed.

Also, I think we can make use of static code blocks to show what the correct usage might look like within the context of an application.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,31 @@
...
+thumbnail: #TODO
+---
+
+<p>{frontmatter.description}</p>
+
+<Placeholder componentPageName='table--basic' />
+
+## Usage
+
+`SkipToContentLink` is designed for screen reader and keyboard users and will only become visible when focused. It provides a shortcut to the main content of the page, bypassing the header and other navigation elements. The `href` prop should be set to the `id` of the main content section, and the value must include the '#' prefix.
```
mayank99, line 13: The first two sentences here do not belong under "Usage", they should be placed in the intro section.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,31 @@
+---
+title: Skip to content link
+description: A skip-to-content link can be used to navigate directly to certain desired contents.
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,31 @@
+---
+title: Skip to content link
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,31 @@
...
+description: A skip-to-content link can be used to navigate directly to certain desired contents.
+thumbnail: #TODO
+---
+
+<p>{frontmatter.description}</p>
+
+<Placeholder componentPageName='table--basic' />
```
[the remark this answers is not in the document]
smmr-dn: I have removed the placeholder link.

apps/website/src/content/docs/skiptocontentlink.mdx
smmr-dn: Hmm, I added a code block under the description and modified the examples. Hmmm, those are just my idea for the examples, I'm not sure if they look good.

apps/website/src/content/docs/skiptocontentlink.mdx
smmr-dn: I have moved these in the intro section.

apps/website/src/content/docs/skiptocontentlink.mdx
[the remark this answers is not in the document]
mayank99: Have you?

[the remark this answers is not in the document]
smmr-dn: That's odd, I updated this along with the title... I have made the change.

apps/website/src/content/docs/skiptocontentlink.mdx
mayank99: It doesn't feel quite right; the wording, instructions and code samples do not accurately convey what a skip-to-content link is and how to use it.

I'll make some concrete suggestions tomorrow, but in the meantime I would suggest reading up on skip links in general and within the context of other design systems. That will help you understand what it is and how it should be documented.

- https://webaim.org/techniques/skipnav/
- https://benmyers.dev/blog/skip-links/
- https://css-tricks.com/a-deep-dive-on-skipping-to-content/
- https://design-system.service.gov.uk/components/skip-link/
- https://www.skyscanner.design/latest/components/skip-link/web-xYpbV1AB
- https://ebay.gitbook.io/mindpatterns/navigation/skip-navigation

smmr-dn: I have tried moving things around and changed the examples in the recent commits.

mayank99: It looks much better now! I'll leave some suggestions separately.

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+<p>{frontmatter.description}</p>
+
+In some applications, numerous navigation links and other focusable elements can require a lot of keyboard presses before users reach their desired content. What is more, if users tab into a link that leads to another page with the same structure, they must tab through the same list of non-essential links again.
+
+While this may not be an issue for sighted or mouse users, it can be cumbersome for those who rely on keyboard navigation. Therefore, a skip link is crucial to improve accessibility and streamline navigation, designed for screen-reader and keyboard users.
+
+```jsx
+<SkipToContentLink href='#main-content-id' />
+```
```
mayank99, line 15: Do we need this code snippet here? Feels like it should be in the usage section.

```suggestion
```

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,53 @@
...
+- The link pointing to that target element.
+
+The `href` prop can be used to direct focus to the target element. This prop must be set to the `id` of the main content section, including the `"#"` prefix.
+
+```jsx {5}
+<div className='App'>
+  <SkipToContentLink href='#main-content-id' />
+  {/* Other elements such as header, navigation, etc. */}
+
+  <div id='main-content-id'>{/* Some main content that will be skipped to */}</div>
```
mayank99, line 31: We could show `<main>` to encourage the use of landmarks.

```suggestion
  <main id='main-content-id'>{/* Main content that will be skipped to */}</main>
```

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,53 @@
...
+## Usage
+
+The `SkipToContentLink` component setup requires two elements:
+
+- The target element to skip to.
+- The link pointing to that target element.
+
+The `href` prop can be used to direct focus to the target element. This prop must be set to the `id` of the main content section, including the `"#"` prefix.
+
+```jsx {5}
```
mayank99, line 26: I think line 2 is more important.

````suggestion
```jsx {2}
````

File: apps/website/src/content/docs/skiptocontentlink.mdx [hunk 9]
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Usage
+
+The `SkipToContentLink` component setup requires two elements:
+
+- The target element to skip to.
+- The link pointing to that target element.
+
+The `href` prop can be used to direct focus to the target element. This prop must be set to the `id` of the main content section, including the `"#"` prefix.
```
mayank99, line 24:
```suggestion
The `href` prop should be used to direct focus to the target element. This prop must be set to the `id` of the main content section, including the `"#"` prefix.
```

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,53 @@
...
+```jsx {5}
+<div className='App'>
+  <SkipToContentLink href='#main-content-id' />
+  {/* Other elements such as header, navigation, etc. */}
+
+  <div id='main-content-id'>{/* Some main content that will be skipped to */}</div>
+</div>
+```
+
+As the skip link is visually hidden from view until focused, users typically access the link by pressing `Tab` at the start of the page. The skip link component should appear on top of the page to make sure it is one of the first items the screen readers get to and that keyboard users are able to navigate to using the keyboard.
```
mayank99, line 35:
```suggestion
As the skip link is visually hidden from view until focused, users typically access the link by pressing <kbd>Tab</kbd> at the start of the page. The skip link component should appear on top of the page to make sure it is one of the first items that keyboard users are able to navigate to using the keyboard.
```

apps/website/src/content/docs/skiptocontentlink.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+<LiveExample src='SkipToContentLink.main.jsx' truncate={true}>
+  <AllExamples.SkipToContentLinkMainExample client:load />
+</LiveExample>
+
+### Custom Text
+
+As the link is designed for screen reader users, localization is crucial for making the application accessible and understandable. The skip link component accepts custom text through its `children` prop, allowing end users to provide a localized message.
+
+If no custom text is provided, the default text displayed is "Skip to main content". This flexibility helps create a more inclusive and user-friendly experience for a global audience.
```
mayank99, line 45: Can be made less wordy.

```suggestion
The skip link component accepts custom text through its `children` prop, which is especially useful for providing a localized message.

If no `children` is passed, the default text displayed is "Skip to main content".
```

examples/SkipToContentLink.main.jsx
mayank99, line 1: Could we improve the demos so that the Skip link shows up inside the demo rectangle? Since this component uses `position: fixed`, you might need to create a containing block in CSS (iirc we do this in Dialog if you want to see an example).

mayank99 (approved):
Pre-approving with some suggestions.

examples/SkipToContentLink.main.jsx
smmr-dn: I have updated the PR with all suggested wording changes above. As for the demos, I set `position: relative` to the containers and `position: absolute` for the skip links so they will be positioned accordingly to the parent containers. I feel a bit iffy about overriding `position: fixed`.

mayank99: We do not need `position: relative` and `position: absolute`. Here's the code where we are creating a containing block in `dialog.scss`:

https://github.com/iTwin/iTwinUI/blob/9fc891e24ef62ed2ccdbc4ded31e9057576f74d1/packages/itwinui-css/src/dialog/dialog.scss#L16

We can do the same in the demo CSS. A code comment would also be useful to indicate that this CSS is only being added for demonstration purposes.

---

I noticed also that the demo doesn't work as per the instructions. If you click the box and press <kbd>Tab</kbd>, the skip link doesn't appear. We might want to change the demo to be more like this:

```jsx
    <div className='demo-container'>
      <div>
        <button>Click here</button> then press <kbd>Tab</kbd> to see the skip
        link.
      </div>
      {/* The SkipToContentLink component should normally be placed at the
          beginning of the page and point to the main content. This is a
          contrived example for visual demonstration. */}
      <SkipToContentLink href='#' />
    </div>
```

Further, I think we should:
1. Show this live demo at the beginning of the page.
2. Hide the code for the demo (can be done in a separate PR, as it will require changes to the `LiveExample` component).
3. Replace the live demo for "custom content" with a static code example.

smmr-dn: Oh I was looking at the CSS in the Dialog doc. I have moved the live demo and added the after-PR todo item to hide the code for the demo. I have also used static code example in place of the live demo.

mayank99: Two nits:
1. We do not need use `Button` and `Kbd` from iTwinUI. Native `<button>` and `<kbd>` would be better, to draw increased focus on `SkipToContentLink`.
2. A short comment on the CSS to create containing block would be useful.

smmr-dn: I have made the change.

PR #2336, opened 2024-11-12
Add initial low-fidelity threat model

ben-polinsky:
Add threagile threat model for iTwinUI Documentation. Since these docs are released through an ADO classic pipeline, work there will follow once this is merged.


ben-polinsky:
> Tried to review it as best as I could and LGTM.

Appreciate it! Pretty simple case for a static website. If things change, it's here to update.

PR #2340, opened 2024-11-14
Fix regression in components with overflow handling

r100-stack:
## Changes

@mayank99 noticed a regression after #2154 when StrictMode is enabled and testing on the local version. The hosted versions worked correctly. Upon investigation, I noticed that the issue started from https://github.com/iTwin/iTwinUI/pull/2154/commits/d667eafb495e4120bf3d2ac7404f30b421b8b402 from that PR.

Because it doesn't work only when StrictMode is enabled, I believe the issue has something to do with an impure reducer function. However, I was unable to find any impurity. I also tried changing the useLayoutEffect temporarily to a useEffect since I thought that might cause some stale values. However, that too didn't help.

To fix the error as of now, I just pass the state in each dispatch. However, open to better ideas since I feel this is hacky.

## Testing

Confirmed that the errors seen in the overflow related components/stories are no longer there. E.g. `ButtonGroup → Overflow` story:

|Before|After|
|-|-|
|image|image|

## Docs

Didn't add a changeset since #2154 is not released.

## After PR TODO

- [ ] Investigate why state argument in reducer function sometimes gives incorrect values in StrictMode. Then try removing the `action.currentState` workaround. ([#2340 (comment)](https://github.com/iTwin/iTwinUI/pull/2340#discussion_r1844103492)).
  - Also maybe investigate other TODO at the same time: [#2340 (comment)](https://github.com/iTwin/iTwinUI/pull/2340#discussion_r1844133265)

### DIFF
--- a/apps/website/README.md
+++ b/apps/website/README.md
@@ -1,1 +1,11 @@
 # iTwinUI docs
+
+[iTwinUI documentation website](https://itwinui.bentley.com).
+
+## Threat Model
+
+This repo uses Threagile to generate a threat model. To ensure that we keep our threat model up-to-date (both in terms of the architecture of our service and any potential threats) we will review the threat model under two different circumstances. Each review entails making sure the threat model continuously aligns with our service, captures known threats, and reflects the current status of any risk-tracking items.
+
+During every PR, care should be taken to identify any potential architectural changes (e.g., a new endpoint/service dependency, significant new features, etc.) that would necessitate changes to the system model in threagile.yaml. It is both the developer's and reviewer's responsibility to ask, "what could go wrong?" resulting from direct changes or changes to prior security assumptions. Any new manual threats identified must be captured and tracked in threagile.yaml. Developers should use each PR as an opportunity to preemptively update threagile.yaml to expedite this process. No PRs will be approved or merged without including any required threagile.yaml changes.
+
+A quarterly review of threagile.yaml. Each quarterly review will be tracked via a GitHub issue in this repo with the tag 'Threagile Review', and should be approved by the security champion and dev lead by typing 'APPROVED' in the issue once the review is complete. Threat model review consists of validating system model accuracy, identifying any new manual threats, and verifying up-to-date statuses on all risk tracking items.


packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -251,28 +251,36 @@ const useOverflow = (
...
         // not overflowing = maybe we guessed too low? So, increase min guess
-        dispatch({ type: 'increaseMinGuess' });
+        dispatch({ type: 'increaseMinGuess', state: guessState });
       }
     } finally {
       isGuessing.current = false;
     }
-  }, [isStabilized, itemsCount, maxGuess, minGuess, orientation, visibleCount]);
+  }, [
+    guessState,
```
mayank99, line 276: I'll repeat what I said internally, which is that we should move `guessVisibleCount` inside the Effect, to make the execution flow more direct.

Additionally, we can make the dependency array smaller, by destructuring things from `guessState` inside the Effect.

packages/itwinui-react/src/utils/components/OverflowContainer.tsx
```diff
@@ -370,51 +382,51 @@ const overflowGuessReducer = (
...
-      if (state.isStabilized) {
-        return state;
+      if (action.state.isStabilized) {
+        return action.state;
       }
 
-      let newMinGuess = state.minGuess;
-      let newMaxGuess = state.maxGuess;
+      let newMinGuess = action.state.minGuess;
+      let newMaxGuess = action.state.maxGuess;
```
mayank99, line 395: I'll repeat what I said internally, which is that we should only use `action.state` should only be used where it is actually necessary, rather than replacing all instances of `state`.

Also, a more descriptive name would help distinguish it from `state`.

mayank99: > Since we're using `action.state` in some places, I thought it could lead to two sources of truth and confusion. So, I thought to just use one everywhere.

It's more confusing to use `action.state` because that's not a thing that's usually done. `state` is the more familiar pattern.

Wherever we use `action.state`, we should add a code comment explaining why. This way, we avoid doing the unusual thing where unnecessary and address any possible confusion at the same time.

And since we do not need it for all actions, we can also avoid passing it.

> Thus, I don't think `action.state` is the "previous state."

It is still the "previous state", because the reducer is used for calculating the "next state". When using reducers, an action is often dispatched with information about the "next"/"desired" state, rather than the previous/current one.

You could name it `currentState` if that's more clear to you. The actual name is not super important, I just want to meaningfully differentiate it from the `state` argument. But this is the less important part of the comment. I think code comments would be helpful regardless of the name.

mayank99: Yeah, my thinking is that any workaround should be kept confined to the minimum required surface area.

Investigating it in the future sounds good. There is a related TODO in #2154 which could be investigated at the same time (we did technically reduce the number of dependencies, but by relying on `guessState` we actually re-execute the Effect even more frequently, so the TODO is still unsolved).

mayank99 (approved):
LGTM. Nice find on the quick workaround.

smmr-dn (approved):

PR #2344, opened 2024-11-18
Docs site supports docs for unstable APIs

r100-stack:
## Changes

Moved some unstable API docs related changes from #2001 into this PR. This PR adds an Astro component called `UnstableApiCard` that is similar to the `Placeholder` but has a message that is instead the warning for the unstable API.

I thought to replace the info icon with a flask, similar to the experimental API icon in most other docs site. However, I couldn't find a flask icon. So, I used a warning icon for now.

Also, I added a `componentProp` in `PropsTable` to help choose the component who's props is to be displayed from the `.d.ts` file. (https://github.com/iTwin/iTwinUI/pull/2001#discussion_r1846639610)

## Testing

<details><summary>Old screenshot</summary>
<p>

image

</p>
</details> 

image

## Docs

N/A

### DIFF
--- a/apps/website/src/components/FeedbackLink.astro
+++ b/apps/website/src/components/FeedbackLink.astro
@@ -5,7 +5,7 @@ type Props = { page: string } & astroHTML.JSX.AnchorHTMLAttributes;
 let { page, ...rest } = Astro.props;
 
 const repoUrl = 'https://github.com/iTwin/iTwinUI';
-const githubFeedbackUrl = `${repoUrl}/discussions/new?category=documentation&title=Feedback%20on%20%60${page}%60&body=Replace%20this%20text%20with%20your%20feedback%20on%20the%20%60${page}%60%20page.`;
+const githubFeedbackUrl = `${repoUrl}/discussions/new?category=feedback&title=Feedback%20on%20%60${page}%60&body=Replace%20this%20text%20with%20your%20feedback%20on%20the%20%60${page}%60%20page.`;
 ---
 
 <a href={githubFeedbackUrl} {...rest}>


apps/website/src/components/DocsNoteCard.astro
```diff
@@ -0,0 +1,31 @@
+---
+
+---
+
```
mayank99, line 4: nit: frontmatter is not used

```suggestion
```

apps/website/src/components/UnstableApiCard.astro
```diff
@@ -0,0 +1,12 @@
+---
+import InformationIcon from '~/icons/InformationIcon.astro';
+import DocsNoteCard from './DocsNoteCard.astro';
+---
+
+<DocsNoteCard {...Astro.props}>
+  <InformationIcon aria-hidden='true' slot='icon' />
```
mayank99, line 7: Maybe we should use a warning ⚠️ icon?

mayank99 (approved):

smmr-dn (approved):

apps/website/src/components/UnstableApiCard.astro
```diff
@@ -0,0 +1,12 @@
+---
+import WarningIcon from '~/icons/WarningIcon.astro';
+import DocsNoteCard from './DocsNoteCard.astro';
+---
+
+<DocsNoteCard {...Astro.props}>
+  <WarningIcon aria-hidden='true' slot='icon' />
+  <p slot='content'>
+    <strong>Unstable</strong>: This component is marked unstable to help us gather feedback and
```
mayank99, line 9: I just realized that we are not actually asking for feedback.

Maybe we can create a "Feedback" category on GitHub Discussions, and link it here?

mayank99: We could also repurpose the existing ["Documentation" cateogory](https://github.com/iTwin/iTwinUI/discussions/categories/documentation).

If we do this, will need to also update the `FeedbackLink.astro` component to point to the new category.

mayank99 (approved):

PR #2345, opened 2024-11-18
Allowed passing customized icon for completed steps in `Stepper`

smmr-dn:
## Changes

Added a new `stepCircleRenderer` prop in `Stepper` which allows users passing in their customized icon or content to replace the default step number for completed steps.

image

## Testing

Added unit tests.

## Docs

Added minor changeset. Updated documentation as well as stories for `Stepper`.

## After-PR TODO:
- [ ] Improve documentation example to better explain the enhancement users specified. (https://github.com/iTwin/iTwinUI/pull/2345#discussion_r1847463604)

### DIFF
--- a/.changeset/curly-bikes-protect.md
+++ b/.changeset/curly-bikes-protect.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`Stepper` now allows users to pass in custom icon or content in each step circle to indicate step completion.
--- a/.changeset/curly-bikes-protect.md
+++ b/.changeset/curly-bikes-protect.md
@@ -2,4 +2,4 @@
 '@itwin/itwinui-react': minor
 ---
 
-`Stepper` now allows users to pass in custom icon or content in each step circle to indicate step completion.
+`Stepper` now allows passing custom icon or content in each step circle to indicate step completion. This can be done by using a `stepContent` property in each item of the `steps` array.
--- a/apps/website/src/content/docs/stepper.mdx
+++ b/apps/website/src/content/docs/stepper.mdx
@@ -64,7 +64,7 @@ The ‘Step X of X’ label that appears in the default long stepper can be repl
 
 By default, each step circle shows the step number. However, users can customize the icon or content displayed for each step in the `Stepper` component by providing the `stepContent` property in each item of the `steps` array.
 
-This property is a function that takes the step index and returns custom content to represent the status of each step. A common use case for this customization is to indicate step completion.
+This property is a function that returns custom content to represent the status of each step. A common use case for this customization is to indicate step completion.
 
 <LiveExample src='Stepper.customIcon.jsx'>
   <AllExamples.StepperCustomIconExample client:load />
--- a/apps/website/src/content/docs/stepper.mdx
+++ b/apps/website/src/content/docs/stepper.mdx
@@ -62,7 +62,7 @@ The ‘Step X of X’ label that appears in the default long stepper can be repl
 
 ### Custom Icon
 
-By default, each step circle displays the step number. However, end users can customize the icon or content shown for completed steps in the `Stepper` component by using the `stepCircleRenderer` prop. This prop accepts a function that takes the step index and returns the custom content to represent step completion.
+By default, each step circle displays the step number. However, end users can customize the icon or content shown for completed steps in the `Stepper` component by passing the `stepCircleRenderer` property in each step item. This property is a function that takes the step index and returns the custom content to represent step completion.
 
 <LiveExample src='Stepper.customIcon.jsx'>
   <AllExamples.StepperCustomIconExample client:load />
--- a/apps/website/src/content/docs/stepper.mdx
+++ b/apps/website/src/content/docs/stepper.mdx
@@ -62,7 +62,7 @@ The ‘Step X of X’ label that appears in the default long stepper can be repl
 
 ### Custom Icon
 
-By default, each step circle displays the step number. However, end users can customize the icon or content shown for completed steps in the `Stepper` component by passing the `stepCircleRenderer` property in each step item. This property is a function that takes the step index and returns the custom content to represent step completion.
+By default, each step circle displays the step number. However, end users can customize the icon or content shown for completed steps in the `Stepper` component by passing the `stepCircleRenderer` property in each item of the `steps` array. This property is a function that takes the step index and returns the custom content to represent step completion.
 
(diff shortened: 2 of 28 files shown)


packages/itwinui-react/src/core/Stepper/Stepper.tsx
```diff
@@ -41,6 +41,10 @@ export type StepperProps = {
    *  Option to provide localized strings.
    */
   localization?: StepperLocalization;
+  /**
+   * Custom content passed for completed steps.
+   */
+  stepCircleRenderer?: (completedIndex: number) => React.ReactNode;
```
mayank99, line 47: This API feels a bit strange. Is there a reason we can't allow passing a ReactNode in the steps array itself?

e.g.:

```jsx
const steps = [
  { name: 'First Step', stepContent: () => <SvgCheckmarkSmall /> },
  { name: 'Second Step', stepContent: () => <SvgCheckmarkSmall /> },
  { name: 'Third Step', stepContent: () => <SvgCheckmarkSmall /> },
  { name: 'Last Step', stepContent: () => <SvgCheckmarkSmall /> },
];
```

packages/itwinui-css/src/stepper/stepper.scss
```diff
@@ -82,6 +83,10 @@
   background-color: var(--_iui-stepper-step-background-color);
   color: var(--_iui-stepper-step-number-color);
 
+  :is(svg, img) {
```
mayank99, line 86: Why is `img` included here? `fill` does nothing on `img`.

smmr-dn: I have removed `img`.

packages/itwinui-react/src/core/Stepper/Stepper.tsx
mayank99: `Select` and especially `ComboBox` a bit different because they often require large datasets, and we fully control the rendering of options (and pass certain information back to the consumer). Still, I'm not happy with those item renderer APIs; I think I would design it differently if I was to do it again.

This `Stepper` case is a lot simpler though, so it makes sense to just ask for it within the steps. In some ways it's similar to handling of icons within `Select`/`ComboBox` options.

packages/itwinui-react/src/core/Stepper/StepperStep.tsx
```diff
@@ -132,7 +137,7 @@ export const StepperStep = React.forwardRef((props, forwardedRef) => {
           {...circleProps}
           className={cx('iui-stepper-circle', circleProps?.className)}
         >
-          {index + 1}
+          {stepContent && isPast ? stepContent(index) : index + 1}
```
mayank99, line 140: We do not need to pass the `index` back into it `stepContent`, because it's not useful (the whole point is that the user wants to render something else).

Also I agree with @r100-stack's other comment.

```suggestion
          {stepContent ? stepContent() : index + 1}
```

mayank99 (commented):
LGTM apart from one small comment.

packages/itwinui-react/src/core/Stepper/StepperStep.tsx
mayank99: It should still be a function, so that the steps array can be defined outside a component. (There shouldn't be any lose React Nodes created outside a component; using a function allows us to delay the creation of these nodes)

packages/itwinui-react/src/core/Stepper/Stepper.tsx
```diff
@@ -21,6 +21,10 @@ export type StepProperties = {
    * A tooltip giving detailed description to this step.
    */
   description?: string;
+  /**
+   * Custom content passed for completed steps.
```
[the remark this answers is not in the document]
smmr-dn: I have made the change accordingly. Since we now allow passing custom content in any step (with `isPast` removed), should I add more logic in the docs and stories to indicate which step is completed for the checkmark icon example as in what I have done in the latest [commit](fa6ded83a48e358cd85daafa6b5977e9922d2447), to make it more reasonable for the use of checkmark?
image

apps/website/src/content/docs/stepper.mdx
```diff
@@ -60,6 +60,14 @@ The ‘Step X of X’ label that appears in the default long stepper can be repl
   <AllExamples.StepperLocalizationExample client:load />
 </LiveExample>
 
+### Custom Icon
+
+By default, each step circle displays the step number. However, end users can customize the icon or content shown for completed steps in the `Stepper` component by passing the `stepCircleRenderer` property in each item of the `steps` array. This property is a function that takes the step index and returns the custom content to represent step completion.
```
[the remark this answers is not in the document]
smmr-dn: I have made this change.

packages/itwinui-react/src/core/Stepper/Stepper.tsx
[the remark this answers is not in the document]
smmr-dn: I have added `Set` in the doc example.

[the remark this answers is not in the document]
smmr-dn: I have replaced the `Set` and used `currentStep` instead and also added the after-PR todo item.

apps/website/src/content/docs/stepper.mdx
```diff
@@ -60,6 +60,16 @@ The ‘Step X of X’ label that appears in the default long stepper can be repl
   <AllExamples.StepperLocalizationExample client:load />
 </LiveExample>
 
+### Custom Icon
+
+By default, each step circle shows the step number. However, users can customize the icon or content displayed for each step in the `Stepper` component by providing the `stepContent` property in each item of the `steps` array.
+
+This property is a function that takes the step index and returns custom content to represent the status of each step. A common use case for this customization is to indicate step completion.
```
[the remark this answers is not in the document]
smmr-dn: I have made the change.

.changeset/curly-bikes-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+`Stepper` now allows users to pass in custom icon or content in each step circle to indicate step completion.
```
mayank99, line 5:
```suggestion
`Stepper` now allows passing custom icon or content in each step circle to indicate step completion. This can be done by using a `stepContent` property in each item of the `steps` array.
```

mayank99 (approved):

.changeset/curly-bikes-protect.md
smmr-dn: I have made the change.

PR #2350, opened 2024-11-20
Allowed hiding demo code in `LiveExample`

smmr-dn:
## Changes

This PR provided the option to hide the demo code in documentation by setting the `hideCode` prop to true. This is a TODO item from #2326.

## Testing

Confirmed that the demo code is hidden when passing `hideCode={true}` to `LiveExample` in `SkipToContentLink` doc.

## Docs

N/A


apps/website/src/components/LiveExample.astro
```diff
@@ -8,17 +8,23 @@ import itwinuiStyles from '@itwin/itwinui-react/styles.css?url';
...
+   * A flag to determine whether the first code snippet will truncate and turn into an expandable disclosure.
    * @default true
    */
   truncate?: boolean;
+
+  /**
+   * A flag to determine whether the demo code should be visible.
+   * @default false
+   */
+  hideCode?: boolean;
```
mayank99, line 24: Instead of adding a new prop, maybe you can make `src` optional?

mayank99: Oh right, that would be a problem. Let's leave it as-is to avoid complications.

mayank99 (approved):

PR #2359, opened 2024-11-27
Added buffer to all sides of `Tooltip`'s floating content

smmr-dn:
## Changes

This PR added a buffer of 4px to all sides of the `Tooltip`'s floating content so that when users slightly move their mouse outside of the content box, the tooltip is still visible.

## Testing

Confirmed that the tooltip is still visible when the mouse moves outside of the border.

https://github.com/user-attachments/assets/718f75fd-a63f-451c-89bf-3962cb7cbb52

## Docs

N/A

### DIFF
--- a/.changeset/clean-worms-destroy.md
+++ b/.changeset/clean-worms-destroy.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-css': major
+---
+
+The `Tooltip` now remains visible when hovered up to `4px` outside its border.
--- a/.changeset/clean-worms-destroy.md
+++ b/.changeset/clean-worms-destroy.md
@@ -1,5 +1,5 @@
 ---
-'@itwin/itwinui-css': major
+'@itwin/itwinui-css': patch
 ---
 
 The `Tooltip` now remains visible when hovered up to `4px` outside its border.
--- a/packages/itwinui-css/src/tooltip/tooltip.scss
+++ b/packages/itwinui-css/src/tooltip/tooltip.scss
@@ -9,7 +9,7 @@
   text-align: center;
   border-radius: var(--iui-border-radius-1);
   font-size: var(--iui-font-size-0);
-  overflow: hidden;
+  overflow: visible;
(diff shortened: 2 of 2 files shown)


smmr-dn:
There are two things I would like to ask for more opinions in this PR:
- I have tested with and without the `safePolygon` specified in the `useHover` hook's option and noticed that even with a larger inset that overlaps the trigger element (I have set the pseudo-element's background to white for debugging), the tooltip disappears as soon as the mouse moves out of it. Should I keep this setup so that the transition between the element and the floating content can be kept?

https://github.com/user-attachments/assets/bf9f88b4-b0f6-467b-8da7-ecb8c7abc70a

- I notice the `overflow: hidden` clips off the edges of the `::before` pseudo-element, which results in the buffer size not showing up correctly. Can we either set `overflow` to either `unset` or `visible` for this buffer content to be fully shown?

|overflow: hidden|overflow: visible|
|-|-|
|<video src="https://github.com/user-attachments/assets/b7f688c5-8a67-4ccc-ad22-6bced5f08c2b" />|<video src="https://github.com/user-attachments/assets/13e69b37-bab3-476d-b5d9-1abedaffd7b2" />|

mayank99:
> I have tested with and without the `safePolygon` specified in the `useHover` hook's option and noticed that even with a larger inset that overlaps the trigger element (I have set the pseudo-element's background to white for debugging), the tooltip disappears as soon as the mouse moves out of it. Should I keep this setup so that the transition between the element and the floating content can be kept?

If it doesn't behave properly without `safePolygon`, then yeah we should keep it.

>     * I notice the `overflow: hidden` clips off the edges of the `::before` pseudo-element, which results in the buffer size not showing up correctly. Can we either set `overflow` to either `unset` or `visible` for this buffer content to be fully shown?

Yeah that's fine. I'm not sure why we even had `overflow: hidden` (looks like it was added 3 years ago without any explanation).

mayank99 (approved):
LGTM. Just make sure to add a changeset.

Issue #2356, opened 2024-11-27
`Select` and `ComboBox` with size `small` have different heights


saskliutas:
### Describe the bug (current behavior)

`ComboBox` and `Select` components with size `small` have different height.

### Expected Behavior

Both components should have same height.

### Link to minimal repro

https://stackblitz.com/edit/github-7qaa9m?file=src%2FApp.tsx

### Steps To Reproduce

1. Open sandbox
2. Observer height difference

### Package Version

- [x] I'm using the latest version of iTwinUI.

PR #2362, opened 2024-12-05
Reduced small `Select`'s content height to be consistent with small `ComboBox`

smmr-dn:
## Changes

As user reported the [issue](https://github.com/iTwin/iTwinUI/issues/2356) regarding the height difference between small `ComboBox` and `Select`, this PR reduced the `min-block-size` and `line-height` of the content inside small `Select`  to `--iui-size-m` so that the overall height would not exceed 24px.

|Before|After|
|-|-|
||| 

## Testing

Approved test screenshots.

## Docs

Added changesets for `itwinui-react` and `itwinui-css` packages.


mayank99 (changes requested):
Please fix the centering. There shouldn't be any test image changes in stories that don't use the small Select.

smmr-dn:
> Please fix the centering. There shouldn't be any test image changes in stories that don't use the small Select.

I figured when I used the fallback values (--iui-size-l) for `min-block-size` and `line-height` for other sizes, it returned 21px instead of 24px, which eventually caused the texts to be un-centered. I tried specifying the min-block-size and line-height vars for medium and large size. The screenshots have been corrected as well.

mayank99 (approved):
LGTM

PR #2368, opened 2024-12-11
Fine-tune delays in `Tooltip` and associated components

r100-stack:
> [!NOTE]  
> Merging this PR in the next minor release (https://github.com/iTwin/iTwinUI/pull/2368#discussion_r1882819605).

## Changes

After we fixed the default `Tooltip` delay in #2263, a couple of components had the problem of lingering tooltips. To fix it, this PR wraps some of the parts of the affected components within Floating UI's [`FloatingDelayGroup`](https://floating-ui.com/docs/FloatingDelayGroup).

Components where `FloatingDelayGroup` wrapper is added:
* `SideNavigation`
* `Slider`
* `Stepper`
* `List` ([#2368 (review)](https://github.com/iTwin/iTwinUI/pull/2368#pullrequestreview-2500362933))

While this PR fixes the lingering tooltip problem within the items of the component itself, it doesn't fix the problem between two components. To fix the latter part also, I had considered moving `FloatingDelayGroup` to the `ThemeProvider` but then didn't go that route based on our internal discussion. One of the reasons to not go that route was to prevent unintended issues.

According to our discussion, I also changed the default tooltip delay from `{ open: 50, close: 250 }` to `{ open: 100, close: 200 }`. If we later find research regarding the best delay values, we can further fine-tune this default delay.

## Testing

CI passing after small updates to the tests.

<details><summary>Before and After</summary>

|Component|Before|After|
|-|-|-|
|Stepper|<video src="https://github.com/user-attachments/assets/db999a87-6078-4de8-8565-5f8c9f2dccde" />|<video src="https://github.com/user-attachments/assets/e1d1aa25-39cb-48b1-81ba-bb41dc043512" />|
|SideNavigation|<video src="https://github.com/user-attachments/assets/3ffd8e67-40f8-4fa7-ab34-b883f0dc055f" />|<video src="https://github.com/user-attachments/assets/ffa1fecf-3b9e-4e4f-b078-fa2fe11e9407" />|
|Slider|<video src="https://github.com/user-attachments/assets/3a0f8475-3419-4deb-ab13-0f8bba6d224c" />|<video src="https://github.com/user-attachments/assets/177b3db3-f8a9-44da-8845-60b6809854f1" />|

</details> 

## Docs

Added a changeset.

## After PR TODOs:

* [ ] See how to solve lingering tooltips across different components. ([#2368 (review)](https://github.com/iTwin/iTwinUI/pull/2368#pullrequestreview-2500163801))
* [ ] Document more about the `placement` of the `Tooltip` ([#2368 (review)](https://github.com/iTwin/iTwinUI/pull/2368#pullrequestreview-2500765266))


smmr-dn (commented):
I also notice the problem caused by delay when the `label` prop is passed into `IconButton`. Does it still need to be wrapped with `FloatingDelayGroup`?

https://github.com/user-attachments/assets/568151b3-ed7c-47ce-b669-2ec6d589e8a8

.changeset/seven-parents-complain.md
mayank99, line 1: We should split this into two changesets, one for mentioning the delay grouping (and also mention which components are affected) and the other for the default delay.

Also I wonder we should release this in a minor, even if the changesets are patch. While there are no API changes, this PR is quite large for a patch, and maybe some of our consumers were relying on the current delays (e.g. see https://github.com/iTwin/iTwinUI/discussions/2352#discussioncomment-11341091).

mayank99 (commented):
> > I also notice the problem caused by delay when the `label` prop is passed into `IconButton`. Does it still need to be wrapped with `FloatingDelayGroup`?
> 
> Since those `Tooltip`s are across different instances of `IconButton`s, this PR doesn't address that atm. 

Could we use `FloatingDelayGroup` in `List`? It makes sense because it is an isolated component containing repeated items (similar to `SideNavigation`/`Stepper`/`ButtonGroup`/etc).

.changeset/seven-parents-complain.md
mayank99: Looks good, but I don't think we need to make it a `minor` changeset. I was only mentioning it so we avoid merging the PR for the purposes of the next `patch` release.

packages/itwinui-react/src/core/Tooltip/Tooltip.tsx
```diff
@@ -197,7 +197,7 @@ const useTooltip = (options: TooltipOptions = {}) => {
 
   const interactions = useInteractions([
     useHover(floating.context, {
-      delay: delay !== 0 ? delay : { open: 50, close: 250 }, // If FloatingDelayGroup exists, use its delay
+      delay: delay !== 0 ? delay : { open: 100, close: 200 }, // If FloatingDelayGroup exists, use its delay
```
mayank99, line 200: nit: Could we create a constant in this file for centralizing `{ open: 100, close: 200 }`?

Then we can use it everywhere, instead of creating a new object literal each time.

mayank99 (commented):
> But when going from bottom to top, there's not much improvement.

One fix for that would be to change the tooltip's `placement`, which consumers can do on their own.

I think this is quite a common issue for lists, and we should probably recommend changing `placement` in our documentation somewhere.

mayank99 (approved):
LGTM

smmr-dn (approved):

PR #2370, opened 2024-12-11
Fixed `Panels` animations not working on docs site

r100-stack:
## Changes

I believe the animations were not working on the docs site because the examples were placed in a shadow DOM. Thus, calling `getElementById("…").scrollIntoView()` wouldn't work since `getElementById()` would not return the `Panel` that is in the shadow DOM. Exact code:

https://github.com/iTwin/iTwinUI/blob/78ae4150219a9270ec24fee79b946f66eea28092/packages/itwinui-react/src/core/Panels/Panels.tsx#L87

To fix it, this PR no longer uses the `ownerDocument`. Since the panels are always within the `PanelsWrapper`, we can instead use search for the `Panel` within the ref of the the `PanelsWrapper`:

```diff
- ref.current?.ownerDocument.getElementById(newActiveId)?.scrollIntoView()
+ ref.current?.querySelector(`[id='${newActiveId}']`)?.scrollIntoView()
```

For the selector, I didn't use `` `#${newActiveId}` `` since React's `useId()` gives ids starting with `":"` that gives an error if used in `querySelector()`.

## Testing

Confirmed that the animations now work on the docs site.

## Docs

Added a changeset. But since this issue might be rare, I could remove the changeset if any reviewer feels so.

### DIFF
--- a/packages/itwinui-react/src/core/ProgressIndicators/ProgressLinear.tsx
+++ b/packages/itwinui-react/src/core/ProgressIndicators/ProgressLinear.tsx
@@ -16,7 +16,7 @@ type ProgressLinearProps = {
   /**
    * Progress variant. If true, `value` will be ignored.
    *
-   * Defaults to true if `value` is passed, otherwise false.
+   * Defaults to false if `value` is passed, otherwise true.
    */
   indeterminate?: boolean;
   /**


smmr-dn (approved):

packages/itwinui-react/src/core/Panels/Panels.tsx
```diff
@@ -84,7 +84,7 @@ export const PanelsWrapper = React.forwardRef((props, forwardedRef) => {
       ReactDOM.flushSync(() => setActivePanelId(newActiveId));
       onActiveIdChange.current?.(newActiveId);
 
-      ref.current?.ownerDocument.getElementById(newActiveId)?.scrollIntoView({
+      ref.current?.querySelector(`[id='${newActiveId}']`)?.scrollIntoView({
```
mayank99, line 87: Have you considered using [`getRootNode()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode) to grab the owner document or shadow-root, and then calling `getElementById` on it?

e.g. something like:

```suggestion
      ref.current?.getRootNode()?.querySelector(newActiveId)?.scrollIntoView({
```

mayank99: nit: You can type it as `Document | ShadowRoot`. Type assertions are all lies, so it doesn't matter for the purposes of the runtime code, but it's more accurate and descriptive (since the issue we are trying to solve is specifically related to shadow-roots).

The third possible return value of `getRootNode()` does make me nervous, but hopefully we never run into it 🤞

mayank99 (approved):
LGTM! Nice work on diagnosing this as a shadow DOM issue

packages/itwinui-react/src/core/Panels/Panels.tsx
mayank99: I already mentioned the reasoning: it's more descriptive. The whole point of using `getRootNode()` is so that we get the encompassing shadow-root (if there is one). Think of it like a code comment.

PR #2372, opened 2024-12-13
Correction in ProgressLinear JSDocs

arome:
## Changes

Small fix in `ProgressLinear`'s JSDocs for its `indeterminate` prop.

i.e. the prop's default value is `true` when no `value` is passed and its default value is `false` when `value` is passed. ([sandbox](https://stackblitz.com/edit/github-ujbmpyyv?file=src%2FApp.tsx))

## Testing

N/A

## Docs

N/A


mayank99 (approved):
Thanks.

PR #2373, opened 2024-12-16
Theme bridge to `v5.0.0`

r100-stack:
## Changes

#2364 refactors some code to introduce private css variables that help with the bridge code mapping (more info below). This PR will be merged after that PR.

---

Since [`@itwin/itwinui-react@5.0.0-alpha.0`](https://www.npmjs.com/package/@itwin/itwinui-react/v/5.0.0-alpha.0) is now published, this PR implements a bridge for existing iTwinUI applications and packages to look similar to this future version and help in migration and early adoption.

The `ThemeProvider` now offers a new prop called `future` that has an option called `themeBridge` which when set to true adds `data-iui-bridge='true'` to the `iui-root`. In addition to enabling `future.themeBridge`, users will have to add `as={Root}` to load `itwinui-react v5`'s CSS variables (`--kiwi-…`). ([#2373 (comment)](https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1940062606)).

React:

```diff
<ThemeProvider
+  as={Root}
   themeOptions={{
     bridgeToFutureVersions: true
   }}
>
    …
</ThemeProvider>
```

Rendered:

```diff
<div
-  class="_iui3160-root"
+  class="🥝-root _iui3160-root"

   data-iui-theme="dark"
   data-iui-contrast="default"
+  data-iui-bridge="true"

+  data-kiwi-theme="dark"
+  data-kiwi-density="dense"
>
      …
</div>
```

The presence of this data attribute activates the bridge code that maps the existing `--iui-…` CSS variables to `--kiwi-…` variables from `v5`. It also adds some limited styles only whenever it was needed. Since some of these mappings are subjective, especially when something similar doesn't exist in iTwinUI-react v5, open to other views.

```scss
.iui-root:where([data-iui-bridge='true']) {
  --iui-color-background: var(--kiwi-color-bg-neutral-base);
  … // other mappings
}
```

The `future.themeBridge` option and `data-iui-bridge` names are bikesheddable. I also instead considered adding a new prop like `bridgeToFutureVersions` or new accepted themes itself like `dark-v5` and `light-v5` that the `theme` prop could accept. However, for `bridgeToFutureVersions`, we decided against it as it wasn't very expansible and I felt that the new themes would lead to many code changes and potentially confusion and complexity. Open to other suggestions.

The mapping code lies in `bridge.scss` that is exported by `itwinui-css` in the `itwinui.bridge` layer. There are some TODOs in that file for whenever I wasn't sure of the mapping. Open to comments on those, if any.

This mapping was most tested in the dark theme. Some testing was done with the light theme too and will test more too. The high-contrast dark and high-contrast light themes continue using the regular dark and regular light themes. Once iTwinUI-react v5 supports high-contrast themes, we can make the necessary changes, if any, in the bridge.

TODO:

* [x] Make comments like "To match Kiwi/iTwinUI" to something more descriptive.

## Testing

Manually tested in `react-workshop`, `css-workshop`, and `vite-playground`. Added `bridgeToFutureVersions` toggles in `react-workshop` and `css-workshop` next to the theme toggles.

- [x] Approve any failing images or stop them from failing

## Docs

* [x] Added changesets.
* [x] Add instructions on how to setup `iTwinUI-react@5`'s `<Root>` on the users' side ([#2373 (comment)](https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1890889070)).

## After PR TODOs

* [x] Added `flex-shrink: 0` to iTwinUI _v3_'s `iui-button-icon` ([#2373 (comment)](https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1896845716))
  * https://github.com/iTwin/iTwinUI/pull/2396
* [ ] Are `status-hover` variables not being used anywhere in iTwinUi v3? Investigate. (https://github.com/iTwin/iTwinUI/pull/2373/commits/162008fc3c1e4c16a5313963b492c31a86185e10)
* [ ] Have component level overrides where the kiwi focus rings aren't correct ([#2373 (comment)](https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1941750739), [#2373 (comment)](https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1941783019)).
* [ ] Update the [theme bridge wiki](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-v5-theme-bridge) to reflect the changes from this PR when this PR is included in a release.

### DIFF
--- a/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -67,8 +67,11 @@ type RootProps = {
    *
    * The 'inherit' option is intended to be used by packages, to enable incremental adoption
    * of iTwinUI while respecting the theme set by the consuming app. It will fall back to 'light'
-   * if no parent theme is found. Additionally, it will attempt to inherit `themeOptions.highContrast`,
-   * `portalContainer`, and `future.themeBridge` (if possible).
+   * if no parent theme is found. Additionally, it will attempt to inherit `themeOptions.highContrast`
+   * and `portalContainer` (if possible).
+   *
+   * `future.themeBridge` will be inherited regardless of `theme` value. To disable it, explicitly set
+   * `future.themeBridge` to false.
    *
    * @default 'inherit'
    */


mayank99 (commented):
Isn't #2364 a prerequisite for this? What's the plan there?

apps/css-workshop/src/pages/_layout.astro
```diff
@@ -9,49 +11,53 @@ const slug = Astro.url.pathname.replace('/', '');
   Copyright (c) Bentley Systems, Incorporated. All rights reserved.
   See LICENSE.md in the project root for license terms and full copyright notice.
 -->
-<html lang='en-US' id='theme'>
-  <head>
```
mayank99, line 13: Why did you remove the `<head>`?

apps/css-workshop/package.json
```diff
@@ -4,6 +4,13 @@
   "private": true,
   "version": "0.1.0",
   "description": "Workshop environment containing HTML pages demonstrating itwinui-css",
+  "dependencies": {
```
mayank99, line 7: nit: everything should be under devDependencies, as per the established convention

packages/itwinui-react/src/styles.js/styles.module.css
```diff
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3;
+@layer itwinui-v1, itwinui.v1, itwinui.v2, itwinui.v3, itwinui.bridge;
```
mayank99, line 6: I've realized just now that we should still include `@layer reset` in here as backup, just in case the consuming application does not do it manually.

We will recommend consumers to set the correct layer order in their `<head>`. The first occurrence of a layer name gets used for the purpose of ordering, so our fallback will just be ignored (so there's no downside in doing os).

packages/itwinui-css/src/bridge.scss
```diff
@@ -0,0 +1,513 @@
+// Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+// See LICENSE.md in the project root for license terms and full copyright notice.
+[data-iui-bridge='true'] {
+  * {
```
mayank99, line 4: We should specifically target the element where these variables are being set (i.e. `.iui-root`) instead of `*`.

mayank99: However, this does make me think: the attribute should be set on the same element that has all the variables.

So can't we target `.iui-root[data-iui-bridge="true"]`?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -234,28 +251,60 @@ const MainRoot = React.forwardRef((props, forwardedRef) => {
...
+      shouldApplyDark,
+      shouldApplyHC,
+      themeOptions?.bridgeToFutureVersions,
+    ],
+  );
+
+  return themeOptions?.bridgeToFutureVersions ? (
+    <ITwinUIV5Root
+      colorScheme={shouldApplyDark ? 'dark' : 'light'}
+      density='dense'
```
mayank99, line 302: Continuing the same comment, this `Root` should be set up by consumers to wrap around iTwinUI v3's `ThemeProvider`. We should have well-written instructions on how to set it up.

packages/itwinui-css/src/bridge.scss
mayank99, line 1: This file has a lot more CSS than I was expecting, and I haven't been able to take a closer look. After the remaining comments are handled, let's go through this file synchronously tomorrow. It would be more productive than going through asynchronous code review process, just because of the breadth of changes.

As mentioned in https://github.com/iTwin/iTwinUI/pull/2364#discussion_r1890877446, we need to have a solid explanation for every component-level change, which we can add as code comments.

packages/itwinui-css/src/bridge.scss
mayank99: > I realized that it was happening because the iTwinUI-react variables were actually unlayered and so taking priority over the bridge variable overrides.

This sounds like a bug that we should fix. All our CSS should be layered.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: > `react-workshop` now depends on `iTwinUI-react@5.0.0-alpha.0` and sets up `<ITwinUIV5Root>`/`<Root>` on its end. ([a216d1b](https://github.com/iTwin/iTwinUI/commit/a216d1b0f3e2a1386a7a2444c6987f19626b949b)).

`<Root>` should be wrapped around `<ThemeProvider>`, not the other way around, otherwise the v3 variables will take preference.
 
> Regarding the "well-written instructions", I'm not sure where we want to document everything related to this theme bridge. Once we decide a place, I will add those instructions. For now, added a checklist item in the PR description.

We can start with a page in the github wiki, and move it elsewhere later if necessary.

packages/itwinui-css/src/bridge.scss
mayank99: After investigating this, it's not a bug in our package but rather in the usage. It's happening because we are importing `@itwin/itwinui-variables` in a JS file (outside a layer).

https://github.com/iTwin/iTwinUI/blob/330361346832306f81a63eff04d659adba47faf5/apps/react-workshop/.ladle/components.tsx#L13-L15

We can fix this by moving it into `global.css` and importing it into a layer.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: > shall I create a WIP GitHub wiki for it?

Yeah, let's do it now.

packages/itwinui-css/src/bridge.scss
mayank99: Try not to remove code without investigating why it was added. The variables import was necessary for the background switcher, because the `<div class="ladle-background">` element is rendered outside the ThemeProvider.

mayank99: > To confirm, what do you mean by "CSS modules" here? Do we mean the `_iui3` prefix transformation?

Yes, that's what I mean. We use [Vite's CSS modules](https://vite.dev/guide/features.html#css-modules) feature for transforming the `styles.module.css` file.

We need to ensure that everything in the `@itwin/itwinui-react/styles.css` file is scoped properly. So the loose `[data-iui-bridge='true']` selector needs to changed to `.iui-root[data-iui-bridge='true']` (which will then get transformed to something like `._iui3160-root[data-iui-bridge='true']`.

packages/itwinui-css/src/bridge.scss
mayank99: Pushed a whole bunch of commits during our pairing sessions: https://github.com/iTwin/iTwinUI/pull/2373/files/7733e97da4af46b193e750b529737ac2c6063a1f..55fbde0cda623b1e6ca4ac86999dc235bbc3a7fb

packages/itwinui-css/src/bridge.scss
mayank99: This is resolved in [`4ccd98c`](https://github.com/iTwin/iTwinUI/pull/2373/commits/4ccd98cd3cd0658faeead1b6d18272dcd92a4a7a)

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -40,6 +40,14 @@ export type ThemeOptions = {
...
    */
   highContrast?: boolean;
+  /**
+   * If enabled, the theme resembles the future iTwinUI version's theme (including alphas) *whenever possible*.
+   *
+   * This is useful in making apps looks like future versions of iTwinUI to help with incremental adoption.
+   *
+   * **NOTE**: Since this is a theme bridge to *future* versions, the theme could have breaking changes.
+   */
+  bridgeToFutureVersions?: boolean;
```
mayank99, line 50: We should bikeshed this name a little bit 😅

mayank99 (commented):
🚀 [3.17.0-dev.0](https://www.npmjs.com/package/@itwin/itwinui-react/v/3.17.0-dev.0) is released with [accompanying instructions](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-v5-theme-bridge) for consumers to try it out.

Early next week, we'll do some final cleanup before merging into `main`.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: I like "theme bridge" and "future". We should avoid `use` prefix to avoid confusion with hooks.

I was also thinking we might have other "future" options added (in the future 😅). So what if added a new `future` prop which accepts an object containing a `themeBridge` property?

```jsx
<ThemeProvider future={{ themeBridge: true }}>
```

apps/css-workshop/src/components/theme.js
```diff
@@ -143,15 +146,19 @@ class ThemeButton extends HTMLElement {
   }
 
   changeTheme = ({ target: { value: _theme } }) => {
+    const root = document.body;
+    const v5Root = document.documentElement;
```
mayank99, line 150: Similar comment here as Ladle: can we set all attributes on the same element (`<body>` or `<html>`) instead of mixing and matching?

apps/css-workshop/src/components/theme.js
```diff
@@ -143,15 +146,19 @@ class ThemeButton extends HTMLElement {
...
+    v5Root.dataset.colorScheme = theme;
+    root.dataset.iuiContrast = isHighContrast ? 'high' : undefined;
     this.shadowRoot.querySelector('#theme-color-scheme').innerHTML = `
-      :host {
-        color-scheme: ${theme.includes('light') ? 'light' : 'dark'};
+    :host {
+      color-scheme: ${theme.includes('light') ? 'light' : 'dark'};
       }
-    `;
+      `;
```
mayank99, line 161: nit: I think there's some unintended whitespace changes here.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -164,13 +180,17 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
...
+
   const [portalContainerFromParent] = useScopedAtom(portalContainerAtom);
 
   const contextValue = React.useMemo(
-    () => ({ theme, themeOptions }),
+    () => ({ theme, themeOptions, future }),
     // we do include all dependencies below, but we want to stringify the objects as they could be different on each render
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [theme, JSON.stringify(themeOptions)],
+    [theme, JSON.stringify(themeOptions), future],
```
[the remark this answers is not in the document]
mayank99: By "different" I mean literally different (i.e. fails the `===` check). `{ themeBridge: true }` is an object literal which gets re-created on every render (unless it's already memoized when passed). You can test this out in the browser console.

Using `JSON.stringify` ensures that the contents of the object are compared, so that our memoized context value only invalidates when the options are actually different.

packages/itwinui-css/src/bridge.scss
```diff
@@ -0,0 +1,361 @@
...
+      color: var(--kiwi-color-text-neutral-primary);
+    }
+  }
+
+  /* ---------------------------------------------------------------------------- */
+  /* Breadcrumbs */
+
+  .iui-breadcrumbs {
+    // Neutral color instead of accent
+    --_iui-breadcrumbs-content-color-default: var(--iui-color-text);
```
mayank99, line 120: I think we can get rid of this, right? We decided to use the same set of color variables as current iTwinUI v3 implementation.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -164,13 +180,17 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
   themeOptions.highContrast ??=
     themeProp === 'inherit' ? parent.highContrast : undefined;
 
+  const future = futureProp ?? {};
+  future.themeBridge ??=
```
smmr-dn, line 184: Just out of curiosity, why do we need a `??=` operator here?

apps/css-workshop/package.json
```diff
@@ -11,7 +11,12 @@
     "astro": "^4",
     "astro-relative-links": "^0.3.7",
     "backstopjs": "~6.2.1",
-    "npm-run-all": "^4.1.5"
+    "npm-run-all": "^4.1.5",
+    "@astrojs/react": "^4.1.0",
```
smmr-dn, line 15: nit/optional: I notice 4.1.2 was out recently. Should we update the package accordingly as well?

packages/itwinui-css/src/bridge.scss
```diff
@@ -0,0 +1,360 @@
...
+    background-color: var(--kiwi-color-bg-surface-tertiary);
+
+    // Kiwi uses box shadow only
+    border: none;
+    box-shadow: var(--kiwi-shadow-surface-xl);
+  }
+
+  .iui-list-item {
+    // Since iTwinUI v3 has hover background color but iTwinUI v5 doesn't
+    --_iui-list-item-background-hover: transparent;
```
mayank99, line 219: I don't think this is correct anymore. We do have a background hover in Kiwi.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -336,6 +358,7 @@ const PortalContainer = React.memo(
     isInheritingTheme,
     theme,
     themeOptions,
+    future,
```
mayank99, line 361: Looks like we are not passing this prop currently, so all the dropdowns are showing up with v3 theme.

packages/itwinui-css/src/bridge.scss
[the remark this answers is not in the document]
mayank99: To be clear, I'm asking for code comments that properly describe why every component-level override was added (you have a TODO item for this in the PR description).

Just as an example, this line should be accompanied by a comment such as "Disabled buttons in Kiwi do not have a visible border".

https://github.com/iTwin/iTwinUI/blob/162008fc3c1e4c16a5313963b492c31a86185e10/packages/itwinui-css/src/bridge.scss#L145

And then this should be accompanied with "For button-group, add back borders that we removed above".

https://github.com/iTwin/iTwinUI/blob/162008fc3c1e4c16a5313963b492c31a86185e10/packages/itwinui-css/src/bridge.scss#L158-L160

Generally speaking, every CSS declaration (or set of declaration for related changes) should have a code comment justifying its presence.

Also, I saw one of the TODOs was removed in [`162008f`](https://github.com/iTwin/iTwinUI/pull/2373/commits/162008fc3c1e4c16a5313963b492c31a86185e10). What action did we take to address this TODO?

packages/itwinui-css/src/button/base.scss
```diff
@@ -40,6 +40,7 @@ $disabledButtonSelectors: '[disabled], :disabled, [aria-disabled="true"], [data-
 
 @mixin iui-button-icon {
   display: inline-flex;
+  flex-shrink: 0;
```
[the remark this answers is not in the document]
mayank99: We should make another release `3.17.0-dev.1` before this PR merges. We can temporarily add it back for the purpose of that release (otherwise the SideNavigation would look broken).

pnpm-lock.yaml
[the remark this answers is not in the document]
mayank99: looks like this was undone in [`52252cf`](https://github.com/iTwin/iTwinUI/pull/2373/commits/52252cff21935e47ea500809331f2e6f57ccb31f)

is your formatter not configured properly?

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
mayank99: A related question: why are we using `futureProp ?? {}` here instead of setting the default value to `{}`?

mayank99: I'm asking about the default value, similar to how we do for `themeOptions` here:

https://github.com/iTwin/iTwinUI/blob/be0762e9c938a5b522b06cdb25df8846da9beaa5/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx#L161

We could set `future = {}` when destructuring from `props`, which would allow using `future.themeProps ??=` (because `future` will always be defined).

packages/itwinui-css/src/bridge.scss
mayank99: Well, we shouldn't need this component-level override, right? It should use whatever color we are setting globally.

packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -164,13 +180,16 @@ export const ThemeProvider = React.forwardRef((props, forwardedRef) => {
   themeOptions.highContrast ??=
     themeProp === 'inherit' ? parent.highContrast : undefined;
 
+  future.themeBridge ??=
+    themeProp === 'inherit' ? parent.context?.future?.themeBridge : undefined;
```
mayank99, line 184: In light of the issues raised internally on Teams, we might want to always inherit this flag even when `theme` is not set `"inherit"`.

I'm not sure if it's ever useful to partially enable theme bridge for some parts of the tree, but it would still be achievable by explicitly passing `false`.

```suggestion
  future.themeBridge ??= parent.context?.future?.themeBridge;
```

apps/react-workshop/.ladle/components.tsx
```diff
@@ -51,12 +53,20 @@ export const Provider: GlobalProvider = ({ children }) => {
...
+            applyBackground: false,
+            highContrast,
+          }}
+          future={{
+            themeBridge: futureThemeBridge,
+          }}
+        >
+          {children}
+        </ThemeProvider>
+      </ITwinUiV5Root>
```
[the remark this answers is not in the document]
mayank99: We might also want to set (and recommend setting) `synchronizeColorScheme` on the v5 root. This helps with popovers and other elements that are not DOM descendants of the root.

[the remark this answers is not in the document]
mayank99: > > Can we try rendering both roots as the same element? It might help with [iTwin/kiwi#259](https://github.com/iTwin/kiwi/issues/259).
> 
> I tried it but it unfortunately didn't work. i.e. kiwi variables were not showing up in the devtools styles' tab for `.🥝-root` even though `data-kiwi-theme="dark"` was correctly being applied.

The variables not showing up might be a Firefox devtools bug. As a workaround, you can close dev-tools, refresh the page, and re-open dev-tools. Also try it in a different browser.

Also I released `5.0.0-alpha.4` which includes https://github.com/iTwin/kiwi/pull/305, so try using that version.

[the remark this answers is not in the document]
mayank99: It looks like `<ThemeProvider as={Root}>` might possibly work. I tried it out with the currently released versions: https://stackblitz.com/edit/github-86h4vwqw-dffcv5sn?file=src%2Fmain.tsx,src%2FApp.tsx

apps/react-workshop/.ladle/global.css
```diff
@@ -3,6 +3,10 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+@layer reset, kiwi, variables, itwinui;
```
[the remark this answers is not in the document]
mayank99: I think the focus rings issue is not specific to the bridge; it could also happen if there are other global focus styles. We should add `!important` here to guard against everything:

https://github.com/iTwin/iTwinUI/blob/583e9ad05c5c70917c1fde72f037b123b3cadab7/packages/itwinui-css/src/input-container/input-container.scss#L193-L195

(Probably needs a separate changeset)

apps/react-workshop/.ladle/components.tsx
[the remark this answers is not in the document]
mayank99: This is fine. We want it to look more like Kiwi, and that includes font size changes too (in addition to color).

In the future, we'll probably want to adjust more of our spacing and sizing values to blend in with Kiwi.

apps/react-workshop/.ladle/global.css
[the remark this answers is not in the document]
mayank99: > Ohh, great, added `!important` along with changesets. ([6c2f574](https://github.com/iTwin/iTwinUI/commit/6c2f5741254e041e6155d0d35f643e21ef32c465)).

The changesets should try to be more pertinent to consumers and avoid mentioning implementation details. So it could be rephrased as something like:

> Fixed an issue in `InputWithDecorations` and `SearchBox` where the component-level focus styling was colliding with global focus styling, leading to double focus outlines.

> I'm also wondering if we can use `!important` in other places to give preference to v3 outlines over those of v5. For instance, for iui-field. This is more evident on an Input with status:
> 
> Another place is for the list items:
>
> Or instead of `!important`, maybe `kiwi` can offer a data attribute that when present does not apply the default focus styles. Then, iTwinUI can use it in these places.

`!important` should be carefully used only in cases where we want to override the global focus outline. In the case of `InputWithDecorations`, it was ok because we are doing something out of the ordinary (i.e. hiding the actual `<input>` and setting focus ring on the outer `<div>`). But in other cases, we probably actually want the Kiwi global focus indicator, just with some component-level adjustments.

I do agree there needs to be some coordination to ensure good user experience. Let's make a note of this.

apps/react-workshop/.ladle/components.tsx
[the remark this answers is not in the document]
mayank99: `<ThemeProvider as={Root}>` sounds good to me. I got some feedback from a consumer that it's working well in their app too.

apps/react-workshop/.ladle/global.css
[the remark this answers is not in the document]
mayank99: Let's defer focus outline offset from this PR's scope. I need to do some more thinking around how focus styles should be defined and modified.

.changeset/funny-singers-think.md
mayank99, line 1: Missing a changeset for the theme-bridge itself?

pnpm-lock.yaml
```diff
@@ -15960,22 +17080,21 @@ snapshots:
       - ts-node
       - utf-8-validate
 
-  '@remix-run/dev@2.15.3(@remix-run/react@2.15.3(react-dom@18.2.0(react@18.2.0))(react@18.2.0)(typescript@5.5.2))(@remix-run/serve@2.15.3(typescript@5.5.2))(@types/node@22.5.5)(lightningcss@1.29.1)(sass-embedded@1.64.1)(sass@1.72.0)(typescript@5.5.2)(vite@6.1.0(@types/node@22.5.5)(lightningcss@1.29.1)(sass-embedded@1.64.1)(sass@1.72.0)(yaml@2.5.0))':
+  '@remix-run/dev@2.8.1(@remix-run/serve@2.8.1(typescript@5.5.2))(@types/node@22.5.5)(lightningcss@1.29.1)(sass-embedded@1.64.1)(sass@1.72.0)(typescript@5.5.2)(vite@5.4.14(@types/node@22.5.5)(lightningcss@1.29.1)(sass-embedded@1.64.1)(sass@1.72.0))':
```
mayank99, line 17083: This looks unintentional. In fact, there are more changes in this file than there should be. You might want to revert this file back to `main` and then re-install and dedupe if necessary.

.changeset/few-llamas-smell.md
```diff
@@ -0,0 +1,7 @@
+---
+'@itwin/itwinui-react': minor
+---
+
+Help with migration to *future* iTwinUI versions:
+* Added a `future` prop to `ThemeProvider` to enable certain features from future versions.
+* Added a theme bridge to make v3 components look like v5 components and thus allow using v5 components in existing v3 apps. To enable, pass `{ themeBridge: true }` to `future` ([remaining setup steps and more info](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-v5-theme-bridge)).
```
mayank99, line 7: Rephrasing to frontload the mention of "theme bridge" (since that's what consumers actually care about):

```suggestion
Added a **theme bridge** which adjusts iTwinUI v3 visuals to blend in with iTwinUI v5 look-and-feel.

To enable the theme bridge, use the `future.themeBridge` flag on `<ThemeProvider>`. See [full documentation](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-v5-theme-bridge) for more detailed instructions.

**Note**: The **theme bridge** is still considered an _alpha_ feature, so there may be breaking changes until it's stabilized.
```

I don't think the `future` prop by itself is worthy of mentioning in this changeset, but you can create a separate changeset for it (if you'd like).

mayank99 (approved):
LGTM after https://github.com/iTwin/iTwinUI/pull/2373#discussion_r1949676167

PR #2374, opened 2024-12-16
Fix `HeaderButton`'s `startIcon` & `endIcon` fill color

r100-stack:
## Changes

Originated from https://github.com/iTwin/iTwinUI/pull/2369#issuecomment-2539884925.

This PR fixes the `fill` color of `HeaderButton`'s `startIcon` and `endIcon`. More specifically:
* all svgs inside the `HeaderButton` now have `fill: currentColor`. This fix applies to end icon since start icon has its own styling from `iui-header-breadcrumb-button-icon` that takes priority.
* Disabled fill with and without `forced-colors` now properly applied to start and end icon.

Unrelated: Removed the duplicate `iui-header-breadcrumb-button` class in `HeaderDropdownButton`.

## Testing

[Playground](https://stackblitz.com/edit/github-d992wxck?file=src%2FApp.tsx)

|Before|After|Diff
|-|-|-|
|image|image|image|

Forced colors:

|Before|After|Diff|
|-|-|-|
|image|image|image|

Disabled fill in start icon:

|Before|After|
|-|-|
|image|image|

## Docs

Added css and react changesets.

## After PR TODOs:

* [ ] Maybe consider better ways to target svgs inside of components since the `svg` selector is too general. ([#2374 (comment)](https://github.com/iTwin/iTwinUI/pull/2374#discussion_r1887301379)). This applies to other components too that follow this pattern.

### DIFF
--- a/packages/itwinui-css/src/header/header-buttons.scss
+++ b/packages/itwinui-css/src/header/header-buttons.scss
@@ -160,6 +160,17 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   &[aria-disabled='true'] {
     cursor: not-allowed;
   }
+
+  svg {
+    fill: currentColor;
+
+    :disabled &,
+    [aria-disabled='true'] & {
+      @media (forced-colors: active) {
+        fill: GrayText;
+      }
+    }
+  }
 }
 
 @mixin iui-header-breadcrumb-button-split {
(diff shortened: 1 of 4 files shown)


packages/itwinui-css/src/header/header-buttons.scss
```diff
@@ -160,6 +160,17 @@ $iui-side-navigation-icon-margins: calc(1.5 * var(--iui-size-m));
   &[aria-disabled='true'] {
     cursor: not-allowed;
   }
+
+  svg {
```
mayank99, line 164: Note: This is fine for now because we use this pattern in many places, but we should start thinking about how to consolidate the styling for icons to reduce duplication and make it easier to adjust.

We shouldn't target loose `svg` selectors like this; instead we should be try modify a custom property (e.g. `--_iui-svg-fill`) in the parent itself.

mayank99 (approved):
LGTM, let's include this in the next patch release.

smmr-dn (approved):

PR #2378, opened 2024-12-18
Fix broken folder variant `Tile` layout when name is long

### LINKED ISSUE #2376: Tile with `variant="folder"` does not truncate name

Tile with `variant="folder"` and long name/description does not truncate text and allows it to overflow outside tile.

### Expected Behavior

Text should be truncated or wrapped to avoid overflowing tile.

### Link to minimal repro

https://stackblitz.com/edit/github-xgunix1w?file=src%2FApp.tsx

### Steps To Reproduce

1. Open sandbox.
2. Observe that name and description is not truncated or wrapped and overflows tile.

### Package Version

- [x] I'm using the latest version of iTwinUI.

### Anything else?

_No response_

r100-stack:
## Changes

Fixes #2376

Adds `min-inline-size: 0` to `.iui-tile-name` to prevent the Tile folder variant UI from breaking when there's a long name label.

## Testing

* CI still passing
* Tested some cases similar to the user's sandbox in a separate [sandbox](https://stackblitz.com/edit/github-xgunix1w-bjkxmbki?file=package.json).

|Before|After|
|-|-|
|image|image|

## Docs

Added changesets

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -89,10 +89,10 @@ importers:
         version: 0.20.0
       astro:
         specifier: ~4.16.3
-        version: 4.16.3(@types/node@22.5.5)(lightningcss@1.25.1)(rollup@4.22.4)(sass-embedded@1.64.1)(sass@1.72.0)(typescript@5.5.2)
+        version: 4.16.18(@types/node@22.5.5)(lightningcss@1.25.1)(rollup@4.28.1)(sass-embedded@1.64.1)(sass@1.72.0)(typescript@5.5.2)
       astro-relative-links:
         specifier: ^0.3.7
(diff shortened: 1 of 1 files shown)


mayank99 (approved):
Lets make a patch release right after merging this.

smmr-dn (approved):

HaveSpacesuit:
I have noticed something similar on regular tiles with long names - do you think this will affect those as well?

HaveSpacesuit:
Here is an example. If the name does not contain spaces, it breaks out of the tile. We work around this by using MiddleTextTruncation, though I'm not sure that is explicitly supported for use by package consumers.

https://stackblitz.com/edit/github-h7qvdgke?file=src%2FApp.tsx

image

mayank99:
@HaveSpacesuit This is definitely something we should fix. Could you please open a new issue? This PR is not the best place for this discussion.

PR #2381, opened 2024-12-19
Fixed site documentation layout 

smmr-dn:
## Changes

The `display: block !important` caused unexpected behaviors in live demo block in terms of width and height for some components. Especially, the `Tag` component got stretched out and ended up overlapping with the content table. 

This PR addressed another workaround, following Mayank's [PR](https://github.com/iTwin/iTwinUI/pull/2353), to temporarily fix the problem caused by VoiceOver not working in Safari. Currently, Safari 18.3 [beta version](https://developer.apple.com/documentation/safari-release-notes/safari-18_3-release-notes) has been released, and this bug was marked as major so hopefully, we can remove this workaround soon.

## Testing

Confirmed VoiceOver is still working in Safari.
voiceover

## Docs

N/A


apps/website/src/components/LiveExample.astro
```diff
@@ -49,7 +49,10 @@ const cssCode = await (async () => {
   <template shadowrootmode='open'>
     <style is:inline>
       astro-island {
-        display: block !important; /* https://github.com/iTwin/iTwinUI/pull/2353 */
+        display: flex !important;
```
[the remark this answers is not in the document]
smmr-dn: I have re-added the comment.

apps/website/src/components/LiveExample.astro
```diff
@@ -49,7 +49,10 @@ const cssCode = await (async () => {
...
+        width: 650px;
```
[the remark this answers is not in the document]
smmr-dn: This works as well. I have updated the PR.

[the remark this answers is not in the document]
smmr-dn: My bad, I have updated the PR, I didn't fully push it :(

apps/website/src/components/LiveExample.astro
mayank99, line 1: Can you please explain the thought process behind the specific changes you've made? 

Previously, it was clear to me that `display: block !important` was supposed to override the default `display: contents` (which causes the Safari issue). Whereas now, it's unclear why we are using `display: flex` with these specific set of properties.

It's also not clear why the examples were changed.

smmr-dn: At first, I tried to keep the `block` layout and fixed the width since the broken demos were all shrunk down to zero width (e.g., `Tree`, `Table` or `Tabs`), while the `Tag` doc stretched all the way to the right.

image

I tried `min-inline-size: 0`, to enforce a minimum size of the content's width where `astro-island` could shrink down to, `width: 100%` for `astro-island` to take up the whole width of its parent, and `place-items: center` to center the content. However, some of the elements that are not wrapped around a `<div>`, such as `<a>`, `<button>`, or `<span>` got aligned with the left corner of the `astro-island` since they took the full width of the `astro-island` container without getting centered based on their intrinsic size, which resulted in them being shifted left.
 
image


To fix this problem, I tried another layout to help with centering and width at the same time, which is `flex` in this PR. I had a fixed width for the content at first then Rohan suggested I should do `min-inline-size: 0` and `inline-size: 100%` so that the contents can be preserved without being forced in a fixed width. `justify-content` and `align-items` are for centering. The contents `ComboBox` label and `Surface` elevation examples were wrapped by fragments `<>`, so I added the `<div>` tag to make them flex items so they can be styled correctly. For the `Tree` virtualization example, I specified the height so it can appear within the container and be scrolled correctly.

mayank99: I'm also trying to understand why `display: block` breaks the layout when it previously worked with `display: contents`. Were you able to figure out why exactly it's breaking suddenly? Ideally we wouldn't need any extra CSS at all, especially when the content is not even overflowing (e.g. the tags themselves fit within the box just fine).

smmr-dn: I notice the latest release note of [Safari](https://developer.apple.com/documentation/safari-release-notes/safari-18_3-release-notes) might include the fix for the bug that we're trying to workaround. The changeset under `Accessibility` says: "Fixed VoiceOver to not announce the content of `<script>` elements inside containers with `display: contents`. (141174582)"

If that's the case, should we test it out and close the PR?

mayank99: Yeah, I feel uneasy about this PR, so let's close it. Between the Safari fix and #2415, we don't need this PR.

#2415 explains why the issue was happening. It's a long tag container further down the page. Similarly, in other pages, there is some long element somewhere in the page (not necessarily in the first demo).

smmr-dn:
Closing this PR as it is no longer applicable (https://github.com/iTwin/iTwinUI/pull/2381#discussion_r1929392376)

smmr-dn:
Closing this PR as it is no longer applicable (https://github.com/iTwin/iTwinUI/pull/2381#discussion_r1929392376)

PR #2384, opened 2024-12-23
Remove leading spaces in peer-deps for react deps

Antanas-Bentley:
## Changes

Hi all, with Module Federation coming into use more and more we have fallen quite a few times into this pitfall of declaring peerDep version ranges with extra spaces.

The version resolution logic is quite complex inside MF sources and it does not seem to handle having extra spaces in the version range. We want MF to include shared bundle of react and react-dom as the peerDep for iTwinUI and would mean the peerDep has to be resolved with a "correct" version range.

## NPM Docs

Set of primitive operators are all declared without spaces: https://docs.npmjs.com/cli/v6/using-npm/semver#ranges

## Testing

I am using a private repository to test this but the results are quite appearant:
image
image

Without spaces in the version range:
image
image

Bundles are shared like:
image

## Docs

N/A

### DIFF
--- a/.changeset/large-poets-begin.md
+++ b/.changeset/large-poets-begin.md
@@ -1,0 +1,5 @@
+---
+'@itwin/itwinui-react': patch
+---
+
+Fixed the syntax for `react` version range declared in `peerDependencies`.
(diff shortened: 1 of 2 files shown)


mayank99 (approved):
Nice find, thanks for the PR!

PR #2390, opened 2025-01-06
Fix `ThemeProvider` bug of remounting and losing state

r100-stack:
## Changes

A user pointed out a bug where `ThemeProvider` used to reset the state of its `children` when `inherit` was toggled between `undefined` and defined. ([playground](https://stackblitz.com/edit/github-wqsdfm?file=src%2FApp.tsx,index.html))

Fixes the regression introduced in [3.14.0](https://github.com/iTwin/iTwinUI/releases/tag/%40itwin%2Fitwinui-react%403.14.0) by #2153. This PR also now passes the e2e test in https://github.com/iTwin/appui/pull/1133 meant to confirm that this regression is fixed.

---

The reason for the reset of the state was because of the following code:

https://github.com/iTwin/iTwinUI/blob/b8d5fa22b2e705104493373ce4d4647479882411/packages/itwinui-react/src/core/ThemeProvider/ThemeProvider.tsx#L180-L182

https://github.com/iTwin/iTwinUI/blob/b8d5fa22b2e705104493373ce4d4647479882411/packages/itwinui-react/src/core/Toast/Toaster.tsx#L107-L119

Here, when `portalContainer`/`portalContainerProp` changes to undefined, the `inherit` changes from `false` to `true`. As a result, `ToastProvider` now returned `children` wrapper with `ToasterDispatchContext.Provider` and `ToasterStateContext.Provider` instead of just `children`. When the UI tree position changes, the state is reset ([docs](https://react.dev/learn/preserving-and-resetting-state#state-is-tied-to-a-position-in-the-tree)).

The fix is to thus keep the UI tree position of `children` the same regardless of `inherit`.

## Testing

- [x] Confirm that it works in the [playground](https://stackblitz.com/edit/github-wqsdfm?file=src%2FApp.tsx,index.html).
- [ ] Confirm that the AppUI [e2e test](https://github.com/iTwin/appui/pull/1133) passes. (Will be doing it after the PR since @GerardasB said that if the playground passes, the test will also likely pass)

Not sure why some react-workshop Table test images failed. But since they actually look more correct than the old images (i.e. correct element is focused), I approved them for now.

## Docs

Added changeset

### DIFF
--- a/apps/website/src/content/docs/tree.mdx
+++ b/apps/website/src/content/docs/tree.mdx
@@ -65,7 +65,7 @@ The `getNode` function then needs to map the user `data` to a `NodeData`. The pr
 const getNode = React.useCallback(
   (node) => {
     return {
-      /*…*/
+      /* … */
       subNodes: node.subItems,
       hasSubNodes: node.subItems.length > 0,
     };
@@ -93,7 +93,7 @@ The `isExpanded` flag which indicates whether the node is expanded to display it
 const getNode = React.useCallback(
   (node) => {
     return {
-      /*…*/
+      /* … */
       isExpanded: expandedNodes[node.id],
     };
   },
--- a/examples/Tree.virtualization.css
+++ b/examples/Tree.virtualization.css
@@ -1,5 +1,4 @@
 .demo-tree {
-  width: min(100%, 260px);
-  height: 100%;
+  width: 260px;
   overflow: auto;
 }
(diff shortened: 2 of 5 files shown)


mayank99 (approved):
I don't love that we are now introducing two additional contexts (which are arguably unnecessary), but it solves the problem. Good fix 👍

smmr-dn (approved):

PR #1835, opened 2024-02-07
Fix invisible `Dialog`/`Modal` on first render with `isOpen={true}`

rohan-kadkol:
## Changes

Fixes #1829

After quite a bit of investigation, I exhausted all potential issues in our code but found no fix.

Upon investigating `react-transition-group` itself, I realized that `4.4.2` instead of `4.4.5` fixed the issue. The potential problematic release on their end could be [reactjs/react-transition-group@v4.4.3 (release)](https://redirect.github.com/reactjs/react-transition-group/releases/tag/v4.4.3).

So, this PR:
* ~Uses react-transition-group `4.4.2` instead of `4.4.5`~ Makes no package.json or yarn.lock changes ([#1835 (comment)](https://github.com/iTwin/iTwinUI/pull/1835#discussion_r1483642005)).
* Adds `.iui-dialog-visible` depending on `isOpen` and independant of the animation state
  * This reverts some of the changes from https://github.com/iTwin/iTwinUI/pull/1776/files#diff-1dc61dd4610c277806587eefb3891b0a84430c2547781f57f452e49ea1f37e20
* Replaces dialog/modal animation state related classes (`.iui-dialog-animation-enter` and `.iui-dialog-animation-enter-active`) with css animations.

## Testing

Confirmed that the following work:
  1. enter animation in the Modal → Full Page Modal story
  2. Dialog appears in the repro sandboxes in [#1829](https://github.com/iTwin/iTwinUI/issues/1829)
  3. Modal appears in the repro sandboxes in https://github.com/iTwin/iTwinUI/issues/1770#issuecomment-1927580282
  4. All animations on the css-workshop demo pages still work

<details>
<summary>Screen recordings</summary>

### 1.

Before:

https://github.com/iTwin/iTwinUI/assets/45748283/3a4ce7ad-5e42-4888-9f80-d1fa8642e1d3

After:

Same as before (i.e. still working)

### 2.

Before:

https://github.com/iTwin/iTwinUI/assets/45748283/15655ab6-86c7-4408-ac1e-e2ef80877c62

After:

https://github.com/iTwin/iTwinUI/assets/45748283/e8d8f807-f85e-4f96-8e11-00bd604b8e27

### 3.

Before:

image

After:

image

</details>

## Docs

Added changesets

PR #2391, opened 2025-01-06
Remove `react-transition-group` dependency

### LINKED ISSUE #2199: findDOMNode deprecation warning with newest react 18 version

With newest React 18 version when certain iTwinUi-react components are used and app is in strict mode, an error is logged into the console: findDOMNode is deprecated in StrictMode.

The error is coming from 'react-transition-group' package that is used in WithCSSTransition component. As I understand according to issue https://github.com/reactjs/react-transition-group/issues/904 this can be fixed on iTwinUI side.

### Expected Behavior

_No response_

### Link to minimal repro

_No response_

### Steps To Reproduce

1. Create an app that is using React 18 and is wrapped in strict mode.
2. Use one of the affected components, e.g. WithCSSTransition 
3. You should be seeing the error in browser console when affected component is rendered

### Anything else?

_No response_

r100-stack:
## Changes

Prerequisite for #2387. Fixes #2199.

This PR removes the dependency to the now mostly unmaintained [`react-transition-group`](https://github.com/reactjs/react-transition-group). This has been our plan for a while now because of running into issues with it and having to find workarounds (e.g. #1835). Since React 19 has now removed the deprecated `findDOMNode()` that `react-transition-group` moving away from this package is needed to support React 19 in #2387.

Affected components:
* Dialog/Modal: Exit animation removed temporarily since it's difficult to implement cleanly in JS.
* Toast: Used the [`animate()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/animate) function to implement the same animation in JS.

---

There were some animations that used to work in previous v3 versions and also in v2. Some of these animations stopped working in the previous few versions before this PR. So, I just removed the code for those animations since they anyways weren't working even before this PR. Some of these animations include:
* <details><summary>Table with subcomponent</summary>

  |v2.12.26|v3.16.0|
  |-|-|
  |<video src="https://github.com/user-attachments/assets/efde123f-e7b8-48a3-89a3-43e748b2aba8" />|<video src="https://github.com/user-attachments/assets/cfa055fe-5aa8-462d-8c41-24840325b386" />|

  </details> 

* <details><summary>SideNavigation with submenu</summary>

  |v2.12.26|v3.16.0|
  |-|-|
  |<video src="https://github.com/user-attachments/assets/6eafdcd3-90c3-4bbc-87c2-adacfd69d555" />|<video src="https://github.com/user-attachments/assets/6fc300df-3e51-439d-8fad-be8933fb7ec5" />|

  </details>

This PR also deprecates the `react-transition-group` related CSS classes: `.iui-enter`, `.iui-enter-active`, `.iui-exit`, `.iui-exit-active` since they are no longer needed.

## Testing

Confirmed whenever react-transition-group was used or at least was intended to be used, there is no visual change after this PR. The only exception is where we purposefully removed the animation temporarily (Dialog/Modal exit animation).

## Docs

Added *minor* react and css changesets.

It is minor for css because of class deprecations. It is a minor for react since `react-transition-group` could be a dependency used in users' apps too and so could affect their dependency resolution. An additional reason is that the `Dialog`/`Modal`'s exit animation is removed temporarily.

## After PR TODO:

* [ ] Try to avoid the effect by moving the sync based on visibility logic to where the visibility changes ([#2391 (comment)](https://github.com/iTwin/iTwinUI/pull/2391#discussion_r1924248045)).
* [x] Confirm if there is some flakiness in the Toasts react tests ([#2312 (review)](https://github.com/iTwin/iTwinUI/pull/2312#pullrequestreview-2582637920)).

### DIFF
--- a/packages/itwinui-react/src/core/Dialog/DialogMain.tsx
+++ b/packages/itwinui-react/src/core/Dialog/DialogMain.tsx
@@ -161,10 +161,13 @@ export const DialogMain = React.forwardRef((props, ref) => {
     }));
   }, []);
 
+  /** Focuses dialog when opened. */
   const onEnter = React.useCallback(() => {
     previousFocusedElement.current = dialogRef.current?.ownerDocument
       .activeElement as HTMLElement;
-    setFocus && dialogRef.current?.focus({ preventScroll: true });
+    if (setFocus) {
+      dialogRef.current?.focus({ preventScroll: true });
+    }
   }, [dialogRef, previousFocusedElement, setFocus]);
 
   /** Brings back focus to the previously focused element when closed. */
@@ -180,7 +183,6 @@ export const DialogMain = React.forwardRef((props, ref) => {
 
   const mountRef = React.useCallback(
     (element: HTMLElement | null) => {
-      /** Focuses dialog when opened. */
       if (element) {
         onEnter();
       }
@@ -235,9 +237,7 @@ export const DialogMain = React.forwardRef((props, ref) => {
 
   return (
     <DialogMainContext.Provider
-      value={{
-        beforeClose,
-      }}
+      value={React.useMemo(() => ({ beforeClose }), [beforeClose])}
     >
       <DialogDragContext.Provider value={{ onPointerDown: handlePointerDown }}>
         {trapFocus && <FocusTrap>{content}</FocusTrap>}
--- a/packages/itwinui-react/src/core/Dialog/DialogTitleBar.tsx
+++ b/packages/itwinui-react/src/core/Dialog/DialogTitleBar.tsx
@@ -59,10 +59,13 @@ export const DialogTitleBar = Object.assign(
 
     const { onPointerDown } = useDialogDragContext();
 
-    const onClick = React.useCallback(() => {
-      dialogMainContext?.beforeClose();
-      onClose?.();
-    }, [dialogMainContext, onClose]);
+    const onClick = React.useCallback(
+      (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        dialogMainContext?.beforeClose();
+        onClose?.(e);
+      },
+      [dialogMainContext, onClose],
+    );
 
     return (
       <Box
--- a/.changeset/thick-melons-explain.md
+++ b/.changeset/thick-melons-explain.md
@@ -1,5 +1,0 @@
----
-'@itwin/itwinui-react': minor
----
-
-The exit animation on `Dialog` and `Modal` has been temporarily removed.
--- a/.changeset/tidy-geckos-check.md
+++ b/.changeset/tidy-geckos-check.md
@@ -2,4 +2,6 @@
 '@itwin/itwinui-react': minor
 ---
 
-Removed dependency on `react-transition-group`.
+Removed dependency on `react-transition-group`. Notable changes in components: 
+* `useToaster`: Animations have been reworked to directly use the web animations API.
+* `Dialog` and `Modal`: Exit animations have been temporarily removed.


packages/itwinui-react/src/core/Toast/Toast.tsx
```diff
@@ -172,73 +171,135 @@ export const Toast = (props: ToastProps) => {
...
+  }, [
+    animateIn,
+    animateOut,
+    animateOutTo,
+    calculateOutAnimation,
+    isVisible,
+    motionOk,
+    onRemove,
+    prevIsVisible,
+  ]);
```
mayank99, line 277: First of all, we should not be using an Effect here. We should animate in the same place where the visible state changes.

However, it might be difficult to untangle the mess that exists, so we could keep the Effect as a (hopefully temporary) hack. We should still reduce the number of dependencies in this array, by moving some of them inside Effect, and making some of them `useLatestRef` if we don't need their reactive values. Also, it would be good to move this into a custom hook.

packages/itwinui-react/src/core/Toast/Toaster.test.tsx
```diff
@@ -19,6 +19,15 @@ import {
...
 const mockOnClick = vi.fn();
 
+// Set reduced motion to not trigger animations since JSDom doesn't mock `animate()`.
+const originalMatchMedia = window.matchMedia;
+beforeAll(() => {
+  window.matchMedia = vi.fn().mockReturnValue({ matches: false });
+});
+afterAll(() => {
+  window.matchMedia = originalMatchMedia;
+});
```
mayank99, line 29: Can we use `animate?.()` to get around this issue? Otherwise our consumers will also need to add this mock.

packages/itwinui-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -203,36 +202,53 @@ export const DialogMain = React.forwardRef((props, ref) => {
...
+      setShouldBeMounted(true);
+    } else {
+      onExit();
+
+      // Wait for the animation to end
+      setTimeout(() => {
+        setShouldBeMounted(false);
+      }, 600);
+    }
+  }
```
mayank99, line 244: Can we use a callback ref to get precise timing of when the element mounts/unmounts?

e.g. something like this:

```jsx
ref={
  React.useCallback((element) => {
    if (element) {
      onEnter();
    } else {
      onExit();
    }
  }, [])
}
```

packages/itwinui-react/src/core/Dialog/Dialog.tsx
```diff
@@ -42,37 +41,50 @@ const DialogComponent = React.forwardRef((props, ref) => {
...
+    if (isOpen) {
+      setShouldBeMounted(true);
+    }
+    // Wait for DialogMain to receive the isOpen=false to properly handle its exit. E.g. move focus to trigger.
+    else {
+      setTimeout(() => {
+        setShouldBeMounted(false);
+      }, 600);
+    }
+  }, [isOpen]);
```
mayank99, line 58: Hmm, I don't think we should have this additional state. `isOpen` can be used as the source of truth if we call `onExit()` after the animation ends.

apps/react-workshop/cypress-visual-screenshots/baseline/Toasts.test.ts-Informational.png
mayank99, line 1: Why did these images change? Looks like there is a slight change in positioning, which I think might be not intentional.

.changeset/rude-books-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@itwin/itwinui-css': minor
+---
+
+Animation related classes are made now deprecated: `.iui-enter`, `.iui-enter-active`, `.iui-exit`, `.iui-exit-active`.
```
mayank99, line 5: typo? "made now deprecated" → "now made deprecated"

packages/itwinui-react/src/core/Dialog/Dialog.tsx
mayank99: Sorry I meant to say `onClose()`, not `onExit()`.

Basically, we need to ensure that `isOpen` will be true when the entry animation starts and remain true until until the exit animation ends.

This means we need to 
1. Animate out the dialog when a "close trigger" is activated (i.e. "❌" button or Esc key).
2. After the animation, call `onClose()` (probably in `onExit()`) for removing it from DOM.

If this requires large code changes, let's remove the animation code from this PR and re-introduce it in a future PR. In the meantime, the dialog entry animation should still work, which is good.

packages/itwinui-react/src/core/Toast/Toast.tsx
mayank99: What I mean is, we should execute this code next to the place where we are updating the state, rather than using an Effect to create a chained computation based on the state value. It's similar to the idea of [moving logic out of Effects into event handlers](https://react.dev/learn/separating-events-from-effects).

For the custom hook, I have a few points of feedback:
- The `shouldBeMounted` state can be moved into the hook and returned by it, so that `setShouldBeMounted` does not need to be passed into it.
- `useLatestRef` might be overused in this scenario. We should be more thoughtful about it. `useLatestRef` is most useful for frequently-changing unmemoized dependencies.
  - `thisElement` is already ref and does not need additional `useLatestRef`.
  - `motionOk` is a simple boolean value that likely won't change often, so doesn't need additional `useLatestRef`.
  - `calculateOutAnimation` could be moved inside the Effect so that it doesn't need `useLatestRef`.
- readability: When accepting lots of parameters, use an object instead of positioned arguments.
  - Depending on your preference, all of the parameters can be passed into a single argument, or you can have two arguments: first one would be `isVisible` and second one would be the object containing the rest of the information.
- readability: When defining multiple functions inside an Effect, consider using a hoisted `function` (instead of arrow function) located below the main Effect logic, so that it's easier to read.

mayank99: I think you want to use `animteOutToRef` inside the Effect and remove `animateOutTo` from the dependency array

mayank99: It looks like you did the opposite of what I suggested, but it's ok. `animateOutTo` is an `HTMLElement`, so it should remain stable for the common use-case (animating toasts into the notification bar).

packages/itwinui-react/src/core/Dialog/Dialog.tsx
mayank99: > I think only `isOpen` is what decides whether a dialog is open or not. Because if users pass `isOpen={false}` and don't update `isOpen` in the `onClose` that they pass to it, the dialog doesn't close even if the "❌" button is clicked.

This should be considered a consumption error. The `onClose()` exists for a reason and should not be ignored. I think #1976 would help avoid such issues in the future, but for now it should be ok to assume proper usage of `onClose()` + `isOpen`.

> Here, by the time onExit's logic is executed, `dialogRef.current` becomes `null`. If not having the exit animations, we'd need to remove that if statement.

We could keep the if-statement but move it before we call `onClose()`. This should ensure that `dialogRef.current` is still available, because `isOpen` will not be `false` until the next render. If we are calling `onClose()` in multiple places, it might be useful to create a wrapper over it, named e.g. `requestClose()`.

> Also, I'm wondering, should we check for `setFocus` in `onExit` similar to how we do it in onEntered?

I don't think so. We never want a situation where the end user's focus gets lost after closing the dialog. Additionally, `setFocus` should default to `true` (#1998), so that consumers don't have to think about it.

mayank99: If `requestClose` is only needed by `Dialog.Main` and its descendants, we can create a different context in `DialogMain`.

It's also confusing to store the `requestClose` in an `onClose` field in the root context. It should be renamed or its value should be reverted back to `props.onClose`.

packages/itwinui-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -204,34 +234,16 @@ export const DialogMain = React.forwardRef((props, ref) => {
...
-          dialogRef.current?.contains(
-            dialogRef.current?.ownerDocument.activeElement,
-          )
-        ) {
-          previousFocusedElement.current?.focus();
-        }
+    <DialogMainContext.Provider
+      value={{
+        beforeClose,
       }}
```
mayank99, line 240: nit 1: Context values should be memoized.

```suggestion
      value={React.useMemo(() => ({ beforeClose }), [beforeClose])}
```

(check formatting when committing)

packages/itwinui-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -158,6 +161,33 @@ export const DialogMain = React.forwardRef((props, ref) => {
...
+
+  const mountRef = React.useCallback(
+    (element: HTMLElement | null) => {
+      /** Focuses dialog when opened. */
+      if (element) {
+        onEnter();
+      }
+    },
+    [onEnter],
+  );
```
mayank99, line 189: nit 2: This can be made more readable by moving the comment and using an if-condition.

```suggestion
  /** Focuses dialog when opened. */
  const onEnter = React.useCallback(() => {
    previousFocusedElement.current = dialogRef.current?.ownerDocument
      .activeElement as HTMLElement;
    if (setFocus) {
      dialogRef.current?.focus({ preventScroll: true });
    }
  }, [dialogRef, previousFocusedElement, setFocus]);

  /** Brings back focus to the previously focused element when closed. */
  const beforeClose = React.useCallback(() => {
    if (
      dialogRef.current?.contains(
        dialogRef.current?.ownerDocument.activeElement,
      )
    ) {
      previousFocusedElement.current?.focus();
    }
  }, [dialogRef, previousFocusedElement]);

  const mountRef = React.useCallback(
    (element: HTMLElement | null) => {
      if (element) {
        onEnter();
      }
    },
    [onEnter],
  );
```

packages/itwinui-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -56,6 +59,11 @@ export const DialogTitleBar = Object.assign(
 
     const { onPointerDown } = useDialogDragContext();
 
+    const onClick = React.useCallback(() => {
+      dialogMainContext?.beforeClose();
+      onClose?.();
```
mayank99, line 64: We were previously passing the click event into `onClose`.

packages/itwinui-react/src/core/Toast/Toast.tsx
mayank99, line 1: Should we have a changeset for Toast? Maybe can be combined with the other changesets.

e.g. Something like:

> Removed dependency on `react-transition-group`. Notable changes in components:
>  - `useToaster`: Animations have been reworked to directly use the web animations API.
>  - `Dialog` and `Modal`: Exit animations have been temporarily removed.

mayank99 (approved):

smmr-dn (approved):

PR #2393, opened 2025-01-06
Rearranged `Typography` doc

smmr-dn:
## Changes

This PR split the `Typography` doc into two separate `Text` and `Label` docs, suggested by Mayank's [suggestion](https://github.com/iTwin/iTwinUI/pull/2324#discussion_r1896853507). Also, the `Typography` is hidden from the sidebar as well as the component page.

## Testing

N/A

## Docs

N/A

## After-PR TODOs:

- [ ] Come up with a better way to exclude and include pages without having to hardcode (https://github.com/iTwin/iTwinUI/pull/2393#discussion_r1904729159)
- [ ] Fix layout breaking when no section title is present (https://github.com/iTwin/iTwinUI/pull/2393#discussion_r1906068898)


apps/website/src/content/docs/typography.mdx
```diff
@@ -2,95 +2,7 @@
 title: Typography
 description: Typography components provide consistency in text styling across the application.
 thumbnail: Typography
+related:
+  - label
+  - text
```
mayank99, line 7: This looks weird. We should just add regular content to the page instead of using related links.

apps/website/src/components/LeftSidebarContent.astro
```diff
@@ -7,7 +7,10 @@ const currentPage = Astro.url.pathname.endsWith('/')
   : Astro.url.pathname;
 
 const introPages = await getCollection('docs', ({ data: { group } }) => group === 'intro');
-const mainComponents = await getCollection('docs', ({ data: { group } }) => group === undefined);
+const mainComponents = await getCollection(
+  'docs',
+  ({ data: { group, title } }) => group === undefined && title !== 'Typography',
```
mayank99, line 12: non-blocking/future: We need to come up a better system for including and excluding pages from the sidebar so that we are not having to hardcode it like this.

smmr-dn: I have added this as an after-PR todo.

apps/website/src/content/docs/typography.mdx
smmr-dn: I have added more content to the page and removed the related links.

apps/website/src/content/docs/typography.mdx
```diff
@@ -6,91 +6,22 @@ thumbnail: Typography
...
-## Text
+## Components
 
-`Text` is a [polymorphic](#polymorphic-only-semantics-and-not-size) typography component that can be used to render any kind of text as any kind of element.
+Typography components consist of two main components:
 
-```jsx
-<Text>This is a text.</Text>
-```
+### `Text`
```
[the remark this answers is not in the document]
smmr-dn: I have removed those back ticks.

apps/website/src/content/docs/typography.mdx
mayank99: I was just talking about moving the links into the main content; this page doesn't need any actual documentation or demos, only links.

Something like this (with links) should be enough:

> The documentation for typography components has been split into the following pages:
> - Label
> - Text

mayank99 (approved):
LGTM after https://github.com/iTwin/iTwinUI/pull/2393#discussion_r1905989884

apps/website/src/content/docs/typography.mdx
smmr-dn: I have made the change.

mayank99: This is something we might want to fix in the future and can be added as an after-PR TODO

apps/website/src/content/docs/text.mdx
```diff
@@ -0,0 +1,65 @@
+---
+title: Text
+description: Text components offer various text formatting options for applications.
+thumbnail: #TODO
```
mayank99, line 4: Typography is the correct thumbnail for this

```suggestion
thumbnail: Typography
```

mayank99:
This PR has broken some links (well, not completely "broken" because we still kept the fallback page, which is good). I found `anchor.mdx`, `combobox.mdx` and `inputgrid.mdx` referencing the older pages; there might be more that I haven't found.



@smmr-dn Can you open another PR to fix it?

PR #2408, opened 2025-01-21
Release packages

imodeljs-admin:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   [#2407](https://github.com/iTwin/iTwinUI/pull/2407): Made the light theme's text selection background opacity more consistent with that of the OS.

### Patch Changes

-   [#2407](https://github.com/iTwin/iTwinUI/pull/2407): Slightly increased `--iui-opacity-5` in the light theme to make its text selection background opacity more consistent with that of the OS.


packages/itwinui-react/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 3.16.6
+
+### Patch Changes
+
+- [#2411](https://github.com/iTwin/iTwinUI/pull/2411): Fixed `Panels.Trigger` not working in older React versions where `useId` is not available.
+- [#2407](https://github.com/iTwin/iTwinUI/pull/2407): Made the light theme's text selection background opacity more consistent with that of the OS. This also affects some other opacities of other components (e.g. shadows and hover backgrounds).
```
mayank99, line 8: This is a nit that I should have mentioned in #2407, but the main reason for increasing the opacity is *not* to make it consistent with "the OS"; it's solely to increase contrast.

That it more closely matches macOS in the default accent is a good sign, but is largely irrelevant and does not need to be mentioned, especially since most end users are not going to be on macOS.

mayank99 (approved):

